PR #11, opened 2019-11-21
IOX #10 Fix payloadsize assignment when reusing last chunk

### LINKED ISSUE #10: Payload size not updated by allocateChunk(WithInfo)
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.4.0

When allocating a chunk with ```allocateChunk``` or ```allocateChunkWithInfo``` using the dynamic payload option (```useDynamicPayloadSizes = true```) the ```ChunkInfo``` field ```m_payloadSize``` is initialized correctly by the first call. Subsequent calls of the same function with an increased ```payloadSize``` do not update the member variable.

```ChunkInfo::m_payloadSize``` should reflect the current payload size under all conditions.

```cpp
// publisher side
auto ci = m_publisher->allocateChunkWithInfo(42, true);
m_publisher->sendChunkWithInfo(ci);

auto ci = m_publisher->allocateChunkWithInfo(42+1, true);
m_publisher->sendChunkWithInfo(ci);

...

// subscriber side
const iox::mepoo::ChunkInfo* ci(nullptr);
while (m_subscriber->getChunkWithInfo(&ci))
{
  std::cout << ci->m_payloadSize << std::endl;
  m_subscriber->releaseChunkWithInfo(ci);
}
```

mossmaurice:
Fixes #10


elBoberido:
Please fix the line endings in the PR. The SenderPort is shown as completely changed when whitespaces are not ignored. This screws up the history.

Regarding the changes:
If you have a few slow sender and a fast receiver, then the sender will always recycle the chunk for the highest payload from his history.

Let's assume we have some applications which initially sends data with a 1GB payloads and afterwards only 1kB payloads.
Let's further assume we have an application which always sends 1GB payloads, but is ensured to run after the other applications which initially used the 1GB payload and are now on 1kB payloads.
Now you wast memory which can perfectly be used by the application which needs it, but is occupied by applications which sends 1kB payloads and the user has no chance to fix it.

mossmaurice:
Thanks @elBoberido for your comments. 

Regarding the line ending change: All files in this repo have a LF (Linux) line ending. However sender_port.cpp had CRLF (Windows). So I thought I'd directly fix this to be consistent. You can ignore whitespace changes in the web GUI, by selecting the cogwheel -> "Hide whitespace changes" in the tab "Files changed".

Regarding "==" instead of ">=": I can definitely see your point. We have to think about the contract with the user. Most likely a user would not expect this behaviour. However is sending 1GB and afterwards only a few kB a real-world use case? What do you think @michael-poehnl & @elfenpiff ?

michael-poehnl:
Currently it is a runtime optimization as we don't need to acquire a new chunk . Always using the smallest possible chunk is the memory optimization. Both are somehow valid.  

The main use case are senders with constant payload. For them it's a nice feature. But maybe we should limit the reuse to the case where the same chunk size would be used again. But then I would also say that we extend the ChunkInfo with the necessary information to have the full chunk size not only the size used by the last chunk

elBoberido:
Do we know how expensive it is to get a new chunk? A long long time ago there was a mutex involved, but not anymore, so this might be cheap now.

PR #28, opened 2019-12-17
Minor cleanups

rotu:
1. Make subscriber thread name shorter to prevent truncation of format string
2. Replace use of ushort, uint with more explicit types
3. Use default destructors where appropriate
4. Use std::cerr instead of fprintf
5. Make ACL enum classes extend actual underlying numeric types
6. Use std::remove_pointer<acl_t>:: type instead of implementation-dependent __acl_ext
7. Switch instructions from apt-get to the more friendly apt

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
+++ b/iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
@@ -201,7 +201,7 @@ inline const T& vector<T, Capacity>::at(const uint64_t index) const
 {
     if (index + 1 > m_size)
     {
-        fprintf(stderr, "out of bounds access, current size is %lu but given index is %lu\n", m_size, index);
+        std::cerr << "out of bounds access, current size is " << m_size <<  " but given index is " << index << std::endl;
         std::terminate();
     }
     return reinterpret_cast<const T*>(m_data)[index];
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
@@ -44,7 +44,7 @@ class AccessController
     static constexpr int32_t MaxNumOfPermissions = 20;
 
     /// @brief identifier for a permission entry (user, group, others, ...)
-    enum class Category
+    enum class Category : acl_tag_t
     {
         USER = ACL_USER_OBJ,
         /// a specific user must be identified by a name
@@ -56,11 +56,11 @@ class AccessController
     };
 
     /// @brief access right for a permission entry
-    enum class Permission
+    enum class Permission : acl_perm_t
     {
         READ = ACL_READ,
         WRITE = ACL_WRITE,
-        READWRITE = ACL_READ | ACL_WRITE,
+        READWRITE = Permission::READ | Permission::WRITE,
         NONE = 0
     };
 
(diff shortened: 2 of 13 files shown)


iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -201,7 +201,7 @@ inline const T& vector<T, Capacity>::at(const uint64_t index) const
 {
     if (index + 1 > m_size)
     {
-        fprintf(stderr, "out of bounds access, current size is %lu but given index is %lu\n", m_size, index);
+        std::cerr << "out of bounds access, current size is " << m_size <<  " but given index is " << index << std::endl;
```
michael-poehnl: @elfenpiff What was the reason for chosing fprintf here?

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
```diff
@@ -56,11 +56,11 @@ class AccessController
     };
 
     /// @brief access right for a permission entry
-    enum class Permission
+    enum class Permission : acl_perm_t
     {
         READ = ACL_READ,
         WRITE = ACL_WRITE,
```
michael-poehnl: @rotu Then we should also replace READ and WRITE with Permission::READ and Permission::WRITE ?

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
Karsten1987: might be nit, but within the ROS2 context we generally preferred `fprintf` over `cerr/cout` as it is primarily threadsafe and doesn't produce interleaved output when running concurrently.

Karsten1987:
I guess this has to go through some internal CI, but this PR generally looks good to me.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
rotu: The problem with fprintf here is that index is a `uint64_t` and is not guaranteed to be an `unsigned long`. Indeed on my Mac, it's an `unsigned long long` and generates a build warning.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
rotu: I don't totally understand what you mean.

In `READWRITE = Permission::READ | Permission::WRITE`, the `Permission::` is redundant, but I added it to visually distinguish where we're pulling in platform header constants `ACL_READ` from where we're defining `READWRITE` as a composite of other bit flags.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
Karsten1987: the right specifier would be `PRIu64` in this case. See here:
http://www.cplusplus.com/reference/cinttypes/

so in this case:
```c++
fprintf(stderr, "out of bounds access, current size is %lu but given index is %" PRIu64 "\n", m_size, index);
```

rotu: I’d rather use cerr to be consistent with the surrounding code

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
michael-poehnl: Sorry, now I got it. I thought you introduce something else from acl.h. But it is the own enum class (and acl would be C anyway)

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
michael-poehnl: That's true. Everywhere else in this file std::cerr is used. So it would just make it consistent. But I'm not sure if std::cerr and std::terminate should be our default error handling here.

Karsten1987:
@rotu Any updates on this?

PR #37, opened 2020-01-28
fix qnx build

elBoberido:
the ACL types on QNX are different from the ones on linux, therefore this conditional compilation for the AccessController enums

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
@@ -43,8 +43,12 @@ class AccessController
     /// @brief maximum number of permission entries the AccessController can store
     static constexpr int32_t MaxNumOfPermissions = 20;
 
-    /// @brief identifier for a permission entry (user, group, others, ...)
+/// @brief identifier for a permission entry (user, group, others, ...)
+#if defined(QNX) || defined(QNX__) || defined(__QNX__)
+    enum class Category : std::underlying_type(acl_tag_t)
+#else
     enum class Category : acl_tag_t
+#endif
     {
         USER = ACL_USER_OBJ,
         /// a specific user must be identified by a name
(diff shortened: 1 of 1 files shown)


iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
```diff
@@ -43,8 +43,12 @@ class AccessController
     /// @brief maximum number of permission entries the AccessController can store
     static constexpr int32_t MaxNumOfPermissions = 20;
 
-    /// @brief identifier for a permission entry (user, group, others, ...)
+/// @brief identifier for a permission entry (user, group, others, ...)
+#if defined(QNX) || defined(QNX__) || defined(__QNX__)
```
Karsten1987: the same logic is applied to `defined(__APPLE__)`.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
```diff
@@ -43,8 +43,12 @@ class AccessController
...
+    enum class Category : std::underlying_type(acl_tag_t)
```
Karsten1987: shouldn't that be `typename std::underlying_type<acl_tag_t>::type` or `std::underlying_type_t<acl_tag_t>`? At least this was necessary for clang/osx to compile. Also I am wondering if we still need that `#if/#else` or if linux can also deal with `std::underlying_type`.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
elBoberido: I think for the mac support we need to do some refactoring and introduce something like a platform abstraction where all this types are defined in a single place

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
elBoberido: We need it. I tried it on linux and QNX and that code was compiling. It seems we need to refactor this when we work on the OSX support, though.

PR #39, opened 2020-01-29
Incoming

elBoberido:
reviewed contribution from Dan Rose and fixed QNX build issue

### DIFF
--- a/iceoryx_posh/source/popo/subscriber.cpp
+++ b/iceoryx_posh/source/popo/subscriber.cpp
@@ -103,8 +103,8 @@ void Subscriber::setReceiveHandler(ReceiveHandler_t cbHandler) noexcept
 // name thread if possible
 #ifdef __unix__
     char threadname[16];
-    static ushort thread_index = 0;
-    snprintf(threadname, sizeof(threadname), "receiver-cb_%d", ++thread_index);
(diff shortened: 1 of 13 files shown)


iceoryx_posh/source/popo/subscriber.cpp
```diff
@@ -103,8 +103,8 @@ void Subscriber::setReceiveHandler(ReceiveHandler_t cbHandler) noexcept
 // name thread if possible
 #ifdef __unix__
     char threadname[16];
-    static ushort thread_index = 0;
-    snprintf(threadname, sizeof(threadname), "receiver-cb_%d", ++thread_index);
+    static uint16_t thread_index = 0;
+    snprintf(threadname, sizeof(threadname), "Receive_%d", ++thread_index);
```
elfenpiff: We either change thread_index to an uint32_t or %d will be updated into %hd since thread_index is a short.
Why is thread_index written in snake-case, shouldn't it be named threadIndex

elBoberido: Why? It will automatically get promoted to an int which is bigger than an unsigned short int.

Don't know why there is snake_case. We need to cleanup our code regarding the style.

btw, your snake-case is actually kebab-case ;)

Issue #12, opened 2019-11-28
compilation error in utils


Jekyll555:
## Required information

**Operating system:**
Ubuntu 16.04.6 LTS
(as per ubiquity robotics [image](https://ubiquity-pi-image.sfo2.cdn.digitaloceanspaces.com/2019-06-19-ubiquity-xenial-lxde-raspberry-pi.img.xz) compiled on raspberryPi 3b)

**Compiler version:**
GCC 7.4.0 (for arm ->raspi)

**Observed result or behaviour:**
allocator.cpp:44:75: error: no matching function for call to ‘align(uintptr_t&, const uint64_t&)’
     uintptr_t l_alignedPosition = cxx::align(l_currentAddress, f_alignment);
--> template class has only a shared parameter type for both parameters of cxx::align. Add a second template type or make second parameter (alignment) a static type e.g. uint64_t.

**Expected result or behaviour:**
Does compile.

**Conditions where it occurred / Performed steps:**
Using ubiquity robotics [image](https://ubiquity-pi-image.sfo2.cdn.digitaloceanspaces.com/2019-06-19-ubiquity-xenial-lxde-raspberry-pi.img.xz) compiled on raspberryPi 3b
additional to install instruction added GCC v7 as default.
(Compilation did finish after changing cxx::align 2nd parameter type (in helplets.hpp) to static type uint64_t.)

michael-poehnl:
Thanks @Jekyll555 . Funny that this is no problem with gcc 7.4 on Intel. We also have ARM targets in our build pipeline but maybe not the combination ARM and gcc 7.4

dkroenke:
Hi @Jekyll555 i can confirm this issue when i try to compile iceoryx on Raspbian with a Raspberry Pi 3B.
The reason for that is that Raspbian is based on 32-Bit but the hardware processor has 64-bit. The 32-bit GCC compiler can then not handle the alignment for the different datatypes. 

I guess that the ubiquity robotics image is also based on 32-Bit. You can check that with uname -a on which platform the operating system is based. If it is "armv7" then it is 32 Bit.

I created already an internal proposal how to fix that issue. I need to review it with my colleague and would then create a Pull-Request with you as reviewer.

Jekyll555:
HI @dkroenke Confirming the raspbian being 32bit image: 
[...]

PR #40, opened 2020-01-31
IOX-#12 32-bit support - fixing alignment problem

elfenpiff:
- fixed alignment problems in 32-bit mode
- duration rounding error in 32-bit fixed
- objectpool bug fixed which occurred only in 32-bit
- convert support for long long (unsigned) int is gone since long long cannot be supported in 32-bit
- 32-bit problems im memory management solved

Signed-off-by: Eltzschig Christian (CC-AD/ESW1) <christian.eltzschig2@de.bosch.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,13 @@ Examples for a "porcelain" API would be e.g.
 [Adaptive Autosar Foundation](https://www.autosar.org/fileadmin/Releases_TEMP/Adaptive_Platform_19-03/AdaptiveFoundation.zip)
 (see AUTOSAR_EXP_ARAComAPI.pdf) or [ROS](https://www.ros.org).
 
+### Supported Platforms
+
+ * Linux
+ * QNX
+ * macOS (not yet - currently in progress with high priority)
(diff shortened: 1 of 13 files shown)


elfenpiff:
> Can you maybe add a comment in the Readme.md about our supported platforms with a prio list?

Done!

README.md
```diff
@@ -26,6 +26,13 @@ Examples for a "porcelain" API would be e.g.
 [Adaptive Autosar Foundation](https://www.autosar.org/fileadmin/Releases_TEMP/Adaptive_Platform_19-03/AdaptiveFoundation.zip)
 (see AUTOSAR_EXP_ARAComAPI.pdf) or [ROS](https://www.ros.org).
 
+### Supported Platforms
+
+ * Linux
+ * QNX
+ * macOS (not yet - currently in progress with high priority)
```
michael-poehnl: I guess we then have to stop talking about zero-copy for POSIX in our headline ;-)

elfenpiff: Zero-Copy for POSIX and Friends ;)

Issue #33, opened 2020-01-15
Basic Windows 10 support


budrus:
## Brief feature description

Having the possibility to use iceoryx on Windows 10 

## Detailed information

* Needed for ROS2 tier 1 support [ROS2 Tiers](https://www.ros.org/reps/rep-2000.html#support-tiers)

- [ ] Unit and integration tests, except timing tests, of the supported features need to run on Windows
    - [ ] remove `define public private` #529 
    - [ ] all tests build and are green

## The following features have to be implemented
- [x] Implement and integrate named pipes for IPC in windows
- [ ] inter process capable mutex
- [ ] inter process capable semaphore
- [ ] use named ipc semaphore in named pipe

rex-schilasky:
Highly welcomed by the [eCAL](https://github.com/continental/ecal) project too.

elfenpiff:
I am on it ...

MatteoRagni:
Today I tried a compilation in Windows 10 Pro (VS2017 x64):

```
cmake -Bbuild -Hiceoryx_meta -Ax64
cmake --build build
```

## Patching

In order to successfully compile the current main (347224e):

 * I had to force the usage of the x64 toolchain for dependencies (`set(CMAKE_GENERATOR_PLATFORM "x64")` in the `if(WIN32)` section of `cmake/[cpptoml,cyclonedds,googletest]/CmakeListes.txt`. I also added a `"-Ax64"` option in `execute_process`. I set the same `CMAKE_GENERATOR_PLATFORM` in the `*.cmake.in` files, but it should be useless (imho)
 * I had to define `target_compile_definitions(iceoryx_posh PUBLIC _ENABLE_EXTENDED_ALIGNED_STORAGE)` in `iceoryx_posh`
 * I had to manually make a downcast for few tests (with `static_cast<T>`), since CL raises a compilation error:
   * `iceoryx_posh\test\moduletests`: `test_mepoo_mempool.cpp:28`, `test_mepoo_typed_mempool.cpp:[42, 93]`
   * `iceoryx_utils\test\moduletests`: `test_cxx_forward_list.cpp:1837`, `test_cxx_list.cpp:2140`, `test_semaphore_module.cpp:78`
[...]

Issue #32, opened 2020-01-15
macOS support


budrus:
## Brief feature description

Having the possibility to use iceoryx on macOS

## Detailed information

Needed for ROS2 tier 1 support
[ROS2 Tiers](https://www.ros.org/reps/rep-2000.html#support-tiers)

marthtz:
Tasks for macOS support:
- [ ] ACL needs to be adapted to macOS
- [ ] Semaphore `sem_timedwait()` not available and needs to be created
- [ ] Implement new OS timer based on "dispatch source"
- [ ] Replace message queues with socket communication
- [ ] Replace or find appropriate RT library if necessary
- [ ] Modify `pthread_setname_np()` calls
- [ ] Resolve type conflicts for `cxx::align` in
  - `requiredManagementMemorySize()`
  - `requiredManagementMemorySize()`
  - `requiredManagementMemorySize()`
  - `Allocator::allocate()`
- [ ] Resolve type conflict for `gid_t` in `PosixUser::getGroups()`
- [ ] Resolve type conflict for `wprintw()` in `printProcessIntrospectionData()`

michael-poehnl:
I guess we start without acl or a replacement for it. Having different access groups on macOS is low prio

elfenpiff:
Mac OS works!

PR #45, opened 2020-02-13
Iox #33 windows support : part 1 / 2

elfenpiff:
The amount of changes are quiet huge and therefore I would like to create a first pull request to ease the pain.
The current state is that everything compiles in Windows (Visual Studio 2019) thanks to stub functions. Linux (64-bit & 32-bit) as well as QNX is running perfectly. Additionally, I implemented a posix interface for Windows Semaphores, SharedMemory and FileHandlers. What remains is the message queue or the reimplementation of the RouDi2App communication with sockets.
@michael-poehnl : its your turn ... give me some sockets


michael-poehnl:
On the weekened I will meet my old friend Richard from the 90s 
https://en.wikipedia.org/wiki/UNIX_Network_Programming

iceoryx_posh/source/roudi/roudi_lock.cpp
```diff
@@ -13,22 +13,22 @@
...
-#include <sys/socket.h>
-#include <sys/types.h>
-#include <unistd.h>
 
 namespace iox
 {
 namespace roudi
 {
 RouDiLock::RouDiLock()
+    : m_socket_fd(cxx::makeSmartC(socket, cxx::ReturnMode::PRE_DEFINED_ERROR_CODE, {-1}, {}, AF_INET, SOCK_STREAM, 0)
```
michael-poehnl: Once the unix domain socket wrapper is merged, we could think of using this one here and not directly the socket calls. 
What do you think?

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -145,24 +114,15 @@ bool MqBase::send(const MqMessage& msg) const noexcept
...
-    }
-    else
-    {
-        return true;
-    }
+    auto logLengthError = [&msg](cxx::expected<posix::IpcChannelError>& error) {
+        if (error.get_error() == posix::IpcChannelError::MESSAGE_TOO_LONG)
+        {
+            const size_t messageSize =
+                static_cast<size_t>(msg.getMessage().size()) + posix::MessageQueue::NULL_TERMINATOR_SIZE;
```
michael-poehnl: We need the IPC channel abstraction here isf we want to resuse this also for unix domain socket. This will become relevant when introducing unix domain socket for MAC (and also Windows?)

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
michael-poehnl: This is work in progress. First functionality already in this PR but not finalized and not yet used

iceoryx_utils/platform/win/include/iceoryx_utils/platform/acl.hpp
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
michael-poehnl: @elfenpiff  will we use this at all on Windows? Or will the acl functionality be skipped on a higher level?

iceoryx_posh/source/roudi/roudi_lock.cpp
elfenpiff: You are completely right. I will use your abstraction as soon as it is available ...

iceoryx_utils/platform/win/include/iceoryx_utils/platform/acl.hpp
elfenpiff: I wanted to keep the ability to use this in Windows. In Windows we have something called DACLs (https://docs.microsoft.com/en-us/windows/win32/secauthz/dacls-and-aces) and I think it is possible to map these ACL calls to Windows.

If in the future the need arises to have some more security in Windows we still have the possibility to go that road without changing our actual code base.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
```diff
@@ -97,7 +97,7 @@ class RouDiMultiProcess
     bool m_killProcessesInDestructor;
     std::atomic_bool m_runThreads;
 
-    const uint32_t m_MessageQueueTimeoutMilliseconds = 100;
+    const units::Duration m_messageQueueTimeout{units::Duration::milliseconds(static_cast<unsigned long long>(100))};
```
MatthiasKillat: why not use 100ULL directly instead of the static cast from an int?

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -64,65 +63,35 @@ std::string mqMessageErrorTypeToString(const MqMessageErrorType msg) noexcept
 MqBase::MqBase(const std::string& InterfaceName, const long maxMessages, const long messageSize) noexcept
     : m_interfaceName(InterfaceName)
 {
-    auto maxMessageSize = messageSize;
-    if (messageSize > MAX_MESSAGE_LENGTH)
+    m_maxMessages = maxMessages;
```
MatthiasKillat: could be initialized in the initializer list.

iceoryx_posh/CMakeLists.txt
```diff
@@ -198,6 +207,11 @@ if(TOML_CONFIG)
     target_compile_definitions(RouDi PRIVATE -DTOML_CONFIG_FILE)
 endif(TOML_CONFIG)
 
+if(test)
+    # since the unit tests have the roudi environment as dependency
+    set(roudi_environment ON) 
+endif(test)
+
 if(roudi_environment)
```
elBoberido: if (roudi_environment OR test)

iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
```diff
@@ -77,4 +81,3 @@ class MemoryManager
 
 } // namespace mepoo
 } // namespace iox
-
```
elBoberido: no new line

iceoryx_posh/include/iceoryx_posh/internal/popo/receiver_port.hpp
```diff
@@ -94,9 +94,11 @@ class ReceiverPort : public BasePort
...
-    const MemberType_t* getMembers() const;
-    MemberType_t* getMembers();
+    const MemberType_t* getMembers() const noexcept;
+    MemberType_t* getMembers() noexcept;
 };
 
 } // namespace popo
 } // namespace iox
+
+#include "iceoryx_posh/internal/popo/receiver_port.inl"
```
elBoberido: missing new line

iceoryx_posh/source/runtime/message_queue_message.cpp
```diff
@@ -84,7 +84,7 @@ void MqMessage::setMessage(const std::string& msg) noexcept
     clearMessage();
 
     m_msg = msg;
-    if (m_msg.back() != m_separator && m_msg.size() > 0)
+    if (m_msg.back() != m_separator && m_msg.size() > 0) /// @todo shouldn't this be "if (m_msg.size() == 0 || m_msg.back() != m_separator)"???
```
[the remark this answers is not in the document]
MatthiasKillat: I would think so, the message is invalid when it does not end with the separator or the size is zero. It cannot end with separator when it has size zero, which means the size needs to be checked first. Depends a bit on how back() behaves when the size is zero.

iceoryx_utils/test/moduletests/test_semaphore_module.cpp
```diff
@@ -226,4 +230,23 @@ TEST_F(Semaphore_test, MoveCTorUnnamed)
...
+
+    std::thread t([&] {
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        ts.tv_nsec += 2000000; // wait 2 ms
+        semaphore->timedWait(&ts, false);
+        timedWaitFinish.store(true);
+    });
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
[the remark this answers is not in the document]
elfenpiff: You are right, but I really have to test the timed_wait that it actual waits! I could move it into an integration test but your argument would still be valid then. I could also remove this test, but then the sem_timedwait implementation would not be tested, but it is important to test it since in Windows we have a complex conversion into UTC.
@elBoberido what is your opinion?

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -0,0 +1,437 @@
...
+    struct timeval tv = timeout;
+
+    auto setsockoptCall = cxx::makeSmartC(setsockopt,
+                                          cxx::ReturnMode::PRE_DEFINED_ERROR_CODE,
+                                          {static_cast<ssize_t>(ERROR_CODE)},
+                                          {EWOULDBLOCK},
+                                          m_sockfd,
+                                          SOL_SOCKET,
+                                          SO_RCVTIMEO,
+                                          (const char*)&tv,
```
elBoberido: C-style casts are not allowed

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -0,0 +1,437 @@
...
+    struct timeval tv = timeout;
+
+    auto setsockoptCall = cxx::makeSmartC(setsockopt,
+                                          cxx::ReturnMode::PRE_DEFINED_ERROR_CODE,
+                                          {static_cast<ssize_t>(ERROR_CODE)},
+                                          {EWOULDBLOCK},
+                                          m_sockfd,
+                                          SOL_SOCKET,
+                                          SO_SNDTIMEO,
+                                          (const char*)&tv,
```
elBoberido: C-style casts are not allowed

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -0,0 +1,437 @@
...
+                                        nullptr);
+
+        if (recvCall.hasErrors())
+        {
+            // timeout is also an error and would return here
+            return createErrorFromErrnum(recvCall.getErrNum());
+        }
+        else
+        {
+            return cxx::success<std::string>(std::string(&(message[0])));
```
elBoberido: - please add a 0 at message[MAX_MESSAGE_SIZE]
- you can remove the &..[0]; just "message" is sufficient

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -0,0 +1,437 @@
...
+        if (IpcChannelSide::SERVER == m_channelSide)
+        {
+            unlink(m_name.c_str());
+
+            auto bindCall = cxx::makeSmartC(bind,
+                                            cxx::ReturnMode::PRE_DEFINED_ERROR_CODE,
+                                            {ERROR_CODE},
+                                            {},
+                                            sockfd,
+                                            (struct sockaddr*)&m_sockAddr,
```
elBoberido: C-style casts are not allowed

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -0,0 +1,437 @@
...
+    /// @todo mode handling
+    if (m_name.empty())
+    {
+        return cxx::error<IpcChannelError>(IpcChannelError::INVALID_CHANNEL_NAME);
+    }
+
+    auto socketCall =
+        cxx::makeSmartC(socket, cxx::ReturnMode::PRE_DEFINED_ERROR_CODE, {ERROR_CODE}, {}, AF_LOCAL, SOCK_DGRAM, 0);
+
+    if (!socketCall.hasErrors())
```
elBoberido: consider early returns for better readability

elBoberido:
everything related to unix domain sockets should probably fixed in elfefpiff/iox32 with the mac support

elfenpiff:
> everything related to unix domain sockets should probably fixed in elfenpiff/iox #32 with the mac support

done in iox-#32-add-mac-os-support-part-II

iceoryx_utils/CMakeLists.txt
```diff
@@ -185,17 +208,23 @@ endif(PERFORM_CLANG_TIDY)
 #
 ########## exporting library ##########
 #
+
+#TODO add support for multiple include directories to setup_install_directories_and_export_package
+install(
+    DIRECTORY ${ICEORYX_PLATFORM_INCLUDES}
```
elBoberido: this breaks out of tree builds. please change to
DIRECTORY ${ICEORYX_PLATFORM}/include/
and please add
$<INSTALL_INTERFACE:include/${PREFIX}>
to target_include_directories(iceoryx_platform ...

PR #48, opened 2020-02-21
Iox #32 add macos support

marthtz:
Two modifications required to compile on macOS:
- [ ]  Resolve type conflict for `gid_t groups[] / getgrouplist` in posix_access_rights.cpp
- [ ]  Resolve type conflict for `size_t length / fromString()` in serialization.inl

Test do also not compile right now.


iceoryx_utils/test/moduletests/test_relative_pointer.cpp
```diff
@@ -123,7 +123,7 @@ TYPED_TEST(relativeptrtests, ConstrTests)
 
     {
         auto offset = ShmSize / 2;
-        void* adr = static_cast<uint8_t*>(memMap1.getMappedAddress()) + offset;
+        void* adr = reinterpret_cast<void*>(reinterpret_cast<size_t>(memMap.getMappedAddress() + offset));
```
MatthiasKillat: Should probably be casted to uint64_t to avoid issues where depending on platform size_t is something else (should not really happen but still, I am not sure). Other than that I prefer the reinterpret casts to void because its quite low level.

iceoryx_utils/test/moduletests/test_relative_pointer.cpp
```diff
@@ -13,13 +13,13 @@
...
+#include "iceoryx_utils/platform/mman.hpp"
+#include "iceoryx_utils/platform/stat.hpp" /* For mode constants */
 
 #include "test.hpp"
 
 #include <cstring>
 #include <fcntl.h> /* For O_* constants */
-#include <sys/mman.h>
-#include <sys/stat.h> /* For mode constants */
 using namespace ::testing;
```
MatthiasKillat: A more serious issue is that the tests are not robust when the OS does not respect the shared memory mapping hint. This seems not to be fixed here but an internal fix is in progress which will later be merged to OSS. (There are still problems with QNX in this fix, so I cannot provide a fix right now...)

elfenpiff:
we will directly merge iox-#32-add-macos-support-partII

PR #49, opened 2020-02-22
Fix string comparison so it compares past the first '\0'

rotu:


mossmaurice:
Hi @rotu thanks for this fix! The CString100 class definitely needs rework ;-) However, I think I would not merge this pull request as @FerdinandSpitzschnueffler removed the CString100 class from the complete code base and replaced it with a better cxx::string<Capacity> in the last few weeks. The patches have not landed upstream yet.

@FerdinandSpitzschnueffler: What do you think? When will the new cxx::string class be upstreamed?

FerdinandSpitzschnueffler:
@mossmaurice, @rotu: I think the new cxx::string will be upstreamed within the next 3-4 weeks and it already provides the requested comparison. So I would not merge this pull request neither.
@rotu: Is this delay ok for you or do you need the fix urgently?

rotu:
@FerdinandSpitzschnueffler, @mossmaurice No worries. Closing unmerged.

PR #52, opened 2020-02-24
Fix operator mistake

rotu:
Thanks, Clang!
```
--- stderr: iceoryx_utils
In file included from /opt/ros/master/src/eclipse/iceoryx/iceoryx_utils/source/posix_wrapper/timer.cpp:15:
/opt/ros/master/src/eclipse/iceoryx/iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp:87:59: warning: result of '2 ^ 24' is 26; did you mean '1 << 24' (16777216)? [-Wxor-used-as-pow]
        static constexpr uint32_t MAX_DESCRIPTOR_VALUE{(2 ^ 24) - 1};
                                                        ~~^~~~
                                                        1 << 24
/opt/ros/master/src/eclipse/iceoryx/iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp:87:59: note: replace expression with '0x2 ^ 24' or use 'xor' instead of '^' to silence this warning
1 warning generated.
```

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp
@@ -84,7 +84,7 @@ class Timer
     class OsTimer;
     struct OsTimerCallbackHandle
     {
-        static constexpr uint32_t MAX_DESCRIPTOR_VALUE{(2 ^ 24) - 1};
+        static constexpr uint32_t MAX_DESCRIPTOR_VALUE{(1 << 24) - 1};
         static sigval indexAndDescriptorToSigval(uint8_t index, uint32_t descriptor);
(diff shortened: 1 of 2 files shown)


rotu:
Please check that XOR was not, in fact, intended. I don't understand this code yet.

elBoberido:
I'm feeling quite stupid right now :D
you are right, it must be (1 << 24) -1

Since there is no possibility to know if the timer callback is still running when it is deactivated from somewhere else in the code, this is an indicator if the timer is still valid before locking the mutex and executing the actual callback. When the mutex is locked, it is save to execute the callback, since the call to deactivate the callback will block until the mutex is unlocked.
If we were allowed to use the heap, this would be much easier to implement.

PR #69, opened 2020-03-21
Iox #32 implement posix timer and introduced timing test

elfenpiff:


iceoryx_utils/testutils/timing_test.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+namespace iox
+{
+namespace testutils
+{
+inline bool performingTimingTest(const std::function<void()>& testCallback,
+                                 const uint64_t repetitions,
+                                 std::atomic_bool& testResult) noexcept
+{
+    for (uint64_t i = 0; i < repetitions; ++i)
```
dkroenke: unsigned literal needed

iceoryx_utils/testutils/timing_test.hpp
```diff
@@ -0,0 +1,142 @@
...
+    GTestType(Name, TimingTest_##Case)                                                                                 \
+    {                                                                                                                  \
+        std::atomic_bool timingTestResult{true};                                                                       \
+        std::string errorMessages;                                                                                     \
+        bool testResult = iox::testutils::performingTimingTest(Test, Repetitions, timingTestResult);                   \
+        EXPECT_TRUE(testResult);                                                                                       \
+        if (!testResult)                                                                                               \
+        {                                                                                                              \
+            std::cout << "\n" << errorMessages << std::endl;                                                           \
```
dkroenke: I'm curious why here two times newline is used in different shapes. As far as i understood std::endl is flushing the buffer and \n not. Is this here the reason for it?

iceoryx_utils/test/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
-project(test_iceoryx_posh VERSION ${test_iceoryx_posh})
+set(test_iceoryx_utils_VERSION 0.16.1)
```
dkroenke: We should not set version numbers for test libs because its harder to maintain and we have to set more version numbers in different CMake files if a release come. 
The tests are always attached to a specific release version, therefore i would not set it. Additionally they are already versioned with git tags.

iceoryx_posh/test/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
-project(test_iceoryx_posh VERSION ${test_iceoryx_posh})
+set(test_iceoryx_posh_VERSION 0.16.1)
```
dkroenke: We should not set version numbers for test libs because its harder to maintain and we have to set more version numbers in different CMake files if a release come.
The tests are always attached to a specific release version, therefore i would not set it. Additionally they are already versioned with git tags.

iceoryx_utils/platform/mac/source/time.cpp
```diff
@@ -0,0 +1,150 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "iceoryx_utils/platform/time.hpp"
+
+static std::chrono::nanoseconds getNanoSeconds(const timespec& value)
+{
+    static constexpr uint64_t NANOSECONDS = 1000000000;
```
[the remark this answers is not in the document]
elfenpiff: done

iceoryx_utils/testutils/timing_test.hpp
elfenpiff: The \n in the beginning adds a new line which improves the readability quiet a bit. Otherwise you barely see where the normal gtest message ends and the timing error trace begins.

iceoryx_utils/test/CMakeLists.txt
elfenpiff: If we do not set the version we have a CMake syntax error. It causes only a warning but it is still there and complains that you have to set a version. This is somehow sloppy and I would like to have this fixed.
@dkroenke I set the version to zero so that it is somehow unversionised.

iceoryx_posh/test/CMakeLists.txt
elfenpiff: @dkroenke I will set the version to zero

iceoryx_utils/testutils/timing_test.hpp
```diff
@@ -0,0 +1,142 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <string>
+
+/// @brief This headers provides TIMING_TEST unit test infrastructure. The idea
```
dkroenke: The name "TIMING_TEST" is a litle bit misleading here because there are no timing constrints checked here. What is here done is to repeat the test until the condition is met or to fail. I know the background why we are doing that but then i would rename it to something like "REPEAT_TEST". That name is more meaningful. Iknow that we have different timing constraints on different platforms like arm or x64. But this should be tested different.

Maybe we can use the timeout mechanism of the semaphore to offer a real timing test which checks for a certain timing constraint?

iceoryx_posh/source/roudi/roudi_multi_process.cpp
```diff
@@ -150,13 +149,11 @@ void RouDiMultiProcess::processMessage(const runtime::MqMessage& message,
 {
     switch (cmd)
     {
-    case runtime::MqMessageType::SERVICE_REGISTRY_CHANGE_COUNTER:
-    {
+    case runtime::MqMessageType::SERVICE_REGISTRY_CHANGE_COUNTER: {
```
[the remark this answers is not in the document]
dkroenke: @elfenpiff is clang-format reworked?

iceoryx_utils/testutils/timing_test.hpp
elfenpiff: I think we should stick to the name TIMING_TEST since with this name we communicate to fellow developers that we are testing a construct with timing constraints. Our approach is to test it repeatedly but maybe later we have an even better approach and then the name should be the same and the rest should just change under the hood.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -41,147 +46,363 @@ class Timer_test : public Test
...
+    auto startTime = std::chrono::system_clock::now();
+    sut.stop();
+    auto endTime = std::chrono::system_clock::now();
+    auto elapsedTime = std::chrono::duration_cast<std::chrono::milliseconds>(endTime - startTime).count();
+
+    TIMING_TEST_EXPECT_TRUE(elapsedTime < 10);
+});
+
+// This test has to be repeated more often since we are spawning a lot
+// of threads in here
```
mossmaurice, line 311: Is this meant as an `@todo`?

iceoryx_utils/platform/mac/source/time.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+    timespec currentTime, startTime;
+    clock_gettime(CLOCK_REALTIME, &currentTime);
+    {
+        std::lock_guard<std::mutex> l(timerid->parameter.mutex);
+        curr_value->it_interval = timerid->parameter.timeParameters.it_interval;
+        startTime = timerid->parameter.startTime;
+    }
+    curr_value->it_value.tv_sec = curr_value->it_interval.tv_sec - (currentTime.tv_sec - startTime.tv_sec);
+    curr_value->it_value.tv_nsec = curr_value->it_interval.tv_nsec - (currentTime.tv_nsec - startTime.tv_nsec);
```
elBoberido, line 138: the calculation could become negative, are positive seconds and negative nanoseconds allowed?

iceoryx_utils/platform/mac/source/time.cpp
```diff
@@ -0,0 +1,146 @@
...
+    timer->callbackParameter = sevp->sigev_value;
+
+    timer->thread = std::thread([timer] {
+        while (timer->keepRunning.load())
+        {
+            if (waitForExecution(timer))
+            {
+                timer->parameter.mutex.lock();
+                bool doCallCallback =
+                    !timer->parameter.runOnce || timer->parameter.runOnce && !timer->parameter.wasCallbackCalled;
```
elBoberido, line 82: please add braces to make the intention clear

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -41,147 +46,363 @@ class Timer_test : public Test
...
-
-    // Save the time before start()
-    auto start = Timer::now();
-    osTimer.start(Timer::RunMode::ONCE);
-
-    std::this_thread::sleep_for(std::chrono::milliseconds(1100));
-
-    auto end = osTimer.now();
+  public:
+    iox::units::Duration timeout{4_ms};
```
elBoberido, line 52: could you make it const and uppercase to indicate that it's a constant

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -41,147 +46,363 @@ class Timer_test : public Test
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(2 * timeout.milliSeconds<int>() / 3));
+    TIMING_TEST_EXPECT_FALSE(sut.hasExpiredComparedToCreationTime());
+    std::this_thread::sleep_for(std::chrono::milliseconds(2 * timeout.milliSeconds<int>() / 3));
+    TIMING_TEST_EXPECT_TRUE(sut.hasExpiredComparedToCreationTime());
+});
 
-    osTimer.stop();
+TEST_F(TimerStopWatch_test, ResetWithDurationIsExpired)
+{
+    Timer sut(1_ms);
```
elBoberido, line 74: magic number

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -41,147 +46,363 @@ class Timer_test : public Test
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(2));
```
elBoberido, line 75: if no magic number was used, this could be 2 times the constant

Issue #70, opened 2020-03-21
Memory abstraction and RouDi modularization step 2, fixed string improvements and other fixes


elBoberido:
## Brief feature description

Step 2 of the RouDi modularisation and shared memory abstraction

## Detailed information

* building blocks for shared memory abstraction (step 2, followup of #59 )
* modularization of RouDi (step 2, followup of #59)
* building blocks for sender and receiver port refactoring
* fixed string improvements
* fix warnings signed/unsigned at numeric literals

PR #74, opened 2020-03-23
Iox #70 roudi modularization part2

mossmaurice:
Implements #70


Karsten1987:
Is there a new release for this? I would need to put rmw_iceoryx on the same tag then.

Issue #25, opened 2019-12-16
Support for n:m communciation (multiple senders on a topic / CaProID)


budrus:
## Brief feature description

Currently for each CaProID only one sender is allowed. This shall be extended to support m:n communication.

## Detailed information

ROS2 is based on the DDS communication philosophy. In contrast to AUTOSAR, it is allowed that several publishers are writing to the same topic. This is currently not fully supported in iceoryx.  For this feature the receiver ports must support multi-pusher queues.
To be more flexible for different communication patterns, we will have a bigger refactoring. Also the iceoryx public API will be affected

- [x] chunk queue building block 

- [x] chunk distributor building block

- [x] chunk sender building block

- [x] chunk receiver building blocks

- [x] publisher port

- [x] subscriber port for single producer

- [x] condition variable

- [x] Integration of condition variable in existing building blocks

- [x] MPMC queue and integration in variant queue + harmonization of queue interfaces and port interfaces

- [x] subscriber port for multi producer

- [x] adaptation of RouDi for managing new ports and condition variable

- [ ] adaptation of public API

- [ ] new examples and go live

mossmaurice:
We often get the issue <-> pull request linkage wrong if we work on a part of an issue. We should check that.

Hence, in this case reopen as just a part of #25 was merged.

mossmaurice:
From #165: Implement move c'tor and move assignment for ChunkQueuePopper, ChunkQueuePusher, ConditionVariableWaiter and ConditionVariableSignaler (https://github.com/eclipse/iceoryx/pull/165#discussion_r450746434)

mossmaurice:
Will be continued in #252

PR #81, opened 2020-04-02
Iox #25 chunk sender building block

budrus:
This is one of the steps for #25. It introduces the new building block ChunkSender


iceoryx_posh/test/moduletests/test_popo_chunk_distributor.cpp
```diff
@@ -61,16 +63,16 @@ class ChunkDistributor_test : public Test
...
-        return std::make_shared<ChunkDistributorData>();
+        return std::make_shared<ChunkDistributorData<MAX_NUMBER_QUEUES, ThreadSafePolicy>>(HISTORY_SIZE);
     }
 };
 
 TEST_F(ChunkDistributor_test, AddingNonAddedQueueWorks)
 {
     auto sutData = getChunkDistributorData();
-    ChunkDistributor sut(sutData.get());
+    ChunkDistributor<MAX_NUMBER_QUEUES, ThreadSafePolicy> sut(sutData.get());
```
elBoberido, line 75: a typed test could be used to also test the SingleThreadedPolicy

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -0,0 +1,550 @@
...
+{
+    auto chunk = m_chunkSender.allocate(sizeof(DummySample));
+    EXPECT_FALSE(chunk.has_error());
+    EXPECT_THAT(m_memoryManager.getMemPoolInfo(0).m_usedChunks, Eq(1u));
+
+    auto errorHandlerCalled{false};
+    auto errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler([&errorHandlerCalled](
+        const iox::Error, const std::function<void()>, const iox::ErrorLevel) { errorHandlerCalled = true; });
+
+    iox::mepoo::ChunkHeader* myCrazyChunk = new iox::mepoo::ChunkHeader();
```
elBoberido, line 173: just a nitpick, but unique_ptr could be used to be more idiomatic with modern C++

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
```diff
@@ -0,0 +1,172 @@
...
+    // chunk
+    if (getMembers()->m_lastChunk && getMembers()->m_lastChunk.hasNoOtherOwners()
+        && getMembers()->m_lastChunk.getChunkHeader()->m_info.m_usedSizeOfChunk
+               >= getMembers()->m_memoryMgr->sizeWithChunkHeaderStruct(payloadSize))
+    {
+        if (getMembers()->m_chunksInUse.insert(getMembers()->m_lastChunk))
+        {
+            getMembers()->m_lastChunk.getChunkHeader()->m_info.m_payloadSize = payloadSize;
+            getMembers()->m_lastChunk.getChunkHeader()->m_info.m_usedSizeOfChunk =
+                getMembers()->m_memoryMgr->sizeWithChunkHeaderStruct(payloadSize);
```
elBoberido, line 53: small performance optimization: sizeWithChunkHeaderStruct is already called in the if condition. this could be moved out of the if statement

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
```diff
@@ -0,0 +1,172 @@
...
+            return cxx::success<mepoo::ChunkHeader*>(getMembers()->m_lastChunk.getChunkHeader());
+        }
+        else
+        {
+            return cxx::error<ChunkSenderError>(ChunkSenderError::TOO_MANY_CHUKS_ALLOCATED_IN_PARALLEL);
+        }
+    }
+    else
+    {
+        // START of critical section, chunk will be lost if process gets hard terminated in between
```
elBoberido, line 63: some IDEs support code folding if BEGIN/END is used

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
budrus: I also extended it to be able to reuse the chunk if the payload size increased but it still fits the chunks

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
```diff
@@ -0,0 +1,171 @@
...
+    {
+        auto& chunkInfo = chunk.getChunkHeader()->m_info;
+        if (!chunkInfo.m_externalSequenceNumber_bl)
+        {
+            chunkInfo.m_sequenceNumber = getMembers()->m_sequenceNumber;
+            getMembers()->m_sequenceNumber++;
+        }
+        else
+        {
+            getMembers()->m_sequenceNumber++; // for Introspection, else nobody updates.
```
mossmaurice, line 159: It is a bit odd to have special introspection data handling in this class. Could we move this to another place? This method only cares about incrementing the `sequenceNumber`. Maybe we want to rename it?

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -0,0 +1,573 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
mossmaurice, line 1: Year

budrus:
> 
> 
> The test `RoudiFindService_test.OfferSingleMethodServiceSingleInstance` is constantly segfaulting. I suppose it is not connected to this pull request? We should have a look.

No problem on my side ....

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
budrus: If the sequence number is set on user side, we use this one for the ChunkInfo. But for the throughput calculation stuff, the sequence number that belongs to the port data in shared memory is used. Therefore this must be incremented. I copied this section from the old sender port which will be replaced with the new building blocks. So I would say this is no special introspection data handling but more keeping an internal sequence number updated also if the one that is visible in the chunk will be set by the user. The comment could be better here

mossmaurice:
> 
> 
> > The test `RoudiFindService_test.OfferSingleMethodServiceSingleInstance` is constantly segfaulting. I suppose it is not connected to this pull request? We should have a look.
> 
> No problem on my side ....

Clean builld fixed the issue.

budrus:
The minor findings will be fixed in the follow up branch iox-#25-chunk-receiver-building-block

Issue #78, opened 2020-03-27
Reduce default memory consumption with config and mempools 


budrus:
## Brief feature description

The current defaults for compile time configuration and mempools are quite high. Better have reasonable defaults and let people increase the values if needed

## Detailed information

With the current default configuration we need ~200 MB for the internal data structures and the default memory pools have ~ 600 MB. This is a challenge on systems which only have some hundred MBs of RAM left. Beginners could be confused and get out of memory errors on RouDi startup. We support 4096 publishers and subscribers with the default config. Because of max size allocation with fixed size containers this blows up the memory. For most users a lower number is sufficient. The mempools can be easily configured with the config file now, so having the large default does not make sense IMHO

mossmaurice:
Reopening, to make `RouDiEnvironment` more configurable and adapt TOML files.

PR #82, opened 2020-04-07
Iox #78 Minor RouDi config adaptions

mossmaurice:
* Make `RouDi_GTest` class more configurable
* Reflect RouDiConfig changes in TOML files


budrus:
Fine so far, but here, the old default config is hardcoded
https://github.com/eclipse/iceoryx/blob/master/doc/usage-guide.md

Can you replace this with a link instead of having a config written there?

budrus:
Oh, and we have to adapt the ipceperf example to not use the large chunks (or ensure a proper mempool config)

doc/usage-guide.md
```diff
@@ -80,54 +81,4 @@ size = 1024
...
-[[segment.mempool]]
-size = 16777216
-count = 5
-
-[[segment.mempool]]
-size = 33554432
-count = 2
-```
\ No newline at end of file
+When no config file is specified, the [default config](../iceoryx_posh/etc/iceoryx/roudi_config_medium_default.toml) will be used.
```
elBoberido, line 84: this is wrong, when no config file is available, then a configuration similar to the above will be used

iceoryx_posh/etc/iceoryx/roudi_config_medium_default.toml
```diff
@@ -0,0 +1,32 @@
+[general]
```
elBoberido, line 1: why a new file? just adjust the old one. If somebody needs a custom config it is easy to create one.

Also, I wouldn't copy the file to /etc/iceoryx/roudi_config.toml but /etc/iceoryx/roudi_config.toml.example and hardcode sane defaults in the code. If a user needs a custom config he/she can just copy the file to /etc/iceoryx/roudi_config.toml and that one would be used. Just have a look how other deamons handle this.

doc/usage-guide.md
mossmaurice: Fixed

iceoryx_posh/etc/iceoryx/roudi_config_example.toml
```diff
@@ -1,48 +1,33 @@
+# Adapt this config to your needs and rename it to e.g. roudi_config.toml
```
elBoberido, line 1: not "to e.g. roudi_config.toml" but just "to roudi_config.toml so that RouDi finds it at startup"

mossmaurice: True, will fix in follow up pull request.

PR #95, opened 2020-05-08
Fix cross-compilation issues 

mauropasse:
Define some pthread vars and perform cast to match template definition

### DIFF
--- a/iceoryx_posh/source/roudi/memory/mempool_collection_memory_block.cpp
+++ b/iceoryx_posh/source/roudi/memory/mempool_collection_memory_block.cpp
@@ -36,7 +36,8 @@ MemPoolCollectionMemoryBlock::~MemPoolCollectionMemoryBlock() noexcept
 
 uint64_t MemPoolCollectionMemoryBlock::size() const noexcept
 {
-    return cxx::align(sizeof(mepoo::MemoryManager), mepoo::MemPool::MEMORY_ALIGNMENT)
+    return cxx::align(sizeof(mepoo::MemoryManager),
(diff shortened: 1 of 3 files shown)


budrus:
Thanks @mauropasse 
As we are an Eclipse project you would need to sign the Eclipse Contributor Agreement. 
see also https://wiki.eclipse.org/Development_Resources/Contributing_via_Git

Then we can merge your changes and you did a contribution. I would appreciate this. If you are not interested in this extra step we could integrate your patches in another commit.

iceoryx_posh/source/roudi/memory/mempool_collection_memory_block.cpp
```diff
@@ -36,15 +36,17 @@ MemPoolCollectionMemoryBlock::~MemPoolCollectionMemoryBlock() noexcept
...
            + mepoo::MemoryManager::requiredFullMemorySize(m_memPoolConfig);
 }
 
 uint64_t MemPoolCollectionMemoryBlock::alignment() const noexcept
 {
     // algorithm::align doesn't like constexpr values
     auto memPoolAlignment = mepoo::MemPool::MEMORY_ALIGNMENT;
-    return algorithm::max(alignof(mepoo::MemoryManager), memPoolAlignment);
+    return algorithm::max(alignof(mepoo::MemoryManager),
+        static_cast<unsigned int>(memPoolAlignment));
```
elfenpiff, line 49: Is this tested on mac os and linux arm 32-bit? Does it compile and work there? If not we should definitely do this before merging this!

elfenpiff:
No one seems to be working on this pull request. It will be closed for now, if the work continues we can reopen it.

PR #97, opened 2020-05-10
iox-#89 add github workflow for build & test

### LINKED ISSUE #89: Usage of github actions to build pull requests
## Brief feature description

External Jenkins jobs are used to build and test the pull requests. Usage of github actions would be better, 

## Detailed information

With the actions, github provides the infrastructure do build the pull requests, this is much more transparent than using external Jenkins servers. The pull requests are also better documented then. In the rmw_iceoryx they were already introduced. We should also use them here. 
e.g. see https://github.com/ros2/rmw_iceoryx/blob/master/.github/workflows/main.yml

prasannabhat:
Add github workflow to build & test iceoryx. (There is an additional build action to build with clang compiler)
This workflow is triggered on push to master or pull request events but only for the master branch.
The job defined in the workflow can be used as a check for PR
closes #89

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,0 +1,50 @@
+# This workflow builds & runs test cases in iceoryx
+
+name: Build & Test
+
+# Triggers the workflow on push or pull request events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build-test"
+  build-test:
+    # The type of runner that the job will run on
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+#       os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-18.04]
+      fail-fast: false
+      
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+       # To build against multiple OSs dependencies installation needs to be OS specific
+      - name: Install iceoryx dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          echo This workflow is triggered by $GITHUB_EVENT_NAME
+          echo Running iceoryx build in directory $GITHUB_WORKSPACE
+          echo Home directory is $HOME
+
+      - name: Build with GNU
+        run: ./tools/iceoryx_build_test.sh
+      
+      - name: Execute Tests
+        run: "./tools/iceoryx_build_test.sh test"
```
elBoberido, line 44: Do you know if this is executed from a clean environment?
If yes, the whole build runs again, which might be time consuming and we should look if the artifacts from the previous build could be used.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Build with GNU
+        run: ./tools/iceoryx_build_test.sh
+      
+      - name: Execute Tests
+        run: "./tools/iceoryx_build_test.sh test"  
+        
+      - name : Build with clang
+        run: |
+          export CC=/usr/bin/clang
```
elBoberido, line 48: nice :)

.github/workflows/build-test.yml
elBoberido: as discussed, this will reuse the same environment and doesn't rebuild iceoryx for the test

prasannabhat: Individual steps run in the same environment. In this case only "test" is built incrementally.

prasannabhat:
@budrus @elBoberido 
Now we have the github action available.
We need to introduce additional status check in PR so as not to merge PR if the action fails.
Steps to [conifgure](https://help.github.com/en/github/administering-a-repository/about-required-status-checks)

Issue #76, opened 2020-03-24
Remove const_cast wherever possible


dkroenke:
**Brief feature description:** 
Check the necessity of all occurring const_casts and remove them if possible.

**Detailed information:** 
To check/improve const correctness in iceoryx we should take a look at all occurring const_casts and refactor them if possible.

elBoberido:
in some places its just like
```
int foo() const;
int foo();
```
where `foo` calls `foo() const`
this could probably be fixed by 
```
int foo() const;
int foo();
private:
int foo_impl() const;
```
then `foo()` and `foo() const` can just call `foo_impl() const`

PR #103, opened 2020-05-14
Iox #76 remove const cast wherever possible

elBoberido:
#76 implemented by @smmathad

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/internal/cxx/optional.inl
+++ b/iceoryx_utils/include/iceoryx_utils/internal/cxx/optional.inl
@@ -174,13 +174,13 @@ inline const T& optional<T>::operator*() const noexcept
 template <typename T>
 inline T* optional<T>::operator->() noexcept
 {
-    return const_cast<T*>(const_cast<const optional*>(this)->operator->());
+    return &value();
 }
 
 template <typename T>
 inline T& optional<T>::operator*() noexcept
 {
-    return const_cast<T&>(const_cast<const optional*>(this)->operator*());
+    return value();
 }
 
 template <typename T>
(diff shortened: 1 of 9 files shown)


iceoryx_utils/include/iceoryx_utils/internal/cxx/optional.inl
```diff
@@ -174,13 +174,13 @@ inline const T& optional<T>::operator*() const noexcept
...
 {
-    return const_cast<T*>(const_cast<const optional*>(this)->operator->());
+    return &value();
 }
 
 template <typename T>
 inline T& optional<T>::operator*() noexcept
 {
-    return const_cast<T&>(const_cast<const optional*>(this)->operator*());
+    return value();
```
elBoberido, line 183: to get rid of the the const_cast from `const T& value() const&` a `T& value_ref() const&` function could be introduced, which would contain the code from `T& value() &`. `m_data` must be made mutable though, like `alignas(alignof(T)) mutable byte_t m_data[sizeof(T)];`
What do you guys think?

elfenpiff: I think this approach makes the code even dirtier. If someone asks us why we have made ```m_data``` mutable our answer in the end would be to get rid of the ```const_cast``` which sounds just wrong.

Your solution would solve the issue but would generate new issues. I think mutable should only be used for really hidden structures inside a class like a mutex and not the actual data structure.

We should actually use const_cast to save ourself from code duplication! A detailed explanation on why we should always prefer this way is offered by Scott Meyers - Effective C++ 3rd Edition, Item 3

Issue #105, opened 2020-05-14
Make SOFI real size equal to the specified one


elBoberido:
## Required information

**Operating system:**
all

**Compiler version:**
all

**Observed result or behaviour:**
The SOFI holds one extry element before the overflowing mechanism takes action.
So in the single threaded case, a full SOFI always provides one more element than the configured size. This is a surprising behaviour.

**Expected result or behaviour:**
A full SOFI should not provide more elements than specified.

**Conditions where it occurred / Performed steps:**
- push to the SOFI until an overflow occurs
- pop until empty and count the pops
- the pop count should be equal to the configured size, but it's off by one

PR #107, opened 2020-05-14
Iox #105 make sofi real size equal to the specified one

elBoberido:


iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
```diff
@@ -139,46 +139,40 @@ inline bool SoFi<ValueType, CapacityValue>::popIf(ValueType& valueOut, const Ver
 template <class ValueType, uint32_t CapacityValue>
 bool SoFi<ValueType, CapacityValue>::push(const ValueType& valueOut, ValueType& f_paramOut_r) noexcept
 {
-    bool hasOverflow = false;
+    constexpr bool OVERFLOW{false};
 
     uint64_t currentWritePosition = m_writePosition.load(std::memory_order_relaxed);
     uint64_t nextWritePosition = currentWritePosition + 1U;
 
+    m_data[static_cast<int32_t>(currentWritePosition) % m_size] = valueOut;
```
elfenpiff, line 147: We should rename valueOut and paramOut to something more fitting like: inValue, outValue otherwise we are confusing people like me who are just wondering why the out value is written into the SoFi

iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
```diff
@@ -139,46 +139,40 @@ inline bool SoFi<ValueType, CapacityValue>::popIf(ValueType& valueOut, const Ver
...
-    // read his own data
-    if (hasOverflow)
+    // we need to update the read position
+    // a) it works, then we need to pass the overflow value back
+    // b) it doesn't work, which means that the pop thread already took the value in the meantime an no further action
+    // is required
+    if (m_readPosition.compare_exchange_strong(
+            currentReadPosition, nextReadPosition, std::memory_order_acq_rel, std::memory_order_relaxed))
     {
+        // no atomic synchronization required because writer can always read his own data
```
elfenpiff, line 170: I would disagree! We have the contract that one push and one pop can happen concurrently but we do not force the user to push always from the same core. What if:
> SoFi with 2 elements, 2 threads from 2 different cpu cores:
```
CPU 1 Thread 1: push(1)    - SoFi = {1}
CPU 1 Thread 1: push(2)    - SoFi = {1,2}
CPU 2 Thread 2: push(3)    - SoFi = {2,3} ... and this push has to synchronize with the previous push from CPU 1 to correctly return the value 1 
```

iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
```diff
@@ -139,46 +139,40 @@ inline bool SoFi<ValueType, CapacityValue>::popIf(ValueType& valueOut, const Ver
...
-        if (nextWritePosition < currentReadPosition + m_size)
-        {
-            hasOverflow = false;
-            break;
-        }
+        return !OVERFLOW;
+    }
 
-        nextReadPosition = currentReadPosition + 1U;
+    // this is an overflow situation, which means that the next push has no free position, therefore the oldesz value
```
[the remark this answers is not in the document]
elBoberido: everything knorke, that's Berlin style ;)

iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
elBoberido: That might have been an issue before this change, but now the access happens only if the CAS was successful which synchronizes with acq_rel.

Even before this change that shouldn't have been an issue.
Either you have to ensure that push is not called concurrently, which might require a mutex or another synchronization mechanism or you have to move the ownership to the SoFi to another thread which also requires a synchronisation mechanism.

Nevertheless, with the current changes, this comment is obsolete and should be removed

iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
elBoberido: should probably be done in a separate PR

PR #118, opened 2020-05-29
Iox #32 Initial github workflow for mac

prasannabhat:
Introduce initial github workflow for mac (build source & tests, running test cases would come later).

**Summary of changes**
- Separate job for ubuntu & mac
- Build script update with new options
_build-test_ : build tests , do not execute
_skip-introspection_ : Skip building iceoryx introspection (introspection doesnt build for mac, at the moment)


README.md
```diff
@@ -117,6 +117,17 @@ In default-mode, the script is not building the provided test. For a clean build
 
 If the script is not used, keep in mind to pass `-DTOML_CONFIG=on` to cmake if the optional RouDi config file feature shall be built.
 
+**NOTE:** Users who wish to build iceoryx with cmake can follow the below steps (Requires CMake 3.14 or higher).
```
dkroenke, line 120: is cmake 3.14 really required?

prasannabhat: For building the project using iceoryx_meta, cmake 3.14 is required (for building with script )

tools/introspection/CMakeLists.txt
```diff
@@ -55,10 +57,19 @@ target_link_libraries(iceoryx_introspection
     ncurses
 )
 
+
+
+# Find tinfo library for APPLE, for other platforms its part of package Curses
+if(CMAKE_SYSTEM_NAME MATCHES Darwin)
```
elBoberido, line 63: we have a platform detection in iceoryx_utils, just `include(IceoryxPlatformDetection)` and then you can do `if(APPLE)` ... maybe a ``find_package(iceoryx_utils REQUIRED)` is necessary before using the include

you could also add all OSes here and make an empty LIB_TINFO for QNX,  then the if statement below could be removed and target_link_libraries could be called directly

tools/introspection/CMakeLists.txt
```diff
@@ -34,6 +34,8 @@ set(${PROJECT_NAME}_DIR ${CMAKE_CURRENT_LIST_DIR}/cmake
     FORCE
 )
 
+# This is required for mac build (set to c++17 globally)
+add_compile_options(-std=c++11)
```
elBoberido, line 38: @elfenpiff didn't we find another solution for this? e.g. like let the linked library pass this?

README.md
elBoberido: this should be a temporary situation. I've a patch somewhere to make this build with cmake 3.5 ... need to remove the dust from it, though

tools/introspection/CMakeLists.txt
elfenpiff: We have to set the C++ standard via the property CXX_STANDARD for every target like this here: https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html

README.md
elfenpiff: For now I would leave it as it is since this is not our default build process ... not yet, but It should be.

tools/introspection/CMakeLists.txt
prasannabhat: This makes cmake file a little simpler.
Taken care

prasannabhat:
> Please use CXX_STANDARD 11 with set_property for the introspection target

This is taken care

Issue #119, opened 2020-06-02
Memory Synchronisation Issue in FiFo


elfenpiff:
## Required information

**Operating system:**
Every Operating System on a weak or C++ memory model

**Compiler version:**
All compiler

**Observed result or behaviour:**
The data which is pushed into the fifo is not synchronised with a different thread which pop's it out of the fifo.

**Expected result or behaviour:**
The data has to be synchronised.

**Conditions where it occurred / Performed steps:**
Unknown

PR #123, opened 2020-06-02
Iox #119 fix fifo memsync

elfenpiff:


iceoryx_utils/include/iceoryx_utils/internal/concurrent/fifo.inl
```diff
@@ -27,11 +27,14 @@ inline bool FiFo<ValueType, Capacity>::push(const ValueType& f_param_r)
     }
     else
     {
-        m_data[m_write_pos.load(std::memory_order_relaxed) % Capacity] = f_param_r;
+        auto currentWritePos = m_write_pos.load(std::memory_order_relaxed);
+        m_data[currentWritePos % Capacity] = f_param_r;
```
elBoberido, line 31: should we maybe create a local variable named "writeIndex"?

iceoryx_utils/include/iceoryx_utils/internal/concurrent/fifo.inl
```diff
@@ -51,17 +54,23 @@ inline bool FiFo<ValueType, Capacity>::empty() const
...
+                    // the producer pop - this is done here
+                    m_write_pos.load(std::memory_order_acquire));
+    if (isEmpty)
     {
         return cxx::nullopt_t();
     }
     else
     {
-        ValueType out = m_data[m_read_pos.load(std::memory_order_acquire) % Capacity];
+        auto currentReadPos = m_read_pos.load(std::memory_order_relaxed);
```
elBoberido, line 67: why do you read the read position here again? you already did it for the empty check, just reuse that one

elfenpiff: This was on purpose - I wanted to see if you can spot it ;)

iceoryx_utils/include/iceoryx_utils/internal/concurrent/fifo.inl
elfenpiff: I would stick to ```currentWritePos``` since it corresponds to ```m_write_pos```.

elBoberido: no, keep both
```
auto currentWritePos = m_write_pos.load(std::memory_order_relaxed);
auto writeIndex = currentWritePos % Capacity;
m_data[writeIndex] = f_param_r;
```

Issue #80, opened 2020-03-28
lock free mpmc queue


MatthiasKillat:
## Brief feature description

Create a lock-free multi producer multi consumer queue to be used in multicast scenarios (as in e.g. ROS) and for internal control flow.

## Detailed information

Properties:
1) lock-free and hence thread-safe
2) usable by an arbitrary number of multiple producers and consumers concurrently
3) no dynamic memory for safety reasons (uses its own memory buffer, determined at compile time, later configured during initialization)
4) robust -when an application terminates unexpectedly while using the queue 
the middleware daemon has to stay operational
5) suited to be placed in shared memory
6) low performance overhead when there is little contention

budrus:
is needed for #25

PR #125, opened 2020-06-03
Iox #80 lock free mpmc queue release version1

MatthiasKillat:
There will be minor corrections but there will also be a follow up pull request were the change capacity feature will be added to the queue. 

Note that adding the feature is relatively easy, but creating proper tests for operations when this feature is used will be more time consuming. 

This is the reason to do this in another pull request.

Another issue we need to consider is how to deal with long running/expensive tests, in this case the concurrent tests for the lock-free queue which run quite some time (>1min in total). This might warrant a separate target which is disabled by default in our internal Jenkins runs.


iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
```diff
@@ -0,0 +1,117 @@
...
+// @todo: unit tests - will be tested with the resize feature implementation
+// which might require minor adaptations/extensions of the CyclicIndex
+
+/// @brief index structure that can contain logical values 0, ..., CycleLength-1
+/// but also stores an internal cycle counter to be used in compare_exchange
+template <uint64_t CycleLength, typename ValueType = uint64_t>
+class CyclicIndex
+{
+  public:
+    using value_t = ValueType;
```
ithier, line 33: What is this for ?

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+        // even if we are not able to publish, we check whether some other push has already updated the writePosition
+        // before trying again to publish
+        auto writePositionRequiresUpdate = oldValue.getCycle() == writePosition.getCycle();
+
+        if (writePositionRequiresUpdate)
+        {
+            // case (2)
+            // the writePosition was not updated yet by another push but the value was already written
+            // help with the update
+            updateNextWritePosition(writePosition);
```
ithier, line 79: Why does the concurrent push need help ?

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
MatthiasKillat: This is to access the type of the template parameter. However, the naming is not ideal. If we have something like template<typename ValueType> we would sometimes like to use the type ValueType in other constructions that depend on it. But we cannot without a using directive, a shortcoming of the language in my opinion. But we also cannot have using ValueType = ValueType, we need a different name. What should this name be? We need a convention here I guess.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
MatthiasKillat: This probably requires more explanation in a comment, I think we cannot have code easily expressing this on its own (the ideal world). I will add this explanation, but the main reason is this: to complete a push operation.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
```diff
@@ -0,0 +1,117 @@
...
+        {
+            return otherCycle == 0;
+        }
+        return isBehind(thisCycle, otherCycle, 1);
+    }
+
+  private:
+    ValueType m_value{0};
+
+    static bool isBehind(ValueType cycle1, ValueType cycle2, ValueType cyclesBehind = 1)
```
elfenpiff, line 112: Why are we having here the extra parameter cyclesBehind? This method is only used once in isOnCycleBehind with the default argument 1. Since it is a one liner and with the context of isOnCycleBehind I would suggest of getting rid of this static function and move the line directly into isOnCycleBehind like:
```
return (thisCycle + 1 == otherCycle);
```
With this you even reduce your code in isOneCycleBehind ;)

I think this is overabstraction

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
elfenpiff, line 1: Please move implementation into separate inl file

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
```
elfenpiff, line 1: Please add the apache 2.0 header.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+
```
elfenpiff, line 2: Please move implementation into separate inl file

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+namespace iox
+{
+template <typename T>
+class unique
+{
+  public:
+    struct invalid_t
+    {
+        explicit constexpr invalid_t()
```
elfenpiff, line 13: An explicit constexpr constructor is not required and can be deleted in your use case.
The following line ```static constexpr invalid_t invalid{};``` will use the compiler generated constexpr ctor

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    static constexpr invalid_t invalid{};
+
+    template <typename... Args>
+    unique(Args&&... args)
+        : m_value(std::forward<Args>(args)...)
+    {
+    }
+
+    unique()
```
elfenpiff, line 26: Isn't this ctor ambiguous to the variadic template argument constructor? Are they both tested in unit tests?

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+        return m_value;
+    }
+
+    bool isValid() const
+    {
+        return m_valid;
+    }
+
+    // note: interferes with operator const T& for some types, we cannot have both(which do we want ?)
+    // operator bool()
```
elfenpiff, line 81: I think this operator would be always wrong. Let's say you have a ```unique<bool> a(false)``` but then you use it like that
```
if ( a ) {
  printf("true");
} else {
  printf("false");
}
```
It would print false even though the ```unique<bool> a``` is true since the bool returns if its set or not and this is very confusing.
I would leave it as it is and remove this note

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+template <typename T>
+class unique
+{
+  public:
+    struct invalid_t
+    {
+        explicit constexpr invalid_t()
+        {
+        }
+    };
```
elfenpiff, line 16: I have at the moment no idea why we need this class. This is the implementation of an ```cxx::optional```. I would strongly recommend removing this class and use cxx::optional since it is fully tested and it does exactly what this class does! Additionally, why is the name called unique?

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+    }
+
+    unique()
+        : m_value()
+    {
+    }
+
+    unique(invalid_t)
+        : m_valid(false)
```
elfenpiff, line 32: This is weird behavior. Lets say I would like to create a unique with an invalid value like
```
unique<ComplexObject> bla{unique<ComplexObject>::invalid};
```
Then the default constructor of the ComplexObject is still being called and the object is constructed. This would be very inefficient and unexpected

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+{
+  public:
+    struct invalid_t
+    {
+        explicit constexpr invalid_t()
+        {
+        }
+    };
+
+    static constexpr invalid_t invalid{};
```
elfenpiff, line 18: I recommend moving invalid outside of unique. Then you could use it more easily. Here take a look
```
unique<FatClassName> a(unique<FatClassName>::invalid);
// or better version
unique<FatClassName> a(invalid);
```

Since you are reimplementing an optional here. ```invalid``` corresponds to std::nullopt and the STL uses this approach

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+        , m_valid(other.m_valid)
+    {
+        other.m_valid = false;
+    };
+
+    unique& operator=(unique&& other)
+    {
+        if (this != &other)
+        {
+            m_value = std::move(other.m_value);
```
elfenpiff, line 52: What happens when other is invalid. Then this move would be undefined behavior!

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+    }
+
+    // no copies
+    unique(const unique&) = delete;
+    unique& operator=(const unique&) = delete;
+
+    // only move
+    unique(unique&& other)
+        : m_value(std::move(other.m_value))
```
elfenpiff, line 42: What happens when other is invalid then this move is undefined behavior

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    const T& operator*() const
+    {
+        return m_value;
+    }
+
+    T release()
+    {
+        m_valid = false;
```
elfenpiff, line 66: What happens when it is in an invalid state? Then this move would be undefined behavior and this is forbidden in our world.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <atomic>
+
+namespace iox
+{
+/// @brief lockfree queue capable of storing indices 0,1,... Capacity-1
+template <uint64_t Capacity, typename ValueType = uint64_t>
+class IndexQueue
+{
+  public:
+    class UniqueIndex : public unique<ValueType>
```
elfenpiff, line 30: Please inherit from ```cxx::optional<ValueType>``` instead of unique. The reasons are
- it does the same like unique
- it is very well tested, reviewed and already proven in use
- unique has undefined behavior and bugs

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    struct ConstructEmpty_t
+    {
+    };
+
+    // @todo: a compile time check whether Index is actually lock free would be nice
+    // note: there is a way  with is_always_lock_free in c++17 (which we cannot use here)
+
+    static constexpr ConstructFull_t ConstructFull{};
```
elfenpiff, line 66: When you move it outside of the index queue then the index queue can be used more easily. Take a look at the std::nullopt from std::optional<T> for example. Thats how the STL uses this concept.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    /// @return capacity of the IndexQueue
+    /// threadsafe, lockfree
+    constexpr uint64_t capacity();
+
+    /// @brief check whether the queue is empty
+    /// @return true iff the queue is empty
+    /// note that if the queue is used concurrently it might
+    /// not be empty anymore after the call
+    /// (but it was at some point during the call)
+    bool empty();
```
elfenpiff, line 90: Please make this method const

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    /// @brief constructs an empty IndexQueue
+    IndexQueue(ConstructEmpty_t = ConstructEmpty);
+
+    /// @brief constructs IndexQueue filled with all indices 0,1,...capacity-1
+    IndexQueue(ConstructFull_t);
+
+    /// @brief get the capacity of the IndexQueue
+    /// @return capacity of the IndexQueue
+    /// threadsafe, lockfree
+    constexpr uint64_t capacity();
```
elfenpiff, line 83: I think this method should also be const.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    using Cell = std::atomic<Index>;
+    Cell m_cells[Capacity];
+
+    std::atomic<Index> m_readPosition;
+    std::atomic<Index> m_writePosition;
+
+  private:
+    Index loadNextReadPosition() const;
+    Index loadNextWritePosition() const;
+    Index loadValueAt(Index position) const;
```
elfenpiff, line 130: Please add const to the argument

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    bool tryToPublishAt(Index writePosition, Index& oldValue, Index newValue);
```
elfenpiff, line 131: Please add const to the arguments

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    bool tryToGainOwnershipAt(Index& readPosition);
+    void updateNextWritePosition(Index& oldWritePosition);
+
+    // internal raw value (ValueType) interface
+    // private, since it does not prevent multiple of the same index pushes by design
+
+    // push index into the queue in FIFO order
+    void push(ValueType index);
```
elfenpiff, line 139: Please add const to the argument

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return toPtr(index);
+    }
+
+    const ElementType* ptr(index_t index) const noexcept
+    {
+        return toPtr(index);
+    }
+
+    uint64_t capacity()
```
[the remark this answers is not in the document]
elfenpiff: And noexcept

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
```diff
@@ -0,0 +1,117 @@
...
+    {
+        return m_value % CycleLength;
+    }
+
+    ValueType getCycle() const noexcept
+    {
+        return m_value / CycleLength;
+    }
+
+    CyclicIndex operator+(ValueType value) const
```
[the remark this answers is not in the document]
elfenpiff: And noexcept

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
```diff
@@ -0,0 +1,117 @@
...
+            // but this has other drawbacks (and the overflow will not occur often if at all with 64 bit)
+            delta = m_value - delta - 1;
+            return CyclicIndex(OVERFLOW_START_INDEX + delta);
+        }
+
+        // no overflow, regular case
+        return CyclicIndex(m_value + value);
+    }
+
+    CyclicIndex next()
```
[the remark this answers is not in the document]
elfenpiff: And add noexcept

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
```diff
@@ -0,0 +1,117 @@
...
+    {
+        if (m_value == MAX_VALUE)
+        {
+            return CyclicIndex(OVERFLOW_START_INDEX);
+        }
+        return CyclicIndex(m_value + 1);
+    }
+
+    bool isOneCycleBehind(const CyclicIndex& other)
```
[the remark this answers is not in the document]
elfenpiff: and add noexcept

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+    {
+      public:
+        using value_t = ValueType;
+        using invalid_t = typename unique<ValueType>::invalid_t;
+        static constexpr invalid_t invalid{};
+
+        friend class IndexQueue<Capacity, ValueType>;
+
+        // only an invalid index can be constructed by anyone other than the IndexQueue itself
+        UniqueIndex(invalid_t)
```
elfenpiff, line 40: Please add noexcept to every method in this file

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+    , m_writePosition(Index(Capacity))
+{
+}
+
+template <uint64_t Capacity, typename ValueType>
+IndexQueue<Capacity, ValueType>::IndexQueue(ConstructFull_t)
+    : m_readPosition(Index(0))
+    , m_writePosition(Index(Capacity))
+{
+    for (uint64_t i = 0; i < Capacity; ++i)
```
elfenpiff, line 15: We have to write ```uint64_t i = 0u``` otherwise stupid tools will babble about it

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+template <uint64_t Capacity, typename ValueType>
+constexpr uint64_t IndexQueue<Capacity, ValueType>::capacity()
+{
+    return Capacity;
+}
+
+template <uint64_t Capacity, typename ValueType>
+void IndexQueue<Capacity, ValueType>::push(ValueType index)
+{
+    // we need the CAS loop here since we may fail due to concurrent push operations
```
elfenpiff, line 30: You are writing here is somewhere a CAS loop hidden?! I have to check every method call inside this method to get the memory order and find all the atomic access and this really irritates me and makes these lock free code even harder to understand, to fix and when we do mistakes with this approach they are even harder to spot.
Please do not do this.

My wish would be that every relevant atomic operation is right in place in this function

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+    auto oldReadIndex = m_readPosition.load(std::memory_order_relaxed);
+    auto value = m_cells[oldReadIndex.getIndex()].load(std::memory_order_relaxed);
+
+    // if m_readPosition is ahead by one cycle compared to the value stored at head,
+    // the queue was empty at the time of the loads above (but might not be anymore!)
+    return value.isOneCycleBehind(oldReadIndex);
+}
+
+template <uint64_t Capacity, typename ValueType>
+typename IndexQueue<Capacity, ValueType>::Index IndexQueue<Capacity, ValueType>::loadNextReadPosition() const
```
elfenpiff, line 200: Please get rid of all these unnecessary atomic abstractions. This makes the code ultra hard to review since you have to jump around all the time and brings no benefit. Lock free code is hard to understand but when you call
```
m_readPosition.load()
```
it is pretty clear to me that you loadNextReadPosition. Its the same which is already written above. But the memory order is important. I would like to understand and confirm that your memory_order_relaxed is correct but this jumping around makes it hard for me to understand

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+{
+    return m_readPosition.load(std::memory_order_relaxed);
+}
+
+template <uint64_t Capacity, typename ValueType>
+typename IndexQueue<Capacity, ValueType>::Index IndexQueue<Capacity, ValueType>::loadNextWritePosition() const
```
elfenpiff, line 206: Please get rid of this unnecessary abstraction. Same argument as in loadNextReadPosition

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+{
+    return m_writePosition.load(std::memory_order_relaxed);
+}
+
+template <uint64_t Capacity, typename ValueType>
+typename IndexQueue<Capacity, ValueType>::Index
+IndexQueue<Capacity, ValueType>::loadValueAt(typename IndexQueue<Capacity, ValueType>::Index position) const
```
elfenpiff, line 213: Please get rid of this unnecessary abstraction, same argument as in loadNextReadPosition

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+{
+    return m_cells[position.getIndex()].load(std::memory_order_relaxed);
+}
+
+template <uint64_t Capacity, typename ValueType>
+bool IndexQueue<Capacity, ValueType>::tryToPublishAt(typename IndexQueue<Capacity, ValueType>::Index writePosition,
```
elfenpiff, line 219: Please get rid of this unnecessary abstraction, same argument as in loadNextReadPosition

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+                                                     Index& oldValue,
+                                                     Index newValue)
+{
+    return m_cells[writePosition.getIndex()].compare_exchange_strong(
+        oldValue, newValue, std::memory_order_relaxed, std::memory_order_relaxed);
+}
+
+template <uint64_t Capacity, typename ValueType>
+void IndexQueue<Capacity, ValueType>::updateNextWritePosition(Index& writePosition)
```
elfenpiff, line 228: Please get rid of this unnecessary abstraction, same argument as in loadNextReadPosition

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,274 @@
...
+{
+    // compare_exchange updates writePosition if NOT successful
+    Index newWritePosition(writePosition + 1);
+    auto updateSucceeded = m_writePosition.compare_exchange_strong(
+        writePosition, newWritePosition, std::memory_order_relaxed, std::memory_order_relaxed);
+}
+
+template <uint64_t Capacity, typename ValueType>
+bool IndexQueue<Capacity, ValueType>::tryToGainOwnershipAt(Index& oldReadPosition)
```
elfenpiff, line 237: Please get rid of this unnecessary abstraction, same argument as in loadNextReadPosition

iceoryx_utils/include/iceoryx_utils/lockfree_queue/lockfree_queue.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include "iceoryx_utils/cxx/optional.hpp"
+#include "iceoryx_utils/internal/lockfree_queue/buffer.hpp"
+#include "iceoryx_utils/internal/lockfree_queue/index_queue.hpp"
+
+#include <atomic>
+
+
+namespace iox
+{
+/// @todo finalize interface, configuration of actual capacity at runtime (change capacity feature)
```
elfenpiff, line 26: Please remove the todo are change it into a note where you list upcoming features

iceoryx_utils/include/iceoryx_utils/lockfree_queue/lockfree_queue.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /// but unsafe copying (i.e. where synchronization is up to the user) would be possible
+    /// can be implemented when it is needed
+    LockFreeQueue(const LockFreeQueue&) = delete;
+    LockFreeQueue(LockFreeQueue&&) = delete;
+    LockFreeQueue& operator=(const LockFreeQueue&) = delete;
+    LockFreeQueue& operator=(LockFreeQueue&&) = delete;
+
+    /// @brief returns the capacity of the queue
+    /// threadsafe, lockfree
+    constexpr uint64_t capacity() noexcept;
```
elfenpiff, line 47: I think you should make this method const

iceoryx_utils/include/iceoryx_utils/lockfree_queue/lockfree_queue.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /// @return value if removal was successful, empty optional otherwise
+    /// threadsafe, lockfree
+    iox::cxx::optional<ElementType> pop() noexcept;
+
+    /// @brief check whether the queue is empty
+    /// @return true iff the queue is empty
+    /// note that if the queue is used concurrently it might
+    /// not be empty anymore after the call
+    ///  (but it was at some point during the call)
+    bool empty();
```
elfenpiff, line 85: Please make this method const

iceoryx_utils/include/iceoryx_utils/lockfree_queue/lockfree_queue.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+    /// @brief get the number of stored elements in the queue
+    /// @return number of stored elements in the queue
+    /// note that this will not be perfectly in sync with the actual number of contained elements
+    /// during concurrent operation but will always be at most capacity
+    uint64_t size();
```
elfenpiff, line 91: Please make this method const

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  private:
+    using byte_t = uint8_t;
+
+    alignas(alignof(ElementType)) byte_t m_buffer[Capacity * sizeof(ElementType)];
+
+    ElementType* toPtr(index_t index) const noexcept
+    {
+        auto ptr = &(m_buffer[index * sizeof(ElementType)]);
+        return reinterpret_cast<ElementType*>(const_cast<byte_t*>(ptr));
```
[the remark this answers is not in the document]
ithier: It might be worth running the toolchain to check if this is really a problem. 
I had something similar which was not flagged.

iceoryx_utils/include/iceoryx_utils/lockfree_queue/readme.md
```diff
@@ -0,0 +1,140 @@
...
+### Capacity and Cycle Length
+The index queue has some capacity n and stores indices in the range [0, n[.
+The number n is also called cycle length.
+
+Assume n = 4 in this explanation, so we can store the indices 0, 1, 2, 3.
+
+
+### Index representation
+Each index i corresponds to an equivalence class [i] modulo n, i.e. 
+[i] = {j | j = c*n + i, n >= 0}
```
FerdinandSpitzschnueffler, line 18: c is not yet defined

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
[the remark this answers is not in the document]
MatthiasKillat: We had a similar case in the code and decided to go with the const_cast to avoid code duplication. I would prefer leaving it like this (I question the toolchain in this regard and we should consider dropping those rules if they cause problems...)

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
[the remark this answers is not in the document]
MatthiasKillat: Regarding the atomic abstractions: readability is fairly subjective. I see pros and cons for them, really hard to decide. It makes sense to have helper functions with expressive names to get an idea what is going on. When the details are of interest, one has to look into the functions. Personally I favor keeping them as is, but will wait for other opinions...

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
```diff
@@ -0,0 +1,53 @@
+#pragma once
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, can be done. It is only a little helper class as of now, but might get extended with cacheline optimzations later. The question is a more general one: do we really want to have inl files for every little helper template? We are not entirely consistent across the codebase in this regard, but we should strive for an agreement in the long run (other coding rules may also need some team review).

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <stdint.h>
+
+namespace iox
+{
+// @todo: add more functionality (policies for cache line size, redzoning), rename in a suitable way
```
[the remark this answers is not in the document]
MatthiasKillat: Will be renamed, it exists as a reminder that there is extension potential (possibly in the near future).

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
MatthiasKillat: Same counter argument as well.

iceoryx_utils/include/iceoryx_utils/lockfree_queue/lockfree_queue.hpp
MatthiasKillat: Yes

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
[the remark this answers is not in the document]
MatthiasKillat: Yes, it just serves as a reminder for extension potential.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    struct ConstructFull_t
+    {
+    };
+
+    struct ConstructEmpty_t
+    {
+    };
+
+    // @todo: a compile time check whether Index is actually lock free would be nice
```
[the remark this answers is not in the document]
MatthiasKillat: By the way, this compile time check would be possible with is_always_lock_free in c++17. Would be nice to be able to use this...oh well.

iceoryx_utils/test/moduletests/test_lockfree_queue_cyclic_index.cpp
```diff
@@ -0,0 +1,381 @@
...
+    const auto i = Index::MAX_INDEX;
+    const auto c = Index::MAX_CYCLE / 2;
+    const auto m = Index::MAX_INDEX + 1;
+
+    Index cycleStart(0, c);
+    Index cycleMid(m / 2, c);
+    Index cycleEnd(i, c);
+
+    Index nextCycleStart = cycleStart + m;
+    Index nextCycleMid = cycleMid + m;
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, a copy and paste error. Those tests are important to check that the overflow works correctly (when we are using 64 bit this is more a theoretical issue). Would be nice if someone could take a look whether the tests are plausible, I guess the helper variables i, c (for brevity) are not really expressive. But they are only used at a small scope were I prefer this for readability.

iceoryx_utils/include/iceoryx_utils/lockfree_queue/readme.md
MatthiasKillat: I will correct this and rework the whole explanation together with the resize feature.

iceoryx_utils/test/moduletests/test_lockfree_queue_cyclic_index.cpp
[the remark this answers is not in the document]
MatthiasKillat: Yes, a copy and paste error. Those tests are important to check that the overflow works correctly (when we are using 64 bit this is more a theoretical issue). Would be nice if someone could take a look whether the tests are plausible, I guess the helper variables i, c (for brevity) are not really expressive. But they are only used at a small scope were I prefer this for readability.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
```diff
@@ -0,0 +1,53 @@
...
+template <typename ElementType, uint64_t Capacity, typename index_t = uint64_t>
+class Buffer
+{
+  public:
+    Buffer() = default;
+    Buffer(const Buffer&) = delete;
+    Buffer& operator=(const Buffer&) = delete;
+
+    ElementType& operator[](index_t index) noexcept
```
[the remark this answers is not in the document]
MatthiasKillat: I will do this but *only* where it is constistent with the implation (i.e. it does not change the value). I am strictly against making it const everywhere and losing the potential of changing the variable in the function. Sometimes an elegant implementation will require this.

iceoryx_utils/include/iceoryx_utils/lockfree_queue/readme.md
MatthiasKillat: should be c>=0, oh well

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
```diff
@@ -0,0 +1,117 @@
...
+    static_assert(CycleLength < MAX_VALUE / 2); // need at least one bit for the cycle
+    static_assert(CycleLength > 0);
+
+    explicit CyclicIndex(ValueType value = 0) noexcept
+        : m_value(value)
+    {
+    }
+
+    CyclicIndex(ValueType index, ValueType cycle) noexcept
+        : m_value(index + cycle * CycleLength)
```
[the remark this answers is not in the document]
MatthiasKillat: Does not really matter here, depending on the optimization might be even less efficient I guess

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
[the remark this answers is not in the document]
MatthiasKillat: Possible here but I consider const value types bad style.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/cyclic_index.hpp
MatthiasKillat: This is artifact from a while ago when there were more "combinations" of cycle and CyclicIndex object argument. There it reduced some duplication, but is not needed any more.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
MatthiasKillat: It is also a bit more heavyweight for this case than I would like it to be. Will take a look, the unique construction may be removed altogether because it costs performance while not adding much (a bit "safety" in the internals) .

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
MatthiasKillat: I disagree here. This is used mainly internally for now and would pollute the namespace. For example nullopt_t in iceoryx is defined in the namespace cxx. It is also a bit different here I would argue. Assume I want to do something similar with another container, how would the struct be called then? ConstructFull_t would already be in use, so we would have to introduce a namespace or choose another name (which will get unwieldy fast). No, this ConstructFull_t is strongly related to the IndexQueue itself and therefore best placed in its implicit namespace.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/buffer.hpp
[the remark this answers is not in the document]
mossmaurice: Yeah, separating the declaration from the definition is a must for templates as well IMHO.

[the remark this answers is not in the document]
MatthiasKillat: Will do, but I am not convinced it is good practice for helper classes which are not part of the public interface (i.e. internal hpp files). This will clutter up the filestructure (ok, this can be mitigated with a good file browser) but more importantly lead to more/longer lines of code which carry not much meaning due to the necessary template and namespace qualifiers in the inl  files.

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
MatthiasKillat: I am against doing this for value types all over the place without critical analysis. We will refine our static analysis ruleset to drop this requirement should this be a problem. This is not really needed and as far as I know does only very rarely offer optimization opportunities to the compiler it would not use anyway. @mossmaurice

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
MatthiasKillat: yeah, I am sure if the ranges and signs fit it is no problem, I am not sure if cast overhead is introduced. For integral template types this is especially problematic...we do not know the instantiated type and length there (do we need 0ull, or 0u or whatever).

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
MatthiasKillat: The CAS loop is the main and only loop of the method. The comment shall indicate that it is needed because the operation can fail due to concurrent operations. The code abstraction is very much a matter of opinion. It basically was the way you mention a long long time ago...when dinosaurs walked the earth (in march this year if i recall correctly). We can argue both ways, the way it is, the messy details are hidden and (maybe) the intention of the call captured by its name (the call overhead will most likely be optimized away). On the other hand it hides important algorithmic details inside those helpers. So, Clean Code ... quo vadis?

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
MatthiasKillat: A matter that can be argued back and forth, unfortunately. It depends on the abstraction level we have in mind when we look at the code. I will take a look whether this can be made easier, but we may need to discuss this. @elfenpiff

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.hpp
mossmaurice: I agree with you. You can hardly ever generalise. The philosophical discussion about clean code is very important as well. However, before running off the track: Please just ran QAC++ tomorrow and see if coding rules are violated here. Const correctness is not always the highest good IMHO.

Nitpick: Before claiming that abstraction costs performance, have you measured it? ;-)

MatthiasKillat: Abstraction is almost never free, sometimes it does not cost you performance but you still pay something, such as longer compile time. But here it is mainly about const ness of value types, which has no real upsides and is just cluttering the interface. Still, there is no universal agreement. See for example this discussion https://stackoverflow.com/questions/8532095/is-it-a-good-idea-of-maintaining-const-ness-as-much-as-possible. Just to be clear, I am all for const correctness of methods and reference arguments (but still think C++ is to verbose here). But most importantly it is an information for the caller and, for references, helps to enforce the contract. For value types it provides not many optimization opportunities for the compiler, since it has to do an analysis if the value is really not changed in any case (constness can be circumvented).

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/index_queue.inl
MatthiasKillat: It should also e noted that if no push comes to help, it is not a problem since this means there are no pushes needing the update. Pops are unaffected by the incomplete operation, they do not require writePosition (well popIfFull does, but that is more complicated, this one would fail which would be correct, because technically the incomplete push means the queue cannot be full...).

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
MatthiasKillat: unique is currently not tested separately but only in its use UniqueIndex. The ctor is redundant. I considered implementing it completely in UniqueIndex itself, but I think it makes more sense to derive from some common base class and adapt the functionality in UniqueIndex a little (mainly to limit construction to the IndexQueue itself, otherwise I would have used unique directly).

iceoryx_utils/include/iceoryx_utils/internal/lockfree_queue/unique.hpp
MatthiasKillat: Yeah, this is just a leftover. I mainly wanted to have the object behave like the underlying type T, using implicit conversion. It is also intended to be used with small types T, otherwise this gets expensive. To be true, this abstraction does more harm than good (performance and maintainance-wise)

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,249 @@
...
+            // case (3) and (4)
+            // note: we do not update with CAS here, the CAS is bound to fail anyway
+            // (since our value of writePosition is not up to date so needs to be loaded again)
+            writePosition = m_writePosition.load(std::memory_order_relaxed);
+        }
+
+    } while (notPublished);
+
+    Index newWritePosition(writePosition + 1);
+    m_writePosition.compare_exchange_strong(
```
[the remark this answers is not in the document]
MatthiasKillat: Yes. Interestly static code analysis did not complain about the ignored return value...

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,249 @@
...
+    // (3) loaded value is more than one cycle behind:
+    //     this should only happen due to wrap around when push is interrupted for a long time
+    //     reload write position and try again
+    //     note that a complete wraparound can lead to a false detection of 1) (ABA problem)
+    //     but this is very unlikely with e.g. a 64 bit value type
+    // (4) loaded value is some cycle ahead:
+    //     write position is outdated, there must have been other pushes concurrently
+    //     reload write position and try again
+
+    bool notPublished = true;
```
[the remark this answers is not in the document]
MatthiasKillat: Oversight during refactoring, will change.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -0,0 +1,255 @@
...
+    // (3) loaded value is more than one cycle behind:
+    //     this should only happen due to wrap around when push is interrupted for a long time
+    //     reload write position and try again
+    //     note that a complete wraparound can lead to a false detection of 1) (ABA problem)
+    //     but this is very unlikely with e.g. a 64 bit value type
+    // (4) loaded value is some cycle ahead:
+    //     write position is outdated, there must have been other pushes concurrently
+    //     reload write position and try again
+
+    constexpr bool notPublished = true;
```
elBoberido, line 52: according to our style guilde, constants should be UPPER_CAMEL_CASE

MatthiasKillat: Yes, but this just looks bad here...I would prefer a while true loop over that. Another commit just for this seems excessive.

iceoryx_utils/doc/lockfree_queue/lockfree_queue.md
```diff
@@ -0,0 +1,140 @@
...
+Afterwards Tail is increased by 1 (potentially not immediately, but before the next push takes effect).
+```
+[ (c,?), (c,x), (c-1,?), (c-1,?) ]   H=(c,1)     T=(c,2)
+```
+push(y)
+```
+[ (c,?), (c,x), (c,y), (c-1,?) ]   H=(c,1)     T=(c,3)
+```
+
+We only push iff the cycle at the element Tail points to is exactly one behind Tails cycle.
```
elBoberido, line 66: typo `iff`

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.inl
elBoberido: I would also be fine with UpperCamelCase

iceoryx_utils/doc/lockfree_queue/lockfree_queue.md
```diff
@@ -0,0 +1,140 @@
+# Lock Free queue
```
mossmaurice, line 1: Add copyright notice?

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.hpp
```diff
@@ -0,0 +1,204 @@
...
+    /// @param index to be pushed, any index can only be pushed once by design
+    void push(UniqueIndex& index) noexcept;
+
+    /// @brief tries to remove index in FIFO order
+    /// threadsafe, lockfree
+    /// @return valid UniqueIndex if removal was successful (i.e. queue was not empty),
+    /// invalid UnqiueIndex otherwise
+    UniqueIndex pop() noexcept;
+
+    /// @brief tries to remove index in FIFO order iff the queue is full
```
mossmaurice, line 168: Typo

iceoryx_utils/doc/lockfree_queue/lockfree_queue.md
MatthiasKillat: Not really: https://en.wikipedia.org/wiki/If_and_only_if

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.inl
MatthiasKillat: Technically reserved for types...oh well.

iceoryx_utils/doc/lockfree_queue/lockfree_queue.md
MatthiasKillat: It is not a source code file but can be done.

iceoryx_utils/doc/lockfree_queue/lockfree_queue.md
MatthiasKillat: But may remove it for consistency with everything. This file will be reworked during capacity change feature anyway.

Issue #121, opened 2020-06-02
Update iceoryx version to 0.17.0.2


elBoberido:
## Brief feature description

update iceroryx version to 0.17.0.2

## Detailed information

preparation for release

PR #134, opened 2020-06-09
iox-#121 Store version in central file

elBoberido:
Signed-off-by: Ithier Jeff (CC-AD/EYF1) <jeff.ithier@de.bosch.com>


iceoryx_meta/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 cmake_minimum_required(VERSION 3.14)
-set(iceoryx_VERSION 0.17.0.2)
-
+file (STRINGS "../VERSION" iceoryx_VERSION)
```
elfenpiff, line 2: I do not like this relative path but I think there is no alternative?! @elBoberido you are our CMake hero! Please help us. do you know stuff?

ithier: Yeah I also wasn't too happy about it. Would be glad to learn an alternative.

elBoberido: I also don't have a better idea. The only other idea I have is to use the VERSION file in iceoryx_utils and iceoryx_posh and use a script to change the version number. But that also no nice solution

PR #138, opened 2020-06-15
Iox #32 Mac OS fully supported

elfenpiff:


iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/argv_inspection.hpp
```diff
@@ -44,7 +44,7 @@ class ArgvInspector
 
   private:
     cxx::optional<std::string> m_cmdline;
-};
+} m
```
Karsten1987, line 47: ?

elfenpiff:
> This doesn't compile for me:
> 
> ```
> --- stderr: iceoryx_utils
> /Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/iceoryx_utils/source/posix_wrapper/timer.cpp:65:15: warning: result of comparison of constant 16777215 with expression of type 'unsigned char' is always false [-Wtautological-constant-out-of-range-compare]
>     if (index >= Timer::OsTimerCallbackHandle::MAX_DESCRIPTOR_VALUE)
>         ~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 1 warning generated.
> ```
> 
> ```
> -- Build files have been written to: /Users/karsten/workspace/ros2/iceoryx_ws/osx/build/icedelivery
> CMake Error at CMakeLists.txt:30 (find_package):
>   Could not find a package configuration file provided by
>   "CycloneDDS_CXX_API" with any of the following names:
> 
>     CycloneDDS_CXX_APIConfig.cmake
>     cyclonedds_cxx_api-config.cmake
> 
>   Add the installation prefix of "CycloneDDS_CXX_API" to CMAKE_PREFIX_PATH or
>   set "CycloneDDS_CXX_API_DIR" to a directory containing one of the above
>   files.  If "CycloneDDS_CXX_API" provides a separate development package or
>   SDK, be sure it has been installed.
> 
> 
> -- Configuring incomplete, errors occurred!
> See also "/Users/karsten/workspace/ros2/iceoryx_ws/osx/build/iceoryx_dds_gateway/CMakeFiles/CMakeOutput.log".
> --- stderr: iceoryx_dds_gateway
> ```
> 
> ```
> -- Found Curses: /usr/lib/libcurses.dylib
> CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> Please set them or make sure they are set and tested correctly in the CMake files:
> LIB_TINFO
>     linked by target "iceoryx_introspection" in directory /Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/tools/introspection
> 
> -- Configuring incomplete, errors occurred!
> See also "/Users/karsten/workspace/ros2/iceoryx_ws/osx/build/iceoryx_introspection/CMakeFiles/CMakeOutput.log".
> --- stderr: iceoryx_introspection
> CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> Please set them or make sure they are set and tested correctly in the CMake files:
> LIB_TINFO
>     linked by target "iceoryx_introspection" in directory /Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/tools/introspection
> 
> ---
> Aborted  <<< iceoryx_introspection
> ```

@Karsten1987 Thanks for the hint. I think we have to build IceOryx maybe differently on MacOS but we have to document this first.
Could you please try:
```
cd iceoryx
cmake -Bbuild -Hiceoryx_meta -Dtest=ON -DTOML_CONFIG=ON
cd build
make -j8
```
This should work and in the future we should only use this way but I will take a look at the script for now and maybe I can fix it.

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -238,6 +241,7 @@ MqRuntimeInterface::MqRuntimeInterface(const std::string& roudiName,
 
     int64_t transmissionTimestamp{0};
     auto regState = RegState::WAIT_FOR_ROUDI;
+    // regState = RegState::SEND_REGISTER_REQUEST;
```
mossmaurice, line 244: Please remove commented-out code

Karsten1987:
@elfenpiff that works, but I'll get quite a handful of warnings such as below:

```
--- stderr: iceoryx_utils
/Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/iceoryx_utils/source/posix_wrapper/timer.cpp:65:15: warning: result of comparison of constant 16777215 with expression of type 'unsigned char' is always false [-Wtautological-constant-out-of-range-compare]
    if (index >= Timer::OsTimerCallbackHandle::MAX_DESCRIPTOR_VALUE)
        ~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```
```
[ 98%] Linking CXX executable ../../posh/test/posh_integrationtests
/Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/iceoryx_utils/test/moduletests/test_cxx_string.cpp:31:1: warning: 'TypedTestCaseIsDeprecated' is deprecated: TYPED_TEST_CASE is deprecated, please use TYPED_TEST_SUITE [-Wdeprecated-declarations]
TYPED_TEST_CASE(stringTyped_test, Implementations);
^
/Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/build/_deps/googletest-src/googletest/include/gtest/gtest-typed-test.h:227:38: note: expanded from macro 'TYPED_TEST_CASE'
  static_assert(::testing::internal::TypedTestCaseIsDeprecated(), ""); \
                                     ^
/Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/build/_deps/googletest-src/googletest/include/gtest/internal/gtest-internal.h:1217:1: note: 'TypedTestCaseIsDeprecated' has been explicitly marked deprecated here
GTEST_INTERNAL_DEPRECATED(
^
/Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/build/_deps/googletest-src/googletest/include/gtest/internal/gtest-port.h:2224:59: note: expanded from macro 'GTEST_INTERNAL_DEPRECATED'
#define GTEST_INTERNAL_DEPRECATED(message) __attribute__((deprecated(message)))
                                                          ^
1 warning generated.
```

I would strongly advice against a custom script, iceoryx has to be able to be built via plain CMake to provide compatibility with Colcon. Otherwise supporting `rmw_iceoryx` and ROS2 so to say will be quite tricky. I should be able to put iceoryx into my ROS2 workspace and just build it.

elfenpiff:
> I would strongly advice against a custom script, iceoryx has to be able to be built via plain CMake to provide compatibility with Colcon. Otherwise supporting `rmw_iceoryx` and ROS2 so to say will be quite tricky. I should be able to put iceoryx into my ROS2 workspace and just build it.

@Karsten1987 Alright, the documentation is written and Ice0ryx should be buildable with the CMakeLists.txt stored in iceoryx_meta without warnings except one in TOML. If you would like to use the introspection client you can also find an installation description for the ncurses requirement directly in the readme.md in iceoryx. The only thing left to fix are these funky timing test ...

Karsten1987:
If I manually skip `iceoryx_introspection` and `iceoryx_dds_gateway` my workspace compiles locally.

I would recommend having something along the following lines in the CMakeLists.txt in these unsupported packages:

```cmake
if(NOT UNIX)
  message(WARNING "${PROJECT_NAME} is currently only supported on Linux")
  return()
endif(UNIX)
```

elBoberido:
@elfenpiff iceoryx_meta is supposed to be used by developers to easily start contributing, not for tools like colcon. colcon should build iceoryx_utils and iceoryx_posh separately. The introspection client resides in tools/introspection, so we either have to put a COLCON_IGNORE file in there or the ncurses dependency has to be found on the system.

iceoryx_posh/test/mocks/chunk_mock.hpp
```diff
@@ -50,9 +58,10 @@ class ChunkMock
         {
             m_chunkHeader->~ChunkHeader();
         }
-        if (m_rawMemory != nullptr)
+        if (m_unalignedMemory != nullptr)
```
elBoberido, line 61: this is weird. You allocate to m_rawMemory and free m_unalignedMemory. How about reverting this and introducing an m_alignedMemory
```
    m_rawMemory = static_cast<uint8_t*>(malloc(Size + Alignment));
    auto memAddress = reinterpret_cast<uint64_t>(m_rawMemory);
    auto offset = memAddress % Alignment;
    auto alignedAddress =  offset = 0 ? memAddress : memAddress + Alignment - offset;
    m_alignedMemory = reinterpret_cast<uint8_t*>(alignedAddress );
```

This has also the benefit to fix the bugs in the implementation ;)

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
```diff
@@ -42,6 +42,11 @@ class MemoryProviderTestImpl : public iox::roudi::MemoryProvider
 
 #if defined(QNX) || defined(QNX__) || defined(__QNX__)
         dummyMemory = static_cast<uint8_t*>(memalign(alignment, size));
+#elif defined(__APPLE__)
+        uint64_t memOffset = reinterpret_cast<uint64_t>(malloc(size + alignment));
+        dummyMemory = reinterpret_cast<uint8_t*>(memOffset);
+        memOffset += memOffset % alignment;
```
elBoberido, line 48: this has bugs, please use the proposal from chunck_mock.hpp

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/argv_inspection.hpp
```diff
@@ -44,7 +44,7 @@ class ArgvInspector
```
elBoberido, line 44: since you deleted the corresponding cpp file, I guess this should also be deleted

iceoryx_utils/platform/mac/include/iceoryx_utils/platform/acl.hpp
```diff
@@ -33,17 +33,17 @@ using acl_perm_t = int;
 using acl_entry_t = int;
 using acl_tag_t = int;
 
-inline int acl_valid(acl_t acl)
+inline int acl_valid(acl_t)
```
elBoberido, line 36: it's preferred to use [[gnu::unused]] instead of deleting the parameter name, especially if there is a function like foo(int, int, int)

iceoryx_utils/platform/mac/source/semaphore.cpp
```diff
@@ -236,6 +236,11 @@ int iox_sem_unlink(const char* name)
 
 iox_sem_t* iox_sem_open_impl(const char* name, int oflag, ...)
 {
+    if (strlen(name) == 0)
```
elBoberido, line 239: how about `strlen(name) == 0 || name[0] == 0`?

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -61,13 +61,6 @@ void Timer::OsTimer::callbackHelper(sigval data)
     auto index = Timer::OsTimerCallbackHandle::sigvalToIndex(data);
     auto descriptor = Timer::OsTimerCallbackHandle::sigvalToDescriptor(data);
 
-    /// @todo cxx::expect
```
elBoberido, line 64: I know there is a warning, but it's just because `index` is currently a `uint8_t`. It could be changed in the future and then this check is necessary. Please cast `index` to an `uint32_t` instead

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -899,3 +899,138 @@ TEST_F(vector_test, ConstructorWithSizeParameterGreaterThanCapacity)
...
+    vector<int, 20> b;
+    b.emplace_back(4);
+    b.emplace_back(5);
+    b.emplace_back(6);
+
+    EXPECT_TRUE(a == b);
+    EXPECT_FALSE(a != b);
+}
+
+TEST_F(vector_test, TwoNonEqualVectorsWithSameCapacityAreNotEqual)
```
elBoberido, line 951: shouldn't this be TwoVectorsWithDifferentSizeAreNotEqual?

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -28,11 +28,26 @@ namespace popo
...
 
 using SenderPortType = iox::popo::SenderPort;
 using ReceiverPortType = iox::popo::ReceiverPort;
 
 constexpr char MQ_ROUDI_NAME[] = "/roudi";
+/// @brief The socket is created in the current path if no absolute path is given hence
+///      we need an absolut path so that every application knows where our sockets can
+///      be found.
+#if defined(__APPLE__)
+constexpr char MQ_ROOT_PATH[] = "/tmp";
```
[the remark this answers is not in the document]
elfenpiff: @elBoberido No if we use sockets on linux or QNX we have to provide the same path otherwise the sockets are created in the current working directory and all the other apps have no idea where to look for them

iceoryx_utils/test/moduletests/test_relocatable_ptr.cpp
```diff
@@ -75,8 +75,6 @@ TEST_F(RelocatablePointer_test, relocation)
     uint8_t* base1 = block1[0];
     uint8_t* base2 = block2[0];
 
-    EXPECT_EQ(base2 - base1, BLOCK_SIZE);
```
elBoberido, line 78: why is this removed?

iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
```diff
@@ -426,7 +426,7 @@ TEST_F(Mepoo_IntegrationTest, DISABLED_SampleOverflow)
     EXPECT_DEATH({ sendreceivesample<samplesize1>(repetition1); }, ".*");
 }
 
-TEST_F(Mepoo_IntegrationTest, MempoolCreationTimeDefaultConfig)
+TEST_F(Mepoo_IntegrationTest, DISABLED_MempoolCreationTimeDefaultConfig)
```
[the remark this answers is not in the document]
elfenpiff: I think this test is obsolete since we do not have any requirements for the time how long it should take to create a mempool with the default config. Additionally, this time varies a lot! It is different for MacOS, Windows and GitHub Ubuntu and fails a lot there. Either we make a timing test out of it or we remove it completely. I would recommend the latter one as long as no one can state to me what the specific goal of this test is.

tools/run_all_tests.sh
```diff
@@ -18,6 +18,32 @@
 
 component_folder="posh utils"
 
+GTEST_FILTER="*"
+
+for arg in "$@"
```
elBoberido, line 23: nice

README.md
```diff
@@ -75,24 +75,30 @@ the interwebs on a lonely evening, he finds out about iceoryx: Free-to-use, high
...
+the introspection client. To install ncurses locally into your build folder follow these steps
+```
+cd iceoryx
+ICEORYX_DIR=$PWD
+mkdir -p build
+cd build
+git clone https://github.com/mirror/ncurses.git 
+cd ncurses
+git checkout v6.2          
+./configure  --prefix=$ICEORYX_DIR/build/install/prefix/ --exec-prefix=$ICEORYX_DIR/build/install/prefix/ --with-termlib
```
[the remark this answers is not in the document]
elfenpiff: done

README.md
```diff
@@ -103,40 +109,62 @@ You will need to install the following packages:
...
-#### Build from sources
+### Build with CMake (all supported platforms)
 
-iceoryx_utils and iceoryx_posh are deployed as independent cmake packages. Posh is using some functions from utils and is depending on it. You are able to build posh and utils and integrate in into existing cmake projects.
+**NOTE:** Requires CMake version 3.14 or higher. If you only have CMake version 3.5 or higher available please use the
+build script.
 
-##### Build Script
-For the first start, we advise to use our build-test script for building everything.
+We recommend using the ```CMakeLists.txt``` inside of ```iceoryx_meta``` provided by us so that you can
```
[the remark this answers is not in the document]
elfenpiff: done

README.md
```diff
@@ -103,40 +109,62 @@ You will need to install the following packages:
...
+    ./tools/iceoryx_build_test.sh
+    ```
 
-The Googletest-Framework will be automatically fetched from github and the test will be executed and the end of the script.
+With the following arguments you can add additional features:
+ 
+ |  switch  |  description |
+ |:---------|:-------------|
+ | ```clean``` | Removes the build directory and performs a clean build. If you have installed ncurses locally into your build directory you have to reinstall it first. |
+ | ```test``` | Enables module-, integration- and component-tests. The Googletest-Framework will be automatically fetched from github and the test will be executed and the end of the script. |
```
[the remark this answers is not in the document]
elfenpiff: yeah the internal CMakeLists of posh and utils are having even more switches but I would ignore them for now.

iceoryx_posh/test/integrationtests/test_mq_interface_startup_race.cpp
```diff
@@ -33,12 +34,12 @@ using iox::runtime::MqMessage;
 using iox::runtime::MqMessageType;
 using iox::runtime::MqRuntimeInterface;
 
-using MQueue = iox::posix::MessageQueue;
-
-constexpr char MqRouDiName[] = "/roudi";
-constexpr char MqAppName[] = "/racer";
 
+#if !defined(__APPLE__)
```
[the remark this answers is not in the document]
elfenpiff: Yes, otherwise I get an unused variable warning and then I have to add clang warning suppression pragmas to avoid that you call my firstborn Jean Luc Runkelruebe

iceoryx_utils/test/moduletests/test_relocatable_ptr.cpp
elfenpiff: I discussed it with @MatthiasKillat and this line is not needed for the test since it actually does not test any kind of behavior.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
elfenpiff: done

iceoryx_utils/source/posix_wrapper/timer.cpp
elfenpiff: done

iceoryx_utils/platform/mac/source/semaphore.cpp
elfenpiff: done

iceoryx_utils/platform/mac/include/iceoryx_utils/platform/acl.hpp
elfenpiff: You are right in general but in the case of the Mac OS posix platform implementation where we just provide empty functions so that the code compiles I think it adds a lot of code bloat with nothing to gain. If we would have an underlying implementation than I would agree with you but since they are only empty stub functions it is not of any advantage when we know what parameter represents what since we are always `return 0;`

iceoryx_posh/test/mocks/chunk_mock.hpp
elBoberido: btw, you could also remove the platform defines and use the mac solution for all platforms. This reduces the complexity of the mock

iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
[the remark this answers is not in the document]
elBoberido: I think this was a check for QNX to spot regressions, since it takes quite long there

iceoryx_posh/test/mocks/chunk_mock.hpp
elfenpiff: done

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
elfenpiff: done

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/argv_inspection.hpp
elfenpiff: done

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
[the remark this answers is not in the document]
elBoberido: then I would favour to put this into UnixDomainSocket. Just prepend the name with this and you are done. It's also more save to use, since you cannot forget to set it.

elfenpiff:
> If I manually skip `iceoryx_introspection` and `iceoryx_dds_gateway` my workspace compiles locally.
> 
> I would recommend having something along the following lines in the CMakeLists.txt in these unsupported packages:
> 
> ```cmake
> if(NOT UNIX)
>   message(WARNING "${PROJECT_NAME} is currently only supported on Linux")
>   return()
> endif(UNIX)
> ```

@Karsten1987  The introspection and our cyclone adds gateway are supported on Mac OS but you have to install the prerequisites (ncurses, CycloneDDS) first and follow our installation instructions from the readme - I added them some seconds ago.

iceoryx_posh/source/runtime/message_queue_interface.cpp
elfenpiff: done

mossmaurice:
> 
> 
> > If I manually skip `iceoryx_introspection` and `iceoryx_dds_gateway` my workspace compiles locally.
> > I would recommend having something along the following lines in the CMakeLists.txt in these unsupported packages:
> > ```cmake
> > if(NOT UNIX)
> >   message(WARNING "${PROJECT_NAME} is currently only supported on Linux")
> >   return()
> > endif(UNIX)
> > ```
> 
> @Karsten1987 The introspection and our cyclone adds gateway are supported on Mac OS but you have to install the prerequisites (ncurses, CycloneDDS) first and follow our installation instructions from the readme - I added them some seconds ago.

I guess we would need a ROS2 vendor package like `cpptoml_vendor`?

iceoryx_posh/include/iceoryx_posh/internal/roudi_environment/roudi_environment.hpp
```diff
@@ -62,7 +62,7 @@ class RouDiEnvironment
 
   private:
     RuntimeTestInterface m_runtimes;
-    std::chrono::milliseconds m_interOpWaitingTime = std::chrono::milliseconds(200);
+    std::chrono::milliseconds m_interOpWaitingTime = std::chrono::milliseconds(1000);
```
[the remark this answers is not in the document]
elfenpiff: done

iceoryx_posh/include/iceoryx_posh/internal/runtime/runnable_data.hpp
```diff
@@ -23,8 +23,9 @@ namespace iox
 namespace runtime
 {
 /// @brief struct which contains all the members of an object of type Runnable
-struct RunnableData
+class RunnableData
```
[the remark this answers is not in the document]
elfenpiff: There was a mixture of `class RunnableData` and `struct RunnableData` in multiple forward declarations. I think the confusion and mixture would be solved if this becomes a class since most developers are assuming that this is a class when they declare it.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -3,10 +3,14 @@ namespace iox
...
 {
 template <uint64_t Capacity, typename ValueType>
-constexpr IndexQueue<Capacity, ValueType>::IndexQueue(ConstructEmpty_t) noexcept
+IndexQueue<Capacity, ValueType>::IndexQueue(ConstructEmpty_t) noexcept
     : m_readPosition(Index(Capacity))
     , m_writePosition(Index(Capacity))
 {
+    for (uint64_t i = 0u; i < Capacity; ++i)
+    {
+        m_cells[i].store(Index(0));
```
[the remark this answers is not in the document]
elfenpiff: done

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -28,11 +28,26 @@ namespace popo
...
+using CommunicationType = iox::posix::UnixDomainSocket;
```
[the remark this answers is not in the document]
elfenpiff: done

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -160,10 +160,13 @@ bool MqBase::openMessageQueue(const posix::IpcChannelSide channelSide) noexcept
...
-    posix::MessageQueue::create(
-        m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
+    CommunicationType::create(std::string(MQ_ROOT_PATH) + m_interfaceName,
+                              posix::IpcChannelMode::BLOCKING,
+                              m_channelSide,
+                              m_maxMessageSize,
+                              m_maxMessages)
         .on_success(
-            [this](cxx::expected<posix::MessageQueue, posix::IpcChannelError>& mq) { this->m_mq = std::move(*mq); });
+            [this](cxx::expected<CommunicationType, posix::IpcChannelError>& mq) { this->m_mq = std::move(*mq); });
```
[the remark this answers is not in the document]
elfenpiff: yeah yeah I know ... it is done ;)

iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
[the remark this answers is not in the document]
elfenpiff: Alright, to make it clear that this test is only intended for QNX I added a macro around the test

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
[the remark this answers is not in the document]
elfenpiff: It is part of UnixDomainSocket and the sockets can be created with root path (default) or explicitly without a root path.

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -160,10 +160,13 @@ bool MqBase::openMessageQueue(const posix::IpcChannelSide channelSide) noexcept
     m_mq.destroy();
 
     m_channelSide = channelSide;
-    posix::MessageQueue::create(
-        m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
+    CommunicationType::create(std::string(MQ_ROOT_PATH) + m_interfaceName,
```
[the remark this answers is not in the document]
elfenpiff: Its inside the socket wrapper

iceoryx_posh/test/mocks/chunk_mock.hpp
```diff
@@ -32,10 +32,19 @@ class ChunkMock
...
+        auto alignedAddress = offset = 0 ? memAddress : memAddress + Alignment - offset;
+        m_alignedMemory = reinterpret_cast<uint8_t*>(alignedAddress);
 #elif defined(_WIN32)
         m_rawMemory = static_cast<uint8_t*>(_aligned_malloc(Alignment, Size));
+        m_alignedMemory = m_rawMemory;
 #else
         m_rawMemory = static_cast<uint8_t*>(aligned_alloc(Alignment, Size));
+        m_alignedMemory = m_rawMemory;
 #endif
         assert(m_rawMemory != nullptr && "Could not get aligned memory");
```
elBoberido, line 50: memset and placement new for m_chunkHeader must use the m_alignedMemory member. I think it would also be sufficient if m_alignedMemory would be a local variable.

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
```diff
@@ -42,6 +42,12 @@ class MemoryProviderTestImpl : public iox::roudi::MemoryProvider
 
 #if defined(QNX) || defined(QNX__) || defined(__QNX__)
         dummyMemory = static_cast<uint8_t*>(memalign(alignment, size));
+#elif defined(__APPLE__)
+        dummyMemory = static_cast<uint8_t*>(malloc(size + alignment));
+        auto memAddress = reinterpret_cast<uint64_t>(dummyMemory);
+        auto offset = memAddress % alignment;
+        auto alignedAddress = offset = 0 ? memAddress : memAddress + alignment - offset;
```
elBoberido, line 49: same issue like in chunk_mock.hpp

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
```diff
@@ -42,6 +42,12 @@ class MemoryProviderTestImpl : public iox::roudi::MemoryProvider
...
+        m_alignedMemory = reinterpret_cast<uint8_t*>(alignedAddress);
 #else
         dummyMemory = static_cast<uint8_t*>(aligned_alloc(alignment, size));
 #endif
```
elBoberido, line 53: you have to return the aligned memory

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
```diff
@@ -56,14 +62,19 @@ class MemoryProviderTestImpl : public iox::roudi::MemoryProvider
             destroyMemoryMock();
         }
 
-        free(dummyMemory);
-        dummyMemory = nullptr;
+        if (dummyMemory != nullptr)
```
elBoberido, line 65: this check is superfluous and just adds visual noise

iceoryx_utils/test/moduletests/test_message_queue.cpp
```diff
@@ -24,7 +24,11 @@ using namespace ::testing;
 using namespace iox;
 using namespace iox::posix;
 
+#if defined(__APPLE__)
+using IpcChannel = UnixDomainSocket;
+#else
 using IpcChannel = MessageQueue;
```
elBoberido, line 30: why not just IpcChannel = IpcChannelType?

iceoryx_posh/test/moduletests/test_roudi_mempool_introspection.cpp
```diff
@@ -226,12 +226,17 @@ TEST_F(MemPoolIntrospection_test, thread)
     // we use the hasSubscribers call to check how often the thread calls the send method
     mock->hasSubscribersReturn = false;
 
-    m_introspection.setSnapshotInterval(10);
+    using namespace iox::units::duration_literals;
+    iox::units::Duration timeout(100_ms);
```
elBoberido, line 230: please rename this to `snapshotInterval`

.github/workflows/build-test.yml
```diff
@@ -73,8 +73,8 @@ jobs:
...
         run: $GITHUB_WORKSPACE/tools/iceoryx_build_test.sh build-test
 
-      # - name: Run tests
-      #   run: |
-      #     cd $GITHUB_WORKSPACE/build
-      #     ../tools/run_all_tests.sh
+      - name: Run tests
+        run: |
+          cd $GITHUB_WORKSPACE/build
+          ../tools/run_all_tests.sh disable-timing-tests
```
mossmaurice, line 79: Why did you disable the timing tests? Could we please fix them instead of disabling?

README.md
```diff
@@ -103,40 +112,66 @@ You will need to install the following packages:
...
+
+With the following CMake switches you can add additional features:
+ 
+ |  switch  |  description |
+ |:---------|:-------------|
+ | ```dds``` | builds the iceoryx dds gateway which requires an installed CycloneDDS, see [https://github.com/eclipse-cyclonedds/cyclonedds](https://github.com/eclipse-cyclonedds/cyclonedds) |
+ | ```examples``` | builds all examples |
+ | ```introspection``` | the console introspection client which requires an installed ncurses library with terminfo support |
+ | ```test``` | enables module-, integration- and component-tests |
+ | ```TOML_CONFIG``` | activates config file support by using toml, if this is deactivated the central broker ```RouDi``` is not being build |
```
mossmaurice, line 149: This list will become outdated very soon as development continues. Can we not add the CMake option that lists all the available targets?

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -31,13 +31,45 @@ UnixDomainSocket::UnixDomainSocket() noexcept
...
+    : UnixDomainSocket(
+        NoPathPrefix,
+        [&]() -> std::string {
+            /// invalid names will be forwarded and handled by the other constructor
+            /// separately
+            if (!isNameValid(name))
+            {
+                return name;
+            }
+            return std::string(PATH_PREFIX) + name;
```
mossmaurice, line 47: Please don't return a `std::string` but a `cxx::string` instead

iceoryx_posh/test/mocks/chunk_mock.hpp
```diff
@@ -32,10 +32,19 @@ class ChunkMock
     {
 #if defined(QNX) || defined(QNX__) || defined(__QNX__)
         m_rawMemory = static_cast<uint8_t*>(memalign(Alignment, Size));
+        m_alignedMemory = m_rawMemory;
+#elif defined(__APPLE__)
+        m_rawMemory = static_cast<uint8_t*>(malloc(Size + Alignment));
+        auto memAddress = reinterpret_cast<uint64_t>(m_rawMemory);
+        auto offset = memAddress % Alignment;
+        auto alignedAddress = offset = 0 ? memAddress : memAddress + Alignment - offset;
```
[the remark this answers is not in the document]
elfenpiff: Whoopsie

iceoryx_posh/test/mocks/chunk_mock.hpp
elfenpiff: fixed by removing all platform macros and introduced `alignedAlloc` and a corresponding `alignedFree` in cxx helpless

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
elfenpiff: fixed

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
elfenpiff: fixed

iceoryx_posh/test/mocks/roudi_memory_provider_mock.hpp
elfenpiff: fixed

iceoryx_utils/test/moduletests/test_message_queue.cpp
elfenpiff: done

iceoryx_posh/test/moduletests/test_roudi_mempool_introspection.cpp
elfenpiff: done

.github/workflows/build-test.yml
elfenpiff: Yes we have to fix them but not in this pull request. The timing tests are constantly failing on Mac OS and we need a better concept to handle them. @mossmaurice  If you have an idea how to fix them please let me know!
Otherwise I would suggest a special Jenkins target which handles only timing tests and therefore I would keep them disabled here.

README.md
elfenpiff: This is the `iceoryx_meta` CMakeLists.txt documentation. Yes we maybe add new options there but then it is not a problem to keep this documentation up to date otherwise the user has to crawl through the Cmake file to see what configuration options are possible - and this is just horrible.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elfenpiff: This has to be done in a separate task when we get rid of `std::string` in our code base. It is not in the scope of this pull request to remove `std::string` from `UnixDomainSocket`.

mossmaurice: Got your point. Could you please create an new issue for this topic? Cheers!

iceoryx_utils/test/moduletests/test_message_queue.cpp
```diff
@@ -24,7 +24,11 @@ using namespace ::testing;
 using namespace iox;
 using namespace iox::posix;
 
-using IpcChannel = MessageQueue;
+#if defined(__APPLE__)
+using IpcChannelType = UnixDomainSocket;
```
elBoberido, line 28: why not just include the `iceoryx_posh_types.hpp`?
I think you misunderstood my previous request. The idea was to include `iceoryx_posh_types.hpp` and just add a `using IpcChannel = iox::IpcChannelType;` to get rid of the platform dependent code.

iceoryx_utils/test/moduletests/test_message_queue.cpp
```diff
@@ -239,7 +243,11 @@ TEST_F(MessageQueue_test, sendAfterServerDestroy)
     std::string message = "Try to send me";
     auto sendResult = client.send(message);
     EXPECT_TRUE(sendResult.has_error());
+#if defined(__APPLE__)
+    EXPECT_THAT(sendResult.get_error(), Eq(IpcChannelError::CONNECTION_RESET_BY_PEER));
```
elBoberido, line 247: it's unfortunate to have different error codes depending on the channel type. This make error handling more error prone if you exchange the IpcChannelType. Is there a possibility to find a common error code for this behaviour?

iceoryx_utils/test/moduletests/test_message_queue.cpp
```diff
@@ -307,7 +316,6 @@ TEST_F(MessageQueue_test, timedSend)
         auto after = system_clock::now();
         if (result.has_error())
         {
-            ASSERT_THAT(result.get_error(), Eq(IpcChannelError::TIMEOUT));
```
elBoberido, line 310: why is this removed?

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.hpp
```diff
@@ -177,6 +177,10 @@ class IndexQueue
     using Index = CyclicIndex<Capacity>;
 
     using Cell = std::atomic<Index>;
+    /// note: this member has to be initialized explicitly in the constructor since
```
dkroenke, line 180: please use @note here for doxygen

iceoryx_utils/test/moduletests/test_message_queue.cpp
elfenpiff: Yes I did this intentionally. If I would use `iceoryx_posh_types.hpp` from `iceoryx_posh` the `iceoryx_utils` would have a dependency to posh which I do not want.

elBoberido: oh, my bad. I feel kind of stupid now

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.hpp
elfenpiff: I removed the note since this is not part of the developer documentation. It is a hint for fellow developers to explicitly initialise this array with a loop when they implement a new constructor.

iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
```diff
@@ -426,6 +426,7 @@ TEST_F(Mepoo_IntegrationTest, DISABLED_SampleOverflow)
     EXPECT_DEATH({ sendreceivesample<samplesize1>(repetition1); }, ".*");
 }
 
+#if defined(QNX) || defined(QNX__) || defined(__QNX__)
```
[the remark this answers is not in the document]
elfenpiff: I converted it into a timing test which are disabled for macOS at the moment.

iceoryx_meta/CMakeLists.txt
```diff
@@ -7,9 +7,17 @@ project(iceoryx VERSION ${iceoryx_VERSION})
...
 
+option(examples "build with iceoryx examples" ON)
+option(introspection "builds the introspection client which requires the ncurses library with an activated terminfo feature" OFF)
+option(diagnostic "enables allmost all compiler warnings" OFF)
+option(dds "enables dds support by compiling the iceoryx dds gateway" OFF)
 
 include(FetchContent)
 
+if (diagnostic)
+    add_compile_options(-Wall -W -Wextra -pedantic)
```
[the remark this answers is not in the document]
elfenpiff: You are right and I created and issue for that: https://github.com/eclipse/iceoryx/issues/153

iceoryx_meta/CMakeLists.txt
```diff
@@ -49,5 +57,15 @@ endif(TOML_CONFIG)
...
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../iceoryx_examples/iceperf ${CMAKE_BINARY_DIR}/examples/iceperf )
+endif(examples)
+
+if(introspection)
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../tools/introspection ${CMAKE_BINARY_DIR}/tools/introspection )
+endif(introspection)
+
+if(dds)
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../iceoryx_dds_gateway ${CMAKE_BINARY_DIR}/iceoryx_dds_gateway )
+endif(dds)
```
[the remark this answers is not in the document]
elfenpiff: Not needed.

iceoryx_utils/test/moduletests/test_message_queue.cpp
elfenpiff: reintroduced...

iceoryx_utils/test/moduletests/test_message_queue.cpp
elfenpiff: The `IpcChannelError` value depends on the implementation but not that we expect some kind of error. I would suggest to remove our expectation for a specific kind of error and just expect one.

iceoryx_utils/testutils/timing_test.cpp
```diff
@@ -0,0 +1,44 @@
+#include "timing_test.hpp"
```
elBoberido, line 1: add license header

elfenpiff: done

Issue #131, opened 2020-06-09
Replace std::function occurrences


mossmaurice:
## Brief feature description

´std::function´ is hard to validate for [ASIL](https://en.wikipedia.org/wiki/Automotive_Safety_Integrity_Level), hence replace it where possible.

## Detailed information

* Step 1): Replace all occurrences when used as function parameter
* Step 2): All other places e.g. ´posix::Timer´

mossmaurice:
Step 2) consists of
* `ErrorHandler`
* `posix::Timer`
* `Iceoryx2DDSGateway`
* `Subscriber_t::setReceiveHandler`
* `PoshRuntime::s_runtimeFactory`
* `GenericRAII`
* `ActiveObject`

mossmaurice:
Just discussed this topic with @elBoberido 
> 
> Step 2) consists of
> 
>     * `ErrorHandler`
`function_ref` can be used
> 
>     * `posix::Timer`
will be removed
> 
>     * `Iceoryx2DDSGateway`
not relevant, quality level 4
> 
>     * `Subscriber_t::setReceiveHandler`
will be removed with iceoryx 1.0 release
> 
>     * `PoshRuntime::s_runtimeFactory`
`function_ref` can be used
> 
>     * `GenericRAII`
`function_ref` can be used for c'tor but not for d'tor **open point**
> 
>     * `ActiveObject`
**open point,** reference counting on callable user might solve the lifetime issue (save_ref)

PR #139, opened 2020-06-15
Iox #131 Remove std::function in expected and optional

mossmaurice:
Thanks @prasannabhat for the help with the service discovery


iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp
```diff
@@ -25,6 +25,7 @@
 #include <condition_variable>
 #include <cstdint>
 #include <ctime>
+#include <functional>
```
elfenpiff, line 28: Why did you add the functional header?

iceoryx_utils/test/moduletests/test_cxx_function_ref.cpp
```diff
@@ -60,6 +61,17 @@ ComplexType returnComplexType(ComplexType foo)
     return foo;
 }
 
+int SameSignature(function_ref<void(void)> callback)
+{
+    callback();
+    return -1;
+}
+
+int SameSignature(function_ref<int(int)> callback)
```
elfenpiff, line 70: I know it is a bit pedantic but could you please also add unit tests for

- ```int SameSignature(function_ref<int(void)> callback)```
- ```int SameSignature(function_ref<void(int)> callback)```
- ```int SameSignature(function_ref<int(int, int)> callback)```

iceoryx_utils/test/moduletests/test_cxx_function_ref.cpp
```diff
@@ -274,3 +286,10 @@ TEST_F(function_refTest, StoreInStdFunctionResultEqual)
     std::function<int()> sut(moep);
     EXPECT_THAT(sut(), Eq(37));
 }
+
+
+TEST_F(function_refTest, CallOverloadedFunctionResultsInCallOfCorrect)
```
elfenpiff, line 291: Could you please also add a test for the void(void) function and check if the value is equal to ```-1```

iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp
mossmaurice: Used to be implicitly included via `optional` before.

iceoryx_utils/test/moduletests/test_cxx_function_ref.cpp
mossmaurice: Added. However, `int SameSignature(function_ref<void(int)> callback)` would be ambiguous to `int SameSignature(function_ref<int(int)> callback)` as the return value is not part of the signature, so this one is left out.

iceoryx_utils/test/moduletests/test_cxx_function_ref.cpp
mossmaurice: Replaced `void(void)` with `int(void)` to have similar test cases.

iceoryx_utils/test/moduletests/test_cxx_function_ref.cpp
elfenpiff: You are right.

PR #154, opened 2020-06-25
Iox #25 subscriber port

budrus:
This is the first PR for the subscriber ports (and fixing of QAC warnings in the other ports and building blocks)
The subscriber port module is not complete yet. It needs to be extended with the condition variable. There are also variants for single and multi producer. For the multi producer version, first the MPMC queue has to be integrated into the variant queue. 

So there are some todos in the code. As the subscriber port is not yet used this should be fine . Also the tests of the subscriber port will be extended if we finalize things with the condition variable


iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -87,10 +90,11 @@ ChunkDistributor<ChunkDistributorDataType>::removeQueue(cxx::not_null<ChunkQueue
 {
     typename MemberType_t::LockGuard_t lock(*getMembers());
 
-    auto iter = std::find(getMembers()->m_queues.begin(), getMembers()->m_queues.end(), queueToRemove);
+    const auto iter = std::find(getMembers()->m_queues.begin(), getMembers()->m_queues.end(), queueToRemove);
     if (iter != getMembers()->m_queues.end())
     {
-        getMembers()->m_queues.erase(iter);
+        /// @remark we don't use iter any longer so return value can be ignored
```
mossmaurice, line 96: Please use @ rationale as described in our CONTRIBUTING.md

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
```diff
@@ -81,7 +80,7 @@ class BasePort
     MemberType_t* getMembers() noexcept;
 
   private:
-    MemberType_t* m_basePortDataPtr{nullptr};
+    MemberType_t* m_basePortDataPtr;
```
mossmaurice, line 83: Why did you remove the value initialisation?

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port_data.hpp
```diff
@@ -44,19 +42,19 @@ struct BasePortData
...
     BasePortData& operator=(BasePortData&&) = delete;
+    ~BasePortData() = default;
 
     capro::ServiceDescription m_serviceDescription;
-    cxx::CString100 m_processName;
+    ProcessName_t m_processName;
 
     static std::atomic<uint64_t> s_uniqueIdCounter;
-    const std::atomic<uint64_t> m_uniqueId{0};
+    const std::atomic<uint64_t> m_uniqueId;
```
mossmaurice, line 57: Why this change?

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/interface_port.hpp
```diff
@@ -26,19 +26,21 @@ namespace popo
...
+    explicit InterfacePort(InterfacePortData* const interfacePortDataPtr) noexcept;
+
     InterfacePort(const InterfacePort& other) = delete;
     InterfacePort& operator=(const InterfacePort& other) = delete;
     InterfacePort(InterfacePort&& other) = default;
     InterfacePort& operator=(InterfacePort&& other) = default;
+    ~InterfacePort() = default;
 
-    bool dispatchCaProMessage(const capro::CaproMessage& f_message);
+    bool dispatchCaProMessage(const capro::CaproMessage& message) noexcept;
```
mossmaurice, line 37: ❤

iceoryx_posh/source/popo/ports/interface_port.cpp
```diff
@@ -19,38 +19,38 @@ namespace iox
...
     }
     return returner;
 }
 
-bool InterfacePort::getCaProMessage(capro::CaproMessage& f_message)
+bool InterfacePort::getCaProMessage(capro::CaproMessage& message) noexcept
 {
-    auto msg = getMembers()->m_caproMessageFiFo.pop();
-    if (msg.has_value())
+    auto maybeMessage = getMembers()->m_caproMessageFiFo.pop();
```
elBoberido, line 39: how about
```
    return getMembers()
        ->m_caproMessageFiFo.pop()
        .and_then([&message](capro::CaproMessage& receivedMessage) { message = receivedMessage; })
        .has_value();
```

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/subscriber_port_user.hpp
```diff
@@ -0,0 +1,101 @@
...
+    bool hasNewChunks() noexcept;
+
+    /// @brief check if there was a queue overflow since the last call of hasLostChunks
+    /// @return true if the underlying queue overflowed since last call of this method, otherwise false
+    bool hasLostChunks() noexcept;
+
+    /// @todo we first need the new condition variable
+    void attachConditionVariable() noexcept;
+    void detachConditionVaribale() noexcept;
+    bool isConditionVariableAttached() noexcept;
```
mossmaurice, line 86: Please change return value to `bool` as it will later derive from `Condition` which provides these three methods, see https://github.com/mossmaurice/iceoryx/blob/iox-%2325-semaphore-refactoring/iceoryx_posh/include/iceoryx_posh/internal/popo/waitset/condition.hpp#L40

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
```diff
@@ -79,22 +79,28 @@ void ChunkQueuePopper::setCapacity(const uint32_t newCapacity) noexcept
...
     {
         auto retVal = getMembers()->m_queue.pop();
-        if (retVal)
+        if (retVal.has_value())
         {
-            auto chunkTupleOut = *retVal;
-            auto chunkManagement =
-                iox::relative_ptr<mepoo::ChunkManagement>(chunkTupleOut.m_chunkOffset, chunkTupleOut.m_segmentId);
-            auto chunk = mepoo::SharedChunk(chunkManagement);
+            /// @remark d'tor of SharedChunk will release the memory, so RAII has the side effect here
```
mossmaurice, line 99: Please use @ rationale

iceoryx_posh/source/popo/ports/interface_port.cpp
```diff
@@ -19,38 +19,38 @@ namespace iox
...
+InterfacePort::InterfacePort(InterfacePortData* const interfacePortDataPtr) noexcept
+    : BasePort(interfacePortDataPtr)
 {
 }
 
-bool InterfacePort::dispatchCaProMessage(const capro::CaproMessage& f_message)
+bool InterfacePort::dispatchCaProMessage(const capro::CaproMessage& message) noexcept
 {
-    bool returner = getMembers()->m_caproMessageFiFo.push(f_message);
+    const bool returner = getMembers()->m_caproMessageFiFo.push(message);
```
elBoberido, line 29: I'm quite sure this isn't a returner ;)
https://www.dict.cc/?s=returner

How about retVal or pushResult?

iceoryx_posh/source/popo/ports/subscriber_port_multi_producer.cpp
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
mossmaurice, line 1: Year?

iceoryx_posh/source/popo/ports/subscriber_port_data.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
mossmaurice, line 1: Year?

iceoryx_posh/source/popo/ports/subscriber_port_multi_producer.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+SubscriberPortMultiProducer::SubscriberPortMultiProducer(
+    cxx::not_null<MemberType_t* const> subscriberPortDataPtr) noexcept
+    : SubscriberPortRouDi(subscriberPortDataPtr)
+{
+}
+
+cxx::optional<capro::CaproMessage> SubscriberPortMultiProducer::getCaProMessage() noexcept
+{
+    /// @todo
```
mossmaurice, line 29: Please document this with a comment in #25 as follow-up work is needed here.

iceoryx_posh/source/popo/ports/subscriber_port_roudi.cpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
mossmaurice, line 1: Year?

iceoryx_posh/source/popo/ports/application_port.cpp
```diff
@@ -18,33 +18,33 @@ namespace iox
...
-    return getMembers()->m_caproMessageFiFo.push(f_message);
+    return getMembers()->m_caproMessageFiFo.push(message);
 }
 
-bool ApplicationPort::getCaProMessage(capro::CaproMessage& f_message)
+bool ApplicationPort::getCaProMessage(capro::CaproMessage& message) noexcept
 {
-    auto msg = getMembers()->m_caproMessageFiFo.pop();
-    if (msg.has_value())
+    auto maybeMessage = getMembers()->m_caproMessageFiFo.pop();
```
elBoberido, line 33: how about
```
    return getMembers()
        ->m_caproMessageFiFo.pop()
        .and_then([&message](capro::CaproMessage& receivedMessage) { message = receivedMessage; })
        .has_value();
```

iceoryx_posh/source/popo/ports/subscriber_port_single_producer.cpp
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
mossmaurice, line 1: Year?

iceoryx_posh/source/popo/ports/subscriber_port_user.cpp
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
mossmaurice, line 1: Year?

iceoryx_posh/source/popo/ports/subscriber_port_user.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void SubscriberPortUser::detachConditionVaribale() noexcept
+{
+}
+
+bool SubscriberPortUser::isConditionVariableAttached() noexcept
+{
+    return false;
+}
```
mossmaurice, line 113: Will be implemented on our marriage branch of `WaitSet` and `Subscriber` as discussed.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_types.hpp
```diff
@@ -31,14 +31,14 @@ enum class ChunkQueueError
...
-    ChunkTuple(iox::relative_ptr<mepoo::ChunkManagement> f_chunk) noexcept
+    explicit ChunkTuple(iox::relative_ptr<mepoo::ChunkManagement> f_chunk) noexcept
         : m_segmentId(f_chunk.getId())
         , m_chunkOffset(f_chunk.getOffset())
     {
     }
 
-    RelativePointer::id_t m_segmentId{iox::RelativePointer::NULL_POINTER_ID};
-    RelativePointer::offset_t m_chunkOffset{iox::RelativePointer::NULL_POINTER_OFFSET};
+    RelativePointer::id_t m_segmentId;
```
elBoberido, line 40: why is this changed? are we sure this is correctly initialized with the default c'tor?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_receiver.hpp
```diff
@@ -26,7 +26,7 @@ namespace iox
 {
 namespace popo
 {
-enum class ChunkReceiverError
+enum class ChunkReceiveError
```
elBoberido, line 29: why this change?

iceoryx_posh/source/popo/ports/interface_port.cpp
mossmaurice: +1 for pushResult

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
budrus: I tried @ rationale, My IDE did not recognize this as keyword and I also found no clear evidence that @ rationale is proper doxygen when searching the internet. So I came up with @ remark as a new proposal ;-)

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
budrus: As this is provided in the c'tor, it is set in the member initializer list. So it is somehow redundant and as far as I remember I realized it as QAC came up with a warning. Here it also gives you a feeling that you can end up with an object that has nullptr for m_basePortDataPtr. Which is not the case because it comes with cxx::not_null in the c'tor

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port_data.hpp
budrus: same same

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/subscriber_port_user.hpp
budrus: I think we can change this when merging with your branch. In my branch there is also no ConditionVariableData yet, so I cannot finalize the signature anyway

iceoryx_posh/source/popo/ports/subscriber_port_multi_producer.cpp
budrus: Done, I added separate check boxes in the task list of #25

iceoryx_posh/source/popo/ports/subscriber_port_roudi.cpp
budrus: oops

iceoryx_posh/source/popo/ports/interface_port.cpp
budrus: You're right. Seems that I have to go the extra mile for adding the const ;-)

iceoryx_posh/source/popo/ports/interface_port.cpp
budrus: @elBoberido: Is this now free of std::function and has better or at least same performance?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_types.hpp
budrus: I think I removed it as QAC complained. Which would be OK if there is only the c'tor with relative ptr to ChunkManagement. But yes, there is also a default c'tor. Will check it again

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_receiver.hpp
budrus: I use the same error codes again for the subscriber. So it is no more the error for the chunk receiver but more general an error that can occur when a chunk is received, either with ChunkReceiver or with SubscriberPort

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
elBoberido: didn't we decide on storing references?

iceoryx_posh/source/popo/ports/interface_port.cpp
elBoberido: yes, it's free of std::function and the performance should also similar. I wouldn't bet the naming rights of my first born on it, though.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
budrus: Then let's do it with the justification possibility the tool provides. Like this
`// PRQA S 3804 1 # we checked the capacity, so pushing will be fine`

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
budrus: The reference refactoring will be done in another PR. This affects also some old modules and I didn't want to blow up this PR

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port_data.hpp
budrus: I will give this a try `const std::atomic<uint64_t> m_uniqueId{s_uniqueIdCounter.fetch_add(1u, std::memory_order_relaxed)};`

iceoryx_posh/source/popo/ports/interface_port.cpp
budrus: You gotta prove it ;-)

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
budrus: new strategy, see adapted CONTRIBUTING.md

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
elfenpiff: > didn't we decide on storing references?

@elBoberido this is not yet possible. If you have references as members in a class the class is neither move- nor copy-able and at the moment we store all port data in a vector. We have to refine this first and need to use the upcoming `cxx::list` before we can go in this direction.

iceoryx_posh/source/popo/ports/interface_port.cpp
elfenpiff: Change the signature to `cxx::optional<capro::CaproMessage> getCaProMessage()` and just write `return getMembers()->m_caproMessageFiFo.pop()`

iceoryx_posh/source/popo/ports/subscriber_port_data.cpp
elfenpiff: 2525?

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+TEST_F(SubscriberPort_test, initialStateReturnsNoCaProMessage)
+{
+    auto maybeCaproMessage = m_sutRouDiSideSingleProducer.getCaProMessage();
+
+    EXPECT_FALSE(maybeCaproMessage.has_value());
+}
+
+TEST_F(SubscriberPort_test, subscribeCallResultsInSubCaProMessage)
```
elfenpiff, line 86: What happens when we call `subscribe` multiple times? If we call it 5 times do we then receive 5 capro messages? Maybe we should test the current behavior...

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -0,0 +1,275 @@
...
+    m_sutRouDiSideSingleProducer.getCaProMessage(); // only RouDi changes state
+    caproMessage.m_type = iox::capro::CaproMessageType::NACK;
+    m_sutRouDiSideSingleProducer.dispatchCaProMessage(caproMessage);
+
+    const auto subscriptionState = m_sutUserSideSingleProducer.getSubscriptionState();
+
+    EXPECT_THAT(subscriptionState, Eq(iox::SubscribeState::NOT_SUBSCRIBED));
+}
+
+TEST_F(SubscriberPort_test, invalidMessageResultsInError)
```
elfenpiff, line 263: I think here are some cases missing. 
- What happens when RouDi breaks the protocol and replies on a subscribe with a PING or FIND?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
budrus: I'm unsure if this makes the code nicer, but it's a valid idea. `discardReturnValue` would be more explicit than the cast

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
budrus: @elfenpiff Sure, we have to collect the alternatives and agree on the best solution. I already had some discussions with @elBoberido

iceoryx_posh/source/popo/ports/interface_port.cpp
budrus: @elfenpiff valid point. We already switched to optional return in the other ports. But I would do that when we are harmonizing the queue interfaces. This will be one of the next tasks in #25 when we are integrating the new MPMC queue in the variant queue

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -136,23 +140,25 @@ template <typename ChunkDistributorDataType>
 inline void ChunkDistributor<ChunkDistributorDataType>::deliverToQueue(cxx::not_null<ChunkQueueData_t* const> queue,
                                                                        mepoo::SharedChunk chunk) noexcept
 {
-    // We intentionally do not return anything here as from a ChunkDistributor point of view it doesn't matter if the
-    // push succeeds or fails
-    ChunkQueuePusher_t(queue).push(chunk);
+    /// @remark We intentionally do not return anything here as from a ChunkDistributor
+    /// point of view it doesn't matter if the push succeeds or fails
+    ChunkQueuePusher_t(queue).push(chunk); // PRQA S 3803
```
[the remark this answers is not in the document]
budrus: Maybe not a topic for this PR but we should discuss this options

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor_data.hpp
```diff
@@ -31,16 +32,23 @@ namespace popo
...
-        m_mutex.unlock();
+        if (!m_mutex.unlock())
+        {
+            errorHandler(Error::kPOPO__CHUNK_DISTRIBUTOR_LOCKING_ERROR, nullptr, ErrorLevel::FATAL);
+        }
     }
-    bool tryLock() const 
+    bool tryLock() const
     {
         return m_mutex.try_lock();
```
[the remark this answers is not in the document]
budrus: My assumption is that it returns false when the mutex is already locked. Which is a valid behavior. Do we have to distinguish an error like it can happen with lock/unlock and the "busy" error which is OK for a try lock?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.hpp
```diff
@@ -65,9 +65,13 @@ class ChunkQueuePopper
     /// @concurrent not thread safe
     void setCapacity(const uint32_t newCapacity) noexcept;
```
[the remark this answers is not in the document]
budrus: Good catch. I also realized it some days ago. And then I also realized that this is already broken in SOFI. I'll put it on the list for queue API harmonization

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/application_port.hpp
```diff
@@ -26,19 +26,20 @@ class ApplicationPort : public BasePort
...
     ApplicationPort& operator=(const ApplicationPort& other) = delete;
-
     ApplicationPort(ApplicationPort&& other) = default;
     ApplicationPort& operator=(ApplicationPort&& other) = default;
+    ~ApplicationPort() = default;
 
-    bool dispatchCaProMessage(const capro::CaproMessage& f_message);
-    bool getCaProMessage(capro::CaproMessage& f_message);
+    bool dispatchCaProMessage(const capro::CaproMessage& message) noexcept;
+    bool getCaProMessage(capro::CaproMessage& message) noexcept;
```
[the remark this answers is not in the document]
budrus: see above

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
elBoberido: another possibility would be to hide the cast behind a macro, e.g. DISCARD_RESULT

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port.hpp
elBoberido: only move assignment is not possible, move c'tors still work

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
budrus: No, this is no state change. We store the subscribe request. If you call multiple times subscribe this has no influence on the request

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
mossmaurice: I don't remember the exact talk but remember "There's only one thing the C++ Community can agree on: Macros are evil" ;-)

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
budrus: will check

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
elBoberido: yes, macros are evil, but in cases like this they have their value and I'd rather take this macro than increasing the build time due to monomorphization of templates

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
budrus: All transitions of the state machine are tested and the case that a (somehow arbitrary) CaPro message is provided by RouDi which breaks the protocol. What I identified is the case when ACK or NACK are received but we do not wait for them. With these additional two test cases we should have coverage of all statements and branches

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor_data.hpp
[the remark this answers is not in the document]
elfenpiff: According to the documentation https://linux.die.net/man/3/pthread_mutex_trylock  the underlying call `pthread_mutex_trylock` can fail when the thread already owns the mutex which would lead to undefined behavior or when the handle to the mutex is invalid. This should in theory never happen since we have our wrapper which ensures the lifetime but when it does the returning bool would not be enough... And in the other calls like `lock` and `unlock` we return false when an failure like invalid handle happens.

PR #164, opened 2020-07-01
iox-#153 always enable warnings and set -Werror for CI

### LINKED ISSUE #153: Enable Compiler Warnings and set -Werror
## Brief feature description

Add target_compile_options for all ice oryx targets which include all required warning flags. These warning flags should be declared as a separate variable inside of ice oryx utils so that they can be managed centrally.
After that the diagnostic option and the compile flags in `iceoryx_meta/CMakeLists.txt`can be removed.

elBoberido:
closes #153

Signed-off-by: Kraus Mathias (CC-AD/ESW1) <mathias.kraus2@de.bosch.com>


iceoryx_utils/cmake/IceoryxPlatformDetection.cmake
```diff
@@ -14,4 +14,16 @@ elseif(WIN32)
     set(ICEORYX_CXX_STANDARD 11)
 elseif(APPLE)
     set(ICEORYX_CXX_STANDARD 17)
-endif(LINUX)
\ No newline at end of file
+endif(LINUX)
+
+set (ICEORYX_WARNINGS -W -Wall -Wextra -Wuninitialized -Wpedantic -Wstrict-aliasing -Wcast-align -Wno-noexcept-type)
```
elBoberido, line 19: I put this here since the `ICEORYX_CXX_STANDARD` is also here, but this file should probably be rename from `IceoryxPlatformDetection.cmake` to `IceoryxFoo.cmake`. Any suggestions for `Foo`?

PR #169, opened 2020-07-02
Iox #167 extend posix timer

### LINKED ISSUE #167: Posix Timer Issues when CallbackRuntime > PeriodicTriggerTime
## Required information

**Operating system:**
Conceptual Problem

**Compiler version:**
Conceptual Problem

**Solution:**
We need to differentiate the following timer:

- Soft Timer: If the callback is still running when the callback should be retriggered: _do nothing_
- ASAPTimer: If the callback is still running when the callback should be retriggered: _rerun callback as soon as the callback is finished_
- Hard Timer: If the callback is still running when the callback should be retriggered: _terminate_

elfenpiff:
Closes #167


iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp
```diff
@@ -76,6 +76,18 @@ class Timer
...
     };
 
+    /// @brief
+    ///   when the callback is still running when the callback should be retriggered:
+    ///     - SOFT_TIMER = do nothing
+    ///     - ASAP_TIMER = retrigger callback right after the callback is finished
+    ///     - HARD_TIMER = terminate
+    enum class TimerType
+    {
+        SOFT_TIMER,
```
elBoberido, line 86: it would be nicer if the name would reflect the behaviour, e.g. if the `TimerType` would be a `CatchUpPolicy` we could have `SKIP_TO_NEXT_BEAT`, `IMMEDIATE` and `TERMINATE`

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -69,32 +69,46 @@ void Timer::OsTimer::callbackHelper(sigval data)
...
     {
         return;
     }
 
-    std::lock_guard<std::mutex> lock(callbackHandle.m_accessMutex);
-    if (!callbackHandle.m_inUse.load(std::memory_order::memory_order_relaxed))
-    {
-        return;
-    }
+    uint64_t currentCycle = (callbackHandle.m_timerType == TimerType::SOFT_TIMER)
```
elBoberido, line 78: there is still a race and under some conditions a callback execution might be skipped
- timer triggers and executes the callback
    - the callbacks takes always almost as much time as the interval time
- the callback is executed, but the mutex is still locked
- the timer triggers again and tries to get the lock
    - since it cannot get the lock, the callback is not executed
- the previous timer invokation continues and unlocks the mutex
- a callback execution is skipped

-> there needs to be a loop around the mutex until currentCycle equals callbackExecutionCycle. This loop must only be executed if tryLock doesn't fail

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -69,32 +69,73 @@ void Timer::OsTimer::callbackHelper(sigval data)
...
+    // without the loop it is possible that a previous thread is interrupted right
+    // after the executeCallback call at the end and the succeeding thread skips the
+    // callback call since it is unable to acquire the mutex with try_lock.
+    // therefore we would skip a callback. to avoid this we have to make sure
+    // the callback is called by checking the callbackExecutionCycle in a while loop
+    if (callbackHandle.m_catchUpPolicy == CatchUpPolicy::IMMEDIATE)
     {
-        return;
-    }
+        while (currentCycle >= callbackHandle.m_callbackExecutionCycle.load(std::memory_order_relaxed))
```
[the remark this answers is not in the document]
elfenpiff: I already feared that something like this can happen and created an issue for that: https://github.com/eclipse/iceoryx/issues/170
For the moment I thought it would be robust enough since this issue should only arise under heavy load with a very small duration set but it seems that under Mac OS the problem already there...

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -69,32 +69,73 @@ void Timer::OsTimer::callbackHelper(sigval data)
...
-    if (!callbackHandle.m_isTimerActive.load(std::memory_order::memory_order_relaxed))
+void Timer::OsTimer::verifyAndExecuteCallback(OsTimerCallbackHandle& handle,
+                                              const uint32_t descriptor,
+                                              const uint64_t currentCycle) noexcept
+{
+    if (handle.m_accessMutex.try_lock() == true)
     {
-        return;
-    }
+        iox::cxx::GenericRAII lockGuard([] {}, [&handle] { handle.m_accessMutex.unlock(); });
```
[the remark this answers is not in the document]
elfenpiff: Please elaborate? When this lock guard goes out of scope it unlocks the `handle.m_accessMutex`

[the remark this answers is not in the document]
elBoberido: I just overlocked that there was no lock guard with the tryLock

iceoryx_utils/source/posix_wrapper/timer.cpp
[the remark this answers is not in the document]
elfenpiff: It is again refactored - hopefully now without any more mistakes

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
     ASSERT_THAT(call.has_error(), Eq(true));
     EXPECT_THAT(call.get_error(), Eq(TimerError::TIMER_NOT_INITIALIZED));
 }
+
+TIMING_TEST_F(Timer_test, SoftTimerContinuesWhenCallbackIsLongerThenTriggerTime, Repeat(5), [&] {
```
elBoberido, line 411: please adjust name, this is now `CatchUpPolicySkipToNextBeat` instead of `SoftTimer`

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+    Timer sut(TIMEOUT,
+              [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10)); });
+
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10));
+    TIMING_TEST_EXPECT_FALSE(hasTerminated);
+});
+
+TIMING_TEST_F(Timer_test, ASAPTimerContinuesWhenCallbackIsLongerThenTriggerTime, Repeat(5), [&] {
```
elBoberido, line 425: please adjust name, there is no `ASAPTimer` anymore

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+    Timer sut(TIMEOUT,
+              [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10)); });
+
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::IMMEDIATE);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10));
+    TIMING_TEST_EXPECT_FALSE(hasTerminated);
+});
+
+TIMING_TEST_F(Timer_test, HardTimerTerminatesWhenCallbackIsLongerThenTriggerTime, Repeat(5), [&] {
```
elBoberido, line 439: please adjust name, there is no `HarTimer` anymore

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 100));
+    // every second callback is skipped since the runtime is slightly longer therefore
+    // the counter must be in that range
+    TIMING_TEST_EXPECT_TRUE(50 <= counter && counter <= 70);
+});
+
+TIMING_TEST_F(Timer_test, ASAPTimerCallsCallbackImmediatelyAfterFinishing, Repeat(5), [&] {
```
elBoberido, line 485: please adjust name, there is no `ASAPTimer` anymore

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::IMMEDIATE);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 100));
+
+    // the asap timer should in theory call the callback 90 times since it is calling it right
+    // after the last one finished and one callback takes 1.1 ms and we run for 100ms.
+    TIMING_TEST_EXPECT_TRUE(80 <= counter && counter <= 100);
+});
+
+TIMING_TEST_F(Timer_test, SoftTimerCallsLessCallbacksThanASAPTimer, Repeat(5), [&] {
```
elBoberido, line 502: please adjust name, there is no `SoftTimer` anymore

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT);
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10));
+    sut.restart(TIMEOUT, Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::TERMINATE);
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10));
+
+    TIMING_TEST_EXPECT_TRUE(hasTerminated);
+});
+
+TIMING_TEST_F(Timer_test, SoftTimerSkipsCallbackWhenStillRunning, Repeat(5), [&] {
```
elBoberido, line 469: please adjust name, there is no `SoftTimer` anymore

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+    Timer sut(TIMEOUT,
+              [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10)); });
+
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::TERMINATE);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 10));
+    TIMING_TEST_EXPECT_TRUE(hasTerminated);
+});
+
+TIMING_TEST_F(Timer_test, SoftToHardTimerChangesBehaviorToTerminate, Repeat(5), [&] {
```
elBoberido, line 453: please adjust name, there is no `SoftTimer` or `HardTimer` anymore

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -406,3 +407,117 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
+        // wait slightly longer then the timeout so that the effect is better measurable
+        std::this_thread::sleep_for(std::chrono::microseconds(TIMEOUT.milliSeconds<int>() * 1100));
+    });
+
+    sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.milliSeconds<int>() * 100));
+    // every second callback is skipped since the runtime is slightly longer therefore
+    // the counter must be in that range
+    TIMING_TEST_EXPECT_TRUE(50 <= counter && counter <= 70);
```
elBoberido, line 482: the tests would run faster if e.g. the interval would be 100ms and the callback execution time 150ms. If the time from exiting the first timer invocation to the time from entering the second timer invocation would be less than 1 ms the tests fails. On the second timer invocation a semaphore could be posted to continue a blocking timed wait instead of the thread sleep in the main thread.
The `SoftTimerCallsLessCallbacksThanASAPTimer` test should also not be needed with this approach.

Don't know which approach would be more robust, though

elfenpiff: At the moment I would leave it as it is since this test is also marked as TIMING_TEST.

Issue #171, opened 2020-07-03
Roudi message queue thread startup


dkroenke:
When starting RouDi immediatlly after the applications , it can happen that the Message Queue in RouDi is not yet ready and the applications already start to send messages. This could lead to a segmentation fault in RouDi. 

To prevent that we let the Message Thread start only when the RouDi has started complety.

Issue #150, opened 2020-06-23
The destroy method of MessageQueue and UnixDomainSocket does not fully invalidate the object


elBoberido:
## Required information

**Operating system:**
All

**Compiler version:**
All supported

**Observed result or behaviour:**
When a client is started with an outdate roudi message queue, the console is flooded with error messages, when roudi is started and the obsolete message queue is deleted. Additionally, there is a high CPU load until the new message queue is created.

**Expected result or behaviour:**
The situation should be gracefully handled with just one error message.

**Conditions where it occurred / Performed steps:**
1. Start `roudi`
2. pkill -9 roudi
3. start any client
4. start `roudi`

PR #172, opened 2020-07-03
Iox #171 roudi message queue thread startup

dkroenke:


elBoberido:
could you maybe rebase on master? The 20e00f4 commit already went in with #150

ithier:
No description ?

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
```diff
@@ -41,17 +41,33 @@ using namespace iox::units::duration_literals;
...
+   
+
     RouDiMultiProcess& operator=(const RouDiMultiProcess& other) = delete;
     RouDiMultiProcess(const RouDiMultiProcess& other) = delete;
 
     RouDiMultiProcess(RouDiMemoryInterface& roudiMemoryInteface,
                       PortManager& portManager,
                       const MonitoringMode f_monitoringMode = MonitoringMode::ON,
-                      const bool f_killProcessesInDestructor = true);
+                      const bool f_killProcessesInDestructor = true,
```
ithier, line 59: Very minor but the variable is inconsistent with the `.cpp`.
`killProcessesInDestructor` vs `f_killProcessesInDestructor`.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
```diff
@@ -41,17 +41,33 @@ using namespace iox::units::duration_literals;
 class RouDiMultiProcess
 {
   public:
+
+    // indicate whether the message queue thread will start directly or deferred
+    // this is important for derived classes which may need to initialize their members before the thread starts
+    enum class MQThreadMode {
+      START,
+      DEFER_START
```
ithier, line 49: This option seems redundant since the deferred start is not managed by this class. Could we get away with just the one option?

I would also suggest a slightly more  descriptive name like:
```
    enum class MQThreadMode {
        IMMEDIATE_START
    };
```

PR #174, opened 2020-07-06
 iox-#171 added option to defer mq thread startup

dkroenke:
Later startup of Roudi Message Queue Thread

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
@@ -44,8 +44,8 @@ class RouDiMultiProcess
 
     // indicate whether the message queue thread will start directly or deferred
     // this is important for derived classes which may need to initialize their members before the thread starts
-    enum class MQThreadMode {
-      START,
+    enum class MQThreadStart {
+      IMMEDIATE,
       DEFER_START
     };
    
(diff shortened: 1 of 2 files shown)


iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
```diff
@@ -41,17 +41,33 @@ using namespace iox::units::duration_literals;
 class RouDiMultiProcess
 {
   public:
+
+    // indicate whether the message queue thread will start directly or deferred
+    // this is important for derived classes which may need to initialize their members before the thread starts
+    enum class MQThreadMode {
```
elBoberido, line 47: why not call it like in the comment above?
```
enum class MqThreadStart
{
    DIRECT,
    DEFERRED
}
```
This would make the intention more obvious when reading, e.g. `MqThreadStart::DIRECT` ... maibe `IMMEDIATE` would also work

MatthiasKillat: Yes, would probably be more explicit and make the intention obvious.

dkroenke: its renamed

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_multi_process.hpp
```diff
@@ -41,17 +41,33 @@ using namespace iox::units::duration_literals;
 class RouDiMultiProcess
 {
   public:
+
+    // indicate whether the message queue thread will start directly or deferred
+    // this is important for derived classes which may need to initialize their members before the thread starts
+    enum class MQThreadStart {
+      IMMEDIATE,
+      DEFER_START
```
elBoberido, line 49: since "start" is already in the name of the enum, it's superfluous here

budrus:
Short question. When the parameter is not immediate, the derived class has to call the method for starting the thread?

PR #176, opened 2020-07-07
Iox #160 optional and expected adjustments

### LINKED ISSUE #160: cxx::optional and cxx::expected adjustments
## Brief feature description

- The methods of `cxx::expected` `on_success` and `on_error` should be renamed to `and_then` and `or_else`.
- It should be possible to create an `cxx::optional` directly from an `cxx::expected` with a method called `.to_optional()` which returns the value of the `cxx::expected` packed inside an `cxx:optional` if it contains a value otherwise it returns the `cxx::nullopt`
- If a value is moved outside the `cxx::optional` the method `has_value` should no longer return true since we would like to avoid undefined behavior
```
cxx::optional<int> a = 5;
if ( a.has_value() ); // returns true
int b = std::move(*a);
if ( a.has_value() ); // should now return false since the value is moved outside the optional

// this case would false assume that the optional has no more contained value but at the moment I do not see a problem with that.
std::move(*a);
```
- If a value is moved outside the `cxx::expected` it should no longer be contained. We need the methods `has_value()` and `has_error()` which both should return false in that case. If the methods `and_then()` or `or_else()` are called on an `cxx::expected` in that state they should terminate.

## Detailed information

We would further reduce undefined behavior.

### LINKED ISSUE #175: unable to compile on OSX
## Required information

**Operating system:**
Mojave

**Compiler version:**
clang-1001.0.46.4

```
Starting >>> iceperf
--- stderr: iceoryx_introspection
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LIB_TINFO
    linked by target "iceoryx_introspection" in directory /Users/karsten/workspace/ros2/iceoryx_ws/osx/src/iceoryx/tools/introspection

---
```

clean compilation

`colcon build` in a workspace with `iceoryx` and `rmw_iceoryx`.

elfenpiff:
Fixes #160 
Fixes #175

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_components.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_components.hpp
@@ -48,9 +48,9 @@ struct IceOryxRouDiComponents
             // if there is an outdated mqueue, the start of the apps will be terminated
             runtime::MqBase::cleanupOutdatedMessageQueue(MQ_ROUDI_NAME);
 
-            m_rouDiMemoryManager.createAndAnnounceMemory().on_error([](cxx::expected<RouDiMemoryManagerError> error) {
+            m_rouDiMemoryManager.createAndAnnounceMemory().or_else([](RouDiMemoryManagerError error) {
                 /// @todo errorHandler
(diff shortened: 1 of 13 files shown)


budrus:
How about the evaluation of expected and optional costs @elfenpiff @elBoberido?

PR #178, opened 2020-07-08
Iox #33 windows compiles again

### LINKED ISSUE #179: Deadlock in Posix Timer implementation for Mac Os and Windows
## Required information

**Operating system:**
MacOS Mojave & Windows 10

**Compiler version:**
clang 11.0 & msvc 16

The following posix timer unit tests are hanging sporadically:
  - StoppingIsNonBlocking
  - CallbackExecutedOnceAfterStart

No dead locks

Call the unit tests very often and 1 in 100 cases it will deadlock there (but even maybe less)

### LINKED ISSUE #182: Undefined Behavior in SoFi
## Required information

**Operating system:**
All OS

**Compiler version:**
All compiler

The variable `nextWritePosition` is not being initialized on startup in the method `popIf`. If sofi is not empty (`currentReadPosition != m_writePosition`), the verificator returns false and the `currentReadPosition` equals `m_readPosition` then this uninitialized value is written into `m_readPosition` by the `compare_exchange_weak` call.
This cannot be fixed by initializing the `nextWritePosition` to `0u`. It must be done in the condition ` if (m_readPosition.load(std::memory_order_relaxed) == currentReadPosition && verificator(valueOut) == false)` in line 117 and the value has to be `currentReadPosition`

It could lead to a theoretical corrupted sofi

elfenpiff:
Fixes #179 
Fixes #182


iceoryx_utils/platform/win/source/time.cpp
```diff
@@ -0,0 +1,187 @@
...
+static std::chrono::nanoseconds getNanoSeconds(const timespec& value)
+{
+    static constexpr uint64_t NANOSECONDS = 1000000000u;
+    return std::chrono::nanoseconds(static_cast<uint64_t>(value.tv_sec) * NANOSECONDS
+                                    + static_cast<uint64_t>(value.tv_nsec));
+}
+
+static void stopTimerThread(timer_t timerid)
+{
+    timerid->parameter.mutex.lock();
```
ithier, line 26: Why no camelCase ?

iceoryx_utils/platform/win/source/time.cpp
```diff
@@ -0,0 +1,187 @@
...
+    else
+    {
+        setTimeParameters(timerid, *new_value, false, true);
+    }
+    timerid->parameter.wakeup.notify_one();
+
+    return 0;
+}
+
+int timer_gettime(timer_t timerid, struct itimerspec* curr_value)
```
ithier, line 135: The mismatch in variable naming convention is making me cringe. Is this mimicing a system API ? We have camelCase, snake_case and lowercase mixed in the same function !

elfenpiff: Yes you are right but we have to reimplement the posix functions like `timer_gettime` https://linux.die.net/man/2/timer_gettime here since they are not available on Windows. Therefore, we have to stick strictly to the code styles of the posix standard.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
```diff
@@ -116,6 +116,7 @@ inline bool SoFi<ValueType, CapacityValue>::popIf(ValueType& valueOut, const Ver
             ///        changed
             if (m_readPosition.load(std::memory_order_relaxed) == currentReadPosition && verificator(valueOut) == false)
             {
+                nextReadPosition = currentReadPosition;
```
mossmaurice, line 119: A bugfix should always result in a new unit test case. Can you add one that reconstructs the bug?

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -21,7 +21,7 @@ using namespace iox::units;
 constexpr int KILO = 1000;
 constexpr int MEGA = 1000000;
 constexpr int GIGA = 1000000000;
-constexpr double doubleEpsilon = std::numeric_limits<double>::epsilon();
+constexpr double doubleEpsilon = std::numeric_limits<double>::epsilon() * 10.;
```
mossmaurice, line 24: Misleading name

iceoryx_utils/include/iceoryx_utils/internal/concurrent/sofi.inl
elfenpiff: Nice idea but sadly this is impossible since it is a freaky race condition. With total control over all threads it would be possible but there is no such thing available.

mossmaurice: @GerdHirsch always created check-points in the code. I guess we would need something similar here. However, you are right, the rule "bugfix==at least one additional test case" is a general rule. I've not fully understood the race yet, can we somehow formalize it?

elfenpiff: I think that the check point approach is not the right solution for this particular problem.
@mossmaurice The race is explained in detail in the following issue: https://github.com/eclipse/iceoryx/issues/182

iceoryx_utils/include/iceoryx_utils/cxx/variant.hpp
```diff
@@ -62,7 +63,7 @@ struct in_place_type
 ///     // variant with setted value therefore the index is not invalid
 ///     if ( someVariant.index() != INVALID_VARIANT_INDEX ) ...
 /// @endcode
-static constexpr uint64_t INVALID_VARIANT_INDEX = -1ULL;
+static constexpr uint64_t INVALID_VARIANT_INDEX = std::numeric_limits<uint64_t>::max();
```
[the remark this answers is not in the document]
elfenpiff: Numeric limits is using internally a hardcoded compile time value and is therefore safe to use and very easy to reimplement. Additionally, it is the safer and cleaner approach. @elBoberido what do you think?

iceoryx_utils/test/moduletests/test_unit_duration.cpp
elfenpiff: done

.github/workflows/build-test.yml
```diff
@@ -77,3 +77,24 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Build sources
+        run: |
+          cmake -Bbuild -Hiceoryx_meta -Dtest=ON && cmake --build build
+
+      - name: Run tests
+      # at the moment we do not support windows therefore we just run a subset of all the tests
+        run: |
+          build\utils\test\Debug\utils_moduletests.exe --gtest_filter="Semaphore*"
```
dkroenke, line 99: We need to ensure full windows compability and enable the other tests too. I created a ticket for this: https://github.com/eclipse/iceoryx/issues/185

iceoryx_utils/platform/win/source/time.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+    constexpr int64_t NANO_SECONDS{1000000000};
+    timespec currentTime;
+    clock_gettime(CLOCK_REALTIME, &currentTime);
+    int64_t currentTimeNs = currentTime.tv_sec * NANO_SECONDS + currentTime.tv_nsec;
```
dkroenke, line 140: time should be uint64 or?

dkroenke:
The branch should be rebased because there are changes from @elfenpiff in it which are already there.

iceoryx_utils/test/moduletests/test_file_reader.cpp
```diff
@@ -12,15 +12,28 @@
...
+#include <string>
+
 
 using namespace ::testing;
 
-const char* TestFileName = "/tmp/FileReader_test.tmp";
+const std::string TestFile = "FileReader_test.tmp";
+#ifndef _WIN32
+const std::string TempPath = "/tmp";
+const std::string CrapPath = "/All/Hail/Hypnotoad";
```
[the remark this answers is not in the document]
dkroenke: The branch should be rebased because these are changes from @elfenpiff

iceoryx_utils/platform/win/source/time.cpp
dkroenke: ok int64 is fine

iceoryx_utils/platform/mac/source/time.cpp
```diff
@@ -23,7 +23,10 @@ static std::chrono::nanoseconds getNanoSeconds(const timespec& value)
 
 static void stopTimerThread(timer_t timerid)
 {
-    timerid->keepRunning.store(false);
+    timerid->parameter.mutex.lock();
```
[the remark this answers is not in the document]
elfenpiff: Do we need traceability for non safety supported platforms like mac os?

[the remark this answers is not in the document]
mossmaurice: Yes, we do.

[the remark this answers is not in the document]
elfenpiff: Alright, it is fixed.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -496,7 +496,7 @@ TIMING_TEST_F(Timer_test, CatchUpPolicyImmediateCallsCallbackImmediatelyAfterFin
 
     // the asap timer should in theory call the callback 90 times since it is calling it right
     // after the last one finished and one callback takes 1.1 ms and we run for 100ms.
-    TIMING_TEST_EXPECT_TRUE(80 <= counter && counter <= 100);
+    TIMING_TEST_EXPECT_TRUE(70 < counter && counter <= 100);
```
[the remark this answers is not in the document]
elfenpiff: The two tests have disjunctive intervals but I would leave it as it is since these timing tests are extremely fragile

Issue #168, opened 2020-07-02
Add unittests for capro message


dkroenke:
increase the test coverage by writing unittests for capro message in https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/capro/capro_message.hpp

PR #181, opened 2020-07-08
Iox #168 unit test for capro message

dkroenke:
adding unittests for capro message class.

### DIFF
--- a/iceoryx_posh/test/moduletests/test_capro_message.cpp
+++ b/iceoryx_posh/test/moduletests/test_capro_message.cpp
@@ -1,0 +1,51 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "iceoryx_posh/internal/capro/capro_message.hpp"
+#include "iceoryx_posh/internal/roudi/introspection/port_introspection.hpp"
+#include "test.hpp"
+
+using namespace ::testing;
+// using ::testing::Return;
+using namespace iox::capro;
+
+class CaproMessage_test : public Test
+{
+  public:
+    void SetUp(){};
+    void TearDown(){};
+};
+
+TEST_F(CaproMessage_test, CtorNoParams)
+{
+    CaproMessage testObj = CaproMessage();
+    EXPECT_EQ(testObj.m_requestPort, nullptr);
+    EXPECT_EQ(testObj.m_type, CaproMessageType::NOTYPE);
+    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::NOSUBTYPE);
+}
+
+
+TEST_F(CaproMessage_test, CtorParams)
+{
+    uint16_t testServiceID = 1;
+    uint16_t testEventID = 2;
+    uint16_t testInstanceID = 3;
+    ServiceDescription sd = ServiceDescription(testServiceID, testEventID, testInstanceID);
+    iox::popo::ReceiverPortData recData;
+
+    CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd, CaproMessageSubType::SERVICE, &recData);
+    EXPECT_EQ(testObj.m_requestPort, &recData);
+    EXPECT_EQ(testObj.m_type, CaproMessageType::OFFER);
+    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::SERVICE);
+}
--- a/iceoryx_posh/test/moduletests/test_capro_message.cpp
+++ b/iceoryx_posh/test/moduletests/test_capro_message.cpp
@@ -29,11 +29,11 @@ class CaproMessage_test : public Test
 TEST_F(CaproMessage_test, CtorNoParams)
 {
     CaproMessage testObj = CaproMessage();
-    EXPECT_EQ(testObj.m_requestPort, nullptr);
-    EXPECT_EQ(testObj.m_type, CaproMessageType::NOTYPE);
-    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::NOSUBTYPE);
-    EXPECT_EQ(testObj.m_chunkQueueData, nullptr);
-    EXPECT_EQ(testObj.m_historyCapacity, 0u);
+    EXPECT_EQ(nullptr, testObj.m_requestPort);
+    EXPECT_EQ(CaproMessageType::NOTYPE, testObj.m_type);
+    EXPECT_EQ(CaproMessageSubType::NOSUBTYPE, testObj.m_subType);
+    EXPECT_EQ(nullptr, testObj.m_chunkQueueData);
+    EXPECT_EQ(0u, testObj.m_historyCapacity);
 }
 
 
@@ -46,12 +46,12 @@ TEST_F(CaproMessage_test, CtorParams)
     iox::popo::ReceiverPortData recData;
 
     CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd, CaproMessageSubType::SERVICE, &recData);
-    EXPECT_EQ(testObj.m_requestPort, &recData);
-    EXPECT_EQ(testObj.m_type, CaproMessageType::OFFER);
-    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::SERVICE);
-    EXPECT_EQ(testObj.m_chunkQueueData, nullptr);
-    EXPECT_EQ(testObj.m_historyCapacity, 0u);
-    EXPECT_EQ(testObj.m_serviceDescription, sd);
+    EXPECT_EQ(&recData, testObj.m_requestPort);
+    EXPECT_EQ(CaproMessageType::OFFER, testObj.m_type);
+    EXPECT_EQ(CaproMessageSubType::SERVICE, testObj.m_subType);
+    EXPECT_EQ(nullptr, testObj.m_chunkQueueData);
+    EXPECT_EQ(0u, testObj.m_historyCapacity);
+    EXPECT_EQ(sd, testObj.m_serviceDescription);
 }
 
 
@@ -64,6 +64,6 @@ TEST_F(CaproMessage_test, CtorDefaultArgs)
     iox::popo::ReceiverPortData recData;
 
     CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd);
-    EXPECT_EQ(testObj.m_requestPort, nullptr);
-    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::NOSUBTYPE);
+    EXPECT_EQ(nullptr, testObj.m_requestPort);
+    EXPECT_EQ(CaproMessageSubType::NOSUBTYPE, testObj.m_subType);
 }
--- a/iceoryx_posh/test/moduletests/test_capro_message.cpp
+++ b/iceoryx_posh/test/moduletests/test_capro_message.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2019-2020 by Robert Bosch GmbH. All rights reserved.
 //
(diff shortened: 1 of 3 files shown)


iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,68 @@
...
+using namespace iox::capro;
+
+class CaproMessage_test : public Test
+{
+  public:
+    void SetUp(){};
+    void TearDown(){};
+};
+
+TEST_F(CaproMessage_test, CtorNoParams)
```
elfenpiff, line 29: Since the default constructor of `CaproMessage` is initializing `m_requestPort` to `nullptr` which is already done directly in the header I would suggest to remove the default constructor implementation and let the compiler generate it like:
```
CaproMessage() =default;
```

With that done we could also remove the `CtorNoParams` test since the compiler is generating everything for us and we do not test compiler generated code.

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,68 @@
...
+    CaproMessage testObj = CaproMessage();
+    EXPECT_EQ(testObj.m_requestPort, nullptr);
+    EXPECT_EQ(testObj.m_type, CaproMessageType::NOTYPE);
+    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::NOSUBTYPE);
+    EXPECT_EQ(testObj.m_chunkQueueData, nullptr);
+    EXPECT_EQ(testObj.m_historyCapacity, 0u);
+}
+
+
+TEST_F(CaproMessage_test, CtorParams)
```
elfenpiff, line 40: Please state in unit test names what you are doing and what you are expecting. As suggestion:
`CTorSetsParametersCorrectly`

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,68 @@
...
+    CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd, CaproMessageSubType::SERVICE, &recData);
+    EXPECT_EQ(testObj.m_requestPort, &recData);
+    EXPECT_EQ(testObj.m_type, CaproMessageType::OFFER);
+    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::SERVICE);
+    EXPECT_EQ(testObj.m_chunkQueueData, nullptr);
+    EXPECT_EQ(testObj.m_historyCapacity, 0u);
+}
+
+
+TEST_F(CaproMessage_test, CtorDefaultArgs)
```
elfenpiff, line 57: Maybe a name like `DefaultCaproMessageSubTypeIsNOSUBTYPE`. Additionally we could split this test and write another one called `DefaultRequestPortIsNullptr`

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+TEST_F(CaproMessage_test, CtorNoParams)
+{
+    CaproMessage testObj = CaproMessage();
+    EXPECT_EQ(testObj.m_requestPort, nullptr);
+    EXPECT_EQ(testObj.m_type, CaproMessageType::NOTYPE);
+    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::NOSUBTYPE);
+    EXPECT_EQ(testObj.m_chunkQueueData, nullptr);
+    EXPECT_EQ(testObj.m_historyCapacity, 0u);
```
budrus, line 36: After getting an extra lesson from @elfenpiff, I'm a fan of the AAA pattern. Having three sections that are separated by an empty line is sufficient IMHO. This already makes it more readable
https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+TEST_F(CaproMessage_test, CtorParams)
+{
+    uint16_t testServiceID = 1;
+    uint16_t testEventID = 2;
+    uint16_t testInstanceID = 3;
+    ServiceDescription sd = ServiceDescription(testServiceID, testEventID, testInstanceID);
+    iox::popo::ReceiverPortData recData;
+
+    CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd, CaproMessageSubType::SERVICE, &recData);
```
budrus, line 48: here a blank line in between

iceoryx_posh/test/moduletests/test_capro_message.cpp
nihalchari: Adding separate test case for each default value arguments might be redundant, will simply increase testcases

iceoryx_posh/test/moduletests/test_capro_message.cpp
nihalchari: introduced blank line

iceoryx_posh/test/moduletests/test_capro_message.cpp
dkroenke: done

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,68 @@
...
+    uint16_t testServiceID = 1;
+    uint16_t testEventID = 2;
+    uint16_t testInstanceID = 3;
+    ServiceDescription sd = ServiceDescription(testServiceID, testEventID, testInstanceID);
+    iox::popo::ReceiverPortData recData;
+
+    CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd);
+    EXPECT_EQ(testObj.m_requestPort, nullptr);
+    EXPECT_EQ(testObj.m_subType, CaproMessageSubType::NOSUBTYPE);
+}
```
dkroenke: missing empty line at the end of file

iceoryx_posh/test/moduletests/test_capro_message.cpp
dkroenke: done

iceoryx_posh/test/moduletests/test_capro_message.cpp
dkroenke: done

iceoryx_posh/test/moduletests/test_capro_message.cpp
dkroenke: done

iceoryx_posh/test/moduletests/test_capro_message.cpp
dkroenke: done

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,63 @@
...
+    virtual void TearDown(){};
+};
+
+
+TEST_F(CaproMessage_test, CTorSetsParametersCorrectly)
+{
+    uint16_t testServiceID = 1;
+    uint16_t testEventID = 2;
+    uint16_t testInstanceID = 3;
+    ServiceDescription sd = ServiceDescription(testServiceID, testEventID, testInstanceID);
```
MatthiasKillat, line 43: [Prefer direct Ctor call] Preferably use ```ServiceDescription sd(testServiceID, testEventID, testInstanceID);``` Another option is ```auto sd = ServiceDescription(testServiceID, testEventID, testInstanceID);```. All of these just call the Ctor, but the preferred option is the shortest.

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,63 @@
...
+    iox::popo::ReceiverPortData recData;
+
+    CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, sd, CaproMessageSubType::SERVICE, &recData);
```
MatthiasKillat, line 46: [Prefer direct Ctor call], as above.

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    virtual void SetUp(){};
+    virtual void TearDown(){};
+};
+
+
+TEST_F(CaproMessage_test, CTorSetsParametersCorrectly)
+{
+    uint16_t testServiceID = 1;
```
MatthiasKillat, line 40: These three could be constexpr, like so```constexpr uint16_t testServiceId{1}``` This leads to (slight) optimization potential. General rule: try to make const or constexpr what is possible. Note that in this case I would not advocate to use CAPITAL_LETTER_STYLE for such constexprs (only for constants with a large scope, e.g. at class scope), i.e. leave the names as is.

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,63 @@
...
+// limitations under the License.
+
+#include "iceoryx_posh/internal/capro/capro_message.hpp"
+#include "iceoryx_posh/internal/roudi/introspection/port_introspection.hpp"
+#include "test.hpp"
+
+using namespace ::testing;
+using namespace iox::capro;
+
+class CaproMessage_test : public Test
```
MatthiasKillat, line 22: As it is, we do not need a test fixture CaproMessage_test and could use TEST instead of TEST_F. (We generate all necessary data in the test itself and not in the test fixture). But we can keep it for future use (could be handy when the tests are be extended).

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,63 @@
...
+    EXPECT_EQ(CaproMessageSubType::SERVICE, testObj.m_subType);
+    EXPECT_EQ(nullptr, testObj.m_chunkQueueData);
+    EXPECT_EQ(0u, testObj.m_historyCapacity);
+    EXPECT_EQ(sd, testObj.m_serviceDescription);
+}
+
+
+TEST_F(CaproMessage_test, DefaultCtorArgs)
+{
+    CaproMessage testObj = CaproMessage(CaproMessageType::OFFER, ServiceDescription{});
```
MatthiasKillat, line 59: [Prefer direct Ctor call], as above.

iceoryx_posh/test/moduletests/test_capro_message.cpp
```diff
@@ -0,0 +1,63 @@
...
+    EXPECT_EQ(&recData, testObj.m_requestPort);
+    EXPECT_EQ(CaproMessageType::OFFER, testObj.m_type);
+    EXPECT_EQ(CaproMessageSubType::SERVICE, testObj.m_subType);
+    EXPECT_EQ(nullptr, testObj.m_chunkQueueData);
+    EXPECT_EQ(0u, testObj.m_historyCapacity);
+    EXPECT_EQ(sd, testObj.m_serviceDescription);
+}
+
+
+TEST_F(CaproMessage_test, DefaultCtorArgs)
```
MatthiasKillat, line 57: I do not fully understand the test. We do not call a default Ctor here but we call it with a default ServiceDescription. Maybe the test needs a better name, such as ```constructWithDefaultServiceDescription```. Also, does it make sense to check more fields in the via EXPECT afterwards? (I am not sure here, we only need to check what is specific to this case and was not tested elsewhere.)

nihalchari: This test is to check default arguments in constructor if initialized properly, 
ServiceDescription is not default argument here, so it is passed(default/initialized may not matter), 
and we are checking for other two (CaproMessageSubTyp , ReceiverPortData) if  initialized properly.
for name ConstructWithDefaultArguments is fine?

nihalchari: CaproMessageType is also passed as it is not a default argument

iceoryx_posh/test/moduletests/test_capro_message.cpp
nihalchari: okay

iceoryx_posh/test/moduletests/test_capro_message.cpp
nihalchari: okay

iceoryx_posh/test/moduletests/test_capro_message.cpp
nihalchari: okay

iceoryx_posh/test/moduletests/test_capro_message.cpp
MatthiasKillat: Yes, but the main point seems to be that we initialize it with a default constructed ServiceDescription, i.e. ```ServiceDescription{}```. I was misled to believe it has something to do with the default constructor. I would just change the name of the test (e.g. as suggested) to make that more clear.

Issue #155, opened 2020-06-25
Add Helix QAC support to build scripts


mossmaurice:
## Brief feature description

CI shall be able to easily build iceoryx for QAC++ scan

## Detailed information

Necessary JSON can be generated with CMake (-DCMAKE_EXPORT_COMPILE_COMMANDS=ON)

PR #195, opened 2020-07-15
iox-#155 Add qacpp option to build script

mossmaurice:


tools/iceoryx_build_test.sh
```diff
@@ -49,6 +50,11 @@ do
         "strict")
             STRICT_FLAG="on"
             ;;
+        "qacpp")
+            BUILD_TYPE="Release"
+            QACPP_JSON="on"
+            INTROSPECTION_FLAG="off"
```
dkroenke, line 56: Please enable introspection, we want to scan it too.

mossmaurice: Done

Issue #198, opened 2020-07-20
Introduce 


mossmaurice:
## Brief feature description

Different CMake targets will be developed according to different quality levels. This shall be documented.

## Detailed information

* Consider https://github.com/ros-infrastructure/rep/blob/master/rep-2004.rst

Issue #83, opened 2020-04-07
ASIL-D compliance 


ShawnSchaerer:
Is there any updates on ASIL-D compliance and how do you plan on getting there?
We are looking at using Iceoryx in a commercial product and would like to understand the roadmap for safety.

budrus:
Hi @ShawnSchaerer 
Nice to here. The safety roadmap is linked to the roadmaps of our customers. Current plan is to have an ASIL B in 2021 and an ASIL D in 2022 or 2023. In the end It is a matter of prioritization.  
There are two dimensions and the question of their balance
a) Having more features and more platforms: e.g. #27 and #33. 
b) Having the focus on production readiness and safety level X for a specific feature set and platform (like pub/sub on QNX)

a) is important for topics like becoming ROS2 tier 1. On the other side, ASIL compliance will not target all operating systems and has some other constraints. 

The safety topic has also two dimensions
a) how iceoryx is developed
b) how iceoryx is used  
  
a) is a question of development process, testing, etc. We have the know how for this. Here the challenge is that it is not easy to fulfill everything with just a github project. So we have some extra work with established internal processes and tools. One thing that we provide soon online are the QAC++ warning reports, here we are just setting up things with Perforce. The code base that we will have on ASIL level is currently restricted to QNX as OS.

b) If you want to use iceoryx in a specific environment, a safety manual is needed that guides you and it must be released/validated for this environment. Here we are talking about which compiler with version and flags you are using, on which OS and version and on which hardware.  You have to get in touch with us for these questions. Only the code is not the full answer to the ASIL question

ShawnSchaerer:
@budrus Thanks for the repsonce.
I will email to discuss in more detail, but I understand what you are saying.  
We are just making sure that their is a road map for ASIL certification or that it will be possible later.

Our system is running on QNX so there should be a way forward for us.

PR #199, opened 2020-07-20
iox-#198 Quality level proposal

mossmaurice:
This is pull request is meant to start the discussion on the quality levels, I suppose it is not ready yet ;-)

@Karsten1987 @budrus @elBoberido What do you guys think?


README.md
```diff
@@ -40,12 +40,20 @@ An example for a "porcelain" API would be [ROS2](https://www.ros.org/). Others a
...
-### Supported Platforms
+### Targeted quality levels & platforms
 
- * Linux
- * QNX
- * macOS (initial support - please test and report bugs)
- * Windows 10 (not yet - currently in progress)
+> [Quality level](./CONTRIBUTING.md#quality-levels) are default, A, B, C, D, where D is highest
+
+|CMake project          | Quality level D on QNX | Quality level B on QNX | Default quality on Linux, Windows, Mac, QNX  | Comment                             |
```
elBoberido, line 47: QNX is twice
How about something like
```
| CMake project/target | QNX | Linux | Windows | macOS |
| :--- | :---: | :---: | :---: | :---: |
|iceoryx_utils| B, D | Q | Q | Q |
```
and so on

CONTRIBUTING.md
```diff
@@ -194,3 +194,39 @@ Each source file needs to have this header:
...
+### Default quality level
+
+This quality level is meant for examples and
+
+* Reviewed by two approver
+* License and copyright statement available
+* No version policy required
+* No unit tests required
+
+### Quality level A
```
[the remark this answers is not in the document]
mossmaurice: I would explicitly not mention ASIL here to avoid any misunderstandings. Only the code is not the full answer to the ASIL question, see #83

CONTRIBUTING.md
```diff
@@ -194,3 +194,39 @@ Each source file needs to have this header:
...
+
+## Quality levels
+
+CMake targets can be developed according to different quality levels. All requirements of a lower level are included.
+
+Also see [ROS quality levels](https://github.com/ros-infrastructure/rep/blob/master/rep-2004.rst).
+
+### Default quality level
+
+This quality level is meant for examples and
```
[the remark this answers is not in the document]
mossmaurice: Yeah, seems like I did a `git add` too early :blush:

CONTRIBUTING.md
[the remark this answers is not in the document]
elBoberido: In this case I would probably use number instead of letter to avoid misunderstandings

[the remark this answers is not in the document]
mossmaurice: Done

README.md
```diff
@@ -40,12 +40,20 @@ An example for a "porcelain" API would be [ROS2](https://www.ros.org/). Others a
...
+> [Quality level](./CONTRIBUTING.md#quality-levels) are 5..1, where 1 is highest
+
+|CMake project/target   | QNX  | Linux, Windows, MacOS | Comment                             |
+|-----------------------|:----:|:---------------------:|:-----------------------------------:|
+| example_icedelivery   | 4    | 4                     |                                     |
+| example_iceperf       | 4    | 4                     |                                     |
+| example_singleprocess | 4    | 4                     |                                     |
+| iceoryx_dds           | 4    | 4                     |                                     |
+| iceoryx_meta          | 4    | 4                     |                                     |
+| iceoryx_posh          | 1, 2 | 4                     | Will be split into separate targets |
```
elBoberido, line 54: For Linux, I would assume this to be at least a 3, don't you think so?
And since we want to support all tier 1 targets or ROS, the same should apply for Windows and MacOS. Maybe by adding compiler and QAC warnings to level 2.

README.md
```diff
@@ -40,12 +40,20 @@ An example for a "porcelain" API would be [ROS2](https://www.ros.org/). Others a
...
+| iceoryx_utils         | 1    | 4                     |                                     |
+| iceoryx_introspection | 4    | 4                     |                                     |
```
elBoberido, line 56: shouldn't this be `iox-introspecton-client`? It might confuse people since we also have introspection parts in iceoryx_posh

CONTRIBUTING.md
```diff
@@ -194,3 +194,36 @@ Each source file needs to have this header:
     // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
     // See the License for the specific language governing permissions and
     // limitations under the License.
+
+## Quality levels
+
+CMake targets can be developed according to different quality levels. For production usage in a specific environment (e.g. compiler version, compiler flags, etc.), please contact one of the maintainers beforehand. All requirements of a lower quality level are included in higher quality levels e.g. quality level 4 is included in quality level 3.
```
budrus, line 200: What is meant with production usage in this context? We cannot provide any guarantees here even if compiler version and flags are provided

CONTRIBUTING.md
```diff
@@ -194,3 +194,36 @@ Each source file needs to have this header:
...
+* No compiler warnings
+* Doxygen and documentation available
+* Test specification available
+* Unit tests available
+* Version policy required
+* Level 8 and 9 warnings in Helix QAC addressed
+
+### Quality level 2
+
+* Code coverage statement and branch
```
budrus, line 224: maybe more clear with ": " Code coverage: statement and branch, or something like unit tests have full statement and branch coverage

CONTRIBUTING.md
```diff
@@ -194,3 +194,36 @@ Each source file needs to have this header:
...
+
+### Quality level 1
+
+* Warnings in Helix QAC addressed
```
budrus, line 228: This means no warnings or all warnings are evaluated and maybe some are disabled with the respective code comments?

README.md
```diff
@@ -40,12 +40,20 @@ An example for a "porcelain" API would be [ROS2](https://www.ros.org/). Others a
...
 | [Cyclone DDS](https://github.com/eclipse-cyclonedds/cyclonedds) | Performant and robust open-source DDS implementation maintained by [ADLINK Technology Inc.](https://www.adlinktech.com/) |
 
-### Supported Platforms
+### Targeted quality levels & platforms
 
- * Linux
- * QNX
- * macOS (initial support - please test and report bugs)
- * Windows 10 (not yet - currently in progress)
+> [Quality level](./CONTRIBUTING.md#quality-levels) are 5..1, where 1 is highest
```
budrus, line 45: Shall this table represent the target or the current state. I guess the first one. Do we also want to document the current state?

CONTRIBUTING.md
marthtz: Do we want to specify a way how to handle/describe a disabled warning?

README.md
elBoberido: we could also define iceoryx 1.0.0 should fulfil the requirements of the table, but we still should track the current state somewhere. Maybe like 42% done.

mossmaurice: Where would the 42% come from? Some points on this list are hard to measure. I'd suggest to provide coverage reports as described in #217

CONTRIBUTING.md
budrus: It is already described how to disable them with code comments and we also collect the justification

README.md
elBoberido: well, my gut feelings are telling me it's the answer to the question how much we achieved 😀

mossmaurice: Can we measure your gut feeling with a sensor and upload it? :-D

CONTRIBUTING.md
mossmaurice: The idea is to state, that just the code is not enough for production usage, but more documents are needed.

README.md
mossmaurice: Good point, but I'd suggest to extend level 4 instead of switch to level 3. Should we document the compilers in the CONTRIBUTING.md where we say which targets we have?

CONTRIBUTING.md
mossmaurice: > This means no warnings or all warnings are evaluated and maybe some are disabled with the respective code comments?

Yeah, that's meant here. Wanna rephrase it?

README.md
elBoberido: sure, but we need LoRa for that 😉

README.md
elBoberido: you teased me with 4 levels on line 45 but didn't specify it. Maybe 4 could become 5 and some parts of 3 could be moved to 4

mossmaurice: Done

README.md
mossmaurice: I would use the names of the CMake targets/projects `project(iceoryx_introspection ..)`. I guess the introspection parts of RouDi should be move to the other CMake target. What do you think?

CONTRIBUTING.md
mossmaurice: Done

CONTRIBUTING.md
mossmaurice: Rephrase it

README.md
elBoberido: yes, ideally there should be only one target per CMakeLists.txt and especially in iceoryx_posh it's confusing what's all included there

PR #202, opened 2020-07-21
iox-#25 Remove deprecated semaphore methods in ChunkQueuePopper

mossmaurice:


iceoryx_posh/test/moduletests/test_popo_condition_variable.cpp
```diff
@@ -140,3 +140,23 @@ TEST_F(ConditionVariable_test, NotifyWhileWaitingResultsNoTimeoutMultiThreaded)
...
+
+TEST_F(ConditionVariable_test, MoveAssignConditionVariableSignalerIsSuccessful)
+{
+    /// @todo
+}
+
+TEST_F(ConditionVariable_test, MoveAssignConditionVariableWaiterIsSuccessful)
+{
+    /// @todo
+}
```
mossmaurice, line 162: @elBoberido @elfenpiff When is a good timeslot to implement move and move assign for all building blocks? I'd say shortly before going live with #25

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
```diff
@@ -113,27 +113,41 @@ void ChunkQueuePopper::clear() noexcept
     } while (true);
 }
 
-/// @deprecated #25
-cxx::expected<ChunkQueueError>
-ChunkQueuePopper::attachSemaphore(const mepoo::SharedPointer<posix::Semaphore>& semaphore) noexcept
+bool ChunkQueuePopper::attachConditionVariableSignaler(ConditionVariableData* ConditionVariableDataPtr) noexcept
```
marthtz, line 116: Shouldn't the parameter start with lower case like "conditionVariableDataPtr"?

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
```diff
@@ -144,56 +146,52 @@ TEST_P(ChunkQueue_test, ClearWithData)
...
-    auto ret = m_popper.attachSemaphore(*semaphore);
-    EXPECT_FALSE(ret.has_error());
+    auto ret = m_popper.attachConditionVariableSignaler(&condVar);
+    EXPECT_TRUE(ret);
 
-    EXPECT_THAT(m_popper.isSemaphoreAttached(), Eq(true));
+    EXPECT_THAT(m_popper.isConditionVariableSignalerAttached(), Eq(true));
 }
 
 TEST_P(ChunkQueue_test, DISABLED_PushAndTriggersSemaphore)
```
marthtz, line 159: Rename test for condition variable. Is this test still required as it's disabled?

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
```diff
@@ -144,56 +146,52 @@ TEST_P(ChunkQueue_test, ClearWithData)
...
     auto chunk = allocateChunk();
     m_pusher.push(chunk);
 
-    EXPECT_THAT(semaphore->get()->tryWait(), Eq(true));
-    EXPECT_THAT(semaphore->get()->tryWait(), Eq(false)); // shouldn't trigger a second time
+    EXPECT_THAT(condVarWaiter.timedWait(1_ms), Eq(true));
+    EXPECT_THAT(condVarWaiter.timedWait(1_ms), Eq(false)); // shouldn't trigger a second time
 }
 
 TEST_P(ChunkQueue_test, DISABLED_AttachSecondSemaphore)
```
marthtz, line 174: Rename test for condition variable. Is this test still required as it's disabled?

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
```diff
@@ -144,56 +146,52 @@ TEST_P(ChunkQueue_test, ClearWithData)
...
-    auto ret2 = m_popper.attachSemaphore(*semaphore2);
-    EXPECT_TRUE(ret2.has_error());
-    ASSERT_THAT(ret2.get_error(), Eq(ChunkQueueError::SEMAPHORE_ALREADY_SET));
+    auto ret2 = m_popper.attachConditionVariableSignaler(&condVar2);
+    EXPECT_FALSE(ret2);
 
-    EXPECT_THAT(semaphore1->get()->tryWait(), Eq(false));
-    EXPECT_THAT(semaphore2->get()->tryWait(), Eq(false));
+    EXPECT_THAT(condVarWaiter1.timedWait(1_ms), Eq(false));
+    EXPECT_THAT(condVarWaiter2.timedWait(1_ms), Eq(false));
```
marthtz, line 188: Is it a good idea to use fix time durations? Is that why the test is currently disabled?

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
```diff
@@ -144,56 +146,52 @@ TEST_P(ChunkQueue_test, ClearWithData)
...
 
     auto chunk = allocateChunk();
     m_pusher.push(chunk);
 
-    EXPECT_THAT(semaphore1->get()->tryWait(), Eq(true));
-    EXPECT_THAT(semaphore2->get()->tryWait(), Eq(false));
+    EXPECT_THAT(condVarWaiter1.timedWait(1_ms), Eq(true));
+    EXPECT_THAT(condVarWaiter2.timedWait(1_ms), Eq(false));
```
marthtz, line 194: Is it a good idea to use fix time durations? Is that why the test is currently disabled?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.hpp
```diff
@@ -81,22 +81,14 @@ class ChunkQueuePopper
     /// @return True if successful, false if not
     bool attachConditionVariableSignaler(ConditionVariableData* ConditionVariableDataPtr) noexcept;
```
marthtz, line 82: Shouldn't the parameter start with lower case like "conditionVariableDataPtr"?

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
mossmaurice: Renamed the test. I don't know why @budrus disabled this test case. I suppose due to the missing lock in `attachConditonVariableSignaler` and `detachConditionVariableSignaler`.

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
mossmaurice: Renamed the test case. See answer above.

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
mossmaurice: I've used `timedWait()` here to make sure having a clean unit test. By using `timedWait()` it is ensure that the methods always returns. Please ping @budrus on why the tests are disabled.

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
mossmaurice: Same same

iceoryx_posh/include/iceoryx_posh/popo/condition.hpp
```diff
@@ -40,12 +40,12 @@ class Condition
   private:
     friend class WaitSet;
     /// @brief User interface for specific attach of condition variable
-    virtual bool setConditionVariable(ConditionVariableData* const ConditionVariableDataPtr) noexcept = 0;
+    virtual bool setConditionVariable(ConditionVariableData* const conditionVariableDataPtr) noexcept = 0;
```
elBoberido, line 43: do we really want this notation with `Ptr` postfix?

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
```diff
@@ -113,27 +113,41 @@ void ChunkQueuePopper::clear() noexcept
     } while (true);
 }
 
-/// @deprecated #25
-cxx::expected<ChunkQueueError>
-ChunkQueuePopper::attachSemaphore(const mepoo::SharedPointer<posix::Semaphore>& semaphore) noexcept
+bool ChunkQueuePopper::attachConditionVariableSignaler(ConditionVariableData* conditionVariableDataPtr) noexcept
 {
-    if (isSemaphoreAttached())
+    /// @todo Add lock guard here
```
elBoberido, line 118: keep in mind, this doesn't need to be added only here and in detach. It has to be added everywhere it is accessed. I would suggest to put the two members into a struct and use the smart_lock for that struct

iceoryx_posh/test/moduletests/test_popo_condition_variable.cpp
elBoberido: Are they currently deleted? If yes, then this test should be disabled with a `/// @note move c'tor currently deleted` or completely removed.
@elfenpiff and I had an idea how to easily check for compile failures, e.g. if you want to check if the move c'tor is really deleted. I should probably write an issue for that

mossmaurice: Yep, they are currently deleted. We decided to implement them as one of the last points of #25, hence a compile failure check is not needed here. Added the comment and disabled them.

iceoryx_posh/include/iceoryx_posh/popo/condition.hpp
mossmaurice: We have this everywhere in the new building blocks, so I decided to stick to the name to be consistent. No hard feels on this, I don't mind the `Ptr` abbrevation too much.

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
mossmaurice: Thanks for the hint. Using the `smart_lock` is a neat idea. Will implement it as part of the next step of #25 after #202 and #219 have been merged. Then we'll add the `ThreadSafePolicy` to both `ChunkQueuePopper`.

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
```diff
@@ -113,27 +113,41 @@ void ChunkQueuePopper::clear() noexcept
...
+        LogWarn() << "Condition variable signaler already set. Attaching a second time will be ignored!";
+        return false;
     }
     else
     {
-        getMembers()->m_semaphore = semaphore;
-        getMembers()->m_semaphoreAttached.store(true, std::memory_order_release);
-        return cxx::success<void>();
+        getMembers()->m_conditionVariableDataPtr = conditionVariableDataPtr;
+        getMembers()->m_conditionVariableAttached.store(true, std::memory_order_release);
```
budrus, line 127: Do we really need the bool flag? nullptr or != nullptr has the same information and it will be protected with a lock when introducing the locking policy as next step

iceoryx_posh/source/popo/building_blocks/chunk_queue_pusher.cpp
```diff
@@ -59,14 +59,11 @@ cxx::expected<ChunkQueueError> ChunkQueuePusher::push(mepoo::SharedChunk chunk)
             auto returnedChunk = mepoo::SharedChunk(chunkManagement);
         }
 
-
-        if (getMembers()->m_semaphoreAttached.load(std::memory_order_acquire) && getMembers()->m_semaphore)
+        if (getMembers()->m_conditionVariableAttached.load(std::memory_order_acquire) && getMembers()->m_conditionVariableDataPtr)
```
budrus, line 62: You also would not need this double check if your single source is the `m_conditionVariableDataPtr`. This double check would also be a race without a lock . You have to do the lock before the check and not after the check as indicated by the comment below

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.hpp
```diff
@@ -79,24 +79,16 @@ class ChunkQueuePopper
     /// @brief Attaches a condition variable signaler
     /// @param[in] ConditionVariableDataPtr, pointer to an condition variable data object
     /// @return True if successful, false if not
-    bool attachConditionVariableSignaler(ConditionVariableData* ConditionVariableDataPtr) noexcept;
+    bool attachConditionVariableSignaler(ConditionVariableData* conditionVariableDataPtr) noexcept;
```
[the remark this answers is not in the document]
mossmaurice: True, will fix.

iceoryx_posh/test/moduletests/test_popo_condition_variable.cpp
elBoberido: Well, we do not have something like this for other classes were we deleted the move c'tor and assignment operator. Not sure if it should just be deleted. We should maybe also create an issue to create compile failure tests for all classes which have deleted c'tors/assignment operators

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
elBoberido: well, is this much work?
If it doesn't blow up this PR, then I'd rather see it here. I fear we postpone to much to follow up PRs and might forget something and especially with concurrent code I'm rather save than sorry.

budrus:
Timing sensitive test?

iceoryx_posh/source/popo/building_blocks/chunk_queue_pusher.cpp
mossmaurice: Jup, that's easier. Done.

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
mossmaurice: Done

iceoryx_posh/source/popo/building_blocks/chunk_queue_popper.cpp
budrus: @marthtz just starts working on introducing the locking policy. So this will not be a long lasting topic

mossmaurice: Is this a blocker for you? If you're fine with it, I'd prefer to do it in the follow-up. Upfront, promised ;-)

mossmaurice:
> 
> 
> Timing sensitive test?

Fixed on master branch.

iceoryx_posh/test/moduletests/test_popo_condition_variable.cpp
mossmaurice: Sure, created #222

PR #207, opened 2020-07-24
Iox #206 solve cmake bug

### LINKED ISSUE #206: CMake Fails on Ubuntu & Debian
## Required information

**Operating system:**
Ubuntu 18.04 LTS, Ubuntu 19.04 LTS, Debian 10.4

**Compiler version:**
GCC 7.4.0, GCC 9.2, GCC 10.0

CMake fails with the following output
```
-- Build files have been written to: /builds/el.chris/larry-services/build/dependencies/cpptoml/download
Unknown argument -j1
Usage: cmake --build <dir> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  -j [<jobs>] --parallel [<jobs>] = Build in parallel using
                   the given number of jobs. If <jobs> is omitted
                   the native build tool's default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.
  --target <tgt> = Build <tgt> instead of default targets.
                   May only be specified once.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --use-stderr   = Ignored.  Behavior is default in CMake >= 3.0.
  --             = Pass remaining options to the native tool.
CMake Error at build/_deps/iceoryx-src/cmake/cpptoml/CMakeLists.txt:51 (message):
  Build step [download] for cpptoml failed: 1
-- Configuring incomplete, errors occurred!
See also "/builds/el.chris/larry-services/build/CMakeFiles/CMakeOutput.log".
See also "/builds/el.chris/larry-services/build/CMakeFiles/CMakeError.log".
```

CMake performs correctly

` cmake -Bbuild -Hiceoryx_meta`

elfenpiff:
Closes #206


cmake/cpptoml/CMakeLists.txt
```diff
@@ -17,11 +17,13 @@ project(cpptoml-build CXX)
...
-  set(CMAKE_BUILD_FLAGS -j${N})
-elseif(LINUX OR QNX)
-  set(CMAKE_BUILD_FLAGS -- -j${N})
-endif()
+
+### does not work on Debian 10.4 or Ubuntu 18.04
+# if(NOT N EQUAL 0 AND ((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
+#   set(CMAKE_BUILD_FLAGS -j${N})
+# elseif(LINUX OR QNX)
+#   set(CMAKE_BUILD_FLAGS -- -j${N})
```
elBoberido, line 25: could you try setting this to `set(CMAKE_BUILD_FLAGS "-- -j${N}")`

cmake/cpptoml/CMakeLists.txt
```diff
@@ -17,11 +17,13 @@ project(cpptoml-build CXX)
...
-if(NOT N EQUAL 0 AND ((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
-  set(CMAKE_BUILD_FLAGS -j${N})
-elseif(LINUX OR QNX)
-  set(CMAKE_BUILD_FLAGS -- -j${N})
-endif()
+
+### does not work on Debian 10.4 or Ubuntu 18.04
+# if(NOT N EQUAL 0 AND ((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
+#   set(CMAKE_BUILD_FLAGS -j${N})
+# elseif(LINUX OR QNX)
```
elBoberido, line 24: oh, and this should be `elseif(NOT N EQUAL 0 AND (LINUX OR QNX))`
maybe it's easier to graps if it's like this
```
if(NOT N EQUAL 0)
    if(bla bla)
    elseif(LINUX OR QNX)
    endif()
endif()
```

cmake/cpptoml/CMakeLists.txt
elfenpiff: This did not work but adding some space between `-j` and `${N}` worked ... How lovely

cmake/cpptoml/CMakeLists.txt
elBoberido: could you also take care of the above? Else linux and qnx will break if `N == 0`

cmake/cpptoml/CMakeLists.txt
```diff
@@ -17,10 +17,13 @@ project(cpptoml-build CXX)
 
 include(ProcessorCount)
 ProcessorCount(N)
-if(NOT N EQUAL 0 AND ((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
-  set(CMAKE_BUILD_FLAGS -j${N})
-elseif(LINUX OR QNX)
-  set(CMAKE_BUILD_FLAGS -- -j${N})
+
+if(NOT N EQUAL 0)
+    if(((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
```
elBoberido, line 22: I think there was a decision to drop ubuntu 16.04 support. I should create an issue to update `cmake_minimum_required` to `(VERSION 3.10)` then this could be simplified to `if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")`.

Which cmake version do we have on Debian 10, Raspbian, mac and Windows?

cmake/cyclonedds/CMakeLists.txt
```diff
@@ -17,10 +17,13 @@ project(cyclonedds-build CXX)
...
 include(ProcessorCount)
 ProcessorCount(N)
-if(NOT N EQUAL 0 AND ((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
-  set(CMAKE_BUILD_FLAGS -j${N})
-elseif(LINUX OR QNX)
-  set(CMAKE_BUILD_FLAGS -- -j${N})
+
+if(NOT N EQUAL 0)
+    if(((${CMAKE_VERSION} VERSION_GREATER "3.12.0") OR ${CMAKE_VERSION} VERSION_EQUAL "3.12.0"))
+        set(CMAKE_BUILD_FLAGS -j ${N})
```
ithier, line 23: The issue was the space between "j" and `${N}` ?
Strange...

.github/workflows/build-test.yml
```diff
@@ -89,9 +89,3 @@ jobs:
       - name: Build sources
         run: |
           cmake -Bbuild -Hiceoryx_meta -Dtest=ON && cmake --build build
-
-      - name: Run tests
```
ithier, line 93: Do we know what broke these ? They were working a few days ago.

elfenpiff: I have why they are broken but we do not support windows currently and therefore I do not want to be blocked by windows issues. When I continue my windows support implementation I have to take care of it but these are the problems of future elfenpiff

Issue #177, opened 2020-07-08
Add unit test for posh runtime


nihalchari:
increase the test coverage by writing unit tests for pos_runtime in https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp

PR #214, opened 2020-07-30
Iox #177 unit test for posh runtime

FerdinandSpitzschnueffler:
code contributed by @nihalchari


iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
+// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
elfenpiff, line 1: The issues #177 states that the posh_runtime should be tested but the tests here are incomplete. The following tests cases are at least missing:

- findService
   * finding offered service
   * not offered service
   * reoffer service
   * mix up of AnyService, AnyInstance, AnyEvent
- offerService
   * call with non invalid serviceDescription like `{AnyService, AnyInstance, AnyEvent}`
   * call with non existing serviceDescription
   * call offer service twice
   * call it successfully
- stopOfferService
   * call with non invalid serviceDescription like `{AnyService, AnyInstance, AnyEvent}`
   * call with non existing serviceDescription
   * call stop offer service twice
   * call it successfully

Could you please create the previous mentioned tests and maybe complete the list of unit tests since I am not sure if I have thought of all the cases.

nihalchari: I think the scope of this task was to add missing unit test(existing tests are at https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/test/moduletests/test_runtime.cpp)

Also i guess tests suggested by you comes under integration tests, which are already present at 
https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp

MatthiasKillat: I am not sure what the scope of the task is(up for discussion I guess), but some of those test cases make sense and are missing in the files provided by the links (reoffering, offer service twice for example). Also, I am not sure whether those tests should be classified unit- or integrationtests. They require just the posh runtime (or multiple runtimes) so could very well be unit tests. 

In addition, I guess the proposed test of calling offerService with a nonexisting service description means to call it with a service description that was not offered yet, in contrast to calling it twice with the same service description (just for clarification).

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    auto& sut = PoshRuntime::getInstance(appname);
+
+    EXPECT_EQ(sut.getInstanceName(), appname);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ReturnValue)
+{
+    uint32_t uniqueIdCounter = iox::popo::BasePortData::s_uniqueIdCounter;
```
MatthiasKillat, line 114: Has nothing to do with the test itself, but it is somewhat dangerous to have public access to this id counter.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    EXPECT_EQ(std::string("/sender"), applicationPortData->m_processName);
+    EXPECT_EQ(iox::capro::ServiceDescription(0u, 0u, 0u), applicationPortData->m_serviceDescription);
+    EXPECT_EQ(false, applicationPortData->m_toBeDestroyed);
+    EXPECT_EQ(uniqueIdCounter, applicationPortData->m_uniqueId);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ApplicationlistOverflow)
+{
+    m_errorHandlerCalled = false;
```
MatthiasKillat, line 127: Could be done in the setup code as well (but will then be done every time, so probably just do it when needed).

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    EXPECT_TRUE(m_errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, GetServiceRegistryChangeCounter_ReturnValue)
+{
+    auto runnableData = m_senderRuntime->getServiceRegistryChangeCounter();
+
+    // Roudi internally  calls 5 servieces before application registers its service
```
MatthiasKillat, line 285: This test is probably brittle and will break when internals change as it relies on implementation details. As far as I know, the only requirement for the counter is to monotonically change when services register/deregister, i.e. some registration status changes. Personally I think only this should be tested: whether the counter changes (by increasing) if something happens that is supposed to cause a change (as mentioned above).

MatthiasKillat: I see this is done below, in this case I think that this test can be removed entirely, we do not care for the initial value to be 5.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+        m_senderRuntime->getMiddlewareReceiver(iox::capro::ServiceDescription(i, i + 1u, i + 2u));
+    }
+
+    EXPECT_TRUE(m_errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, GetServiceRegistryChangeCounter_ReturnValue)
+{
+    auto runnableData = m_senderRuntime->getServiceRegistryChangeCounter();
```
MatthiasKillat, line 283: Variable name is misleading.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    // Roudi internally  calls 5 servieces before application registers its service
+    EXPECT_EQ(5u, *runnableData);
+}
+
+
+TEST_F(PoshRuntime_test, GetServiceRegistryChangeCounter_OfferStopOfferService)
+{
+    auto initial_value = m_senderRuntime->getServiceRegistryChangeCounter();
```
MatthiasKillat, line 292: Please no snake_case unless it is a library call (e.g. STL) or STL-like code.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    const uint32_t runnableDeviceIdentifier = 1u;
+    iox::runtime::RunnableProperty runnableProperty(iox::cxx::CString100("testRunnable"), runnableDeviceIdentifier);
+
+    auto runableData = m_senderRuntime->createRunnable(runnableProperty);
+
+    EXPECT_EQ(std::string("/sender"), runableData->m_process);
+    EXPECT_EQ(iox::cxx::CString100("testRunnable"), runableData->m_runnable);
+
+    /// @todo I am passing runnableDeviceIdentifier as 1, but it returns 0, is this expected?
+    // EXPECT_EQ(runnableDeviceIdentifier, runableData->m_runnableDeviceIdentifier);
```
MatthiasKillat, line 319: Probably not intended but currently unused, need to be kept in mind and revisited when we run on multiple devices.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
ithier: @elfenpiff what do you mean by "non invalid" ? valid ? :stuck_out_tongue:

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    iox::cxx::GenericRAII m_errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler(
+        [](const iox::Error, const std::function<void()>, const iox::ErrorLevel) { m_errorHandlerCalled = true; });
+
+    RouDiEnvironment m_roudiEnv{iox::RouDiConfig_t().setDefaults()};
+    PoshRuntime* m_receiverRuntime{&iox::runtime::PoshRuntime::getInstance("/receiver")};
+    PoshRuntime* m_senderRuntime{&iox::runtime::PoshRuntime::getInstance("/sender")};
+    MqMessage m_sendBuffer;
+    MqMessage m_receiveBuffer;
+    const iox::cxx::CString100 m_runnableName = iox::cxx::CString100("testRunnable");
+    const iox::cxx::CString100 m_invalidRunnableName = iox::cxx::CString100("invalidRunnable,");
```
ithier, line 49: Shouldn't we be using the fixed string API ? I believe `CString100` is deprecated.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    static bool m_errorHandlerCalled;
+};
+
+bool PoshRuntime_test::m_errorHandlerCalled{false};
+
+
+TEST_F(PoshRuntime_test, AppnameLength_TooLong)
+{
+    const std::string string104chars =
+        "/MXIYXHyPF9KjXAPv9ev9jxofYDArZzTvf8FF5uaWWC4dwabcjW75DurqeN645IabAsXVfngor7784446vb4vhArwBxLZlN1k1Qmrtz";
```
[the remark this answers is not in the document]
ithier: Also the name should by such that it will only fail based on length. There isn't really any value of using random characters.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    EXPECT_DEATH({ PoshRuntime::getInstance(string104chars); },
+                 "Application name has more than 100 characters, including null termination!");
+}
+
+
+TEST_F(PoshRuntime_test, AppnameLength_ok)
+{
+    const std::string string100chars =
```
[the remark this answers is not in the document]
ithier: Here you are testing app name length AND valid characters.
These should be their own tests.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+        "/MXIYXHyPF9KjXAPv9ev9jxofYDArZzTvf8FF5uaWWC4dwabcjW75DurqeN645IabAsXVfngor7784446vb4vhArwBxLZlN1k1";
+
+    EXPECT_NO_FATAL_FAILURE({ PoshRuntime::getInstance(string100chars); });
+}
+
+
+TEST_F(PoshRuntime_test, NoAppname)
```
ithier, line 75: "App Name" is two words. Why no camel case ?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    const std::string wrong = "wrongname";
+
+    EXPECT_DEATH({ PoshRuntime::getInstance(wrong); },
+                 "Cannot initialize runtime. Application name wrongname does not have the required leading slash '/'");
+}
+
+
+// test class creates instance of Poshruntime, so when getInstance() is called without name it reuturns extisting
+// instance
+TEST_F(PoshRuntime_test, DISABLED_AppnameEmpty)
```
ithier, line 95: Is this actually `DISABLED` ?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+    EXPECT_DEATH({ iox::runtime::PoshRuntime::getInstance(); },
+                 "Cannot initialize runtime. Application name has not been specified!");
+}
+
+
+TEST_F(PoshRuntime_test, GetInstanceName)
```
ithier, line 102: What are you testing here ? The test name should explain the test well enough so if it is seen in an error report we know immediately what is failing.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+    const std::string appname = "/app";
+
+    auto& sut = PoshRuntime::getInstance(appname);
+
+    EXPECT_EQ(sut.getInstanceName(), appname);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ReturnValue)
```
ithier, line 112: This is a pointless test ... What business logic is being tested here ?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    for (auto i = 0u; i < iox::MAX_PROCESS_NUMBER; ++i)
+    {
+        m_senderRuntime->getMiddlewareApplication();
+    }
+
+    EXPECT_TRUE(m_errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareInterface_ReturnValue)
```
ithier, line 138: Another pointless test... Or maybe I am missing something ?
@dkroenke @elfenpiff @MatthiasKillat @elBoberido

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    EXPECT_TRUE(m_errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, SendMessageToRouDi_ValidMessage)
+{
+    m_sendBuffer << mqMessageTypeToString(MqMessageType::IMPL_INTERFACE) << std::string("/sender")
+                 << static_cast<uint32_t>(iox::capro::Interfaces::INTERNAL) << m_runnableName;
+
+    const auto status = m_senderRuntime->sendMessageToRouDi(m_sendBuffer);
```
ithier, line 168: Nitpicking here, but naming this variable "success" instead of "status" makes the code nicer to read.
`EXPECT_EQ(true, success)` is immediately understandable.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: It should also not be necessary to repeat the type, just use `{"testRunnable"}` instead of `= iox::cxx::string...`

Also, don't we have a type alias for this? If not, maybe we should create one. This should be done in a separate PR, though.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: personally I don't find success much better than status. how about `delivered`, `sent` or `successfullySent`?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ReturnValue)
+{
+    uint32_t uniqueIdCounter = iox::popo::BasePortData::s_uniqueIdCounter;
+
+    const auto applicationPortData = m_senderRuntime->getMiddlewareApplication();
+
+    EXPECT_EQ(std::string("/sender"), applicationPortData->m_processName);
+    EXPECT_EQ(iox::capro::ServiceDescription(0u, 0u, 0u), applicationPortData->m_serviceDescription);
```
elBoberido, line 119: I'm not sure if this checks should be done here. That's something RouDi creates for us, the runtime just passes the pointer the the caller. I don't know if we can check it directly in RouDi, though.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: If it's done this way, we should always check that the error handler was not called in all the testcases where we do not expect it to be called.
By changing the error handler for all tests, this is necessary since a test might have triggered the error handler but since it doesn't do anything except setting the flag, a situation which might lead to a termination in real word usage might result in a passed test by accident.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: it's not pointless, the name could be a little more descriptive, like GetMittlewareApplicationIsSuccessful but it just test if we get an application port

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    EXPECT_EQ(false, applicationPortData->m_toBeDestroyed);
+    EXPECT_EQ(uniqueIdCounter, applicationPortData->m_uniqueId);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ApplicationlistOverflow)
+{
+    m_errorHandlerCalled = false;
+
+    for (auto i = 0u; i < iox::MAX_PROCESS_NUMBER; ++i)
```
elBoberido, line 129: I'm not sure why this calls the error handler. I would expect to happen with `i <= iox::MAX_PROCESS_NUMBER`.

Could you check what happens with `i < iox::MAX_PROCESS_NUMBER - 1`?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    {
+        m_senderRuntime->getMiddlewareApplication();
```
elBoberido, line 131: the return value should be checked for not being a nullptr for the number of allowed ports. So this test needs to be split into a for loop which requests ports up to the limit and another request outside of the loop which triggers the error handler.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: it's the same like with the application port test

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+    EXPECT_EQ(true, status);
+}
+
+
+TEST_F(PoshRuntime_test, SendMessageToRouDi_InvalidMessage)
+{
+    m_sendBuffer << mqMessageTypeToString(MqMessageType::IMPL_INTERFACE) << std::string()
+                 << static_cast<uint32_t>(iox::capro::Interfaces::INTERNAL) << m_invalidRunnableName;
+
+    const auto status = m_senderRuntime->sendMessageToRouDi(m_sendBuffer);
```
elBoberido, line 179: could you check if `PoshRuntime::sendMessageToRouDi` is used somewhere? My gut feelings tell me we could remove this function from the runtime

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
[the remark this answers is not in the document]
nihalchari: done

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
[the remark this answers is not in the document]
nihalchari: done

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: changed to counter

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: updated

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: Is there a chance this might change? If not, what's the point of a permanently disabled test?
At least comment it!

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: We should make use of more descriptive names like "...IsSuccessful", "...ResultsInXY" or "...ResultsInFailure".

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: +1 for `successfullySent` or `sentSuccessfully`

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: btw. this test could actually be split into two, the first one would check if we can get `MAX_PROCESS_NUMBER` of application ports and the second would check if the error handler is called if we request more.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,319 @@
...
+        m_senderRuntime->getMiddlewareSender(iox::capro::ServiceDescription(i, i + 1u, i + 2u));
+    }
+
+    EXPECT_TRUE(m_errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareReceiver_ReturnValue)
+{
+    auto receiverPort = m_receiverRuntime->getMiddlewareReceiver(
```
elBoberido, line 258: you do not need an additional runtime to get the receiver port. You can take the same you use to get a sender port and maybe call it just `m_runtime`

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: now I know. There are two runtimes created in the test fixture, the m_receiverRuntime and m_senderRuntime and the runtime implicitly creates an application port.
I think we do not need two runtimes but just one and then the loop can start at `1` instead of `0`, maybe with a comment that there is already an active runtime which acquired an application port.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,319 @@
...
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ApplicationlistOverflow)
+{
+    m_errorHandlerCalled = false;
+
+    for (auto i = 0u; i < iox::MAX_PROCESS_NUMBER; ++i)
+    {
+        m_senderRuntime->getMiddlewareApplication();
+    }
+
+    EXPECT_TRUE(m_errorHandlerCalled);
```
elBoberido, line 142: instead of just checking if the error handler is called, we should check if it's only once called and also if it was called with the expected error. So it might make sense to instantiate the `TemporaryErrorHandler` in the test case itself instead of the test fixture

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: fixed

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,320 @@
...
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplication_ReturnValue)
+{
+    uint32_t uniqueIdCounter = iox::popo::BasePortData::s_uniqueIdCounter;
+
+    const auto applicationPortData = m_senderRuntime->getMiddlewareApplication();
+
+    EXPECT_EQ(std::string("/sender"), applicationPortData->m_processName);
```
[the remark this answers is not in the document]
nihalchari: fixed

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: fixed

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: done

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: yeah, its not used

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: done

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elBoberido: could you then please remove it from PoshRuntime?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,339 @@
...
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareApplicationIsSuccessful)
+{
+    uint32_t uniqueIdCounter = iox::popo::BasePortData::s_uniqueIdCounter;
+
+    const auto applicationPortData = m_runtime->getMiddlewareApplication();
+
+    ASSERT_THAT(m_runtimeName, applicationPortData->m_processName);
```
elBoberido, line 126: this is not what I meant. You need to check that `applicationPortData` is not a `nullptr` before you access `m_processName`

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,339 @@
...
+    m_runtime->offerService({"service1", "instance1"});
+    this->InterOpWait();
+    auto counter = m_runtime->getServiceRegistryChangeCounter();
+
+    EXPECT_EQ(6u, *counter);
+
+    m_runtime->stopOfferService({"service1", "instance1"});
+    this->InterOpWait();
+
+    EXPECT_EQ(7u, *counter);
```
MatthiasKillat, line 323: Only check whether the counter has increased compared to the previous value (which needs to be stored for this purpose).

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
MatthiasKillat: Will not be changed in the test scope.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: @elfenpiff  created issue for remaining tests https://github.com/eclipse/iceoryx/issues/240

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+
+TEST_F(PoshRuntime_test, SendRequestToRouDi_ValidMessage)
+{
+    m_sendBuffer << mqMessageTypeToString(MqMessageType::IMPL_INTERFACE) << m_runtimeName
+                 << static_cast<uint32_t>(iox::capro::Interfaces::INTERNAL) << m_runnableName;
+
+    const auto successfullySent = m_runtime->sendRequestToRouDi(m_sendBuffer, m_receiveBuffer);
+
+    EXPECT_TRUE(m_receiveBuffer.isValid());
```
MatthiasKillat, line 223: Ok, should be fine to check that the general request response mechanism works.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    EXPECT_TRUE(errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, GetServiceRegistryChangeCounter_OfferStopOfferService)
+{
+    auto initialValue = m_runtime->getServiceRegistryChangeCounter();
+    auto cout = initialValue->load();
```
MatthiasKillat, line 316: strange variable name, meant count? Why not initialCount?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+    m_runtime->offerService({"service1", "instance1"});
+    this->InterOpWait();
+    auto counter = m_runtime->getServiceRegistryChangeCounter();
```
MatthiasKillat, line 320: Logically this is correct (I still would have liked EXPECT_GT more...but ok). We need to try to find variable names that convey the meaning of what is happening. I.e. initialCount. It is about comprehensibility for other developers:-)

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
MatthiasKillat: This nullptr check is still missing.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: using EXPECT_GT after offerService()  is fine, but using EXPECT_GT again after stopOfferService() may not fit, because value is already incremented for 1st time(during offerService )
It will be like performing EXPECT_GT on value that we have already confirmed to be greater

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: return value of getServiceRegistryChangeCounter() is atomic and updates when there is change in service, so i renamed initialValue to serviceCounter, and count to initialCount

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,342 @@
...
+    EXPECT_EQ(false, successfullySent);
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareSenderIsSuccessful)
+{
+    const auto senderPort = m_runtime->getMiddlewareSender(
+        iox::capro::ServiceDescription(99u, 1u, 20u), m_runnableName, iox::runtime::PortConfigInfo(11u, 22u, 33u));
+
+    EXPECT_EQ(iox::capro::ServiceDescription(99u, 1u, 20u), senderPort->m_serviceDescription);
```
elBoberido, line 244: please add an `ASSERT_THAT(senderPort, Ne(nullptr))` before you access the members of the senderPort. This could be a nullptr and then the test would crash.

Please also do it everywhere you call getMiddlewareX and then de-reference the returned pointer

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,342 @@
...
+            errorHandlerCalled = true;
+        });
+
+    // i=1 because there is already an active runtime in test fixture class which acquired an application port
+    for (auto i = 1u; i < iox::MAX_PROCESS_NUMBER; ++i)
+    {
+        auto appPort = m_runtime->getMiddlewareApplication();
+        EXPECT_NE(nullptr, appPort);
+    }
+    auto appPort = m_runtime->getMiddlewareApplication();
```
elBoberido, line 148: please add a check that the error handler was not called before you force the overflow. With the current checks it is possible that the error handler is already called with any invocation of getMiddlewareApplication and we wouldn't notice it

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,342 @@
...
+{
+    auto errorHandlerCalled{false};
+    auto errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler(
+        [&errorHandlerCalled](const iox::Error, const std::function<void()>, const iox::ErrorLevel) {
+            errorHandlerCalled = true;
+        });
+
+    for (uint32_t i = 0u; i < iox::MAX_PORT_NUMBER; ++i)
+    {
+        m_runtime->getMiddlewareSender(iox::capro::ServiceDescription(i, i + 1u, i + 2u));
```
elBoberido, line 269: please add to all the overflow tests the same checks like in the GetMiddlewareApplication_ApplicationlistOverflow test

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,339 @@
...
+    EXPECT_EQ(true, successfullySent);
+}
+
+
+TEST_F(PoshRuntime_test, SendRequestToRouDi_ValidMessage)
+{
+    m_sendBuffer << mqMessageTypeToString(MqMessageType::IMPL_INTERFACE) << m_runtimeName
+                 << static_cast<uint32_t>(iox::capro::Interfaces::INTERNAL) << m_runnableName;
+
+    const auto successfullySent = m_runtime->sendRequestToRouDi(m_sendBuffer, m_receiveBuffer);
```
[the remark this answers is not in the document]
elBoberido: the `getMiddlewareX` methods are using this method to get the ports, so basically if those tests work, this will also work. I think the only reason this method is public is because we have users who might extend RouDi and send custom messages

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: added

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
nihalchari: done

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,342 @@
...
+        EXPECT_NE(nullptr, interfacePort);
+    }
+    auto interfacePort = m_runtime->getMiddlewareInterface(iox::capro::Interfaces::INTERNAL);
+
+    EXPECT_EQ(nullptr, interfacePort);
+    EXPECT_TRUE(errorHandlerCalled);
+}
+
+
+TEST_F(PoshRuntime_test, SendMessageToRouDi_ValidMessage)
```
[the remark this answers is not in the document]
dkroenke: sendMessageToRouDi is removed.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,359 @@
...
+}
+
+
+TEST_F(PoshRuntime_test, GetMiddlewareSenderSenderlistOverflow)
+{
+    auto senderlistOverflowDetected{false};
+
+    auto errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler(
+        [&senderlistOverflowDetected](const iox::Error error, const std::function<void()>, const iox::ErrorLevel) {
+            if (error == iox::Error::kPORT_POOL__SENDERLIST_OVERFLOW)
```
MatthiasKillat, line 256: Ok, note that this was necessary because the error handler is called twice but with another error the second time (we are only interested in the overflow error here).

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,359 @@
...
+            {
+                senderlistOverflowDetected = true;
+            }
+        });
+
+    ///@note 5 sender ports are alloted for internal services of Roudi.
+    /// hence getServiceRegistryChangeCounter() is used
+    auto serviceCounter = m_runtime->getServiceRegistryChangeCounter();
+    auto availableSenderPort = serviceCounter->load();
```
MatthiasKillat, line 265: It is a bit unfortunate to derive the number of used ports from this. This is an API limitation of the runtime. Maybe the runtime could be extended with way to query this information.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,359 @@
...
+{
+    std::string appName("/valid_name");
+
+    EXPECT_NO_FATAL_FAILURE({ PoshRuntime::getInstance(appName); });
+}
+
+
+TEST_F(PoshRuntime_test, AppNameLength_OutOfLimit)
+{
+    std::string tooLongName(100, 's');
```
elfenpiff, line 74: Please get rid of the magic number `100` here. The Maximum app name length is defined in `iceoryx_posh_types.hpp` in the variable `MAX_PROCESS_NAME_LENGTH`

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -0,0 +1,359 @@
...
+    tooLongName.insert(0, 1, '/');
+
+    EXPECT_DEATH({ PoshRuntime::getInstance(tooLongName); },
+                 "Application name has more than 100 characters, including null termination!");
+}
+
+
+TEST_F(PoshRuntime_test, MaxAppNameLength)
+{
+    std::string maxValidName(99, 's');
```
elfenpiff, line 84: No magic numbers. Please use `MAX_PROCESS_NAME_LENGTH - 1`

PR #215, opened 2020-07-30
Iox #186 Add cxx::list implementation to utils

### LINKED ISSUE #186: Add cxx::list to utils
## Brief feature description
As mentioned in [CONTRIBUTING.md](https://github.com/eclipse/iceoryx/blob/master/CONTRIBUTING.md) non-heap, shared memory supporting library functions are needed for the project targets. In the same way as already a cxx::vector implementation can be found, a cxx::list (or possibly cxx::forward_list) should be available and replace std::list use cases in iceoryx.

Please describe in a few sentences what should be implemented
The new list implementation shall be able to replace a list (or forward_list) implementation currently used in iceoryx project (like for RouDi process management). 
The interface shall be held most compatible to the standard library interface, however - as with the cxx::vector implementation - certain restrictions might apply (fixed max_size at compile time, non-returning functions become returning a success flag on it's operations).

## Detailed information
Pros and cons about a list (bi-directional) vs a forward_list implementation
- operator--, ~back functions (like emplace_back, pop_back) only available in bi-directional list
- forward_list has less memory (~half) overhead
- forward_list could be more efficient in processing speed (less operations) and safety accreditation (less review and tests)

The cxx list library should only use stack memory and be limited to a max_size (->no data races). 
It should not use c-pointers to reference between list elements (an element indexing-based management comes to mind). This will ensure safe moving of the list through shared memory from different processes. Concurrency-safety however should be ensured by further (available) utils on list user side.

At the time of realizing this first implementation the 'error handling strategy' #130 is not available to align the implementation of cxx::list error handling (usage of return values or calling cerr or terminate). This has to followed up by the mentioned issue.

bishibashiB:
As a follow-up to PR #189 (cxx::forward_list) here is also a 'list' implementation allowing more freedom of choice for the developer in switching from other list implementations like std::list. (No refactoring of source code needed like in case of switching from std::list to cxx::forward_list)
Fixes #186


iceoryx_utils/include/iceoryx_utils/internal/cxx/list.inl
```diff
@@ -0,0 +1,713 @@
...
+    }
+}
+
+// iterator validition on iterator operations or iterators are function arguments
+// not coherent-save, as not each single iterator dereferencing operation is checked for validity.
+template <typename T, uint64_t Capacity>
+inline bool list<T, Capacity>::handleInvalidIterator(const const_iterator& iter) const noexcept
+{
+    // freeList / invalid elements will have the prevIdx set to INVALID_INDEX
+    if ((getPrevIdx(iter) < INVALID_INDEX) && (getNextIdx(iter) < INVALID_INDEX)
```
MatthiasKillat, line 669: Works, due to the specific value of INVALID_INDEX being Capacity+1. Note that we cannot and should not check for inequality here as it is implemented.

Changing this would lead to logic errors but this is ok for intricate data structures (implementation details are somewhat rigid, i.e. hard to change, but interface is not).

One thing that does not necessarily needs to be changed is separating the valid check from the handling, i.e. having a method such as isValid(iterator) and a handleInvalidIterator method. As they are always used together right now this is not needed. Should be done when it becomes apparent that valid checks are used on their own.

iceoryx_utils/test/moduletests/test_cxx_list.cpp
```diff
@@ -0,0 +1,2283 @@
...
+    ASSERT_THAT(cTor, Eq(TESTLISTCAPACITY));
+    ASSERT_THAT(dTor, Eq(TESTLISTCAPACITY));
+}
+
+TEST_F(list_test, PopFrontFullPlusOneToEmptyList)
+{
+    for (uint64_t i = 0; i < TESTLISTCAPACITY; ++i)
+    {
+        sut.emplace(sut.cbegin());
+        EXPECT_THAT(sut.size(), Eq((i + 1) > TESTLISTCAPACITY ? TESTLISTCAPACITY : (i + 1)));
```
MatthiasKillat, line 974: Does not matter much here but normally it is better to add an extra statement after the loop for the final emplace to eliminate the branch in the loop which is only needed in the final iteration.

Does not have to be changed though. Just an observation which is important for performance  and arguably readability. (The reader may think on first sight the condition could be true every iteration if it would be more complex.)

iceoryx_utils/test/moduletests/test_cxx_list.cpp
```diff
@@ -0,0 +1,2283 @@
...
+
+    ASSERT_THAT(sut.size(), Eq(2u));
+    ASSERT_THAT(cTor, Eq(1u));
+    ASSERT_THAT(customCTor, Eq(1u));
+    auto iter = sut.begin();
+    EXPECT_THAT((*iter).m_value, Eq(DEFAULT_VALUE));
+    EXPECT_THAT((++iter)->m_value, Eq(TEST_LIST_ELEMENT_DEFAULT_VALUE));
+}
+
+TEST_F(list_test, InsertBeforeBeginListRValue)
```
MatthiasKillat, line 1136: I am thinking about how one could have reused much the code for test cases when the difference is just an lvalue or an rvalue. Probably with some kind of policy and typed tests, where the policy type is varied and provides a generation interface the allows us to obtain rvalues or lvalues (since returned values are always rvalues there would be some tricks needed which seem ugly).

Summary: a strategy write tests in a more generic way suitable for for lvalues and  rvalues would be nice but it is unclear how this could be done elegantly. Food for thought for the future. 
No change needed, of course.

iceoryx_utils/test/moduletests/test_cxx_list.cpp
```diff
@@ -0,0 +1,2283 @@
...
+    auto iter2 [[gnu::unused]] = sut.erase(iter);
+
+    EXPECT_DEATH(dummyFunc(iter->m_value == 12), "");
+}
+
+TEST_F(list_test, ListIsCopyableViaMemcpy)
+{
+    uint64_t i = 0;
+    using TestFwdList = list<TestListElement, TESTLISTCAPACITY>;
+    alignas(alignof(TestFwdList)) uint8_t* otherSutPtr = new uint8_t[sizeof(TestFwdList)];
```
MatthiasKillat, line 2254: I am not sure this does what you expect. I think it will align the pointer itself, not the pointee (!). The memory returned by new will not be aligned. This will fail in weird ways when not handled correctly. One way is to allocate more memory (size + align) via new or malloc and align it afterwards with std::align. But also, otherSut could be allocated on the stack as a ```alignas(alignof(TestFwdList)) uint8_t buffer[sizeof(...)];``` Then take a pointer from this and use it as before, just no delete at the end but possibly a dtor call since this is somewhat like a placement new.

MatthiasKillat (changes requested):

iceoryx_utils/test/moduletests/test_cxx_forward_list.cpp
```diff
@@ -398,7 +400,7 @@ TEST_F(forward_list_test, EmplaceAfterWithOneElements)
     constexpr uint64_t CAPACITY{42u};
```
dkroenke, line 400: Please use U in uppercase letters

iceoryx_utils/test/moduletests/test_cxx_list.cpp
```diff
@@ -0,0 +1,2283 @@
...
+
+namespace
+{
+static constexpr uint64_t TESTLISTCAPACITY{10u};
+static constexpr int64_t TEST_LIST_ELEMENT_DEFAULT_VALUE{-99};
+
+class list_test : public Test
+{
+  public:
+    static int64_t cTor;
```
dkroenke, line 30: Please rework to uint usage

dkroenke (changes requested):

iceoryx_utils/test/moduletests/test_cxx_list.cpp
bishibashiB: Background: this implementation code stems from a parametrized test case (w/ list filling as parameter). 
The test implementation being however does not use parametrized tests (yet).

iceoryx_utils/test/moduletests/test_cxx_list.cpp
bishibashiB: Thx bringing this up. Addressed~!

iceoryx_utils/include/iceoryx_utils/internal/cxx/list.inl
```diff
@@ -0,0 +1,713 @@
...
+// init
+template <typename T, uint64_t Capacity>
+void list<T, Capacity>::init() noexcept
+{
+    // all list elements are concatenated and accessed via the 'freeListHead'
+    // used-list is empty, element-links for used-list are stubbed by BEGIN_END_LINK_INDEX
+    setPrevIdx(0U, INVALID_INDEX);
+    setNextIdx(0U, 1U);
+
+    for (size_type i = 1; (i + 1U) < BEGIN_END_LINK_INDEX; ++i)
```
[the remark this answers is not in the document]
bishibashiB: addressed in 8b1033c72d943f7e50c00a42ea452943c584850a

MatthiasKillat (approved):

iceoryx_utils/include/iceoryx_utils/cxx/list.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+    size_type m_size{0U};
+}; // list
+
+} // namespace cxx
+} // namespace iox
+
+#include "iceoryx_utils/internal/cxx/list.inl"
+
+#endif // IOX_UTILS_CXX_LIST_HPP
```
dkroenke, line 390: please add empty endline

iceoryx_utils/include/iceoryx_utils/cxx/list.hpp
```diff
@@ -0,0 +1,390 @@
...
+    /// @param[in] beforeToBeErasedIter iterator linking the element before the to-be-removed element
+    /// @return an (non-const_) iterator to the element after the removed element,
+    ///         returns end() element when reached end of list
+    iterator erase(const_iterator beforeToBeErasedIter) noexcept;
+
+    /// @brief remove the first element which matches the given comparing element (compare by value)
+    ///         requires a the template type T to have operator== defined.
+    /// @param[in] data value to compare to
+    /// @return the number of elements removed, return is C++20-conform
+    size_type remove(const T& data) noexcept;
```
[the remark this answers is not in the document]
bishibashiB: According to [std::list (as per C++20](https://en.cppreference.com/w/cpp/container/list/remove)) this function will return the **count** of removes (each element comparing-equal (aka. T::operator==) to argument `data` will be removed). 
For singular instance removal `erase(iterator)` has to be utilized.

iceoryx_utils/include/iceoryx_utils/cxx/list.hpp
```diff
@@ -0,0 +1,390 @@
...
+    bool pop_back() noexcept;
+
+    /// @brief remove all elements from the list, list will be empty
+    ///         element destructors will be invoked
+    void clear() noexcept;
+
+    /// @brief remove next element from linked iterator position
+    ///         element destructors will be invoked
+    ///         recursive calls to erase_after only delete each 2nd element
+    /// @param[in] beforeToBeErasedIter iterator linking the element before the to-be-removed element
```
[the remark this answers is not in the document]
bishibashiB: Comments are copy&paste errors, thx! 
Has been addressed!

dkroenke (approved):

dkroenke (approved):

MatthiasKillat (approved):

PR #224, opened 2020-08-07
Iox #221 apply cxx list

### LINKED ISSUE #221: Replace occurence of std::list by cxx::list
## Brief feature description
After finishing cxx::list and cxx::forward_list implementation with #186 the replacement of std::list is consequently the next step. 

## Detailed information
After replacement of std::list additional replacements might make (1)sense performance-wise or (2)for the sake of a tidy code architecture. 
However, for these 'additional replacements' a discussion / agreement should take place before any code rework.
Feel free to put your opinions and hints here.
E.g. so far I can think of discussing 

- RouDi (a)processes and (b)runnables mangement, in app and (c)in introspection => review and establish a picture where to utilize lists, where vectors
- Could a cxx::list replace the UsedChunkList? Would that justify a rework of code (different member function signature), would concurrency requirements not allow to do so?
- PortPool 'lists' which use cxx::vector aka Fixed**Position**Container (sender/receiver/interface/applicationPortDataList, runnableDataList)
- PortIntrospection Fixed**Size**Container

bishibashiB:
Within the attached source std::list usage was removed from roudi process list handling (replaced with cxx::list implementation).
Additionally the FixedPositionContainer of roudi port pool was picked to replace with cxx::list as well... see details in #221. (fixes #221)

This PR requires cxx::list to be approved (merged) #215 first.

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/process_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/process_introspection.hpp
@@ -17,9 +17,9 @@
 #include "iceoryx_posh/iceoryx_posh_types.hpp"
 #include "iceoryx_posh/internal/popo/sender_port.hpp"
 #include "iceoryx_posh/roudi/introspection_types.hpp"
+#include "iceoryx_utils/cxx/list.hpp"
 
 #include <atomic>
-#include <list>
 #include <mutex>
 #include <thread>
 
@@ -111,9 +111,7 @@ class ProcessIntrospection
     void setSendInterval(unsigned int interval_ms);
 
   private:
-    /// @todo use a fixed, stack based list once available
-    // using ProcessList_t = cxx::list<ProcessIntrospectionData, MAX_PROCESS_NUMBER>;
-    using ProcessList_t = std::list<ProcessIntrospectionData>;
+    using ProcessList_t = cxx::list<ProcessIntrospectionData, MAX_PROCESS_NUMBER>;
     ProcessList_t m_processList;
     bool m_processListNewData{true}; // true because we want to have a valid field, even with an empty list
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
@@ -23,12 +23,12 @@
 #include "iceoryx_posh/mepoo/chunk_header.hpp"
 #include "iceoryx_posh/version/compatibility_check_level.hpp"
 #include "iceoryx_posh/version/version_info.hpp"
+#include "iceoryx_utils/cxx/list.hpp"
 #include "iceoryx_utils/posix_wrapper/posix_access_rights.hpp"
 
 #include <csignal>
 #include <cstdint>
 #include <ctime>
-#include <list>
 
 namespace iox
 {
@@ -100,9 +100,7 @@ class ProcessManagerInterface
 class ProcessManager : public ProcessManagerInterface
 {
   public:
-    /// @todo use a fixed, stack based list once available
-    // using ProcessList_t = cxx::list<RouDiProcess, MAX_PROCESS_NUMBER>;
-    using ProcessList_t = std::list<RouDiProcess>;
+    using ProcessList_t = cxx::list<RouDiProcess, MAX_PROCESS_NUMBER>;
     using PortConfigInfo = iox::runtime::PortConfigInfo;
 
     ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,


iceoryx_posh/include/iceoryx_posh/internal/roudi/iceoryx_port_pool.hpp
```diff
@@ -28,8 +28,8 @@ class IceOryxPortPool : public PortPool
   public:
     IceOryxPortPool(PortPoolData& portPool) noexcept;
 
-    cxx::vector<SenderPortType::MemberType_t*, MAX_PORT_NUMBER> senderPortDataList() noexcept override;
-    cxx::vector<ReceiverPortType::MemberType_t*, MAX_PORT_NUMBER> receiverPortDataList() noexcept override;
+    cxx::list<SenderPortType::MemberType_t, MAX_PORT_NUMBER>& senderPortDataList() noexcept override;
```
elfenpiff, line 31: Uh this is dangerous. What if someone takes the list and modifies it arbitrarily? Is this what we want? If yes, then we can just make the whole list public, otherwise I would return here a copy.

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
```diff
@@ -47,11 +47,11 @@ class PortPool
     /// @todo don't create the vector with each call but only when the data really change
     /// there could be a member "cxx::vector<popo::SenderPortData* m_senderPorts;" and senderPorts() would just update
     /// this member if the sender ports actually changed
-    virtual cxx::vector<SenderPortType::MemberType_t*, MAX_PORT_NUMBER> senderPortDataList() noexcept = 0;
-    virtual cxx::vector<ReceiverPortType::MemberType_t*, MAX_PORT_NUMBER> receiverPortDataList() noexcept = 0;
-    cxx::vector<popo::InterfacePortData*, MAX_INTERFACE_NUMBER> interfacePortDataList() noexcept;
-    cxx::vector<popo::ApplicationPortData*, MAX_PROCESS_NUMBER> appliactionPortDataList() noexcept;
-    cxx::vector<runtime::RunnableData*, MAX_RUNNABLE_NUMBER> runnableDataList() noexcept;
+    virtual cxx::list<SenderPortType::MemberType_t, MAX_PORT_NUMBER>& senderPortDataList() noexcept = 0;
```
elfenpiff, line 50: Again, it is dangerous to return the reference to internal members. Then everyone can modify, delete and add anything they like - even invalid data. If this is what we want make the list public otherwise we have to return a copy here.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -98,9 +98,13 @@ void PortManager::doDiscovery()
...
     // get the changes of sender port offer state
-    for (auto senderPortData : m_portPool->senderPortDataList())
+
+    auto& senderList = m_portPool->senderPortDataList();
+    for (auto iter = senderList.begin(); iter != senderList.cend();)
     {
-        SenderPortType senderPort(senderPortData);
+        auto checkIter = iter;
+        ++iter;
+        SenderPortType senderPort(checkIter.operator->());
```
elfenpiff, line 107: Could you please write `(*checkIter)` instead of using the complicated `operator->()` method. I think it was not intended to be used like this.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -98,9 +98,13 @@ void PortManager::doDiscovery()
 void PortManager::handleSenderPorts()
 {
     // get the changes of sender port offer state
-    for (auto senderPortData : m_portPool->senderPortDataList())
+
+    auto& senderList = m_portPool->senderPortDataList();
+    for (auto iter = senderList.begin(); iter != senderList.cend();)
     {
-        SenderPortType senderPort(senderPortData);
+        auto checkIter = iter;
```
elfenpiff, line 105: I think it would be better to write
```
SenderPortData * dataPtr = *iter;
```
Then we have better readable code in the whole function

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -134,17 +138,20 @@ void PortManager::handleSenderPorts()
...
 
 void PortManager::handleReceiverPorts()
 {
     // get requests for change of subscription state of receivers
-    for (auto receiverPortData : m_portPool->receiverPortDataList())
+    auto& receiverList = m_portPool->receiverPortDataList();
+    for (auto iter = receiverList.begin(); iter != receiverList.cend();)
     {
-        ReceiverPortType receiverPort(receiverPortData);
+        auto checkIter = iter;
```
elfenpiff, line 152: Please write
```
ReceiverPortData * dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -163,29 +170,31 @@ void PortManager::handleReceiverPorts()
...
-    cxx::vector<popo::InterfacePortData*, MAX_INTERFACE_NUMBER> interfacePortsForInitialForwarding;
-
+    cxx::list<popo::InterfacePortData*, MAX_INTERFACE_NUMBER> interfacePortsForInitialForwarding;
 
-    for (auto interfacePortData : m_portPool->interfacePortDataList())
+    auto& interfaceList = m_portPool->interfacePortDataList();
+    for (auto iter = interfaceList.begin(); iter != interfaceList.cend();)
     {
-        if (interfacePortData->m_doInitialOfferForward)
+        auto checkIter = iter;
```
elfenpiff, line 186: Please write
```
InterfacePortData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -246,9 +255,12 @@ void PortManager::handleApplications()
 {
     capro::CaproMessage caproMessage;
 
-    for (auto applicationPortData : m_portPool->appliactionPortDataList())
+    auto& appList = m_portPool->appliactionPortDataList();
+    for (auto iter = appList.begin(); iter != appList.cend();)
     {
-        iox::popo::ApplicationPort applicationPort(applicationPortData);
+        auto checkIter = iter;
```
elfenpiff, line 261: Please write
```
ApplicationPortData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -296,11 +308,15 @@ void PortManager::handleRunnables()
     // m_processIntrospection->removeRunnable(cxx::CString100(process.c_str()),
     // cxx::CString100(runnable.c_str()));
 
-    for (auto runnableData : m_portPool->runnableDataList())
+    auto& runnableList = m_portPool->runnableDataList();
+    for (auto iter = runnableList.begin(); iter != runnableList.cend();)
     {
-        if (runnableData->m_toBeDestroyed)
+        auto checkIter = iter;
```
elfenpiff, line 314: Please write
```
RunnableData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -400,57 +416,72 @@ bool PortManager::areAllReceiverPortsSubscribed(std::string appName)
 
 void PortManager::deletePortsOfProcess(std::string processName)
 {
-    for (auto port : m_portPool->senderPortDataList())
+    auto& senderList = m_portPool->senderPortDataList();
+    for (auto iter = senderList.begin(); iter != senderList.cend();)
     {
-        SenderPortType sender(port);
+        auto checkIter = iter;
```
elfenpiff, line 422: Please write
```
SenderPortData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -400,57 +416,72 @@ bool PortManager::areAllReceiverPortsSubscribed(std::string appName)
...
+            destroySenderPort(*checkIter);
         }
     }
 
-    for (auto port : m_portPool->receiverPortDataList())
+    auto& receiverList = m_portPool->receiverPortDataList();
+    for (auto iter = receiverList.begin(); iter != receiverList.cend();)
     {
-        ReceiverPortType receiver(port);
+        auto checkIter = iter;
```
elfenpiff, line 434: Please write
```
ReceiverPortData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -400,57 +416,72 @@ bool PortManager::areAllReceiverPortsSubscribed(std::string appName)
...
+            destroyReceiverPort(*checkIter);
         }
     }
 
-    for (auto port : m_portPool->interfacePortDataList())
+    auto& interfaceList = m_portPool->interfacePortDataList();
+    for (auto iter = interfaceList.begin(); iter != interfaceList.cend();)
     {
-        popo::InterfacePort interface(port);
+        auto checkIter = iter;
```
elfenpiff, line 446: Please write
```
InterfacePortData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -400,57 +416,72 @@ bool PortManager::areAllReceiverPortsSubscribed(std::string appName)
...
             LogDebug() << "Deleted Interface of application " << processName;
         }
     }
 
-    for (auto port : m_portPool->appliactionPortDataList())
+    auto& appList = m_portPool->appliactionPortDataList();
+    for (auto iter = appList.begin(); iter != appList.cend();)
     {
-        popo::ApplicationPort application(port);
+        auto checkIter = iter;
```
elfenpiff, line 459: Please write
```
ApplicationPortData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -400,57 +416,72 @@ bool PortManager::areAllReceiverPortsSubscribed(std::string appName)
...
             LogDebug() << "Deleted ApplicationPort of application " << processName;
         }
     }
 
-    for (auto runnableData : m_portPool->runnableDataList())
+    auto& runnableList = m_portPool->runnableDataList();
+    for (auto iter = runnableList.begin(); iter != runnableList.cend();)
     {
-        if (processName == runnableData->m_process)
+        auto checkIter = iter;
```
elfenpiff, line 472: Please write
```
RunnableData *dataPtr = *iter;
```

iceoryx_posh/source/roudi/port_pool.cpp
```diff
@@ -25,27 +25,27 @@ PortPool::PortPool(PortPoolDataBase& portPoolDataBase) noexcept
...
 
 cxx::expected<popo::InterfacePortData*, PortPoolError>
 PortPool::addInterfacePort(const std::string& applicationName, const capro::Interfaces interface) noexcept
 {
-    if (m_portPoolDataBase->m_interfacePortMembers.hasFreeSpace())
+    if (!m_portPoolDataBase->m_interfacePortMembers.full())
     {
-        auto interfacePortData = m_portPoolDataBase->m_interfacePortMembers.insert(
+        auto interfacePortData = &m_portPoolDataBase->m_interfacePortMembers.emplace_front(
             iox::cxx::string<100>(iox::cxx::TruncateToCapacity, applicationName), interface);
```
elfenpiff, line 49: No magic numbers. Why `string<100>`?
You could use the alias `using ProcessName_t = cxx::string<100>;` which can be found in `iceoryx_posh_types.hpp`

I know you did not write this particular line of code but the boyscout principle applies and you should always leave the code a little bit cleaner then you found it. 🦘

iceoryx_posh/source/roudi/port_pool.cpp
```diff
@@ -91,17 +92,19 @@ PortPool::addRunnableData(const iox::cxx::CString100& process,
...
-    m_portPoolDataBase->m_interfacePortMembers.erase(portData);
+    m_portPoolDataBase->m_interfacePortMembers.remove_if(
+        [&](popo::InterfacePortData& port) { return &port == portData; });
 }
 
 void PortPool::removeApplicationPort(popo::ApplicationPortData* const portData) noexcept
 {
-    m_portPoolDataBase->m_applicationPortMembers.erase(portData);
+    m_portPoolDataBase->m_applicationPortMembers.remove_if(
+        [&](popo::ApplicationPortData& port) { return &port == portData; });
```
elfenpiff, line 102: Please do not capture everything. In this case we only require to capture the `portData` pointer so please change the lambda to `[portData](...`

iceoryx_posh/source/roudi/port_pool.cpp
```diff
@@ -91,17 +92,19 @@ PortPool::addRunnableData(const iox::cxx::CString100& process,
...
 }
 
 void PortPool::removeRunnableData(runtime::RunnableData* const runnableData) noexcept
 {
-    m_portPoolDataBase->m_runnableMembers.erase(runnableData);
+    m_portPoolDataBase->m_runnableMembers.remove_if([&](runtime::RunnableData& r) { return &r == runnableData; });
```
elfenpiff, line 107: Please do not capture everything. In this case we only require to capture the `portData` pointer so please change the lambda to `[runnableData](...`

elfenpiff (changes requested):

bishibashiB (commented):
Addressed comments from @elfenpiff 
Ether committed or forwarded to #257

iceoryx_posh/source/roudi/iceoryx_port_pool.cpp
```diff
@@ -77,26 +77,28 @@ IceOryxPortPool::addReceiverPort(const capro::ServiceDescription& serviceDescrip
 }
 
 /// @deprecated #25
-void IceOryxPortPool::removeSenderPort(SenderPortType::MemberType_t* const portData) noexcept
+void IceOryxPortPool::removeSenderPort(const SenderPortType::MemberType_t& portData) noexcept
 {
-    m_portPoolData->m_senderPortMembers.erase(portData);
```
MatthiasKillat, line 82: Ok, worked before because the iterators where pointers (for cxx::vector). Now to ensure deletion we need to provide a reference to an object actually in the container (equivalent to a pointer in this regard). Should be fine, but worth thinking about that it is still correct and equivalent to before. We cannot delete by id or name or anything else. The address of the object to be deleted is essentially the id.

bishibashiB: > We cannot delete by id or name or anything else. The address of the object to be deleted is essentially the id.

Q: Could there appear a race where the pointer-is-ID approach becomes corrupted due to different address ranges in processes... ?

iceoryx_posh/include/iceoryx_posh/internal/roudi/iceoryx_port_pool.hpp
MatthiasKillat: A discussion lead to the following proposal: return a container (probably still a vector if it is just read from) of pointers to the ports as is instead of a reference to the underlying container, which is now a list. 
This more overhead since this container needs to be constructed everytime the function is called (this is also the case now), but does not expose the internal list (but the ports, as before).

In the long run this can maybe be improved even more by passing the list reference as proposed, at the cost of losing encapsulation.

The underlying container should still be changed to a list to allow efficient insertion and removal.

iceoryx_posh/include/iceoryx_posh/internal/popo/receiver_handler.hpp
```diff
@@ -53,7 +53,7 @@ class ReceiverHandler : public LockingPolicy
     using this_type = ReceiverHandler_t;
 
   public:
-    using ReceiverVector_t = cxx::vector<relative_ptr<ReceiverPortType::MemberType_t>, MaxReceivers>;
+    using ReceiverList_t = cxx::list<relative_ptr<ReceiverPortType::MemberType_t>, MaxReceivers>;
```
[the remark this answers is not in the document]
MatthiasKillat: You may be right here. Cache Locality is much worse with the list and this is probably the reason for a large performance loss while iterating over its elements.

elBoberido (approved):
I just wanted to propose to split this up into multiple PR since I feared you would beat me for the longest open PR

elfenpiff (approved):

PR #232, opened 2020-08-10
Iox #25 building block ports cleanups

budrus:


iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -26,10 +26,10 @@ namespace iox
 {
 namespace roudi
 {
-capro::Interfaces StringToEInterfaces(std::string f_str)
+capro::Interfaces StringToEInterfaces(std::string str)
```
mossmaurice, line 29: ❤️

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -245,27 +244,30 @@ void PortManager::handleInterfaces()
...
 
-    for (auto l_applicationPortData : m_portPool->appliactionPortDataList())
+    for (auto applicationPortData : m_portPool->appliactionPortDataList())
     {
-        iox::popo::ApplicationPort l_applicationPort(l_applicationPortData);
-        while (l_applicationPort.getCaProMessage(l_caproMessage))
+        iox::popo::ApplicationPort applicationPort(applicationPortData);
+
+        for (auto maybeCaproMessage = applicationPort.getCaProMessage(); maybeCaproMessage.has_value();
+             maybeCaproMessage = applicationPort.getCaProMessage())
```
mossmaurice, line 254: 😎

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: I'm watching https://www.youtube.com/watch?v=MBRoCdtZOYg right now, so why not give it a more descriptive name like `caproInterface`
... and why is this called `StringToEInterfaces` when there is no `EInterface` anymore?

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: reminds me that I wanted try to code something like this in C++ for ages
```
for(auto caproMessage: applicationPort.iterCaproMessage())
{
    // do something with caproMessage
}
```
something generic which calls getNext until it's a `nullopt`

iceoryx_posh/source/roudi/port_manager.cpp
budrus: I just wanted to get rid of f_ . You're dialing another number ;-)

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/subscriber_port_user.hpp
```diff
@@ -48,7 +48,7 @@ class SubscriberPortUser
     /// @brief try to subscribe to all matching publishers
     /// @param[in] queueCapacity, capacity of the queue where chunks are stored before they are passed to the user with
     /// getChunk. Caution: Depending on the underlying queue there can be a different overflow behavior
-    void subscribe(const uint32_t queueCapacity = MAX_RECEIVER_QUEUE_CAPACITY) noexcept;
+    void subscribe(const uint32_t queueCapacity = MemberType_t::ChunkQueueData_t::MAX_CAPACITY) noexcept;
```
marthtz, line 51: Must have missed that one in mine...

iceoryx_posh/source/popo/gateway_generic.cpp
```diff
@@ -28,15 +28,25 @@ GatewayGeneric::GatewayGeneric(const capro::Interfaces f_interface) noexcept
...
 GatewayGeneric::~GatewayGeneric() noexcept
 {
-    if(m_interfaceImpl)
+    if (m_interfaceImpl)
     {
         m_interfaceImpl.destroy();
     }
 }
 
 bool GatewayGeneric::getCaProMessage(CaproMessage& msg) noexcept
```
dkroenke, line 37: maybe const here?

iceoryx_posh/source/popo/ports/application_port.cpp
```diff
@@ -23,20 +24,27 @@ ApplicationPort::ApplicationPort(ApplicationPortData* const applicationPortDataP
...
+        return cxx::make_optional<capro::CaproMessage>(maybeMessage.value());
+    }
+    else
+    {
+        return cxx::nullopt_t();
+    }
 }
 
-bool ApplicationPort::getCaProMessage(capro::CaproMessage& message) noexcept
+void ApplicationPort::dispatchCaProMessage(const capro::CaproMessage& caProMessage) noexcept
```
dkroenke, line 41: caProMessage looks ugly for consistency i would take msg here. But i have not strong feelings about it as @elBoberido would say ;-)

iceoryx_posh/source/popo/gateway_generic.cpp
elfenpiff, line 37: I would suggest changing the signature to `cxx::optional<CaproMessage> GatewayGeneric::getCaProMessage() noexcept`
This would then be consistent to all the other methods which can return a value but do not have to.
Additionally, all the other ports have this signature.

iceoryx_posh/source/popo/ports/application_port.cpp
```diff
@@ -23,20 +24,27 @@ ApplicationPort::ApplicationPort(ApplicationPortData* const applicationPortDataP
 {
 }
 
-bool ApplicationPort::dispatchCaProMessage(const capro::CaproMessage& message) noexcept
+cxx::optional<capro::CaproMessage> ApplicationPort::getCaProMessage() noexcept
 {
-    return getMembers()->m_caproMessageFiFo.push(message);
+    auto maybeMessage = getMembers()->m_caproMessageFiFo.pop();
```
elfenpiff, line 29: You could also write: 
```
return getMembers()->m_caproMessageFiFo.pop();
```

and get rid off the additional if else.

iceoryx_posh/source/popo/ports/interface_port.cpp
```diff
@@ -24,25 +24,27 @@ InterfacePort::InterfacePort(InterfacePortData* const interfacePortDataPtr) noex
 {
 }
 
-bool InterfacePort::dispatchCaProMessage(const capro::CaproMessage& message) noexcept
+cxx::optional<capro::CaproMessage> InterfacePort::getCaProMessage() noexcept
 {
-    const bool pushResult = getMembers()->m_caproMessageFiFo.push(message);
-    if (!pushResult)
+    auto maybeMessage = getMembers()->m_caproMessageFiFo.pop();
```
elfenpiff, line 29: You could also here write:
```
return getMembers()->m_caproMessageFiFo.pop();
```

iceoryx_posh/source/roudi/port_manager.cpp
elfenpiff: You touched it, now you have to change the argument type to: `const std::string & str` ;) ... Please

iceoryx_posh/source/popo/gateway_generic.cpp
budrus: Where? The provided CaProMessage cannot be const as it is set by this method. This is an API like the often used FiFo pop() without optional

budrus: This is no more port but the next layer. Here we have a new game and breaking this API has consequences for all legacy gateways. The costs are high for having it more the modern C++ way. And it is not inconsistent as it is another layer...

iceoryx_posh/source/popo/ports/application_port.cpp
budrus: good point, I somehow had in mind that the FiFo has no optional but a bool return.

iceoryx_posh/source/roudi/port_manager.cpp
budrus: OK, will touch it again ;-)

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -205,21 +205,30 @@ TEST_F(RoudiFindService_test, InterfacePort)
...
+                break;
+            }
+        }
+        else
         {
-            serviceFound = true;
             break;
         }
-    }
+    } while (true);
```
elBoberido, line 230: please use either `while (!serviceFound)` or something like this
```
while (auto caproMessage = interfacePort.getCaProMessage())
{
    if ((caproMessage.m_serviceDescription.getServiceIDString() == IdString("service1"))
        && (caproMessage.m_serviceDescription.getInstanceIDString() == IdString("instance1"))
        && ((caproMessage.m_serviceDescription.getEventIDString() == IdString(iox::capro::AnyEventString))))
    {
        serviceFound = true;
        break;
    }
}
```
It does exactly the same as the code above but with less code ;)

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: It just ocured to me that we already can do something similar
```
while (auto caproMessage = applicationPort.getCaProMessage())
{
    // do something with capro message
}
```
the while loop uses the bool operator of the optional and stops iterating if it's a nullopt instead of a caproMessage

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.inl
```diff
@@ -109,22 +109,14 @@ inline uint64_t ChunkQueuePopper<ChunkQueueDataType>::getMaximumCapacity() const
 template <typename ChunkQueueDataType>
 inline void ChunkQueuePopper<ChunkQueueDataType>::clear() noexcept
 {
-    do
+    while (auto maybeChunkTuple = getMembers()->m_queue.pop())
```
elBoberido, line 112: nitpicking, but this is not a maybeChunkTuple. The loop is not entered if it's a nullopt

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -205,13 +205,13 @@ TEST_F(RoudiFindService_test, InterfacePort)
     auto interfacePortData = receiverRuntime->getMiddlewareInterface(iox::capro::Interfaces::SOMEIP);
     iox::popo::InterfacePort interfacePort(interfacePortData);
 
-    iox::capro::CaproMessage caproMessage;
     this->InterOpWait();
 
     bool serviceFound = false;
 
-    while (interfacePort.getCaProMessage(caproMessage))
+    while (auto maybeCaProMessage = interfacePort.getCaProMessage())
```
elBoberido, line 212: `maybe` sounds here a bit wrong, but I also don't have a better name

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -245,27 +244,29 @@ void PortManager::handleInterfaces()
...
+    for (auto applicationPortData : m_portPool->appliactionPortDataList())
     {
-        iox::popo::ApplicationPort l_applicationPort(l_applicationPortData);
-        while (l_applicationPort.getCaProMessage(l_caproMessage))
+        iox::popo::ApplicationPort applicationPort(applicationPortData);
+
+        while (auto maybeCaproMessage = applicationPort.getCaProMessage())
         {
-            switch (l_caproMessage.m_type)
+            auto caproMessage = maybeCaproMessage.value();
```
elBoberido, line 255: please use `auto&` to prevent a copy of caproMessage

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
budrus: I have the feeling we somehow agreed that maybe* is a good name for the optional return

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.inl
mossmaurice: Still, I would suggest to use the `maybe*` everywhere where an `optional` or `expected` is returned. This triggers the reminder in my brain "Hey that's an optional/expected!"

budrus: Valid point. In the loop body it is a chunk tuple. But when we do the assignment it is maybe one ;-)

Issue #234, opened 2020-08-11
Build script attempts to run DDS Gateway tests when they are not built


ithier:
**Observed result or behaviour:**
Building with command: `tools/iceoryx_build_test clean release test` 
Results in error:
    cd: iceoryx/build/dds_gateway/test: No such file or directory

**Expected result or behaviour:**
Script does not attempt to run dds_gateway tests as they have not been explicitly request using the `with-dds-gateway` flag

PR #235, opened 2020-08-11
iox-#234 Only run DDS tests when DDS Gateway is built.

MatthiasKillat:
Signed-off-by: Ithier Jeff (CC-AD/EYF1) <jeff.ithier@de.bosch.com>


ithier:
This fixed #234

PR #237, opened 2020-08-12
Iox #25 subscriber port for multi producer

budrus:


iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -301,3 +301,165 @@ TEST_F(SubscriberPort_test, nackWhenNotWaitingForResultsInError)
...
+TEST_F(SubscriberPortMultiProducer_test, subscribeCallResultsInSubCaProMessage)
+{
+    m_sutUserSideMultiProducer.subscribe();
+
+    auto maybeCaproMessage = m_sutRouDiSideMultiProducer.getCaProMessage();
+
+    EXPECT_TRUE(maybeCaproMessage.has_value());
+    auto caproMessage = maybeCaproMessage.value();
+    EXPECT_THAT(caproMessage.m_type, Eq(iox::capro::CaproMessageType::SUB));
+    EXPECT_THAT(caproMessage.m_serviceDescription, Eq(iox::capro::ServiceDescription("x", "y", "z")));
```
mossmaurice, line 353: Move the service description to a struct and check for `Eq(TestServiceDescription)`? x, y and z are magic strings 🦄

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -301,3 +301,165 @@ TEST_F(SubscriberPort_test, nackWhenNotWaitingForResultsInError)
...
+    m_sutRouDiSideMultiProducer.getCaProMessage(); // only RouDi changes state
+    iox::capro::CaproMessage caproMessage(iox::capro::CaproMessageType::OFFER,
+                                          iox::capro::ServiceDescription("x", "y", "z"));
+
+    auto maybeCaproMessage = m_sutRouDiSideMultiProducer.dispatchCaProMessage(caproMessage);
+
+    EXPECT_TRUE(maybeCaproMessage.has_value());
+    auto caproMessageResponse = maybeCaproMessage.value();
+    EXPECT_THAT(caproMessageResponse.m_type, Eq(iox::capro::CaproMessageType::SUB));
+    EXPECT_THAT(caproMessageResponse.m_serviceDescription, Eq(iox::capro::ServiceDescription("x", "y", "z")));
```
mossmaurice, line 405: Using other strings than x, y and z for each test case would be safer.

iceoryx_posh/source/popo/ports/subscriber_port_multi_producer.cpp
```diff
@@ -26,15 +26,64 @@ SubscriberPortMultiProducer::SubscriberPortMultiProducer(
...
+        caproMessage.m_chunkQueueData = static_cast<void*>(&getMembers()->m_chunkReceiverData);
+        caproMessage.m_historyCapacity = getMembers()->m_historyRequest;
+
+        return cxx::make_optional<capro::CaproMessage>(caproMessage);
+    }
+    else if ((capro::CaproMessageType::ACK == caProMessage.m_type)
+             || (capro::CaproMessageType::NACK == caProMessage.m_type)
+             || (capro::CaproMessageType::STOP_OFFER == caProMessage.m_type))
+    {
+        // we ignore all these messages for multi-producer
```
elfenpiff, line 78: If we ignore all those messages then where is the subscription state updated? I think at the moment this class has only the state WAIT_FOR_OFFER or NOT_SUBSCRIBED. Is this intended? 

I think for the multi producer subscriber we need a list of all the producer to which the subscriber is connected. When we receive a STOP_OFFER from them we have to cross them from our list and if no more producers are available we switch to the NOT_SUBSCRIBED state.

We could also go in the direction of ignoring the subscription state but then the multi producer subscriber port would behave differently and then the question arises: is this something a user would expect from such a subscriber?

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
elfenpiff: :moyai:

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -301,3 +301,165 @@ TEST_F(SubscriberPort_test, nackWhenNotWaitingForResultsInError)
     EXPECT_FALSE(maybeCaproMessage.has_value());
     EXPECT_TRUE(errorHandlerCalled);
 }
+
+class SubscriberPortMultiProducer_test : public Test
```
elfenpiff, line 305: Please move the following code into a separate file called `test_popo_subscriber_port_multi_producer.cpp`

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -30,14 +30,14 @@
```
elfenpiff, line 30: Please rename the file into `test_popo_subscriber_port_single_producer.cpp`

iceoryx_posh/source/popo/ports/subscriber_port_multi_producer.cpp
budrus: We keep it simple first. In an n:m pub/sub setup the subscription state is no more easy. When are you subscribed, when there is at least one publisher. Or maybe there are two and you would need both to be fine, would you then be subscribed if you are subscribed to all the ones you have in mind? A bookkeeping as you propose is a lot effort and I would first get some experience and see if we want to go for this. I also have the feeling that in other n:m technologies like DDS there is no subscription state like this for a subscriber. They more look on available publishers and subscribers on a topic.  WAIT_FOR_OFFER was maybe not the best choice. I would propose that in a multi-pusher setup we have NOT_SUBSCRIBED and SUBSCRIBED. And this state reflects the subscriber wish. I.e. no matter how many publishers there are, if you want to be subscribed we subscribe to all available ones and are in the SUBSCRIBED state. We can make it more complex later if we see the need. What do you think @elfenpiff

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
budrus: This is a question about SW units. We have a separation of data structures in the shared memory and access classes without own data members (besides pointer to this shared memory object). We also separate the multi-threaded ones in access classes for different threads of execution (like RouDi and User). You could have everything in one class, it is more an architectural decision to do it that way. For me all these classes are the SubscriberPort SW unit. And if each unit has a unit test this would all be in one file.

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
budrus: See above

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
```diff
@@ -301,3 +301,165 @@ TEST_F(SubscriberPort_test, nackWhenNotWaitingForResultsInError)
...
+    auto errorHandlerCalled{false};
+    auto errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler([&errorHandlerCalled](
+        const iox::Error, const std::function<void()>, const iox::ErrorLevel) { errorHandlerCalled = true; });
+    iox::capro::CaproMessage caproMessage(iox::capro::CaproMessageType::UNSUB,
+                                          iox::capro::ServiceDescription("x", "y", "z"));
+
+    auto maybeCaproMessage = m_sutRouDiSideMultiProducer.dispatchCaProMessage(caproMessage);
+
+    EXPECT_FALSE(maybeCaproMessage.has_value());
+    EXPECT_TRUE(errorHandlerCalled);
```
dkroenke, line 470: It should be also against the corresponding errortype tested:
This can be done like this:

    auto errorHandlerCalled{false};
    iox::Error receivedError;

    auto errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler(
        [&errorHandlerCalled,
         &receivedError](const iox::Error error, const std::function<void()>, const iox::ErrorLevel) {
            errorHandlerCalled = true;
            receivedError = error;
        });

    ASSERT_TRUE(errorHandlerCalled);
    ASSERT_THAT(receivedError, Eq(iox::Error::kPOPO__CAPRO_PROTOCOL_ERROR));

iceoryx_posh/test/moduletests/test_popo_subscriber_port.cpp
mossmaurice: Can we write this down somewhere?

iceoryx_posh/source/popo/ports/subscriber_port_multi_producer.cpp
elfenpiff: I agree!

Issue #243, opened 2020-08-14
Segmentation Fault in Posix Timer


elfenpiff:
## Required information

**Operating system:**
Ubuntu 18.04 LTS, Archlinux

**Compiler version:**
GCC 7.4.0, CLANG 10.0, GCC 10.2

**Observed result or behaviour:**
The now disabled unit test: `Timer_test.SelfTriggeringTimerWorksAndDoesNotCauseSegFault` seg faults

**Expected result or behaviour:**
The unit does not seg fault

shankar-in:
@elfenpiff could you pls let me know if this seg fault issue is fixed? Still the test looks disabled `DISABLED_SelfTriggeringTimerWorksAndDoesNotCauseSegFault`

elfenpiff:
@shankar-in the issue seems to be solved.

shankar-in:
@shankar-in Okay I will enable the test again.

Issue #242, opened 2020-08-14
Timer unit tests should be TIMING_TESTS for timing tests


elfenpiff:
## Required information

If the timing tests are not declared as TIMING_TESTS it is possible that they are sporadically failing

elfenpiff:
Fixed with pull request https://github.com/eclipse/iceoryx/pull/244

PR #244, opened 2020-08-14
iox-#242 timing test fix

elfenpiff:
- … of segfault in posix timer

Signed-off-by: Eltzschig Christian (CC-AD/ESW1) <christian.eltzschig2@de.bosch.com>

Found a segmentation fault in posix timer see issue #243


iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -520,3 +520,24 @@ TIMING_TEST_F(Timer_test, CatchUpPolicySkipToNextBeatCallsLessCallbacksThanASAPT
 
     TIMING_TEST_EXPECT_TRUE(softTimerCounter < asapTimerCounter);
 });
+
+TEST_F(Timer_test, DISABLED_SelfTriggeringTimerWorksAndDoesNotCauseSegFault)
+{
+    Duration selfTriggerTimeout = 1_ns;
```
dkroenke, line 526: When you start the timer with 1 ns the value is truncated to zero because the cxx::duartion has an truncation error. See https://github.com/eclipse/iceoryx/issues/190

elfenpiff: No this is not the case here. If you uncomment line 531 the test will work fine and the condition `counter == repetition` will hold. The line 531 causes the seg fault since there seems to be some weird race condition occuring.

MatthiasKillat: I assume the timeout is chosen this small but nonzero for a reason (e.g. to trigger almost immediately). I do not know what the variable name is supposed to mean (what is self triggering?). Does the timer not just have a callback in which user code (with some restrictions) can be executed? As far as I see from the test, the timer restarts itself in the callback, but with the ONCE policy. But it is also initially started with this policy.

I have not researched it, but the underlying structure is a POSIX timer. Could it be that the fire ONCE policy causes it to cleanup something in the kernel before the other call has properly finished? Each callback runs in some kind of context/thread, but I assume the ONCE policy is used by the kernel data for the timer to indicate it is not needed after the callback is complete. This might cause the other instance to access an invalid timer struct.
Maybe worth investigating if the cause for the segmentation fault is not yet known...my best guess from my current understanding.

Bottom line: maybe restart is not allowed when RunMode::Once is used.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -520,3 +520,24 @@ TIMING_TEST_F(Timer_test, CatchUpPolicySkipToNextBeatCallsLessCallbacksThanASAPT
...
+    int repetitions = 100;
+    std::atomic_int counter{0};
+    {
+        Timer sut(selfTriggerTimeout, [&] {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
MatthiasKillat, line 531: What is this sleep for? To provoke the error?  If the error depends on the sleep, the test is not well-formed (or the timer used wrongly). I think if the mechanism works properly, there should be no need for a sleep, we should be able to restart it as soon as we are in the callback.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -520,3 +520,24 @@ TIMING_TEST_F(Timer_test, CatchUpPolicySkipToNextBeatCallsLessCallbacksThanASAPT
 
     TIMING_TEST_EXPECT_TRUE(softTimerCounter < asapTimerCounter);
 });
+
+TEST_F(Timer_test, DISABLED_SelfTriggeringTimerWorksAndDoesNotCauseSegFault)
+{
+    Duration selfTriggerTimeout = 1_ns;
+    int repetitions = 100;
```
MatthiasKillat, line 527: Arbitrary number, I assume we repeat it to test that it can restart itself in succession.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -520,3 +520,24 @@ TIMING_TEST_F(Timer_test, CatchUpPolicySkipToNextBeatCallsLessCallbacksThanASAPT
...
+            if (counter < repetitions)
+            {
+                EXPECT_FALSE(
+                    sut.restart(selfTriggerTimeout, Timer::RunMode::ONCE, Timer::CatchUpPolicy::IMMEDIATE).has_error());
+            }
+            ++counter;
+        });
+
+        sut.start(Timer::RunMode::ONCE, Timer::CatchUpPolicy::IMMEDIATE);
+        std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
MatthiasKillat, line 541: If this sleep expires, the timer may still be active if I did not miss something. Is this intended? The sut dtor will be called, the counter go out of scope, the callback could access data now that is gone (either from the timer, but in any case the counter itself). There is no other join or wait mechanism that waits until all repetitions (or even one) were performed.

Conclusion: I may not fully understand the test but I think object lifetime is a problem here.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -25,6 +25,7 @@
 
 using namespace ::testing;
 
+using namespace iox::units;
```
mossmaurice, line 28: Please do not start `using namespace`. Will lead to namespace pollution very easily, when used in other places

elfenpiff: @mossmaurice This cannot happen in cpp files. The `using namespace` statement is only a problem in header files since everyone who is including this header file is also executing the `using namespace` statement.

A cpp file is on the other hand a compilation unit where no one knows how the implementation looks like since no one is including a cpp file. You can access symbols in a cpp file by knowing that they are present (done in the header) and linking against them later. 

But `using namespace` is not a symbol. It is a compile time directive to access a certain scope so we do not have a problem there!

iceoryx_utils/test/moduletests/test_posix_timer.cpp
elfenpiff: The segfault is documented internally and a issue is already created, see #243 .
The goal of this unit test is to ease the investigation of the segmentation fault when someone starts digging here.
- `restart` is allowed and makes absolute sense when the policy is `RunMode::Once`

iceoryx_utils/test/moduletests/test_posix_timer.cpp
elfenpiff: I added some documentation

iceoryx_utils/test/moduletests/test_posix_timer.cpp
elfenpiff: From my point of view this tests is valid code and it should not segfault. We just wait here 1000 ms since it is sufficient time to provoke the segfault.
I share the same conclusion but at the moment I do not have the time to solve this segfault, therefore I created #243 .

MatthiasKillat: AFAIK you cannot stop a POSIX timer reliably once the kernel has decided to start the callback (but is not yet running). I encountered this issue while trying to implement a timed_wait function in another project. And if you cannot stop it, you have to make sure that the data the callback tries to access is still there (even if it is just a flag/counter).

Hard for me to say whether the timer is broken, but I guess what you are saying is that the test is not responsible to block (a bit dangerous in case of a malfunction of the timer but maybe unavoidable).

TLDR: I suspect there is a fundamental flaw in the timer in the small window before the callback actually runs.

iceoryx_utils/test/moduletests/test_posix_timer.cpp
MatthiasKillat: @elfenpiff is correct, it is fine this way.

PR #245, opened 2020-08-14
Iox #25 integrate unique

elfenpiff:


iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
```diff
@@ -11,33 +11,46 @@
...
 namespace iox
 {
-namespace cxx
+namespace popo
 {
-/// @brief Unique ID depending on type. If you would like to assign different
+namespace internal
+{
+/// @brief Has to be set on roudi startup of so that a unique roudi id is set
+///         for all newly generated unique ids.
```
dkroenke, line 30: param[in] id missing

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
```diff
@@ -11,33 +11,46 @@
...
+void setUniqueRouDiId(const uint16_t id) noexcept;
+
+/// @brief Sets the RouDi id to an undefined state
+void unsetUniqueRouDiId() noexcept;
+
+/// @brief returns the unique roudi id
```
dkroenke, line 36: doxygen return explanation missing

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.inl
```diff
@@ -11,21 +11,27 @@
...
+namespace popo
 {
 template <typename T>
 std::atomic<uint64_t> TypedUniqueId<T>::globalIDCounter{0u};
 
 template <typename T>
 inline TypedUniqueId<T>::TypedUniqueId() noexcept
-    : ThisType(newtype::internal::ProtectedConstructor, globalIDCounter.fetch_add(1u, std::memory_order_relaxed))
+    : ThisType(cxx::newtype::internal::ProtectedConstructor,
+               (static_cast<uint64_t>(internal::getUniqueRouDiId()) << 48)
```
dkroenke, line 27: Where does the magic numbers come from?

iceoryx_posh/include/iceoryx_posh/internal/roudi_environment/roudi_environment.hpp
```diff
@@ -56,7 +57,7 @@ class RouDiEnvironment
         BASE,
     };
     /// @brief for implementations on top of RouDiEnvironment
-    RouDiEnvironment(BaseCTor);
+    RouDiEnvironment(BaseCTor, const uint16_t uniqueRouDiId = 0);
```
dkroenke, line 60: u suffix is missing

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.hpp
```diff
@@ -124,7 +124,7 @@ class ChunkDistributor
     MemberType_t* getMembers() noexcept;
 
   private:
-    MemberType_t* m_chunkDistrubutorDataPtr;
+    MemberType_t* m_chunkDistrubutorDataPtr{nullptr};
```
budrus, line 127: There is only on c'tor which sets this pointer (with cxx::not_null). When I see this change I could get the impression that this member could also be nullptr. I also already saw QAC warnings that complained when you set a member in the initializer list and with this initializer. So I decided to not have this nullptr initialization. But maybe this is a matter of taste..

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor_data.hpp
```diff
@@ -33,6 +33,7 @@ namespace popo
 template <typename ChunkDistributorDataProperties, typename LockingPolicy, typename ChunkQueuePusherType>
 struct ChunkDistributorData : public LockingPolicy
 {
+    using ThisType_t = ChunkDistributorData<ChunkDistributorDataProperties, LockingPolicy, ChunkQueuePusherType>;
     using LockGuard_t = std::lock_guard<
         const ChunkDistributorData<ChunkDistributorDataProperties, LockingPolicy, ChunkQueuePusherType>>;
```
budrus, line 38: maybe reuse your brand new thisType for the LockGuard?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_data.hpp
```diff
@@ -29,6 +29,7 @@ namespace popo
 template <typename ChunkQueueDataProperties, typename LockingPolicy>
 struct ChunkQueueData : public LockingPolicy
 {
+    using ThisType_t = ChunkQueueData<ChunkQueueDataProperties, LockingPolicy>;
     using LockGuard_t = std::lock_guard<const ChunkQueueData<ChunkQueueDataProperties, LockingPolicy>>;
```
budrus, line 33: same same

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.hpp
```diff
@@ -32,6 +33,9 @@ enum class AllocationError
     TOO_MANY_CHUNKS_ALLOCATED_IN_PARALLEL
 };
 
+struct BasePortData;
```
budrus, line 36: Ugly dependency I would say. No problem with the forward declaration but not nice. There is no ports in ChunkSender

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.hpp
```diff
@@ -56,7 +61,8 @@ class ChunkSender : public ChunkDistributorType
     /// @param[in] payloadSize, size of the user paylaod without additional headers
     /// @return on success pointer to a ChunkHeader which can be used to access the payload and header fields, error if
     /// not
-    cxx::expected<mepoo::ChunkHeader*, AllocationError> allocate(const uint32_t payloadSize) noexcept;
+    cxx::expected<mepoo::ChunkHeader*, AllocationError> allocate(const uint32_t payloadSize,
+                                                                 const UniqueId_t originId) noexcept;
```
budrus, line 65: please adapt the doxygen header

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
```diff
@@ -11,33 +11,46 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#ifndef IOX_UTILS_CXX_TYPED_UNIQUE_ID_HPP
-#define IOX_UTILS_CXX_TYPED_UNIQUE_ID_HPP
+#ifndef IOX_POSH_POPO_BUILDING_BLOCKS_TYPED_UNIQUE_ID_HPP
+#define IOX_POSH_POPO_BUILDING_BLOCKS_TYPED_UNIQUE_ID_HPP
```
budrus, line 15: why now in the popo bulding blocks? I thought that this is a generic unique ID helper that we can use for other things too. I guess because of the RouDi ID that is in now. But than this becomes a very specific thing and is more ore less a Port ID and no more a generic UniqueID. Is it still worth the effort?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
budrus: set on roudi startup of so?

elfenpiff: done

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
elfenpiff: done

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.inl
elfenpiff: done

iceoryx_posh/include/iceoryx_posh/internal/roudi_environment/roudi_environment.hpp
elfenpiff: done

iceoryx_posh/source/popo/building_blocks/typed_unique_id.cpp
```diff
@@ -0,0 +1,48 @@
...
+void unsetUniqueRouDiId() noexcept
+{
+    hasDefinedUniqueRouDiId = false;
+}
+
+uint16_t getUniqueRouDiId() noexcept
+{
+    if (!hasDefinedUniqueRouDiId)
+    {
+        errorHandler(Error::kPOPO__TYPED_UNIQUE_ID_ROUDI_HAS_NO_DEFINED_UNIQUE_ID);
```
budrus, line 45: please always set an error level

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -44,6 +46,7 @@ void CmdLineParser::parse(int argc, char* argv[], const CmdLineArgumentParsingMo
             std::cout << "Options:" << std::endl;
             std::cout << "-h, --help                        Display help." << std::endl;
             std::cout << "-v, --version                     Display version." << std::endl;
+            std::cout << "-u, --unique-roudi-id <INT>       Display version." << std::endl;
```
budrus, line 49: Display version? copy 'n' waste

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.hpp
elfenpiff: Here I have the C++ Core Guidelines on my side and I think the CTor approach is actually a bad code style: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer 

In this particular case I would suggest to ignore the QAC warning and perform all the initialization inside the header and only specialize the member variables during the constructor initialization.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
elfenpiff: The newtype pattern is still worth the effort since we would like to use it for instance to distinct:
 - Service, Instance, Event string by creating own types which would allow us to fetch even more errors at compile time
 - Service, Instance, Event id integers

The UniqueId is now part of the building blocks since it is now consisting of two parts:
- 16 bit RouDi id
- 48 bit Unique id

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
elfenpiff: whoopsie

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -60,6 +63,18 @@ void CmdLineParser::parse(int argc, char* argv[], const CmdLineArgumentParsingMo
             m_run = false;
             break;
 
+        case 'u':
+        {
+            uint16_t roudiId{0u};
+            if (!cxx::convert::fromString(optarg, roudiId))
+            {
+                LogError() << "The RouDi id must be in the range of [0, " << ((1 << 16) - 1) << "]";
```
mossmaurice, line 71: Could we have a `constexpr` here instead of shifts?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.hpp
elfenpiff: solved

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor_data.hpp
```diff
@@ -33,6 +33,7 @@ namespace popo
 template <typename ChunkDistributorDataProperties, typename LockingPolicy, typename ChunkQueuePusherType>
 struct ChunkDistributorData : public LockingPolicy
 {
+    using ThisType_t = ChunkDistributorData<ChunkDistributorDataProperties, LockingPolicy, ChunkQueuePusherType>;
```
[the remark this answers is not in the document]
elfenpiff: Never enough ;). In my opinion `ThisType_t` is more readable.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
elfenpiff: done

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
```diff
@@ -61,11 +62,13 @@ class CmdLineParser
     bool getRun() const;
     iox::log::LogLevel getLogLevel() const;
     MonitoringMode getMonitoringMode() const;
+    cxx::optional<uint16_t> getUniqueRouDiId() const noexcept;
 
   protected:
     bool m_run{true};
     iox::log::LogLevel m_logLevel{iox::log::LogLevel::kWarn};
     MonitoringMode m_monitoringMode{MonitoringMode::ON};
+    cxx::optional<uint16_t> m_uniqueRouDiId;
```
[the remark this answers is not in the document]
elfenpiff: @mossmaurice The command line parser class reads the optional command line setting of a roudi id from the command line. If there is a roudi id set, then the roudi id is used for the whole system to identify entities uniquely. If there is no roudi id set then we use 0 as default roudi id.

The optional uniqueRouDiId which is read from the command line is stored in this member

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.inl
```diff
@@ -11,21 +11,27 @@
...
+                   + ((globalIDCounter.fetch_add(1u, std::memory_order_relaxed) << 16) >> 16))
```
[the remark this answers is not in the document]
elfenpiff: Got rid of the magic numbers! Shifts are needed since the first 16-bit of this 64-bit integer is used for the RouDi id and the remaining 48-bits are used for the unique id itself.

dkroenke:
@elfenpiff There are merge conflicts

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
```diff
@@ -0,0 +1,130 @@
...
+    ///         previous created id
+    TypedUniqueId() noexcept;
+
+    /// @brief constructor which creates an invalid id
+    TypedUniqueId(CreateInvalidId_t) noexcept;
+
+    bool isValid() const noexcept;
+
+  private:
+    static constexpr uint64_t INVALID_UNIQUE_ID = 0u;
```
MatthiasKillat, line 119: An explanation which bits are used for the different id parts would have been useful. I.e. highest 16 bits for the Roudi Id part it seems, the lower 48 bit for the specific part generated by the global counter associated with type T.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.inl
```diff
@@ -0,0 +1,53 @@
...
+template <typename T>
+std::atomic<uint64_t> TypedUniqueId<T>::globalIDCounter{0u};
+
+template <typename T>
+inline TypedUniqueId<T>::TypedUniqueId() noexcept
+    : ThisType(cxx::newtype::internal::ProtectedConstructor,
+               (static_cast<uint64_t>(internal::getUniqueRouDiId()) << UNIQUE_ID_BIT_LENGTH)
+                   + ((globalIDCounter.fetch_add(1u, std::memory_order_relaxed) << ROUDI_ID_BIT_LENGTH)
+                      >> ROUDI_ID_BIT_LENGTH))
+{
```
MatthiasKillat, line 30: The shift back and forth of ROUDI_ID_BIT_LENGTH is to zero the highest bits right? I would have preferred a constexpr mask for both parts (and using & to access the different parts). For positive numbers (i.e. uint) I am sure that it is guaranteed that the upper bits will be zero when we shift right. 

But note that there are some undefined behaviour cases like this which seemingly work, masking always works.

PR #266, opened 2020-09-02
iox-#264 fix char array assignment to cxx::string

### LINKED ISSUE #264: cxx::string doesn't respect the null termination in an assigned char array
## Required information

**Operating system:**
all

**Compiler version:**
all

the following code fails
```
char foo[50] = "bar";
cxx::string<100> bar(foo);
assert(bar.size() == strnlen(foo, 50));
```
`bar.size()` would actually return 50, the size of the char array.

The code above doesn't fail.

create a char array (not string literal) which has a null termination and assign it to a cxx::string.

elBoberido:
closes #264 

Signed-off-by: Kraus Mathias (CC-AD/ESW1) <mathias.kraus2@de.bosch.com>

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
+++ b/iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
@@ -119,9 +119,17 @@ inline string<Capacity>& string<Capacity>::operator=(const char (&rhs)[N]) noexc
     {
         return *this;
     }
-    std::memcpy(m_rawstring, rhs, N - 1u);
-    m_rawstring[N - 1u] = '\0';
-    m_rawstringSize = N - 1u;
+
+    m_rawstringSize = strnlen(rhs, Capacity);
+    std::memcpy(m_rawstring, rhs, m_rawstringSize);
+    m_rawstring[m_rawstringSize] = '\0';
+
+    if (rhs[m_rawstringSize] != '\0')
+    {
+        std::cerr << "iox::cxx::string: Assignment of array which is not zero-terminated! Last value of array "
+                     "overwritten with 0!"
+                  << std::endl;
+    }
     return *this;
 }
 
--- a/iceoryx_utils/test/moduletests/test_cxx_string.cpp
+++ b/iceoryx_utils/test/moduletests/test_cxx_string.cpp
@@ -361,6 +361,23 @@ TEST(String100, UnsafeCharToStringConstrIncludingNullCharWithCountResultsInSizeC
     EXPECT_THAT(testSubject.c_str(), StrEq("ice\0ryx"));
 }
 
+TEST(CharArrayAssignment, AssignZeroTerminatedCharArrayOfSizeForFullCapa)
+{
+    char testString[8] = "iceoryx";
+    string<7> testSubject(testString);
+    EXPECT_THAT(testSubject.size(), Eq(7));
+    EXPECT_THAT(testSubject.c_str(), StrEq("iceoryx"));
+}
+
+TEST(CharArrayAssignment, AssignNonZeroTerminatedCharArrayOfSizeForFullCapa)
+{
+    char testString[8] = "iceoryx";
+    testString[7] = 'x'; // overwrite the 0 termination
+    string<7> testSubject(testString);
+    EXPECT_THAT(testSubject.size(), Eq(7));
+    EXPECT_THAT(testSubject.c_str(), StrEq("iceoryx"));
+}
+
 TYPED_TEST(stringTyped_test, UnsafeCharToStringConstrWithNullPtrResultsEmptyString)
 {
     using myString = typename TestFixture::stringType;
--- a/iceoryx_utils/test/moduletests/test_cxx_string.cpp
+++ b/iceoryx_utils/test/moduletests/test_cxx_string.cpp
@@ -361,6 +361,14 @@ TEST(String100, UnsafeCharToStringConstrIncludingNullCharWithCountResultsInSizeC
     EXPECT_THAT(testSubject.c_str(), StrEq("ice\0ryx"));
 }
 
+TEST(CharArrayAssignment, AssignCharArrayWithStringSizeLessThanArraySize)
+{
+    char testString[20] = "iceoryx";
+    string<20> testSubject(testString);
+    EXPECT_THAT(testSubject.size(), Eq(7));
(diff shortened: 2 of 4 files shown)


iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
```diff
@@ -119,9 +119,17 @@ inline string<Capacity>& string<Capacity>::operator=(const char (&rhs)[N]) noexc
...
-    m_rawstring[N - 1u] = '\0';
-    m_rawstringSize = N - 1u;
+
+    m_rawstringSize = strnlen(rhs, Capacity);
+    std::memcpy(m_rawstring, rhs, m_rawstringSize);
+    m_rawstring[m_rawstringSize] = '\0';
+
+    if (rhs[m_rawstringSize] != '\0')
+    {
+        std::cerr << "iox::cxx::string: Assignment of array which is not zero-terminated! Last value of array "
```
budrus, line 129: time for an error handling workshop. OK for now

budrus (approved):

iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
```diff
@@ -119,9 +119,17 @@ inline string<Capacity>& string<Capacity>::operator=(const char (&rhs)[N]) noexc
     {
         return *this;
     }
-    std::memcpy(m_rawstring, rhs, N - 1u);
-    m_rawstring[N - 1u] = '\0';
-    m_rawstringSize = N - 1u;
+
+    m_rawstringSize = strnlen(rhs, Capacity);
+    std::memcpy(m_rawstring, rhs, m_rawstringSize);
+    m_rawstring[m_rawstringSize] = '\0';
```
MatthiasKillat, line 125: Should be safe since m_rawstringSize is at most Capacity (due to strlen), so it is ok.

iceoryx_utils/test/moduletests/test_cxx_string.cpp
```diff
@@ -361,6 +361,23 @@ TEST(String100, UnsafeCharToStringConstrIncludingNullCharWithCountResultsInSizeC
     EXPECT_THAT(testSubject.c_str(), StrEq("ice\0ryx"));
 }
 
+TEST(CharArrayAssignment, AssignZeroTerminatedCharArrayOfSizeForFullCapa)
```
MatthiasKillat, line 364: The test name confuses me, but ok. OfSize? SizeCapacity? How about AssignZeroTerminatedCharArrayOfSizeFullCapacity. Does not really matter though, those names are bound to be strange.

MatthiasKillat (approved):

iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
```diff
@@ -119,9 +119,17 @@ inline string<Capacity>& string<Capacity>::operator=(const char (&rhs)[N]) noexc
     {
```
marthtz, line 119: I know you just fixed a bug but wouldn't it be clearer with just a single return and write
```
if (c_str() != rhs)
{
  // add null termination
}
return * this;
```

iceoryx_utils/test/moduletests/test_cxx_string.cpp
marthtz: Yeah, basically it's not telling what the size is, CharArrayOfSize???
I also prefer capacity.

iceoryx_utils/test/moduletests/test_cxx_string.cpp
```diff
@@ -361,6 +361,23 @@ TEST(String100, UnsafeCharToStringConstrIncludingNullCharWithCountResultsInSizeC
...
+    string<7> testSubject(testString);
+    EXPECT_THAT(testSubject.size(), Eq(7));
+    EXPECT_THAT(testSubject.c_str(), StrEq("iceoryx"));
+}
+
+TEST(CharArrayAssignment, AssignNonZeroTerminatedCharArrayOfSizeForFullCapa)
+{
+    char testString[8] = "iceoryx";
+    testString[7] = 'x'; // overwrite the 0 termination
+    string<7> testSubject(testString);
```
marthtz, line 376: You could also just assign it to a smaller string?

iceoryx_utils/test/moduletests/test_cxx_string.cpp
elBoberido: how about AssignZeroTerminatedCharArray**With**SizeForFullCapa?

iceoryx_utils/test/moduletests/test_cxx_string.cpp
elBoberido: that's not possible. This is a corner case with the knowledge of the implementation. The c'tor accepts char array with a size of `Capacity + 1` at max

iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
elBoberido: probably, but I wanted to change as less as possible

dkroenke (approved):

iceoryx_utils/test/moduletests/test_cxx_string.cpp
marthtz: Ok. Good to know!

iceoryx_utils/include/iceoryx_utils/internal/cxx/string.inl
marthtz: Yeah. Maybe this comment fits better in PR #253 .

iceoryx_utils/test/moduletests/test_cxx_string.cpp
marthtz: Better

MatthiasKillat: Fine with me.

MatthiasKillat (approved):

iceoryx_utils/test/moduletests/test_cxx_string.cpp
elBoberido: maybe @FerdinandSpitzschnueffler has an even better idea. This could probably be renamed in #253

PR #270, opened 2020-09-04
iox #243 fix posix timer

MatthiasKillat:
These are small fixes that should address the segmentation fault in the destructor and several races that may lead to lost callbacks. 

I also tried to simplify some code and protect it against certain races, but due to time constraints and the current design this could only be done in a limited way.

There are still potential problems in the timer and a redesign is advised, potentially including a variant that is not using a heavyweight callback mechanism which creates threads for each callback.


iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -254,21 +306,26 @@ cxx::expected<TimerError> Timer::OsTimer::start(const RunMode runMode, const Cat
...
     // Set the timer
     auto result = cxx::makeSmartC(
         timer_settime, cxx::ReturnMode::PRE_DEFINED_ERROR_CODE, {-1}, {}, m_timerId, 0, &interval, nullptr);
 
     if (result.hasErrors())
     {
+        // undo optimistically setting m_isTimerActive before
+        // not entirely safe against concurrent starts, we cannot detect these in this way
+        // needs extensive refactoring to achieve this (e.g. start and stop with mutex protection?)
+        handle.m_isTimerActive.exchange(wasActive, std::memory_order_relaxed);
```
MatthiasKillat, line 325: Could be a store. But we may miss concurrent timer starts/stops which have set it to another value. Maybe use a compare exchange.

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -76,52 +76,93 @@ void Timer::OsTimer::callbackHelper(sigval data)
         return;
     }
 
-    handle.m_timerInvocationCounter.fetch_add(1u, std::memory_order_relaxed);
```
MatthiasKillat, line 79: Logic change in this method together with start is supposed to address lost callbacks invocations.

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -208,6 +254,10 @@ Timer::OsTimer::~OsTimer() noexcept
     {
         stop();
 
+        // do not delete the timer while the callback is running, it could access the timer which is about to be deleted
```
MatthiasKillat, line 257: Should address the segfault in the destructor, at least it could not be reproduced any more and it definitely was a problem, but there may be additional causes.

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -254,21 +306,26 @@ cxx::expected<TimerError> Timer::OsTimer::start(const RunMode runMode, const Cat
...
+        // undo optimistically setting m_isTimerActive before
+        // not entirely safe against concurrent starts, we cannot detect these in this way
+        // needs extensive refactoring to achieve this (e.g. start and stop with mutex protection?)
+        handle.m_isTimerActive.exchange(wasActive, std::memory_order_relaxed);
         return createErrorFromErrno(result.getErrNum());
     }
 
-    auto& handle = OsTimer::s_callbackHandlePool[m_callbackHandleIndex];
-
-    handle.m_catchUpPolicy = catchUpPolicy;
```
elBoberido, line 268: this has either to be set here, maybe made atomic, or removed from the function signature

iceoryx_utils/source/posix_wrapper/timer.cpp
elBoberido: I think this was never designed to have concurrent start/stops. If it's done this way, you have the problem you described a while ago with the callback beeing executed before this flag is set and then the callback doesn't do anything because the flag is still false

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -76,52 +76,93 @@ void Timer::OsTimer::callbackHelper(sigval data)
...
-        }
+            invocationCounter = handle.m_timerInvocationCounter.exchange(0u, std::memory_order_acq_rel);
+
+            // did someone else execute the callback for us?
+            if (invocationCounter != 0u)
+            {
+                handle.m_timer->executeCallback();
+            }
+
+            handle.m_callbackIsAboutToBeExecuted.clear(std::memory_order_release);
```
elBoberido, line 136: from here up to the point where the flag is set again, there is a window where the CatchUpPolicy::TERMINATE will be handled wrong

dkroenke:
Please merge latest master into your branch for coverage update.

iceoryx_utils/source/posix_wrapper/timer.cpp
marthtz: If I remember right, one problem is when a timer is modified (especially restarted) within its callback. At least we should mentioned it somewhere that it might be risky to do that?!

MatthiasKillat: Risky is an understatement. It was not race-free to do this before. I think I improved this issue (somewhat), but there will be probably more problems. And as @marthtz says it was used like this internally, hence the fixes.

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -76,52 +76,93 @@ void Timer::OsTimer::callbackHelper(sigval data)
...
         {
-            if (handle.m_catchUpPolicy == CatchUpPolicy::TERMINATE)
+            // prohibits other threads from entering the flag protected region
+            handle.m_callbackIsAboutToBeExecuted.test_and_set(std::memory_order_acq_rel);
+
+            if (handle.m_timer == nullptr)
             {
-                errorHandler(Error::kPOSIX_TIMER__CALLBACK_RUNTIME_EXCEEDS_RETRIGGER_TIME);
+                errorHandler(Error::kPOSIX_TIMER__INCONSISTENT_STATE);
+                return;
```
MatthiasKillat, line 110: The complaints of the Codecov are valid (i.e. it is true), but it is also hard to inject the faults leading into those branches. We may need a general fault injection strategy for scenarios like this. Part of the error handling strategy I guess.

iceoryx_utils/source/posix_wrapper/timer.cpp
```diff
@@ -218,8 +268,10 @@ Timer::OsTimer::~OsTimer() noexcept
 
         m_timerId = INVALID_TIMER_ID;
 
-        auto& callbackHandle = OsTimer::s_callbackHandlePool[m_callbackHandleIndex];
-        std::lock_guard<std::mutex> lock(callbackHandle.m_accessMutex);
+        // note that some of those relaxed loads and stores could be potential problems since they might
+        // be reordered within certain constraints which may lead to subtle but serious logical errors due to races
+        // (especially on architectures with weaker memory model)
+        // this cannot be fixed without detailed analysis and a redesign is advised
         callbackHandle.m_inUse.store(false, std::memory_order_relaxed);
```
MatthiasKillat, line 275: seq_consistent

mossmaurice (approved):
Approved under the condition that code will be deprecated with the next release and removed with the one after. A completely re-written, new version of the `posix::timer` is advised to use signals and handle its own thread for the sake of simplicity.

elBoberido (approved):

Issue #276, opened 2020-09-09
Unable to build with colcon


evshary:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.5.0

**Observed result or behaviour:**
Unable to build colcon build with the following steps: https://github.com/eclipse/iceoryx/blob/master/doc/installation-guide.md#build-with-colcon

**Expected result or behaviour:**
Able to build successfully.

The error message is like that:

```
--- stderr: example_icedelivery_on_c
CMake Error at CMakeLists.txt:8 (find_package):
  Could not find a package configuration file provided by "iceoryx_binding_c"
  with any of the following names:

    iceoryx_binding_cConfig.cmake
    iceoryx_binding_c-config.cmake

  Add the installation prefix of "iceoryx_binding_c" to CMAKE_PREFIX_PATH or
  set "iceoryx_binding_c_DIR" to a directory containing one of the above
  files.  If "iceoryx_binding_c" provides a separate development package or
  SDK, be sure it has been installed.
```

I think the root cause is the existence of COLCON_IGNORE in iceoryx_binding_c.
This will cause colcon ignore this package.
However, this package is used by example_icedelivery_on_c in iceoryx_examples.
IMHO, I think it's OK to remove the COLCON_IGNORE in iceoryx_binding_c.

PR #277, opened 2020-09-09
iox-#276 Remove COLCON_IGNORE and able to build with colcon.

evshary:
As said in #276, remove COLCON_IGNORE in iceoryx_binding_c to make colcon work.


elfenpiff (approved):

elBoberido (approved):

elBoberido:
@evshary keep in mind, this is not yet working with RouDi, since the new building blocks are not yet integrated into RouDi. This will happen the next weeks

evshary:
> @evshary keep in mind, this is not yet working with RouDi, since the new building blocks are not yet integrated into RouDi. This will happen the next weeks when #25 (last point in list) is finished

Thanks for your kind reminders. I'm looking forward to #25. 
Not only the C API binding but also the multiple publisher. :-)

PR #281, opened 2020-09-10
Iox #25 add new pubsub ports to roudi

marthtz:
- Added new publisher/subscriber ports to RouDi (in parallel, old sender/receiver still in place for now)
- Marked all (old) sender/receiver as `@deprecated #25` so they can be deleted in follow-up PR
- Marked some code with `@todo #25` as it has to be fixed before RouDi can actually use new pub/sub ports (e.g. introspection)
- Follow-up PR will flip the switch - use pub/sub ports and delete sender/receiver ports

Open questions:
- Isn't `iceoryx_port_pool` obsolete now? Shouldn't`port_pool`  be enough? @elBoberido 
- I'm not quite happy using a preprocessor define to select the roudi producer type (`ENABLE_1_TO_N_COMMUNICATION_ONLY`). Please speak out if you have a better idea.


iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -41,71 +44,106 @@ namespace iox
...
-                                                      const RunnableName_t& runnable = "");
+                                                      const RunnableName_t& runnable = "") noexcept;
 
-    popo::ApplicationPortData* acquireApplicationPortData(const ProcessName_t& processName);
+    popo::ApplicationPortData* acquireApplicationPortData(const ProcessName_t& processName) noexcept;
 
-    runtime::RunnableData* acquireRunnableData(const ProcessName_t& process, const RunnableName_t& runnable);
+    runtime::RunnableData* acquireRunnableData(const ProcessName_t& process, const RunnableName_t& runnable) noexcept;
 
     cxx::expected<popo::ConditionVariableData*, PortPoolError> acquireConditionVariableData();
```
mossmaurice, line 100: `noexcept` missing?

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
```diff
@@ -49,39 +55,60 @@ class PortPool
...
                   const std::string& applicationName,
                   const mepoo::MemoryInfo& memoryInfo = mepoo::MemoryInfo()) noexcept = 0;
 
+    /// @deprecated #25
     virtual cxx::expected<ReceiverPortType::MemberType_t*, PortPoolError>
     addReceiverPort(const capro::ServiceDescription& serviceDescription,
                     const std::string& applicationName,
                     const mepoo::MemoryInfo& memoryInfo = mepoo::MemoryInfo()) noexcept = 0;
 
+    virtual cxx::expected<PublisherPortRouDiType::MemberType_t*, PortPoolError>
```
mossmaurice, line 82: I think there's no need to make this method `virtual`, same for the other one

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
```diff
@@ -49,39 +55,60 @@ class PortPool
...
-    cxx::expected<runtime::RunnableData*, PortPoolError>
-    addRunnableData(const ProcessName_t& process,
-                    const RunnableName_t& runnable,
-                    const uint64_t runnableDeviceIdentifier) noexcept;
+    cxx::expected<runtime::RunnableData*, PortPoolError> addRunnableData(
+        const ProcessName_t& process, const RunnableName_t& runnable, const uint64_t runnableDeviceIdentifier) noexcept;
 
-    cxx::expected<popo::ConditionVariableData*, PortPoolError>
-    addConditionVariableData();
+    cxx::expected<popo::ConditionVariableData*, PortPoolError> addConditionVariableData();
```
mossmaurice, line 104: add `noexcept`

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -573,10 +746,73 @@ ReceiverPortType::MemberType_t* PortManager::acquireReceiverPortData(const capro
...
+    //         {
+    //             errorHandler(Error::kPOSH__PORT_MANAGER_PUBLISHERPORT_NOT_UNIQUE, nullptr, ErrorLevel::MODERATE);
+    //             return cxx::error<PortPoolError>(PortPoolError::UNIQUE_PUBLISHER_PORT_ALREADY_EXISTS);
+    //         }
+    //         else
+    //         {
+    //             break;
+    //         }
+    //     }
+    // }
```
mossmaurice, line 779: Not a big fan of this huge out-commented section. Do you promise to fix it asap in a follow-up PR? ;-)

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -552,6 +527,101 @@ void ProcessManager::addSenderForProcess(const ProcessName_t& name,
...
+        /// pursuing this further could lead to a separate management entity for ports
+        /// which could support queries like: find all ports with a given service or some other
+        /// specific attribute (to allow efficient and well encapsulated lookup)
+
+        auto maybeSubscriber =
+            m_portManager.acquireSubscriberPortData(service, historyRequest, name, runnable, portConfigInfo);
+
+        if (!maybeSubscriber.has_error())
+        {
+            // send ReceiverPort to app as a serialized relative pointer
```
mossmaurice, line 555: copy & waste ;-) The term "Receiver" is deprecated

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -552,6 +527,101 @@ void ProcessManager::addSenderForProcess(const ProcessName_t& name,
...
+    if (nullptr != process)
+    {
+        // create a SubscriberPort
+
+        /// @todo: it might be useful to encapsulate this into some kind of port factory
+        /// (which maybe contains a m_shmMgr)
+        /// main goal would be to isolate the port creation logic as it becomes more complex
+        /// pursuing this further could lead to a separate management entity for ports
+        /// which could support queries like: find all ports with a given service or some other
+        /// specific attribute (to allow efficient and well encapsulated lookup)
```
mossmaurice, line 548: copy & waste, I don't think this is valid anymore? Check with @MatthiasKillat

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -324,6 +324,106 @@ TEST_F(PoshRuntime_test, GetMiddlewareReceiverReceiverlistOverflow)
...
+{
+    const auto publisherPort = m_runtime->getMiddlewarePublisher(
+        iox::capro::ServiceDescription(99U, 1U, 20U), 0U, m_runnableName, iox::runtime::PortConfigInfo(11U, 22U, 33U));
+
+    ASSERT_NE(nullptr, publisherPort);
+    EXPECT_EQ(iox::capro::ServiceDescription(99U, 1U, 20U), publisherPort->m_serviceDescription);
+    /// @todo #25 historyCapacity no member?
+    /// @todo #25 memoryInfo not used in publisher port right now
+    // EXPECT_EQ(22u, publisherPort->m_memoryInfo.deviceId);
+    // EXPECT_EQ(33u, publisherPort->m_memoryInfo.memoryType);
```
mossmaurice, line 337: Please don't check in dead code ;-) I don't understand what's the issue here?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -324,6 +324,106 @@ TEST_F(PoshRuntime_test, GetMiddlewareReceiverReceiverlistOverflow)
...
+
+TEST_F(PoshRuntime_test, getMiddlewarePublisherDefaultArgs)
+{
+    const auto publisherPort = m_runtime->getMiddlewarePublisher(iox::capro::ServiceDescription(99U, 1U, 20U));
+
+    ASSERT_NE(nullptr, publisherPort);
+
+    /// @todo #25 memoryInfo not used in publisher port right now
+    // EXPECT_EQ(0u, publisherPort->m_memoryInfo.deviceId);
+    // EXPECT_EQ(0u, publisherPort->m_memoryInfo.memoryType);
```
mossmaurice, line 348: Same :-)

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -324,6 +324,106 @@ TEST_F(PoshRuntime_test, GetMiddlewareReceiverReceiverlistOverflow)
...
+        iox::capro::ServiceDescription(99U, 1U, 20U), 0U, m_runnableName, iox::runtime::PortConfigInfo(11U, 22U, 33U));
+
+    ASSERT_NE(nullptr, subscriberPort);
+    EXPECT_EQ(iox::capro::ServiceDescription(99U, 1U, 20U), subscriberPort->m_serviceDescription);
+    EXPECT_EQ(0U, subscriberPort->m_historyRequest);
+
+    /// @todo #25 memoryInfo not used in subscriber port right now
+    // EXPECT_EQ(22u, subscriberPort->> m_memoryInfo.deviceId);
+    // EXPECT_EQ(33u, subscriberPort->m_memoryInfo.memoryType);
+}
```
mossmaurice, line 389: Same same

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -324,6 +324,106 @@ TEST_F(PoshRuntime_test, GetMiddlewareReceiverReceiverlistOverflow)
...
+
+TEST_F(PoshRuntime_test, GetMiddlewareSubscriberDefaultArgs)
+{
+    auto subscriberPort = m_runtime->getMiddlewareSubscriber(iox::capro::ServiceDescription(99U, 1U, 20U));
+
+    ASSERT_NE(nullptr, subscriberPort);
+
+    /// @todo #25 memoryInfo not used in subscriber port right now
+    // EXPECT_EQ(0u, subscriberPort->m_memoryInfo.deviceId);
+    // EXPECT_EQ(0u, subscriberPort->m_memoryInfo.memoryType);
```
mossmaurice, line 399: Same same

mossmaurice (changes requested):

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -30,19 +30,35 @@ template <typename>
...
-using SenderPortType = iox::popo::SenderPort;
-using ReceiverPortType = iox::popo::ReceiverPort;
+using SenderPortType = iox::popo::SenderPort;     /// @deprecated #25
+using ReceiverPortType = iox::popo::ReceiverPort; /// @deprecated #25
+using PublisherPortRouDiType = iox::popo::PublisherPortRouDi;
+using PublisherPortUserType = iox::popo::PublisherPortUser;
+using SubscriberPortUserType = iox::popo::SubscriberPortUser;
 using UniquePortId = popo::TypedUniqueId<popo::BasePortData>;
 
+#if defined(ENABLE_1_TO_N_COMMUNICATION_ONLY)
```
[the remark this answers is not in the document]
marthtz: That's the decision to be made. Right now as implemented n-m is default. But we can change that. Can you elaborate on how this is handled in other scripts? So maybe it needs to be adjusted here as well. Currently, as long as you don't specify anything n-m is active.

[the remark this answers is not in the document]
marthtz: See above. We agreed to n-m as default when we discussed it.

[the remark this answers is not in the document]
mossmaurice: But we should offer the possibility to switch it off, right? That part is missing.

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: Remember, when we discussed it we agreed to add `@todo #25` to everything that'll be cleaned-up by the follow-up PR, which will be done right after this one is merged. So it will stay just a couple of days.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: See above regarding `@todo #25` tags.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: See above regarding `@todo #25` tags.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: See above regarding `@todo #25` tags.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: See above regarding `@todo #25` tags.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: He won't come back until next week. Either delay this PR or add `#25` as marker.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
mossmaurice: What's the blocker to clarify this?

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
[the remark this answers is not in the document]
marthtz: You can have either 1:m or n:m.
n:m is default and you can switch by defining `ENABLE_1_TO_N_COMMUNICATION_ONLY`.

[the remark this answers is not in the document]
marthtz: Done.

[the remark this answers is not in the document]
marthtz: Done.

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
marthtz: Fixed

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
marthtz: Fixed

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: Fixed

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: Fixed

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
marthtz: Fixed

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Fixed

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
marthtz: missed

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: missed

mossmaurice (dismissed):
LGTM

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -244,6 +248,157 @@ ReceiverPortType::MemberType_t* PoshRuntime::requestReceiverFromRoudi(const MqMe
...
+                LogError() << "No valid publisher port received from RouDi.";
+                errorHandler(Error::kPOSH__RUNTIME_NO_VALID_PUBLISHER_RECEIVED, nullptr, iox::ErrorLevel::MODERATE);
+                return cxx::error<MqMessageErrorType>(stringToMqMessageErrorType(mqMessage2.c_str()));
+            }
+        }
+    }
+
+    LogError() << "Wrong response from message queue :'" << receiveBuffer.getMessage() << "'";
+    errorHandler(Error::kPOSH__RUNTIME_WRONG_MESSAGE_QUEUE_RESPONSE, nullptr, iox::ErrorLevel::SEVERE);
+    return cxx::success<PublisherPortUserType::MemberType_t*>(nullptr);
```
elBoberido, line 325: since one line above this the error handler is called, I would assume that this is an error, not a success

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -244,6 +248,157 @@ ReceiverPortType::MemberType_t* PoshRuntime::requestReceiverFromRoudi(const MqMe
...
+                LogError() << "No valid subscriber port received from RouDi.";
+                errorHandler(Error::kPOSH__RUNTIME_NO_VALID_SUBSCRIBER_RECEIVED, nullptr, iox::ErrorLevel::MODERATE);
+                return cxx::error<MqMessageErrorType>(stringToMqMessageErrorType(mqMessage2.c_str()));
+            }
+        }
+    }
+
+    LogError() << "Wrong response from message queue";
+    errorHandler(Error::kPOSH__RUNTIME_WRONG_MESSAGE_QUEUE_RESPONSE, nullptr, iox::ErrorLevel::SEVERE);
+    return cxx::success<SubscriberPortUserType::MemberType_t*>(nullptr);
```
elBoberido, line 399: again, shouldn't this be error instead of success?

elBoberido (commented):
it's almost impossible to review port_manager.cpp with the codecov comments every second line. Is there a possibility to hide that?

mossmaurice:
> 
> 
> it's almost impossible to review port_manager.cpp with the codecov comments every second line. Is there a possibility to hide that?

Yep, just click on the three dots in the top right corner and deselect "Show annotations"

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -166,13 +168,89 @@ void PortManager::handleReceiverPorts()
...
 }
 
-void PortManager::handleInterfaces()
+void PortManager::handlePublisherPorts() noexcept
+{
+    // get the changes of publisher port offer state
+    for (auto publisherPortData : m_portPool->getPublisherPortDataList())
+    {
+        PublisherPortRouDiType publisherPort(publisherPortData);
+        auto maybeCaproMessage = publisherPort.getCaProMessage();
```
elBoberido, line 177: have you thought about using this pattern
```
publisherPort.getCaProMessage(). and_then(...).or_else(...)
```

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -166,13 +168,89 @@ void PortManager::handleReceiverPorts()
...
+
+            if (!sendToAllMatchingPublisherPorts(caproMessage, subscriberPort))
+            {
+                LogDebug() << "capro::SUB/UNSUB, no matching sender!!";
+                capro::CaproMessage nackMessage(capro::CaproMessageType::NACK,
+                                                subscriberPort.getCaProServiceDescription());
+                subscriberPort.dispatchCaProMessage(nackMessage);
+            }
+        }
+
```
elBoberido, line 238: shouldn't there be an else branch like with `PortManager::handlePublisherPorts`

iceoryx_posh/source/roudi/iceoryx_port_pool.cpp
```diff
@@ -72,15 +76,83 @@ IceOryxPortPool::addReceiverPort(const capro::ServiceDescription& serviceDescrip
...
+
+cxx::expected<SubscriberPortProducerType::MemberType_t*, PortPoolError>
+IceOryxPortPool::addSubscriberPort(const capro::ServiceDescription& serviceDescription,
+                                   const uint64_t& historyRequest,
+                                   const ProcessName_t& applicationName,
+                                   const mepoo::MemoryInfo& memoryInfo) noexcept
+{
+    if (m_portPoolData->m_subscriberPortMembers.hasFreeSpace())
+    {
+#if defined(ENABLE_1_TO_N_COMMUNICATION_ONLY)
```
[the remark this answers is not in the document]
budrus: We currently do not think of mixing. This could get complicated and we normally live in one world

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -365,38 +445,84 @@ void PortManager::sendToAllMatchingReceiverPorts(const capro::CaproMessage& mess
...
 {
-    int32_t numberOfReceiverPorts{0};
-    int32_t numberOfConnectedReceiverPorts{0};
-    for (auto receiverPortData : m_portPool->receiverPortDataList())
+    for (auto subscriberPortData : m_portPool->getSubscriberPortDataList())
     {
-        ReceiverPortType receiver(receiverPortData);
-        if (receiver.getProcessName() == processName)
+        SubscriberPortProducerType subscriberPort(subscriberPortData);
+        if (subscriberPort.getCaProServiceDescription() == publisherSource.getCaProServiceDescription())
```
elBoberido, line 481: somehow I always think this checks if getting the service description is equal, not the service description itself. That's why I dislike the `get` prefix 😉

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -166,13 +168,89 @@ void PortManager::handleReceiverPorts()
...
+        auto returnedCaproMessage = subscriberPort.getCaProMessage();
+        if (returnedCaproMessage.has_value())
+        {
+            auto& caproMessage = returnedCaproMessage.value();
+
+            m_portIntrospection.reportMessage(caproMessage);
+
+            if (!sendToAllMatchingPublisherPorts(caproMessage, subscriberPort))
+            {
+                LogDebug() << "capro::SUB/UNSUB, no matching sender!!";
```
budrus, line 232: no matching publisher

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -365,38 +445,84 @@ void PortManager::sendToAllMatchingReceiverPorts(const capro::CaproMessage& mess
...
         {
-            numberOfReceiverPorts++;
-            numberOfConnectedReceiverPorts += receiver.isSubscribed() ? 1 : 0;
+            auto subscriberResponse = subscriberPort.dispatchCaProMessage(message);
+
+            // if the subscribers react on the change, process it immediately on publisher side
+            if (subscriberResponse.has_value())
+            {
+                // we only expect reaction on OFFER
+                assert(capro::CaproMessageType::OFFER == message.m_type);
```
elBoberido, line 489: since `cxx::Ensures` is used below, how about using `cxx::Expects`?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -487,12 +615,56 @@ void PortManager::destroyReceiverPort(ReceiverPortType::MemberType_t* const rece
...
+    m_portIntrospection.reportMessage(message);
+
+    sendToAllMatchingPublisherPorts(message, subscriberPort);
+
+    /// @todo #25 Fix introspection
+    // m_portIntrospection.removeSubscriber(subscriberPort.getProcessName(), serviceDescription);
+
+    // delete subscriber impl from list after unsubscribe was processed
+    m_portPool->removeSubscriberPort(subscriberPortData);
+    LogDebug() << "Destroyed SubscriberPortImpl";
```
budrus, line 659: Impl is legacy, call it Data

budrus (changes requested):

iceoryx_posh/source/roudi/iceoryx_port_pool.cpp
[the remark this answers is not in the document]
elBoberido: I didn't look into the implementation, but shouldn't the variant queue just handle this?

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: Missed that one. Done.

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: Missed it. Done.

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: Yes, I thought about it. Building blocks mostly use the `maybe` pattern. That's why I used it here as well. Is the `and_then(...)` pattern to way we want to go with iceoryx?

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: What is it supposed to do in case of the else branch? It used to do nothing. Should it dispatch an `ok` message?

elBoberido:
> * Isn't `iceoryx_port_pool` obsolete now? Shouldn't`port_pool`  be enough? @elBoberido

I don't think we have users who extends RouDi anymore, therefore this could be merged

> * I'm not quite happy using a preprocessor define to select the roudi producer type (`ENABLE_1_TO_N_COMMUNICATION_ONLY`). Please speak out if you have a better idea.

Yes, if someone is not using cmakes find_package, then this approach will break. We already had this with some users regarding the large deployment. An idea would be to add this to `iox_posh_deployment.hpp.in`, e.g.
```
using SubscriberPortProducerType = @IOX_SUBSCRIBER_PORT_PRODUCER_TYPE@;
```
and in the CMakeLists.txt
```
set(IOX_SUBSCRIBER_PORT_PRODUCER_TYPE iox::popo::SubscriberPortSingleProducer)
```
If there is code which is different depending on the type, an additional define could be used, e.g. `define IOX_SUBSCRIBER_PORT_PRODUCER_TYPE_N2M false;` or something similar. Maybe there are better solutions, e.g. SFINEA.

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: if possible yes, since it makes the code in some situations cleaner

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: I would assume it's an error, like with `PortManager::handlePublisherPorts`. @budrus, what do you think?
Keep in mind, this was quite old code where the error handling was not optimal

marthtz:
> 
> I don't think we have users who extends RouDi anymore, therefore this could be merged
> 
Thanks. Will be done in follow-up PR.

> 
>  An idea would be to add this to `iox_posh_deployment.hpp.in`, e.g.
> ```
> using SubscriberPortProducerType = @IOX_SUBSCRIBER_PORT_PRODUCER_TYPE@;
> ```
> and in the CMakeLists.txt
> ```
> set(IOX_SUBSCRIBER_PORT_PRODUCER_TYPE iox::popo::SubscriberPortSingleProducer)
> ```
> If there is code which is different depending on the type, an additional define could be used, e.g. `define IOX_SUBSCRIBER_PORT_PRODUCER_TYPE_N2M false;` or something similar. Maybe there are better solutions, e.g. SFINEA.
There's basically one other method  `addSubscriberPort()` that's different, where the port is created. This could also be done in a different way as implicitly know the queue type for the producer type. Any idea for that? Then it could be moved to deployment and is at least in-line with these other deployed consts.

iceoryx_posh/source/roudi/port_manager.cpp
marthtz: Ok. I'm in!

iceoryx_posh/source/roudi/iceoryx_port_pool.cpp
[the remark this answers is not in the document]
budrus: You can have variant queues for single and multi producer. But I see another challenge. When you create another publisher on the same topic, how do you know that all subscribers have the multi pusher queue. Would you need a per topic configuration with the information whether multi-producer or not. It's getting complicated

mossmaurice (dismissed):

budrus (approved):

mossmaurice (approved):

elBoberido:
> There's basically one other method `addSubscriberPort()` that's different, where the port is created. This could also be done in a different way as implicitly know the queue type for the producer type. Any idea for that? Then it could be moved to deployment and is at least in-line with these other deployed consts.

This could probably be done with `std::enable_if` and `std::is_same`

Issue #286, opened 2020-09-15
Clang formatting has not been applied to entire repo


ithier:
Clang formatting has not been applied consistently to the repo as new code has been added.
Running clang format now changes code all over the repo and thus pollutes new feature branches with unrelated changes.
Clang format needs to be run across the entire repository and merged in its own PR to fix this.

PR #288, opened 2020-09-15
iox-#286 Clang format.

elfenpiff:
contribution from @ithier

### DIFF
--- a/iceoryx_posh/source/mepoo/memory_info.cpp
+++ b/iceoryx_posh/source/mepoo/memory_info.cpp
@@ -23,5 +23,5 @@ MemoryInfo::MemoryInfo(uint32_t deviceId, uint32_t memoryType)
     , memoryType(memoryType)
 {
 }
-}
-}
+} // namespace mepoo
+} // namespace iox
--- a/iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/cyclic_index.inl
+++ b/iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/cyclic_index.inl
@@ -82,4 +82,4 @@ bool CyclicIndex<CycleLength, ValueType>::isOneCycleBehind(const CyclicIndex& ot
     return (thisCycle + 1 == otherCycle);
 }
 } // namespace concurrent
-}
+} // namespace iox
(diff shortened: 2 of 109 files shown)


elBoberido (approved):

iceoryx_posh/source/mepoo/memory_info.cpp
```diff
@@ -23,5 +23,5 @@ MemoryInfo::MemoryInfo(uint32_t deviceId, uint32_t memoryType)
     , memoryType(memoryType)
 {
 }
-}
-}
\ No newline at end of file
+} // namespace mepoo
+} // namespace iox
```
mossmaurice, line 27: New line missing?

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/cyclic_index.inl
```diff
@@ -82,4 +82,4 @@ bool CyclicIndex<CycleLength, ValueType>::isOneCycleBehind(const CyclicIndex& ot
     return (thisCycle + 1 == otherCycle);
 }
 } // namespace concurrent
-}
\ No newline at end of file
+} // namespace iox
```
mossmaurice, line 85: New line missing?

mossmaurice (approved):

dkroenke (approved):

iceoryx_posh/source/mepoo/memory_info.cpp
elfenpiff: We do not need this since C++11. It is a requirement from the old C world which was abandoned with C++11.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/cyclic_index.inl
elfenpiff: We do not need this since C++11. It is a requirement from the old C world which was abandoned with C++11.

iceoryx_posh/source/mepoo/memory_info.cpp
dkroenke: This comes from the POSIX standard where several tools like cat or wc threat only files with a newline at the end correctly. Otherwise it can lead to strange behaviour when the last line of the file is missing while processing it. I would vote for following this rule.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/cyclic_index.inl
dkroenke: See comment above

Issue #130, opened 2020-06-08
Logger and error handling strategy


budrus:
## Brief feature description

Having a clear strategy for logging and error handling 

## Detailed information

In some very low layer modules of iceoryx_utils, cerr and terminate are used and not the logger and error handler. A clear strategy is needed whether different approaches can be used and if so which criteria apply.

MatthiasKillat:
While we have error-handling.md which summarizes our current error handling strategy and mentions open points (cf. https://github.com/eclipse/iceoryx/pull/343)

How do specifically deal with errors in iceoryx::utils is not addressed in much detail there (we have yet to find a clear strategy for this).

@budrus 
I propose to leave this issue open but maybe rename the issue to Logger and error handling strategy in iceoryx::utils. 

Even if we do not want to do this, we have not yet verified that all methods are indeed noexcept.

budrus:
@MatthiasKillat OK, but this is then no more a topic for the 1.0 release. I will update the description

PR #292, opened 2020-09-16
Iox #25 implement unique publisher detection for new building blocks

mossmaurice:
* Addresses one `@todo` of #281
* Remove class `IceOryxPortPool`


mossmaurice:
I suppose an additional test run for the 1:n variant makes sense? What do you think @elBoberido?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -764,7 +764,23 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
                                       const RunnableName_t& runnable [[gnu::unused]], // @todo #25 Fix introspection
                                       const PortConfigInfo& portConfigInfo) noexcept
 {
-    /// @todo #25 Add unique check for single producer type - see acquireSenderPortData()
```
elfenpiff, line 767: I would prefer having a boolean member in port manager which is called like `m_hasMultiPublisherSupport` instead of working with macros.
Why are we using macros here?

Or it could be possible to store such a boolean or maybe enum in `iceoryx_posh_types.hpp` as a constexpr.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -366,6 +366,20 @@ TEST_F(PoshRuntime_test, getMiddlewarePublisherPublisherlistOverflow)
     EXPECT_TRUE(publisherlistOverflowDetected);
 }
 
+#if defined(RESTRICT_TO_1_TO_N_COMMUNICATION)
```
elfenpiff, line 369: Please do not use macros here. Is it possible to use the roudi environment and create a port manager which supports and one which does not support n to m communication?

elfenpiff (changes requested):
Please do not use macro definitions.

Add a enum or boolean member to ` PortManager` which ensures the behavior instead of using a macro definition.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -764,7 +764,23 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
...
+        popo::PublisherPortRouDi publisherPort(publisherPortData);
+        if (service == publisherPort.getCaProServiceDescription())
+        {
+            LogWarn() << "Process '" << processName
+                      << "' tried to register an unique PublisherPort which is already used by '"
+                      << publisherPortData->m_processName << "' with service '"
+                      << service.operator cxx::Serialization().toString() << "'.";
+
+            errorHandler(Error::kPOSH__PORT_MANAGER_PUBLISHERPORT_NOT_UNIQUE, nullptr, ErrorLevel::MODERATE);
+            return cxx::error<PortPoolError>(PortPoolError::UNIQUE_PUBLISHER_PORT_ALREADY_EXISTS);
```
marthtz, line 780: Is it necessary to call the error handler here? We already return an error that is sent via mq and finally evaluated in the runtime.
I guess our error handling needs to be revised anyway.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
mossmaurice: Yeah, I hate macros, too 😃 In #281 we decided to go forward with the macro approach and replace it later on. I'm happy to do that as part of this pull request.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: I suppose that's part of #130

budrus: This will be an error reported by RouDi, the one sent over the message queue will be in the user application. So we somehow log here that an application does not act according to the contract. I would like to have the moderate error here

iceoryx_posh/source/roudi/port_manager.cpp
budrus: This has to switch the used types during compile time. I'm wondering if we can do it similar to the CMake parameters for MAX_PUBLISHERS etc. What do you think @elfenpiff @elBoberido

elBoberido: yes, I already talked with @mossmaurice about that. He is working on it

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
elfenpiff: @mossmaurice I fixed the macro approach in the pull request: https://github.com/eclipse/iceoryx/pull/295

mossmaurice: I already fixed it on my local branch in a nicer SFINAE way ;-) Could you revert your changes?

elfenpiff: @mossmaurice No problem, I will do that on monday!

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -150,14 +150,52 @@ class PortManager
     void addEntryToServiceRegistry(const capro::IdString& service, const capro::IdString& instance) noexcept;
     void removeEntryFromServiceRegistry(const capro::IdString& service, const capro::IdString& instance) noexcept;
 
+    // @todo move this to cxx type_traits and create inl
```
mossmaurice, line 153: Will carve out `cxx/type_traits.hpp` from #291 and then address this todo

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
```diff
@@ -48,49 +51,87 @@ enum class PortPoolError : uint8_t
...
-    virtual cxx::expected<SubscriberPortProducerType::MemberType_t*, PortPoolError>
+    cxx::expected<SubscriberPortProducerType::MemberType_t*, PortPoolError>
     addSubscriberPort(const capro::ServiceDescription& serviceDescription,
                       const uint64_t& historyRequest,
                       const ProcessName_t& applicationName,
-                      const mepoo::MemoryInfo& memoryInfo = mepoo::MemoryInfo()) noexcept = 0;
+                      const mepoo::MemoryInfo& memoryInfo = mepoo::MemoryInfo()) noexcept;
+
+
+    // @todo move this to cxx type_traits and create inl
```
mossmaurice, line 98: Will carve out cxx/type_traits.hpp from #291 and then address this todo

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: Remove macros

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -366,6 +366,36 @@ TEST_F(PoshRuntime_test, getMiddlewarePublisherPublisherlistOverflow)
...
+
+    const auto publisherPort1 = m_runtime->getMiddlewarePublisher(
+        sameServiceDescription, 0U, m_runnableName, iox::runtime::PortConfigInfo(11U, 22U, 33U));
+
+    const auto publisherPort2 = m_runtime->getMiddlewarePublisher(
+        sameServiceDescription, 0U, m_runnableName, iox::runtime::PortConfigInfo(11U, 22U, 33U));
+
+    ASSERT_NE(nullptr, publisherPort1);
+
+    if (typeid(policy) == typeid(iox::build::OneToManyPolicy))
```
mossmaurice, line 389: Unsure whether this is a nice solution. Any ideas @elBoberido?

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
mossmaurice: Removed the macros

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
mossmaurice: Fixed with e5c0467. Thanks @elfenpiff

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -150,14 +151,22 @@ class PortManager
     void addEntryToServiceRegistry(const capro::IdString& service, const capro::IdString& instance) noexcept;
     void removeEntryFromServiceRegistry(const capro::IdString& service, const capro::IdString& instance) noexcept;
 
+    template <typename T, cxx::enable_if_t<std::is_same<T, iox::build::OneToManyPolicy>::value>* = nullptr>
+    bool hasDuplicatePublisher(const capro::ServiceDescription& service, const ProcessName_t& processName) noexcept;
```
elfenpiff, line 155: I think this method should be const.

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
```diff
@@ -0,0 +1,53 @@
...
+                      << service.operator cxx::Serialization().toString() << "'.";
+            errorHandler(Error::kPOSH__PORT_MANAGER_PUBLISHERPORT_NOT_UNIQUE, nullptr, ErrorLevel::MODERATE);
+            return true;
+        }
+    }
+    return false;
+}
+
+template <typename T, cxx::enable_if_t<std::is_same<T, iox::build::ManyToManyPolicy>::value>*>
+inline bool PortManager::hasDuplicatePublisher(const capro::ServiceDescription& service [[gnu::unused]],
```
elfenpiff, line 43: I think the name is misleading. Maybe we would like to know later if there are multiple publishers in a m to n environment and then this method is lying to us. Therefore, I would like to rename it or to handle the logic if n to m is allowed or not in the method which is calling `hasDuplicatePublishers`. I would prefer the second approach.

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
```diff
@@ -0,0 +1,53 @@
...
+inline bool PortManager::hasDuplicatePublisher(const capro::ServiceDescription& service,
+                                               const ProcessName_t& processName) noexcept
+{
+    // check if the publisher is already in the list
+    for (auto publisherPortData : m_portPool->getPublisherPortDataList())
+    {
+        popo::PublisherPortRouDi publisherPort(publisherPortData);
+        if (service == publisherPort.getCaProServiceDescription())
+        {
+            LogWarn() << "Process '" << processName
```
elfenpiff, line 31: I would move the error handling outside of this method into the method which is calling `hasDuplicatePublishers`. The reason is that no one would expect that a call to this method breaks the whole system. Maybe you would like to know in an 1 to m environment if this publisher is already created and if it is not created you would create a publisher.

But at the moment you would ask this method in an innocent way and would crash for no reasons from your point of view when there is already a publisher which is not an error! It becomes only then an error if you create one.

I would call the error handler exactly where the error has the effect of screwing up the system. This is not the place. The place is the method where this method is called.

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
```diff
@@ -0,0 +1,53 @@
...
+    return false;
+}
+
+template <typename T, cxx::enable_if_t<std::is_same<T, iox::build::ManyToManyPolicy>::value>*>
+inline bool PortManager::hasDuplicatePublisher(const capro::ServiceDescription& service [[gnu::unused]],
+                                               const ProcessName_t& processName [[gnu::unused]]) noexcept
+{
+    // Duplicates are allowed when using n:m policy
+    return false;
+}
```
elfenpiff, line 48: My suggestion is to get rid of the enable if. Have only one method and decide in the calling method how to handle the result.

ithier (approved):

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -770,7 +770,10 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
                                       const RunnableName_t& runnable [[gnu::unused]], // @todo #25 Fix introspection
                                       const PortConfigInfo& portConfigInfo) noexcept
 {
-    /// @todo #25 Add unique check for single producer type - see acquireSenderPortData()
+    if (hasDuplicatePublisher<iox::build::CommunicationPolicy>(service, processName))
+    {
+        return cxx::error<PortPoolError>(PortPoolError::UNIQUE_PUBLISHER_PORT_ALREADY_EXISTS);
```
elfenpiff, line 775: We should call the errorHandler here if we have 1 to m communication. Otherwise `hasDuplicatePublisher` has unforseen side effects.

elfenpiff (changes requested):
Move the errorHandling

iceoryx_meta/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ file (STRINGS "../VERSION" iceoryx_VERSION)
 project(iceoryx_meta VERSION ${iceoryx_VERSION})
 
 option(TOML_CONFIG "activates or deactivates TOML support - without TOML RouDi will not be build" ON)
+option(ONE_TO_MANY "Restricts communication to 1:n pattern" OFF)
```
budrus, line 20: Maybe ONE_TO_MANY_ONLY, to make more clear that it is a restriction. When I read MANY here I get the feeling that it extends and not restricts

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -54,14 +52,7 @@ using PublisherPortUserType = iox::popo::PublisherPortUser;
...
 
-/// @brief 1 to m communication enabled
-///     for n to m communication use
-///         iox::popo::SubscriberPortMultiProducer
-///         iox::cxx::VariantQueueTypes::SoFi_MultiProducerSingleConsumer;
-using SubscriberPortProducerType = iox::popo::SubscriberPortSingleProducer;
-constexpr cxx::VariantQueueTypes SUBSCRIBER_PORT_QUEUE_TYPE =
-    iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer;
-
+using SubscriberPortProducerType = iox::build::CommunicationPolicy;
```
budrus, line 55: Maybe we find a better name? One can get the impression that it is about a producer Type, but the type is a subscriber type. Maybe only SubscriberPortType?

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
mossmaurice: TBH I like the SFINAE approach for embedded, then this is purely compile-time and no run-time checks are necessary. What do you think about `duplicatePublisherAllowed`?

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
mossmaurice: Got your point. My approach was to included as few code in the text segment as possible for both variants. Other opinions? @elBoberido @budrus

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: That makes sense. However, I had the issue that some of the infos that are printed are only available when looping over publisher port list. I would remove the following:
````
<< publisherPortData->m_processName << "' with service '"
<< service.operator cxx::Serialization().toString() << "'.";
````
What do you think?

File: iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl (hunk 8 above)
Comment (budrus) [inline iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl:43]: That's a bit imprecise IMHO. Wenn you have many-to-many, there will be duplicated publishers but you don't check it as it is OK. But you have a method with name hasDuplicatePublisher that returns false. 
Maybe somthing like violatesCommunicationPolicy()

**Review by budrus — CHANGES_REQUESTED** [review]

File: iceoryx_meta/CMakeLists.txt (hunk 12 above)
Comment (mossmaurice) [reply iceoryx_meta/CMakeLists.txt:20]: Done

File: iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp (hunk 13 above)
Comment (mossmaurice) [reply iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp:55]: You're right done.

File: iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl (hunk 9 above)
Comment (budrus) [reply iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl:31]: I'm with @elfenpiff. 
If you rename it like I proposed, yopu would end up in something like
`if(violatesCommunicationPoliy())
{errorHandler(Bäääm);
}`

File: iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl (hunk 8 above)
Comment (mossmaurice) [reply iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl:43]: You're right. I like `violatesCommunicationPolicy()` 👍

File: iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl (hunk 9 above)
Comment (mossmaurice) [reply iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl:31]: Done

File: iceoryx_posh/source/roudi/port_manager.cpp [hunk 14]
```diff
@@ -770,7 +770,10 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
                                       const RunnableName_t& runnable [[gnu::unused]], // @todo #25 Fix introspection
                                       const PortConfigInfo& portConfigInfo) noexcept
 {
-    /// @todo #25 Add unique check for single producer type - see acquireSenderPortData()
+    if (hasDuplicatePublisher<iox::build::CommunicationPolicy>(service, processName))
```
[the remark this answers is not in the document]
mossmaurice: Renamed to `violatesCommunicationyPolicy()`

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: Moved the `errorHandler` call outside and left the `LogWarn` inside

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
mossmaurice: Renamed to `violatesCommunicationPolicy()`

budrus (approved):

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
elfenpiff: I would prefer the name `areDuplicatePublishersAllowed` but I would be fine with it. I would also move the warning message out since here you just ask the if it is allowed.

iceoryx_posh/source/roudi/port_manager.cpp
elfenpiff: This would be fine by me.

budrus (dismissed):

elfenpiff (approved):

budrus (approved):

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.inl
elBoberido: The caller already handles the result. This just tells the caller about the violation, the caller can still ignore it.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -770,7 +770,17 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
                                       const RunnableName_t& runnable [[gnu::unused]], // @todo #25 Fix introspection
                                       const PortConfigInfo& portConfigInfo) noexcept
 {
-    /// @todo #25 Add unique check for single producer type - see acquireSenderPortData()
+    if (doesViolateCommunicationPolicy<iox::build::CommunicationPolicy>(service).and_then(
```
elBoberido, line 773: this is really hard to read and quite unintuitive with 7 lines for the if condition. how about this
```
if(auto violatingProcessName = doesViolateCommunicationPolicy<iox::build::CommunicationPolicy>(service))
{
    // do stuff
}
```
or just move the call to `doesViolateCommunicationPolicy` out of the if condition

elfenpiff: Yes you are right, we have to improve this piece of code.

But I think the code snippet below would always be true since it is an assignment
> if(auto violatingProcessName = doesViolateCommunicationPolicy<iox::build::CommunicationPolicy>(service))
> {
>     // do stuff
> }

PR #294, opened 2020-09-16
Iox #226 cxx::vector erase returning an iterator position

### LINKED ISSUE #226: cxx::vector::erase() should return an iterator following the last removed element
## Brief feature description

The `std::vector::erase()` method returns an iterator following the last removed element. We should be compatible in this aspect to the STL.

bishibashiB:
…and , calling destructor of erased element
Fixes #226 
@elfenpiff : please have a look~


iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -293,18 +293,23 @@ inline typename vector<T, Capacity>::const_iterator vector<T, Capacity>::end() c
...
 inline typename vector<T, Capacity>::iterator vector<T, Capacity>::erase(iterator position)
 {
+    iterator ret = nullptr;
+
     if (begin() <= position && position < end())
     {
         uint64_t index = static_cast<uint64_t>(position - begin()) % (sizeof(element_t) * Capacity);
         size_t n = index;
+
+        at(n).~T();
```
elfenpiff, line 303: This is undefined behavior. You essential are doing the following.
1. deleting element at position n
2. in for loop: moving element at position n+1 to element at position n which was deleted above.

The correct behavior would be moving
1. moving element at position n + 1 to element at position n
2. increment n (`++n`) and repeat step 1 and 2 until the second last element is reached
3. destruct the last element. the content of the last element was moved to its predecessor and therefore it should be alright.

This would be essentially the old behavior. Would you agree on @bishibashiB  this or are here some kind of problems which I have overlooked?

elfenpiff (changes requested):
Fix undefined behavior.
Merge tests.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -293,18 +293,24 @@ inline typename vector<T, Capacity>::const_iterator vector<T, Capacity>::end() c
...
         for (; n + 1u < size(); ++n)
         {
-            at(n) = std::move(at(n + 1u));
+            new (&at(n)) T(std::move(at(n + 1u)));
+            at(n + 1u).~T();
```
marthtz, line 307: Could you use upper case literals here as you did in the other files?! You've touched these lines anyway.

bishibashiB: Was about doing it, but thought it looks awkward when it differs from all functions surrounding it (format-styles mixed each file).... 
Still keen on upper case literals then? Applying it to the whole file?

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -1034,3 +1175,25 @@ TEST_F(vector_test, PartiallyEqualVectorsWithDifferentCapacityAreNotEqual)
...
+
+    EXPECT_THAT(cTor, Eq(0U));
+    EXPECT_THAT(customCTor, Eq(10U));
+    EXPECT_THAT(copyCTor, Eq(0U));
+    EXPECT_THAT(moveCTor, Eq(7U));
+    EXPECT_THAT(moveAssignment, Eq(0U));
+    EXPECT_THAT(copyAssignment, Eq(0U));
+    EXPECT_THAT(dTorList.front(), Eq(2U));
+    EXPECT_THAT(dTorList.size(), Eq(8U));
+}
```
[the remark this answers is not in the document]
bishibashiB: sure

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
bishibashiB: Anticipate a 'change file to use upper case' here... 😄

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
MatthiasKillat: I do not understand the problem. But calling the destructor of the element to erased and moving all trailing elemnts forward by one (beginning with the one immediately following) and the returning the iterator to the element which now points to the next element before deletion is the expected behaviour.

MatthiasKillat: Comparing the code before and after the changes suggests that the new variant (with placement new and element-wise destructor call) should be preferred for the following reasons:

1) It is a little less restrictive, since it just requires T to be copy (or move) constructible but not copy (or move) assignable. It is rare that a type is copy constructible but not copy assignable, but possible nonetheless.
2) If move construction is available (i.e. implemented by T), the source object is only required to be destructible after the move but NOT necessarily assignable (though they often are). For this reason calling the destructor on the source object is safe but assigning is not in general.

Returning the iterator to the element  immediately after the erased element is of course STL-conform.

Issue #278, opened 2020-09-10
Introduce API IsEnabled(LogLevel logLevel) in Logger


prasannabhat:
## Brief feature description

Introduce API `IsEnabled(LogLevel logLevel)` in Logger

## Detailed information

This API checks current configured log reporting level.
Applications may want to check the actual configured reporting log level of certain loggers before doing log data preparation that is runtime intensive.

ithier:
@dkroenke @prasannabhat Can this be closed ?

prasannabhat:
@ithier It can be closed
Here is the  [PR](https://github.com/eclipse/iceoryx/pull/309)
Not sure why its not linked to the issue (commits and branch name have issue no)

PR #302, opened 2020-09-21
Iox #278 logger add is enabled

dkroenke:
Code contributed by @prasannabhat and @nihalchari


iceoryx_utils/include/iceoryx_utils/log/logger.hpp
```diff
@@ -44,6 +44,7 @@ class Logger
 
     void SetLogLevel(const LogLevel logLevel) noexcept;
     void SetLogMode(const LogMode logMode) noexcept;
+    bool IsEnabled(const LogLevel logLevel) const noexcept;
```
elfenpiff, line 47: Wrong Code Style. Aren't we using lowerCamelCase?! I think we have to adjust this.

elBoberido: This was based on the ARA API, that's the reason for this coding style

iceoryx_utils/include/iceoryx_utils/log/logger.inl
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef IOX_UTILS_LOG_LOGGER_INL
+#define IOX_UTILS_LOG_LOGGER_INL
+namespace iox
+{
+namespace log
+{
+inline bool Logger::IsEnabled(const LogLevel logLevel) const noexcept
```
elBoberido, line 21: why the `logger.inl`? There is already a `logger.cpp`.

elBoberido (changes requested):
please move the code from the `logger.inl` to the `logger.cpp`

iceoryx_utils/include/iceoryx_utils/log/logger.inl
prasannabhat: Previously this check was done inside public API Logger::Log()
Now Logger::Log() calls IsEnabled(). 
Hence this is marked as inline , not to loose the performance as compared to  reference.

elfenpiff: @prasannabhat could you please explain why it has to be an inline method?

MatthiasKillat: The inline keyword has no use here, as it is not templated and thus can be implemented in a cpp file. As there should not be multiple definitions if the implementation is moved to a cpp file it can just be removed.

iceoryx_utils/include/iceoryx_utils/log/logger.inl
```diff
@@ -0,0 +1,32 @@
...
+    if (logLevel <= m_logLevel.load(std::memory_order_relaxed))
+    {
+        return true;
+    }
+    return false;
+}
+
+} // namespace log
+} // namespace iox
+#endif // IOX_UTILS_LOG_LOGGER_INL
```
marthtz, line 32: Newline is missing.

marthtz (changes requested):

iceoryx_utils/include/iceoryx_utils/log/logger.inl
prasannabhat: > @prasannabhat could you please explain why it has to be an inline method?

Previously check against set LogLevel was done inside public API Logger::Log() (which is called for every logging operation)
Logger::Log() delegates this to IsEnabled().
Hence IsEnabled() is marked as inline , not to loose the performance compared to previous revision.

PR #307, opened 2020-09-22
Iox #25 add cxx type traits

mossmaurice:
* Carve out `type_traits.hpp` from #291
* Add tests
* Refactor `cxx::function_ref` to use the new type traits


iceoryx_utils/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "iceoryx_utils/cxx/type_traits.hpp"
+#include "test.hpp"
+
+using namespace ::testing;
+using namespace iox::cxx;
+
+TEST(TypeTraitsTest, IsInvocableResolvesToTrue)
```
ithier, line 21: It might be worth testing function pointers & callable structs too.

iceoryx_utils/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    auto lambda = [](int foo) -> void { foo++; };
+    auto sut = is_invocable<decltype(lambda), int>::value;
+    EXPECT_TRUE(sut);
+}
+
+TEST(TypeTraitsTest, IsInvocableResolvesToFalse)
+{
+    int beeblebrox{42};
```
elfenpiff, line 30: Long live Zaphod our president of the universe !

elfenpiff (approved):

ithier (approved):

elBoberido (approved):

iceoryx_utils/test/moduletests/test_cxx_type_traits.cpp
elBoberido: isn't this a magic number? shouldn't it be like this
```
constexpr int ANSWER {42};
int beeblebrox{ANSWER};
```

iceoryx_utils/test/moduletests/test_cxx_type_traits.cpp
mossmaurice: I guess that's not necessary. Lambdas are Functor classes which implement the `operator()(...)`

PR #310, opened 2020-09-23
Iox #25 service discovery for new building blocks

elfenpiff:


iceoryx_examples/icecallback_on_c/README.md
```diff
@@ -1,4 +1,4 @@
-# [NOT YET IMPLEMENTED] icecallback_on_c - Using the WaitSet to implement callbacks
+# icecallback_on_c - Using the WaitSet to implement callbacks
```
ithier, line 1: Can you name it "icecallback in C" without the underscores to be consistent with "icedelivery in C".

iceoryx_utils/include/iceoryx_utils/internal/concurrent/trigger_queue.inl
```diff
@@ -60,7 +60,9 @@ inline bool TriggerQueue<T, CAPACITY>::blocking_pop(T& out)
 template <typename T, uint64_t CAPACITY>
 inline bool TriggerQueue<T, CAPACITY>::try_pop(T& out)
 {
-    if (!m_semaphore->tryWait())
+    if (!m_semaphore->tryWait()
```
elBoberido, line 63: I don't really like this. There is too much going on in the condition of the if statement. Older developer like me have already forgotten to be in the if statement when reaching `get_value` ;)
Please decouple to error printing with e.g. an else clause. Terminating because of trying to access an non existing value should also be explicit

elBoberido (dismissed):

iceoryx_posh/source/popo/building_blocks/condition_variable_waiter.cpp
```diff
@@ -27,21 +27,40 @@ ConditionVariableWaiter::ConditionVariableWaiter(cxx::not_null<ConditionVariable
...
+        errorHandler(Error::kPOPO__CONDITION_VARIABLE_WAITER_SEMAPHORE_CORRUPTED_IN_WAIT, nullptr, ErrorLevel::FATAL);
+    }
 }
 
 bool ConditionVariableWaiter::timedWait(units::Duration timeToWait) noexcept
 {
     auto timeout = timeToWait.timespec(units::TimeSpecReference::Epoch);
     auto continueOnInterrupt{false};
-    return getMembers()->m_semaphore.timedWait(&timeout, continueOnInterrupt);
+    auto result = getMembers()->m_semaphore.timedWait(&timeout, continueOnInterrupt);
```
budrus, line 54: so we can have spurious wakeups?

iceoryx_utils/source/posix_wrapper/semaphore.cpp
```diff
@@ -79,27 +81,33 @@ void Semaphore::closeHandle() noexcept
...
 }
 
-bool Semaphore::getValue(int& value) const noexcept
+cxx::expected<int, SemaphoreError> Semaphore::getValue() const noexcept
 {
-    return m_isInitialized
-           && !cxx::makeSmartC(iox_sem_getvalue, cxx::ReturnMode::PRE_DEFINED_ERROR_CODE, {-1}, {}, m_handlePtr, &value)
-                   .hasErrors();
-}
+    int value;
```
marthtz, line 86: Shouldn't we prefer fixed width integer types?

iceoryx_examples/icecallback_on_c/README.md
elfenpiff: done

iceoryx_posh/source/popo/building_blocks/condition_variable_waiter.cpp
elfenpiff: Spurious wakeups are handled in `timedWait` inside the `semaphore` and cannot occur on this level.

iceoryx_utils/source/posix_wrapper/semaphore.cpp
elfenpiff: In general you are correct but the posix semaphore explicitly contains an integer therefore we have to stick to that type for now.

mossmaurice: Couldn't we then hit issues with platform independence? TBH I'd prefer a cast.

budrus (approved):

iceoryx_dds/source/dds2iceoryx_app/main.cpp
```diff
@@ -47,7 +47,7 @@ class ShutdownManager
     static std::atomic_bool s_shutdownRequested;
     ShutdownManager() = default;
 };
-iox::posix::Semaphore ShutdownManager::s_semaphore = iox::posix::Semaphore::create(0u).get_value();
+iox::posix::Semaphore ShutdownManager::s_semaphore = iox::posix::Semaphore::create(iox::posix::CreateUnnamedSingleProcessSemaphore, 0u).get_value();
```
dkroenke, line 50: please use uppercase U

iceoryx_posh/include/iceoryx_posh/internal/popo/subscriber.inl
```diff
@@ -166,7 +166,9 @@ inline bool Subscriber_t<ReceiverPortType>::tryWaitForChunk() noexcept
     const auto semaphore = m_receiver.GetShmSemaphore();
     assert(semaphore != nullptr && "TryWaitForChunk: semaphore is not set");
 
-    return semaphore->tryWait();
+    auto call = semaphore->tryWait();
+    assert(call.has_error() && "Semaphore is corrupted");
```
dkroenke, line 170: Shouldn't this maybe also handled in Releasebuild?

iceoryx_posh/source/popo/wait_set.cpp
```diff
@@ -89,12 +89,47 @@ void WaitSet::detachAllConditions() noexcept
 
 WaitSet::ConditionVector WaitSet::timedWait(const units::Duration timeout) noexcept
 {
-    return waitAndReturnFulfilledConditions<WaitPolicy::TIMED_WAIT>(cxx::make_optional<units::Duration>(timeout));
+    return waitAndReturnFulfilledConditions([this, timeout] { return !m_conditionVariableWaiter.timedWait(timeout); });
```
mossmaurice, line 92: Nice!

mossmaurice (approved):

iceoryx_utils/source/posix_wrapper/semaphore.cpp
```diff
@@ -201,30 +229,53 @@ Semaphore::Semaphore(iox_sem_t* handle, const unsigned int value) noexcept
...
     }
+    else
+    {
+        m_isInitialized = false;
+        m_errorValue = SemaphoreError::CREATION_FAILED;
+    }
 }
 
-Semaphore::Semaphore(const char* name, const int oflag) noexcept
+Semaphore::Semaphore(OpenNamedSemaphore_t, const char* name, const int oflag) noexcept
```
dkroenke, line 239: The Fixedstring should be used for the name to avoid unsafe strncpy operation below.

iceoryx_utils/source/posix_wrapper/semaphore.cpp
```diff
@@ -286,5 +337,25 @@ bool Semaphore::isNamedSemaphore() noexcept
 {
     return m_isNamedSemaphore;
 }
+
+SemaphoreError Semaphore::errnoToEnum(const int errnoValue) const noexcept
```
dkroenke, line 341: shouldnt we use exact-width types like uint32_t here?

dkroenke (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/popo/subscriber.inl
mossmaurice: `Expects`?

elfenpiff: You are completely right but this is the old API which will be deleted in the next weeks. The assert was already there I just adapted it to the new semaphore API

iceoryx_utils/source/posix_wrapper/semaphore.cpp
elfenpiff: No since we get the errno (type int) has input argument here.

iceoryx_utils/source/posix_wrapper/semaphore.cpp
elfenpiff: The argument is forwarded to `sem_open` which is using a `const char *`, see: https://www.man7.org/linux/man-pages/man3/sem_open.3.html

Therefore, if we run into issues `smart_c` will catch them and inform the user via the creation pattern.

The fixed string should be used in the upper layers.

budrus (approved):

dkroenke (approved):

Issue #298, opened 2020-09-17
cxx::unique_ptr implementation


ithier:
## Brief feature description

We need a unique_ptr implementation for the new publisher/subscriber API.
Since we cannot use the STL, we need to implement our own.

## Detailed information
The implementation should:
* Allow users to provide a custom deleter
* Not encode the deleter in the type (like the STL does)
* Not use any heap ... this one will be a challenge - where do we store the deleter function ? Consider this idea: https://github.com/eclipse/iceoryx/pull/291#discussion_r490762168

The following short-comings have been identified in the existing (minimal) cxx:unique_ptr implementation:
* Missing a few operator implementations
* No dereference operator

PR #316, opened 2020-09-25
Iox #298 add unique ptr implementation

marthtz:


ithier (approved):

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,113 @@
...
+    unique_ptr() = delete;
+
+    ///
+    /// @brief unique_ptr Creates an empty unique ptr that owns nothing. Can be passed ownership later via reset.
+    ///
+    unique_ptr(cxx::function_ref<void(ptr_t)>&& deleter) noexcept;
+
+    ///
+    /// @brief unique_ptr Creates a unique pointer that takes ownership of an object.
+    /// @details A deleter must always be provided as no default can be provided given that no head is used.
```
mossmaurice, line 46: heap?

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,113 @@
...
+    /// The unique_ptr must know how to delete the managed object when pointer out of scope.
+    /// @param ptr The raw pointer to the object to be managed.
+    /// @param deleter The deleter function for cleaning up the managed object.
```
mossmaurice, line 49: We should mention that function_ref is non-owning. So the user needs to care about the lifetime of the callable.

iceoryx_utils/test/moduletests/test_cxx_unique_ptr.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+        auto anotherSut = std::move(sut);
+
+        // no deleter call during movie
+        EXPECT_FALSE(m_deleterCalled);
+        EXPECT_FALSE(sut);
+        EXPECT_EQ(anotherSut.get(), object);
+    }
+    // anotherSUT is out of scope and should have called deleter
+    EXPECT_TRUE(m_deleterCalled);
```
mossmaurice, line 136: You could verify that the deleter wasn't called twice?

iceoryx_utils/test/moduletests/test_cxx_unique_ptr.cpp
```diff
@@ -0,0 +1,433 @@
...
+        anotherSut = std::move(sut);
+
+        // anotherSUT is overwritten so it should have called its anotherDeleter
+        EXPECT_TRUE(m_anotherDeleterCalled);
+        EXPECT_FALSE(m_deleterCalled);
+        EXPECT_FALSE(sut);
+        EXPECT_EQ(anotherSut.get(), object);
+    }
+    // SUT is out of scope and should have called deleter
+    EXPECT_TRUE(m_deleterCalled);
```
mossmaurice, line 156: Same as above

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
marthtz: Done

iceoryx_utils/test/moduletests/test_cxx_unique_ptr.cpp
marthtz: Done

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
marthtz: Done

iceoryx_utils/test/moduletests/test_cxx_unique_ptr.cpp
marthtz: Done

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef IOX_UTILS_CXX_UNIQUE_PTR_HPP
+#define IOX_UTILS_CXX_UNIQUE_PTR_HPP
+
+#include "iceoryx_utils/cxx/function_ref.hpp"
+
+#include <functional>
```
ithier, line 20: This can be removed now.

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// @brief unique_ptr Creates a unique pointer that takes ownership of an object.
+    /// @details A deleter must always be provided as no default can be provided given that no heap is used.
+    /// The unique_ptr must know how to delete the managed object when the pointer goes out of scope.
+    /// @param ptr The raw pointer to the object to be managed.
+    /// @param deleter The deleter function for cleaning up the managed object. As cxx:function_ref used for the deleter
+    ///                is non-owning the user needs to care about the lifetime of the callable!
+    ///
+    unique_ptr(ptr_t ptr, function_ref<void(ptr_t)>&& deleter) noexcept;
+
+    unique_ptr(std::nullptr_t) noexcept;
```
budrus, line 54: Can I create a unique_ptr with this c'tor and then later use the reset method to provide a valid pointer. And is then the deleter not set as required and the "deleter must always be provided" rule is violated?

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+}
+
+template <typename T>
+void unique_ptr<T>::reset(T* ptr) noexcept
+{
+    if (m_ptr && m_deleter)
+    {
+        m_deleter(m_ptr);
+    }
+    m_ptr = ptr;
```
budrus, line 98: So is it an error if ptr != nullptr and !m_deleter?
Do we need a cxx::Expects here? @elfenpiff @ithier @elBoberido

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// @brief operator bool Returns true if it points to something.
+    ///
+    explicit operator bool() const noexcept;
+
+    ///
+    /// @brief get Retrieve the underlying raw pointer.
+    /// @details The unique_ptr retains ownership, therefore the "borrowed" pointer must not be deleted.
+    /// @return Pointer to managed object or nullptr if none owned.
+    ///
+    ptr_t get() const noexcept;
```
elfenpiff, line 83: It is a const method but the underlying type can be changed by derereferencing `ptr_t`. I would suggest to create two methods
1. `const ptr_t get() const noexcept;`
2. `ptr_t get() noexcept;`

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    ///
+    /// Automatically deletes the owned object on destruction.
+    ///
+    ~unique_ptr() noexcept;
+
+    ///
+    /// Return the stored pointer.
+    ///
+    ptr_t operator->() noexcept;
```
elfenpiff, line 71: I would recommend adding a const overload of this method.
`const ptr_t operator->() const noexcept;`

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    ///
+    /// @brief unique_ptr Creates a unique pointer that takes ownership of an object.
+    /// @details A deleter must always be provided as no default can be provided given that no heap is used.
+    /// The unique_ptr must know how to delete the managed object when the pointer goes out of scope.
+    /// @param ptr The raw pointer to the object to be managed.
+    /// @param deleter The deleter function for cleaning up the managed object. As cxx:function_ref used for the deleter
+    ///                is non-owning the user needs to care about the lifetime of the callable!
+    ///
+    unique_ptr(ptr_t ptr, function_ref<void(ptr_t)>&& deleter) noexcept;
```
elfenpiff, line 52: Const correctness. Please write `unique_ptr(ptr_t const ptr...` since `ptr_t` is in the end `T*`

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+#ifndef IOX_UTILS_CXX_UNIQUE_PTR_INL
+#define IOX_UTILS_CXX_UNIQUE_PTR_INL
+
+namespace iox
+{
+namespace cxx
+{
+template <typename T>
+unique_ptr<T>::unique_ptr(function_ref<void(ptr_t)>&& deleter) noexcept
+    : m_deleter(deleter)
```
elfenpiff, line 24: I would use constructor delegation here.
`: unique_ptr(nullptr, deleter)` otherwise we end up with code duplication and the code can become more error prone.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+template <typename T>
+unique_ptr<T>::unique_ptr(ptr_t ptr, function_ref<void(ptr_t)>&& deleter) noexcept
+    : m_ptr(ptr)
+    , m_deleter(std::move(deleter))
+{
+}
+
+template <typename T>
+unique_ptr<T>::unique_ptr(std::nullptr_t) noexcept
+{
```
elfenpiff, line 37: Please try to use always constructor delegation. You have on constructor which provides access for the whole construction process and the other constructors delegate their stuff to this constructor. In the `unique_ptr` case it would be the ctor: `unique_ptr<T>::unique_ptr(ptr_t ptr, function_ref<void(ptr_t)>&& deleter) noexcept`
and every other ctor should delegate to it.
Then you would realize that this `unique_ptr` ctor has a problem since it does not define a deleter.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
elfenpiff: It is allowed since you are setting the `unique_ptr` to a state where it does not contain a pointer which is a valid state.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+template <typename T>
+T* unique_ptr<T>::release() noexcept
+{
+    auto ptr = m_ptr;
+    m_ptr = nullptr;
+    return ptr;
+}
+
+template <typename T>
+void unique_ptr<T>::reset(T* ptr) noexcept
```
elfenpiff, line 92: Please add const: `void unique_ptr<T>::reset(T * const ptr) noexcept`
Why are we using here `T` again and not `ptr_t`.
I think we should either use `ptr_t` consequently or we should get rid of it and only use `T`

elfenpiff (changes requested):

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+template <typename T>
+T* unique_ptr<T>::operator->() noexcept
+{
+    return get();
+}
+
+template <typename T>
+unique_ptr<T>::operator bool() const noexcept
+{
+    return get() == ptr_t() ? false : true;
```
bishibashiB, line 74: or `return get() != ptr_t;` ...

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+}
+
+template <typename T>
+bool operator==(const unique_ptr<T>& x, std::nullptr_t)
+{
+    return !x;
+}
+
+template <typename T>
+bool operator==(std::nullptr_t, const unique_ptr<T>& x)
```
bishibashiB, line 131: wondering whether this could default to a `bool operator==(const unique_ptr<T>& x, const unique_ptr<U>& y)` where one argument gets generated by the constructor `unique_ptr<T>::unique_ptr(std::nullptr_t)`

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+{
+    return !x;
+}
+
+template <typename T, typename U>
+bool operator!=(const unique_ptr<T>& x, const unique_ptr<U>& y)
```
bishibashiB, line 137: Maybe implement via `operator==`

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
```diff
@@ -0,0 +1,157 @@
...
+template <typename T, typename U>
+bool operator==(const unique_ptr<T>& x, const unique_ptr<U>& y)
+{
+    return x.get() == y.get();
+}
+
+template <typename T>
+bool operator==(const unique_ptr<T>& x, std::nullptr_t)
+{
+    return !x;
```
bishibashiB, line 127: are you relying on the `operator!` implementation of `T`?

iceoryx_utils/test/moduletests/test_cxx_unique_ptr.cpp
```diff
@@ -0,0 +1,453 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
bishibashiB, line 1: :snail:

bishibashiB (commented):
didn't make it to the tests yet

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
ithier: `void unique_ptr<T>::reset(T * const ptr)` or `void unique_ptr<T>::reset(T * ptr) const` ?

Do we need an overload for the const ptrs ?

ithier: Ping me when you're free, got a question about this.

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
ithier: Removed the nullptr constructor.

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+///
+/// Only a subset of STL's 'std::unique_ptr' has been implemented so far.
+/// Main difference is that 'Deleter' is a function reference ('cxx:function_ref') here compared to a 'std::function' in
+/// the STL implementation.
+///
+template <typename T>
+class unique_ptr
+{
+  public:
+    using ptr_t = T*;
```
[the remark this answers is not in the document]
ithier: ?

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+    ///
+    unique_ptr(ptr_t ptr, function_ref<void(ptr_t)>&& deleter) noexcept;
+
+    unique_ptr(std::nullptr_t) noexcept;
+
+    // Not copy-able to ensure uniqueness.
+    unique_ptr(const unique_ptr& other) = delete;
+    unique_ptr& operator=(const unique_ptr&) = delete;
+
+    unique_ptr(unique_ptr&& rhs) noexcept;
```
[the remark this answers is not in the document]
ithier: This is the move constructor.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
ithier: That's way more intuitive haha.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
ithier: Maybe... the STL implementation has both for some reason though and I don't understand why so I thought it would be safest to do the same.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
ithier: Removed the nullptr constructor.

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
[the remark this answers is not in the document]
bishibashiB: thought it's not mentioned in ".[clang-tidy](https://github.com/eclipse/iceoryx/blob/master/iceoryx_utils/.clang-tidy)", syncing `TypeAliasCase` with e.g. `ClassCase`, I did assume a `CamelCase` format for this type alias (and without postfix, maybe `TPtrType`).
Again, if choosing the `ptr_t` format makes sense e.g. to follow stl or have typetraits working ... that's something to take into concern. Some lessons learned.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
budrus: I'm not sure if you got me right. I think you can set it to state where it does contain a pointer but no deleter

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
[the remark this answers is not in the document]
bishibashiB: Surely: no magic involved here. But will doxygen mention the available moveConstr/Assignment, if you omit the concerning comment here?
Following up on that: even adding doxygen comments to the deleted Copy/MoveConstructor will draw a more complete picture of what's available when consulting the docs.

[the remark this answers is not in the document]
ithier: I think only the available constructors will be shown. 
I think the fact that they exist (or don't) is enough information to the user. I can't think of any extra useful documentation to add ...

iceoryx_utils/include/iceoryx_utils/cxx/unique_ptr.hpp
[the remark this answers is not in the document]
ithier: Removed the alias, it was kind of getting in the way when considering constness anyway.

iceoryx_utils/include/iceoryx_utils/internal/cxx/unique_ptr.inl
bishibashiB: after checking ... seems safe (operator! on pointer only).

mossmaurice (approved):

budrus (approved):

Issue #325, opened 2020-10-30
Deadlock in shared mem mutex when apps are killed


marthtz:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.4.0

**Observed result or behaviour:**
Deadlock in shared mem mutex when apps are killed

**Expected result or behaviour:**
No deadlock.

**Conditions where it occurred / Performed steps:**
Kill an app while some system is running

elBoberido:
I've had an idea how this could be solved in an elegant way without a mutex but an atomic and semaphore instead. 
Let's assume we have the following class
```c++
class DualAccessTransactionTray
{
    private:
    class AccessGuard;

    public:
    AccessGuard acquireRouDiAccess()
    {
        switch (m_accessToken.exchange(AccessToken::ROUDI, std::memory_order_acquire))
        {
        case AccessToken::NONE:
            break;
        case AccessToken::ROUDI:
            errorHandler(kBROKEN_INVARIANT);
            break;
        case AccessToken::APP:
            m_waitingLine.wait();
            break;
        }
        return AccessGuard{*this, AccessToken::ROUDI};
    }
    AccessGuard acquireAppAccess()
    {
        switch (m_accessToken.exchange(AccessToken::APP, std::memory_order_acquire))
        {
        case AccessToken::NONE:
            break;
        case AccessToken::ROUDI:
            m_waitingLine.wait();
            break;
        case AccessToken::APP:
            errorHandler(kBROKEN_INVARIANT);
            break;
        }
        return AccessGuard{*this, AccessToken::APP};
    }
    void cleanupAfterAppWalkedThePlank()
    {
        releaseAccess(AccessToken::APP);
        // just to be sure the memory is synchronized
        AccessToken expected = AccessToken::APP;
        m_accessToken.compare_exchange_strong(expected, AccessToken::NONE, std::memory_order_acq_rel);
    }

    private:
    enum class AccessToken
    {
        NONE,
        ROUDI,
        APP
    };

    void releaseAccess(AccessToken expected)
    {
        auto casSuccessful =
            m_accessToken.compare_exchange_strong(expected, AccessToken::NONE, std::memory_order_release);
        if (!casSuccessful)
        {
            if (expected == AccessToken::NONE)
            {
                errorHandler(kBROKEN_INVARIANT);
            }
            m_waitingLine.post();
        }
    }

    class AccessGuard
    {
        public:
[...]

PR #330, opened 2020-11-02
Iox #325 deadlock in shared mem mutex

marthtz:
Introduce a robust, recursive mutex.
If recursive is enabled, the same thread, which has already locked the mutex, can lock the mutex again without getting blocked.
If robust is enabled, a process or thread can exit while having the lock without causing an invalid mutex. In the next lock call to this mutex the system recognizes the mutex is locked by a dead process or thread and allows the mutex to be restored.

### DIFF
--- a/iceoryx_utils/source/posix_wrapper/mutex.cpp
+++ b/iceoryx_utils/source/posix_wrapper/mutex.cpp
@@ -23,7 +23,7 @@ namespace iox
 {
 namespace posix
 {
-mutex::mutex(bool f_isRecursive)
+mutex::mutex(const Recursive recursive, const Robust robust)
 {
     pthread_mutexattr_t attr;
     bool isInitialized{true};
@@ -41,7 +41,7 @@ mutex::mutex(bool f_isRecursive)
                                       {0},
                                       {},
                                       &attr,
-                                      f_isRecursive ? PTHREAD_MUTEX_RECURSIVE_NP : PTHREAD_MUTEX_FAST_NP)
+                                      recursive == Recursive::ON ? PTHREAD_MUTEX_RECURSIVE_NP : PTHREAD_MUTEX_FAST_NP)
                           .hasErrors();
     isInitialized &= !cxx::makeSmartC(pthread_mutexattr_setprotocol,
                                       cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE,
@@ -50,6 +50,16 @@ mutex::mutex(bool f_isRecursive)
                                       &attr,
                                       PTHREAD_PRIO_NONE)
                           .hasErrors();
+    if (robust == Robust::ON)
+    {
+        isInitialized &= !cxx::makeSmartC(pthread_mutexattr_setrobust,
+                                          cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE,
+                                          {0},
+                                          {},
+                                          &attr,
+                                          PTHREAD_MUTEX_ROBUST)
+                              .hasErrors();
+    }
 
     isInitialized &=
         !cxx::makeSmartC(pthread_mutex_init, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, &m_handle, &attr)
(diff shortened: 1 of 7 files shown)


marthtz:
Robust mutex is not supported on Win/Mac. We need to find a solution!

mossmaurice:
Windows and Mac builds are broken :-( IMHO best solution would be to add empty wrappers for e.g. `pthread_mutexattr_setrobust` in the `iceoryx_utils/platform/*/include/iceoryx_utils/platform/pthread.hpp`

budrus:
@marthtz @sculpordwarf is this feature also available in MacOS and Windows?

budrus:
This will prevent a deadlock and makes the mutex consistent again. 
But how to deal with the problem that the process that died while holding the mutex could leave some shared data objects in an inconsistent state. The locking will work now but depending on the critical section and the involved shared data objects this opens the road to undefined behavior. 

What's your feeling on this @marthtz @sculpordwarf ?

sculpordwarf:
Windows: WAIT_ABANDONED
Mac: EDEADLK -> you need to make an example an step in with the debugger, there isn´t enough docu

marthtz:
> 
> What's your feeling on this @marthtz @sculpordwarf ?

You're right @budrus, we're getting into difficult terrain here. All shared data protected by an invalid mutex should be treated as corrupted. Off the top of my head we could do two things when an invalid mutex has been detected. We could either re-init these data blocks or we throw an error (and may shutdown gracefully).

iceoryx_utils/source/posix_wrapper/mutex.cpp
```diff
@@ -50,6 +50,16 @@ mutex::mutex(bool f_isRecursive)
                                       &attr,
                                       PTHREAD_PRIO_NONE)
                           .hasErrors();
+    if (robust == Robust::ON)
+    {
+        isInitialized &= !cxx::makeSmartC(pthread_mutexattr_setrobust,
```
elfenpiff, line 55: I think this call is not supported for Mac OS and therefore has to be reimplemented and I am not aware if Windows is supporting robust mutexes?! 

But maybe you can follow up this link: https://stackoverflow.com/questions/24946886/is-there-pthread-mutex-robust-equivalent-in-mac-os-x or contact me and we can work together on a solution.

iceoryx_utils/test/moduletests/test_posix_mutex.cpp
```diff
@@ -47,7 +48,7 @@ class Mutex_test : public Test
         }
     }
 
-    iox::posix::mutex sut{false};
+    mutex_t sut{mutex_t::Recursive::OFF, mutex_t::Robust::OFF};
```
elfenpiff, line 51: Test cases for recursive and robust mutexes are missing.
Like:
- normal creation and usage should work for recursive, robust and recursive and robust mutex
- multiple locks should not lead to deadlock in a recursive mutex in a single thread
- multiple locks should cause a deadlock in a recursive mutex from different threads (start multiple threads, let the first thread deadlock and then unlock the mutex in the thread it was locked before)
- robust mutex cleanup should work ... here we maybe have to think how we can test it. @marthtz do you have an idea?

elfenpiff (changes requested):
- Fix Mac OS behavior so that is consistent with linux (if possible)
- Find robust mutex solution for windows (if possible)
- Add additional unit tests for mutex

marthtz:
Isn't it the case that with the introduction of the building blocks / new API the number of mutexes is reduced?! So, maybe the robust approach is a bit over the top?

budrus:
> You're right @budrus, we're getting into difficult terrain here. All shared data protected by an invalid mutex should be treated as corrupted. Off the top of my head we could do two things when an invalid mutex has been detected. We could either re-init these data blocks or we throw an error (and may shutdown gracefully).

So we would have to provide this information and not silently make the mutex consistent again.

budrus:
> 
> 
> Isn't it the case that with the introduction of the building blocks / new API the number of mutexes is reduced?! So, maybe the robust approach is a bit over the top?

They are reduced but not completely eliminated

budrus:
My proposal after some discussion with @sculpordwarf : 

- Windows and Mac. No support of robust mutex so far. We would fall back to the deadlock situation as we have today
- We need the information whether a mutex was held by a dead application and the `lock()` call made it consistent again. The user of the mutex needs to be able to decide what the consequence is. I.e. if the data protected by the mutex could be inconsistent or hast to be reset etc. The idea is to introduce a `was_inconsistent()` method that can be called after lock
- For the old ports we came to the conclusion that a robust mutex is fine for now. For the locking policy in the new building blocks we start with a non-robust mutex and first have to analyze if and how the robust mutex could help
- For tests we think it should be possible to have a global mutex and several threads to force the interesting situations

elfenpiff:
@marthtz is this PR dead? Can we remove it?

elfenpiff:
@marthtz I assume that this PR is dead and remove it. If I am mistaken just reopen it.

Issue #324, opened 2020-10-29
RouDi does not free shared memory properly on x86 Linux


marthtz:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.4.0

**Observed result or behaviour:**
RouDi does not free shared memory properly on x86 Linux.
When an app is closed RouDi might free the apps shared mem while the app is still cleaning up.

**Expected result or behaviour:**
No bus error or ipcrm shm

**Conditions where it occurred / Performed steps:**
After enough iterations, all memory will be exhausted and RouDi will not start any more with a Bus Error.

PR #333, opened 2020-11-02
Iox #324 roudi improve app shutdown

marthtz:
Introduce 2-step shutdown procedure for apps/processes.
First with a SIGTERM and if they have not terminated after finallKillTime they're killed with SIGKILL.


iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -142,6 +147,21 @@ void CmdLineParser::parse(int argc, char* argv[], const CmdLineArgumentParsingMo
             }
             break;
         }
+        case 'f':
+        {
+            uint32_t finalKillTimeInSeconds{0u};
+            constexpr uint64_t MAX_FINAL_KILL_TIME = ((1ul << 32) - 1);
```
mossmaurice, line 153: `std::numeric_limits<uint64_t>::max`?

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -278,23 +419,33 @@ bool ProcessManager::removeProcess(const ProcessName_t& name) noexcept
...
             return true; // we can assume there are no other processes with this name
         }
         ++it;
     }
     return false;
 }
 
+bool ProcessManager::removeProcess(ProcessList_t::iterator& processIter) noexcept
+{
+    // don't take the lock, else it needs to be recursive
```
budrus, line 438: what does this mean? Do you expect this method to be called only with an already locked mutex? Calling erase on an unlocked list could end up in undefined behavior, or?

iceoryx_utils/README.md
```diff
@@ -69,12 +72,27 @@ queue which is thread-safe when combined with `smart_lock`. To learn more about
...
 |`LockedLoFFLi`       | i | X | LIFO based index manager (Lock Free Free List). One building block of our memory manager. After construction it contains the indices {0 ... n} which you can acquire and release. |
 |`LoFFLi`             | i |   | Lock-free version of LockedLoFFLi. |
-|`smart_lock`         | i |   | Creates arbitrary thread safe constructs which then can be used like smart pointers. If some STL type should be thread safe use the smart_lock to create the thread safe version in one line. Based on some ideas presented in [Wrapping C++ Member Function Calls](https://stroustrup.com/wrapper.pdf) |
 |`SoFi`               | i |   | Single Producer, Single Consumer Lock Free Safely overflowing FiFo (SoFi). |
 |`TACO`               | i |   | Thread Aware exChange Ownership (TACO). Solution if you would like to use `std::atomic` with data types larger than 64 bit. Wait free data synchronization mechanism between threads.|
-|`TriggerQueue`       | i | X | Queue with a `push` - `pop` interface where `pop` is blocking as long as the queue is empty. Can be used as a building block for active objects. |
+
+an atrribute overview
```
budrus, line 82: attribute?

iceoryx_utils/README.md
```diff
@@ -69,12 +72,27 @@ queue which is thread-safe when combined with `smart_lock`. To learn more about
...
 |`LoFFLi`             | i |   | Lock-free version of LockedLoFFLi. |
-|`smart_lock`         | i |   | Creates arbitrary thread safe constructs which then can be used like smart pointers. If some STL type should be thread safe use the smart_lock to create the thread safe version in one line. Based on some ideas presented in [Wrapping C++ Member Function Calls](https://stroustrup.com/wrapper.pdf) |
 |`SoFi`               | i |   | Single Producer, Single Consumer Lock Free Safely overflowing FiFo (SoFi). |
 |`TACO`               | i |   | Thread Aware exChange Ownership (TACO). Solution if you would like to use `std::atomic` with data types larger than 64 bit. Wait free data synchronization mechanism between threads.|
-|`TriggerQueue`       | i | X | Queue with a `push` - `pop` interface where `pop` is blocking as long as the queue is empty. Can be used as a building block for active objects. |
+
+an atrribute overview 
+
+| Data Structure | Shared Memory usable  | Thread-Safe | Lock-Free | Concurrent Producers : Consumers | Bounded Capacity | Data Type Restriction |Use Case |
```
budrus, line 84: I like :-)

iceoryx_utils/README.md
```diff
@@ -69,12 +72,27 @@ queue which is thread-safe when combined with `smart_lock`. To learn more about
...
+|----------------|-----------------------|-------------|-----------|----------------------------------|------------------|---|-----------------------|
+|`FiFo`          | Yes | No  | No  | n/a | Yes | Copyable            | FIFO Data transfer |
```
budrus, line 86: The FiFo is thread safe and lock-free, or? And is SPSC

iceoryx_utils/platform/win/include/iceoryx_utils/platform/wait.hpp
```diff
@@ -19,4 +19,9 @@ inline int WEXITSTATUS(int)
     return 0;
 }
 
+inline int waitpid(int pid, int* stat_loc, int options)
```
budrus, line 22: what is the consequence of this default implementation?

iceoryx_utils/README.md
```diff
@@ -69,12 +72,27 @@ queue which is thread-safe when combined with `smart_lock`. To learn more about
...
-|`TriggerQueue`       | i | X | Queue with a `push` - `pop` interface where `pop` is blocking as long as the queue is empty. Can be used as a building block for active objects. |
+
+an atrribute overview 
+
+| Data Structure | Shared Memory usable  | Thread-Safe | Lock-Free | Concurrent Producers : Consumers | Bounded Capacity | Data Type Restriction |Use Case |
+|----------------|-----------------------|-------------|-----------|----------------------------------|------------------|---|-----------------------|
+|`FiFo`          | Yes | No  | No  | n/a | Yes | Copyable            | FIFO Data transfer | 
+|`smart_lock`    | No  | Yes | No  | n/a | n/a | Copyable            | Wrapper to make classes thread-safe (by using a lock)| 
+|`TriggerQueue`  | No  | Yes | No  | n:m | Yes | Copyable            | Process events in a blocking way| 
+|`LockedLoFFLi`  | Yes | Yes | No  | n:m | Yes | int32               | manage memory access, LIFO order |
```
budrus, line 89: we don't need the LockedLoFFli any longer, or @elfenpiff ?

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
elBoberido: should actually be std::numeric_limits<uint32_t>::max

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -72,6 +72,7 @@ constexpr char SHM_NAME[] = "/iceoryx_mgmt";
 using namespace units::duration_literals;
 
 // Timeout
+constexpr units::Duration PROCESS_FINAL_KILL_TIME = 45_s;
```
ithier, line 75: Name could be revised... it's not immediately clear what this means without context.
Maybe: `PROCESS_MAXIMUM_KILL_DELAY` or something. 

Maybe a bit pedantic.

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
     m_memoryManagerOfCurrentProcess = m_segmentInfo.m_memoryManager;
 }
 
-void ProcessManager::killAllProcesses() noexcept
+void ProcessManager::killAllProcesses(const units::Duration finalKillTime) noexcept
 {
     std::lock_guard<std::mutex> g(m_mutex);
+    cxx::vector<bool, MAX_PROCESS_NUMBER> processStillRunning(m_processList.size(), true);
```
ithier, line 140: Did you mean processes** ?
i.e. `processesStillRunning`

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+        // if it was killed we need to check if it really has terminated, if we can't kill it, we consider it
+        // terminated
+        processStillRunning[i] = requestShutdownOfProcess(process, ShutdownPolicy::SIG_TERM, ShudownLog::FULL);
+        ++i;
+    }
 
-    for (; itEnd != it; ++it)
+    // wait for process to complete
+    bool allProcessFinished = false;
+    posix::Timer finalKillTimer(finalKillTime);
```
ithier, line 156: Can you please avoid using the posix timer since we want to scrap it ?

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+
+    yield();
+    auto waitForKillProcessed = [&]() {
```
ithier, line 159: Weird wording. How about `awaitProcessTermination` ?

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+        bool processStateChanged = true;
```
ithier, line 160: This variable name is really confusing. 
Reading the could for the first time, the first question I have is "why has the process state changed before doing anything" ?

Something like `shouldCheckProcessState` would be clearer (to me at least).

ithier (changes requested):

iceoryx_posh/source/roudi/roudi_process.cpp
elBoberido: I think this is only called from a method which already holds the lock. This could maybe made more clear by the method name.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
```diff
@@ -57,7 +57,8 @@ class RouDi
           const config::MonitoringMode f_monitoringMode = config::MonitoringMode::ON,
           const bool f_killProcessesInDestructor = true,
           const MQThreadStart mqThreadStart = MQThreadStart::IMMEDIATE,
-          const version::CompatibilityCheckLevel compatibilityCheckLevel = version::CompatibilityCheckLevel::PATCH);
+          const version::CompatibilityCheckLevel compatibilityCheckLevel = version::CompatibilityCheckLevel::PATCH,
```
elfenpiff, line 60: I think we reached now the maximum number of arguments for the good old roudi ctor. I would suggest the following:
1. Create a struct called `struct RoudiStartupParameters` and move all the arguments from here in there.
2. Use this struct `RoudiStartupParameters` in this ctor instead of all the arguments which provides us with the big advantage that we only have to change one place if we add another argument to roudi.
3. Use it like
```cpp
RouDi myRoudi(RoudiStartupParameters{memoryInterface, portManager, config::MonitoringMode::ON ....});
```
then we do not have to change much in the source code. We just have to add `RoudiStartupParameters`

iceoryx_posh/source/roudi/roudi_process.cpp
elfenpiff: @ithier I would leave the name as it is now since you would read it like `processStillRunning[5]` which sound like is process ```5``` still running

iceoryx_utils/README.md
elfenpiff: @budrus No. We can remove it safely since we have the LoFFli and if there is something wrong we can always use ```cxx::smart_lock<LoFFli>``` which produces a threadsafe LoFFli with mutexes lick LockedLoFFli

iceoryx_utils/README.md
elfenpiff: @budrus: Correct!

iceoryx_utils/README.md
```diff
@@ -69,12 +72,27 @@ queue which is thread-safe when combined with `smart_lock`. To learn more about
...
 |`SoFi`               | i |   | Single Producer, Single Consumer Lock Free Safely overflowing FiFo (SoFi). |
 |`TACO`               | i |   | Thread Aware exChange Ownership (TACO). Solution if you would like to use `std::atomic` with data types larger than 64 bit. Wait free data synchronization mechanism between threads.|
-|`TriggerQueue`       | i | X | Queue with a `push` - `pop` interface where `pop` is blocking as long as the queue is empty. Can be used as a building block for active objects. |
+
+an atrribute overview 
+
+| Data Structure | Shared Memory usable  | Thread-Safe | Lock-Free | Concurrent Producers : Consumers | Bounded Capacity | Data Type Restriction |Use Case |
+|----------------|-----------------------|-------------|-----------|----------------------------------|------------------|---|-----------------------|
+|`FiFo`          | Yes | No  | No  | n/a | Yes | Copyable            | FIFO Data transfer | 
+|`smart_lock`    | No  | Yes | No  | n/a | n/a | Copyable            | Wrapper to make classes thread-safe (by using a lock)|
```
elfenpiff, line 87: The `smart_lock` can be used in the shared memory and has no data type restrictions!

iceoryx_utils/platform/win/include/iceoryx_utils/platform/wait.hpp
elfenpiff: Where do we use it?

elfenpiff (changes requested):

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
marthtz: Done.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Agreed. Should we just call it `removeProcessWithoutLock`?
As @budrus said it's dangerous if it's not called from a method that already holds the lock. How should we address this? Add a comment in the description or in the method's name?

iceoryx_utils/README.md
marthtz: Done

iceoryx_utils/README.md
marthtz: Fixed.

iceoryx_utils/README.md
marthtz: Removed from readme. Is there any code to be removed as well?

iceoryx_utils/README.md
marthtz: Fixed!

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
```diff
@@ -206,7 +213,42 @@ class ProcessManager : public ProcessManagerInterface
...
+    /// @param [in] shudownLog Defines the logging detail.
+    /// @return Returns true if the sent kill signal was successful.
+    bool requestShutdownOfProcess(const RouDiProcess& process,
+                                  ShutdownPolicy shutdownPolicy,
+                                  ShudownLog shudownLog) noexcept;
+
+    /// @brief Checks if the given process has terminated.
+    /// @param [in] process The process to be checked.
+    /// @return True, if the process has terminated.
+    bool isProcessGone(const RouDiProcess& process) noexcept;
```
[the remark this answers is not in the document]
marthtz: It definitely is! Fixed.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: I'm with @elfenpiff and would keep it as it is.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
marthtz: That's right. I renamed it and all related variables, incl. the command line option.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
```diff
@@ -206,7 +213,42 @@ class ProcessManager : public ProcessManagerInterface
                     const uint64_t sessionId,
                     const version::VersionInfo& versionInfo) noexcept;
 
+    /// @brief Removes the process from the managed client process list, identified by its id.
+    /// @param [in] name The process name which should be removed.
+    /// @return Returns true if the process was found and removed from the internal list.
     bool removeProcess(const ProcessName_t& name) noexcept;
+
+    /// @brief Removes the given process from the managed client process list without taking the list's lock!
```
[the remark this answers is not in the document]
marthtz: That's right. Could be dangerous. See @budrus, @elBoberido comments above. Will be addressed.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
marthtz: What about creating a new issue for that with a separate PR to follow?! That'll help keep to keep our PRs smaller.

iceoryx_utils/platform/win/include/iceoryx_utils/platform/wait.hpp
marthtz: I don't think it's used in iceoryx. The only matches I get are within googletest, which is obviously not including this header.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
elfenpiff: In this case I would disagree. The code changes should be minor and the struct would only need some member without any logic at all.

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
     m_memoryManagerOfCurrentProcess = m_segmentInfo.m_memoryManager;
 }
 
-void ProcessManager::killAllProcesses() noexcept
+void ProcessManager::killAllProcesses(const units::Duration processKillDelay) noexcept
 {
     std::lock_guard<std::mutex> g(m_mutex);
+    cxx::vector<bool, MAX_PROCESS_NUMBER> processStillRunning(m_processList.size(), true);
+    int i = 0;
```
elfenpiff, line 141: Please use `uint64_t` here. Try to avoid types where the size is not stated.

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+
+    auto yield = []() { std::this_thread::sleep_for(std::chrono::milliseconds(250)); };
```
elfenpiff, line 143: 250 is a magic number

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+        processStillRunning[i] = requestShutdownOfProcess(process, ShutdownPolicy::SIG_TERM, ShudownLog::FULL);
+        ++i;
+    }
 
-    for (; itEnd != it; ++it)
+    // wait for process to complete
+    bool allProcessFinished = false;
+    posix::Timer finalKillTimer(processKillDelay);
+
+    yield();
```
elfenpiff, line 158: Why are we waiting here?
I think we are using `yield()` only once so I would recommend writing here directly `std::this_thread::sleep_for(std::chrono::milliseconds(250));`

iceoryx_posh/source/roudi/roudi_process.cpp
elfenpiff: I have a not so strict rule to let all boolean start with words like is, has or like @ithier suggested should!

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+                          << (shutdownPolicy == ShutdownPolicy::SIG_KILL ? "SIGKILL" : "SIGTERM")
+                          << " for unknown reason: ’" << killC.getErrorString() << "'";
+            }
+        }
+        return false;
     }
+    return true;
+}
+
+bool ProcessManager::isProcessTerminated(const RouDiProcess& process) noexcept
```
elfenpiff, line 282: This is naming is weird. This method does not verify if a process is terminated it kills a process inside of `requestShutdownOfProcess`. I would get rid of this method and use `requestShutdownOfProcess` directly.

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+{
+    return !requestShutdownOfProcess(process, ShutdownPolicy::SIG_TERM, ShudownLog::NONE);
```
elfenpiff, line 284: You made a Bob. A `t` is missing in `ShudownLog::NONE`

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -278,23 +419,33 @@ bool ProcessManager::removeProcess(const ProcessName_t& name) noexcept
     auto it = m_processList.begin();
     while (it != m_processList.end())
```
elfenpiff, line 420: This should be a ranged based loop like
`for(auto & process : m_processList) {...`
Which would reduce the code and would make it even more readable

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,154 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
+                    LogWarn() << "Process ID " << process.getPid() << " named '" << process.getName()
+                              << "' is still running after SIGKILL was sent " << processKillDelay.seconds<int>()
+                              << " seconds ago. RouDi is ignoring this process.";
+                }
+                ++i;
+            }
+        }
+    }
+
+    auto it = m_processList.begin();
```
elfenpiff, line 228: What happens when `m_processList` is empty. Then `it` would point to `m_processList.end()` since it is equal to `begin` in this case and `removeProcess` would segfault since we are accessing an invalid element.

iceoryx_posh/source/roudi/roudi_process.cpp
elBoberido: the posix timer is schizophrenic and has a good and a bad part. This one uses the good part, which should be refactored out as `CountdownTimer` or `DeadlineTimer`

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
elBoberido: we already have something which we could use, it's the `RouDiConfig` ;)

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
elBoberido: how about passing a reference to a lock guard to this function, then it's impossible to call this without locking a mutex beforehand.
This should maybe be a rule of thumb when we have functions which need to be called with a locked mutex. Let the type system help us, memories are like tears in the rain ;)

iceoryx_posh/source/roudi/roudi_process.cpp
elBoberido: the method checks if `it` is `iterator.end()` so nothing bad happens. But you are right, it's not obvious from the function call

iceoryx_posh/source/roudi/roudi_process.cpp
elBoberido: nope, the list is mutated in the loop

iceoryx_utils/README.md
```diff
@@ -69,12 +72,26 @@ queue which is thread-safe when combined with `smart_lock`. To learn more about
 |:-----------------------:|:--------:|:--------------:|:------------|
 |`ActiveObject`       | i | X | Active object base skeleton implementation inspired by [Prefer Using Active Objects Instead Of Naked Threads](https://www.drdobbs.com/parallel/prefer-using-active-objects-instead-of-n/225700095)  |
 |`FiFo`               | i |   | Single Producer, Single Consumer Lock Free FiFo |
+|`smart_lock`         | i |   | Creates arbitrary thread safe constructs which then can be used like smart pointers. If some STL type should be thread safe use the smart_lock to create the thread safe version in one line. Based on some ideas presented in [Wrapping C++ Member Function Calls](https://stroustrup.com/wrapper.pdf) |
```
elBoberido, line 75: could you please restore the alphabetic ordering?

elBoberido (changes requested):

iceoryx_posh/source/roudi/roudi_process.cpp
mossmaurice: I think the name `StopWatch` fits best ;-)

marthtz: Ok, so what do you guys think? Keep it for now and refactor later when a countdown timer is implemented or use some sleeps and chrono right now?

elBoberido: @mossmaurice as stop watch counts up, that timer counts down ;)

iceoryx_posh/source/roudi/roudi_process.cpp
elfenpiff: @elBoberido @mossmaurice by definition a range based loop like:
`for(auto & element : list)` is expanded to
```cpp
  auto && __range = range_expr ; 
  for (auto __it = begin(__range), __e = end(__range); __it != __e; ++__it) { 
    element_decl = *it; 
    statement 
  } 
```

Which is in the end the construct we are using here.
In the end you could translate this while loop one to one into:
```cpp
for(auto it = m_processList.begin(); it != m_processList.end(); ++it)
```

Which is exactly what a range based for loop would expand to according to the standard.
Take a look at the technical explanation: https://en.cppreference.com/w/cpp/language/range-for
And the more user friendly one: https://arne-mertz.de/2015/07/new-c-features-stdbeginend-and-range-based-for-loops/

elfenpiff: @elBoberido I think you confused it with a vector. If a vector is modified then the iterator become invalid but a list can be modified without invalidating the iterator.

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -278,23 +419,33 @@ bool ProcessManager::removeProcess(const ProcessName_t& name) noexcept
...
         {
-            m_portManager.deletePortsOfProcess(name);
-
-            m_processIntrospection->removeProcess(it->getPid());
-
-            // delete application
-            it = m_processList.erase(it);
-
-            LogDebug() << "New Registration - removed existing application " << name;
+            if (removeProcess(it))
```
elfenpiff, line 425: This is a bug. If you iterate through a list and remove elements you have to increment the iterator first and then remove the element otherwise you have undefined behavior

elfenpiff (changes requested):
resolve bug in `ProcessManager::registerProcess`

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Fixed.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: That's right. Removed.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Fixed.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Of course.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Agreed.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: I agree with @elBoberido 's definition :) 
Timer refactoring is #337.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: Give processes some time to execute the terminate before checking on them. Code has been re-ordered.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: @elfenpiff are you sure it's a bug? The stackoverflow accepted answer mentions the following valid code to do this: `items.erase(i++);  // alternatively, i = items.erase(i);`

That's what `removeProcess(it)` does:
`processIter = m_processList.erase(processIter); // delete application`

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: So, you're saying we can also use a range based loop?!
But we don't want to do a  `++it` after each iteration but `auto it = m_processList.begin();` instead.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
marthtz: Checkout latest commit.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
marthtz: Added a lock guard reference.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: See below. Added lock guard reference.

iceoryx_posh/source/roudi/roudi_process.cpp
elBoberido: @elfenpiff I would even accept a slap bet that your proposal results in a segmentation fault ;)

elfenpiff: @elBoberido you are right this would lead to a segfault ...

marthtz: So, this one can be dismissed?!

elBoberido: @elfenpiff no worries
@marthtz yep

iceoryx_posh/source/roudi/roudi_process.cpp
elBoberido: This is again not a bug. Hidden in `removeProcess(it)` there is a `it = list.erase(it);` like @marthtz mentioned. It's just not obvious from the caller side. Maybe this could be solved differently but I've currently no good idea how.

marthtz: I'm ready to dismiss it too! ;)

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
```diff
@@ -206,7 +213,39 @@ class ProcessManager : public ProcessManagerInterface
...
+        SIG_KILL
+    };
+
+    enum class ShutdownLog
+    {
+        NONE,
+        FULL
+    };
+
+    /// @brief Kills the given process in m_processList with the given signal.
```
ithier, line 240: A bit misleading... it's not killing unless `shutdownPolicy == ShutdownPolicy.SIG_KILL`.

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -126,21 +134,151 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
 
-    // send SIGTERM to all running applications
-    typename ProcessList_t::iterator it = m_processList.begin();
-    const typename ProcessList_t::iterator itEnd = m_processList.end();
+    auto awaitProcessTermination = [&]() {
+        bool shouldCheckProcessState = true;
+        finalKillTimer.resetCreationTime();
 
-    for (; itEnd != it; ++it)
+        while (!haveAllProcessesFinished && !finalKillTimer.hasExpiredComparedToCreationTime())
```
ithier, line 149: It's still a bit hard for me to quickly grasp what's happening here again after forgetting... Maybe I'm just dumb.
Could we add some short comments or self-documenting functions for dumb people like me? e.g. What is this while loop doing ? Waiting for the applications to shut themselves down ?

ithier (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
MatthiasKillat: Can the list not be wrapped in a smart lock to avoid this? In most cases it is best that the data structure itself only provides access protected by mutex. If this is not easily possible (requires much refactoring, or undesired for some technical reason) the approach suggested by @elBoberido is the next best thing I guess (we could unfortunately pass an unreleated mutex in this lock_guard...).

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
```diff
@@ -206,7 +213,39 @@ class ProcessManager : public ProcessManagerInterface
                     const uint64_t sessionId,
                     const version::VersionInfo& versionInfo) noexcept;
 
+    /// @brief Removes the process from the managed client process list, identified by its id.
+    /// @param [in] name The process name which should be removed.
+    /// @return Returns true if the process was found and removed from the internal list.
     bool removeProcess(const ProcessName_t& name) noexcept;
+
+    /// @brief Removes the given process from the managed client process list without taking the list's lock!
+    /// @param [in] lockGuard This method has to be called within a lock guard context. By providing this lock guard it
```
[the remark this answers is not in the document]
MatthiasKillat: Should probably be mentioned that the lock is associated with the  process list.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -32,10 +32,12 @@ void CmdLineParser::parse(int argc, char* argv[], const CmdLineArgumentParsingMo
                                       {"log-level", required_argument, nullptr, 'l'},
                                       {"ignore-version", required_argument, nullptr, 'i'},
                                       {"unique-roudi-id", required_argument, nullptr, 'u'},
+                                      {"compatibility", required_argument, nullptr, 'c'},
+                                      {"final-kill-time", required_argument, nullptr, 'f'},
```
elBoberido, line 36: shouldn't this be `kill-delay`

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -56,14 +58,19 @@ void CmdLineParser::parse(int argc, char* argv[], const CmdLineArgumentParsingMo
...
                       << std::endl;
             std::cout << "                                  off: no check" << std::endl;
             std::cout << "                                  major: same major version " << std::endl;
             std::cout << "                                  minor: same minor version + major check" << std::endl;
             std::cout << "                                  patch: same patch version + minor check" << std::endl;
             std::cout << "                                  commitId: same commit ID + patch check" << std::endl;
             std::cout << "                                  buildDate: same build date + commId check" << std::endl;
+            std::cout << "-k, --kill-delay <UINT>           Sets the delay when RouDi kills the apps hard, if they"
+                      << std::endl;
+            std::cout << "                                  have't responded after the first soft kill, in seconds."
```
elBoberido, line 71: how about `graceful termination request` instead of `soft kill`?

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -142,6 +147,22 @@ void CmdLineParser::parse(int argc, char* argv[], const CmdLineArgumentParsingMo
             }
             break;
         }
+        case 'f':
```
elBoberido, line 150: this should probably be `k` now

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
```diff
@@ -49,15 +49,35 @@ class RouDi
         DEFER_START
     };
 
+    struct RoudiStartupParameters
```
elBoberido, line 52: at one day, this needs to go to `RouDiConfig` and also made available in the TOML config

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
elfenpiff: I agree with @MatthiasKillat . I would also explore and try to wrap the `smart_lock` around this list. Is this possible @marthtz

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
```diff
@@ -131,7 +131,14 @@ class ProcessManager : public ProcessManagerInterface
                          const uint64_t sessionId,
                          const version::VersionInfo& versionInfo) noexcept;
 
-    void killAllProcesses() noexcept;
+    /// @brief Kills all registered processes. First try with a SIGTERM and if they have not terminated after
+    /// finallKillTime they are killed with SIGKILL. If RouDi doesn't have sufficient rights to kill the process, the
+    /// process is considered killed.
+    /// @param [in] processKillDelay RouDi On termination RouDi kills the applications and watches for the specified
```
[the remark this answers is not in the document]
marthtz: That's right!

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
```diff
@@ -131,7 +131,14 @@ class ProcessManager : public ProcessManagerInterface
                          const uint64_t sessionId,
                          const version::VersionInfo& versionInfo) noexcept;
 
-    void killAllProcesses() noexcept;
+    /// @brief Kills all registered processes. First try with a SIGTERM and if they have not terminated after
+    /// finallKillTime they are killed with SIGKILL. If RouDi doesn't have sufficient rights to kill the process, the
```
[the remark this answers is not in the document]
marthtz: Of course.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
marthtz: Done.

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
marthtz: Renamed all kills.

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: That's true.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
marthtz: Being explicit now.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
marthtz: Most likely :D

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
marthtz: That's probably also valid for some other predefined values (e.g. discovery interval, etc.), isn't it?

dkroenke: I would see this topic in a separate PR/issue. I created it here: https://github.com/eclipse/iceoryx/issues/353

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
elBoberido: @elfenpiff then this has to be massively refactored or a recursive mutex has to be used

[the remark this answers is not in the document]
elfenpiff: Okay then I would say leave it ... I didn't see that coming.

iceoryx_posh/source/roudi/roudi_process.cpp
ithier: So... this was resolved but still no comment in code ?

elBoberido (approved):

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi_process.hpp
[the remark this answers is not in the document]
sculpordwarf: ``removeProcess`` is private and only called twice in ``killAllProcesses``. ``killAllProcesses`` already takes the lock. Making a mutex recursive, because you can`t manage the calls correctly in your private functions is not the right way in my opinion. If you are afraid of accidentally use your own private function wrong, I would recommend to use a subfunction in ``killAllProcesses`` via lambda in c++ and define it after the lock to make it even or clear:
```
void ProcessManager::killAllProcesses(const units::Duration finalKillTime) noexcept
{
    std::lock_guard<std::mutex> g(m_mutex);
    auto removeProcess = [](ProcessList_t::iterator& processIter) 
    {
       if (processIter != m_processList.end())
       {
          m_portManager.deletePortsOfProcess(processIter->getName());
          m_processIntrospection->removeProcess(processIter->getPid());
          processIter = m_processList.erase(processIter); // delete application
          return true;
       }
       return false;
    };
    cxx::vector<bool, MAX_PROCESS_NUMBER> processStillRunning(m_processList.size(), true);
    int i = 0;
    .
    .
    .
}
```

dkroenke (approved):

budrus:
@ithier @elfenpiff LGTM?

elfenpiff (approved):

iceoryx_posh/source/roudi/roudi_process.cpp
marthtz: I added a comment just before `while`, that's why it's resolved. Check out second to last commit 1e130ed.

ithier (approved):

Issue #252, opened 2020-08-19
Building blocks in iceoryx


dkroenke:
Brief feature description

For a better modularization and customization of the iceoryx ports, we need to have building blocks where the customer can build the ports with his own flavor. For example we want to make the queue in the receiver port portable so that the customer can either have a Sofi or a MPMC queue in it.

We want to do the following points:

- [x] 1: Implement NewType Pattern for UniqueID 
- [x] 2: use UniqueID implementation in building Blocks 
- [x] 3: C Language binding
- [x] 4: Modern C++ API implementation
- [x] 5: runnable abstraction #349
- [x] 6: Modern C++ API proposal - cxx::unique_ptr 
- [x] 7: Modern C++ API proposal - Typed Publisher 
- [x] 8: Modern C++ API proposal - Typed Subscriber 
- [x] 9: Replacement of old ports in iceoryx with new building block ones
- [x] 10: Modern C++ API Proposal - Untyped
- [x] 11: Top layer integration tests #378 
- [x] 12: API documentation #283 
- [x] 13: callbacks in the new APIs #350

dkroenke:
@budrus Maybe you want to add something if you want.

mossmaurice:
This is a duplicate of #25?

dkroenke:
@mossmaurice whats is duplicated?

mossmaurice:
Reopening as @marthtz and myself are still working on this.

PR #335, opened 2020-11-04
iox #252 example applications using new api

ithier:
In this PR:
* Adds example applications working with new pub/sub api
* Renames `receive()` method in subscriber API to `take()`
* Automatically default-constructs types used by the `TypedPublisher`
* Cleanup leftover findings from previous PR

Next Steps:
* Following this PR, the old pub/sub API and its examples will be removed


iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
```diff
@@ -49,36 +49,44 @@ int main(int argc, char* argv[])
     {
         ++ct;
 
-        // Retrieve a typed sample from shared memory.
-        // Sample can be held until ready to publish.
+        // API Usage #1
```
mossmaurice, line 52: ❤️

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+void subscriberHandler(iox::popo::WaitSet& waitSet)
+{
+    // run until interrupted
+    while(!killswitch)
+    {
+        auto triggeredConditions = waitSet.wait();
+        for(auto& condition : triggeredConditions)
+        {
+            auto subscriber = dynamic_cast<iox::popo::TypedSubscriber<Position>*>(condition);
```
mossmaurice, line 44: `static_cast` should be enough? We don't have RTTI available, so a dynamic_cast doesn't make sense here.

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,84 @@
...
+    iox::popo::WaitSet waitSet{};
+    waitSet.attachCondition(typedSubscriber);
+    waitSet.attachCondition(shutdownGuard);
+
+    // delegate handling of received data to another thread
+    std::thread subscriberHandlerThread(subscriberHandler, std::ref(waitSet));
+    subscriberHandlerThread.join();
+
+    // clean up
+    waitSet.detachAllConditions();
```
mossmaurice, line 81: The user does not have to care about this. It is being called in the d'tor of `WaitSet`

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
```diff
@@ -0,0 +1,83 @@
...
+    iox::popo::WaitSet waitSet{};
+    waitSet.attachCondition(untypedSubscriber);
+    waitSet.attachCondition(shutdownGuard);
+
+    // delegate handling of received data to another thread
+    std::thread untypedSubscriberThread(subscriberHandler, std::ref(waitSet));
+    untypedSubscriberThread.join();
+
+    // clean up
+    waitSet.detachAllConditions();
```
mossmaurice, line 80: Not needed

iceoryx_examples/icedelivery/topic_data.hpp
```diff
@@ -23,6 +23,7 @@ struct CounterTopic
 
 struct Position
 {
+    Position() = default;
```
mossmaurice, line 26: Add `noexcept`

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_subscriber.hpp
```diff
@@ -93,6 +100,7 @@ class BaseSubscriber : public Condition
     virtual bool hasTriggered() const noexcept override;
 
   protected:
+    BaseSubscriber() = default; // Required for testing.
```
mossmaurice, line 103: Add `noexcept`

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -722,6 +722,8 @@ void ProcessManager::monitorProcesses() noexcept
                 // @todo Check if ShmManager and Process Manager end up in unintended condition
                 m_portManager.deletePortsOfProcess(processIterator->getName());
 
+                /// @todo #25 Need to delete condition variables used by terminating processes...
```
mossmaurice, line 725: Good catch!

mossmaurice (changes requested):

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
ithier: We should update the example in wait_set.hpp too then.

mossmaurice: You mean this snippet from `wait_set.hpp:89`?
````
/// myWaitSet.detachCondition(guardCond);
/// myWaitSet.detachCondition(mySubscriber2);
/// myWaitSet.detachCondition(mySubscriber1);
````

File: iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp [hunk 8]
```diff
@@ -49,36 +49,44 @@ int main(int argc, char* argv[])
...
         {
             auto& sample = result.get_value();
-            sample->x = ct * 1.1;
-            sample->y = ct * 1.1;
-            sample->z = ct * 1.1;
+            sample->x = ct;
+            sample->y = ct;
+            sample->z = ct;
             sample.publish();
         }
```
elBoberido, line 64: Would an else branch make sense, e.g. printing an error?
Just asking because I think it would make sense with the lambda to show how to access the error, e.g.
```
typedPublisher.load()
    .and_then(...)
    .or_else(...) { // do something with the eroor}
```

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
dkroenke, line 1: 2020

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
```diff
@@ -49,36 +49,44 @@ int main(int argc, char* argv[])
...
+        //  * Sample can be held until ready to publish.
+        //
         auto result = typedPublisher.loan();
         if (!result.has_error())
         {
             auto& sample = result.get_value();
-            sample->x = ct * 1.1;
-            sample->y = ct * 1.1;
-            sample->z = ct * 1.1;
+            sample->x = ct;
```
elBoberido, line 60: could you do a placement new, like with the lambda example, just to keep the differences at a minimun

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
dkroenke, line 1: 2020

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
```diff
@@ -49,36 +49,44 @@ int main(int argc, char* argv[])
...
-        // allocation. The first argument of the callable must be T*, this will point to the memory
-        // allocation.
+        // API Usage #4
+        //  * Provide a callable that will be used to populate the loaned sample.
+        //  * The first argument of the callable must be T* and is the location that the callable should
+        //      write its result to.
+        //
         typedPublisher.publishResultOf(getVehiclePosition, ct);
-        typedPublisher.publishResultOf([](Position* allocation) { new (allocation) Position(0, 0, 0); });
+        typedPublisher.publishResultOf([&ct](Position* allocation) { new (allocation) Position(ct, ct, ct); });
```
elBoberido, line 89: 💖

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_publisher.hpp
```diff
@@ -59,6 +59,12 @@ class BasePublisher : public PublisherInterface<T>
     ///
     uid_t getUid() const noexcept;
 
+    ///
+    /// @brief getServiceDescription Get the service description of the subscriber.
```
dkroenke, line 63: not publisher?

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_publisher.hpp
```diff
@@ -103,12 +109,13 @@ class BasePublisher : public PublisherInterface<T>
...
 
   protected:
+    BasePublisher() = default; // Required for testing.
     BasePublisher(const capro::ServiceDescription& service);
 
   private:
     ///
     /// @brief The PublisherSampleDeleter struct is a custom deleter in functor form which releases loans to a sample's
-    /// underlying memory chunk via a publishers publisher port.
+    /// underlying memory chunk via a publisher's publisher port.
```
dkroenke, line 118: sounds weird ;-)

iceoryx_examples/icedelivery/iox_publisher_untyped_modern.cpp
```diff
@@ -29,25 +29,40 @@ static void sigHandler(int f_sig [[gnu::unused]])
...
     untypedPublisher.offer();
 
     float_t ct = 0.0;
     while (!killswitch)
     {
+        ++ct;
+
+        // API Usage #1
+        //  * Loaned sample can be held until ready to publish
+        auto maybeSample = untypedPublisher.loan(128);
```
elBoberido, line 49: in the typed publisher this was called `result` instead of `maybeSample`. Should it be harmonized?

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_subscriber.hpp
```diff
@@ -76,14 +76,21 @@ class BaseSubscriber : public Condition
     bool hasNewSamples() const noexcept;
 
     ///
-    /// @brief receive Receive the next sample if available.
-    /// @return
-    /// @details Sample is automatically released when it goes out of scope.
+    /// @brief hasMissedSamples Check if samples have been missed since the last subscriber call.
```
dkroenke, line 79: what is meant with subscriber call? You mean subscribe() right?

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
elBoberido: Can the condition also be the `shutdownGuard`? I guess no? I have to get up to date with the new API :)

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
```diff
@@ -0,0 +1,83 @@
...
+int main()
+{
+    // register sigHandler for SIGINT
+    signal(SIGINT, sigHandler);
+
+    // initialize runtime
+    iox::runtime::PoshRuntime::getInstance("/iox-ex-subscriber-untyped-modern");
+
+    // initialized subscribers
+    iox::popo::UntypedSubscriber untypedSubscriber({"Odometry", "Position", "Vehicle"});
```
elBoberido, line 67: do we really need `untyped` in the variable name?

elBoberido (commented):
Just as a side note. I might sound like a broken record, but this PR got a bit large, just from the amount of changed files. There are also unrelated changes, so this would have made sense to split it up into two or three PRs, e.g. just the example and the refactoring. I know from myself that this happens quite easily, just wanted to recollect the "small PR" mantra whenever I see a large PR :)

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
ithier: The idea was to present different alternatives that work.

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
ithier: Why not ? It is in fact the star of the show in this example.

iceoryx_examples/icedelivery/iox_publisher_untyped_modern.cpp
ithier: Optional vs. expected.

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
ithier: It can, which is why I did the dynamic_cast.
The point of the cast is to check the type of condition. I can't use a static cast.
I'm not too happy with this approach though, anyone got an alternative ?

mossmaurice: Yeah, it can be a `shutdownGuard` / `GuardCondition` IIRC. Should we introduce a flag in the base class for this? Or fast forwarding, if we have additional new Conditions, how do we differentiate here?

ithier: One way would be to use unique ids.
The waitset could return a list of unique IDs then the user can determine which conditions are triggered by matching them.

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
ithier: Yes, let's remove it in the future.

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
elBoberido: Yes, I wanted to propose something like that
```
#include <iostream>

class Base {
public:
    uint32_t getConditionId() {
        return m_conditionId;
    }
    
    template <typename T>
    static uint32_t getConditionIdForType() {
        static uint32_t conditionId = conditionIdGenerator();
        return conditionId;
    }
    
protected:
    Base (uint32_t conditionId) : m_conditionId(conditionId) {}
        
private:
    static uint32_t conditionIdGenerator() {
        static uint32_t conditionId = 0; 
        conditionId++;
        return conditionId;
    }
    
    uint32_t m_conditionId {0};
};

class Foo : public Base {
public:
    Foo() : Base(Base::getConditionIdForType<Foo>()) {}
};

class Bar : public Base {
public:
    Bar() : Base(Base::getConditionIdForType<Bar>()) {}
};

int main() {
    Foo foo;
    
    Base* base = &foo;
    
    if (base->getConditionId() == Base::getConditionIdForType<Foo>()) {
        std::cout << "It's a Foo" << std::endl;
    }
    
    if (base->getConditionId() != Base::getConditionIdForType<Bar>()) {
        std::cout << "It's not a Bar" << std::endl;
    }
    
    return EXIT_SUCCESS;
}
```
I couldn't resist and write some code. Something like that might be feasible. In the end we reinvent RTTI. For the TypeId we could use the newtype pattern.

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
elBoberido: I just assumed the idea was to have different alternatives how to get the sample, not how to initialize it. By directly assigning instead of doing a placement new, it's harder to tell the actual differences in how the different styles are supposed to be used.

iceoryx_examples/icedelivery/iox_publisher_untyped_modern.cpp
elBoberido: Why the difference? I guess this loan can also fail

ithier: Ah I got confused (looked at the wrong line in my IDE). this call is in fact an expected and should be named result.

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
ithier: @budrus Thoughts ?

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // run until interrupted
+    while(!killswitch)
+    {
+        auto triggeredConditions = waitSet.wait();
+        for(auto& condition : triggeredConditions)
+        {
+            auto untypedSubscriber = dynamic_cast<iox::popo::UntypedSubscriber*>(condition);
+            if(untypedSubscriber)
+            {
+                untypedSubscriber->take().and_then([](iox::cxx::optional<iox::popo::Sample<const void>>& maybeSample){
```
elfenpiff, line 46: We need to simplify this. I would recommend a signature like:
```cpp
ReceivedSample<const T> take() noexcept;
```

The `ReceivedSample` should then get an API like (it is just a draft to get an idea):
```cpp
template<typename T>
class ReceivedSample {
    void and_then(std::function<void(T&)> ); // is only called when the expected contains a value and the optional contains a value
    // is called when the expected contains an error or the optional does not have a value. 
    // The user has to determine the case and therefore the ReceivedSample is provided here as reference
    void or_else(std::function<void(ReceivedSample &)>); 
    bool hasError(); // returns true if the expected contains an error
    bool hasValue();; // returns true if the expected contains a value as well as the optional
};
```

@ithier @elBoberido @budrus what do you think?

iceoryx_examples/icedelivery/iox_publisher_untyped_modern.cpp
```diff
@@ -29,25 +29,40 @@ static void sigHandler(int f_sig [[gnu::unused]])
...
     {
+        ++ct;
+
+        // API Usage #1
+        //  * Loaned sample can be held until ready to publish
+        auto result = untypedPublisher.loan(128);
+        if(!result.has_error())
+        {
+            auto& sample = result.get_value();
+            new (sample.get()) Position(ct, ct, ct);
```
budrus, line 53: Here we could explicitly mention that the placement new is needed for untyped. For the typed API, the placement new is not the normal way IMHO. And therefore returning a default constructed sample with the typed API is what I prefer

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <iostream>
+
+bool killswitch = false;
+iox::popo::GuardCondition shutdownGuard;
+
+static void sigHandler(int f_sig [[gnu::unused]])
+{
+    // caught SIGINT, now exit gracefully
+    killswitch = true;
+    shutdownGuard.trigger(); // unblock waitsets
```
budrus, line 33: cool. That's exactly the guard use case 🥇

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // run until interrupted
+    while(!killswitch)
+    {
+        auto triggeredConditions = waitSet.wait();
+        for(auto& condition : triggeredConditions)
+        {
+            auto subscriber = dynamic_cast<iox::popo::TypedSubscriber<Position>*>(condition);
+            if(subscriber)
+            {
+                subscriber->take().and_then([](iox::cxx::optional<iox::popo::Sample<const Position>>& maybePosition){
```
budrus, line 47: As discussed it has a complicated touch for me 
take -> and_then -> has_value -> value -> now we are ready to read. The question is if we can make it simpler without loosing the strength of expected and optional

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
budrus: This is an idea we should discuss. This could hide the expected and optional behind the scenes, so having a simpler API but not loosing the benefits of expected and optional which are both valid choices for the take

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
```diff
@@ -15,15 +15,15 @@
 #ifndef IOX_POSH_POPO_BASE_PUBLISHER_INL
 #define IOX_POSH_POPO_BASE_PUBLISHER_INL
```
budrus, line 17: The new C API is in the root directory. We need some repo refactoring later, but should we have it already symmetric. I.e. this would be in a iceoryx_binding_modern_cpp directory parallel to iceoryx_binding_c ?

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/typed_publisher.inl
```diff
@@ -36,7 +43,10 @@ inline uid_t TypedPublisher<T, base_publisher_t>::getUid() const noexcept
 template <typename T, typename base_publisher_t>
 inline cxx::expected<Sample<T>, AllocationError> TypedPublisher<T, base_publisher_t>::loan() noexcept
 {
-    return base_publisher_t::loan(sizeof(T));
+    // Call default constructor here to ensure type is immediately ready to use by the caller.
+    // There is a risk that the type will be re-constructed by the user (e.g. by using a placement new in
+    // publioshResultOf(), however the overhead is considered to be insignificant and worth the additional safety.
+    return std::move(base_publisher_t::loan(sizeof(T)).and_then([](Sample<T>& sample) { new (sample.get()) T(); }));
```
budrus, line 49: Agreed

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_subscriber.hpp
```diff
@@ -76,14 +76,21 @@ class BaseSubscriber : public Condition
     bool hasNewSamples() const noexcept;
 
     ///
-    /// @brief receive Receive the next sample if available.
-    /// @return
-    /// @details Sample is automatically released when it goes out of scope.
+    /// @brief hasMissedSamples Check if samples have been missed since the last take() call.
```
budrus, line 79: I guess it is since the last `hasMissedSamples()` call. Only when you check with this call the state is reset and not with every take. @mossmaurice right, or?

iceoryx_posh/include/iceoryx_posh/popo/modern_api/typed_publisher.hpp
```diff
@@ -27,6 +27,7 @@ template <typename T, typename base_publisher_t = BasePublisher<T>>
 class TypedPublisher : public base_publisher_t
 {
     static_assert(!std::is_void<T>::value, "Type must not be void. Use the UntypedPublisher for void types.");
+    static_assert(std::is_default_constructible<T>::value, "The TypedPublisher requires default-constructable types.");
```
budrus, line 30: Nice

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
```diff
@@ -15,15 +15,15 @@
...
 namespace iox
 {
 namespace popo
 {
-// ============================== BasePublisher ============================== //
-
 template <typename T, typename port_t>
-inline BasePublisher<T, port_t>::BasePublisher(const capro::ServiceDescription&)
-/// @todo #25 : m_port(iox::runtime::PoshRuntime::getInstance().getMiddlewareSender(service, ""))
+BasePublisher<T, port_t>::BasePublisher(const capro::ServiceDescription& service)
```
FerdinandSpitzschnueffler, line 25: Why did you remove ```inline```?

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
budrus: The `getValue()` is missing. and this would return a T& ?

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
budrus: How about this approach:

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
ithier: Added #340 to explore this further.

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
ithier: Will be addressed in #340

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
ithier: @elfenpiff @elBoberido @mossmaurice Thoughts ?

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
ithier: Good question!

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_subscriber.hpp
mossmaurice: Neither reset on `hasMissedSamples()` nor with `take()`, it just calls `empty` on the `ChunkQueuePopper` and the underlying queue. I guess you mixed it up with `hasLostChunksSinceLastCall()`? Should this be replaced?

mossmaurice: My bad... `ChunkQueuePopper::hasOverflown()` is called here and the flag is reset with every call.

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
mossmaurice: Yeah, makes sense. As stated in our main Readme we need to split up the CMake target `iceoryx_posh` anyway at some point. I would call it `iceoryx_binding _cpp`.

budrus: As we have these new API on the master branch now and also the new examples. People who clone this maybe already use it. Is this cool? I'm not sure which stability we reached with the new APIs. Having things more hidden or flagged as "experimental" would be a more conservative way. Or we have to point more to releases and make clear that the master branch is work in progress and maybe unstable

mossmaurice: I don't think we need to explicitly mention this, it's the same with every project master = unstable, release tag = stable

iceoryx_examples/icedelivery/iox_publisher_untyped_modern.cpp
```diff
@@ -29,25 +29,42 @@ static void sigHandler(int f_sig [[gnu::unused]])
...
+        {
+            auto& sample = result.get_value();
+            // In the untyped API, the returned sample is a void pointer, therefore the data must be constructed
+            // in place.
+            new (sample.get()) Position(ct, ct, ct);
+            sample.publish();
+        }
+
+        // API Usage #2
+        // * Loan sample and provide logic for it immediately with via a lambda
```
FerdinandSpitzschnueffler, line 60: remove 'with'

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
ithier: This shall be addressed in #341

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
ithier: I will leave it as is for now.

iceoryx_posh/include/iceoryx_posh/popo/modern_api/base_subscriber.hpp
ithier: OK so @budrus is correct, will update the doc.

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
ithier: I don't think this API should be moved into its own directory. 
It's not a "binding", it's part of the core iceoryx library.

budrus: The C binding isn't either, or? Please align with @elfenpiff who made the C "binding". Similar things should have the same naming conventions and should be found in similar places. Like the client libraries in ROS with rclcpp, rclpy and now rclrs for rust

iceoryx_examples/icedelivery/CMakeLists.txt
```diff
@@ -72,6 +75,28 @@ set_target_properties(iox-ex-publisher-untyped-modern PROPERTIES
     POSITION_INDEPENDENT_CODE ON
 )
 
+add_executable(iox-ex-subscriber-typed-modern ./iox_subscriber_typed_modern.cpp)
+target_link_libraries(iox-ex-subscriber-typed-modern
+    iceoryx_posh::iceoryx_posh
+)
+set_target_properties(iox-ex-subscriber-typed-modern PROPERTIES
```
dkroenke, line 82: You can add the properties to multiple targets at once.
```
set_target_properties(iox-ex-subscriber-untyped-modern iox-ex-subscriber-typed-modern PROPERTIES
    CXX_STANDARD_REQUIRED ON
    ...
)
```

You can even add "iox-ex-publisher-typed-modern" and "iox-ex-publisher-untyped-modern"  from above to save more.

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
elBoberido: I would keep the native API/language within posh or at least not call it binding. When I see a C++ binding, I would wonder in which language iceoryx is written

iceoryx_examples/icedelivery/CMakeLists.txt
ithier: This was copying the structure used by the old examples. 
I can change it though if you prefer. I don't think it really matters though...

ithier:
@elBoberido @elfenpiff @budrus @mossmaurice @dkroenke @FerdinandSpitzschnueffler @marthtz 
Good to merge ?

iceoryx_examples/icedelivery/CMakeLists.txt
dkroenke: According to a popular committer here i would say: I have no strong feelings about this ;-)

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
budrus: OK fine. Then the modern C++ API will replace old one that is in posh. And the C API is some kind of extra API that will be more like an additional module. Having different APIs side by side would be another option but now understand your thoughts

budrus (approved):

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // initialize runtime
+    iox::runtime::PoshRuntime::getInstance("/iox-ex-subscriber-typed-modern");
+
+    // initialized subscribers
+    iox::popo::TypedSubscriber<Position> typedSubscriber({"Odometry", "Position", "Vehicle"});
+    typedSubscriber.subscribe();
+
+    // set up waitset
+    iox::popo::WaitSet waitSet{};
+    waitSet.attachCondition(typedSubscriber);
```
MatthiasKillat, line 73: Terminolgy-wise I also think the subscriber *being* a condition does not feel right (it reads like this). The subscriber corresponds to a condition in a natural way, the condition of receiving data or not.
One could also argue to just call the function on the waitSet attach, not attachCondition for this reason.

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void subscriberHandler(iox::popo::WaitSet& waitSet)
+{
+    // run until interrupted
+    while(!killswitch)
+    {
+        auto triggeredConditions = waitSet.wait();
+        for(auto& condition : triggeredConditions)
+        {
+            auto untypedSubscriber = dynamic_cast<iox::popo::UntypedSubscriber*>(condition);
```
MatthiasKillat, line 43: As discussed above, I think we need to get rid of the need of the dynamic_cast here by an improved design.

iceoryx_examples/icedelivery/iox_subscriber_untyped_modern.cpp
```diff
@@ -0,0 +1,80 @@
...
+            if(untypedSubscriber)
+            {
+                untypedSubscriber->take().and_then([](iox::cxx::optional<iox::popo::Sample<const void>>& maybeSample){
+                    if(maybeSample.has_value())
+                    {
+                        auto position = reinterpret_cast<const Position*>(maybeSample->get());
```
MatthiasKillat, line 49: I am not sure and did not try it out, but it could define and_then to be a template<T> which will do the cast before passing it to the lambda. But then the lambda would need to take a T& (and, as previously suggested, only be called by and_then when there actually is data). 
Type inference would most likely not work based on the lambda type I think, so we would call something like  ```untypedSubscriber->take().and_then<T>(...)```. 

Having some API call such as ```Sample<T> typedSample = untypedSample.as<T>;``` which converts the untyped ```Sample<void>``` to a typed one may also be nice, if it can be done in a lightweight manner (i.e. no heavy copies). This would require some further thought to get it right (usable, efficient) though.

We then could do something like ```untypedSubscriber->take().as<T>().and_then(...)```. Maybe build it in take() as an alternative, i.e. ```untypedSubscriber->take<T>().and_then(...)```. The idea is to get a typed sample as early as possible to be able to use the typed API.

Note that this is not fully thought through, just some ideas. I would prefer to avoid having the user do too many reinterpret or dynamic casts (much boilerplate and error-prone).

iceoryx_posh/include/iceoryx_posh/popo/modern_api/typed_publisher.hpp
MatthiasKillat, line 30: Why does this have to be the case? Are we not constructing the data into raw memory, given arbitrary constructor arguments? It may be a limitation right now, but I imagine this due to the problem itself, only due to the current design. I may be wrong though, there may be a deeper underlying reason.

iceoryx_posh/include/iceoryx_posh/popo/modern_api/typed_subscriber.hpp
```diff
@@ -42,7 +42,8 @@ class TypedSubscriber : public base_subscriber_t
     void unsubscribe() noexcept;
 
     bool hasNewSamples() const noexcept;
-    cxx::expected<cxx::optional<Sample<const T>>, ChunkReceiveError> receive() noexcept;
+    bool hasMissedSamples() noexcept;
+    cxx::expected<cxx::optional<Sample<const T>>, ChunkReceiveError> take() noexcept;
```
MatthiasKillat, line 46: This expected of optional is supposed to change in the future right? Since it is a little cumbersome to use. Maybe as a specific sample type combining the two...not sure what would be the best approach.

MatthiasKillat:
I think certain functional approaches to work with the samples can be improved in the future and the ```expected<optional<T>,E>``` sample types are cumbersome. However this can be adapted in a future pull request.

The dynamic_casts to check for guard conditions cannot be a long term solution from my point of view.

MatthiasKillat (approved):

iceoryx_posh/include/iceoryx_posh/popo/modern_api/typed_subscriber.hpp
ithier: See https://github.com/eclipse/iceoryx/issues/340#issuecomment-724149533 
Feedback appreciated!

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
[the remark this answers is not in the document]
ithier: I have already implemented this and will submit the PR as soon as this one is merged :)

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,81 @@
...
+void subscriberHandler(iox::popo::WaitSet& waitSet)
+{
+    // run until interrupted
+    while(!killswitch)
+    {
+        auto triggeredConditions = waitSet.wait();
+        for(auto& condition : triggeredConditions)
+        {
+            auto subscriber = dynamic_cast<iox::popo::TypedSubscriber<Position>*>(condition);
+            if(subscriber)
```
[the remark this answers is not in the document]
ithier: I don't think hard-coding a check for specific conditions is the way to go.
What if users implement their own conditions ?
We should discuss this in more detail in #341

iceoryx_examples/icedelivery/iox_subscriber_typed_modern.cpp
```diff
@@ -0,0 +1,81 @@
...
+    killswitch = true;
+    shutdownGuard.trigger(); // unblock waitsets
+}
+
+void subscriberHandler(iox::popo::WaitSet& waitSet)
+{
+    // run until interrupted
+    while(!killswitch)
+    {
+        auto triggeredConditions = waitSet.wait();
```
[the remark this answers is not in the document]
ithier: Please raise your points in #341 !

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
```diff
@@ -49,36 +49,47 @@ int main(int argc, char* argv[])
...
```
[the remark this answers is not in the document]
ithier: I don't understand this one. What do you mean by "called loan to get the location"?

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
```diff
@@ -49,36 +49,47 @@ int main(int argc, char* argv[])
...
         });
 
-        // Simple copy-and-publish. Useful for smaller data types.
-        auto position = Position(ct * 111.111, ct * 111.111, ct * 111.111);
+
+        // API Usage #3
+        //  * Basic copy-and-publish. Useful for smaller data types.
+        //
+        auto position = Position(ct, ct, ct);
         typedPublisher.publishCopyOf(position);
```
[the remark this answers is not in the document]
ithier: Got a suggestion for an alternative name ?

iceoryx_posh/include/iceoryx_posh/popo/modern_api/typed_publisher.hpp
ithier: The answer lies here: https://github.com/eclipse/iceoryx/pull/335#discussion_r518087695

In short, apparently the automotive world expects samples to be default-constructable and not doing so will be confusing.

iceoryx_examples/icedelivery/iox_publisher_typed_modern.cpp
elBoberido: @budrus just to clarify, so when I don't have a default c'tor for whatever reason, I cannot use the typed API?

Issue #240, opened 2020-08-13
Integration test for posh-runtime class


nihalchari:
This is in continuation with Iox #177 unit test for posh runtime.
Test to be covered 

    findService
        reoffer service
        mix up of AnyService, AnyInstance, AnyEvent
    offerService
        call offer service twice
    stopOfferService
        call with invalid serviceDescription like {AnyService, AnyInstance, AnyEvent}
        call with non existing serviceDescription
        call stop offer service twice

dkroenke:
@nihalchari this can be closed or?

nihalchari:
yes @dkroenke , ill close it

nihalchari:
closed by mistake, reopening,

ithier:
@nihalchari Why was this reopenned ?

nihalchari:
@ithier   This is in continuation with https://github.com/eclipse/iceoryx/pull/214  unit test for posh runtime., this issue was closed by mistake, actually 177 should hvae closed

Issue #148, opened 2020-06-19
File-based Configuration for DDS Gateways


ithier:
## Brief feature description

Enable the services to be forwarded via DDS gateways via a Toml configuration rather than needing to be dynamically discovered.

## Detailed information

Pros:
* Quicker setup time
* Lower the chance of missing data points

ithier:
Forgot to mention - this is required for the dds gateway using cyclonedds as dds service discovery is not implemented there.

budrus:
So this is the workaround for https://github.com/eclipse-cyclonedds/cyclonedds/issues/42

ithier:
This issue was resolved in #162.

PR #342, opened 2020-11-09
iox-#240 addedintegration test for posh runtime

ithier:
Code contributed by @nihalchari


iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
 
+TEST_F(RoudiFindService_test, OfferServiceWithDefaultDescription)
+{
+    senderRuntime->offerService(iox::capro::ServiceDescription());
+    this->InterOpWait();
+    iox::runtime::InstanceContainer instanceContainer;
+    receiverRuntime->findService(iox::capro::ServiceDescription(), instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    senderRuntime->stopOfferService(iox::capro::ServiceDescription());
```
dkroenke, line 51: Why is here stoppoffer called when nothing is tested afterwards?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
     ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
 }
 
+TEST_F(RoudiFindService_test, OfferServiceWithDefaultDescription)
+{
+    senderRuntime->offerService(iox::capro::ServiceDescription());
+    this->InterOpWait();
+    iox::runtime::InstanceContainer instanceContainer;
+    receiverRuntime->findService(iox::capro::ServiceDescription(), instanceContainer);
+
```
dkroenke, line 49: Please verify that the returned instanceContainer has the default ServiceDescription in it

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+    this->InterOpWait();
+    InstanceContainer instanceContainer;
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
+}
+
+TEST_F(RoudiFindService_test, FindServiceMultipleTimes)
```
dkroenke, line 94: Name could be more concrete because here the same ServiceDescription is found multiple times

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+{
+    senderRuntime->offerService({"service1", "instance1"});
+
+    this->InterOpWait();
+    InstanceContainer instanceContainer;
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(2u));
+    ASSERT_EQ(instanceContainer.at(0), instanceContainer.at(1));
```
dkroenke, line 104: I know it's maybe nitpicking but here should be explicitly verified that the ServiceDescription {"service1", "instance1"} is found at the first and second element in the instancecontainer.

dkroenke (changes requested):

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
     ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
 }
 
+TEST_F(RoudiFindService_test, OfferServiceWithDefaultDescription)
+{
+    senderRuntime->offerService(iox::capro::ServiceDescription());
```
budrus, line 45: We try to write our test according to the AAA pattern. Please also do it that way that makes it easier for all the others
[AAA pattern](https://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern/)

[iceoryx example](https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp#L134-L148)

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+    iox::runtime::InstanceContainer instanceContainer;
+    receiverRuntime->findService(iox::capro::ServiceDescription(), instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    senderRuntime->stopOfferService(iox::capro::ServiceDescription());
+}
+
+TEST_F(RoudiFindService_test, ReOfferSingleMethodServiceSingleInstance)
+{
+    senderRuntime->offerService({"service1", "instance1"});
```
budrus, line 56: This is maybe more than one test? StopOffering and ReOfffering?

budrus (changes requested):

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff, line 45: I think this makes no sense. A default constructed ServiceDescription is invalid, see https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/source/capro/service_description.cpp line 99.

How can a service offer an invalid service? I think we should generate an error in this case and not accept an offer.
Additionally, I think I would not like to find invalid service descriptions with `findService`. Then I would get the impression that there is something substantially wrong in the system.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
     ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
 }
 
+TEST_F(RoudiFindService_test, OfferServiceWithDefaultDescription)
```
elfenpiff, line 43: @nihalchari Please remember to question our behavior. The goal of the unit tests is not only to verify if something is working you should also verify if it makes sense that iceoryx behaves in this way. I think here you discovered a behavior which makes no sense, therefore it would make sense to correct it and verify it with a unit test.

- [ ] I think `offerService` should be declined when any of the following member values are set in the service description
```cpp
static constexpr uint16_t InvalidID = 0u;
static const IdString InvalidIDString{"0"};
static constexpr uint16_t AnyService = 0xFFFFu;
static constexpr uint16_t AnyInstance = 0xFFFFu;
static constexpr uint16_t AnyEvent = 0xFFFFu;
static constexpr char AnyServiceString[]{"65535"};
static constexpr char AnyInstanceString[]{"65535"};
static constexpr char AnyEventString[]{"65535"};
```
 - [ ] test the new `offerService` behavior with valid and invalid values

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff, line 56: Does it make sense to offer a service with an invalid event id?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    senderRuntime->stopOfferService(iox::capro::ServiceDescription());
+}
+
+TEST_F(RoudiFindService_test, ReOfferSingleMethodServiceSingleInstance)
+{
+    senderRuntime->offerService({"service1", "instance1"});
+    this->InterOpWait();
+    iox::runtime::InstanceContainer instanceContainer;
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
```
elfenpiff, line 59: The lines with `findService` and verifying the contents of the `instanceContainer` could be skipped since you already covered that in a previous tests. Try to follow the pattern
1. arrange
2. act
3. assert

https://wiki.c2.com/?ArrangeActAssert

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff, line 43: Could you please adjust your unit test names so that they reflect what you are testing and what you are expecting. Something like: OfferedServiceWithValidServiceDescriptionCanBeFound

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+    senderRuntime->offerService(iox::capro::ServiceDescription());
+    this->InterOpWait();
+    iox::runtime::InstanceContainer instanceContainer;
+    receiverRuntime->findService(iox::capro::ServiceDescription(), instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    senderRuntime->stopOfferService(iox::capro::ServiceDescription());
+}
+
+TEST_F(RoudiFindService_test, ReOfferSingleMethodServiceSingleInstance)
```
elfenpiff, line 54: Please adjust the test name to something more understandable. Are you expecting that a reoffered service can be found? Maybe a name like: ReofferedServiceWithValidServiceDescriptionCanBeFound

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
+
+    senderRuntime->stopOfferService({"service1", "instance1"});
+    this->InterOpWait();
+    instanceContainer.clear();
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(0u));
```
elfenpiff, line 69: Are you testing that stop offer works here? If so please create a separate test for it!
Please try to avoid to test multiple things in one tests.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+
+    senderRuntime->offerService({"service1", "instance1"});
+    this->InterOpWait();
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
+}
+
+TEST_F(RoudiFindService_test, OfferSameServiceMultipleTimes)
```
elfenpiff, line 79: The name should state what you are expecting. You expect that offering the same service multiple times fails.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+{
+    senderRuntime->offerService({"service1", "instance1"});
```
elfenpiff, line 81: I think `offerService` should return a bool since it is failing when there is the service already offered.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+    ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
+}
+
+TEST_F(RoudiFindService_test, FindServiceMultipleTimes)
+{
+    senderRuntime->offerService({"service1", "instance1"});
+
+    this->InterOpWait();
+    InstanceContainer instanceContainer;
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
```
elfenpiff, line 100: What is the goal of this test? That `findService` works even after it was called?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -180,6 +244,35 @@ TEST_F(RoudiFindService_test, StopOfferMultiMethodServiceSingleInstance)
     receiverRuntime->findService({"service3", "instance1"}, instanceContainer);
     ASSERT_THAT(instanceContainer.size(), Eq(0u));
 }
+
+TEST_F(RoudiFindService_test, StopOfferServiceSingleMultipleCalls)
```
elfenpiff, line 248: Please let the name reflect what you test and what you expect.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -180,6 +244,35 @@ TEST_F(RoudiFindService_test, StopOfferMultiMethodServiceSingleInstance)
...
+    senderRuntime->stopOfferService({"service1", "instance1"});
+    this->InterOpWait();
+
+    iox::runtime::InstanceContainer instanceContainer;
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+    ASSERT_THAT(instanceContainer.size(), Eq(0u));
+}
+
+
+TEST_F(RoudiFindService_test, StopNonExistingService)
```
elfenpiff, line 263: I think the test here is to verify that stopping a non existing service does not stop an existing service?!

elfenpiff (changes requested):
- [ ] Fix review findings
- [ ] adjust `offerService` behavior (do not allow invalid ServiceDescriptions)
- [ ] `offerService` should return a bool signalling if it was successful
- [ ] add more test cases to verify the behavior fully

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
dkroenke: This is some kind of ServiceDescription 'light' which you can use for service discovery in gateways where the protocol does not necessarily need  an event id. According to the implementation of Servicedescription you can construct it with a service and instance id only. In this case is the bool flag m_hasServiceOnlyDescription set. 

From my point of view it is valid to have it because not everyone needs this event id and we give more freedom to the users.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: i want to check for a sequence here, i.e. offer -> stopOffer -> offere, then may be naming  OfferStopofferOfferServiceSequence might hold good?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: i dint split test here as i m checking for scenario  where service might get stopped and start again

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: no, i am checking for a sequence, start -> stop -> start

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: i think removing of findService here will be against the scope of the test, where the scope is to check sequence(start->stop-startOfferService), yeah i should change the name but

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff: @nihalchari I would prefer reoffer, I think the name already implies a sequence like offer -> stopOffer -> offer. And what is the expected outcome? Or how do you verify that the sequence works?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff: @nihalchari I think @budrus meant a test for:
- that stopOffer works
- that offered and service after it was stopped works (which is your current test)

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff: Could you please then add a stopOffer is working test.
And then you could remove this `ASSERT_THAT` since you verified it in the stopOffer test

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: okay, test for stopOffer works is present at https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp #148

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: yeah,  test for stopOffer works is present at https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp line148

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: OfferService returns void, and when i offer same service again its add service to concurrent::FiFo, there is no check i guess

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: currently it returns void

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: yeah

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: test is to check stopping non existing services will silently ignore the request, additionally check was introduced to confirm that it does not stop existing service

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: yeah, not required in this case... i added it as a habit over some other tests where it was required to release some resources acquired for testing purpose.. i will remove it from here

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: okay, i will verify explicitly

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: okay , @elfenpiff  for adjusting offerService behavior not to allow invalid ServiceDescriptions, i will create a separate issue and link here, and adapt test OfferServiceWithDefaultDescription accordingly and disable for now, is it fine?

elfenpiff: @nihalchari This is fine by me.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: done

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: @elfenpiff  i created new issue https://github.com/eclipse/iceoryx/issues/359 and fixed other review points.
Could u please review

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,63 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
     ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
 }
 
+
+///@todo offere service should return bool signalling success/failure
```
dkroenke, line 44: please add a link to the iox issue here for tracking purposes

elfenpiff:
@nihalchari could you please resolve all conversation where you solved the criticized issue.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: added

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: name changed

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: done

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: created issue for same #359

nihalchari: done

nihalchari:
@elfenpiff i think i do not have rights to resolve conversations, i replied by comments that it is resolved

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
budrus: I'm not sure if this is something that makes sense. You have the same instance twice in the instanceContainer. This is because we just push but don't clear or check the container if this instance is already in? 
Having the same instance twice in the instance container is very dangerous IMHO. We should fix this!

budrus (changes requested):

nihalchari:
@budrus, yes instance container needs to be cleared when entering the findService() call.

dkroenke:
> 
> 
> @budrus, yes instance container needs to be cleared when entering the findService() call.

Do you want to do it in this PR or should it be tracked it in a issue?

nihalchari:
> 
> 
> > @budrus, yes instance container needs to be cleared when entering the findService() call.
> 
> @nihalchari Do you want to do it in this PR or should it be tracked it in a issue?

@dkroenke , i will do it in this PR

budrus (approved):

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -484,6 +484,7 @@ cxx::expected<Error> PoshRuntime::findService(const capro::ServiceDescription& s
         return cxx::error<Error>(Error::kMQ_INTERFACE__REG_UNABLE_TO_WRITE_TO_ROUDI_MQ);
     }
 
+    instanceContainer.clear();
```
dkroenke, line 487: It is better not to clear the container because it could contain other elements given by the user. 

I would prefer to check if the given ServiceDescription is already exist in the instanceContainer. It is a cxx::vector, so you can easily iterate through it with a range-based loop. This should be done before sendRequestToRouDi(). If the ServiceDescription already exist then just print a Logmessage like ("ServiceDescription already exist") in Debuglevel and return a cxx::success.

elfenpiff: @dkroenke you are right it is somehow a mess here. What should we do if we get a container which is already full of garbage, then we are unable to return anything. 

I would recommend to change the signature of `findService` to: 
```cpp
cxx::expected<InstanceContainer, Error> PoshRuntime::findService(const capro::ServiceDescription& serviceDescription) noexcept
```
and return the `InstanceContainer` directly otherwise we have a lot of issues to deal with, like
1. What to do if the provided `InstanceContainer` is full
2. If we found a service should we look in the `InstanceContainer` if its already present.
3. What is with entries which are out of date?

@nihalchari could you please adapt `findService` so that it is returning the `InstanceContainer` directly?

nihalchari: yeah @elfenpiff  i could adapt this change, but ill create separate issue and continue. i will link it here

budrus: In ara::com the instance container is provided as parameter to avoid returning it by value and doing a copy. But yes, we see that it has some drawbacks. As the iceoryx API does not need to follow ara::com here, we can change it. I see only the options a) clear or b) the proposal of @elfenpiff
Are we sure that compilers of today with things like return value optimization are good enough that b) is not way more expensive than a)? The instance container has a max capacity of 50 and an ID string which is a cxx::string<100> I guess. So we are talking about 5 kBytes

elfenpiff: @budrus I would say we are save here that the compiler will optimize that away. Take a look at the minimalistic example: https://godbolt.org/z/z565x6 . There seems to be only one allocation in assembler (line 32: add rsp, 3888) so that in the end the instance container would actually live in the caller (here main) and is allocated just once and then reused.

nihalchari: Yes, container will be reused, but in case of multiple calls container will be overwritten, then it will be callers job to duplicate/store data if necessary… in this case, can it be still achieved by option a) by @budrus  where we clear the container and document in deoxygen, that if user needs initial data from container he needs to store it before passing to findService()

Even if iceoryx API does not follow ara:com, does is it make sense that findService is oneshot call which should only return available instance of requested service at the time of call.
@elfenpiff  @budrus

elfenpiff: @nihalchari what do you mean with multiple calls?
In the godbolt example the container was not overridden. With every iteration it was recreated again since the end of the loop is the end of the scope and then the scope starts again with the next iteration. But the assembler code showed us that the memory position stays the same and that the compiler optimizes it behind the scene in a way that the memory is reused.

To the oneshot call question: At the moment findService does exactly what you described and if you return the InstanceVector instead of using a reference it will still perform the same thing. I think I do not understand question here either!? :kangaroo:

budrus: My expectation from a container that was filled/returned by the finService call is that it contains the instances that are now available. Having already some instances in the container and further will be added by the call does not feel right. Therefore I'm not for the option that @dkroenke proposed. Clearing the vector is the workaround to fulfill this contract. But this is also a bit dirty. With the assumption that the compiler will do a good job, the option @elfenpiff proposed seems for me the best solution

nihalchari: okay, if clearing container is dirty then i will go with option proposed by @elfenpiff

nihalchari: created new issue #387 and disabled failing test

budrus (dismissed):

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -40,6 +40,70 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
+    senderRuntime->offerService({"service1", "instance1"});
+    this->InterOpWait();
+
+    receiverRuntime->findService({"service1", "instance1"}, instanceContainer);
+
+    ASSERT_THAT(instanceContainer.size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.begin(), Eq(IdString("instance1")));
+}
+
+///@todo #387 findService to return InstanceContainer directly
```
elfenpiff, line 87: Please do not add `@todo` into the code. In my opinion this is a minor change but it would maybe break the API. Therefore, I suggest the following: 
Please adapt the poshRuntime method in this pull request to:
```cpp
    cxx::expected<Error, InstanceContainer> findService(const capro::ServiceDescription& serviceDescription) noexcept;
```

Deprecate the old `findService` method like:
```cpp
    [[gnu::deprecated]] cxx::expected<Error> findService(const capro::ServiceDescription& serviceDescription,
                                     InstanceContainer& instanceContainer) noexcept;
```
and test it here.  

In my opinion an integration- or unit-test story does not only contain writing test. It also contains fixing bugs and adjust behavior if you discover them (if it is possible and feasible). If we create a new issue for every misbehavior/bug we find we have to track too much and the gain is minor since we could also fix it right away.

elfenpiff (changes requested):
Adjust `findService` as suggested.

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
elfenpiff: @nihalchari if you need some assistance I would be happy to help you!

nihalchari: okay @elfenpiff  ill do it in this PR, and yes i will ask for assistance if required 👍

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: The lines with findService and verifying the contents of the instanceContainer are removed

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: This will be fixed with #359

iceoryx_posh/source/runtime/posh_runtime.cpp
nihalchari: changed signature as discussed in this PR, #387 is closed

dkroenke (approved):

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
```diff
@@ -62,14 +62,24 @@ class PoshRuntime
...
+    /// @deprecated #240
     /// @brief find all services that match the provided service description
     /// @param[in] serviceDescription service to search for
     /// @param[out] instanceContainer container that is filled with all matching instances
     /// @return cxx::expected<Error> Error, if any, encountered during the operation
     /// Error::kPOSH__SERVICE_DISCOVERY_INSTANCE_CONTAINER_OVERFLOW : Number of instances can't fit in instanceContainer
     /// Error::kMQ_INTERFACE__REG_UNABLE_TO_WRITE_TO_ROUDI_MQ : Find Service Request could not be sent to RouDi
-    cxx::expected<Error> findService(const capro::ServiceDescription& serviceDescription,
-                                     InstanceContainer& instanceContainer) noexcept;
+    [[gnu::deprecated]] cxx::expected<Error> findService(const capro::ServiceDescription& serviceDescription,
```
dkroenke, line 72: @nihalchari You can remove this deprecated function because we will not need it for the upcoming 1.0 release

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -33,52 +33,102 @@ TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
...
     this->InterOpWait();
+
+    auto instanceContainer = receiverRuntime->findService({"service1", "instance1"});
+
+    ASSERT_THAT(instanceContainer.get_value().size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.get_value().begin(), Eq(IdString("instance1")));
+}
+
+
+///@todo #359 offereService should return bool signalling success/failure
```
dkroenke, line 44: Can you maybe integrate this follow-up in this PR here directly? It should be not much work to change the signature.

dkroenke (changes requested):

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
nihalchari: okay, i will remove it

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: okay

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -31,54 +31,118 @@ class RoudiFindService_test : public RouDiServiceDiscoveryTest
 
 TEST_F(RoudiFindService_test, OfferSingleMethodServiceSingleInstance)
 {
+    auto isServiceOffered = senderRuntime->offerService({"service1", "instance1"});
+    this->InterOpWait();
+
+    auto instanceContainer = receiverRuntime->findService({"service1", "instance1"});
+
+    ASSERT_THAT(instanceContainer.get_value().size(), Eq(1u));
```
dkroenke, line 39: The API of the expected changed with the last big PR from @elfenpiff. Now `get_value()` is marked as deprecated and will not be accepted by the strict build in iceoryx CI. 
Instead please use `value()` in every occurrence.

dkroenke (changes requested):

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -507,10 +507,19 @@ PoshRuntime::findService(const capro::ServiceDescription& serviceDescription) no
 }
 
 
-void PoshRuntime::offerService(const capro::ServiceDescription& serviceDescription) noexcept
+bool PoshRuntime::offerService(const capro::ServiceDescription& serviceDescription) noexcept
 {
+    if (serviceDescription.getServiceIDString() == iox::capro::InvalidIDString
```
dkroenke, line 512: I would propose to do the check not here because this is something which is somehow only related to the servicedescription.  We will also in other cases need a function which checks if the servicedescription is invalid.

Please create in service_description.hpp a function called `isvalid` which maybe return a bool. Inside of that you can check then against the invalid values. Check will be easier here because internally the two friends serviceid and instanceid are represented as u_int values which you can easy compare without converting.

dkroenke (changes requested):

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -289,6 +289,12 @@ Interfaces ServiceDescription::getSourceInterface() const noexcept
     return m_interfaceSource;
 }
 
+bool ServiceDescription::isValid() const noexcept
+{
+    return !(m_serviceString == iox::capro::InvalidIDString || m_serviceID == iox::capro::AnyService
+             || m_instanceString == iox::capro::InvalidIDString || m_instanceID == iox::capro::AnyInstance);
```
nihalchari, line 295: For default ServiceDescription, which is an invalid case, m_serviceID and m_instanceId (represented as u_int ) are initialized with InvalidID which is zero, on other hand for valid ServiceDescription also... say {"service1", "instance1"}  m_serviceID and m_instanceId are also deduced to zero, hence need to check for m_serviceString and m_instanceString against iox::capro::InvalidIDString instead of checking against InvalidID with m_serviceID and m_instanceId.

dkroenke: You need to check if the Servicedescription is full blown or not. If the reduced one is constructed then a bool member `m_hasServiceOnlyDescription` is set and can be checked with `hasServiceOnlyDescription()`.

Also when you construct a ServiceDescription{"service1", "instance1"} then the members m_serviceID and m_instanceId are set. 
The ctor of ServiceDescription(IDString, IDString) is redirecting to  ServiceDescription(IDString, IDString, InvalidString). And there is also a conversion from string to uint_16t happening. Therefore you only need to check the uint16_t members. Take a look in the constructors in service_description.cpp

dkroenke (changes requested):

iceoryx_posh/source/capro/service_description.cpp
nihalchari: Conversion is happening if stringIsNumber, in case of {"service1", "instance1"}, stringIsNumber() returns false, and m_serviceID and m_instanceId are set to InvalidID.

Also i dint get why i need to check for Servicedescription is full blown or not, I guess we are not checking for event, do u want to consider for future case where event may consider ??

dkroenke: Yes you are right, i somehow didn't want to accept that the term "service1" can't be represented as int, shame on me. :-) Then it is the safest way to compare the strings. 
What if you are internally using full blown service descriptions like the ara::com API? This could cause trouble for the case that someone constructs a full blown servicedescription with `m_hasServiceOnlyDescription = false` and someone sets the `m_eventID` to zero. It is maybe a special case but can rarely happen.

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
```diff
@@ -67,16 +67,17 @@ class PoshRuntime
...
     /// Error::kPOSH__SERVICE_DISCOVERY_INSTANCE_CONTAINER_OVERFLOW : Number of instances can't fit in instanceContainer
     /// Error::kMQ_INTERFACE__REG_UNABLE_TO_WRITE_TO_ROUDI_MQ : Find Service Request could not be sent to RouDi
-    cxx::expected<Error> findService(const capro::ServiceDescription& serviceDescription,
-                                     InstanceContainer& instanceContainer) noexcept;
+    cxx::expected<InstanceContainer, Error> findService(const capro::ServiceDescription& serviceDescription) noexcept;
 
     /// @brief offer the provided service, sends the offer from application to RouDi daemon
     /// @param[in] serviceDescription service to offer
-    void offerService(const capro::ServiceDescription& serviceDescription) noexcept;
+    /// @return bool, if service is offered returns true else false
```
elfenpiff, line 79: Could you please write when this method will return false. E.g. when you are providing an invalid service description.

iceoryx_posh/source/capro/service_description.cpp
nihalchari: okay, let me share my understanding.
If full blown service description is constructed and `m_eventID` sets to zero then it should be considered as invalid case and should be ignored, on other hand for description like {"service1", "instance1"}, where `m_hasServiceOnlyDescription` will be true and m_`eventID` will be zero should be considered as valid case. This is what do you mean? @dkroenke

dkroenke: @nihalchari Yes correct

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -418,13 +419,22 @@ cxx::expected<Error> PoshRuntime::findService(const capro::ServiceDescription& s
...
+    if (serviceDescription.isValid())
+    {
+        capro::CaproMessage msg(
+            capro::CaproMessageType::OFFER, serviceDescription, capro::CaproMessageSubType::SERVICE);
+        m_applicationPort.dispatchCaProMessage(msg);
+        return true;
+    }
+    LogWarn() << "Could not offer service " << serviceDescription.getServiceIDString() << ","
+              << " ServiceDescription is invalid\n";
+    return false;
```
dkroenke, line 437: Can you please extend the tests to the new added functionality and resolve the codecov warnings?

dkroenke (changes requested):
Please resolve codecov warnings

iceoryx_posh/source/runtime/posh_runtime.cpp
nihalchari: Tests are already added for newly added functionality, but that is under integration tests not module

nihalchari:
> 
> 
> Please resolve codecov warnings

Resolved

dkroenke (approved):

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -32,54 +32,154 @@ class RoudiFindService_test : public RouDiServiceDiscoveryTest
...
+    auto instanceContainer = receiverRuntime->findService({"service1", "instance1"});
+    ASSERT_THAT(instanceContainer.value().size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.value().begin(), Eq(IdString_t("instance1")));
+
+    instanceContainer = receiverRuntime->findService({"service1", "instance1"});
+    ASSERT_THAT(instanceContainer.value().size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.value().begin(), Eq(IdString_t("instance1")));
 }
 
 TEST_F(RoudiFindService_test, DISABLED_OfferMultiMethodServiceSingleInstance_PERFORMANCETEST42)
```
elBoberido, line 150: would it be possible to enable this test?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -32,54 +32,154 @@ class RoudiFindService_test : public RouDiServiceDiscoveryTest
...
     senderRuntime->offerService({"service1", "instance2"});
     senderRuntime->offerService({"service1", "instance3"});
-
     this->InterOpWait();
-    InstanceContainer instanceContainer;
     InstanceContainer instanceContainerExp;
     InitContainer(instanceContainerExp, {"instance1", "instance2", "instance3"});
 
-    receiverRuntime->findService({"service1", "65535"}, instanceContainer);
+    auto instanceContainer = receiverRuntime->findService({"service1", "65535"});
```
elBoberido, line 179: Shouldn't this be `AnyInstanceString`?

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
```diff
@@ -32,54 +32,154 @@ class RoudiFindService_test : public RouDiServiceDiscoveryTest
...
+    instanceContainer = receiverRuntime->findService({"service1", "instance1"});
+    ASSERT_THAT(instanceContainer.value().size(), Eq(1u));
+    ASSERT_THAT(*instanceContainer.value().begin(), Eq(IdString_t("instance1")));
 }
 
 TEST_F(RoudiFindService_test, DISABLED_OfferMultiMethodServiceSingleInstance_PERFORMANCETEST42)
 {
     senderRuntime->offerService({"service1", "instance1"});
     senderRuntime->offerService({"service2", "instance1"});
     senderRuntime->offerService({"service3", "instance1"});
```
elBoberido, line 154: Could copy this test and make this `"instance1_1"`, `"instance2_1"` and `"instance3_1"` or `"instance_a"`, `"instance_b"`, `"instance_c"`? Just to catch the error that findService returns the instances of the wrong serviceId

elBoberido (commented):
Just a few minor things. Looks good overall

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: enabled

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: yeah, changed to AnyInstanceString

iceoryx_posh/test/integrationtests/service_discovery/test_roudi_findservice.cpp
nihalchari: current implementation for mentioned scenario doesn't return error, it returns empty container

elBoberido: This is not regarding returning an error, but catching an error. As far as I can tell, there is no check for `findService({"service2", "instance1"})` looking in `service1` for that instance. Since the instances have all the same name, this is an untested error case. If the instances have different names in one test, it could be checked that the returned container is not empty.

nihalchari: @elBoberido sorry but i dint get the test you are asking for, what i got is 

    senderRuntime->offerService({"service1", "instance1_1"});
    senderRuntime->offerService({"service2", "instance2_1"});
    senderRuntime->offerService({"service3", "instance3_1"});

and then what combination you want in findService()  ??
is it 
findService({"service1", "instance1_1"});
findService({"service2", "instance2_1"});
findService({"service3", "instance3_1"});

and then check for each returned container for non empty?

elBoberido: let's assume we have the following services offered
```
senderRuntime->offerService({"service1", "instance1"});
senderRuntime->offerService({"service2", "instance2"});
```
a weird bug in `findService` could ignore the passed service ID and `findService({"service2", "instance1"});`would return `instance1`. When all instances have the same name, this error case is not found. So this would be a `OfferMultiMethodServiceWithDistinctSingleInstance`

nihalchari: i think that is not the case, following test passes

    senderRuntime->offerService({"service1", "instance1"});
    senderRuntime->offerService({"service2", "instance2"});
    this->InterOpWait();

    auto instanceContainer = receiverRuntime->findService({"service1", "instance1"});
    ASSERT_THAT(instanceContainer.value().size(), Eq(1u));
    ASSERT_THAT(*instanceContainer.value().begin(), Eq(IdString_t("instance1")));

    instanceContainer = receiverRuntime->findService({"service2", "instance1"});
    ASSERT_THAT(instanceContainer.value().size(), Eq(0u));

    instanceContainer = receiverRuntime->findService({"service2", "instance2"});
    ASSERT_THAT(instanceContainer.value().size(), Eq(1u));
    ASSERT_THAT(*instanceContainer.value().begin(), Eq(IdString_t("instance2")));

elBoberido: I would assume that test to pass, but where can I find it? It seems it's not in this file.

nihalchari: @elBoberido i pushed it now

dkroenke (approved):

elBoberido (approved):

Issue #340, opened 2020-11-09
Simplify API for "expected optionals"


ithier:
## Brief feature description
The current API for expected optionals is a little convoluted, requiring many layers of "unwrapping" before the data can finally be accessed.

## Detailed information
We have this API:
```
    auto result = cxx::expected<cxx::optional<ValueType>>();
    if(!result.has_error())
    {
        auto optional = result.get_value();
        if(optional.has_value())
        {
            auto value = optional.value();
            // What did I want to do again ?
        }
    }
```
And this one (loaning a sample from a publisher):
```
    TypedPublisher<ValueType> typedPublisher{"service", "instance", "event"};
    result = typedPublisher.loan().and_then([](cxx::optional<ValueType> optional){
        if(optional.has_value())
        {
            auto value = optional.value();
            // Now I can finally play the game !
        }
    })
```

These steps are a little tedious to need to go through and can be a little confusing for newcomers.

budrus:
I'm wondering if we need a new issue for this topic or if this is just another pull request for #252 
As this is relevant for 1.0 you would have to assign it to this project etc. Now it is a new issue that is not assigned

ithier:
Adding additional "todo" items within another issue is not great for visibility.
Also, I'm not sure if "creating an issue per feature" fits the github workflow (but I am no expect). 
For example, what happens if a user would like to raise an issue ?

budrus:
It is about issue granularity I guess. And this has currently a wide range. If there is an issue for "Write a new C++ API" and another issue for "improve this specific aspect of the not yet finalized C++ API", I have the feeling that it should be nested but with the single layer of issues they are side by side. Too heterogeneous for me. Sure a user can raise whatever issue but if we implement new features we should have some best practise.

ithier:
My preference would be to have higher granularity issues, i.e. in the "improve this specific aspect of the not yet finalized C++ API" form, and then group these using milestones.
We should definitely have a consensus though.

ithier:
I have a working prototype that might fix this problem.
Behold:
```
subscriber->take()
       .and_then([](iox::popo::Sample<const Position>& position){
            std::cout << "Got value: (" << position->x << ", " << position->y << ", " << position->z << ")" << std::endl;
        })
        .if_empty([]{
            std::cout << "Didn't get a value, but do something anyway." << std::endl;
        });
```

The expected has a specialization for when its type is an optional.
In this case it will:
* automatically retrieve the value from the optional when it is not empty
[...]

PR #352, opened 2020-11-11
Iox #340 simplify expected optionals

ithier:
* adds an `and_then` specialization for `cxx::expected<cxx::optional<T>>` that automatically retrieves the value from non-empty optionals
* the and_then specialization does nothing if the optional is empty
* adds `if_empty()` method to `cxx::expected<cxx::optional<T>>` types that allows users to react to retrieving an empty optional without error

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
+++ b/iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
@@ -347,12 +347,10 @@ expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(ValueType&
 }
 
 template <typename ValueType, typename ErrorType>
-template<typename _ValueType = ValueType, typename std::enable_if<is_optional<_ValueType>::value, int>::type = 0>
+template<typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
 inline expected<ValueType, ErrorType>&
-expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename _ValueType::type&)>& callable) noexcept
+expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename Optional::type&)>& callable) noexcept
 {
-    std::cout << "THIS IS AN OPTIONAL" << std::endl;
-
     if (!this->has_error())
     {
         auto optional = get_value();
@@ -364,6 +362,24 @@ expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename _
 
     return *this;
 }
+
+template <typename ValueType, typename ErrorType>
+template<typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
+inline expected<ValueType, ErrorType>&
+expected<ValueType, ErrorType>::if_empty(const cxx::function_ref<void(void)>& callable) noexcept
+{
+    if (!this->has_error())
+    {
+        auto optional = get_value();
+        if(!optional.has_value())
+        {
+            callable();
+        }
+    }
+
+    return *this;
+}
+
 
 template <typename ValueType, typename ErrorType>
 inline expected<ValueType, ErrorType>&
--- a/iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
+++ b/iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
@@ -347,22 +347,23 @@ expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(ValueType&
 }
 
 template <typename ValueType, typename ErrorType>
-template<typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
+template <typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
 inline const expected<ValueType, ErrorType>&
-expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename Optional::type&)>& callable) const noexcept
+expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename Optional::type&)>& callable) const
+    noexcept
 {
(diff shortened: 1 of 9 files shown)


elBoberido:
@elfenpiff told me yesterday about an idea which makes your proposal more generic, maybe you can discuss with him about that. I'm fine with either solution

ithier:
> @elfenpiff told me yesterday about an idea which makes your proposal more generic, maybe you can discuss with him about that. I'm fine with either solution

@elBoberido I spoke briefly with him about it. I would still like to merge this as a stepping stone towards the generic solution.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -346,6 +346,58 @@ expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(ValueType&
     return *this;
 }
 
+template <typename ValueType, typename ErrorType>
+template <typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
```
mossmaurice, line 350: @elfenpiff Can the 2nd template parameter be omitted as it's already in the hpp? I remember we did in some first version of the `WaitSet`.

mossmaurice (approved):
LGTM

mossmaurice (changes requested):
Fix build error related to 2nd template param in new `expected` methods

elBoberido:
> Fix build error related to 2nd template param in new `expected` methods

and there is also this warning in the Colcon build check
```
The `set-env` command is deprecated and will be disabled on November 16th. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```
Not related to this PR, but we should fix it the next few days.

mossmaurice (approved):

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -248,3 +257,51 @@ TEST_F(expected_test, ExpectedWithErrorConvertsToNullopt)
     optional<int> value = sut.to_optional();
     ASSERT_THAT(value.has_value(), Eq(false));
 }
+
+TEST_F(expected_test, AndThenUnpacksOptionalWhenNonEmptyOptionalValue)
+{
+    auto sut = expected<iox::cxx::optional<int>, float>::create_value(123);
+    MockCallables mocks{};
+    EXPECT_CALL(mocks, onSuccess).Times(1);
```
elfenpiff, line 265: I think you could get rid of `MockCallables` completely, reduce the code and test everything anyway.
Look here:
```cpp
bool andThenWasCalled = false;
sut.and_then([&andThenWasCalled](int& val) {
        andThenWasCalled = true;
        ASSERT_THAT(val, Eq(123));
    });
EXPECT_TRUE(andThenWasCalled);
```

You could also create an integer which is incremented every time you call the the lambda to verify that it is only called once.

elfenpiff (approved):
@ithier If you have time you could maybe look again over the tests but it is mandatory...

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -346,6 +346,58 @@ expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(ValueType&
...
+{
+    return const_cast<expected*>(this)->and_then(callable);
+}
+
+template <typename ValueType, typename ErrorType>
+template <typename Optional, typename std::enable_if<is_optional<Optional>::value, int>::type>
+inline expected<ValueType, ErrorType>&
+expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename Optional::type&)>& callable) noexcept
+{
+    if (!this->has_error())
```
MatthiasKillat, line 363: Can we add a has_value function to the expected in this pull request? I know that it is not strictly needed, but I would prefer to have both (has_error and has_value) or only has_value for consistency.

If we do so, we could use it here.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -346,6 +346,58 @@ expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(ValueType&
...
+template <typename Optional, typename std::enable_if<is_optional<Optional>::value, int>::type>
+inline const expected<ValueType, ErrorType>&
+expected<ValueType, ErrorType>::and_then(const cxx::function_ref<void(typename Optional::type&)>& callable) const
+    noexcept
+{
+    return const_cast<expected*>(this)->and_then(callable);
+}
+
+template <typename ValueType, typename ErrorType>
+template <typename Optional, typename std::enable_if<is_optional<Optional>::value, int>::type>
```
MatthiasKillat, line 359: There is also enable_if_t (C++14) which can be used to shorten this a bit (not required).

MatthiasKillat (approved):

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
mossmaurice: This was already added to our cxx/type_trait.hpp and can be removed once we switched to C++14

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
ithier: Will get it in the next PR.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
ithier: Will get it in a follow-up.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
ithier: Ah, I forgot about this.

PR #360, opened 2020-11-12
Iox #252 replacement of old ports with building blocks

marthtz:
## Pre-Review Checklist for the PR Author

1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php

## Notes for Reviewer
This is about removing all old ports with building blocks. Unfortunately, it turned out this is way too big for one PR. Cleaning-up the introspection with its tests is a big task for itself, it's still missing an overhaul to only store pointers to `PortData`s and not using `PortUser` as members. In addition, introspection throughput calculation has to be re-implemented using a more sophisticated approach (e.g. only calculate it when an introspection is connected).
~~Last, roudi portmanager (previously roudi shm) tests are currently more or less disabled. I'm still working on it.~~

Note: ~~Has to be built with `one-to-many` as `get/setCapacity`hasn't been implemented to multi pusher queues!~~ Resolved, was an error when accidentally calling `subscribe()` with an argument.

Open points:
1. Refactor introspection to remove `PortUser` and templates: #401
1. Implement throughput calculation: #402
1. Introspection clean-up: #414
1. Add tests for new (previously missed) transitions in subscriber port: #406
1. [x] Fix roudi portmanager tests.
1. [x] Resolve merge conflicts with master

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## Post-review Checklist for the Eclipse Committer

1. [x] All checkboxes in the PR checklist are checked or crossed out
[...]

PR #361, opened 2020-11-12
iox-#260 CString100 ... gone for good

### LINKED ISSUE #260: Clean-up string usage
## Brief feature description
Iceoryx uses a variety of strings (`std::string`, `cxx::string<>`, `cxx::CString100`, etc).
Once #253 is merged this mishmash should be cleaned-up.

## Detailed information
Only use `cxx:string<>` or types created from it.

- [x] No std::string in public API

- [ ] No more std::strings in code base
    - [ ] iceoryx_utils
    - [ ] iceoryx_posh
    - [ ] iceoryx_dds

- [x] No more CSring100 in code base

- [ ] Common convention for string aliases. Currently we have:
    - alias without `_t`, e.g. `IdString`
    - alias with `_t`, e.g. `ProcessName_t`
    - alias nested in a class, e.g. `PosixUser::string_t`
    - alias with a constexpr as string length, e.g. `cxx::string<MAX_LENGTH>`
    - alias without a constexpr as string length, e.g. `cxx::string<100>`
    - see also https://github.com/eclipse/iceoryx/pull/373/files#r525922015 and https://github.com/eclipse/iceoryx/pull/373/files#r525922656
    - [ ] cleanup iceoryx_utils
    - [x] cleanup iceoryx_posh
    - [x] cleanup iceoryx_dds

elBoberido:
We are gathered here today in memories to CString100. RIP.

partially closes #260 

Signed-off-by: Mathias Kraus <mathias.kraus@apex.ai>


iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -335,8 +334,7 @@ void RouDi::processMessage(const runtime::MqMessage& message,
         }
         else
         {
-            capro::ServiceDescription service(
-                cxx::Serialization(cxx::CString100(cxx::TruncateToCapacity, message.getElementAtIndex(2))));
+            capro::ServiceDescription service(cxx::Serialization(std::string(message.getElementAtIndex(2))));
```
elBoberido, line 337: to use CString100 was an error in the first place. Changed it back to std::string and made it a problem of the future elBoberido

budrus:
We've come a long long way together, but now it's time to say goodbye

ithier:
Who has the champaigne ? :champagne:

iceoryx_posh/include/iceoryx_posh/runtime/runnable.hpp
```diff
@@ -47,11 +47,11 @@ class Runnable
 
     /// @brief returns the name of the runnable
     /// @return string which contains the runnable name
-    cxx::CString100 getRunnableName() const noexcept;
+    RunnableName_t getRunnableName() const noexcept;
 
     /// @brief returns the name of the process
     /// @return string which contains the process name
-    cxx::CString100 getProcessName() const noexcept;
+    RunnableName_t getProcessName() const noexcept;
```
FerdinandSpitzschnueffler, line 54: Why not ``ProcessName_t``?

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
```diff
@@ -65,7 +65,7 @@ class ProcessIntrospection_test : public Test
 
         introspection.send();
 
-        EXPECT_THAT(m_senderPortImpl_mock->deliverChunk, Eq(1));
+        EXPECT_THAT(m_senderPortImpl_mock->deliverChunk, Eq(1u));
```
FerdinandSpitzschnueffler, line 68: Did we decide to use 1u or 1U? I'm not sure...

iceoryx_posh/include/iceoryx_posh/runtime/runnable.hpp
marthtz: Copy/paste?!

iceoryx_posh/source/runtime/runnable.cpp
```diff
@@ -57,12 +57,12 @@ Runnable& Runnable::operator=(Runnable&& rhs) noexcept
...
 }
 
-cxx::CString100 Runnable::getRunnableName() const noexcept
+RunnableName_t Runnable::getRunnableName() const noexcept
 {
     return m_data->m_runnable;
 }
 
-cxx::CString100 Runnable::getProcessName() const noexcept
+RunnableName_t Runnable::getProcessName() const noexcept
```
marthtz, line 65: Same same?!

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
marthtz: I'd also like to know.
If I remember right MISRA prefers upper case because lower case 'i' and number '1' might look pretty similar depending on font type.

iceoryx_posh/include/iceoryx_posh/runtime/runnable.hpp
elBoberido: c&p mistake

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
elBoberido: so 50% more typing ... and there goes the productivity

iceoryx_posh/include/iceoryx_posh/roudi/introspection_types.hpp
```diff
@@ -61,11 +61,11 @@ const capro::ServiceDescription IntrospectionPortService(INTROSPECTION_SERVICE_I
 /// @todo if future fixed string is aligned to 8 byte, the alignment here can be removed
 struct PortData
 {
-    alignas(8) cxx::CString100 m_name;
-    alignas(8) cxx::CString100 m_caproInstanceID;
-    alignas(8) cxx::CString100 m_caproServiceID;
-    alignas(8) cxx::CString100 m_caproEventMethodID;
-    alignas(8) cxx::CString100 m_runnable;
+    alignas(8) ProcessName_t m_name;
```
ithier, line 64: Should we have a convention for the string types ?
Here we have `IdString` and `RunnableName_t` which are both string aliases.

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
dkroenke: According to MISRA-CPP-2.13.4 and AUTOSAR-M2.13.4 all literal suffixes should not contain lowercase characters. This removes potential ambiguity between “1” (digit 1) and  “l” (letter el) for declaring literals (from: https://rules.sonarsource.com/c/RSPEC-818).

Clang-Tidy offers an automatic check for it: https://clang.llvm.org/extra/clang-tidy/checks/readability-uppercase-literal-suffix.html

iceoryx_posh/source/roudi/roudi.cpp
dkroenke: I leave this comment here to remind the future dkroenke that he reminds the future elBoberido about that problem ;-)

iceoryx_posh/include/iceoryx_posh/roudi/introspection_types.hpp
elBoberido: @ithier we should probably have that for all aliases. From my point of view, it's out of scope for this PR though.

Karsten1987:
Are you coming up with a follow up PR for the RMW interfaces as well? I believe removing CString100 will break the ROS2 rmw compilation.

mossmaurice:
> Are you coming up with a follow up PR for the RMW implementation as well? I believe removing CString100 will break the ROS2 rmw compilation.

We need to talk about that. Till the v0.90 (pre-release for v1.0) that it scheduled to arrive as Christmas present, there will be dozens of breaking changes. So we might want to consider to wrap up the adaption of rmw_iceoryx. Would you stick to a step-by-step approach keeping the master's compatible all the time?

Karsten1987:
It's essentially boiling down to how much attention you want to give to the ROS2 interoperability. I am fine with putting a disclaimer on the RMW master branch, saying that the latest and greatest of iceoryx can't currently be used with ROS2. I could also imagine to simply erase the master branch (or rename the branch to `develop`), so that only tagged release branches are compatible with their respective iceoryx version. 

I've raised that question in the past where we have to decide a little bit on a release strategy for iceoryx as well as the RMW layer.

iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
```diff
@@ -34,7 +34,7 @@ class Publisher_t
     /// @param[in] service Information on service , service, instance, event Id
     /// @param[in] runnableName optional name of the runnable the publisher belongs to
     Publisher_t(const capro::ServiceDescription& service,
-                const cxx::CString100& runnableName = cxx::CString100("")) noexcept;
+                const RunnableName_t& runnableName = RunnableName_t("")) noexcept;
```
elfenpiff, line 37: Doesn't have the `string` a default ctor which constructs an empty string? It is just minor but I would write `RunnableName_t()`.
Also below `Publisher` ctor

iceoryx_posh/include/iceoryx_posh/roudi/introspection_types.hpp
elfenpiff: @elBoberido look at the comment in line 61. Could we remove the `alignas(8)` therefore or do I miss something?!

elfenpiff (changes requested):

elfenpiff:
> Are you coming up with a follow up PR for the RMW implementation as well? I believe removing CString100 will break the ROS2 rmw compilation.

@Karsten1987 @mossmaurice @elBoberido @budrus  I think we require some kind of integrationtest which is directly in our CI pipeline which always build rmw_iceoryx with the new version.
How many committers are required to merge a pull request in rmw_iceoryx? We should have at least one (or better two) more then required otherwise this could really block us.

iceoryx_posh/include/iceoryx_posh/roudi/introspection_types.hpp
elBoberido: you are right

iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
elBoberido: yes

iceoryx_posh/source/roudi/roudi.cpp
budrus: This PR removes the old CStrin100 and not the std::string, therefore fine for now, right?

budrus (approved):

budrus:
> 
> 
> It's essentially boiling down to how much attention you want to give to the ROS2 interoperability. I am fine with putting a disclaimer on the RMW master branch, saying that the latest and greatest of iceoryx can't currently be used with ROS2. I could also imagine to simply erase the master branch (or rename the branch to `develop`), so that only tagged release branches are compatible with their respective iceoryx version.
> 
> I've raised that question in the past where we have to decide a little bit on a release strategy for iceoryx as well as the RMW layer.

With the new API we have something like a new iceoryx. The rmw_iceoryx needs to be heavily adapted. I would also assume that we switch to the C API. So as long as we do not write the next gen rmw there is no working master/master combination. Best would be to prevent people to be able to try  

Ready for a rmw hackathon in spring 2021 @jeising @Karsten1987?

iceoryx_posh/source/roudi/roudi.cpp
elBoberido: yes, replacing std::string is the next topic on my todo list

elBoberido:
@Karsten1987 on top of what the other already said, with 1.0 we commit the API stability. A breaking change would therefore considered a bug. We need to setup some tooling to notice accidental breakages, but dkroenke has already some ideas how to do it.

elfenpiff (approved):

budrus (approved):

elBoberido:
@Karsten1987 are you fine with the merge?
There is also a possibility that rmw_iceoryx master doesn't break with this PR since we only delete the alias and depending on how it's used in rmw_iceoryx it might have no impact. What definitely will have an impact is #360

MatthiasKillat (approved):

Karsten1987 (approved):

Karsten1987:
all good. Go ahead with the merge. I'll follow up with a comment on rmw_iceoryx to pay attention to the versioning.

PR #364, opened 2020-11-13
Iox#216 lockfree queue preliminary simplification refactoring

MatthiasKillat:
A preliminary simplificaton before implementing the resize/setCapacity functionality.

Essentially the UniqueIndex was removed.
Reason: It cannot really protect from misuse and will cause minor inefficiencies. Using it makes the queue code harder to understand for little/no gain. 

Since it will become a little more complicated with the setcapacity feature, I opted to simplify it before starting the real feature.
The additional pull request will make it easier to see what actually is part of the feature and what was part of the simplification.


iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/index_queue.hpp
```diff
@@ -145,38 +70,37 @@ class IndexQueue
...
-    // UniqueIndex from it.
-
     /// @brief push index into the queue in FIFO order
-    /// always succeeds if the UniqueIndex to be pushed is popped from another equallysized
-    /// IndexQueue (and should only be used this way)
-    /// threadsafe, lockfree
-    /// @param index to be pushed, any index can only be pushed once by design
-    void push(UniqueIndex& index) noexcept;
+    /// @param index to be pushed
+    /// note that do the way it is supposed to be used
```
MatthiasKillat, line 75: Comment is broken...will fix it.

elfenpiff (approved):

elBoberido (approved):

elfenpiff (approved):

Issue #327, opened 2020-10-30
Global Instantiation of Publisher/Subscriber created core dump


surendra210:
## Required information

**Operating system:**
Ubuntu 18.04

**Compiler version:**
GCC 7.4.0

**Observed result or behavior:**
Process with global Publisher/Subscriber exits with core dump.

**Expected result or behavior:**
No core dump for global Publisher/Subscriber based communication.

**Conditions where it occurred / Performed steps:**

Step1: Create a _global instance of a wrapper class_ which calls iox::runtime::PoshRuntime::getInstance("/ProcessUniqueString") in its constructor.
Step2: Create a global Publisher/Subscriber Instances, to communicate over a topic.
Step3: Compile code and run RouDi and start Publisher/Subscriber.
Step4: You will observe core dump with SIGABT.

**NOTE:** 
We were able to get rid of this core dump issue by replacing **PoshRuntime::s_runtimeFactory()** with **PoshRuntime::defaultRuntimeFactory()**  inside PoshRuntime::getInstance() in filename posh_runtime.cpp

We suspect PoshRuntime::s_runtimeFactory() has not been defined by the time PoshRuntime::getInstance() is called by _global instance of the wrapper class_

budrus:
Thanks for reporting @surendra210 

Some questions: 

- In your wrapper class, do you also pass a name to the iox::runtime::PoshRuntime::getInstance() call? With the first call a name for your process must be provided and this call must also be made before creating publishers and subscribers
- This are all global objects, or? So we have some kind of [static initilization fiasco](http://www.cs.technion.ac.il/users/yechiel/c++-faq/static-init-order.html)? 

Our current contract is that first a runtime must be created an then you are allowed to create publishers and subscribers. We do some singleton magic behind the scenes. With that approach you don't have to provide the runtime when creating publishers and subscribers but you have to fulfil this contract.

surendra210:
>>In your wrapper class, do you also pass a name to the iox::runtime::PoshRuntime::getInstance() call? With the first call a name for your process must be provided and this call must also be made before creating publishers and subscribers
<< Sorry for the confusion with my description, we do send a unique string something like  iox::runtime::PoshRuntime::getInstance("/ProcessUniqueString"), we do ensure that getInstance is called before pub/sub, by creating the Global instances in order of WrapperClass followed by pub/sub.

Note: The main essence of the wrapper class itself is to do the getInstance before pub/sub.

[...]

PR #373, opened 2020-11-17
Iox #260 remove std string from posh part 1

### LINKED ISSUE #260: Clean-up string usage

elBoberido:
This PR removes the std::string from the RouDiEnvironment and from the public API from PoshRuntime.
The C binding is adjusted and some bugs were fixed

partly closes #260

### DIFF
--- a/iceoryx_binding_c/include/iceoryx_binding_c/runnable.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/runnable.h
@@ -33,7 +33,8 @@ void iox_runnable_destroy(iox_runnable_t const self);
 /// @param[in] name pointer to a memory location where the name can be written to
 /// @param[in] nameCapacity size of the memory location where the name is written to
 /// @return the actual length of the runnable name, if the return value is greater
-///         then nameCapacity the name is truncated
+///         then nameCapacity the name is truncated.
+///         If name is a nullptr, 0 will be returned.
 uint64_t iox_runnable_get_name(iox_runnable_t const self, char* const name, const uint64_t nameCapacity);
 
 /// @brief acquires the name of the process in which the runnable is stored
@@ -41,7 +42,8 @@ uint64_t iox_runnable_get_name(iox_runnable_t const self, char* const name, cons
 /// @param[in] name pointer to a memory location where the name can be written to
 /// @param[in] nameCapacity size of the memory location where the name is written to
 /// @return the actual length of the process name, if the return value is greater
-///         then nameCapacity the name is truncated
+///         then nameCapacity the name is truncated.
+///         If name is a nullptr, 0 will be returned.
 uint64_t iox_runnable_get_process_name(iox_runnable_t const self, char* const name, const uint64_t nameCapacity);
 
 #endif
--- a/iceoryx_binding_c/source/c_runtime.cpp
+++ b/iceoryx_binding_c/source/c_runtime.cpp
@@ -45,9 +45,9 @@ uint64_t iox_runtime_get_instance_name(char* const name, const uint64_t nameLeng
     }
 
     auto instanceName = PoshRuntime::getInstance().getInstanceName();
-    uint64_t instanceNameSize = instanceName.size();
-    std::strncpy(name, instanceName.c_str(), std::min(nameLength, instanceNameSize));
+    std::strncpy(name, instanceName.c_str(), nameLength);
     name[nameLength - 1] = '\0'; // strncpy doesn't add a null-termination if destination is smaller than source
 
-    return instanceNameSize;
+    return instanceName.size();
+    ;
 }
--- a/iceoryx_binding_c/source/c_runtime.cpp
+++ b/iceoryx_binding_c/source/c_runtime.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2020 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,13 +23,31 @@ extern "C" {
 
 void iox_runtime_register(const char* const name)
 {
-    PoshRuntime::getInstance(name);
+    if (name == nullptr)
+    {
+        LogError() << "Application name is a nullptr!";
+        std::terminate();
+    }
+    else if (strnlen(name, iox::MAX_PROCESS_NAME_LENGTH + 1) > MAX_PROCESS_NAME_LENGTH)
+    {
+        LogError() << "Application name has more than 100 characters!";
+        std::terminate();
+    }
+
+    PoshRuntime::getInstance(ProcessName_t(iox::cxx::TruncateToCapacity, name));
 }
 
 uint64_t iox_runtime_get_instance_name(char* const name, const uint64_t nameLength)
 {
+    if (name == nullptr)
+    {
+        return 0;
+    }
+
     auto instanceName = PoshRuntime::getInstance().getInstanceName();
     uint64_t instanceNameSize = instanceName.size();
-    strncpy(name, instanceName.c_str(), std::min(nameLength, instanceNameSize));
+    std::strncpy(name, instanceName.c_str(), std::min(nameLength, instanceNameSize));
+    name[nameLength - 1] = '\0'; // strncpy doesn't add a null-termination if destination is smaller than source
+
     return instanceNameSize;
 }
--- a/iceoryx_binding_c/source/c_runtime.cpp
+++ b/iceoryx_binding_c/source/c_runtime.cpp
@@ -41,13 +41,12 @@ uint64_t iox_runtime_get_instance_name(char* const name, const uint64_t nameLeng
 {
     if (name == nullptr)
     {
-        return 0;
+        return 0U;
     }
 
     auto instanceName = PoshRuntime::getInstance().getInstanceName();
     std::strncpy(name, instanceName.c_str(), nameLength);
-    name[nameLength - 1] = '\0'; // strncpy doesn't add a null-termination if destination is smaller than source
+    name[nameLength - 1U] = '\0'; // strncpy doesn't add a null-termination if destination is smaller than source
 
     return instanceName.size();
-    ;
 }
--- a/iceoryx_binding_c/test/integrationtests/test_runtime.cpp
+++ b/iceoryx_binding_c/test/integrationtests/test_runtime.cpp
@@ -1,0 +1,101 @@
+// Copyright (c) 2020 by Apex.AI Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extern "C" {
+#include "iceoryx_binding_c/runtime.h"
+}
+
+#include "iceoryx_posh/iceoryx_posh_types.hpp"
+#include "testutils/roudi_gtest.hpp"
+
+using namespace iox;
+using namespace iox::runtime;
+
+class BindingC_Runtime_test : public RouDi_GTest
+{
+  public:
+    void SetUp()
+    {
+    }
+
+    void TearDown()
+    {
+    }
+};
+
+TEST_F(BindingC_Runtime_test, SuccessfulRegistration)
+{
+    constexpr char EXPECTED_APP_NAME[iox::MAX_PROCESS_NAME_LENGTH + 1] = "/chucky";
+    iox_runtime_register(EXPECTED_APP_NAME);
+
+    char actualAppName[iox::MAX_PROCESS_NAME_LENGTH + 1];
+    auto nameLength = iox_runtime_get_instance_name(actualAppName, iox::MAX_PROCESS_NAME_LENGTH + 1);
+
+    ASSERT_THAT(nameLength, Eq(strnlen(EXPECTED_APP_NAME, iox::MAX_PROCESS_NAME_LENGTH + 1)));
+    EXPECT_THAT(actualAppName, StrEq(EXPECTED_APP_NAME));
+}
+
+TEST_F(BindingC_Runtime_test, AppNameLengthIsMax)
+{
+    std::string maxName(iox::MAX_PROCESS_NAME_LENGTH, 's');
+    maxName.front() = '/';
+
+    iox_runtime_register(maxName.c_str());
+
+    char actualAppName[iox::MAX_PROCESS_NAME_LENGTH + 1];
+    auto nameLength = iox_runtime_get_instance_name(actualAppName, iox::MAX_PROCESS_NAME_LENGTH + 1);
+
+    ASSERT_THAT(nameLength, Eq(iox::MAX_PROCESS_NAME_LENGTH));
+}
+
+TEST_F(BindingC_Runtime_test, AppNameLengthIsOutOfLimit)
+{
+    std::string tooLongName(iox::MAX_PROCESS_NAME_LENGTH + 1, 's');
+    tooLongName.insert(0, 1, '/');
+
+    EXPECT_DEATH({ iox_runtime_register(tooLongName.c_str()); }, "Application name has more than 100 characters!");
+}
+
+TEST_F(BindingC_Runtime_test, AppNameIsNullptr)
+{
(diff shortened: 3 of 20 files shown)


FerdinandSpitzschnueffler (dismissed):

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -185,7 +185,7 @@ struct DefaultChunkQueueConfig
 
 // alias for cxx::string
 using ConfigFilePathString_t = cxx::string<1024>;
-using ProcessName_t = cxx::string<100>;
+using ProcessName_t = cxx::string<MAX_PROCESS_NAME_LENGTH>;
 using RunnableName_t = cxx::string<100>;
```
marthtz, line 189: Why not MAX_PROCESS_NAME_LENGTH as well?

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
```diff
@@ -49,7 +50,9 @@ class SegmentManager
     struct SegmentMapping
     {
       public:
-        SegmentMapping(const std::string& sharedMemoryName,
+        using string_t = cxx::string<128>;
```
marthtz, line 53: No constexpr for `128`?

iceoryx_binding_c/source/c_runtime.cpp
```diff
@@ -23,13 +23,31 @@ extern "C" {
 
 void iox_runtime_register(const char* const name)
 {
-    PoshRuntime::getInstance(name);
+    if (name == nullptr)
+    {
+        LogError() << "Application name is a nullptr!";
+        std::terminate();
```
dkroenke, line 29: Is it not a little bit to hard to terminate the complete process when the given name does not match the requirements? Maybe it should just print a message like "application name not set, runtime not created" and exit the function without creating a runtime.
Same for the process name length.

iceoryx_binding_c/source/c_runtime.cpp
```diff
@@ -23,13 +23,31 @@ extern "C" {
...
+    }
+    else if (strnlen(name, iox::MAX_PROCESS_NAME_LENGTH + 1) > MAX_PROCESS_NAME_LENGTH)
```
dkroenke, line 31: Is the check for length not already done in verifyInstanceName() when calling getInstance() in posh_runtime.cpp?

iceoryx_binding_c/source/c_runtime.cpp
elfenpiff: Maybe we return a `bool` if it was successful. Or an enum which states the problem like:
- `APPLICATION_NAME_IS_NULLPTR`
- `EXCEEDING_ALLOWED_NAME_LENGTH`

iceoryx_binding_c/test/integrationtests/test_runtime.cpp
```diff
@@ -0,0 +1,102 @@
...
+    constexpr char EXPECTED_APP_NAME[iox::MAX_PROCESS_NAME_LENGTH + 1] = "/chucky";
+    iox_runtime_register(EXPECTED_APP_NAME);
+
+    char actualAppName[iox::MAX_PROCESS_NAME_LENGTH + 1];
+    auto nameLength = iox_runtime_get_instance_name(nullptr, iox::MAX_PROCESS_NAME_LENGTH + 1);
+
+    ASSERT_THAT(nameLength, Eq(0U));
+}
+
+TEST_F(BindingC_Runtime_test, GetInstanceNameLengthIsLessThanAppNameLength)
```
elfenpiff, line 86: What about tests when the app name length is max or is exceeding the allowed name length?

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
elfenpiff: Magic Number? I think the shared memory needs also a `MAX_SHARED_MEMORY_NAME_LENGTH` variable in `posh_types`

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -29,22 +29,29 @@ namespace iox
...
-PoshRuntime& PoshRuntime::getInstance(const std::string& name) noexcept
+PoshRuntime& PoshRuntime::getInstance(const ProcessName_t& name) noexcept
 {
     return PoshRuntime::s_runtimeFactory(name);
 }
 
-PoshRuntime::PoshRuntime(const std::string& name, const bool doMapSharedMemoryIntoThread) noexcept
+ProcessName_t& PoshRuntime::defaultRuntimeInstanceName() noexcept
+{
+    static ProcessName_t defaultInstanceName = "dummy";
```
elfenpiff, line 50: Again magic name? I think we should move the default name into `posh_types` like
`static constexpr char DEFAULT_INSTANCE_NAME[MAX_PROCESS_NAME_LENGTH] = "dummy";` and use it here.

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -66,34 +73,29 @@ PoshRuntime::~PoshRuntime() noexcept
...
         std::terminate();
     }
-    else if (name.compare(DEFAULT_RUNTIME_INSTANCE_NAME) == 0)
+    else if (name.compare(defaultRuntimeInstanceName()) == 0)
     {
         LogError() << "Cannot initialize runtime. Application name has not been specified!";
         std::terminate();
     }
-    else if (name.front() != '/')
+    else if (name.c_str()[0] != '/')
```
elfenpiff, line 88: Magic slash? To understand the context better you could maybe create a variable
```cpp
constexpr char PATH_SEPARATOR = '/';
if ( name[0] != PATH_SEPARATOR ) ...
```
Then a reviewer maybe understands better why we require this.

elfenpiff (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
elBoberido: I don't consider this as magic number. It's obvious what the number means in the context of the string and we also don't have it for most of the string aliases. The constexpr was only used when the value of the capacity was needed. I'd rather create a static `capacity()` function in the `cxx::string` than to have everywhere a constexpr for the length. Just in case, you also don't want a constexpr for the length of every vector, do you?

iceoryx_binding_c/source/c_runtime.cpp
elBoberido: it's the same behavior like in the PoshRuntime. I also don't consider this a place for a graceful degradation. We need a runtime. Iceoryx doesn't work without a runtime. It's mandatory and nothing which only might be called under some conditions. It's essentially the same as when you don't get memory. Your application basically doesn't work when this call fails, so immediate termination makes sense here.

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: I deliberately changed the constexpr to this, because it's used as default value for the instance name in `getInstance(..)`. If we keep the constecpr, every call to `getInstance()` will construct a cxx::string

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elBoberido: I'd rather create a static `capacity` function in `cxx::string` than doing this. Is this okay for you?

iceoryx_binding_c/test/integrationtests/test_runtime.cpp
elBoberido: like the test in line 49 and 62 ;)

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: Maybe we should get rid of the slash. As far as I know it's removed when the unix domain socket is used. I also wouldn't call it `PATH_SEPARATOR` since it's the same on Windows, even with `\` as path separator. How about `APP_PREFIX`

iceoryx_posh/source/runtime/posh_runtime.cpp
MatthiasKillat: In support of @elBoberido : we also do not want or need this name to be publicly accessible, which it would be in posh_types.I would keep it hidden, it is not really intended to construct a dummy runtime anyway.

iceoryx_posh/source/runtime/posh_runtime.cpp
MatthiasKillat: Were should this live? If locally it does not add much, just makes the code artificially longer. If globally is not really useful...unless we want to make it configurable somehow. This is not needed here I would say.

MatthiasKillat (approved):

iceoryx_posh/source/runtime/posh_runtime.cpp
budrus: I guess we have still the requirement that the name provided when creating the runtime has to start with a slash. And I guess the reason was that this string is used to name a message queue. As we heavily break this API now, we could clean this up, or?

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: btw. I never liked the `getInstance()` call with a default parameter which is just a random string to indicate that the function was called without a parameter. I would suggest to have two functions `initRuntim(ProcessName_t& name)` and `getInstance()` with
```
private:
    PoshRuntime& getInstance(cxx::optional<ProcessName_t*> name);
public:
    PoshRuntime& initRuntime(ProcessName_t& name) {
        return getInstance(cxx::make_optional<ProcessName_t*>(&name));
    }
    PoshRuntime& getInstance() {
        return getInstance(cxx::nullopt);
    }
```
With this approach we get rid of a magic string and make the code more robust.
 Maybe @FerdinandSpitzschnueffler can implement this with #327 
@budrus, @elfenpiff @MatthiasKillat what do you think?

iceoryx_binding_c/include/iceoryx_binding_c/runnable.h
```diff
@@ -33,15 +33,17 @@ void iox_runnable_destroy(iox_runnable_t const self);
 /// @param[in] name pointer to a memory location where the name can be written to
 /// @param[in] nameCapacity size of the memory location where the name is written to
 /// @return the actual length of the runnable name, if the return value is greater
-///         then nameCapacity the name is truncated
+///         then nameCapacity the name is truncated.
```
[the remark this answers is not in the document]
elBoberido: will fix the typo with one of my next commits

iceoryx_posh/source/runtime/posh_runtime.cpp
budrus: Sounds good 👍

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
elfenpiff: Actually, I would disagree. And here the argument is not that the user does not know what the `128` means it is more like we would like to change sometime the support name length and then we find every property in the posh_types file except this one which  is stored in some header file.
I think now that we have started the thing with the posh_types we should be consistent.

iceoryx_posh/source/runtime/posh_runtime.cpp
elfenpiff: Great idea! I think @FerdinandSpitzschnueffler had the same idea for the runtime factory.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elfenpiff: No please don't. Again here the argument is that all settings are stored in the posh_types. Maybe we support at some point in the future a different process name length. Every setting is then stored in posh_types except this one which lives in some header file.

But a static capacity function is a good idea anyway!

iceoryx_posh/source/runtime/posh_runtime.cpp
elfenpiff: Maybe we should hide this from the user. If the name like for message queues has to start with a slash then the message queue adds one if no slash is present otherwise it leaves it as it is. 
It would also have no effect for the user since he would still find the corresponding file handle under the specified name.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
marthtz: I agree with @elfenpiff . Maybe we don't need a `constexpr` for every single one and use a generic/default one like `MAX_DEFAULT_NAME_LENGTH` for all `128` for example?

iceoryx_posh/source/runtime/posh_runtime.cpp
FerdinandSpitzschnueffler: Good idea and quite similar to the runtime factory changes I implemented for [#327](https://github.com/eclipse/iceoryx/issues/327). I will add @elBoberido's approach.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elBoberido: @elfenpiff the alias for `ProcessName_t` also lives in the posh_types. It comes down to go to posh_types and either change the constexpr or the template parameter of the alias.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
elBoberido: @elfenpiff `ProcessName_t` **is** stored in the posh_types ;)

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: @FerdinandSpitzschnueffler since this will need changes is several places and make the PR of the bugfix quite large, maybe it would be better to have a separate PR for this

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
elBoberido: @elfenpiff in this case, the alias should probably live there. There is an item in #260 which takes care of the convention on how to name string aliases and I wanted to do it when we decide on that one.

elBoberido:
@elfenpiff @marthtz would you agree to continue the discussion on string length as constexpr in the #260 issue. We can also discuss which convention we want for our string aliases and I would create a final PR with the outcome. I could continue working on the `std::string` removal with several smaller PRs until we have a decision on this

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: I've created #381 for this

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: will be done by #382

elfenpiff:
> @elfenpiff @marthtz would you agree to continue the discussion on string length as constexpr in the #260 issue. We can also discuss which convention we want for our string aliases and I would create a final PR with the outcome. I could continue working on the `std::string` removal with several smaller PRs until we have a decision on this

Agreed!

iceoryx_binding_c/source/c_runtime.cpp
elBoberido: @dkroenke since this is not a behavioral change (see PoshRuntime) are you okay with keeping it as is?

dkroenke: yes it's fine

dkroenke (approved):

elBoberido:
@elfenpiff can we merge this?

elfenpiff (approved):

PR #379, opened 2020-11-18
Iox #374 fix strncpy usage

### LINKED ISSUE #374: Incorrect usage of strncpy
## Required information

strncpy is incorrectly used in some places. If count is reached before the null-termination of the source, no null-termination is added. This leads to a non null-terminated string if the null-termination is not added manually.

Furthermore, instead of setting the destination buffer size as count, often the source buffer size is set, leading to a potential buffer overflow.

Occurences:
- unix_domain_socket.cpp
- semaphore.cpp
- uds.cpp from examples

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
- fixed ~~possible~~ (we triggered it in a test) buffer overflow due to incorrect strncpy usage in `posix::UnixDomainSocket` and `UDS` from the examples
- adjusted `MAX_PROCESS_NAME_LENGTH` for maxOS since the current one exceeds the buffer size for the unix domain socket path
- removed strncpy from `posixs::Semaphore`; in hindsight this was not necessary since there was a check earlier in the code, but I already replaced the `char array` with a `cxx::string` and removed some code

1. [ ] Merge

## References

- Closes #374


iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -326,6 +327,11 @@ cxx::expected<int32_t, IpcChannelError> UnixDomainSocket::createSocket(const Ipc
     // initialize the sockAddr data structure with the provided name
     memset(&m_sockAddr, 0, sizeof(m_sockAddr));
     m_sockAddr.sun_family = AF_LOCAL;
+    const uint64_t maxDestinationLength = cxx::strlen2(m_sockAddr.sun_path);
+    if (m_name.length() > maxDestinationLength)
+    {
+        return cxx::error<IpcChannelError>(IpcChannelError::INVALID_CHANNEL_NAME);
```
elBoberido, line 333: How do we want to test this? Currently the UnixDomainSocket uses the same tests as the MessageQueue and is only tested on maxOS.
Shall I make the MessageQueue test a typed test and add a test for UnixDomainSocket specific behavior?

MatthiasKillat: I would say this requires a set of separate tests specific to the UnixDomainSocket. To make a typed test reasonable the types would always need to have the same interface and semantics (the message queue will not have a createSocket method). E.g. are the maximum allowed name length the same? Otherwise both would behave differently, unless some additional measures are implemented (minimum of both allowed length etc.).

So just one UnixDomainSocket test were the name is (just) short enough and another one were it is too long and construction fails would be my proposal.

Those tests could (and should, if we provide the UnixDomainSocket class in utils) also run on Linux I guess, they are in the POSIX standard if I am not mistaken.

MatthiasKillat: Creating additional tests unrelated to string is not part of this pull request in my opinion.

iceoryx_examples/iceperf/uds.cpp
```diff
@@ -46,6 +47,8 @@ void UDS::init() noexcept
     // initialize the sockAddr data structure with the provided name
     memset(&m_sockAddrPublisher, 0, sizeof(m_sockAddrPublisher));
     m_sockAddrPublisher.sun_family = AF_LOCAL;
+    const uint64_t maxDestinationLength = iox::cxx::strlen2(m_sockAddrPublisher.sun_path);
+    assert(maxDestinationLength >= m_publisherName.size() && "Socketname too large!");
```
MatthiasKillat, line 51: Is an assert here really sufficient? I think this check needs to be there in release mode as well or do we have full control over what is passed via m_publisherName? 

If yes, an assert suffices.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: Well, currently the MessageQueue tests are used for the UnixDomainSocket on macOS, so there is a common interface and semantics. There are just fully equal, like with the name length.

iceoryx_examples/iceperf/uds.cpp
elBoberido: It's just a tests, therefore I took the easiest route, but I can also add a "std::terminate there".

@budrus why wasn't the `UnixDomainSocket` class used for the example? Because it doesn't support zero copy at the moment? I'd like to add a zero copy interface to if when I port it away from std::string

budrus: @elBoberido I wanted to use the UDS as raw as possible. Otherwise people could argue that we loose some performance because of our abstraction layers

iceoryx_utils/include/iceoryx_utils/posix_wrapper/semaphore.hpp
```diff
@@ -185,8 +186,7 @@ class Semaphore : public DesignPattern::Creation<Semaphore, SemaphoreError>
     iox_sem_t* getHandle() noexcept;
 
   private:
-    static constexpr int m_nameSize = 128;
-    char m_name[m_nameSize] = {'\0'};
+    cxx::string<128> m_name;
```
marthtz, line 189: I guess that'll be cleaned-up when a common convention for string aliases is found in #260?!

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
marthtz: >
> Shall I make the MessageQueue test a typed test 

Would that take lots of effort?
In the end we want to have universal tests that run on every system and not just on a specific one, don't we?

elBoberido: No, I guess that wouldn't be much afford

iceoryx_utils/include/iceoryx_utils/posix_wrapper/semaphore.hpp
elBoberido: yes, the idea is to have cxx::strings in the classes itself and for the last PR of #260 do a grep for all string aliases or direct usages and consolidate it. We could also decide to refactor the Semaphore and use cxx::string in the API instead of a `const char*`, but that is a bigger refactoring topic and maybe something for v1.1

iceoryx_examples/iceperf/uds.cpp
elBoberido: @MatthiasKillat it's the example with the benchmark, not a test like I mentioned before, and we have full control, so this is just to appease my inner chicken

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -223,6 +223,11 @@ MqRuntimeInterface::MqRuntimeInterface(const std::string& roudiName,
     , m_AppMqInterface(appName)
     , m_RoudiMqInterface(roudiName)
 {
+    if (!m_AppMqInterface.isInitialized()) {
+        errorHandler(Error::kMQ_INTERFACE__UNABLE_TO_CREATE_APPLICATION_MQ);
```
elBoberido, line 227: I have no idea how to test this. Especially with the unix domain socket on mac, which is the reason for this. This can only happen if the unix domain socket cannot be created, e.g. when the path name is too long. But that is already taken care of with `MAX_PROCESS_NAME` and there is already a test in PoshRuntime for MAX_PROCESS_NAME

ithier (approved):
Looks good to me.

budrus:
@elBoberido What is the consequence of this review check "Copyright owner are updated in the changed files"? If a guy from one company touches a file that was originally written by a guy from another company, the header must include both copyrights?

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -148,7 +148,15 @@ constexpr uint32_t MAX_NUMBER_OF_INSTANCES = 50U;
 constexpr uint32_t MAX_RUNNABLE_NUMBER = 1000U;
 constexpr uint32_t MAX_RUNNABLE_PER_PROCESS = 50U;
 
+#if defined(__APPLE__)
+/// @note on macOS the process name length nedds to be decreased since the process name is used for the unix domain
+/// socket path which has a capacity for only 103 characters. The full path consists of UnixDomainSocket::PATH_PREFIX,
+/// which is currently 5 characters and the specified process name.
+constexpr uint32_t MAX_PROCESS_NAME_LENGTH = 98U;
```
budrus, line 155: So the length of PATH_PREFIX is implicitly hidden in the 98. Can you calculate this with the given 103 and PATH_PREFIX. Would be more robust to PATH_PREFIX changes

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
budrus: We already differentiate in the message queue test https://github.com/eclipse/iceoryx/blob/master/iceoryx_utils/test/moduletests/test_message_queue.cpp#L163

So you can add a test that is specific for UDS?

elBoberido:
> @elBoberido What is the consequence of this review check "Copyright owner are updated in the changed files"? If a guy from one company touches a file that was originally written by a guy from another company, the header must include both copyrights?

@budrus exactly. But common sense should be applied. So fixing a typo or changing a variable name may not imply a copyright change

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elBoberido: @budrus elfenpiff has plans to make PATH_PREFIX customizable, then this calculation might be obsolete. But could right now be done easily. There are runtime checks for the issue, though.

budrus: So we leave it for now?

elBoberido: I also noticed that I would need to include the unix_domain_socket.hpp in iceoryx_posh_types.hpp so I would rather leave it for now.

elBoberido:
Okay, it's now done. Btw, I should have been waiting with adding reviewers until the PR was ready. Sorry for that.

budrus (dismissed):

elBoberido:
Oh nuts! Since I rebased on master which has already replaced std::string in some places, we get an compile error on mac since we do not use the correct alias for ProcessName_t everywhere. I've update #383 and checked that it compile with a MAX_PROCESS_NAME_LENGTH of 98 on linux. So once that is merged, this one should also build on mac.

marthtz (approved):

iceoryx_examples/iceperf/uds.cpp
elfenpiff: @elBoberido I would prefer to use `cxx::Expects` here since `assert` is the old C way and `cxx::Expects` provides you with a more detailed output like 
  - file 
  - line
  - function
  - condition 

when something goes wrong.

Additionally, it is more expressive..

iceoryx_utils/source/posix_wrapper/message_queue.cpp
```diff
@@ -400,6 +400,11 @@ cxx::error<IpcChannelError> MessageQueue::createErrorFromErrnum(const std::strin
         std::cerr << "message queue \"" << name << "\" does not exist" << std::endl;
         return cxx::error<IpcChannelError>(IpcChannelError::NO_SUCH_CHANNEL);
     }
+    case ENAMETOOLONG:
+    {
+        std::cerr << "message queue \"" << name << "\" has too long name" << std::endl;
```
elfenpiff, line 405: It sound so german ... maybe 
```cpp
std::cerr << "the message queue name \"" << name << "\" is too long" << std::endl;
```
is better. 

@ithier do you have an even better lyrical euphonious recommendation?

elfenpiff (changes requested):
- please substitute `assert` with `cxx::Expected` if you agree with my assessment

iceoryx_utils/source/posix_wrapper/message_queue.cpp
dkroenke: My proposal: "The amount of characters in the designation of the message queue entity exceeds the restrictions"

elfenpiff: @dkroenke my counter proposal (please read it in your head with the voice of Sean Connery):
"Dear Sir or Madam, would you please be so kind to reduce the amount of characters in your splendid message queue name to help me, your humble program, to run your magnificent algorithm. Yours sincerely, RouDi"

budrus (dismissed):

iceoryx_examples/iceperf/uds.cpp
elBoberido: done

elfenpiff (approved):

iceoryx_utils/source/posix_wrapper/message_queue.cpp
elBoberido: how about
```
O MessageQueue, MessageQueue, wherefore art thou MessageQueue?
Deny thy request and refuse thy name.
Or if thou wilt not, be but sworn my queue
And I'll no longer be as IPC the same.
```

elfenpiff: Uuh this is great. I thought also maybe our error messages should sound more sophisticated, articulated, with a glimmer of passion and of course very polite.
`Name too long` does not fulfill then our criteria. We should create an issue for that called lyrical evaluation off end user communication.
@elBoberido @budrus I think this should be the highest prio.

marthtz (approved):

PR #400, opened 2020-11-30
iox-#141 enable sanizier(-fsanitize=address)

### LINKED ISSUE #141: Enable Sanitizer in Debug Build and Unit Tests
## Brief feature description
Add the following compiler flags for the Linux (and MacOS build, if possible)
```-fsanitize=address -fsanitze=undefined```. All unit tests have to run without undefined behavior, memory leaks or other funky memory effects.

## Detailed information

We can a avoid a wide variety of bugs when this is activated in the github CI.

### PR DESCRIPTION
**Pull request by prasannabhat**

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer

**What is done ?**

- Support AddressSanitizer (all the flags described [here](https://clang.llvm.org/docs/AddressSanitizer.html), including builtin leak detection)
- Suppress known errors (blacklist files are stored under `iceoryx_meta/sanitizer_blacklist`)
- Support Sanitizer github workflow (Linux & Mac)
    - Known errors are suppressed
    - workflow fails (exits upon first error) upon detection any new errors, if its pattern doesnt match blacklist

**Which components are sanitized?**

- Sanitization is enabled for core components (`iceoryx_posh` & `iceoryx_utils`)
However 

> Runtime interposition allows AddressSanitizer to find bugs in code that is not being recompiled

**Next steps**

- Sanization be enabled for components other than `iceoryx_posh` & `iceoryx_utils` ?
- Enable other [sanitizers](https://github.com/google/sanitizers) (e.g. ThreadSanitizer , MemorySanitizer)
- Analyse and fix (if needed) the known issues
 
More details on the sanitizer can be found here ( [Clang](https://clang.llvm.org/docs/AddressSanitizer.html) , [Google](https://github.com/google/sanitizers/wiki/AddressSanitizer))

1. [x] Merge

## References

- Closes #141

### CODE REVIEW

File: .github/workflows/sanitize.yml [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+      - name : Checkout
+        uses: actions/checkout@v2
+
+      - name : Clang Sanitize
+        run: |
+          export CC=/usr/bin/clang
+          export CXX=/usr/bin/clang++
+          export NUM_CORES=`nproc`
+          # build tests , dont run them
+          $GITHUB_WORKSPACE/tools/iceoryx_build_test.sh build-test sanitize clean -j $NUM_CORES
```
elBoberido, line 33: Any reasons the dds gateway is not included? With the removal of the clang build, the dds gateway wouldn't be build with clang anymore

iceoryx_meta/CMakeLists.txt
```diff
@@ -127,5 +140,6 @@ message("       Build Properties")
 message("          project name..............: " ${CMAKE_PROJECT_NAME})
 message("          c compiler................: " ${CMAKE_C_COMPILER})
 message("          c++ compiler..............: " ${CMAKE_CXX_COMPILER})
+message("          c++ flags.................: " ${CMAKE_CXX_FLAGS})
```
elBoberido, line 143: this might mislead since we don't set the flags globally anymore. We set it on the target, e.g. with `${ICEORYX_CXX_STANDARD}` or `${ICEORYX_WARNINGS}`

iceoryx_meta/sanitizer_blacklist.txt
```diff
@@ -0,0 +1,9 @@
+# Suppress specific errors
```
elBoberido, line 1: maybe we need an issue to fix all the errors from this file

iceoryx_meta/CMakeLists.txt
```diff
@@ -43,6 +44,18 @@ if(coverage)
     endif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 endif(coverage)
 
+if(sanitize)
```
elBoberido, line 47: this won't work if iceoryx is not build with `iceoryx_meta`. Maybe it should be moved to `iceoryx_utils/cmake/IceoryxPlatformDetection.cmake` ... and maybe we need to rename that file to make it clear that there is more than just platform detection in in

sanitize_test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
elBoberido, line 1: was this file intentionally be added?

elBoberido (commented):
nice, you will make @elfenpiff quite happy with this

dkroenke (changes requested):
For resolving the memleaks in test_waitset.cpp do the following:
In Teardown extend the delete loop: 
```
      for (auto s : m_subscriber)
        {
            delete s->m_portData;
            delete s;
        }
```


Add 'delete sut2;' to line 115

.github/workflows/sanitize.yml
prasannabhat: > nice, you will make @elfenpiff quite happy with this

I hope I wont make anybody unhappy with this ;-)

prasannabhat: I skipped it because there were no tests. You are right , it would be miss clang build.
Its added

iceoryx_meta/sanitizer_blacklist.txt
prasannabhat: Suppressing errors using `-fsanitize-blacklist` seems not be working. Not sure whats going wrong there !

iceoryx_meta/CMakeLists.txt
prasannabhat: `sanitize` is a special build (invoked in github action of its own). What would be the need to support it for other builds ?

Btw what are build you are referring to @elBoberido (colcon build?)

sanitize_test.sh
prasannabhat: This was private file, somehow sneaked into public !
Removed it now

prasannabhat:
> For resolving the memleaks in test_waitset.cpp do the following:
> In Teardown extend the delete loop:
> 
> ```
>       for (auto s : m_subscriber)
>         {
>             delete s->m_portData;
>             delete s;
>         }
> ```
> 
> Add `delete sut2;` to line 115

Thanks for the hint @dkroenke.
Fixed this "low hanging fruit" :-)

iceoryx_meta/CMakeLists.txt
elBoberido: if you build just iceoryx_utils for example, then the `${SANITIZE_COMPILE_FLAGS}` variable would be empty. If it's put into `IceoryxPlatformDetection.cmake` it would still work

.github/workflows/sanitize.yml
elBoberido: yes, it's more like ensuring it's building with clang and there are also no warnings. But you are right, we should probably also execute the tests with the clang build

iceoryx_meta/sanitizer_blacklist.txt
elBoberido: I guess then we need to fix them now :/

iceoryx_meta/CMakeLists.txt
prasannabhat: Makes sense 
In that case other compiler settings  (for `coverage` ) can also be moved to `IceoryxPlatformDetection.cmake`.

**Whats a good name for this file ?**
`IceoryxCentral.cmake`
Any other suggestions ?

dkroenke: @elBoberido i think it make more sense to leave it here, because iceoryx_meta is some kind of "main" cmake file to build iceoryx with. Placing that into iceoryx_utils/cmake/IceoryxPlatformDetection.cmake will "hide" this option. 
For me it is ok to have it here, because most of the users currently use iceoryx_meta for building. 
Discussion is related to https://github.com/eclipse/iceoryx/issues/230.

elBoberido: The `coverage` is just used in the tests, right? As far as I know, this also needs some configuration passed to cmake.

To be able to use the sanitizer might be useful outside of the tests and iceoryx_meta, e.g. when someone integrates iceoryx into a custom framework and wants to switch the sanitize on. Therefore I would move it some layers down. We can still have the option in iceoryx_meta.

How about calling the file just `IceoryxPlatform.cmake`?

tools/run_all_tests.sh
```diff
@@ -20,6 +20,41 @@ COMPONENTS="utils posh binding_c"
...
+#     #2 0x55c8284c15d1 in ReceiverPort_test_newdata_Test::~ReceiverPort_test_newdata_Test() /home/pbt2kor/data/aos/repos/iceoryx_oss/iceoryx/iceoryx_posh/test/moduletests/test_posh_receiverport.cpp:137
+#     #3 0x55c8284c15ed in ReceiverPort_test_newdata_Test::~ReceiverPort_test_newdata_Test() /home/pbt2kor/data/aos/repos/iceoryx_oss/iceoryx/iceoryx_posh/test/moduletests/test_posh_receiverport.cpp:137
+#     #4 0x55c82857b2fb in testing::Test::DeleteSelf_() (/home/pbt2kor/data/aos/repos/iceoryx_oss/iceoryx/build/posh/test/posh_moduletests+0x3432fb)
+    echo "OSTYPE is $OSTYPE"
+    if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        ASAN_OPTIONS=detect_leaks=1
+    else
+    # other OS (Mac here)
+    # ==23449==AddressSanitizer: detect_leaks is not supported on this platform.
+        ASAN_OPTIONS=detect_leaks=0
```
prasannabhat, line 49: detect_leaks option wont work in github mac instance ([Mac OS X 10.15.7](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md))
However the [document](https://clang.llvm.org/docs/AddressSanitizer.html#memory-leak-detection) reads as below

> The leak detection is turned on by default on Linux, and can be enabled using ASAN_OPTIONS=detect_leaks=1 on macOS; however, it is not yet supported on other platforms.

Not sure whats the problem here!

Alternate could be to try [standalone](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#stand-alone-mode) mode

iceoryx_posh/CMakeLists.txt
```diff
@@ -136,14 +136,16 @@ target_include_directories(iceoryx_posh
     source/version
     source/runtime
 )
-target_link_libraries(iceoryx_posh
+target_link_libraries(iceoryx_posh 
     PUBLIC
     iceoryx_utils::iceoryx_utils
+    #@todo - Is it required to link sanitizer libs here (Its already linked against iceoryx_utils)
+    ${SANITIZER_LINKER_FLAGS}
```
prasannabhat, line 143: Is it required to link sanitizer libs  with iceoryx_posh (& other components as well) ?
(sanitizer libs are already linked against iceoryx_utils)
@elBoberido

iceoryx_meta/sanitizer_blacklist.txt
prasannabhat: I could suppress all the known errors (So far found no hard errors that cant be suppressed :wink: )

iceoryx_meta/CMakeLists.txt
prasannabhat: @elBoberido `coverage` flags are used for compiling the sources (may be dependent libs are linked against tests only)
For now `IceoryxPlatformDetection.cmake` contains platform specific compiler flags. `sanitize` flags are compiler dependent not platform.

@dkroenke @elBoberido do I move `sanitize` flags ?

prasannabhat:
> nice, you will make @elfenpiff quite happy with this

happiness doubled @elfenpiff  :wink: (supported sanitization on macos as well)

iceoryx_posh/CMakeLists.txt
dkroenke: I would do so to be sure that the stacktrace from the sanitizer is complete. My guess is that otherwise the sanitizer cannot show the full path from posh to utils in the stacktrace.

You can add `-fsanitize=address` to the general compiler flags (`CFLAGS` and `CXXFLAGS`). See in CMakelists.txt in iceoryx_meta how it is done with the coverage flags. Then you can remove the `${SANITIZER_LINKER_FLAGS}` from the libraries.

dkroenke (changes requested):

iceoryx_meta/CMakeLists.txt
elBoberido: I think we have a tie here. Where is the tiebreaker 😄

iceoryx_posh/CMakeLists.txt
elBoberido: > You can add `-fsanitize=address` to the general compiler flags (`CFLAGS` and `CXXFLAGS`). See in CMakelists.txt in iceoryx_meta how it is done with the coverage flags. Then you can remove the `${SANITIZER_LINKER_FLAGS}` from the libraries.

I'm not sure about this. We deliberately moved away from using `CFLAGS` and `CXXFLAGS` and used `TARGET_COMPILE_OPTIONS` instead. Now we would revert those changes.

iceoryx_meta/CMakeLists.txt
elBoberido: > For now `IceoryxPlatformDetection.cmake` contains platform specific compiler flags. `sanitize` flags are compiler dependent not platform.

Well, depending on the platform different flags must be set, so it wouldn't be that unrelated.

iceoryx_posh/CMakeLists.txt
elBoberido: Just one thing, `CUSTOM_FLAGS` might be too generic. Just in case iceoryx is part of a bigger framework which already use that variable. It would probably be better to have an iceoryx prefix.

`gcc_build_flags.cmake` is problematic, since this is also used for `clang` and `msvc` compiler ... and maybe in future for even more. I would just drop the `Detection` and then the name is quite generic.

dkroenke: Then we name it like this `ICEORYX_FLAGS` in `build_flags.cmake`

elBoberido: but it's not only build flags, it also set some cmake variables which are used at cmake configure time ;)

prasannabhat: Summarising the discussions (plus new points)

- Rename `IceoryxPlatformDetection.cmake` ->  `IceoryxPlatform.cmake` (as suggested by @elBoberido). This name reflects the purpose somehow. (or more near to describing the purpose :wink: :)
- Move `sanitize` related flags to this file

I would still prefer to use separate flags `SANITIZER_COMPILER_FLAGS` & `SANITIZER_LINKER_FLAGS` because

- `sanitize` option enabled for selected components (`iceoryx_utils` & `iceoryx_platform`)
- This will be expanded with additional flags in future  (other sanitizers)

tools/run_all_tests.sh
prasannabhat: I guess this would be okay because LeakSanitizer runs in Linux build

iceoryx_posh/CMakeLists.txt
dkroenke: Regarding the name for the flags it should be named as instrumentation flags like they do it in GCC: [https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) because sanitizer are only one aspect. There are other tools like gcov who needs also linker flags.
Proposal: INSTRUMENTATION_COMPILE_FLAGS and INSTRUMENTATION_LINKER_FLAGS

To your issue with the paths: You can try it with CMAKE_SOURCE_DIR which gives you the path to the toplevel cmake. In our case it is iceoryx_meta, which maybe can used as starting point. Keep in mind that this only work in standalone build of iceoryx.

elBoberido: We could have a `ICEORYX_SANITIZER_FLAGS` and add them to the `ICEORYX_INSTRUMENTATION_FLAGS` so that we only need to add one variable to the target_compile_options and still have a separation

prasannabhat: I have reworked `sanitize` flags & defined `ICEORYX_SANITIZER_FLAGS`

However I doubt if we can have one global `ICEORYX_INSTRUMENTATION_FLAGS` (which includes `sanitize` & `coverage`), because `sanitize` & `coverage` has to be applied to targets separately 
e.g. `coverage` is applied to all the targets (including tests). However `sanitize` is applied to only core components.
Theoretically its possible to apply `sanitize` to all targets , however that will

- unnecessarily bloat the targets
- applies strict checks for all the components (including tests), which may leave the the **developers**  frustrated :wink:

dkroenke (changes requested):
The only thing left: there is one commit which has no iox tag in his name: https://github.com/eclipse/iceoryx/pull/400/commits/218318947c515ed320dac3e96220462f61b4b888

prasannabhat:
> The only thing left: there is one commit which has no iox tag in his name: [2183189](https://github.com/eclipse/iceoryx/commit/218318947c515ed320dac3e96220462f61b4b888)

oops ..that got missed !
@dkroenke I have created issue #423 to analyse the errors. (will add the details later)

dkroenke (dismissed):

iceoryx_meta/sanitizer_blacklist/lsan_runtime.txt
```diff
@@ -0,0 +1,21 @@
...
+
+leak:libacl.so.1
+leak:iox::posix::MessageQueue::timedReceive
+leak:iox::posix::UnixDomainSocket::timedReceive
+leak:iox::posix::MessageQueue::receive
+leak:test_base_port.cpp
+leak:BaseSubscriberTest_ReceivedSamplesAreAutomaticallyDeletedWhenOutOfScope_Test
+leak:BaseSubscriberTest_ReceiveReturnsAllocatedMemoryChunksWrappedInSample_Test
+leak:BaseSubscriberTest_AttachingAttachedSubscriberToNewWaitsetDetachesItFromOriginalWaitset_Test
+leak:iox_sub_test_deinitSubscriberDetachesTriggerFromWaitSet_Test::iox_sub_test_deinitSubscriberDetachesTriggerFromWaitSet_Test
```
elBoberido, line 20: Could you please create issues for this? At least the ones not in tests.

iceoryx_utils/cmake/IceoryxPlatform.cmake
```diff
@@ -48,3 +48,32 @@ if(BUILD_STRICT)
         set(ICEORYX_WARNINGS ${ICEORYX_WARNINGS} -Werror)
     endif (  )
 endif(BUILD_STRICT)
+
+if(sanitize)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+        # NOTE : This works only when iceoryx is built standalone , in which case CMAKE_SOURCE_DIR point to iceoryx_meta
+        set(ICEORYX_SANITIZER_BLACKLIST -fsanitize-blacklist=${CMAKE_SOURCE_DIR}/sanitizer_blacklist/asan_compile_time.txt)
```
elBoberido, line 55: why not just putting the blacklist right to this file, then it works all the time?

iceoryx_utils/cmake/IceoryxPlatform.cmake
```diff
@@ -48,3 +48,32 @@ if(BUILD_STRICT)
...
+    endif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+    
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+        set(ICEORYX_SANITIZER_COMMON_FLAGS -fno-omit-frame-pointer -fno-optimize-sibling-calls -g -O1)
```
elBoberido, line 59: is `-O1` required? I'm just asking because I'm not sure if it's overiden later on in cmake

iceoryx_posh/CMakeLists.txt
```diff
@@ -152,14 +152,15 @@ target_include_directories(iceoryx_posh
     source/version
     source/runtime
 )
-target_link_libraries(iceoryx_posh
+target_link_libraries(iceoryx_posh 
     PUBLIC
     iceoryx_utils::iceoryx_utils
+    ${ICEORYX_SANITIZER_FLAGS}
```
elBoberido, line 158: can't this be private?

.github/workflows/sanitize.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: sudo apt-get update && sudo apt-get install -y libacl1-dev libncurses5-dev
+
+      - name : Checkout
+        uses: actions/checkout@v2
+
+      - name: Build
+        run: |
+          export NUM_CORES=`nproc`
+          # build tests , dont run them
+          $GITHUB_WORKSPACE/tools/iceoryx_build_test.sh clang build-test with-dds-gateway sanitize clean -j $NUM_CORES
```
elBoberido, line 31: could you add `strict`? Else we might introduce new warnings into clang

iceoryx_posh/CMakeLists.txt
elBoberido: I'm a little bit torn by limiting it just to core targets. It might result in some technical dept in our tests, which we already have. It's okay for now, since fixing all the stuff in the tests would blow up this PR, but we should at least consider it.

iceoryx_meta/sanitizer_blacklist/lsan_runtime.txt
dkroenke: He has begun here: https://github.com/eclipse/iceoryx/issues/423 @prasannabhat

iceoryx_utils/cmake/IceoryxPlatform.cmake
elBoberido: there might be a `beside` missing right of `right`

prasannabhat: > why not just putting the blacklist right to this file, then it works all the time?

This is the method [suggested](https://clang.llvm.org/docs/AddressSanitizer.html#suppressing-errors-in-recompiled-code-blacklist) 



> there might be a `beside` missing right of `right`

I dint quite get that !

iceoryx_posh/CMakeLists.txt
prasannabhat: May be we can create a follow up issue to enable `sanitize` for all the targets ?

iceoryx_utils/cmake/IceoryxPlatform.cmake
prasannabhat: I think it wont break the sanitizer. Its more about getting a readable stack trace of the error.

iceoryx_posh/CMakeLists.txt
prasannabhat: I have set  `ICEORYX_SANITIZER_FLAGS` as PRIVATE for all the used targets.
I am not sure what exactly are the implications.
From [cmake documentation](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) 

>  Libraries and targets following PRIVATE are linked to, but are not made part of the link interface

May be the dependent libraries dont carry over this flag, by marking it PRIVATE

elBoberido: yes, if you use iceoryx_posh in your application and want to use the sanitizer, you need to set the flags yourself. Usually only the libraries should be public which are somehow exposed by the public API, e.g. if a posix header is included in a public header, we should also use PUBLIC in the link interface, else the user needs to add it herself, even if she uses that lib only by including our header. If the library is not exposed to the public API, then the link interface should be PRIVATE

iceoryx_utils/cmake/IceoryxPlatform.cmake
prasannabhat: okay ..now I understand what you meant.
I am not sure how do I get the path to `IceoryxPlatform.cmake` (when this file is included in another cmake file)

iceoryx_posh/CMakeLists.txt
prasannabhat: @elBoberido Created #424 . Please add if I missed anything

dkroenke (approved):

iceoryx_utils/cmake/IceoryxPlatform.cmake
elBoberido: another possibility would be to use `file({WRITE | APPEND} <filename> <content>...)` from cmake and create those files on the fly in e.g. `CMAKE_BUILD_DIR`. It's up to you if you want to do the hassle

elBoberido (approved):

Issue #14, opened 2019-12-05
Reserved chunkinfo user payload header


rex-schilasky:
## Brief feature description

When sending a payload chunk via publishers _sendChunk_ method a user can pass the payload memory pointer and it's size to transfer it with zero copy behavior to the subscribers side. If there is any additional content that needs to be transferred too (like counter, user timestamp, hash values ..) there is no option to apply this information additionally into the chunk header. So the only way to transport the payload data and some additional header information as single chunk is to do an extra memory copy in a new payload memory block before sending it as one chunk. So for this use case the amazing zero memory concept is broken.

## Possible solution

Provide a reserved space in the ChunkInfo header to hold / transfer a user specific payload header that is not in the same linear memory like the payload data.

---

edited by @elBoberido 

# Progress Tracking

- [x] Design Document
- [x] Extend the current `ChunkHeader` and write test for the new functionality
- [x] remove `ChunkInfo` and use the refactored `ChunkHeader` in the code base
- extend the API to access the new functionality
  - [x] typed API
  - [x] untyped API
  - [x] C API
- [x] restrict access to `ChunkHeader` members and hand out references instead of pointer
- [ ] documentation and examples
- cleanup
  - [ ] remove all TODOs
  - [x] remove `chunk_mock_dds.hpp` and user `chunk_mock.hpp` instead
  - [x] in order to reduce ambiguity don't use just `payload` but `chunkPayload` and `userPayload`. The `chunkPayload` consists of the `customHeader/userHeader` and `userPayload`

michael-poehnl:
Thanks for the request @rex-schilasky 

Having a user defined extra space in the chunk header would be possible. But I'm wondering if this could also be solved in your middleware code that uses iceoryx as transport layer. 
I.e. when calling allocateChunk() you provide "your payload size + your specific header size". Then you do some pointer magic to provide to your next API layer a pointer to chunk pointer address + sizeof(your_header). You leave some extra space at the beginning of the iceoryx payload that you could fill when sending. 

This would be similar to what we are doing under the hood. We also provide a pointer to the payload to the user but have in front of that memory space for our chunk header. 

What do you think?

rex-schilasky:
[...]

Issue #27, opened 2019-12-16
Request/Response communication with iceoryx


budrus:
## Brief feature description

Besides publish/subscribe also request/response communication shall be supported

## Detailed information

So far iceoryx only supports publish/subscribe. For frameworks like ROS2 also a request/response communication shall be supported (services in ROS2).  The building blocks in posh can be re-used here, so start with client and server ports and then do the layers above

We will have the same layered architecture as for publishers and subscribers. With `ChunkSender` and `ChunkReceiver` common building blocks are reused but a `Client` and a `Server` will use one of each for the bidirectional data transfer with requests and responses. 
This image is a simplified view that shows the class hierarchy and the public method names for the typed C++ API. 

![component](https://user-images.githubusercontent.com/55156294/109230075-8db9ee00-77c4-11eb-8a29-913808769f35.png)

1. It will also have a zero-copy API, i.e. we can loan requests and take responses on the `Client` side and take requests and loan responses on the `Server` side.   
1. Like a `Publisher`, a `Server` can `offer()` or `stopOffere()`
2. Similar to a `Subscriber` `subscribe()` and `unsubscribe()`, the `Client` can `connect()`, `disconnect()` and getConnectionState()
3. You can poll the `Server` for requests and the `Client` for responses or use a `Waitset` to wait for requests and responses 

- [x] 1. Design and interface implementation for ports

- [x] 2. Implementation and unit tests for ports

- [x] 3. Integration test for ports
   - [x] including QueueFullPolicy tests with a real Client- and ServerPort

- [x] 4. C API 
   - [x] client
   - [x] request/response header
   - [x] server
   - [x] examples
   - [ ] (optional) add `iceoryx_posh_types.hpp` RPC related config stuff to `config.h`.
   - [ ] (optional) add missing functionality to chunk header

- [x] 5. C++ API
   - [x] typed C++ API 
   - [x] tests for typed C++ API
   - [x] untyped C++ API
   - [x] tests for untyped C++ API
   - [x] send returns an expected instead of void

- [x] 6. Integration in runtime and RouDi
   - [x] Make `ServerPort`'s findable via `ServiceDiscovery` 

- [x] 7. Common port queue policies
    - rename `SubscriberTooSlowPolicy` -> `ConsumerTooSlowPolicy`
        - alias for ConsumerTooSlowPolicy for easy migration to new API
        - rename `WAIT_FOR_SUBSCRIBER` -> `WAIT_FOR_CONSUMER`
[...]

PR #403, opened 2020-12-01
iox-#14 chunk header design with custom header

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
This is a test for a workflow with a design document PR before a full implementation.

It's the first PR for a custom extension to the chunk header as requested by @rex-schilasky and also needed for the request response feature #27.

I would like to start implementing this up to the integration into the publisher. For that maybe a PoC is needed. I'm quite confident that the proposed design for the ChunkHeader works regardless of the final decision how to use it in the publisher. So there will be at least two further PRs for this issue.

If the implementation requires design changes. The document will be updated with the corresponding PR.

1. [ ] Merge

## References

- Partly closes #14

### DIFF
--- a/iceoryx_posh/doc/chunk_header.md
+++ b/iceoryx_posh/doc/chunk_header.md
@@ -1,0 +1,218 @@
+# Summary
+
+Chunks are the transport capsules in iceoryx. They store data from a publisher as payload and are sent to one or more subscriber. Furthermore, there is some meta information which is stored alongside the payload, e.g. the size and the origin of the chunk. This data is composed in the `ChunkHeader` and located at the front of the chunk. Custom meta information can be added to extend the data from the `ChunkHeader` and tailor the chunk to specific use cases. While this makes the chunk layout more complex, this complexity would otherwise be distributed at different locations all over the code base, e.g. in the request/response feature. Additionally, the adjustments for the custom header makes arbitrary alignments of the payload trivial to implement.
+
+# Terminology
+
+| Name              | Description                                              |
+| :---------------- | :------------------------------------------------------- |
+| Chunk             | a piece of memory                                        |
+| Chunk Header      | contains meta information related to the chunk           |
+| Custom Header     | contains custom meta information, e.g. timestamps        |
+| Payload           | the user data                                            |
+| Back-Offset       | offset stored in front of the payload to calculate back to the chunk header |
+
+# Design
+
+## Considerations
+
+- it's not uncommon to record chunks for a later replay -> detect incompatibilities on replay
+- iceoryx runs on multiple platforms -> endianness of recorded chunks might differ
+- for tracing, a chunk should be uniquely identifiable -> store origin and sequence number
+- the chunk is located in the shared memory, which will be mapped to arbitrary positions in the address space of various processes -> no absolute pointer are allowed
+- aligning the `ChunkHeader` to 32 bytes will ensure that all member are on the same cache line and will improve performance
+- in order to reduce complexity, the alignment of the custom header must not exceed the alignment of the `ChunkHeader`
+
+## Solution
+
+### ChunkHeader Definition
+```
+struct alignas(32) ChunkHeader
+{
+    uint32_t m_chunkSize;
+    uint8_t m_chunkHeaderVersion;
+    uint8_t m_reserved1;
+    uint8_t m_reserved2;
+    uint8_t m_reserved3;
+    uint64_t m_originId;
+    uint64_t m_sequenceNumber;
+    uint32_t m_payloadSize;
+    uint32_t m_payloadOffset;
+};
+```
+
+- **m_chunkSize** is the size of the whole chunk
+- **m_chunkHeaderVersion** is used to detect incompatibilities for record&replay functionality
+- **m_reserved1**, **m_reserved2**, **m_reserved3** are currently not used and set to `0`
+- **m_originId** is the unique identifier of the publisher the chunk was sent from
+- **m_sequenceNumber** is a serial number for the sent chunks
+- **m_payloadSize** is the size of the used payload of the chunk
+- **m_payloadOffset** is the offset of the payload relative to the begin of the chunk
+
+### Framing
+
+For back calculation from the payload pointer to the `ChunkHeader` pointer, the payload offset must be accessible from a defined position relative to the payload. Lets call this `back-offset`. This is solved by storing the offset in the 4 bytes in front of the payload. In case of a simple layout where the `ChunkHeader` is adjacent to the payload, this nicely overlaps with the position of `m_payloadOffset` and no memory is wasted. In more complex cases, the offset has to be stored a second time. If the payload alignment requires some padding from the header extension, this memory is used to store the offset.
+
+1. No custom header and alignment doesn't exceed the `ChunkHeader` alignment
+
+```
+  sizeof(ChunkHeader)     m_payloadSize
+|-------------------->|------------------->|
+|                     |                    |
++=====================+====================+==================================+
+|                     |                    |                                  |
+|     ChunkHeader     |      Payload       |  Padding                         |
+|                     |                    |                                  |
++=====================+====================+==================================+
+|                     |                                                       |
+|   m_payloadOffset   |                                                       |
+|-------------------->|                                                       |
+|                                m_chunkSize                                  |
+|---------------------------------------------------------------------------->|
+```
+
+2. No custom header and alignment exceeds the `ChunkHeader` alignment
+
+```
+  sizeof(ChunkHeader)            back-offset    m_payloadSize
+|-------------------->|                |<---|------------------->|
+|                     |                |    |                    |
++=====================+=====================+====================+============+
+|                     |                .    |                    |            |
+|     ChunkHeader     |                .    |      Payload       |  Padding   |
+|                     |                .    |                    |            |
++=====================+=====================+====================+============+
+|                                           |                                 |
+|           m_payloadOffset                 |                                 |
+|------------------------------------------>|                                 |
+|                                                                             |
+|                                m_chunkSize                                  |
+|---------------------------------------------------------------------------->|
+```
+
+Depending on the address of the chunk there is the chance that `ChunkHeader` is the still adjacent to the payload. In this case, the framing looks exactly like it case 1.
+
+3. Custom header is used
+
+```
+  sizeof(ChunkHeader)            back-offset    m_payloadSize
+|-------------------->|                |<---|------------------->|
+|                     |                |    |                    |
++=====================+=====================+====================+============+
+|                     |   Custom    |  .    |                    |            |
+|     ChunkHeader     | ChunkHeader |  .    |      Payload       |  Padding   |
+|                     |  Extension  |  .    |                    |            |
++=====================+=====================+====================+============+
+|                                           |                                 |
+|           m_payloadOffset                 |                                 |
+|------------------------------------------>|                                 |
+|                                                                             |
+|                                m_chunkSize                                  |
+|---------------------------------------------------------------------------->|
+```
+
+### Payload Offset Calculation
+
+1. No custom header and payload alignment doesn't exceed the `ChunkHeader` alignment
+
+```
+payloadOffset = sizeof(ChunkHeader);
+```
+
+2. No custom header and payload alignment exceeds the `ChunkHeader` alignment, which means the payload is either adjacent to the `ChunkHeader` or there is a padding with at least the size of `ChunkHeader` in front of the payload and therefore enough space to store the `back-offset`
+
+```
+headerEndAddress = addressof(chunkHeader) + sizeof(chunkHeader);
+payloadAddress = align(headerEndAddress, payloadAlignment);
+payloadOffset = payloadAddress - addressof(chunkHeader);
+```
+
+3. Custom header is used
+
+```
+headerEndAddress = addressof(chunkHeader) + sizeof(chunkHeader) + sizeof(customHeader);
+potentialBackOffsetAddress = align(headerEndAddress, alignof(payloadOffset));
+potentialPayloadAddress = potentialBackOffsetAddress + sizeof(payloadOffset);
+payloadAddress = align(potentialPayloadAddress, payloadAlignment);
+payloadOffset = payloadAddress - addressof(chunkHeader);
+```
+
+### Required Chunk Size Calculation
+
+In order to fit the custom header and the payload into he chunk, a worst case calculation has to be done. We can assume that a chunk is aligned to 32 bytes, which is also the alignment of the `ChunkHeader`.
+
+1. No custom header and payload alignment doesn't exceed the `ChunkHeader` alignment
+
+```
+chunkSize = sizeof(chunkHeader) + payloadSize;
+```
+
+2. No custom header and payload alignment exceeds the `ChunkHeader` alignment
+
+Worst case scenario is when a part of the `ChunkHeader` crosses the payload alignment boundary, so that the payload must be aligned to the next boundary.
+```
+sizeLeftOfThePayloandAlignmentBoundary = sizeof(chunkHeader) - alignof(chunkHeader);
+chunkSize = sizeLeftOfThePayloandAlignmentBoundary + payloadAlignment + payloadSize;
+```
+
+3. Custom header is used
+
+Similar to case 2, but in this case it is the `back-offset` which might cross the payload alignment boundary.
+```
+headerSize = sizeof(chunkHeader) + sizeof(customHeader)
+sizeLeftOfThePayloandAlignmentBoundary = align(headerSize, alignof(payloadOffset));
+maxAlignment = max(alignof(payloadOffset), payloadAlignment);
+chunkSize = sizeLeftOfThePayloandAlignmentBoundary + maxAlignment + payloadSize;
+```
+
+### Accessing Chunk Header Extension
+
+The `ChunkHeader` has a template method to get custom header. There is a risk to use this wrong by accident, but there is currently no better solution for this.
+
+Since the custom header is always adjacent to the `ChunkHeader`, the formula to get the custom header is:
+```
+customHeader = addressOf(chunkHeader) + sizeof(chunkHeader);
+```
+
+### ChunkHeader Methods
+
+- `void* payload()` returns a pointer to the payload
+- `template <typename T> T* customHeader()` returns a pointer to the custom header
+- `static ChunkHeader* fromPayload(const void* const payload)` returns a pointer to the `ChunkHeader` associated to the payload
+
+### Integration Into Publisher/Subscriber API
+
+The `Publisher` has additional template parameters for the custom header and payload alignment. Alternatively this could also be done with the `allocate` method, but that increases the risk of using it wrong by accident.
+
+Additionally, to fully integrate the custom header into the regular control flow, a pointer to a `ChunkHeaderHook` can be passed to the `Publisher` c'tor.
+
+The `ChunkHeaderHook` is not in the shared memory and has the following virtual methods:
+- `virtual void allocateHook(ChunkHeader& chunkHeader)` can be used to initialize the custom header
+- `virtual void deliveryHook(ChunkHeader& chunkHeader)` can be used to set members of the custom header, e.g. a timestamp
+
+Furthermore, the `Publisher` and `Subscriber` have access to the `ChunkHeader` and can use the `customHeader()` method to gain access to the custom header.
+
+### Pitfalls & Testing
+
+- when the payload is adjacent to the `ChunkHeader`, it must be ensured that `m_payloadOffset` overlaps with the `back-offset`, which is `sizeof(m_payloadOffset)` in front of the payload
+- to simplify calculation, it is assumed that the alignment of the custom header doesn't exceed the alignment of the `ChunkHeader`. This has to be enforced with an `assert`
+
+# Open Issues
+
+- the design was done with the intention to have a custom header of arbitrary size, if the size is limited to e.g. 32 bytes, some things could be simplified
+- untyped API and C bindings
+    - PoC is needed
+- user defined sequence number
+    - this can probably be done by a `ChunkHeaderHook`
+    - alternatively, a flag could be introduce into the `ChunkHeader`
+- user defined timestamp
+    - this should probably be in a custom header
+- mempool configuration
+    - currently we specify the payload size and the size of the `ChunkHeader` is added automatically
+    - with the new approach, part of the specified payload size might be used for the custom header
+    - part of the specified payload might also be used as padding for the payload alignment
+    - shall we change this back to just specifying the full chunk size?
+    - shall we add an option to specify the custom header size and the payload alignment?
+- is it necessary to store a flag in `ChunkHeader` if a custom extension is used?
+- do we want to store the version of the custom extension in the `ChunkHeader`, similarly to `m_chunkHeaderVersion`
+- for record&replay the custom header is totally opaque, which might lead to problems if e.g. a timestamp is stored in the custom header and needs to be updated for the replay
--- a/iceoryx_posh/doc/chunk_header.md
+++ b/iceoryx_posh/doc/chunk_header.md
@@ -56,17 +56,15 @@ For back calculation from the payload pointer to the `ChunkHeader` pointer, the 
 1. No custom header and alignment doesn't exceed the `ChunkHeader` alignment
 
 ```
-  sizeof(ChunkHeader)     m_payloadSize
-|-------------------->|------------------->|
-|                     |                    |
-+=====================+====================+==================================+
-|                     |                    |                                  |
-|     ChunkHeader     |      Payload       |  Padding                         |
-|                     |                    |                                  |
-+=====================+====================+==================================+
-|                     |                                                       |
-|   m_payloadOffset   |                                                       |
-|-------------------->|                                                       |
+ sizeof(ChunkHeader)    m_payloadSize
+|------------------>|--------------------->|
+|                   |                      |
++===================+======================+==================================+
+|  ChunkHeader  ¦   |       Payload        |  Padding                         |
++===================+======================+==================================+
+|                   |                                                         |
+|  m_payloadOffset  |                                                         |
+|------------------>|                                                         |
 |                                m_chunkSize                                  |
 |---------------------------------------------------------------------------->|
 ```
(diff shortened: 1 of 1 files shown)


elBoberido:
oh, if someone doesn't know that feature, you can select `Display the rich diff` in the `Files changed` tab, which makes reviewing the markdown more comfortable

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+| Custom Header     | contains custom meta information, e.g. timestamps        |
+| Payload           | the user data                                            |
+| Back-Offset       | offset stored in front of the payload to calculate back to the chunk header |
+
+# Design
+
+## Considerations
+
+- it's not uncommon to record chunks for a later replay -> detect incompatibilities on replay
+- iceoryx runs on multiple platforms -> endianness of recorded chunks might differ
```
elBoberido, line 20: btw. we could store an endianness indicator in the `ChunkHeader`, e.g. by replacing `m_reserved2` and `m_reserved3` with a `uint16_t m_endianness = 0xA55A`. On big endian platforms the memory would be `... m_chunkHeaderVersion m_reserved1 0xA5 0x5A` and on little endian `... m_chunkHeaderVersion m_reserved1 0x5A 0xA5`. Alternatively, one byte only could be used as flag.
I just don't know if it makes sense to store it in the `ChunkHeader` or shift the responsibility to the record framework

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+Worst case scenario is when a part of the `ChunkHeader` crosses the payload alignment boundary, so that the payload must be aligned to the next boundary.
+```
+sizeLeftOfThePayloandAlignmentBoundary = sizeof(chunkHeader) - alignof(chunkHeader);
+chunkSize = sizeLeftOfThePayloandAlignmentBoundary + payloadAlignment + payloadSize;
+```
+
+3. Custom header is used
+
+Similar to case 2, but in this case it is the `back-offset` which might cross the payload alignment boundary.
+```
```
budrus, line 161: Also here an additional image would help

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+
+### ChunkHeader Methods
+
+- `void* payload()` returns a pointer to the payload
+- `template <typename T> T* customHeader()` returns a pointer to the custom header
+- `static ChunkHeader* fromPayload(const void* const payload)` returns a pointer to the `ChunkHeader` associated to the payload
+
+### Integration Into Publisher/Subscriber API
+
+The `Publisher` has additional template parameters for the custom header and payload alignment. Alternatively this could also be done with the `allocate` method, but that increases the risk of using it wrong by accident.
```
budrus, line 185: For me this is fixed for a specific publisher. So no need to have it in the allocate call

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+
+Additionally, to fully integrate the custom header into the regular control flow, a pointer to a `ChunkHeaderHook` can be passed to the `Publisher` c'tor.
+
+The `ChunkHeaderHook` is not in the shared memory and has the following virtual methods:
```
budrus, line 189: Can you write a rough example how the publisher and subscriber class and c'tor signatures would look like

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+- `virtual void allocateHook(ChunkHeader& chunkHeader)` can be used to initialize the custom header
+- `virtual void deliveryHook(ChunkHeader& chunkHeader)` can be used to set members of the custom header, e.g. a timestamp
```
budrus, line 191: This hooks are nice but also further extend the API. Would it be an option for you to not use them and let the user do all the work based on the `customHeader()` call. I.e. do a placement new and setting the values

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+- when the payload is adjacent to the `ChunkHeader`, it must be ensured that `m_payloadOffset` overlaps with the `back-offset`, which is `sizeof(m_payloadOffset)` in front of the payload
+- to simplify calculation, it is assumed that the alignment of the custom header doesn't exceed the alignment of the `ChunkHeader`. This has to be enforced with an `assert`
+
+# Open Issues
+
+- the design was done with the intention to have a custom header of arbitrary size, if the size is limited to e.g. 32 bytes, some things could be simplified
+- untyped API and C bindings
+    - PoC is needed
+- user defined sequence number
+    - this can probably be done by a `ChunkHeaderHook`
```
budrus, line 206: Then the user defined sequence number would be part of the custom chunk header. Maybe this is the best for now. If iceoryx is used in a framework that wants to have additional features based on the sequence number having an additional custom chunk header there is quite likely (e.g. there are timestamps etc.)

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+    - alternatively, a flag could be introduce into the `ChunkHeader`
+- user defined timestamp
+    - this should probably be in a custom header
+- mempool configuration
+    - currently we specify the payload size and the size of the `ChunkHeader` is added automatically
+    - with the new approach, part of the specified payload size might be used for the custom header
+    - part of the specified payload might also be used as padding for the payload alignment
+    - shall we change this back to just specifying the full chunk size?
+    - shall we add an option to specify the custom header size and the payload alignment?
+- is it necessary to store a flag in `ChunkHeader` if a custom extension is used?
```
budrus, line 216: This could be used for some additional checks. Maybe a custom header identifier. But then we would need some rules and maybe something like a reserved range for predefine custom header like the ones for client and server

budrus (changes requested):

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+
+Since the custom header is always adjacent to the `ChunkHeader`, the formula to get the custom header is:
+```
+customHeader = addressOf(chunkHeader) + sizeof(chunkHeader);
+```
+
+### ChunkHeader Methods
+
+- `void* payload()` returns a pointer to the payload
+- `template <typename T> T* customHeader()` returns a pointer to the custom header
```
budrus, line 180: what is returned if there is no custom header?

iceoryx_posh/doc/chunk_header.md
elfenpiff: @budrus I would disagree here.
Isn't the idea of the untyped publisher to acquire memory for any arbitrary type and the type and amount of memory can change with every call? With this also the alignment can change with every call!
For me it seems quite natural to have then a second argument in allocate the define the alignment.
`untypedPublisher.allocate(memSize, alignment)`

In the case of the typed publisher I totally agree with you but here we just call `loan`.

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+- user defined timestamp
+    - this should probably be in a custom header
+- mempool configuration
+    - currently we specify the payload size and the size of the `ChunkHeader` is added automatically
+    - with the new approach, part of the specified payload size might be used for the custom header
+    - part of the specified payload might also be used as padding for the payload alignment
+    - shall we change this back to just specifying the full chunk size?
+    - shall we add an option to specify the custom header size and the payload alignment?
+- is it necessary to store a flag in `ChunkHeader` if a custom extension is used?
+- do we want to store the version of the custom extension in the `ChunkHeader`, similarly to `m_chunkHeaderVersion`
```
elfenpiff, line 217: You could argue that the creator of the custom extension could add a field where they store the version number if they need it.

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+- user defined sequence number
+    - this can probably be done by a `ChunkHeaderHook`
+    - alternatively, a flag could be introduce into the `ChunkHeader`
+- user defined timestamp
+    - this should probably be in a custom header
+- mempool configuration
+    - currently we specify the payload size and the size of the `ChunkHeader` is added automatically
+    - with the new approach, part of the specified payload size might be used for the custom header
+    - part of the specified payload might also be used as padding for the payload alignment
+    - shall we change this back to just specifying the full chunk size?
```
elfenpiff, line 214: I would keep it as simple as possible and would just specify the full chunk size.

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+# Open Issues
+
+- the design was done with the intention to have a custom header of arbitrary size, if the size is limited to e.g. 32 bytes, some things could be simplified
+- untyped API and C bindings
+    - PoC is needed
+- user defined sequence number
+    - this can probably be done by a `ChunkHeaderHook`
+    - alternatively, a flag could be introduce into the `ChunkHeader`
+- user defined timestamp
+    - this should probably be in a custom header
```
elfenpiff, line 209: I completely agree.

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+The `ChunkHeaderHook` is not in the shared memory and has the following virtual methods:
+- `virtual void allocateHook(ChunkHeader& chunkHeader)` can be used to initialize the custom header
+- `virtual void deliveryHook(ChunkHeader& chunkHeader)` can be used to set members of the custom header, e.g. a timestamp
+
+Furthermore, the `Publisher` and `Subscriber` have access to the `ChunkHeader` and can use the `customHeader()` method to gain access to the custom header.
+
+### Pitfalls & Testing
+
+- when the payload is adjacent to the `ChunkHeader`, it must be ensured that `m_payloadOffset` overlaps with the `back-offset`, which is `sizeof(m_payloadOffset)` in front of the payload
+- to simplify calculation, it is assumed that the alignment of the custom header doesn't exceed the alignment of the `ChunkHeader`. This has to be enforced with an `assert`
```
elfenpiff, line 198: I think it is a complex topic and I suggest solving this issue right away otherwise one of us as to get into the offset and alignment calculation next year and figure everything out again. 
Where would be the challenge? Could we get some `maxHeaderAlignment = std::max(alignof(ChunkHeader), customHeaderAlignment)` or do I miss something?

iceoryx_posh/doc/chunk_header.md
mossmaurice: Good point @sculpordwarf  @frank-schmidtner What do you think, should this be take care of by the recording framework? Any preferences?

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+
+# Design
+
+## Considerations
+
+- it's not uncommon to record chunks for a later replay -> detect incompatibilities on replay
+- iceoryx runs on multiple platforms -> endianness of recorded chunks might differ
+- for tracing, a chunk should be uniquely identifiable -> store origin and sequence number
+- the chunk is located in the shared memory, which will be mapped to arbitrary positions in the address space of various processes -> no absolute pointer are allowed
+- aligning the `ChunkHeader` to 32 bytes will ensure that all member are on the same cache line and will improve performance
```
mossmaurice, line 23: Why 32 byte?

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+    uint8_t m_reserved3;
+    uint64_t m_originId;
+    uint64_t m_sequenceNumber;
+    uint32_t m_payloadSize;
+    uint32_t m_payloadOffset;
+};
+```
+
+- **m_chunkSize** is the size of the whole chunk
+- **m_chunkHeaderVersion** is used to detect incompatibilities for record&replay functionality
```
mossmaurice, line 45: Couldn't this be a compile time thing? How are we dealing with this currently? Using `iceoryx_posh/include/iceoryx_posh/version/version_info.hpp`?

iceoryx_posh/doc/chunk_header.md
mossmaurice: I suppose we could use SFINAE (`enable_if` meta programming magic) to prevent the user calling this method in case there is no custom header.

mossmaurice (commented):
Great work and a good example for our future design documents!

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+- **m_reserved1**, **m_reserved2**, **m_reserved3** are currently not used and set to `0`
+- **m_originId** is the unique identifier of the publisher the chunk was sent from
```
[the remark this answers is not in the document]
elBoberido: good question. AFAIK currently it would change. @elfenpiff does it?

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,218 @@
...
+1. No custom header and payload alignment doesn't exceed the `ChunkHeader` alignment
+
+```
+chunkSize = sizeof(chunkHeader) + payloadSize;
+```
+
+2. No custom header and payload alignment exceeds the `ChunkHeader` alignment
+
+Worst case scenario is when a part of the `ChunkHeader` crosses the payload alignment boundary, so that the payload must be aligned to the next boundary.
+```
```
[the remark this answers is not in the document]
elBoberido: yes, you are right, as long as the size of `ChunkHeader` does not exceed the alignment, it's 0. This makes it just future prove in case this changes, e.g. because we decide to decrease the alignment because we would waste memory when the actual size would be e.g. 40 bytes. The chunks could still be aligned to 32 bytes, but the `ChunkHeader` alignment could be 8.

iceoryx_posh/doc/chunk_header.md
elBoberido: I think elfenpiff is right. After thinking a bit about this, it might be helpful to have this in allocate. Maybe we have to write down some use cases for the untyped publisher.

iceoryx_posh/doc/chunk_header.md
elBoberido: sure. I will put that in the `Open Issues` section, since this might change when it's actually used in an example and we notice that it feels weird to use.

iceoryx_posh/doc/chunk_header.md
elBoberido: yes, that would be the simplest way, e.g.
```
auto pub = iox::popo::TypedPublisher<MyPayload, MyHeader>(serviceDescription);
pub.loan()
    .and_then([&](auto& sample) {
        sample.getHeader()->customHeader<MyHeader>().data = 73;
        sample->a = 42;
        sample->b = 13;
        sample.publish();
    })
    .or_else([](iox::popo::AllocationError) {
        // Do something with error.
    });
```
The `CustomHeaderHook` can be used to automate this stuff and also apply data which maybe not valid a allocation, e.g. setting a timestamp right before delivery.

iceoryx_posh/doc/chunk_header.md
elBoberido: either that or the `ChunkHeaderHook`could also be used without a custom header, to overwrite the sequence number in `ChunkHeader`. I'm not yet sure if this is a good idea though. maybe we don't want to give the user write access to the `ChunkHeader`

iceoryx_posh/doc/chunk_header.md
elBoberido: this could be similar to `IANA` https://tools.ietf.org/id/draft-cotton-tsvwg-iana-ports-00.html#privateports. We could maintain a list of known header IDs, and everything with an ID larger than `0xC000` is free to use

iceoryx_posh/doc/chunk_header.md
elBoberido: if we store a flag, a nullptr. Maybe an `optional<T*>` would be better

iceoryx_posh/doc/chunk_header.md
elBoberido: yes, but then every creator of the custom header would need to repeat the code

iceoryx_posh/doc/chunk_header.md
elBoberido: if the alignment of the custom header is larger than the `ChunkHeader` we cannot assume that the position of the custom header is at `addressof(chunkHeader) + sizeof(chunkHeader)` which means we need to also store the custom header offset. I cannot think of a scenario where you want a special alignment for the custom header and also no scenario where it is necessary that the alignment should be larger than 8 bytes

iceoryx_posh/doc/chunk_header.md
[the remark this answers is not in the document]
elfenpiff: @budrus the idea was that the originId is unique in the complete system. Therefore, we would have a unique roudi or system id (first 32 bits) and a unique port id (last 32 bits) which identifies the port. And the combination would be uniquel in the whole universe.

iceoryx_posh/doc/chunk_header.md
elBoberido: historical reasons. Up until now the chunks are aligned to 32 bytes to ensure that we also support an alignment of up to 32 bytes for the payload. With this change, the alignment of the  payload can be specified, so it's not necessary anymore. It might help with performance if the `ChunkHeader` is not split at two cache lines and 32 is currently the size of the `ChunkHeader`. We might want to trade this for more memory efficiency in case we need to increase the `ChunkHeader` to e.g. 40 bytes, then the alignment would enforce a size of 64 bytes with 24 padding bytes. In this case it might be better to have an alignment of `8` bytes with a potential cost of distributing the `ChunkHeader` on multiple cache lines. This could also be a compile time option to shift the decision to the user.

iceoryx_posh/doc/chunk_header.md
elBoberido: yes, it would be ca compile time thing, so this should probably be a `const uint8_t m_chunkHeaderVersion`. Have to think about the implications about this, e.g. the `ChunkHeader` wouldn't be copy asignable anymore

iceoryx_posh/doc/chunk_header.md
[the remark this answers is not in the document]
budrus: @elfenpiff This wasn't the question. If I stop and restart an application, the same publisher will have the same ID or a new one?

iceoryx_posh/doc/chunk_header.md
budrus: Puuh, You really see a use case where one publisher is used with different alignments for different samples? Quite academic IMHO. If you need a special alignment that is > 32 byte, I would assume that every sample would have the same alignment as there is always the same processing for each sample and this has a specific alignment requirement

iceoryx_posh/doc/chunk_header.md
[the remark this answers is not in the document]
elfenpiff: @budrus There are no guarantees that you receive the same ID after a stop and restart of an application.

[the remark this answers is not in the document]
elBoberido: @budrus additionally if you restart RouDi there is a high chance that the IDs are completely arbitrary

iceoryx_posh/doc/chunk_header.md
elfenpiff: Let's say you would like to send samples which are containing data for vector operations implemented with AVX512 or SSE2/3/4. Then you need an alignment of 64 bytes. And currently we explicitly allow the user to send a bool as next sample where an alignment of 1 would suffice.

I think we should go all in in this case and support really the stuff the untyped publisher promises the expert user.

iceoryx_posh/doc/chunk_header.md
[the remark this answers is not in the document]
elBoberido: updated with some ASCII art

iceoryx_posh/doc/chunk_header.md
elBoberido: updated with some ASCII art

iceoryx_posh/doc/chunk_header.md
elBoberido: Like elfenpiff said, we can already specify the payload size when a sample is loaned, therefore it makes sense to also be able to specify the alignment. I would specify the custom header with the publisher/subscriber, though.

iceoryx_posh/doc/chunk_header.md
elBoberido: done

iceoryx_posh/doc/chunk_header.md
elBoberido: clarified in the document

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,323 @@
...
+// subscriber option 2
+auto customHeaderSize = sizeOf(MyHeader);
+auto pub = iox::popo::UntypedSubscriber(serviceDescription, customHeaderSize);
+sub->take()
+    .and_then([](auto& sample){
+        std::cout << "Custom header data: " << sample.getHeader()->customHeader<MyHeader>().data << std::endl;
+        std::cout << "Payload data: " << static_cast<const MyPayload*>(sample->get()).data << std::endl;
+    });
+```
+    - option 1 has the benefit to catch a wrong alignment of the custom header and would be necessary if we make the `Sample` aware of the custom header
```
elBoberido, line 307: @rex-schilasky 
this is an adjustment of our current untyped API. The payload size is accessible by the `ChunkHeader`.

iceoryx_posh/doc/chunk_header.md
budrus: The untyped subscriber is very likely used when iceoryx is integrated in a bigger framework like eCAL. At least I would go for a default as for most of them it will be don't care

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,323 @@
...
+{
+  public:
+    void deliveryHook(ChunkHeader& chunkHeader) override
+    {
+        chunkHeader.customHeader<CustomHeader>().timestamp = myTimeProvider::now();
+    }
+};
+
+auto customHeaderHook = MyChunkHeaderHook<MyHeader>();
+auto pub = iox::popo::TypedPublisher<MyPayload>(serviceDescription, customHeaderHook);
```
budrus, line 270: Don't you need MyHeader as template argument here? Or can you solve it with the provided Hook?

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,323 @@
...
+    .or_else([](iox::popo::AllocationError) {
+        // Do something with error.
+    });
+
+// subscriber option 1
+auto pub = iox::popo::UntypedPublisher<MyHeader>(serviceDescription);
+
+// subscriber option 2
+auto customHeaderSize = sizeOf(MyHeader);
+auto pub = iox::popo::UntypedSubscriber(serviceDescription, customHeaderSize);
```
budrus, line 300: Maybe avoiding a template here makes sense. And default for customHeaderSize would be zero. Then there is no change in the API if the custom header is not used

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,323 @@
...
+sub->take()
+    .and_then([](auto& sample){
+        std::cout << "Custom header data: " << sample.getHeader()->customHeader<MyHeader>().data << std::endl;
+        std::cout << "Payload data: " << static_cast<const MyPayload*>(sample->get()).data << std::endl;
```
elBoberido, line 304: @ithier what do you think about adjusting the `Sample::get` method to this
```
template <typename T>
class Sample {
public:
    template <typename U = T>
    U* get();

private:
    cxx::unique_ptr<T> m_samplePtr{[](T* const) {}};
};

template <typename T>
template <typename U>
U* Sample<T>::get() {
    static_assert(std::is_same<U, T>::value, "incompatible types");
    return m_samplePtr.get();
}

template <>
template <typename U>
U* Sample<void>::get() {
    return static_cast<U*>(m_samplePtr.get(););
}
```
the we could have
```
Sample<void> s;
auto untypedPointer = s.get();
auto typedPointer = s.get<Foo>();
```

iceoryx_posh/doc/chunk_header.md
elBoberido: yes, default alignment would be 8 bytes

iceoryx_posh/doc/chunk_header.md
elBoberido: No, that's the nice thing here. It's deduced from the Hook. Less things to get wrong :)

iceoryx_posh/doc/chunk_header.md
elBoberido: I would still lean towards option 1, since it's more save and I can't think of a use case to change the header between samples. The small change would be to use `UntypedPublisher<>(serviceDescription)` if the custom header is not used

iceoryx_posh/doc/chunk_header.md
```diff
@@ -0,0 +1,327 @@
...
+    - shall we change this back to just specifying the full chunk size?
+    - shall we add an option to specify the custom header size and the payload alignment?
+- is it necessary to store a flag in `ChunkHeader` if a custom extension is used?
+    - we could maintain a list of known header IDs or ranges of IDs similar to `IANA` https://tools.ietf.org/id/draft-cotton-tsvwg-iana-ports-00.html#privateports
+    - the IDs could be stored in the `ChunkHeader` and everything with an ID larger than `0xC000` is free to use
+    - to make this somewhat save, the `ChunkHeaderHook` must be mandatory with e.g. a `virtual uint16_t getId() = 0;` method which will be called in the `ChunkSender`
+- do we want to store the version of the custom extension in the `ChunkHeader`, similarly to `m_chunkHeaderVersion`
+- for record&replay the custom header is totally opaque, which might lead to problems if e.g. a timestamp is stored in the custom header and needs to be updated for the replay
+    - if we maintain a list of known custom header IDs and also store the custom header version, a record&replay framework could implement the required conversions
```
elBoberido, line 327: what do you think of this? Could nicely solve some issues

iceoryx_posh/doc/chunk_header.md
budrus: OK, go for it

iceoryx_posh/doc/chunk_header.md
budrus: I would only allow access to the custom chunk header in the hook. The chunk header is then a header only iceoryx writes. Having these separation of concerns makes sense for me when we have a custom header

budrus (approved):
For me fine for now. The remaining questions and options are to be addressed in the implementation and integration

iceoryx_posh/doc/chunk_header.md
ithier: Much nicer than making the user do a cast themselves.
I like it.

ithier (approved):
Concept looks sound.
Good write-up!

iceoryx_posh/doc/chunk_header.md
elBoberido: @ithier there is also this discussion https://github.com/eclipse/iceoryx/pull/360#discussion_r535725529 in another PR. Shall I create an issue and move the discussion there?

elBoberido: moved to #408

Issue #261, opened 2020-08-26
Clean-up std::terminate usage


mossmaurice:
## Brief feature description

Iceoryx often uses `std::terminate()` instead of `errorHandler()`. With #130 a clear error handling will be available, afterwards all occurrences should be replaced by calls to the errorHandler.

## Detailed information

* [x] Calling `std::terminate` is only allowed in building blocks like the `errorHandler`
* [x] Replace `std::terminate` in `iceoryx_utils` with `cxx::Expects`, where dependencies shall be avoided to higher level modules
* [x] Replace `std::terminate` in `iceoryx_posh` with `cxx::Expects`

PR #417, opened 2020-12-08
Iox #91 Clean up todo's in RouDi classes

### LINKED ISSUE #91: Minor RouDi cleanups
## Brief feature description

* Leftovers from #70 and #78 

## Detailed information

* Deprecate old RouDi c'tors
* Take care of `@todo` in RouDi files

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Cleaned up dozens of `@todo`'s
* Removed deprecated methods

1. [x] Merge

## References

- Closes #91


iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.inl
```diff
@@ -30,9 +30,8 @@ MemPoolIntrospection<MemoryManager, SegmentManager, SenderPort>::MemPoolIntrospe
 {
     m_senderPort.activate(); // corresponds to offer
 
-    /// @todo create a wrapper function which takes care of the 16 character limitation of the thread name
-    // set thread name
-    pthread_setname_np(m_thread.native_handle(), "MemPoolIntr");
+    cxx::string<16> threadName{"MemPoolIntr"};
+    pthread_setname_np(m_thread.native_handle(), threadName.c_str());
```
elBoberido, line 34: This is not really what the `todo` suggested. Now we have a magic number distributed over several files. Could you maybe wrap the `pthread_setname_np` function in the utils?

iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_components.hpp
```diff
@@ -39,25 +39,11 @@ struct IceOryxRouDiComponents
...
-
-            m_rouDiMemoryManager.createAndAnnounceMemory().or_else([](RouDiMemoryManagerError error) {
-                LogFatal() << "Could not create SharedMemory! Error: " << static_cast<uint64_t>(error);
-                errorHandler(Error::kROUDI_COMPONENTS__SHARED_MEMORY_UNAVAILABLE, nullptr, iox::ErrorLevel::FATAL);
-            });
-            return true;
-        },
-        []() {}};
+    /// @brief Prepare the memory and clean up old ressources
+    IceOryxRouDiMemoryManager* initRouDiMemoryManager() noexcept;
```
elBoberido, line 43: Could you make this private?

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
```diff
@@ -37,6 +37,8 @@ enum class MonitoringMode
     OFF
 };
```
elBoberido, line 38: maybe the `MonitoringMode` should be moved to `iceoryx_posh_types.hpp`? It's also used in `roudi.cpp` which probably should not depend on this header

iceoryx_utils/CMakeLists.txt
```diff
@@ -197,6 +197,12 @@ set_target_properties(iceoryx_platform PROPERTIES
 
 target_compile_options(iceoryx_platform PRIVATE ${ICEORYX_WARNINGS})
 
+target_link_libraries(iceoryx_platform 
+    PUBLIC 
+    rt 
+    pthread
+)
```
elBoberido, line 204: after some thinking, I still think `rt` and `pthread` belongs to `iceoryx_platform` but it must only be added when the target is Linux. Maybe we need a separate `CMakeLists.txt` for each `platform/[linux,qnx,...]` folder

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
```diff
@@ -61,6 +63,8 @@ class CmdLineParser
                        char* argv[],
                        const CmdLineArgumentParsingMode cmdLineParsingMode = CmdLineArgumentParsingMode::ALL) noexcept;
 
+    void printParams() noexcept;
```
elfenpiff, line 66: `printParams` sounds weird. I think `printParameters` sounds better.

iceoryx_posh/include/iceoryx_posh/version/compatibility_check_level.hpp
```diff
@@ -28,6 +30,32 @@ enum class CompatibilityCheckLevel
     BUILD_DATE
 };
 
+inline iox::log::LogStream& operator<<(iox::log::LogStream& logstream, const version::CompatibilityCheckLevel& level)
```
elfenpiff, line 33: This streaming operator is not unit tested according to codecov. Could you please add some unit tests.

iceoryx_posh/source/roudi/application/roudi_app.cpp
```diff
@@ -67,37 +65,28 @@ void RouDiApp::registerSigHandler() noexcept
     sigemptyset(&act.sa_mask);
     act.sa_handler = roudiSigHandler;
     act.sa_flags = 0;
-    if (-1 == sigaction(SIGINT, &act, NULL))
+    if (cxx::makeSmartC(sigaction, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, SIGINT, &act, nullptr)
```
elfenpiff, line 68: I think we need the `and_then`/`or_else` pattern for smart_c. I created an issue for that: https://github.com/eclipse/iceoryx/issues/418 and this could become really clean.

iceoryx_posh/source/roudi/application/roudi_app.cpp
```diff
@@ -67,37 +65,28 @@ void RouDiApp::registerSigHandler() noexcept
...
+            .hasErrors())
     {
-        LogError() << "Calling sigaction() failed";
-        std::terminate();
+        LogFatal() << "Calling sigaction() failed";
+        errorHandler(Error::kROUDI_APP__COULD_NOT_REGISTER_SIGNALS, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 72: Please add a return after the `errorHandler`. Otherwise it is possible that the execution continues when the errorHandler does not terminate when the user overrides the default behavior.

iceoryx_posh/source/roudi/application/roudi_app.cpp
```diff
@@ -67,37 +65,28 @@ void RouDiApp::registerSigHandler() noexcept
...
     }
 
-    if (-1 == sigaction(SIGTERM, &act, NULL))
+    if (cxx::makeSmartC(sigaction, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, SIGTERM, &act, nullptr)
+            .hasErrors())
     {
-        LogError() << "Calling sigaction() failed";
-        std::terminate();
+        LogFatal() << "Calling sigaction() failed";
+        errorHandler(Error::kROUDI_APP__COULD_NOT_REGISTER_SIGNALS, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 79: Please add a return after the `errorHandler`.

iceoryx_posh/source/roudi/application/roudi_app.cpp
```diff
@@ -67,37 +65,28 @@ void RouDiApp::registerSigHandler() noexcept
...
     }
 
-    if (-1 == sigaction(SIGHUP, &act, NULL))
+    if (cxx::makeSmartC(sigaction, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, SIGHUP, &act, nullptr)
+            .hasErrors())
     {
-        LogError() << "Calling sigaction() failed";
-        std::terminate();
+        LogFatal() << "Calling sigaction() failed";
+        errorHandler(Error::kROUDI_APP__COULD_NOT_REGISTER_SIGNALS, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 86: Please add a return after the `errorHandler`.

iceoryx_posh/source/roudi/iceoryx_roudi_components.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+IceOryxRouDiMemoryManager* IceOryxRouDiComponents::initRouDiMemoryManager() noexcept
+{
+    // this temporary object will create a roudi mqueue and close it immediatelly
+    // if there was an outdated roudi message queue, it will be cleaned up
+    // if there is an outdated mqueue, the start of the apps will be terminated
+    runtime::MqBase::cleanupOutdatedMessageQueue(MQ_ROUDI_NAME);
+
+    m_rouDiMemoryManager.createAndAnnounceMemory().or_else([](RouDiMemoryManagerError error) {
+        LogFatal() << "Could not create SharedMemory! Error: " << static_cast<uint64_t>(error);
```
elfenpiff, line 29: Providing the user with some integer does not seem so helpful. Could you please implement a streaming operator or another conversion to translate `RouDiMemoryManagerError` into a human readable string.

iceoryx_posh/source/roudi/memory/posix_shm_memory_provider.cpp
```diff
@@ -78,7 +76,7 @@ cxx::expected<void*, MemoryProviderError> PosixShmMemoryProvider::createMemory(c
             .hasErrors())
     {
         LogFatal() << "Could not set signal handler for SIGBUS!";
-        std::terminate();
+        errorHandler(Error::kROUDI_MEMORY__COULD_NOT_REGISTER_SIGBUS, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 79: Please add a return after the `errorHandler`.

iceoryx_posh/source/roudi/memory/posix_shm_memory_provider.cpp
```diff
@@ -89,7 +87,7 @@ cxx::expected<void*, MemoryProviderError> PosixShmMemoryProvider::createMemory(c
             .hasErrors())
     {
         LogFatal() << "Could not reset signal handler for SIGBUS!";
-        std::terminate();
+        errorHandler(Error::kROUDI_MEMORY__COULD_NOT_UNREGISTER_SIGBUS, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 90: Please add a return after the `errorHandler`.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -238,5 +238,29 @@ units::Duration CmdLineParser::getProcessKillDelay() const noexcept
     return m_processKillDelay;
 }
 
+iox::log::LogStream& operator<<(iox::log::LogStream& logstream, const MonitoringMode& mode)
```
elfenpiff, line 241: According to codecov this is not tested. Could you please write a little test for this function.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -238,5 +238,29 @@ units::Duration CmdLineParser::getProcessKillDelay() const noexcept
...
+{
+    switch (mode)
+    {
+    case MonitoringMode::OFF:
+        logstream << "OFF";
```
elfenpiff, line 246: I think this is not expressive enough. `ON` and `OFF` could be anything and the user has to rely on the future developer to provide him some context in a log message what on and off means. 

Could you please instead translate it into: `MonitoringMode::ON` and `MonitoringMode::OFF`

iceoryx_posh/include/iceoryx_posh/version/compatibility_check_level.hpp
```diff
@@ -28,6 +30,32 @@ enum class CompatibilityCheckLevel
...
+{
+    switch (level)
+    {
+    case CompatibilityCheckLevel::OFF:
+        logstream << "OFF";
```
elfenpiff, line 38: The translation is exactly the same then the MonitoringMode translation. This could lead to confusion since the developer maybe assumes that the translation is detailed enough and does not provide some context which OFF is meant.

Could you please therefore translate this to `CompatibilityCheckLevel::OFF` and the values below with the same logic (EnumName::VALUE)

elfenpiff (changes requested):
- Add UnitTests for enum translation
- Adjust translation for enums and add missing enum translation instead of printing just numbers
- Either revert the `pthread_setname_np` changes or write a posix threads wrapper which solves the issue cleanly.

iceoryx_utils/CMakeLists.txt
dkroenke: Yes the linker errors in windows and Mac shows that it needs only to be added for linux. But i do not see the benefit of having more cmake files in comparison to have some platform checks in  utils cmake.

iceoryx_posh/include/iceoryx_posh/version/compatibility_check_level.hpp
elBoberido: oh, that's will be interesting. I guess you have to look at `test_logstream` ... maybe the `Logger_Mock` should be moved to the testutils

iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_components.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
mossmaurice: Only `roudi_app.hpp` depends on `roudi_cmd_line_parser.hpp`. If I'd move it to `iceoryx_posh_types.hpp` we would have a free function in there. I'm happy to move it to the class `RouDiApp`. Opinions?

iceoryx_posh/include/iceoryx_posh/version/compatibility_check_level.hpp
mossmaurice: Fixed.

iceoryx_posh/source/roudi/application/roudi_app.cpp
mossmaurice: I don't get that. How will a `return` help here if `errorHandler()` does not return?

iceoryx_posh/source/roudi/application/roudi_app.cpp
mossmaurice: done

iceoryx_posh/source/roudi/application/roudi_app.cpp
mossmaurice: done

iceoryx_posh/source/roudi/iceoryx_roudi_components.cpp
mossmaurice: Done

iceoryx_posh/source/roudi/memory/posix_shm_memory_provider.cpp
mossmaurice: Done

iceoryx_posh/source/roudi/memory/posix_shm_memory_provider.cpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.inl
mossmaurice: I chose to implement it in a free function as the first step. Once we create our own pthread class, this will become a member method.

iceoryx_utils/include/iceoryx_utils/posix_wrapper/pthread.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace iox
+{
+namespace posix
+{
+enum class PThreadErrorType : int32_t
+{
+    EXCEEDED_RANGE_LIMIT = 0
+};
+
+cxx::expected<PThreadErrorType> setThreadName(pthread_t thread, const char* name);
```
elfenpiff, line 32: Since this is a C++ wrapper you should create an alias like `using threadName_t = cxx::string<16>` and use this type instead of `const char*`. Like `cxx::expected<PThreadErrorType> setThreadName(pthread_t thread, const threadName_t & name);`

iceoryx_utils/include/iceoryx_utils/posix_wrapper/pthread.hpp
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020 by Apex.AI Inc. All rights reserved.
```
elfenpiff, line 1: I would call this class/header just thread instead of pthread. In this case p = posix and we call the semaphore also semaphore and not psemaphore. It is clear from the namespace that this is a `posix::thread`.

iceoryx_utils/include/iceoryx_utils/posix_wrapper/pthread.hpp
```diff
@@ -0,0 +1,37 @@
...
+//
```
elfenpiff, line 2: Since you already created a file wouldn't it be better to go all the way and create a full posix thread wrapper. You do not have to go all the way with all the fancy features but a constructor which creates it then a destructors which cleans it up and some methods like `setName`, `getName` which then do not need the thread handle anymore.

All the files are already there and the code you have to add should be not too much and would fit perfectly into this cleanup story. Otherwise we have a cleanup story where we introduce again new technical debt.

elfenpiff (changes requested):

iceoryx_posh/include/iceoryx_posh/version/compatibility_check_level.hpp
elfenpiff: @mossmaurice either you forgot to commit your fix or my github page is showing me for some reasons the old state :/

mossmaurice: It's showing the **Outdated** version as stated here in the first line ;-)

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
mossmaurice: Fixed.

iceoryx_utils/include/iceoryx_utils/posix_wrapper/pthread.hpp
mossmaurice: I had exactly the same thought, a moment after pushing :smile: Done

iceoryx_utils/include/iceoryx_utils/posix_wrapper/pthread.hpp
mossmaurice: If we are sure, that we will replace all `std::thread` occurrences  with our own implementation, I'd agree to do that now. TBH I don't see this coming :-) `std::thread` is just fine and there are various safety certified versions from vendors available AFAIK. I'll leave this as it is.

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
mossmaurice: Move it to `iceoryx_posh_types.hpp` and started preparing for the split of a separate `iceoryx_roudi_types.hpp` once RouDi is in its own CMake target.

iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_app.hpp
```diff
@@ -29,7 +29,8 @@ class IceOryxRouDiApp : public RouDiApp
     IceOryxRouDiApp(const config::CmdLineParser& cmdLineParser, const RouDiConfig_t& roudiConfig) noexcept;
 
     /// @brief starts the execution of the RouDi daemon
-    void run() noexcept override;
+    /// @return Return code for programm execution
+    uint8_t run() noexcept override;
```
FerdinandSpitzschnueffler, line 33: please update the copyright header

iceoryx_dds/include/iceoryx_dds/dds/dds_config.hpp
```diff
@@ -21,6 +21,7 @@ namespace iox
 {
 namespace dds
 {
+using namespace units::duration_literals;
```
elBoberido, line 24: do we want this in headers?

iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_app.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -192,15 +177,61 @@ struct DefaultChunkQueueConfig
...
 
+/// @todo Move everything in this namespace to iceoryx_roudi_types.hpp once we move RouDi to a separate CMake target
+namespace roudi
+{
+using ConfigFilePathString_t = cxx::string<1024>;
+
+constexpr char MQ_ROUDI_NAME[] = "/roudi";
+
+/// shared memmory segment for the iceoryx managment data
+constexpr char SHM_NAME[] = "/iceoryx_mgmt";
```
elBoberido, line 191: this is not purely used by roudi but also by the runtime. Not sure if it should be in this namespace if all of this will be move to a separate cmake target at one day.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -192,15 +177,61 @@ struct DefaultChunkQueueConfig
...
+
+// Timeout
+using namespace units::duration_literals;
+constexpr units::Duration PROCESS_DEFAULT_KILL_DELAY = 45_s;
+constexpr units::Duration PROCESS_TERMINATED_CHECK_INTERVAL = 250_ms;
+constexpr units::Duration PROCESS_WAITING_FOR_ROUDI_TIMEOUT = 60_s;
```
elBoberido, line 197: this is not used by roudi but only be the runtime during the application startup if roudi was not found

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -192,15 +177,61 @@ struct DefaultChunkQueueConfig
...
+constexpr units::Duration DISCOVERY_INTERVAL = 100_ms;
```
elBoberido, line 198: the `PROCESS_KEEP_ALIVE_INTERVAL` depends on this, so the same as `MQ_ROUDI_NAME` and `SHM_NAME` applies to this

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -192,15 +177,61 @@ struct DefaultChunkQueueConfig
...
+    switch (mode)
+    {
+    case MonitoringMode::OFF:
+        logstream << "MonitoringMode::OFF";
+        break;
+    case MonitoringMode::ON:
+        logstream << "MonitoringMode::ON";
+        break;
+    }
+    return logstream;
```
elBoberido, line 223: I guess we need an `*.inl`

iceoryx_utils/include/iceoryx_utils/posix_wrapper/thread.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace iox
+{
+namespace posix
+{
+enum class ThreadErrorType : int32_t
+{
+    EXCEEDED_RANGE_LIMIT = 0
+};
+
+using ThreadName_t = cxx::string<16>;
```
elBoberido, line 31: the name for pthread_setname_np is restricted to 16 characters, including the `\0` so the cxx::string capacity must be `15` in order to have space for the termination

iceoryx_utils/source/posix_wrapper/thread.cpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+    else
+    {
+        return cxx::success<>();
+    }
+}
+
+cxx::expected<ThreadName_t, ThreadErrorType> getThreadName(pthread_t thread)
+{
+    constexpr uint8_t maxCharCountThreadName = 17;
```
elBoberido, line 36: max is 16, not 17 ;)

iceoryx_utils/test/moduletests/test_posix_thread.cpp
```diff
@@ -0,0 +1,83 @@
...
+/// @todo Renable this test, once "does not compile" tests are possible
+#if 0
+    constexpr char stringLongerThan16Chars[] =
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor "
+        "incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud "
+        "exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.";
+
+    auto setResult = setThreadName(m_thread->native_handle(), stringLongerThan16Chars);
+#endif
+}
```
elBoberido, line 71: please create a test with a name of `ThreadName_t::capacity.

Additionally, a test with `ThreadName_t::capacity + 1` could be created, were `pthread_setname_np` is called manually and `pthread_getname_np` returns a truncated string.

iceoryx_utils/test/moduletests/test_posix_thread.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    ~Thread_test()
+    {
+    }
+
+    void threadFunc()
+    {
+        while (m_run)
+        {
+        }
```
elBoberido, line 53: this will rise the sea level by a few nm. Please either add a sleep or use a semaphore.

iceoryx_dds/include/iceoryx_dds/dds/dds_config.hpp
mossmaurice: Nope, already this discuss with @elfenpiff we want to avoid namespace pollution and keep ADL time low. `inline namespace` does not prevent this.
Steps in follow-up:
1.  Create c'tor for all numeric literal operators e.g. `Duration::Seconds()`
1. Remove `using namespace` from headers

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
mossmaurice: For shared interfaces the common rules goes like this:

> The one defining the interfaces owns the interface. 

In this case this is RouDi.

iceoryx_utils/source/posix_wrapper/thread.cpp
mossmaurice: Good catch ;) However, choosing 15 will result in an `ERANGE` error ;)

iceoryx_posh/include/iceoryx_posh/version/compatibility_check_level.hpp
mossmaurice: Done

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
mossmaurice: Just realised there is not `CmdLineParser` test so far. I remember @budrus didn't want to have issues created for these kind of findings. I can increase the test coverage of RouDi in a follow-up. If you look [here](https://codecov.io/gh/eclipse/iceoryx?branch=master), that's our main weak spot.

@budrus Ok from your side to create an issue on this? Happy to work on it after the 0.9 release

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elBoberido: sure, but how do you solve the problem that the runtime depends on roudi once there is a separate roudi target?

iceoryx_utils/test/moduletests/test_posix_thread.cpp
mossmaurice: Added the first test case.

The latter won't compile. Anyway, it's already contained in the equivalence class of `LargeStringIsTruncated`.

iceoryx_utils/source/posix_wrapper/thread.cpp
mossmaurice: Weird, worked now :confused: Down from 15, I'm getting `ERANGE`. This is the expected behavior. Adapted it.

iceoryx_utils/test/moduletests/test_posix_thread.cpp
mossmaurice: Penny saved is a penny earned. Goes for the climate as well :national_park:

iceoryx_utils/include/iceoryx_utils/posix_wrapper/thread.hpp
mossmaurice: Initially thought `string` does count the null termination as well. Thanks for clearing things up!

iceoryx_utils/test/moduletests/test_posix_thread.cpp
elBoberido: you misunderstood me. You have to call `pthread_setname_np` and `pthread_getname_np` manually in the test, not by the abstraction you introduced. There you can create a char array of `ThreadName_t::capacity + 1`. With this, we can check if our assumption of having the maximum possible `ThreadName_t` is correct

elBoberido: I could live without this test. Just wanted to make sure we are on the same page.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
mossmaurice: Added

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
mossmaurice: Moved `PROCESS_WAITING_FOR_ROUDI_TIMEOUT` to namespace `runtime`. Also, I found the first circular dependency between RouDi and the runtime. roudi::DISCOVERY_INTERVAL -> runtime::PROCESS_KEEP_ALIVE_INTERVAL -> roudi::PROCESS_KEEP_ALIVE_TIMEOUT :rofl: We really need to untangle RouDi and the runtime and separate them clearly in CMake targets. Just wondering if this will be possible in v1.x without breaking the user API. @dkroenke Do we have a list how the folks integrating iceoryx are build it CMake-wise? Is it possible to introduce a separate `iceory_roudi` CMake package in v1.x?

iceoryx_posh/include/iceoryx_posh/gateway/gateway_generic.hpp
```diff
@@ -34,6 +35,8 @@ namespace iox
 {
 namespace gw
 {
+using namespace iox::units::duration_literals;
```
elfenpiff, line 38: I thought we would like to get rid of `using namespace` in headers?

iceoryx_posh/include/iceoryx_posh/roudi/roudi_app.hpp
```diff
@@ -78,19 +70,18 @@ class RouDiApp
...
-                                                 .or_else([](posix::SemaphoreError&) {
-                                                     std::cerr << "Unable to create the semaphore for RouDi"
-                                                               << std::endl;
-                                                     std::terminate();
-                                                 })
-                                                 .value());
+    posix::Semaphore m_semaphore =
+        std::move(posix::Semaphore::create(posix::CreateUnnamedSingleProcessSemaphore, 0u)
+                      .or_else([](posix::SemaphoreError&) {
+                          errorHandler(Error::kROUDI_APP__FAILED_TO_CREATE_SEMAPHORE, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 80: After calling the error handler we have to explicitly terminate. If we don't do this and someone configures the error handling so that they ignore this we run completely into the undefined behavior land. 

This class is unusable when the semaphore acquire fails. The usual way to handle this kind of problem would be the `Creation` design pattern ... or terminate.

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
```diff
@@ -61,20 +49,22 @@ class CmdLineParser
                        char* argv[],
                        const CmdLineArgumentParsingMode cmdLineParsingMode = CmdLineArgumentParsingMode::ALL) noexcept;
 
+    void printParameters() noexcept;
```
elfenpiff, line 52: I think this method should be const

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
dkroenke: Yes we have an instruction here: [https://github.com/eclipse/iceoryx/blob/master/doc/installation-guide.md#build-with-cmake](https://github.com/eclipse/iceoryx/blob/master/doc/installation-guide.md#build-with-cmake) and a separate library for roudi called "iceoryx_posh_roudi" here: [here](https://github.com/eclipse/iceoryx/blob/3c43d88fb385747bdb00c5c48e6cca89053df8a3/iceoryx_posh/CMakeLists.txt#L201) Also the iox-roudi executable is defined in this file.

We can think of to move the roudi lib into an own cmake file.

iceoryx_posh/source/roudi/iceoryx_roudi_components.cpp
```diff
@@ -0,0 +1,36 @@
...
+IceOryxRouDiMemoryManager* IceOryxRouDiComponents::initRouDiMemoryManager() noexcept
+{
+    // this temporary object will create a roudi mqueue and close it immediatelly
+    // if there was an outdated roudi message queue, it will be cleaned up
+    // if there is an outdated mqueue, the start of the apps will be terminated
+    runtime::MqBase::cleanupOutdatedMessageQueue(roudi::MQ_ROUDI_NAME);
+
+    m_rouDiMemoryManager.createAndAnnounceMemory().or_else([](RouDiMemoryManagerError error) {
+        LogFatal() << "Could not create SharedMemory! Error: " << error;
+        errorHandler(Error::kROUDI_COMPONENTS__SHARED_MEMORY_UNAVAILABLE, nullptr, iox::ErrorLevel::FATAL);
```
elfenpiff, line 30: The program continues here and then we return the pointer to an invalid object? Is this code testable?

I would prefer here to maybe return an `expected<IceOryxRouDiMemoryManager*, ErrorType>` to avoid this issue.

Or we could define the error level fatal as non overloadable and that this always has to terminate.

iceoryx_posh/source/roudi_environment/roudi_environment.cpp
```diff
@@ -32,7 +32,7 @@ RouDiEnvironment::RouDiEnvironment(BaseCTor, const uint16_t uniqueRouDiId)
 }
 
 RouDiEnvironment::RouDiEnvironment(const RouDiConfig_t& roudiConfig,
-                                   config::MonitoringMode monitoringMode,
+                                   roudi::MonitoringMode monitoringMode,
```
elfenpiff, line 35: Please add const

iceoryx_posh/test/moduletests/test_compatability_check_level.cpp
```diff
@@ -0,0 +1,121 @@
...
+    {
+        m_loggerMock.m_logs.clear();
+    }
+
+    Logger_Mock m_loggerMock;
+};
+
+TEST_F(CompatibilityCheckLevel_test, OffLeadsToCorrectString)
+{
+    auto sut = CompatibilityCheckLevel::OFF;
```
elfenpiff, line 41: sadly we also can do stuff like:
```cpp
logstream << static_cast<CompatibilityCheckLevel>(18237812);
```

And this should always be translated into something like `CompatibilityCheckLevel::UNDEFINED` or something similar. I think we have to care of this in every enum to string translation function that after the switch case always the undefined string is returned.

iceoryx_utils/source/posix_wrapper/thread.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace iox
+{
+namespace posix
+{
+cxx::expected<ThreadErrorType> setThreadName(pthread_t thread, const ThreadName_t& name)
+{
+    if (cxx::makeSmartC(pthread_setname_np, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, thread, name.c_str())
+            .hasErrors())
+    {
+        return cxx::error<ThreadErrorType>(ThreadErrorType::EXCEEDED_RANGE_LIMIT);
```
elfenpiff, line 26: This error can never occur since `ThreadName_t` has the perfect size and the user is unable to provide here a string which is larger than 15 characters.

According to the posix manual this function fails only when the name is too long. Since this can never happen we can return `void` here and remove the `ThreadErrorType::EXCEEDED_RANGE_LIMIT`.
Otherwise we have produced code here which is untestable!

elfenpiff (changes requested):

iceoryx_posh/include/iceoryx_posh/gateway/gateway_generic.hpp
mossmaurice: Yes, but not in this pull request. We discussed it and said there will be another one, remember? ;-) To not let it slip through, we'll move the issue right into our sprint.

iceoryx_posh/include/iceoryx_posh/roudi/roudi_app.hpp
mossmaurice: Hmm, that seems like an unreasonable exception to me :smiley: If the user is configuring the error handling wrong, than it is not our fault. We need to have a single place where we can catch all the errors, this being the error handler. IIRC we wanted to remove `std::terminate` occurrences except for the classes that need to run independent from `iceoryx_utils` (see #261) e.g. `cxx::vector`

@MatthiasKillat Can you confirm? Did not get a grasp on your error handling document yet.

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
mossmaurice: Done

iceoryx_utils/test/moduletests/test_posix_thread.cpp
```diff
@@ -0,0 +1,96 @@
...
+    constexpr char stringLongerThanThreadNameCapacitiy[] =
+        "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor "
+        "incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud "
+        "exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.";
+
+    auto setResult = setThreadName(m_thread->native_handle(), stringLongerThanThreadNameCapacitiy);
+#endif
+}
+
+TEST_F(Thread_test, ThreadNameCapacityIsNotTruncated)
```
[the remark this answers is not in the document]
mossmaurice: Renamed the test cases

iceoryx_utils/source/posix_wrapper/thread.cpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+    else
+    {
+        return cxx::success<>();
+    }
+}
+
+cxx::expected<ThreadName_t, ThreadErrorType> getThreadName(pthread_t thread)
+{
+    constexpr uint8_t maxCharCountThreadName = 16;
```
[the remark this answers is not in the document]
mossmaurice: Valid point. Done.

iceoryx_utils/source/posix_wrapper/thread.cpp
```diff
@@ -0,0 +1,56 @@
...
+    char tempName[maxCharCountThreadName];
+    if (cxx::makeSmartC(pthread_getname_np,
```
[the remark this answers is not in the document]
mossmaurice: Well, MISRA and other coding rules state to be as explicit as possible and one should always take care of all possible branches, which I am doing here. Imagine on an obscure POSIX platform the limit is set to 10. What happens then? We just blindly trusted QNX that things will be fine? I guess instead of removing the error check, we will rather add a coverage exception here with this explanation. Want me do add a few lines here? Or we will do this mid-2021 when starting with certification?

iceoryx_utils/source/posix_wrapper/thread.cpp
mossmaurice: See https://github.com/eclipse/iceoryx/pull/417#discussion_r542568697

iceoryx_posh/test/moduletests/test_roudi_memory_manager.cpp
```diff
@@ -108,4 +108,9 @@ TEST_F(RouDiMemoryManager_Test, AddMemoryProviderExceedsCapacity)
     ASSERT_THAT(expectError.has_error(), Eq(true));
     EXPECT_THAT(expectError.get_error(), Eq(RouDiMemoryManagerError::MEMORY_PROVIDER_EXHAUSTED));
 }
+
+TEST_F(RouDiMemoryManager_Test, StreamOperatorTranslationIsCorrect)
```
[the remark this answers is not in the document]
mossmaurice: I'd ask you to keep it here. I admit it's not the best style. However, we've done it before. For my part the more reminders there are the better.

iceoryx_posh/test/moduletests/test_compatability_check_level.cpp
mossmaurice: Thanks for spotting this! Added a default branch to the `MonitoringMode` streaming operator as well.

iceoryx_posh/source/roudi_environment/roudi_environment.cpp
mossmaurice: Added

iceoryx_posh/source/roudi/iceoryx_roudi_components.cpp
mossmaurice: If someone configures the errorHandler to be continuing after a fatal error, I can't really help. **Fatal** is clear in the wording, isn't it? I've implemented the solution you suggested for the sake of a cleaner API. However, on the calling side no further error handling is done, just calling `initRouDiMemoryManager().value()`.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elBoberido: short term, you can create a constexpr in runtime and assign that to `roudi::DISCOVERY_INTERVAL` and `runtime::PROCESS_KEEP_ALIVE_INTERVAL` ... you just have to find a good name for it

iceoryx_utils/source/posix_wrapper/thread.cpp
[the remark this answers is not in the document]
elBoberido: I'd rather have a static cxx::string::capacity function and use `ThreadName_t::capacity` instead of introducing new constants ... but that's a different discussion

iceoryx_posh/source/roudi/iceoryx_roudi_components.cpp
MatthiasKillat: In our error handling workshop (and thus in error-handling.md) we defined that a fatal error **always leads to a terminate** in the error handler. It may be preceded with additional code execution, but ultimatetely it has to call a terminate. While this is not enforced by the code right now, when someone reconfigures the error handler to not call terminate he violates this assumption.
This reconfiguration must be possible but should only be used  for testing purpose. A fully configurable error handler is another issue and we need to define a proper user interface for this.

An additional terminate call is therefore redundant and would also be inconsistent to all other places were a fatal error occurs. In theory the error handler should be internal now i.e. not be part of the public API and configurable by the user.

I would postpone this to the point were we refine the error handler and the error handling strategy.

From my point of view it helps to think of returning an expected with an error type as some kind of throw and calling the error handler is analogous to a catch clause which can call a restricted subset of exception handlers (i.e. the error handler with different levels). The decision who handles the error is similarly difficult as for exception (i.e. propagate vs. handle). It must be handled at the latest when it is clear that further operation is not possible, which is the case here.

iceoryx_posh/include/iceoryx_posh/roudi/roudi_app.hpp
MatthiasKillat: The same as in (EH) applies here in my opinion. An error is propagated and at this point we decide that without a semaphore we cannot continue and hence handle this error as a fatal error. An additional terminate would be superfluous. 
Writing a better error handler belongs to a different issue.

iceoryx_utils/source/posix_wrapper/thread.cpp
```diff
@@ -33,10 +33,10 @@ cxx::expected<ThreadErrorType> setThreadName(pthread_t thread, const threadName_
 
 cxx::expected<ThreadErrorType> getThreadName(pthread_t thread, threadName_t& name)
 {
-    constexpr uint64_t size = name.capacity();
-    char tempName[size];
+    constexpr uint8_t maxCharCountThreadName = 17;
```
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_utils/source/posix_wrapper/shared_memory_object/memory_map.cpp
```diff
@@ -47,7 +47,7 @@ MemoryMap::MemoryMap(const void* f_baseAddressHint,
                      const off_t f_offset)
     : m_length(f_length)
 {
-    int32_t l_memoryProtection;
+    int32_t l_memoryProtection{0};
```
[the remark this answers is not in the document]
mossmaurice: @dkroenke Can you answer this question?

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
mossmaurice: We should do that in a follow-up

iceoryx_utils/test/moduletests/test_posix_thread.cpp
mossmaurice: Understood. As for own main platforms the limitations are clear, I won't implement this test case.

iceoryx_utils/source/posix_wrapper/thread.cpp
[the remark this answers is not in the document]
mossmaurice: It is basically a trade-off between "Cover all possible branches, even if not reachable" and "Do write dead/unreachable code"

After a longer discussion with @elfenpiff we decided to remove the unreachable branches and add an additional `cxx::Ensures()`.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
mossmaurice: Ok, we need to consider this when doing the split.

elBoberido (dismissed):

iceoryx_utils/source/posix_wrapper/shared_memory_object/memory_map.cpp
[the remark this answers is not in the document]
dkroenke: yes PROT_NONE is correct here

iceoryx_utils/test/moduletests/test_posix_thread.cpp
```diff
@@ -0,0 +1,96 @@
...
+        {
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        }
+    };
+
+    std::atomic_bool m_run{true};
+    std::thread* m_thread;
+};
+
+TEST_F(Thread_test, DISABLED_LargeStringIsTruncated)
```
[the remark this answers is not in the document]
mossmaurice: Removed

iceoryx_posh/test/moduletests/test_roudi_memory_manager.cpp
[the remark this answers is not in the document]
mossmaurice: Replaced with an `/// @todo`

iceoryx_utils/source/posix_wrapper/shared_memory_object/memory_map.cpp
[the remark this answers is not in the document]
mossmaurice: Ok, will adapt this in the next pull request.

dkroenke (approved):

iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_components.hpp
```diff
@@ -39,25 +44,12 @@ struct IceOryxRouDiComponents
...
-                LogFatal() << "Could not create SharedMemory! Error: " << static_cast<uint64_t>(error);
-                errorHandler(Error::kROUDI_COMPONENTS__SHARED_MEMORY_UNAVAILABLE, nullptr, iox::ErrorLevel::FATAL);
-            });
-            return true;
-        },
-        []() {}};
-
     /// @brief Handles the ports in shared memory
-    PortManager m_portManager{&m_rouDiMemoryManager};
+    PortManager m_portManager{initRouDiMemoryManager().value()};
```
elfenpiff, line 48: If `initRouDiMemoryManager` returns an error we run here directly into a segmentation fault since the expected is dereferencing a nullptr here.

iceoryx_posh/test/moduletests/test_roudi_memory_manager.cpp
```diff
@@ -108,4 +108,7 @@ TEST_F(RouDiMemoryManager_Test, AddMemoryProviderExceedsCapacity)
     ASSERT_THAT(expectError.has_error(), Eq(true));
     EXPECT_THAT(expectError.get_error(), Eq(RouDiMemoryManagerError::MEMORY_PROVIDER_EXHAUSTED));
 }
+
+/// @todo Add TEST_F(RouDiMemoryManager_Test, StreamOperatorTranslationIsCorrect)
```
elfenpiff, line 112: Could you please add this todo into an issue as an open point?

iceoryx_utils/include/iceoryx_utils/posix_wrapper/thread.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "iceoryx_utils/cxx/expected.hpp"
+#include "iceoryx_utils/cxx/smart_c.hpp"
+#include "iceoryx_utils/cxx/string.hpp"
+#include "iceoryx_utils/platform/pthread.hpp"
+
+namespace iox
+{
+namespace posix
+{
+enum class ThreadErrorType : int32_t
```
elfenpiff, line 26: This `ThreadErrorType` is no where used and could be removed.

iceoryx_utils/test/moduletests/test_posix_thread.cpp
```diff
@@ -0,0 +1,92 @@
...
+    std::atomic_bool m_run{true};
+    std::thread* m_thread;
+};
+
+#if !defined(__APPLE__)
+TEST_F(Thread_test, SetAndGetWithEmptyThreadNameIsWorking)
+{
+    ThreadName_t emptyString = "";
+
+    setThreadName(m_thread->native_handle(), emptyString);
```
elfenpiff, line 66: This test will also never work on windows since windows does not have posix threads. I would recommend here to remove the `#if !defined(__APPLE__)` and use `pthread_create` directly. This is our approach to make platform code platform independent, with the use of the platform header files.

iceoryx_utils/test/moduletests/test_posix_thread.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+  public:
+    Thread_test()
+    {
+    }
+
+    void SetUp()
+    {
+        m_run = true;
+        m_thread = new std::thread(&Thread_test::threadFunc, this);
```
elfenpiff, line 35: Please use `pthread_create` instead of `std::thread~

iceoryx_utils/include/iceoryx_utils/posix_wrapper/thread.hpp
mossmaurice: Done

iceoryx_posh/test/moduletests/test_roudi_memory_manager.cpp
mossmaurice: Done: https://github.com/eclipse/iceoryx/issues/91#issuecomment-746314090

iceoryx_utils/test/moduletests/test_posix_thread.cpp
mossmaurice: Ok understood, `native_handle_type` is not portable.

dkroenke (approved):

iceoryx_utils/test/moduletests/test_posix_thread.cpp
mossmaurice: Ok, didn't manage to port to the native `pthread_create` interface in less than 1h. It compiles but getting a segfault, when setting the name. Please do this in the thread wrapper story.

iceoryx_utils/test/moduletests/test_posix_thread.cpp
mossmaurice: See https://github.com/eclipse/iceoryx/pull/417#discussion_r544366501

iceoryx_posh/include/iceoryx_posh/roudi/iceoryx_roudi_components.hpp
mossmaurice: Done. TIL directly invoking lamdas is possbile =)

dkroenke (approved):

elfenpiff (approved):

PR #434, opened 2020-12-15
Iox #431 iox::relative_ptr takes invalid state when constructed/assigned from iox::RelativePointer

### LINKED ISSUE #431: iox::relative_ptr takes invalid state when constructed/assigned from iox::RelativePointer
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.4.0

iox::relative_ptr takes invalid state when constructed/assigned from iox::RelativePointer.
This is because the `id` is initialized with `NULL` and the c'tor/assignment is calling `computeOffset()` directly. It can be fixed by calling the base c'tor/ operator `=` instead.

What do you expect to happen?

1. create iox::relative_ptr a with some value
1. assign ptr a to a iox::RelativePointer b
1. assign ptr b to a iox::relative_ptr c

marthtz:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
1. Fix 1: Base c'tor/ operator `=` is called instead of `computeOffset()`. This fixes #431.
1. Fix 2: `const void` was not supported as type. Removed the specialized template and using a SFINAE approach now. Added a test for this as well.
3. I don't like the naming. Why is the typed version of `iox::RelativePointer` called `iox::relative_ptr` ? I'd like to change it to `iox::TypedRelativePointer`. Thoughts?

1. [ ] Merge

## References

- Closes #431

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
@@ -234,13 +234,13 @@ class relative_ptr : public RelativePointer
 
 
     relative_ptr(const RelativePointer& other)
+        : RelativePointer(other)
     {
-        m_offset = computeOffset(other.computeRawPtr());
     }
 
     relative_ptr& operator=(const RelativePointer& other)
     {
-        m_offset = computeOffset(other.computeRawPtr());
+        RelativePointer::operator=(other);
 
         return *this;
     }
--- a/iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
@@ -253,34 +253,36 @@ class relative_ptr : public RelativePointer
         return *this;
     }
 
-    T& operator*()
+    template <typename SFINAE = T>
+    typename std::enable_if<!std::is_void<SFINAE>::value, SFINAE&>::type operator*()
     {
-        return *(static_cast<T*>(computeRawPtr()));
+        return *get();
     }
 
     T* operator->()
     {
-        return static_cast<T*>(computeRawPtr());
+        return get();
     }
 
-    const T& operator*() const
+    template <typename SFINAE = T>
+    typename std::enable_if<!std::is_void<SFINAE>::value, const SFINAE&>::type operator*() const
     {
-        return *(static_cast<T*>(computeRawPtr()));
+        return *get();
     }
 
     T* operator->() const
     {
-        return static_cast<T*>(computeRawPtr());
+        return get();
     }
 
     T* get() const
     {
-        return reinterpret_cast<T*>(RelativePointer::get());
+        return static_cast<T*>(computeRawPtr());
     }
 
     operator T*() const
     {
-        return reinterpret_cast<T*>(RelativePointer::get());
+        return get();
     }
 
     bool operator==(T* const ptr) const
@@ -289,60 +291,6 @@ class relative_ptr : public RelativePointer
     }
 
     bool operator!=(T* const ptr) const
-    {
-        return ptr != get();
-    }
-};
-
-template <>
-class relative_ptr<void> : public RelativePointer
-{
-  public:
-    relative_ptr(ptr_t ptr, id_t id)
-        : RelativePointer(ptr, id)
-    {
-    }
-
-    relative_ptr(ptr_t ptr = nullptr)
-        : RelativePointer(ptr)
-    {
-    }
-
-    relative_ptr(const RelativePointer& other)
-    {
-        m_offset = computeOffset(other.computeRawPtr());
-    }
-
-    relative_ptr& operator=(const RelativePointer& other)
-    {
-        m_offset = computeOffset(other.computeRawPtr());
-
-        return *this;
-    }
-
-    relative_ptr& operator=(ptr_t ptr)
-    {
-        m_id = searchId(ptr);
-        m_offset = computeOffset(ptr);
-        return *this;
-    }
-
-    void* get() const
-    {
-        return RelativePointer::get();
-    }
-
-    operator void*() const
-    {
-        return RelativePointer::get();
-    }
-
-    bool operator==(void* const ptr) const
-    {
-        return ptr == get();
-    }
-
-    bool operator!=(void* const ptr) const
     {
         return ptr != get();
     }
--- a/iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
@@ -240,7 +240,7 @@ class relative_ptr : public RelativePointer
 
     relative_ptr& operator=(const RelativePointer& other)
     {
-        RelativePointer::operator=(other);
+        *this = other;
 
         return *this;
     }
@@ -253,8 +253,8 @@ class relative_ptr : public RelativePointer
         return *this;
     }
 
-    template <typename SFINAE = T>
-    typename std::enable_if<!std::is_void<SFINAE>::value, SFINAE&>::type operator*()
+    template <typename U = T>
+    typename std::enable_if<!std::is_void<U>::value, U&>::type operator*()
     {
         return *get();
     }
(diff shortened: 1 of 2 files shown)


iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -234,13 +234,13 @@ class relative_ptr : public RelativePointer
...
     relative_ptr(const RelativePointer& other)
+        : RelativePointer(other)
     {
-        m_offset = computeOffset(other.computeRawPtr());
     }
 
     relative_ptr& operator=(const RelativePointer& other)
     {
-        m_offset = computeOffset(other.computeRawPtr());
+        RelativePointer::operator=(other);
```
elBoberido, line 243: doesn't `*this = other` also work?

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -253,34 +253,36 @@ class relative_ptr : public RelativePointer
         return *this;
     }
 
-    T& operator*()
+    template <typename SFINAE = T>
```
elBoberido, line 256: it's weird to have the template parameter call `SFINAE`. It's more idiomatic to call it `U`

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
marthtz: Yes, it does work as well. You prefer `*this`?

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
marthtz: You're probably right. Maybe add a comment regarding SFINAE and use `U`?

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
elBoberido: It's more idiomatic ... at least I've seen it way more than the explicit call of the assignment operator

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
elBoberido: I don't think a comment is necessary. The `std::enable_if` already states that SFINAE is used

budrus:
With respect to the naming you mention in the description. Maybe you dicuss with @MatthiasKillat. I guess it is called relative_ptr to be similar to the STL...

MatthiasKillat:
Yes it is in snake case for similarity reasons. Now as for the functionality I will take a look later. The relative_ptr concept is sound but the implementation could be cleaned up a bit and maybe improved (for efficiency and reducing some indirections) here and there.
I plan to do this when I get time, but it is mainly aesthetics.

I will take a look at this bug and PR later this day.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
MatthiasKillat: Yes it is more idiomatic. For operators I usually use ``rhs`` instead of ``other``, I wonder why I did not do so in the past.
You will also sometimes find ``that`` instead of ``other``, but this is just a matter of style and convention.

Historically it is a bit of prototype code and could be cleaned up, but there seemed to be no glaring bugs (that I know of).

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
MatthiasKillat: Sometimes I would like to have a more beautiful way to say "I want to have this function when a certain compile time condition is true". At the moment this is the best we can do, to my knowledge.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -253,34 +253,36 @@ class relative_ptr : public RelativePointer
...
     T* operator->() const
     {
-        return static_cast<T*>(computeRawPtr());
+        return get();
     }
 
     T* get() const
     {
-        return reinterpret_cast<T*>(RelativePointer::get());
+        return static_cast<T*>(computeRawPtr());
```
MatthiasKillat, line 280: I would use a reinterpret_cast here. Under certain conditions (involving multiple inheritance of the types and maybe others) a static_cast can actually change the address. reinterpret_cast will never do this.
I prefer reinterpret casts if we cast from void to some typed pointer.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
marthtz: Done.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
marthtz: Done.

marthtz:
> 
> 
> Yes it is in snake case for similarity reasons. Now as for the functionality I will take a look later. The relative_ptr concept is sound but the implementation could be cleaned up a bit and maybe improved (for efficiency and reducing some indirections) here and there.
> I plan to do this when I get time, but it is mainly aesthetics.
> 
> I will take a look at this bug and PR later this day.

The issue was that you can call `=` or the c'tor while `m_id` of the parameter is not set. The result was that `computeOffset()` and `getOffset()` return `NULL`. The way the c'tor and operator where written implies that you should get the same offset as the init parameter back when you call them. That's what got fixed and is tested for now.

It still don't like using `iox::RelativePointer` and `iox::relative_ptr`. I get the relation to STL, maybe we could call the untyped version just `iox::UntypedRelativePointer`?

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
marthtz: `computeRawPtr()` returns a `void*` already so performing a `static_cast` won't perform a change in address.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
schuellerstefan: > doesn't `*this = other` also work?

Of course this doesn't work (infinite recursion)

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -234,13 +234,13 @@ class relative_ptr : public RelativePointer
...
     relative_ptr(const RelativePointer& other)
+        : RelativePointer(other)
     {
-        m_offset = computeOffset(other.computeRawPtr());
     }
 
     relative_ptr& operator=(const RelativePointer& other)
     {
-        m_offset = computeOffset(other.computeRawPtr());
+        *this = other;
```
schuellerstefan, line 243: Where is the test that reveals that this is an infinite recursion now? @marthtz

elBoberido: yeah, it seems we are missing some tests ... `*static_cast<RelativePointer*>(this) = other;` should do the trick ... I guess I'm too dependent on my IDE to prevent me from doing stupid things :D

marthtz: Added a test and fixed it.
However, I'm using `RelativePointer::operator=(other);` as this is IMHO straightforward  and easier to comprehend than `static_cast<...`.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
marthtz: Fixed.

elBoberido (approved):

budrus (approved):

PR #441, opened 2020-12-16
Iox #408 add struct with options to publisher

### LINKED ISSUE #408: Refine modern pub/sub API

- take a sample
- something happens

**To-do:**

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
Introduce `PublisherOption` struct to pass options like history size to the publisher.

1. [ ] Merge

## References

- Partly closes #408

### DIFF
--- a/iceoryx_posh/source/roudi/port_manager.cpp
+++ b/iceoryx_posh/source/roudi/port_manager.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2019, 2020 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
 #include "iceoryx_posh/internal/roudi/port_manager.hpp"
 #include "iceoryx_posh/iceoryx_posh_types.hpp"
 #include "iceoryx_posh/internal/log/posh_logging.hpp"
+#include "iceoryx_posh/popo/publisher_options.hpp"
 #include "iceoryx_posh/roudi/introspection_types.hpp"
 #include "iceoryx_posh/runtime/node.hpp"
 #include "iceoryx_utils/cxx/vector.hpp"
@@ -595,8 +596,8 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
     }
 
     // we can create a new port
-    auto maybePublisherPortData = m_portPool->addPublisherPort(
-        service, historyCapacity, payloadMemoryManager, processName, portConfigInfo.memoryInfo);
+    auto options = popo::PublisherOptions{historyCapacity, portConfigInfo.memoryInfo};
+    auto maybePublisherPortData = m_portPool->addPublisherPort(service, payloadMemoryManager, processName, options);
     if (!maybePublisherPortData.has_error())
     {
         m_portIntrospection.addPublisher(maybePublisherPortData.value(), processName, service, node);
@@ -653,7 +654,8 @@ popo::ApplicationPortData* PortManager::acquireApplicationPortData(const Process
     }
 }
 
-void PortManager::addEntryToServiceRegistry(const capro::IdString_t& service, const capro::IdString_t& instance) noexcept
+void PortManager::addEntryToServiceRegistry(const capro::IdString_t& service,
+                                            const capro::IdString_t& instance) noexcept
 {
     m_serviceRegistry.add(service, instance);
     m_portPool->serviceRegistryChangeCounter()->fetch_add(1, std::memory_order_relaxed);
--- a/iceoryx_posh/source/roudi/port_manager.cpp
+++ b/iceoryx_posh/source/roudi/port_manager.cpp
@@ -596,8 +596,9 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
     }
 
     // we can create a new port
-    auto options = popo::PublisherOptions{historyCapacity, portConfigInfo.memoryInfo};
-    auto maybePublisherPortData = m_portPool->addPublisherPort(service, payloadMemoryManager, processName, options);
+    auto options = popo::PublisherOptions{historyCapacity};
+    auto maybePublisherPortData =
+        m_portPool->addPublisherPort(service, payloadMemoryManager, processName, options, portConfigInfo.memoryInfo);
     if (!maybePublisherPortData.has_error())
     {
         m_portIntrospection.addPublisher(maybePublisherPortData.value(), processName, service, node);
--- a/iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
+++ b/iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
@@ -78,7 +78,7 @@ class PublisherPort_test : public Test
     iox::popo::PublisherPortUser m_sutUserSide{&m_publisherPortData};
 
     // publisher port w/ history
-    iox::popo::PublisherOptions m_publisherPortOptions{iox::MAX_PUBLISHER_HISTORY, iox::mepoo::MemoryInfo()};
+    iox::popo::PublisherOptions m_publisherPortOptions{iox::MAX_PUBLISHER_HISTORY};
     iox::popo::PublisherPortData m_publisherPortDataHistory{
         iox::capro::ServiceDescription("x", "y", "z"), "myApp", &m_memoryManager, m_publisherPortOptions};
     iox::popo::PublisherPortUser m_sutWithHistoryUseriSide{&m_publisherPortDataHistory};
--- a/iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
+++ b/iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2020 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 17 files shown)


iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <cstdint>
+
+namespace iox
+{
+namespace popo
+{
+struct PublisherOptions
+{
+    uint64_t historyCapacity{0};
```
elBoberido, line 28: I deliberatelly omitted the `m_` prefix and would suggest to not use it if the member is public. It's just weird to have the `m_` prefix in a public interface.

iceoryx_posh/source/roudi/roudi_cmd_line_parser.cpp
```diff
@@ -244,8 +244,15 @@ void CmdLineParser::printParameters() const noexcept
...
-    LogVerbose() << "Unique RouDi ID: " << m_uniqueRouDiId.value();
-    LogVerbose() << "Process kill delay: " << m_processKillDelay.seconds<uint64_t>() << " ms" ;
+    if (m_uniqueRouDiId)
+    {
+        LogVerbose() << "Unique RouDi ID: " << m_uniqueRouDiId.value();
+    }
+    else
+    {
+        LogVerbose() << "Unique RouDi ID: not set";
+    }
```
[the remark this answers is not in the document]
mossmaurice: whoopsie-daisy :grimacing: I agree 100%!

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
```diff
@@ -22,8 +22,9 @@ namespace iox
 namespace popo
 {
 template <typename T, typename port_t>
-inline BasePublisher<T, port_t>::BasePublisher(const capro::ServiceDescription& service)
-    : m_port(iox::runtime::PoshRuntime::getInstance().getMiddlewarePublisher(service))
+inline BasePublisher<T, port_t>::BasePublisher(const capro::ServiceDescription& service,
+                                               const PublisherOptions& publisherOptions)
+    : m_port(iox::runtime::PoshRuntime::getInstance().getMiddlewarePublisher(service, publisherOptions.historyCapacity))
```
mossmaurice, line 27: According to lcov these lines are not covered. Could you add a test case here?

iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
mossmaurice: I'd vote for sticking to our coding guidelines. If we decide otherwise, please document this exception in the `CONTRIBUTING.md`

iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
```diff
@@ -0,0 +1,34 @@
...
+    mepoo::MemoryInfo memoryInfo;
```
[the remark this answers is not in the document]
budrus: I would not, it is not used so far and I cannot predict the future of it

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -572,8 +573,8 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
     }
 
     // we can create a new port
-    auto maybePublisherPortData = m_portPool->addPublisherPort(
-        service, historyCapacity, payloadMemoryManager, processName, portConfigInfo.memoryInfo);
+    auto options = popo::PublisherOptions{historyCapacity, portConfigInfo.memoryInfo};
```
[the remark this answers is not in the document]
mossmaurice: IIRC @MatthiasKillat wanted to have the `portConfigInfo` removed? Is it actually used anywhere other than with default values?

mossmaurice (dismissed):
Please either document the proposed coding guideline exception or add the "m" in Hungarian notation.

iceoryx_posh/source/roudi/port_manager.cpp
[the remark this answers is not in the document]
budrus: I see. It is already in the runtime as parameter but so far was not exposed to the user with publisher and subscriber API. So far, I would keep it hidden. When we see that it must become a part of the public API we can later integrate it in the publisher options. What do you think @elfenpiff ?

iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
budrus: Aren't there already other m_ member in thinks like ChunkHeader that we already expose to the user?

budrus (changes requested):
I would remove MemoryInfo from the public API

iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
elBoberido: yes, at least in the ChunkHeader. I'm currently not aware of other structs where that's the case. I would also change it there and add it to the `CONTRIBUTING.md` if there are no objections

elBoberido: @mossmaurice it's now or with the 2.0 release

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
elBoberido: I just had a look at it and I noticed that I need to mock the `PoshRuntime`. Alternatively, the RouDiEnvironment could be used, but it's quite heavy weight for a module tests and I think a `PoshRuntimeMock` could help with many other tests. It's quite some work though and if you don't mind, I would create a separate issue for this.

iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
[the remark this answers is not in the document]
elBoberido: done ... and I'm disappointed that you didn't notice there was no `U` suffix ;)

iceoryx_posh/include/iceoryx_posh/popo/modern_api/publisher_options.hpp
elBoberido: any other comments on this?

elfenpiff: @elfenpiff  I vote for getting rid of `m_` for public members. The gain is not present but the pain is!

iceoryx_posh/source/roudi/port_manager.cpp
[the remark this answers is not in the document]
elfenpiff: The `PortConfigInfo`s struct intention was to store the place where the port is running and how the memory/device is configured. But at the moment it is nowhere used and can be deleted or at least hidden from the user API.

Maybe we require it later or maybe we require it never depends on which path we would like to follow for GPUs/TPUs ...

dkroenke:
This commit has a typo: [https://github.com/eclipse/iceoryx/pull/441/commits/0aea15cb8f6b2dd0930b4c8a5e9552f8cbe191c1](https://github.com/eclipse/iceoryx/pull/441/commits/0aea15cb8f6b2dd0930b4c8a5e9552f8cbe191c1)

iceoryx_posh/source/roudi/port_manager.cpp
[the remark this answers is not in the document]
budrus: I would go for the hide option for now

iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
```diff
@@ -370,33 +371,35 @@ TEST_F(PublisherPort_test, sendWhenSubscribedDeliversAChunk)
...
+    iox::popo::PublisherOptions options;
+    options.historyCapacity = 1U;
+    iox::popo::PublisherPortData publisherPortDataHistory{
+        iox::capro::ServiceDescription("x", "y", "z"), "myApp", &m_memoryManager, options};
+    iox::popo::PublisherPortUser sutWithHistoryUseriSide{&publisherPortDataHistory};
+    iox::popo::PublisherPortRouDi sutWithHistoryRouDiSide{&publisherPortDataHistory};
     // do it the ara field like way
     // 1. publish a chunk to a not yet offered publisher
-    auto maybeChunkHeader = m_sutWithHistoryUseriSide.tryAllocateChunk(sizeof(DummySample));
+    auto maybeChunkHeader = sutWithHistoryUseriSide.tryAllocateChunk(sizeof(DummySample));
```
budrus, line 387: Ahh, here the m_ is also no more good enough ;-)

budrus (dismissed):

elBoberido:
> This commit has a typoin his commit message (480 instead of 408): [0aea15c](https://github.com/eclipse/iceoryx/commit/0aea15cb8f6b2dd0930b4c8a5e9552f8cbe191c1)

upsi, fixed

budrus (dismissed):

MatthiasKillat (approved):

MatthiasKillat:
From what I see that enables us to add more options to the publisher construction later and avoids changes on intermediate levels due to the struct passed. Should be useful in the future.

iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
MatthiasKillat: Personally I would encourage all to use the following rule: for structs or classes which do not have to encapsulate members and thus have them public, do not use the m_ prefix. It is just ugly for access then and I question the design in such a case.

Now, which classes are that? Simple: classes that do not really operate on their data AND for which any state set by the user, e.g. points in 3d space. But not something like a struct with two members a and b which always need to add up to some constant to be usable, the user can mess this up.

iceoryx_posh/include/iceoryx_posh/internal/popo/modern_api/base_publisher.inl
elBoberido: adde #449

iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
elBoberido: @MatthiasKillat I will update the guidelines according your words

elBoberido:
had to resolve merge conflicts

elfenpiff (approved):

dkroenke (approved):

Issue #332, opened 2020-11-02
App crash with Mq Error Code 15


marthtz:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.4.0

**Observed result or behaviour:**
Add a session ID to the message queue communication to filter outdated messages when an application restarts.

**Expected result or behaviour:**
 Application registers at RouDi and runs like expected.

**Conditions where it occurred / Performed steps:**
Quite hard, since it is a race. But is is possible with a debugger. Start RouDi and an application. Just stop the activation thread in RouDi right before the WAKEUP_TRIGGER is sent. Stop and restart the application.
Quite rare, but when this bug is triggered, then the application registers at RouDi and receives an outdated message when an application/sender/receiver port is requested. The application terminates itself.

**TODO**
- [ ] Re-write iceoryx_posh/test/moduletests/test_mq_interface.cpp to test only POSH specific functionality of the IpcInterface e.g. getShmTopic(), IpcChannel is already covered by iceoryx_utils/test/moduletests/test_ipc_channel.cpp (follow-up from #381 )

marthtz:
To actually fix this bug, we need to send a session ID with each mqueue message. With the session ID, the recipient can check if the received message is valid or from an outdated session.

The message queue communication needs a more sophisticated protocol with an header and payload. Since the communication can break in several ways, I would propose the following changes. To fix this specific bug, only the first step needs to be implemented.

Step 1
- MqMessage is split into header and payload, where the header is just an fixed, reserved amount of space at the beginning of the internal string and filled in sendMessageToProcess with the session ID which is passed alongside the MqMessage
- to be able to create a more sophisticated header in the next steps, the first byte is the magic number 0xFF
- followed by 8 byte session id
- followed by two bytes payload size and the actual payload, which consists of the current protocol
- to keep our tests running, we need an "always valid" session id. this could be either 0 or -1u. this is not great for security reasons, but doesn't make it worse than the current situation
- framing
   - magic number (0xFF): 1 byte -> maybe have this as protocol version byte and also use it later
   - session id (uint64_t): 8 byte
   - payload length (uint16_t, N): 2 bytes
   - payload: N bytes

Step 2
- split MqMessageType into several enums, since it contains to much unrelated stuff
- MqMessageType (uint8_t)
   - REQUEST-> for e.g. IMPL_SENDER
   - RESPONSE -> for e.g. IMPL_SENDER
[...]

PR #446, opened 2020-12-17
Iox #408 add struct with options to subscriber

### LINKED ISSUE #408: Refine modern pub/sub API

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
This is a follow up to #441 and  and based on the branch from that PR. It adds a `SubscriberOptions` struct to configure `historyRequest` and `queueCapacity`. Documentation and examples will be updated in the next PR.
Please wait with review until #441 is merged.

1. [x] Merge

## References

- Partly closes #408

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,6 +90,7 @@ See [error-handling.md](./doc/error-handling.md) for additional information abou
 * Methods and variables in `lowerCamelCase`: `uint16_t myVariable`
 * Compile time constants, also enum values in `UPPER_SNAKE_CASE`: `static constexpr uint16_t MY_CONSTANT`
 * Class members start with `m_`: `m_myMember`
+    * Public members of structs and classes do not have the `m_` prefix
 * Namespaces in `lower_snake_case` : `my_namespace`
 * Aliases have a `_t` postfix : `using FooString_t = iox::cxx::string<100>;`
 
--- a/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
@@ -27,13 +27,15 @@ typedef struct cpp2c_Subscriber* iox_sub_t;
 /// @param[in] service serviceString
 /// @param[in] instance instanceString
 /// @param[in] event eventString
-/// @param[in] historyCapacity size of the history chunk queue
+/// @param[in] queueCapacity size of the receiver queue
+/// @param[in] historyCapacity max number of chunks  received as history after subscription
 /// @return handle of the subscriber
 iox_sub_t iox_sub_init(iox_sub_storage_t* self,
                        const char* const service,
                        const char* const instance,
                        const char* const event,
-                       uint64_t historyRequest);
+                       const uint64_t queueCapacity,
+                       const uint64_t historyRequest);
 
 /// @brief deinitialize a subscriber handle
 /// @param[in] self the handle which should be removed
--- a/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
@@ -43,8 +43,7 @@ void iox_sub_deinit(iox_sub_t const self);
 
 /// @brief subscribes to the service
 /// @param[in] self handle to the subscriber
-/// @param[in] queueCapacity size of the receiver queue
-void iox_sub_subscribe(iox_sub_t const self, const uint64_t queueCapacity);
+void iox_sub_subscribe(iox_sub_t const self);
 
 /// @brief unsubscribes from a service
 /// @param[in] self handle to the subscriber
--- a/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
@@ -28,7 +28,7 @@ typedef struct cpp2c_Subscriber* iox_sub_t;
 /// @param[in] instance instanceString
 /// @param[in] event eventString
 /// @param[in] queueCapacity size of the receiver queue
-/// @param[in] historyCapacity max number of chunks  received as history after subscription
+/// @param[in] historyRequest of chunks received after subscription if chunks are available
 /// @return handle of the subscriber
 iox_sub_t iox_sub_init(iox_sub_storage_t* self,
                        const char* const service,
--- a/iceoryx_binding_c/test/moduletests/test_event_info.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_event_info.cpp
@@ -61,8 +61,7 @@ class iox_event_info_test : public Test
 
     void Subscribe(SubscriberPortData* ptr)
     {
-        uint64_t queueCapacity = MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY;
-        iox_sub_subscribe(m_subscriberHandle, queueCapacity);
+        iox_sub_subscribe(m_subscriberHandle);
 
         SubscriberPortSingleProducer(ptr).tryGetCaProMessage();
         iox::capro::CaproMessage caproMessage(iox::capro::CaproMessageType::ACK, TEST_SERVICE_DESCRIPTION);
@@ -87,8 +86,11 @@ class iox_event_info_test : public Test
     Allocator m_memoryAllocator{m_memory, MEMORY_SIZE};
     MePooConfig m_mempoolconf;
     MemoryManager m_memoryManager;
-    iox::popo::SubscriberPortData m_portPtr{
-        TEST_SERVICE_DESCRIPTION, "myApp", iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer};
+    SubscriberOptions m_subscriberOPtions{MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY, 0U};
+    iox::popo::SubscriberPortData m_portPtr{TEST_SERVICE_DESCRIPTION,
+                                            "myApp",
+                                            iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer,
+                                            m_subscriberOPtions};
     ChunkQueuePusher<SubscriberPortData::ChunkQueueData_t> m_chunkPusher{&m_portPtr.m_chunkReceiverData};
     cpp2c_Subscriber m_subscriber;
     iox_sub_t m_subscriberHandle = &m_subscriber;
@@ -231,4 +233,3 @@ TEST_F(iox_event_info_test, callbackCanBeCalledMultipleTimes)
 
     EXPECT_EQ(m_lastEventCallbackArgument, &m_userTrigger);
 }
-
--- a/iceoryx_binding_c/test/moduletests/test_subscriber.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_subscriber.cpp
@@ -62,7 +62,6 @@ class iox_sub_test : public Test
 
     void Subscribe(SubscriberPortData* ptr)
     {
-        uint64_t queueCapacity = MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY;
         iox_sub_subscribe(m_sut);
 
         SubscriberPortSingleProducer(ptr).tryGetCaProMessage();
@@ -87,10 +86,11 @@ class iox_sub_test : public Test
 
     iox::cxx::GenericRAII m_uniqueRouDiId{[] { iox::popo::internal::setUniqueRouDiId(0); },
                                           [] { iox::popo::internal::unsetUniqueRouDiId(); }};
+    iox::popo::SubscriberOptions subscriberOptions{MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY, 0U};
     iox::popo::SubscriberPortData m_portPtr{TEST_SERVICE_DESCRIPTION,
                                             "myApp",
                                             iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer,
-                                            iox::popo::SubscriberOptions{1U, 0U}};
+                                            subscriberOptions};
     ChunkQueuePusher<SubscriberPortData::ChunkQueueData_t> m_chunkPusher{&m_portPtr.m_chunkReceiverData};
     std::unique_ptr<cpp2c_Subscriber> m_subscriber{new cpp2c_Subscriber};
     iox_sub_t m_sut = m_subscriber.get();
--- a/iceoryx_binding_c/test/moduletests/test_subscriber.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_subscriber.cpp
@@ -63,7 +63,7 @@ class iox_sub_test : public Test
     void Subscribe(SubscriberPortData* ptr)
     {
         uint64_t queueCapacity = MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY;
-        iox_sub_subscribe(m_sut, queueCapacity);
+        iox_sub_subscribe(m_sut);
 
         SubscriberPortSingleProducer(ptr).tryGetCaProMessage();
         iox::capro::CaproMessage caproMessage(iox::capro::CaproMessageType::ACK, TEST_SERVICE_DESCRIPTION);
@@ -87,8 +87,10 @@ class iox_sub_test : public Test
 
     iox::cxx::GenericRAII m_uniqueRouDiId{[] { iox::popo::internal::setUniqueRouDiId(0); },
                                           [] { iox::popo::internal::unsetUniqueRouDiId(); }};
-    iox::popo::SubscriberPortData m_portPtr{
-        TEST_SERVICE_DESCRIPTION, "myApp", iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer};
+    iox::popo::SubscriberPortData m_portPtr{TEST_SERVICE_DESCRIPTION,
+                                            "myApp",
+                                            iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer,
+                                            iox::popo::SubscriberOptions()};
     ChunkQueuePusher<SubscriberPortData::ChunkQueueData_t> m_chunkPusher{&m_portPtr.m_chunkReceiverData};
     std::unique_ptr<cpp2c_Subscriber> m_subscriber{new cpp2c_Subscriber};
     iox_sub_t m_sut = m_subscriber.get();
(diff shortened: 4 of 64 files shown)


iceoryx_binding_c/test/moduletests/test_subscriber.cpp
```diff
@@ -105,8 +107,8 @@ TEST_F(iox_sub_test, initialStateNotSubscribed)
 
 TEST_F(iox_sub_test, offerLeadsToSubscibeRequestedState)
 {
-    uint64_t queueCapacity = 1u;
-    iox_sub_subscribe(m_sut, queueCapacity);
+    uint64_t queueCapacity = 1U;
```
budrus, line 110: no more used, or?

iceoryx_binding_c/test/moduletests/test_subscriber.cpp
```diff
@@ -115,8 +117,8 @@ TEST_F(iox_sub_test, offerLeadsToSubscibeRequestedState)
 
 TEST_F(iox_sub_test, NACKResponseLeadsToSubscribeWaitForOfferState)
 {
-    uint64_t queueCapacity = 1u;
-    iox_sub_subscribe(m_sut, queueCapacity);
+    uint64_t queueCapacity = 1U;
```
budrus, line 120: same

iceoryx_binding_c/test/moduletests/test_subscriber.cpp
```diff
@@ -127,8 +129,8 @@ TEST_F(iox_sub_test, NACKResponseLeadsToSubscribeWaitForOfferState)
 
 TEST_F(iox_sub_test, ACKResponseLeadsToSubscribedState)
 {
-    uint64_t queueCapacity = 1u;
-    iox_sub_subscribe(m_sut, queueCapacity);
+    uint64_t queueCapacity = 1U;
```
budrus, line 132: same

iceoryx_binding_c/test/moduletests/test_subscriber.cpp
```diff
@@ -139,8 +141,8 @@ TEST_F(iox_sub_test, ACKResponseLeadsToSubscribedState)
 
 TEST_F(iox_sub_test, UnsubscribeLeadsToUnscribeRequestedState)
 {
-    uint64_t queueCapacity = 1u;
-    iox_sub_subscribe(m_sut, queueCapacity);
+    uint64_t queueCapacity = 1U;
```
budrus, line 144: and here

iceoryx_dds/include/iceoryx_dds/internal/gateway/dds_to_iox.inl
```diff
@@ -42,7 +42,7 @@ inline void DDS2IceoryxGateway<channel_t, gateway_t>::loadConfiguration(const co
             LogDebug() << "[DDS2IceoryxGateway] Setting up channel for service: {"
                        << serviceDescription.getServiceIDString() << ", " << serviceDescription.getInstanceIDString()
                        << ", " << serviceDescription.getEventIDString() << "}";
-            setupChannel(serviceDescription);
+            setupChannel(serviceDescription, popo::PublisherOptions());
```
budrus, line 45: So currently in the DDS gateway the history is always 0?

iceoryx_examples/ice_multi_publisher/ice_resubscriber.cpp
```diff
@@ -35,10 +35,11 @@ constexpr uint64_t UNSUBSCRIBED_TIME_SECONDS{3U};
 
 void receive()
 {
-    iox::popo::TypedSubscriber<CounterTopic> subscriber({"Group", "Instance", "Counter"});
+    iox::popo::SubscriberOptions options;
+    options.queueCapacity = MAX_NUMBER_OF_SAMPLES - 2U;
```
budrus, line 39: I feel this makes no more sense as the queue size is no more adapted. So I would remove `MAX_NUMBER_OF_SAMPLES` completely or use it to set the capacity but without the magic -2

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
```diff
@@ -142,7 +141,8 @@ class BaseSubscriber
 
   protected:
     BaseSubscriber() noexcept; // Required for testing.
-    BaseSubscriber(const capro::ServiceDescription& service) noexcept;
+    BaseSubscriber(const capro::ServiceDescription& service,
+                   const SubscriberOptions& subscriberOptions = SubscriberOptions()) noexcept;
```
budrus, line 145: Why do we have a default for the subscriber but not the publisher? Do we need a default in `BaseSubscriber`?

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
```diff
@@ -64,27 +65,27 @@ class PortPool
...
                      mepoo::MemoryManager* const memoryManager,
                      const ProcessName_t& applicationName,
+                     const popo::PublisherOptions& publisherOptions = popo::PublisherOptions(),
                      const mepoo::MemoryInfo& memoryInfo = mepoo::MemoryInfo()) noexcept;
 
     cxx::expected<SubscriberPortType::MemberType_t*, PortPoolError>
     addSubscriberPort(const capro::ServiceDescription& serviceDescription,
-                      const uint64_t& historyRequest,
                       const ProcessName_t& applicationName,
+                      const popo::SubscriberOptions& subscriberOptions,
```
budrus, line 76: also here, why a default for the publisher but not for the subscriber?

iceoryx_posh/source/popo/ports/subscriber_port_user.cpp
```diff
@@ -37,31 +37,13 @@ SubscriberPortUser::MemberType_t* SubscriberPortUser::getMembers() noexcept
...
 
-void SubscriberPortUser::subscribe(const uint64_t queueCapacity) noexcept
+void SubscriberPortUser::subscribe() noexcept
 {
     if (!getMembers()->m_subscribeRequested.load(std::memory_order_relaxed))
     {
         // start with new chunks, drop old ones that could be in the queue
         m_chunkReceiver.clear();
 
-        /// @todo is it safe to change the capacity when it is no more the initial subscribe?
```
budrus, line 47: one worry less :-)

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -225,13 +236,25 @@ PoshRuntime::requestPublisherFromRoudi(const MqMessage& sendBuffer) noexcept
 
 SubscriberPortUserType::MemberType_t*
 PoshRuntime::getMiddlewareSubscriber(const capro::ServiceDescription& service,
-                                     const uint64_t& historyRequest,
+                                     const popo::SubscriberOptions& subscriberOptions,
                                      const NodeName_t& nodeName,
                                      const PortConfigInfo& portConfigInfo) noexcept
 {
+    constexpr uint64_t MAX_QUEUE_CAPACITY = SubscriberPortUserType::MemberType_t::ChunkQueueData_t::MAX_CAPACITY;
+
```
budrus, line 244: We sould also check against MAX_HISTORY. I fear this is missing so far

budrus (changes requested):

iceoryx_dds/include/iceoryx_dds/internal/gateway/dds_to_iox.inl
elBoberido: yes, I wanted to keep the behavior the same. Shall we change it?

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: `MAX_HISTORY_CAPACITY` is checked. See a few lines above. What's not checked is `historyRequest`, but I thought since this can be greater than the configured `historyCapacity`, it doesn't matter if it's also greater than `MAX_HISTORY_CAPACITY`

iceoryx_examples/ice_multi_publisher/ice_resubscriber.cpp
elBoberido: maybe the whole example is obsolete? Do we really need to show how to repeatedly call `subscribe` and `unsubscribe`?

iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
```diff
@@ -28,22 +28,23 @@ typedef CLASS cpp2c_WaitSet* iox_ws_t;
 /// @param[in] service serviceString
 /// @param[in] instance instanceString
 /// @param[in] event eventString
-/// @param[in] historyCapacity size of the history chunk queue
+/// @param[in] queueCapacity size of the receiver queue
+/// @param[in] historyCapacity max number of chunks  received as history after subscription
```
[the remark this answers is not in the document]
elBoberido: right

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
elBoberido: I did it on Thursday, which has an odd number of `u`, hence the default value. Since we now have Friday, it's removed

iceoryx_posh/include/iceoryx_posh/roudi/port_pool.hpp
elBoberido: removed from publisher

iceoryx_examples/ice_multi_publisher/ice_resubscriber.cpp
budrus: That's true. You could even drop this and adapt the README.

iceoryx_posh/source/runtime/posh_runtime.cpp
budrus: OK, that's true. We do not know on subscriber side which historyCapacity is available on publisher side. If the requested history is larger than the available one, we should print a warning on publisher side. Maybe this is even done?

elBoberido: do you know where I have to look?

iceoryx_examples/ice_multi_publisher/ice_resubscriber.cpp
elBoberido: can I do this in a follow up PR? I also need to adjust the example to show how to use the history feature and this PR is already quite large

iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
```diff
@@ -27,22 +27,23 @@ typedef struct cpp2c_Subscriber* iox_sub_t;
...
+/// @param[in] queueCapacity size of the receiver queue
+/// @param[in] historyRequest of chunks received after subscription if chunks are available
 /// @return handle of the subscriber
 iox_sub_t iox_sub_init(iox_sub_storage_t* self,
                        const char* const service,
                        const char* const instance,
                        const char* const event,
-                       uint64_t historyRequest);
+                       const uint64_t queueCapacity,
+                       const uint64_t historyRequest);
```
elBoberido, line 38: @elfenpiff do you think we should also provide a option struct here?

elBoberido:
@ithier could you have a look at the DDS adjustments?

iceoryx_posh/source/runtime/posh_runtime.cpp
budrus: I check

budrus: It is not checked. I remember that I thought about whether to respond with NACK if history request and history capacity do not match. But then we cannot differentiate whether the NACK is because there is no publisher available or because of a history mismatch. And then we need extra info in NACK. Printing a warning in chunk_distributor.inl is all I would do for now. But this can also be done in a follow up PR

budrus (approved):

iceoryx_posh/test/integrationtests/test_popo_chunk_building_blocks.cpp
```diff
@@ -54,7 +54,7 @@ struct ChunkDistributorConfig
 
 struct ChunkQueueConfig
 {
-    static constexpr uint64_t MAX_QUEUE_CAPACITY = iox::MAX_SUBSCRIBER_QUEUE_CAPACITY;
+    static constexpr uint64_t MAX_QUEUE_CAPACITY = NUM_CHUNKS_IN_POOL / 3;
```
elBoberido, line 57: @budrus this time I haven't had luck. Hope this make it more robust. The `MAX_QEUEU_CAPACITY` is now the same like before, just the amount of chunks in the mempools are now three times as high

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: btw, with the proposed fix for #332 it would be possible to send NACK with error codes. Shall I add a comment for refactoring after #332 is finished?

CONTRIBUTING.md
```diff
@@ -90,6 +90,7 @@ See [error-handling.md](./doc/error-handling.md) for additional information abou
 * Methods and variables in `lowerCamelCase`: `uint16_t myVariable`
 * Compile time constants, also enum values in `UPPER_SNAKE_CASE`: `static constexpr uint16_t MY_CONSTANT`
 * Class members start with `m_`: `m_myMember`
+    * Public members of structs and classes do not have the `m_` prefix
```
mossmaurice, line 93: Thanks for adding this exception

iceoryx_posh/include/iceoryx_posh/gateway/channel.hpp
```diff
@@ -17,6 +17,8 @@
 
 #include "iceoryx_posh/capro/service_description.hpp"
 #include "iceoryx_posh/iceoryx_posh_types.hpp"
+#include "iceoryx_posh/popo/publisher_options.hpp"
+#include "iceoryx_posh/popo/subscriber_options.hpp"
```
mossmaurice, line 21: Are the includes needed? It's a template parameter down there

iceoryx_posh/include/iceoryx_posh/gateway/gateway_generic.hpp
```diff
@@ -20,6 +20,8 @@
 #include "iceoryx_posh/gateway/gateway_config.hpp"
 #include "iceoryx_posh/iceoryx_posh_config.hpp"
 #include "iceoryx_posh/iceoryx_posh_types.hpp"
+#include "iceoryx_posh/popo/publisher_options.hpp"
+#include "iceoryx_posh/popo/subscriber_options.hpp"
```
mossmaurice, line 24: Check if includes are needed

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/subscriber_port_user.hpp
```diff
@@ -19,6 +19,7 @@
 #include "iceoryx_posh/internal/popo/ports/base_port.hpp"
 #include "iceoryx_posh/internal/popo/ports/subscriber_port_data.hpp"
 #include "iceoryx_posh/mepoo/chunk_header.hpp"
+#include "iceoryx_posh/popo/subscriber_options.hpp"
```
mossmaurice, line 22: Remove this include?

iceoryx_posh/test/integrationtests/test_popo_chunk_building_blocks.cpp
mossmaurice: IIRC @dkroenke has pinged @jpsamper2009 for dedicated hardware already

mossmaurice (approved):
Please double check the mentioned includes, otherwise LTGM

iceoryx_posh/source/runtime/posh_runtime.cpp
budrus: This NACK here is a CaPro message in the discovery loop. What you mention are responses over the message queue. So I'm not sure if you mix something up

budrus (approved):

iceoryx_posh/test/integrationtests/test_popo_chunk_building_blocks.cpp
elBoberido: The test runs out of chunks again. This should not be possible. I don't understand why it fails this way

budrus: Hmm. Maybe we do not release the chunks we cannot push in a full FiFo. Normally we only use overflowing queues for chunk communication

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/subscriber_port_user.hpp
elBoberido: This makes it kind of easier. It would be needed in `subriber_port_data.hpp` but the `subscriber_options.hpp` need a default value from `subriber_port_data.hpp` and then there would be a cyclic dependency. Currently there is a forward declaration to `SubscriberOptions`. One idea would be to either include `subscriber_options.hpp` at the end of `subriber_port_data.hpp` or move the `ChunkQueueData_t` alias to `iceoryx_posh_types.hpp`

iceoryx_posh/include/iceoryx_posh/gateway/channel.hpp
elBoberido: nope, those are remainders of a failed first attempt. Will remove it in the follow up PR

iceoryx_posh/include/iceoryx_posh/gateway/gateway_generic.hpp
elBoberido: will be removed

iceoryx_binding_c/test/moduletests/test_event_info.cpp
```diff
@@ -87,8 +86,11 @@ class iox_event_info_test : public Test
     Allocator m_memoryAllocator{m_memory, MEMORY_SIZE};
     MePooConfig m_mempoolconf;
     MemoryManager m_memoryManager;
-    iox::popo::SubscriberPortData m_portPtr{
-        TEST_SERVICE_DESCRIPTION, "myApp", iox::cxx::VariantQueueTypes::SoFi_SingleProducerSingleConsumer};
+    SubscriberOptions m_subscriberOPtions{MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY, 0U};
```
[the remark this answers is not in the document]
elBoberido: fixed in follow up PR

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: Whopsi, yes. Although this could also be checked when the SubscriberPort is requested from the runtime

budrus: Hmm. But only if the publisher port is already present...

elBoberido: This is the reason Chuck Norris admires you

Issue #409, opened 2020-12-07
Discussion and meeting on static code analysis


mossmaurice:
## Brief feature description

Several companies are working on Eclipse iceoryx nowadays. The maintainers chose [QAC](https://www.perforce.com/blog/qac/why-eclipse-iceoryx-uses-helix-qac) from Perforce to check the certain rulesets common in the automotive industry w.r.t. to the ISO26262. Beside local scans things like the CI integration or upload to the Perforce server never went live. Therefore, let's revive this topic :-)

Comes from #356

## Detailed information

Proposed proceeding:

1) Re-do assessment with all maintainers
  * Choose a single vendor or agree to a suppression syntax to be able to use different tools?
2) Talk to the vendors
3) Make decision
4) Setup CI and infrastructure (tbd in a future issue, not here)

Pending list of criteria for the assessment (not complete yet):
* Developer experience with the tool (prior knowledge)
* Developer convenience (usability)
  * How easy to use is the tool for developers?
* Suppression syntax
  * How does the suppression syntax look like?
  * Are there dependencies?
  * Common MISRA-C++:2020 suppression syntax supported?
* Commercial agreement needed (affordability)?
* Maintenance
* Ability to export results
* SLA and support
  * What happens if developers run in trouble?
  * How fast can the vendor react?
* CI support
* Coding rules support (extensibility, customizability)
  * Which coding rules are supported (e.g. AUTOSAR, MISRA, CERT)?
* What is experience of developers with false positives (correctness)?
* How long does a scan take with the iceoryx code base (efficiency)?

mossmaurice:
@elfenpiff @elBoberido and myself had a call with [Axivion](https://www.axivion.com/en) today.

Major points:
* Axivion did an analysis of iceoryx
  * No parsing errors
  * Analysis time is 15 minutes
* Plugins for various IDEs available e.g. VSCode and Eclipse IDE
* Webinterface/dashboard to display the scan results is available
* Axivion can parse Lint and PRQA suppression syntax

Next steps:
* [ ] @mossmaurice tries out Axivion
* [ ] Setup meeting with Axivion sales in January

mossmaurice:
Today I learned that MISRAC++:2020 will introduce a [common suppression syntax](https://www.heise.de/hintergrund/MISRA-C-bietet-Richlinien-und-Konformitaet-auch-fuer-neuen-Sprachstandard-C-20-4980750.html?seite=all), so that different SCA tools can be used.

[...]

PR #473, opened 2021-01-02
Iox #391 static memory alternative for std function

### LINKED ISSUE #391: static memory alternative for std::function
## Brief feature description

We need an alternative to std::function as a owning generic function wrapper. This alternative should 
* not use dynamic memory (or make this optional/configurable)
* not employ virtual functions
* own all data to execute the callable
* be compatible with functors (lambdas) and function pointers

## Detailed information

This is needed to store callables without using dynamic memory as in std::function.
 Internal virtual function calls are not allowed if we want to store these callables in shared memory.

This can potentially help to resolve https://github.com/eclipse/iceoryx/issues/131, but here it may not be needed (using function_ref may be sufficient). Nevertheless an owning wrapper compatible with our constraints would be useful in future software design.

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

1. [x] Merge

## References

- Closes #391

## Summary
Added a static storage alternative for std::function including tests. This is supposed to store any callable up to a given storage size. As is also the case for sd::function, it cannot e used in shared memory (this cannot be avoided due to the fact that pointers to e.g. free functions but also lambdas are stored which will usually not be valid in another process).

The implementation uses cxx::function_ref, otherwise parts of those functionality would have been rewritten/duplicated. It is important to note that much of the complications arise due to the fact that information about the underlying stored type is lost but it still must be ensured that it is copied/moved properly (according to RAII). Naturally these problems do NOT arise when only a reference is stored (as in function_ref).

Fixed an issue that function_ref was not constructible from function pointers of free functions (only function names, but those are not function pointers but are converted to them). There is still an open point of a conversion from something potentially not convertible to a void pointer in the function_ref ctor (existed before as well, as it would take a universal reference to anything callable). For function_pointers this is compliant with the POSIX standard, but in general it is not (function pointers cannot be converted to void pointers in general).

This requires some further investigation. A potential fix could be added here as well.


iceoryx_utils/include/iceoryx_utils/internal/cxx/function_ref.inl
```diff
@@ -34,6 +34,21 @@ inline function_ref<ReturnType(ArgTypes...)>::function_ref(CallableType&& callab
 {
 }
 
+template <class ReturnType, class... ArgTypes>
+inline function_ref<ReturnType(ArgTypes...)>::function_ref(ReturnType (*function)(ArgTypes...))
```
mossmaurice, line 38: I'd suggest to avoid an additional overload here. It should be possible with `std::is_object` to determine what kind of callable was passed to the c'tor. However, I suppose `std::addressof` to be [implemented with std::is_object](https://en.cppreference.com/w/cpp/memory/addressof).

mossmaurice (commented):
Only reviewed `function_ref` and `type_trait` changes so far.

iceoryx_utils/include/iceoryx_utils/cxx/type_traits.hpp
```diff
@@ -46,6 +46,30 @@ struct is_invocable
     static constexpr bool value = decltype(test<Callable, ArgTypes...>(nullptr))::value;
 };
 
+///
+/// @brief Verifies whether the passed Callable type is in fact invocable with the given arguments
+///        and the result of the invocation is ReturnType
+///
+template <typename ReturnType, typename Callable, typename... ArgTypes>
+struct is_invocable_r
```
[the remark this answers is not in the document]
MatthiasKillat: Really? I do not remember, it seemed not to be sufficient but I can take a closer look again. ``is_invocable_r`` is as in the C++17 standard, if we need just one we could consider using the one more similar to the standard.

iceoryx_utils/include/iceoryx_utils/internal/cxx/function_ref.inl
MatthiasKillat: Well, for function pointer we cannot use ``std::addressof`` it seems. It was not working at all for function pointers before (but please double check). Personally I consider using an almost unrestricted version of universal reference dangerous, it can essentially take anything. And there is still an issue that it can be used with e.g.method pointers and I am not sure that anything that is happening inside the function is fully portable in this case alone.

If the behavior is (even slighly) different for specific types, there is no way around an overload.

iceoryx_utils/include/iceoryx_utils/internal/cxx/static_storage.inl
```diff
@@ -0,0 +1,92 @@
...
+#define IOX_UTILS_CXX_STATIC_STORAGE_INL
+
+#include "iceoryx_utils/internal/cxx/static_storage.hpp"
+
+namespace iox
+{
+namespace cxx
+{
+template <uint64_t Capacity, uint64_t Align>
+constexpr uint64_t static_storage<Capacity, Align>::align_delta(uint64_t align, uint64_t alignTarget)
```
ithier, line 25: The arguments could be const ref.

ithier: Also, why snake case here ? Does this class reimplement the STL ?

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,162 @@
...
+{
+namespace cxx
+{
+template <typename ReturnType, typename... Args>
+using signature = ReturnType(Args...);
+
+template <typename StorageType, typename T>
+class storable_function;
+
+template <typename StorageType, typename ReturnType, typename... Args>
```
ithier, line 36: Please add class documentation.
What is this class ? How can I use it ?

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,162 @@
...
+            {
+                destroyFunction(f);
+            }
+        }
+    };
+
+    vtable m_vtable;
+
+    /// @note  in general we cannot know the alignment of the type we want to store at construction time
+    StorageType m_storage;
```
ithier, line 84: I might be wrong, but I thought the convention was to put member variables at the end of the class definition.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,250 @@
...
+    g = std::move(tmp);
+}
+
+template <typename S, typename ReturnType, typename... Args>
+template <typename Functor, typename>
+void storable_function<S, signature<ReturnType, Args...>>::storeFunctor(const Functor& functor) noexcept
+{
+    ///@todo we may need something more to get the correct type in general...
+    using StoredType = typename std::remove_reference<Functor>::type;
+    auto p = m_storage.template allocate<StoredType>();
```
ithier, line 149: Whats `p` ? `ptr` ?

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,250 @@
...
+
+    if (p)
```
ithier, line 151: This condition check could be more self-documenting.

ithier (changes requested):

ithier:
Idea looks sound to me. Just needs some documentation !

iceoryx_utils/include/iceoryx_utils/cxx/function.hpp
```diff
@@ -0,0 +1,46 @@
...
+/// @note set the storage type and reorder template arguments for the user API
+/// the storage type must precede the (required) variadic arguments in the internal one
+/// if the static storage is insufficient to store the callable we get a compile time error
+
+/// @note this alias is needed to improve usability
+///       for the API see storable_function
+template <typename Signature, uint64_t Bytes = 128>
+using function = storable_function<static_storage<Bytes>, Signature>;
+
+/// @note the following would essentially be a complete std::function replacement
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, it is a little work in progress but was supposed to show how altering the storage policy can be used to create a type with behavior even closer to std::function. It is also a little unfortunate to use an alias here, but it is needed for this kind of generic programming (create types from other types by setting some types to specific types, like the storage here). I am open for restructuring suggestions but would not like to lose this generality.

iceoryx_utils/include/iceoryx_utils/internal/cxx/static_storage.inl
MatthiasKillat: Nope, but it is low level and I like it in this case. I really dislike Upper Camel Case for those kinds of constructions (similar with e.g. relative_ptr, which are supposed to mimic shared_ptr syntax).

MatthiasKillat: The args are primitive types, same size as pointers on a 64 bit system. No point passing them by ref IMO.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
MatthiasKillat: Sure

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
MatthiasKillat: Can do. Needed a private section for internal types at the beginning and used it for members. We have no strict convention for this AFAIK. I prefer defining types, then the public API, then the private API (including members and potentially private types).

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
MatthiasKillat: It is a ptr, yes. Implied by allocate (maybe). Personally I prefer to keep variable names with very small scope short unless something magical happens which can be explained by better names.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
ithier: Would be nice to have some kind of convention. I completely missed these members when scanning the file.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
MatthiasKillat: Hm, we should not go overboard with this. It is quite low level. When the allocation fails we get a nullptr, as usual in many APIs. consider writing something like 

```
const auto allocationSucceeded = p != nullptr;
if(allocationSucceeded) {
...
}
```
overkill.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
ithier: I guess kind of nitpicky but I would argue for using conventional names like `ptr` when possible.
When I see `ptr` I have to spend 0 brain power to immediately understand what's happening.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
ithier: Yeah, my argument would be naming `p` as `ptr` instead makes it immediately clear we are checking a nullptr.
I won't argue for it further though. As you like.

MatthiasKillat:
@Jeff Ithier. I will improve the documentation. I prefer brief docu though, unless the use cases are very tricky. No one reads long docu for functions or classes. :-)

iceoryx_utils/include/iceoryx_utils/cxx/type_traits.hpp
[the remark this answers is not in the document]
mossmaurice: Fine by me. Please remove `has_signature` and refactor all usages.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
mossmaurice: Throwing `pointer` in the ring. Simple, precise, straightforward. Opinions?

iceoryx_utils/include/iceoryx_utils/cxx/function.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+namespace cxx
+{
+/// @note set the storage type and reorder template arguments for the user API
+/// the storage type must precede the (required) variadic arguments in the internal one
+/// if the static storage is insufficient to store the callable we get a compile time error
+
+/// @note this alias is needed to improve usability
+///       for the API see storable_function
+template <typename Signature, uint64_t Bytes = 128>
```
mossmaurice, line 31: Why not expose the 128 to the user so the can set the magic number?

mossmaurice (changes requested):
Great work! Tests are not yet reviewed.

1) Please check if the implementation of `storable_function` can be simplified by using one generic c'tor, as in `std::function`:
```
template< class F, class Alloc >
function( std::allocator_arg_t, const Alloc& alloc, F f );
```

2) Align `cxx::PoorMansHeap` usage

MatthiasKillat:
I will update the PR in a day or two, until then no further review needed.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
dkroenke: `ptr` is my favourite because it's in the middle between `p` and `ptr` and it is common sense (type ptr in the search field of Github).

iceoryx_utils/include/iceoryx_utils/cxx/function.hpp
dkroenke: And it needs the `U`

iceoryx_utils/include/iceoryx_utils/cxx/function.hpp
```diff
@@ -0,0 +1,46 @@
...
+// template <typename Signature, uint64_t Bytes = 128>
+// using function = detail::storable_function<optimized_storage<Bytes>, Signature>;
+// or alternatively
+// template <typename Signature>
+// using function = detail::storable_function<dynamic_storage, Signature>;
+
+} // namespace cxx
+} // namespace iox
+
+#endif // IOX_UTILS_FUNCTION_HPP
```
dkroenke, line 46: missing endline

iceoryx_utils/include/iceoryx_utils/internal/cxx/static_storage.hpp
```diff
@@ -0,0 +1,83 @@
...
+///       but other allocation policies and use them where we need to store objects
+///       with some interchangable storage policy (e.g. in storable_function)
+///       optimized_storage would have a dynamic memory fallback when static memory is
+///       insufficent
+template <uint64_t Capacity, uint64_t Align = 1>
+class static_storage
+{
+  private:
+    alignas(Align) uint8_t m_bytes[Capacity];
+    void* m_ptr{nullptr};
```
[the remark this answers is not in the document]
elBoberido: @mossmaurice `cxx::PoorMansHeap` is a poor name. a better name would probably be `cxx::PolymorphicOptional`. It's similar to a variant, but you do not care about the content as long as it is derived from a common class and it can be empty, which make it similar to an optional. It's for situations where you have a factory which decides at runtime which concrete implementation to serve. You just want to use it like a pointer to a base class. This might be helpful to get rid of some templated classes and still be able to insert mocks in the tests. Currently the `cxx::PoorMansHeap` is not moveable, so the usage is a little bit cumbersome, but I have a proof of concept how to implement that and then you can use it exactly like a factory method with the heap.
E.g.
```c++
using MyClass_t = PoorMansHeap<MyClass, maxSize, maxAlignment>;
...
MyClass_t myClass = getConcreteMyClass();
myClass->foo();
```

MatthiasKillat:
Reworked, can be reviewed now.

iceoryx_utils/include/iceoryx_utils/cxx/function.hpp
MatthiasKillat: It is exposed but defaulted to a reasonable 128.

iceoryx_utils/include/iceoryx_utils/cxx/type_traits.hpp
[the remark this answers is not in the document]
MatthiasKillat: Will be done later.

iceoryx_utils/include/iceoryx_utils/internal/cxx/function_ref.inl
MatthiasKillat: I will create an example which fails with the original later.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @brief construct from functor (including lambda)
+    template <typename Functor,
+              typename = typename std::enable_if<std::is_class<Functor>::value
+                                                     && is_invocable_r<ReturnType, Functor, Args...>::value,
+                                                 void>::type>
+    storable_function(const Functor& functor) noexcept;
+
+    /// @brief construct from function pointer (including static functions)
+    storable_function(ReturnType (*function)(Args...)) noexcept;
```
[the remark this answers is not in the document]
MatthiasKillat: Because the cases are handled differently internally and too generic constructors with universal references cause all kinds of issues, it is especially easy to overlook a type category that can be passed into them and will not actually work.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
MatthiasKillat: pointer is to long, I can call it ptr. Still, I personally am a fan of short variable names in those kind of local scopes and it is fairly obvious that an allocation will provide one.
I do not like overly long degenerate names for no good reason.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,292 @@
...
+storable_function<S, signature<ReturnType, Args...>>::~storable_function() noexcept
+{
+    m_vtable.destroy(*this);
+}
+
+
+template <typename S, typename ReturnType, typename... Args>
+ReturnType storable_function<S, signature<ReturnType, Args...>>::operator()(Args... args)
+{
+    auto r = m_function(std::forward<Args>(args)...);
```
elfenpiff, line 117: This is not perfect forwarding and here you create under some circumstances an additional move. A simple example would be:
```cpp
// copy but not movable
class B { 
   public:
    B() = default;
    B(const B &) = default;
    B(B &&) = delete;
    B &operator=(const B &) = default;
    B &operator=(B &&) = delete;
};

// function with argument as copy
void someFunction(B blubb) { ... }

int main() {
  storable_function<void(B)> func(someFunction);
  B bla;
  someFunction(bla); // works fine
  // does not compile since bla is not movable since std::forward<Args>(args)... enforces a move
  func(bla);
}
```

The solution is to implement perfect forwarding with variadic template arguments in the method and not in the class.
```cpp
template <typename S, typename ReturnType, typename... Args>
￼template <typename... FuncArgs>
ReturnType storable_function<S, signature<ReturnType, Args...>>::operator()(FuncArgs... args)
￼{
￼    auto r = m_function(std::forward<FuncArgs>(args)...);
}
```

But it would not be required to verify that `FuncArgs` and `Args` are the same since it would not compile if this wouldn't be the case.

elfenpiff (changes requested):
Only partially reviewed I will continue on monday.

iceoryx_utils/include/iceoryx_utils/cxx/type_traits.hpp
```diff
@@ -65,6 +89,20 @@ struct has_signature<Callable,
 {
 };
 
+///
+/// @brief Check whether T is a function pointer with arbitrary signature
+///
+template <typename T>
+struct is_function_pointer
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, I was following the style already present or forgot to put much thought in it because the real challenge is elsewhere.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,165 @@
...
+{
+namespace cxx
+{
+template <typename ReturnType, typename... Args>
+using signature = ReturnType(Args...);
+
+template <typename StorageType, typename T>
+class storable_function;
+
+/// @brief A storable alternative of std::function which not necessarily uses dynamic memory.
```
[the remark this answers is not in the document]
MatthiasKillat: Yes it does, StorageType can essentially define how it will store the underlying object. And it can use dynamic memory...if StorageType does. It is very flexible.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,165 @@
...
+    storable_function(storable_function&& other) noexcept;
+
+    storable_function& operator=(const storable_function& rhs) noexcept;
+
+    storable_function& operator=(storable_function&& rhs) noexcept;
+
+    ~storable_function() noexcept;
+
+    /// @brief invoke the stored function
+    /// @note  invoking the function if there is no stored function (i.e. operator bool returns false)
```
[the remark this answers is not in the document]
MatthiasKillat: Completely against changing the signature to returning ``expected<ReturnType`` for efficiency and semantic reasons (it is not what the user wanted, otherwise he would have said so).

Now, we can check the condition internally, yes. I can live with this but I consider the STL approach (which I am following here) superior (same goes for ``function_ref`` but I did not implement this).

I will put an assert like construct here.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,165 @@
...
+    ///        is undefined
+    ReturnType operator()(Args... args);
```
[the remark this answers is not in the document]
MatthiasKillat: Exactly, this is the problem in general. It is a false guarantee but we then just terminate. Not ideal but I have no ad hoc solution for this.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
MatthiasKillat: Need to think about the best approach here. Likely needs the additional variadic template parameter.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
elfenpiff: Maybe we can restrict it to noexcept callables via a user given template parameter. So that we in our use case can state that we only would like to have noexcept callables but in general the user is free to do whatever they want.

iceoryx_utils/include/iceoryx_utils/internal/cxx/static_storage.hpp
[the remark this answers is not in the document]
MatthiasKillat: I think this abstraction here is lower level, it solves a different problem: creating a simple abstraction of an (algigned) memory buffer whose memory can be created in different ways. Here it is statically created on the stack, but variants for dynamic memory or hybrids are conceivable. The purpose is o make the storage type of the function itself interchangable, i.e. decouple how it works from how it is stored.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
MatthiasKillat: I think it is not completely possible to check at compile time whether something can throw. I also would not like to really restrict it. Mainly because if it throws, then only because the callable you are using it with throws and the user can influence this (by providing a non throwing callable).

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
MatthiasKillat: Uses Expects now, leading to defined termination.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
MatthiasKillat: Whether something can throw cannot be checked, it is not part of the signature and exceptions may originate from code the compiler does not even know when compiling the current compilation unit(other compilation unit or library, the latter being bad practice). Therefore we cannot have this. 
We can verify noexcept (which is not the same as not throwing) with the operator but I do not quite like limiting the callables. Restricting the callables to noexcept is in effect the same as declaring the operator itself noexcept. Should something be thrown, it will terminate.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
MatthiasKillat: The example you provided is wrongly parameterized `storable_function` is missing its storage type template parameter, but that is not the issue. Adding variadic arguments works and leads to perfect forwarding in `function`, but `function_ref` may introduce unwanted copies (in the function pointer case).

I am not entirely happy with the change for this reason, need to look closer at the function pointer case (maybe the ctor in `function_ref` can be avoided (might be not possible without C++17, see my comments regarding this).

mossmaurice:
@MatthiasKillat The last two commits are missing the `#` in the commit message and GitHub does not link them. Maybe you can rebase and force push?

dkroenke (commented):
> @MatthiasKillat The last two commits are missing the `#` in the commit message and GitHub does not link them. Maybe you can rebase and force push?

Did someone said commit-hooks? https://github.com/eclipse-iceoryx/iceoryx/pull/799

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,264 @@
...
+        // erase the functor type and store as reference to the call in storage
+        m_function = *p;
+        m_storedObj = p;
+        m_vtable.copyFunction = copy<StoredType>;
+        m_vtable.moveFunction = move<StoredType>;
+        m_vtable.destroyFunction = destroy<StoredType>;
+    }
+
+    // else we detect the problem at compile time or store nothing when memory is exhausted
+    // note that we have no other choice, it is used in the ctor and we cannot throw
```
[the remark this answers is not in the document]
MatthiasKillat: Can do. As noted below, for the usage with static_storage (as defined in function.hpp which is the version we will be using) this will never be a problem (it will detect that there is not enough storage at compile time and fail to compile).

Of course, there always is the possibility of implementation error but I tried locally to use it with callables that are too large and it fails to compile (we cannot test this properly with unit tests).

iceoryx_utils/include/iceoryx_utils/internal/cxx/static_storage.inl
```diff
@@ -0,0 +1,97 @@
...
+    static_assert(fits_statically<T>(), "type does not fit into static storage");
+    return reinterpret_cast<T*>(allocate(alignof(T), sizeof(T)));
+}
+
+template <uint64_t Capacity, uint64_t Align>
+void* static_storage<Capacity, Align>::allocate(uint64_t align, uint64_t size) noexcept
+{
+    if (m_ptr)
+    {
+        return nullptr; // cannot allocate, already in use
```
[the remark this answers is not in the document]
MatthiasKillat: i would opt against this, this comment is just for developers and it is common to return `nullptr` if allocation fails (or throw an exception which we do not want).  I would not output warnings for things like this. The way we use it, this should not happen since the storage is cleared when we want to store a different callable.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
```diff
@@ -71,22 +71,49 @@ struct is_invocable
...
+template <typename T>
+struct is_function_pointer : std::false_type
+{
+};
+template <typename ReturnType, typename... ArgTypes>
+struct is_function_pointer<ReturnType (*)(ArgTypes...)> : std::true_type
 {
 };
 
+///
```
MatthiasKillat, line 111: Was probably removed and reintroduced by the automatic merge, can remove again. I do not see much point in having this since we can always negate the result of `is_same`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.inl
```diff
@@ -0,0 +1,102 @@
...
+{
+namespace cxx
+{
+template <uint64_t Capacity, uint64_t Align>
+constexpr uint64_t static_storage<Capacity, Align>::align_mismatch(uint64_t align, uint64_t requiredAlign) noexcept
+{
+    auto r = align % requiredAlign;
+
+    // if r != 0 we are not aligned and need to add this amount to an align
+    // aligned address to be aligned with alignTarget
```
MatthiasKillat, line 32: requiredAlign

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,343 @@
...
+    return (required_storage_size<T>() <= S::capacity()) && is_invocable_r<ReturnType, T, Args...>::value;
+}
+
+template <typename S, typename ReturnType, typename... Args>
+void storable_function<S, signature<ReturnType, Args...>>::operations::copy(const storable_function& src,
+                                                                            storable_function& dest) noexcept
+{
+    if (copyFunction)
+    {
+        std::cout << "copyFunction" << std::endl;
```
mossmaurice, line 316: Left-over?

mossmaurice (commented):
8/12 files reviewed

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: Yes, will remove.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    cxx::Ensures(!dest.empty());
+}
+
+template <typename S, typename ReturnType, typename... Args>
+template <typename CallableType>
+void storable_function<S, signature<ReturnType, Args...>>::destroy(storable_function& f) noexcept
+{
+    // TODO: this is not an indication allocate was called anymore (due to fptrs also being stored in m_callable)
```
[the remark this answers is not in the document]
MatthiasKillat: I will solve this in this PR.

iceoryx_hoofs/test/moduletests/test_static_storage.cpp
```diff
@@ -0,0 +1,209 @@
...
+    void SetUp() override
+    {
+    }
+
+    void TearDown() override
+    {
+    }
+};
+
+TEST(static_storage_test, IsAllocatableWithoutAlignmentRestriction)
```
elfenpiff, line 70: Please add a test for `capacity`.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -0,0 +1,627 @@
...
+    Functor::resetCounts();
+
+    {
+        test_function sut(f);
+    }
+
+    EXPECT_EQ(Functor::numDestroyed, 1U);
+}
+
+TEST_F(function_test, CopyCtorCopiesStoredFunctor)
```
elfenpiff, line 293: Could you please add to all successful copy tests that the source and the destination can still call the stored callable.
For all successful move tests could you please verify that the destination can call the stored callable.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
```diff
@@ -71,22 +71,49 @@ struct is_invocable
...
-/// @brief Verfies the signature ReturnType(ArgTypes...) of the provided Callable type
+/// @brief Verifies whether the passed Callable type is in fact invocable with the given arguments
+///        and the result of the invocation is convertible to ReturnType.
+///
+/// @note This is an implementation of std::is_invokable_r (C++17).
 ///
-template <typename Callable = void, typename ReturnType = void, typename ArgTypes = void>
-struct has_signature : std::false_type
+template <typename ReturnType, typename Callable, typename... ArgTypes>
+struct is_invocable_r
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, I know. I chose the style of the already existing `is_invocable` ...

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.inl
```diff
@@ -0,0 +1,102 @@
...
+template <uint64_t Capacity, uint64_t Align>
+void static_storage<Capacity, Align>::deallocate() noexcept
+{
+    m_ptr = nullptr;
+}
+
+template <uint64_t Capacity, uint64_t Align>
+void static_storage<Capacity, Align>::clear() noexcept
+{
+    std::memset(m_bytes, 0, Capacity);
```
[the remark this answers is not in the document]
MatthiasKillat: Currently this is not even used (except to test something, IIRC) and it is useful. If we want to check against this, I would prefer the function to return a bool indicating wether it succeeded or failed (instead of terminate via Expects) . It will fail if currently an object exists,  allocate without deallocate was called.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.hpp
```diff
@@ -0,0 +1,101 @@
...
+    /// @brief set the managed static memory to all zeros
+    void clear() noexcept;
+
+    /// @brief get the storage capacity in bytes
+    /// @return maximum number of bytes available in the static storage
+    /// @note this is an upper bound for the object size it can store
+    static constexpr uint64_t capacity() noexcept;
+
+  private:
+    alignas(Align) uint8_t m_bytes[Capacity];
```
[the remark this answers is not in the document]
MatthiasKillat: No, `m_ptr` has another purpose as well, it does not necessarily point to first byte in `m_bytes` if the stored object is aligned differently. Therefore we need more than a bool. I would also not try to fit it in the storage area `m_bytes` somehow, this is very error-prone and not worth it.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.hpp
```diff
@@ -0,0 +1,101 @@
...
+/// @note We can define optimized_storage (or dynamic_storage) with a similar interface
+///       but other allocation policies and use them where we need to store objects
+///       with some interchangable storage policy (e.g. in storable_function)
+///       optimized_storage would have a dynamic memory fallback when static memory is
+///       insufficent.
+template <uint64_t Capacity, uint64_t Align = 1>
+class static_storage
+{
+  public:
+    static_storage() noexcept = default;
```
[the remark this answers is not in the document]
MatthiasKillat: Likely true, have not checked. Will take a look and yes, upgrading to constexpr what is possible might be useful but mainly if we use this more except as an implementation detail of `cxx::function` as of now.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
MatthiasKillat: Yes.

iceoryx_hoofs/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -57,20 +57,6 @@ TEST(TypeTraitsTest, IsInvocableResolvesToFalse)
     EXPECT_FALSE(sut);
 }
```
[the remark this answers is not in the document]
MatthiasKillat: Will be added. We could also remove this if we do not want the changes in function_ref.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -0,0 +1,343 @@
...
+ReturnType storable_function<S, signature<ReturnType, Args...>>::invokeFreeFunction(void* callable, Args&&... args)
+{
+    return (reinterpret_cast<ReturnType (*)(Args...)>(callable))(std::forward<Args>(args)...);
+}
+
+template <typename S, typename ReturnType, typename... Args>
+template <typename T>
+constexpr uint64_t storable_function<S, signature<ReturnType, Args...>>::required_storage_size() noexcept
+{
+    return sizeof(T) + alignof(T);
```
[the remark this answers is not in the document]
MatthiasKillat: First of all: I could save one 1 byte here and will do this, the pessimistic estimate is size + alignment - 1. 

The value the function here returns is supposed to be sufficient, not best possible (and it cannot be at compile time I think). Maybe we can do a little better, but I am not sure. `function` contains a `static_storage`, which is aligned somehow (Align).  This information can be used to improve on the estimate but it is a little more complicated since I think. Currently Align is set to 1 by default and used this way in function. For this, the estimate above is still best possible.

But `storable_function` has no knowledge about this alignment (and should not have, since the StorageType should be configurable).
TLDR: `storable_function` does not know `static_storage`, it is more generic.

iceoryx_hoofs/test/moduletests/test_static_storage.cpp
MatthiasKillat: Sure.

iceoryx_utils/include/iceoryx_utils/internal/cxx/storable_function.inl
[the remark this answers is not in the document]
MatthiasKillat: Deleted comment, added note to the function and ctor describing the behavior. Note tha in the copy ctor/assignment this must will always work with static_storage (since it worked for the original).

iceoryx_posh/include/iceoryx_posh/internal/popo/publisher.inl
```diff
@@ -46,7 +46,7 @@ inline cxx::expected<AllocationError> PublisherImpl<T, H, BasePublisher_t>::publ
     static_assert(cxx::is_invocable<Callable, T*, ArgTypes...>::value,
                   "Publisher<T>::publishResultOf expects a valid callable with a specific signature as the "
                   "first argument");
-    static_assert(cxx::has_signature<Callable, void(T*, ArgTypes...)>::value,
+    static_assert(cxx::is_invocable_r<void, Callable, T*, ArgTypes...>::value,
```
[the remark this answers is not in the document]
MatthiasKillat: Well, I do not like the the syntax and naming of this STL construct, but it is usable and useful.

iceoryx_hoofs/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -57,20 +57,6 @@ TEST(TypeTraitsTest, IsInvocableResolvesToFalse)
...
 
-TEST(TypeTraitsTest, HasSignatureResolvesToTrue)
```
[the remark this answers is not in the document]
MatthiasKillat: Mistake.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/function_ref.hpp
```diff
@@ -87,10 +87,17 @@ class function_ref<ReturnType(ArgTypes...)>
     /// @brief Creates a function_ref with a callable whose lifetime has to be longer than function_ref
     /// @param[in] callable that is not a function_ref
     template <typename CallableType,
-              typename = std::enable_if_t<!is_function_ref<std::remove_reference_t<CallableType>>::value>,
-              typename = std::enable_if_t<is_invocable<CallableType, ArgTypes...>::value>>
+              typename = std::enable_if_t<!is_function_pointer<CallableType>::value
+                                          && not_same<CallableType, function_ref>::value
```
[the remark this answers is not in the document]
MatthiasKillat: They do not, a function_ref is not a function pointer. I do not think the second conition can be removed without changing the meaning. Also the function_ref test was there before my modification, I just needed to exclude function pointers here because the ctor was not working correctly with them and by excluding them overload resolution is guided to the explicit ctor for function pointers (it would not without this).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+    storable_function& operator=(storable_function&& rhs) noexcept;
+
+    ~storable_function() noexcept;
+
+    /// @brief invoke the stored function
+    /// @param args arguments to invoke the stored function with
+    /// @return return value of the stored function
+    /// @note  Invoking the function if there is no stored function (i.e. operator bool returns false)
+    ///        leads to terminate being called.
```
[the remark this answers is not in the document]
MatthiasKillat: That is not how it is currently though, as far as I know. Therefore the comment should reflect this and will stay this way.

I prefer interface behavior description to be precise, may is not conductive to this (would be better to be able to state the conditions under which it will terminate).

It may work, or maybe not. Maybe it also blows up your computer...who knows. An exciting function to be sure, but may terminate is not a useful claim here.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,199 @@
...
+    /// @note Deliberately not noexcept but can only throw if the stored callable can throw an exception (hence
```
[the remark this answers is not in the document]
MatthiasKillat: I want to leave it that way as it is fairly generic and I prefer giving users the ability to wrap throwing callables. For us it makes no difference if we only ever wrap non-throwing callables (or even a throwing one and never catch). Hence I want to leave it this way for now. Should we see this causing problems (I do not see why it would) we can easily change this.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+    /// @brief indicates whether a function is currently stored
+    /// @return true if a function is stored, false otherwise
+    operator bool() noexcept;
+
+    /// @brief swap this with another storable function
+    void swap(storable_function& f) noexcept;
+
+    /// @brief swap two storable functions
+    static void swap(storable_function& f, storable_function& g) noexcept;
```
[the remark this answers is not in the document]
MatthiasKillat: Defining it outside in a useful manner given the templates may be problematic. I give it a look.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -0,0 +1,199 @@
...
+    // Therefore we cannot store a storable_function in shared memory (the same holds for std::function).
+    // This is inherent to the type erasure technique we (have to) use.
+    struct operations
+    {
+        // function pointers defining copy, move and destroy semantics
+        void (*copyFunction)(const storable_function& src, storable_function& dest){nullptr};
+        void (*moveFunction)(storable_function& src, storable_function& dest){nullptr};
+        void (*destroyFunction)(storable_function& f){nullptr};
+
+        operations() = default;
```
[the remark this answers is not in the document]
MatthiasKillat: Can be, but it is internal and just shuffling around function pointers. Way to much boilerplate.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
MatthiasKillat: Defined out of class as per STL style (out of class template definition delegates to member swap implementation).

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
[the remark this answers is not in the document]
MatthiasKillat: I will leave this as is, it is correct as far as I can see and the code style should be ok as well. We can think about refactoring the type traits in a separate issue (maybe adding more tests for some as well, though this one is tested).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.hpp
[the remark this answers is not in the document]
MatthiasKillat: Was possible, changed all methods to constexpr (except dtor, which can never be constexpr).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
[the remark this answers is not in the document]
MatthiasKillat: Looking at this we are safe here since we are always free to call deallocate, even if we did not call allocate (it is like freeing a nullptr, which is a no-op). This is true for the `static_storage` but must be true with other storage types as well (i.e. behave like freeing a `nullptr` in this regard). In other words, there is nothing to do except to remove the todo, no other variable has to be introduced (and this is good since checking lots of flags if not needed should be avoided).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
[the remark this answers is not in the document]
MatthiasKillat: I introduced optimal allocation size calculation based on alignment, but at the price of adding a new requirement to the storage type of providing an `allocation_size` function (due to the reasons above).

iceoryx_hoofs/include/iceoryx_hoofs/cxx/function_ref.hpp
```diff
@@ -87,10 +93,19 @@ class function_ref<ReturnType(ArgTypes...)>
...
+                                          && is_invocable<CallableType, ArgTypes...>::value>>
     function_ref(CallableType&& callable) noexcept;
 
+    /// @brief Creates a function_ref from a function pointer
+    /// @param[in] function function pointer to function we want to reference
+    ///
+    /// @note This overload is needed, as the general implementation
+    /// will not work properly for function pointers.
+    /// This ctor is not needed anymore once we can use user-defined-deduction guides (C++17)
+    function_ref(ReturnType (*function)(ArgTypes...));
```
elfenpiff, line 107: Please add `noexcept`

elfenpiff (commented):
Looks good! I hope we can merge it til the end of the week.

MatthiasKillat:
The pipeline reports a problem with `alignof` that is used in a non-standard way - applied to expressions (gcc allows this). I have changed this to be semantically equivalent for our case..

MatthiasKillat:
@elfenpiff @mossmaurice I think I addressed all the points and the CI is green.

I think it is mature enough to be tried out internally. Should there be any further issues I will find them when I replace internal `std::function` occurrences next. usage should be fine, within the limitations of the capacity constraint and behavior of `operator()` which should be as `std::function`.

mossmaurice:
> @elfenpiff @mossmaurice I think I addressed all the points and the CI is green.
> 
> It is by no means perfect, but I think it is mature enough to be tried out internally. Should there be any further issues I will find them when I replace internal `std::function` occurrences next. Usage should be fine (according to tests), within the limitations of the capacity constraint and behavior of `operator()` which should be as `std::function`.

:+1: I'll finish off the review tonight.

Could you merge `origin/master` to this branch? There are still some conflicts.

elfenpiff (approved):

mossmaurice (changes requested):
10/13 done

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.hpp
```diff
@@ -0,0 +1,108 @@
...
+    /// @return pointer to memory where a T can be constructed if memory is available, nullptr otherwise
+    /// @note  compilation fails if storage is insufficient for objects of type T
+    template <typename T>
+    constexpr T* allocate() noexcept;
+
+    /// @brief request aligned memory with a specific size in bytes
+    /// @param align alignment of the memory requested
+    /// @param size number of bytes of the memory requested
+    /// @return pointer to aligned memory of the requested size if available, nullptr otherwise
+    constexpr void* allocate(const uint64_t align, const uint64_t size) noexcept;
```
mossmaurice, line 75: You could think about renaming this method to `aligned_allocate`.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -0,0 +1,630 @@
...
+    // (no dependency or side effects)
+    test_function sut(functor);
+
+    ASSERT_TRUE(sut.operator bool());
+
+    // both increment their state independently
+    EXPECT_EQ(sut(1U), functor(1U));
+
+    // clear original (set to 0)
+    p->~Functor();
```
mossmaurice, line 272: Please don't call the d'tor directly in tests. Either use an `cxx::optional` or a `std::unique_ptr` to wrap the object.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -0,0 +1,630 @@
...
+
+    swap(sut1, sut2);
+
+    ASSERT_TRUE(sut1.operator bool());
+    EXPECT_EQ(sut1(1), f2(1));
+    ASSERT_TRUE(sut2.operator bool());
+    EXPECT_EQ(sut2(1), f1(1));
+}
+
+TEST_F(function_test, functorOfSizeSmallerThanStorageBytesCanBeStored)
```
mossmaurice, line 493: Nitpick: From here on all the test cases names are starting with a lowercase letter

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -0,0 +1,630 @@
...
+    {
+    }
+
+    static int32_t staticFunction(int32_t n)
+    {
+        return n + 1;
+    }
+};
+
+TEST_F(function_test, DefaultConstructionCreatesNoCallable)
```
mossmaurice, line 174: What about a 2nd test cases that stores a `nullptr`?

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -0,0 +1,630 @@
...
+{
+    test_function sut;
+
+    EXPECT_FALSE(sut.operator bool());
+}
+
```
mossmaurice, line 180: You could add a death that verifies the termination behaviour in case an empty `function` is invoked.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -0,0 +1,630 @@
...
+    auto lambda = [](Arg& a1, const Arg& a2, Arg&& a3, Arg a4) { return a1.value + a2.value + a3.value + a4.value; };
+    iox::cxx::function<int32_t(Arg&, const Arg&, Arg&&, Arg), 128> sut(lambda);
+
+    ASSERT_TRUE(sut.operator bool());
+    auto result = sut(arg1, arg2, std::move(arg3), arg4);
+
+    EXPECT_EQ(result, sum);
+}
+
```
mossmaurice, line 629: You could add some tests to see if SFINAE is working properly, like [those tests](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_hoofs/test/moduletests/test_cxx_function_ref.cpp#L304-L320).

mossmaurice (commented):
LGTM in general, only minor findings. Let's get this merged!

mossmaurice:
Another thing that just popped up: As this PR introduced a lot of new code, could you run Axivion with [these coding rules](https://github.com/eclipse-iceoryx/iceoryx/blob/master/tools/axivion/axivion_config.json) over this PR?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/function_ref.hpp
```diff
@@ -24,6 +24,7 @@
 #include <cstddef>
 #include <iostream>
 #include <memory>
+#include <type_traits>
```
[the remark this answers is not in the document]
MatthiasKillat: This is bad style though, because if this changes, I get a compile error here. We should not rely on indirect includes for this reason. Rejected.

iceoryx_hoofs/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -125,4 +125,29 @@ TEST(TypeTraitsTest, AddConstConditionallyTypeAliasWorks)
...
+TEST(TypeTraitsTest, IsFunctionPointerResolvesToTrue)
+{
+    auto result = is_function_pointer<void (*)(double)>::value;
+    EXPECT_TRUE(result);
+
+    result = is_function_pointer<int* (*)(double)>::value;
+    EXPECT_TRUE(result);
+
+    result = is_function_pointer<void (*)(int, double)>::value;
+    EXPECT_TRUE(result);
```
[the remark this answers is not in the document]
MatthiasKillat: That is very unreasonable here. Small is only one side of the coin, it should be part of a useful test category/group. They are all positive cases, and the test is very reasonable to read, there is no interference between the checks and so on. Each test started incurs overhead as well, especially with fixtures. So we should not break everything down just for the sake of it.

Rejected.

iceoryx_hoofs/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -125,4 +125,29 @@ TEST(TypeTraitsTest, AddConstConditionallyTypeAliasWorks)
...
+TEST(TypeTraitsTest, IsFunctionPointerResolvesToFalse)
+{
+    auto result = is_function_pointer<int*>::value;
+    EXPECT_FALSE(result);
+
+    result = is_function_pointer<void*>::value;
+    EXPECT_FALSE(result);
+
+    result = is_function_pointer<int>::value;
+    EXPECT_FALSE(result);
```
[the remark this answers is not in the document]
MatthiasKillat: Rejected all the same, as in the positive case. If I follow other examples somewhere else I would only have one negative instance, but I consider this insufficient. I will not spread those multiple negative instances across multiple tests.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.inl
```diff
@@ -0,0 +1,114 @@
...
+    // Capacity >= sizeof(T) + alignof(T) - 1
+    return allocation_size<T>() <= Capacity;
+}
+
+template <uint64_t Capacity, uint64_t Align>
+template <typename T>
+constexpr T* static_storage<Capacity, Align>::allocate() noexcept
+{
+    static_assert(is_allocatable<T>(), "type does not fit into static storage");
+    return reinterpret_cast<T*>(allocate(alignof(T), sizeof(T)));
```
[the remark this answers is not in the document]
MatthiasKillat: I consider 'static_cast' from or to `void*` bad style, but can change it. Is semantically equivalent here, but does not raise the necessary awareness that we are dealing with erased/opaque types.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/static_storage.hpp
MatthiasKillat: No, the parameter reflects this and I do not need the name otherwise. This is not C where we cannot properly overload etc. Function names should be seen combined with the parameters to form something like a "sentence" and redundancy is not required then, if we use proper parameter names (formal and actual) .

iceoryx_hoofs/test/moduletests/test_cxx_function_ref.cpp
```diff
@@ -318,4 +323,45 @@ TEST_F(function_refTest, CallOverloadedFunctionResultsInCallOfIntInt)
...
+    EXPECT_EQ(result, freeFuncTestValue);
+}
+
+TEST_F(function_refTest, CreationWithFunctionPointerWithRefArgWorks)
+{
+    constexpr auto fp = &freeVoidFunction;
+    function_ref<void(int&)> sut(fp);
+
+    ASSERT_TRUE(sut.operator bool());
+    int arg{0};
```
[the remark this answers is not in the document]
MatthiasKillat: No it cannot, the sut takes a `int&`.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
MatthiasKillat: This is needed as it was constructed using placement new and is the way to go. This is not a use case for an `optional`. We cannot really leak anything here (if we leave via ASSERT), but we can prevent the missed call with a `unique_ptr'` acting as guard.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
MatthiasKillat: No, should we decide to have Expects not terminate the test will be invalid. I would not make this terminate a requirement with such a test.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
MatthiasKillat: Why? SFINAE is an implementation detail (compile time technique to switch between implementations), not a requirement of the behavior of the object.

Besides, I do not quite see the point of the tests you linked. They test whether I can pass some callable to a function which takes a `function_ref` (of compatible signature) by value. Internally the `function_ref` is then called.
Would the signature not be compatible, it would not compile.

This is equivalent to just pass the callable to the function_ref itself and callng it, no indirection with an additional function is needed nor useful. The equivalent case mentioned is already tested for `cxx::function` by testing the ctor itself.

TLDR: the tests add nothing since the ctor is already tested so I will add no such test.

MatthiasKillat:
No, in my opinion we will introduce Axivion for the whole code base and from then on we will run it for every PR. I have no license for it yet and do not see that we should block the PR due to this. I will fix any problem occurring due to this in a new PR, but we should introduce Axivion properly before this is done.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
MatthiasKillat: Added.

elfenpiff (dismissed):

MatthiasKillat:
Now it should be ok if the CI builds, wanted to get rid off one merge commit at least hence the force pushes.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -34,11 +34,14 @@ storable_function<S, signature<ReturnType, Args...>>::storable_function(const Fu
 template <typename S, typename ReturnType, typename... Args>
 storable_function<S, signature<ReturnType, Args...>>::storable_function(ReturnType (*function)(Args...)) noexcept
 {
-    m_invoker = invokeFreeFunction;
-    m_callable = reinterpret_cast<void*>(function);
-    m_operations.copyFunction = copyFreeFunction;
-    m_operations.moveFunction = moveFreeFunction;
-    // destroy is not needed for free functions
+    if (function)
```
MatthiasKillat, line 37: Was needed since it otherwise tries to invoke a nullptr. This way if we construct it from a nullptr it is not callable and hence correct (the variables we do not set directly are initialized with nullptr).

mossmaurice:
> I will fix any problem occurring due to this in a new PR, but we should introduce Axivion properly before this is done.

Could you create an issue for that? Axivion was introduced in #409 for the `master` branch, but currently lacks a script for the local scan.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
mossmaurice: What I meant is, to add some tests to check if the overload resolution works as expected. I remember having missed the tests in the first attempt of `function_ref` and when using it I had to narrow down possible set during the overload resolution via type traits.

mossmaurice (approved):

MatthiasKillat:
Before we create any issue we should wait whether anything pops up at all.

MatthiasKillat:
@elfenpiff @mossmaurice should be ready to merge

dkroenke (approved):

Issue #337, opened 2020-11-05
Deprecate posix timer


marthtz:
## Brief feature description

Deprecate posix timer from iceoryx code base.

## Detailed information

Posix timer has several issues (concurrency, truncation) and is pretty heavy weight and a bit slow. As iceorxy needs to be very responsive the posix timer should be removed and replaced with a lightweight stopwatch.

This will also close #190 as it'll be obsolete.

Tasks
- [ ] create `DeadlineTimer` or similar and replace this part of the `PosixTimer` where applicable
- [ ] create a small wrapper around a std::thread which can be used for the keep alive message
    - replace the keep alive message from a `std::chrono` timestamp with a `units::Duration` from a `timespec` initialized with `clock_gettime` with `CLOCK_REALTIME`

elfenpiff:
@marthtz I think a stopwatch is not exactly what we need. We have to rewrite the timer so that the timer does not work with callbacks anymore but instead with signals. If you look at the manpage https://man7.org/linux/man-pages/man2/timer_create.2.html we have to possible approaches:
- send a signal to the whole process
- send a signal to a specific thread specified when creating the timer

I think we should explore the second option. Maybe it has similar drawbacks than the callback approach but if not it should be our way to go.

@elBoberido what do you think?

elBoberido:
@elfenpiff @marthtz please don't call it stopwatch. A stopwatch is measuring time, this functionality is running towards a deadline or countdown. See also the equivalent int boost and Qt.

To the other points. The posix timer was started as a deadline timer and then enhanced with the posix calls. Currently it has both functionalities.
1. cut out the deadline functionality and use that class where appropriate
2. deprecate the current posix timer and replace the remaining occurrences with `std::thread` and `sleep_for`
3. decide if we really need something like the posix timer. If we do, implement the second option from @elfenpiff which is also described in more detail in #170

shankar-bosch:
@elfenpiff I am currently working on this story. Based on the comments from @dkroenke I am working on to replace the timer.start() function with a simple thread sleep. Currently the posixtimer.start() is being called within the posh_runtime, service_discovery_notifier.

elBoberido:
@shankar-bosch a simple `sleep_for` is not enough. The current timer invokes a thread on each timeout. You also have a start a thread where you use the `sleep_for`.

shankar-bosch:
@elBoberido okay I will do so

elfenpiff:
[...]

Issue #190, opened 2020-07-13
nanoseconds in units::Duration are truncated


dkroenke:
## Required information

**Operating system:**
on aarch64 targets like Raspberry PI

**Compiler version:**
E.g. GCC 9.3.0

**Observed result or behaviour:**
The Posix Timer is getting double values on the API which are then converted into int64_t internally. This conversion leads to a loss of data because the compiler will truncate the value. This can lead to abnormal behaviour when using small numbers like 1 nanoseconds.

**Expected result or behaviour:**
Remove the double literals from the posix timer API and expect only integer als time values. Then we avoid this constraint and have a clean calculation.

**Conditions where it occurred / Performed steps:**
Try to start a timer with 1 nanosecond waiting time. It will no execute because the time is internally set to zero. This disarms the timer and it is not proceeding.

budrus:
I can confirm. But first lost some time my Pi before finding this issue...

elBoberido:
@dkroenke I think this related to the underlying `long double` of `units::Duration`. Should we rephrase this bug to fix the problem on its source?

@elfenpiff suggested in #337 to use
```c++
struct duration_t {
     uint64_t seconds;
     uint32_t nanoseconds;
};
```
as base instead of `long double`

PR #490, opened 2021-01-13
Iox #489 replace introspection threads with periodic task

### LINKED ISSUE #489: Replace introspection threads with PeriodicTask
## Brief feature description

Use the `PeriodicTask` to remove the thread from the introspection classes.

## Detailed information

Currently the introspection classes spawn an own thread to periodically update their topics. The shutdown of the threads is done by polling a member every 100ms in the port and process introspection and by a condition variable in the mempool introspection.
By using the `PeriodicTask`, this gets unified and should also speed up the tests where the RouDiEnvironment is used.

Refactor
- [ ] process introspection
- [ ] port introspection
- [ ] mempool introspection

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
This PR replaces the thread in the introspection classes with a `PeriodicTask` and simplifies the code. The test time with the RouDiEnvironment will also be reduced a little bit (1s posh_modultetests and 2s in posh_integrationtests), since the stopping of the threads is not done by a polling with 100ms period.

1. [x] Merge

## References

- Closes #489 and #494

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
@@ -73,13 +73,13 @@ class MemPoolIntrospection
     MemoryManager* m_rouDiInternalMemoryManager{nullptr}; // mempool handler needs to outlive this class (!)
     SegmentManager* m_segmentManager{nullptr};
     PublisherPort m_publisherPort{nullptr};
+    void send() noexcept;
 
   private:
     static void prepareIntrospectionSample(MemPoolIntrospectionInfo& sample,
                                            const posix::PosixGroup& readerGroup,
                                            const posix::PosixGroup& writerGroup,
                                            uint32_t id) noexcept;
-    void send() noexcept;
 
     /// @brief copy data fro internal struct into interface struct
     void copyMemPoolInfo(const MemoryManager& memoryManager, MemPoolInfoContainer& dest) noexcept;
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
@@ -20,14 +20,10 @@
 #include "iceoryx_posh/internal/roudi/roudi_process.hpp"
 #include "iceoryx_posh/mepoo/mepoo_config.hpp"
 #include "iceoryx_posh/roudi/introspection_types.hpp"
+#include "iceoryx_utils/cxx/method_callback.hpp"
+#include "iceoryx_utils/internal/concurrent/periodic_task.hpp"
 
-#include <chrono>
-#include <condition_variable>
 #include <cstdint>
-#include <iostream>
-#include <mutex>
-#include <string>
-#include <thread>
 
 namespace iox
 {
@@ -41,14 +37,6 @@ namespace roudi
 template <typename MemoryManager, typename SegmentManager, typename PublisherPort>
 class MemPoolIntrospection
 {
-  private:
-    enum RunLevel
-    {
-        RUN,
-        WAIT,
-        TERMINATE
-    };
-
   public:
     /// @brief The constructor for the MemPoolIntrospection.
     ///        It starts a thread and set it into a wait condition.
@@ -68,24 +56,18 @@ class MemPoolIntrospection
     MemPoolIntrospection(MemPoolIntrospection&&) = delete;
     MemPoolIntrospection& operator=(MemPoolIntrospection&&) = delete;
 
-    /// @brief This function sets the thread into run mode, which periodically
-    ///        sends snapshots of the mempool introspecton data to the client.
-    ///        The send interval can be set by @ref setSnapshotInterval "setSnapshotInterval(...)".
-    void start() noexcept;
-
-    /// @brief This functions sets the thread into a wait state and
-    ///        the transmission of the introspection data is stopped.
-    void wait() noexcept;
+    /// @brief This function starts the periodic transmission of snapshots of the mempool introspecton data.
+    ///        The send interval can be set by @ref setSendInterval "setSendInterval(...)". By default it's 1 second.
+    void run() noexcept;
 
     /// @brief This function stops the thread which sends the introspection data.
     ///        It is not possible to start the thread again.
-    void terminate() noexcept;
+    void stop() noexcept;
 
     /// @brief This function configures the interval for the transmission of the
     ///        mempool introspection data.
-    ///
-    /// @param snapshotInterval_ms is the interval time in milliseconds
-    void setSnapshotInterval(unsigned int snapshotInterval_ms) noexcept;
+    /// @param[in] interval duration between two send invocations
+    void setSendInterval(const units::Duration interval) noexcept;
 
   protected:
     MemoryManager* m_rouDiInternalMemoryManager{nullptr}; // mempool handler needs to outlive this class (!)
@@ -93,21 +75,6 @@ class MemPoolIntrospection
     PublisherPort m_publisherPort{nullptr};
 
   private:
-    std::chrono::milliseconds m_snapShotInterval{1000};
-
-    std::atomic<RunLevel> m_runLevel{WAIT};
-    std::condition_variable m_waitConditionVar;
-    std::mutex m_mutex;
-    std::thread m_thread;
-
-    void run() noexcept;
-    void waitForRunLevelChange() noexcept;
-    void wakeUp(RunLevel newLevel = RUN) noexcept;
-
-    // wait until start command, run until wait or terminate, restart from wait
-    // is possible  terminate call leads to exit
-    void threadMain() noexcept;
-
     static void prepareIntrospectionSample(MemPoolIntrospectionInfo& sample,
                                            const posix::PosixGroup& readerGroup,
                                            const posix::PosixGroup& writerGroup,
@@ -116,6 +83,10 @@ class MemPoolIntrospection
 
     /// @brief copy data fro internal struct into interface struct
     void copyMemPoolInfo(const MemoryManager& memoryManager, MemPoolInfoContainer& dest) noexcept;
+
+  private:
+    units::Duration m_sendInterval{units::Duration::seconds<unsigned long long int>(1)};
+    concurrent::PeriodicTask<cxx::MethodCallback<void>> m_sender{"MemPoolIntr", *this, &MemPoolIntrospection::send};
 };
 
 /// @brief typedef for the templated mempool introspection class that is used by RouDi for the
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
@@ -87,7 +87,7 @@ class MemPoolIntrospection
   private:
     units::Duration m_sendInterval{units::Duration::seconds<unsigned long long int>(1)};
     concurrent::PeriodicTask<cxx::MethodCallback<void>> m_publishingTask{
-        "MemPoolIntr", *this, &MemPoolIntrospection::send};
+        concurrent::PeriodicTaskManualStart, "MemPoolIntr", *this, &MemPoolIntrospection::send};
 };
 
 /// @brief typedef for the templated mempool introspection class that is used by RouDi for the
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
@@ -86,7 +86,8 @@ class MemPoolIntrospection
 
   private:
     units::Duration m_sendInterval{units::Duration::seconds<unsigned long long int>(1)};
-    concurrent::PeriodicTask<cxx::MethodCallback<void>> m_sender{"MemPoolIntr", *this, &MemPoolIntrospection::send};
+    concurrent::PeriodicTask<cxx::MethodCallback<void>> m_publishingTask{
+        "MemPoolIntr", *this, &MemPoolIntrospection::send};
 };
 
 /// @brief typedef for the templated mempool introspection class that is used by RouDi for the
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2019, 2021 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
@@ -62,12 +62,12 @@ class PortIntrospection
 
         struct PublisherInfo
         {
-            PublisherInfo() = default;
+            PublisherInfo() noexcept = default;
 
             PublisherInfo(typename PublisherPort::MemberType_t* const portData,
                           const ProcessName_t& name,
                           const capro::ServiceDescription& service,
-                          const NodeName_t& node)
+                          const NodeName_t& node) noexcept
                 : portData(portData)
                 , name(name)
                 , service(service)
@@ -92,14 +92,12 @@ class PortIntrospection
 
         struct SubscriberInfo
         {
-            SubscriberInfo()
-            {
-            }
+            SubscriberInfo() noexcept = default;
 
             SubscriberInfo(typename SubscriberPort::MemberType_t* const portData,
                            const ProcessName_t& name,
                            const capro::ServiceDescription& service,
-                           const NodeName_t& node)
+                           const NodeName_t& node) noexcept
                 : portData(portData)
                 , name(name)
                 , service(service)
@@ -115,20 +113,18 @@ class PortIntrospection
 
         struct ConnectionInfo
         {
-            ConnectionInfo()
-            {
-            }
+            ConnectionInfo() noexcept = default;
 
             ConnectionInfo(typename SubscriberPort::MemberType_t* const portData,
                            const ProcessName_t& name,
                            const capro::ServiceDescription& service,
-                           const NodeName_t& node)
+                           const NodeName_t& node) noexcept
                 : subscriberInfo(portData, name, service, node)
                 , state(ConnectionState::DEFAULT)
             {
             }
 
-            ConnectionInfo(SubscriberInfo& subscriberInfo)
+            ConnectionInfo(SubscriberInfo& subscriberInfo) noexcept
                 : subscriberInfo(subscriberInfo)
                 , state(ConnectionState::DEFAULT)
             {
@@ -138,14 +134,14 @@ class PortIntrospection
             PublisherInfo* publisherInfo{nullptr};
             ConnectionState state{ConnectionState::DEFAULT};
 
-            bool isConnected()
+            bool isConnected() const noexcept
             {
                 return publisherInfo && state == ConnectionState::CONNECTED;
             }
         };
 
       public:
-        PortData();
+        PortData() noexcept;
 
         /// @brief add a publisher port to be tracked by introspection
         /// there cannot be multiple publisher ports with the same capro id
@@ -158,7 +154,7 @@ class PortIntrospection
         bool addPublisher(typename PublisherPort::MemberType_t* const port,
                           const ProcessName_t& name,
                           const capro::ServiceDescription& service,
-                          const NodeName_t& node);
+                          const NodeName_t& node) noexcept;
 
         /// @brief add a subscriber port to be tracked by introspection
         /// multiple subscribers with the same capro id are possible as long as the names are different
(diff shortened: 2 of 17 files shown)


iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
```diff
@@ -33,13 +33,20 @@ template <typename T>
 class PeriodicTask
 {
   public:
+    /// @brief Creates a periodic task. The specified callable is stored but not executed.
+    /// To run the task, `void start(const units::Duration interval)` must be called.
+    /// @param[in] taskName will be set as thread name
```
dkroenke, line 38: Additionally comment needed:
`/// @tparam <template-parameter-name>` for template parameters

iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
```diff
@@ -51,15 +58,32 @@ class PeriodicTask
     PeriodicTask& operator=(const PeriodicTask&) = delete;
     PeriodicTask& operator=(PeriodicTask&&) = delete;
 
+    /// @brief Spawns a thread and immediately executes the callable specified with the constructor.
+    /// The execution is repeated after the specified interval is passed.
+    /// @param[in] interval is the time the thread waits between two invocations of the callable
+    /// @attention If there is already a running thread, this will be stopped and started again with the new interval.
```
dkroenke, line 64: You mean a running thread with the same name right?

dkroenke (commented):
Two questions:
1. Could be the TimeSpecReference in the run() method maybe Monotonic? I think the timeout interval does not depend on the current value of time.

2. This is maybe for future use: For validation purpose the user maybe wants to know how often the task was executed. A getter method like `uint64_t getExecutionCount()` could be helpful for tests.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
elBoberido: no, the `PeriodicTask` manages one thread. If that thread is running, it is stopped. Maybe it's a bit misleading from the text

elBoberido:
@dkroenke I will have a look if a monotonic clock is possible with the semaphore.

Regarding the `uint64_t getExecutionCount()`. In tests, you can always add a counter to the function which will be executed. This could be helpful for other situations though. I would consider it as low prio and create an issue for this. This is a good first issue for a new contributor. Do you agree?

elBoberido:
@dkroenke regarding the monotonic clock, I just checked the the semaphore documentation and an absolute timeout is required, therefore we cannot use a monotonic clock

MatthiasKillat (commented):
Improves the sending process of introspection data. I just would consider to use ``chrono`` (why would we not use it?) . Tests are missing but this is a draft anyway.

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
```diff
@@ -330,11 +333,8 @@ class PortIntrospection
   private:
     PortData m_portData;
 
-    std::atomic<bool> m_runThread;
-    std::thread m_thread;
-
-    unsigned int m_sendIntervalCount{10};
-    const std::chrono::milliseconds m_sendIntervalSleep{100};
+    units::Duration m_sendInterval{units::Duration::seconds<long double>(1.0)};
```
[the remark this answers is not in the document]
elBoberido: That's currently a limitation of `units::Duration`. If I'm not completely a moron, that's the only way to create a duration.

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
```diff
@@ -207,7 +207,7 @@ class PoshRuntime
 
     /// @note the m_keepAliveTask should always be the last member, so that it will be the first member to be destroyed
     concurrent::PeriodicTask<cxx::MethodCallback<void>> m_keepAliveTask{
```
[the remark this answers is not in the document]
elBoberido: you can use any callable. If you have a class which overloads `operator()()` that can also be used.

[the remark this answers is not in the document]
elBoberido: and yes, it's basically `std::thread` with syntactic sugar for periodic usage

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
```diff
@@ -169,11 +169,8 @@ TEST_F(ProcessIntrospection_test, thread)
         EXPECT_CALL(introspectionAccess.getPublisherPort().value(), offer()).Times(1);
         EXPECT_CALL(introspectionAccess.getPublisherPort().value(), sendChunk(_)).Times(Between(2, 8));
 
-        std::chrono::milliseconds& sendIntervalSleep =
-            const_cast<std::chrono::milliseconds&>(introspectionAccess.m_sendIntervalSleep);
-        sendIntervalSleep = std::chrono::milliseconds(10);
-
-        introspectionAccess.setSendInterval(10);
+        using namespace iox::units::duration_literals;
+        introspectionAccess.setSendInterval(10_ms);
```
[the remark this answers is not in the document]
elBoberido: I guess you really like chrono :D
Well, I don't mind using chrono either. I think the idea behind duration was the easy conversion to timespec/timeval which we need for the POSIX calls. That's also the reason I use it here, since the `PeriodicTask` is build upon a `Semaphore` which expects a `units::Duration` for `timedWait`.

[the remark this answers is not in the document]
mossmaurice: +1
I would vote for replacing the `double` inside `unit::Duration` with `chrono`. In addition to it would then provide conversion operators for common POSIX time structs.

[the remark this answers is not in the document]
elBoberido: @mossmaurice in an issue @elfenpiff suggested to use this as base
```c++
struct duration_t {
     uint64_t seconds;
     uint32_t nanoseconds;
};
```
I think it's nice for it's use case and would also suggest to do it that way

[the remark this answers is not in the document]
elBoberido: @mossmaurice I noted this also in #190 and I think it's best discussed there or maybe in a new issue

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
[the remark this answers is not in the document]
elBoberido: Well, apparently I am a moron. One could also use `unsigned long long int` instead of `long double`.

In my shame I replace the `long double` internal representation to a fixed size in #501 to repent my stupidity 😁

elBoberido:
> Regarding the `uint64_t getExecutionCount()`. In tests, you can always add a counter to the function which will be executed. This could be helpful for other situations though. I would consider it as low prio and create an issue for this. This is a good first issue for a new contributor. Do you agree?

done with #502

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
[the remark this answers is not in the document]
elBoberido: please use #501 for any Duration related discussions

dkroenke:
@elfenpiff 350cb50 has no iox number.

iceoryx_utils/test/moduletests/test_concurrent_periodic_task.cpp
```diff
@@ -126,49 +166,78 @@ TIMING_TEST_F(PeriodicTask_test, PeriodicTaskWithObjectWithConstructorWithArgume
...
+    auto start = std::chrono::steady_clock::now();
+    concurrent::PeriodicTask<cxx::function_ref<void()>> sut(INTERVAL, "Test", [] {
+        std::this_thread::sleep_for(SLEEP_TIME);
+    });
+    sut.stop();
+    auto stop = std::chrono::steady_clock::now();
+    auto elapsedTime{std::chrono::duration_cast<std::chrono::milliseconds>(stop - start)};
+
+    EXPECT_THAT(elapsedTime, Ge(SLEEP_TIME));
+});
```
mossmaurice, line 243: What about adding a test case that call `start()` again while the read is still running?

iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
```diff
@@ -33,13 +33,22 @@ template <typename T>
 class PeriodicTask
 {
   public:
+    /// @brief Creates a periodic task. The specified callable is stored but not executed.
```
mossmaurice, line 36: Please add some documentation regarding the behaviour of the `PeriodicTask`. This is a lessons learned from the `posix::timer`, back then we called it:
```
    enum class CatchUpPolicy
    {
        SKIP_TO_NEXT_BEAT,
        IMMEDIATE,
        TERMINATE
    };
```

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
```diff
@@ -330,11 +333,8 @@ class PortIntrospection
   private:
     PortData m_portData;
 
-    std::atomic<bool> m_runThread;
-    std::thread m_thread;
-
-    unsigned int m_sendIntervalCount{10};
-    const std::chrono::milliseconds m_sendIntervalSleep{100};
+    units::Duration m_sendInterval{units::Duration::seconds<unsigned long long int>(1)};
+    concurrent::PeriodicTask<cxx::MethodCallback<void>> m_sender{"PortIntr", *this, &PortIntrospection::send};
```
mossmaurice, line 337: You might want to consider a renaming

iceoryx_posh/include/iceoryx_posh/popo/user_trigger.hpp
```diff
@@ -30,7 +30,7 @@ namespace popo
 class UserTrigger
 {
   public:
-    UserTrigger() noexcept = default;
+    UserTrigger() noexcept;
```
mossmaurice, line 33: Changed because of #494?

iceoryx_posh/test/moduletests/test_roudi_mempool_introspection.cpp
```diff
@@ -21,14 +21,8 @@
 using namespace ::testing;
 using ::testing::Return;
 
-#define private public
-#define protected public
-
```
mossmaurice, line 26: :partying_face:

mossmaurice (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
```diff
@@ -68,54 +56,37 @@ class MemPoolIntrospection
...
                                            const posix::PosixGroup& readerGroup,
                                            const posix::PosixGroup& writerGroup,
                                            uint32_t id) noexcept;
-    void send() noexcept;
 
     /// @brief copy data fro internal struct into interface struct
     void copyMemPoolInfo(const MemoryManager& memoryManager, MemPoolInfoContainer& dest) noexcept;
+
+  private:
+    units::Duration m_sendInterval{units::Duration::seconds<unsigned long long int>(1)};
```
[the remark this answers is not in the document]
elBoberido: that's unfortunately not possible within a class

iceoryx_utils/test/moduletests/test_concurrent_periodic_task.cpp
elBoberido: that's what `PeriodicTaskWhichIsActiveAppliesNewIntervalAfterStart` is doing

iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
elBoberido: will add a note that it behaves like `IMMEDIATE`. There is a [proposal](https://github.com/eclipse-iceoryx/iceoryx/issues/337#issuecomment-760909671) to implement that feature for the `posix::Timer` refactoring which will then be used in the `PeriodicTask`

iceoryx_posh/include/iceoryx_posh/popo/user_trigger.hpp
elBoberido: elfenpiff had also issues compiling this branch on Windows while we were looking why the Windows CI build was broken. But it's a nice side effect that it also fixes that bug. Shall I rebase and change the issue number for that commit?

iceoryx_posh/test/moduletests/test_roudi_mempool_introspection.cpp
elBoberido: only 5 remaining. Those also need to go the way of the dodo in order to support Windows.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
elBoberido: @mossmaurice well, apparently it's currently not `IMMEDIATE` but more like `EXTEND`, which means the time the callback takes for execution is not taken into account for simplicity reasons. This will be added once the previous mentioned refactoring is done.

iceoryx_posh/include/iceoryx_posh/popo/user_trigger.hpp
mossmaurice: Yes, please :smiley_cat: I suppose #494 can be closed then.

sculpordwarf: #494 maybe shows only the first problem with '() noexcept = default;'-constructor. If you want to close 494, you will need to search in the hole code for '() noexcept = default;'. It was several times in the codebase and I´m not sure if it was remove completely on the master in the meantime.

mossmaurice: @sculpordwarf Ok, thanks.

elBoberido: @sculpordwarf @mossmaurice moved the discussion to #494

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/mempool_introspection.hpp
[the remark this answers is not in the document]
mossmaurice: Oh, you're right. One thing I saw when checking the coreguidlines. They explicitly exclude `using namespace std::units::duration_literals`, [see](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using-directive). So we might leave them in headers but probably not in iceoryx_posh_types.hpp as this will increase lookup times.

iceoryx_utils/test/moduletests/test_concurrent_periodic_task.cpp
mossmaurice: Hmm, ok you're calling `start()` indirectly via the other c'tor. That's what I missed. Could we make the user more aware that the behaviour of the two c'tors is **very** different? One does start the other not. I have the feeling folks will do the same mistake as I in the near future (and my future self, too).

elBoberido: Any suggestions? An `@attention` doxygen tag?

mossmaurice: I'm torn inbetween. What about removing the "automatic start c'tor"? An explicit start call might be simpler for most users.

elBoberido: Can do that, but then you have the opposite problem to not forget to start it. I'm also not quite sure about this.

Just as food for thought, how about something like this
```c++
class PeriodicTask {
    enum class OnInit {
        STOPPED,
        RUNNING
    }

    PeriodicTask(OnInit onInit, ...)
};

PeriodicTask task{PeriodicTask::OnInit::RUNNING, ...}
```
It could also be the separator between the actual `PeriodicTask` args and the variadic args forwarded to the callable and thus help prevent ambiguities.

mossmaurice: I like your approach, but it might create too much noise. There can be already quite a lot of args. If we leave out the "automatic start c'tor" won't the user notice this right away and then add start?

elBoberido: well, I think it will take as much time as when starting immediately and noticing it's not what you wanted. Maybe even longer because noticing that something is not happening might take longer to realize.

dkroenke (approved):

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
```diff
@@ -322,6 +322,9 @@ class PortIntrospection
     /// @brief sends the subscriberport changing data, this is used from the unittests
     void sendSubscriberPortsData();
 
+    /// @brief calls the three specific send functions from above, this is used from the periodic task
+    void send();
```
elfenpiff, line 326: Please add noexcept

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
```diff
@@ -302,8 +302,8 @@ class PortIntrospection
                                PublisherPort&& publisherPortSubscriberPortsData);
 
     /// @brief set the time interval used to send new introspection data
-    /// @param[in] sendIntervalMs time interval in ms
-    void setSendInterval(unsigned int sendIntervalMs);
+    /// @param[in] interval duration between two send invocations
+    void setSendInterval(const units::Duration interval);
```
elfenpiff, line 306: Please add noexcept

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.inl
```diff
@@ -21,31 +21,30 @@ namespace iox
 namespace roudi
 {
 template <typename PublisherPort, typename SubscriberPort>
-PortIntrospection<PublisherPort, SubscriberPort>::PortIntrospection()
-    : m_runThread(false)
+inline PortIntrospection<PublisherPort, SubscriberPort>::PortIntrospection()
```
elfenpiff, line 24: Please add noexcept to every method.

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/process_introspection.hpp
```diff
@@ -83,8 +83,8 @@ class ProcessIntrospection
 
     /// @brief This function configures the interval for the transmission of the
     ///        port introspection data.
-    /// @param[in] interval_ms is the interval time in milliseconds
-    void setSendInterval(unsigned int interval_ms);
+    /// @param[in] interval duration between two send invocations.
+    void setSendInterval(const units::Duration interval);
```
elfenpiff, line 87: Please add noexcept.

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/process_introspection.inl
```diff
@@ -26,13 +26,12 @@ namespace iox
...
 template <typename PublisherPort>
-ProcessIntrospection<PublisherPort>::ProcessIntrospection()
-    : m_runThread(false)
+inline ProcessIntrospection<PublisherPort>::ProcessIntrospection()
 {
 }
 
 template <typename PublisherPort>
-ProcessIntrospection<PublisherPort>::~ProcessIntrospection()
+inline ProcessIntrospection<PublisherPort>::~ProcessIntrospection()
```
elfenpiff, line 34: If you can add inline you can also add noexcept - Thank you :D

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/process_introspection.inl
```diff
@@ -42,7 +41,7 @@ ProcessIntrospection<PublisherPort>::~ProcessIntrospection()
 }
 
 template <typename PublisherPort>
-void ProcessIntrospection<PublisherPort>::addProcess(int f_pid, const ProcessName_t& f_name)
+inline void ProcessIntrospection<PublisherPort>::addProcess(int f_pid, const ProcessName_t& f_name)
```
elfenpiff, line 44: Please add const

iceoryx_utils/include/iceoryx_utils/internal/concurrent/periodic_task.hpp
```diff
@@ -27,19 +27,29 @@ namespace iox
...
     /// @brief Creates a periodic task by spawning a thread. The specified callable is executed immediately on creation
     /// and then periodically after the interval duration.
-    /// @param[in] taskName will be set as thread name
+    /// @tparam Args are variadic template parameter for which are forwarded to the underlying callable object
     /// @param[in] interval is the time the thread waits between two invocations of the callable
+    /// @param[in] taskName will be set as thread name
     /// @param[in] args are forwarded to the underlying callable object
     template <typename... Args>
-    PeriodicTask(const posix::ThreadName_t taskName, const units::Duration interval, Args&&... args) noexcept;
+    PeriodicTask(const units::Duration interval, const posix::ThreadName_t taskName, Args&&... args) noexcept;
```
elfenpiff, line 52: You could add a constexpr struct to make it more readable for the user.

iceoryx_utils/test/moduletests/test_concurrent_periodic_task.cpp
elBoberido: added a struct to indicate the automatic or manual start of the task

dkroenke (approved):

elBoberido:
oh, and could the one of the reviewer please take care to the reviewer task list ;)

mossmaurice (approved):

iceoryx_posh/include/iceoryx_posh/internal/roudi/introspection/port_introspection.hpp
```diff
@@ -62,12 +62,12 @@ class PortIntrospection
 
         struct PublisherInfo
         {
-            PublisherInfo() = default;
+            PublisherInfo() noexcept {};
```
sculpordwarf, line 65: @elBoberido This will generate a warning in more strict builds:
https://rextester.com/live/WANPI61259
-Weverything
```
504430806/source.cpp:7:32: warning: extra ';' after member function definition [-Wextra-semi]
    PublisherInfo() noexcept {};
                               ^
```
the hint is at the comma

elBoberido: @sculpordwarf thanks for the hint. I'll fix it with my next PR. I'm wondering why our CI didn't catch it

Issue #488, opened 2021-01-13
cxx::vector support of Capacity 0 and erase by index


sschnitzer:
## Brief feature description

1. cxx::vector shall support "0" as Capacity
2. cxx::vector shall allow to erase an element identified by its index, i.e., new method `iterator erase(uint64_t index)`

## Detailed information

1. Support for Capacity 0 makes the cxx::vector implementation more generic. Typically, the capacity is given as configurable compile-time constants. Such a constant might potentially be set to "0", which shall not break compilation. 
2. The existing `iterator erase(iterator position)` method assumes that iterators are used. In case the vector is accessed via index, a new  `iterator erase(uint64_t index)` method is more convenient.

elfenpiff:
@sschnitzer 

1. The `cxx::vector` already supports a capacity of 0 if your compiler supports zero length arrays like `int a[0]`. This extension is supported by clang and gcc.

2. You already can use erase in such a manner. Let's say you have an `index` you could delete an element at that index with `myVector.erase(myVector.begin() + index)`.

elBoberido:
@sschnitzer is there a use case for a vector with a capacity of 0? I'm just asking because it might add dead code.

sculpordwarf:
@elBoberido
```
template <typename T, uint64_t Capacity>
class vector 
```
0 is a valid number here.
Think about MAX_INTERFACE_NUMBER = 0 if you don't want to allow a gateway, you would see the same issue in iceoryx. 
You will get the same problem in every class with preallocated memory with a cxx::vector:
```
template <uint64_t MaxStreets, uint64_t MaxLights>
class City

City<5, 0> myCity;
```
It would also be possible to write for every class a template specialization and some proper 0 handling in e.g.  the interface ports vector, but this would be even more code.

elfenpiff:
Whoever the capacity thing implements I would suggest the following idea to reduce code duplication to a minimum and it could then also be reused by our list implementations:

```cpp
// define struct where the elements are stored
template<typename T, uint64_t NumberOfElements>
struct ContainerStorage {
    using element_t = uint8_t[sizeof(T)];
    alignas(alignof(T)) element_t m_data[NumberOfElements];
};

// overload for storage when NumberOfElements is 0
template<uint64_t ElementSize>
struct ContainerStorage<ElementSize, 0U> {
};

// inherit in a protected way from ContainerStorage to gain access to m_data
template<typename T, uint64_t Capacity>
[...]

PR #491, opened 2021-01-14
Iox #488 cxx vector and list support for capacity 0

sschnitzer:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

1. [ ] Merge

## References

- Closes **TBD**


iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -739,6 +774,19 @@ TEST_F(vector_test, EraseReturnsNullWhenElementIsInvalid)
 {
     auto i = sut.begin() + 5;
     EXPECT_THAT(sut.erase(i), Eq(nullptr));
+    i = sut0.begin() + 5;
+    EXPECT_THAT(sut0.erase(i), Eq(nullptr));
+}
+
+TEST_F(vector_test, EraseReturnsIteratorWhenElementIsValid)
```
elfenpiff, line 781: Here are some tests missing:
erasing with iterator:
  - erase multiple elements returns iterator
  - erase element from vector of size 1 return end()
  - erase iterator which is out of bounds

erasing with index:
  - erase multiple elements with index returns iterator
  - erase element from vector of size 1 returns end()
  - erase index which is out of bounds

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -335,5 +343,185 @@ bool operator!=(const iox::cxx::vector<T, CapacityLeft>& lhs, const iox::cxx::ve
...
+
+    inline const T& back() const noexcept
+    {
+        std::cerr << "Attempting to access the front of an empty vector!" << std::endl;
+        std::terminate();
+    }
+
+    inline uint64_t capacity() const
+    {
+        return 0;
```
elfenpiff, line 476: Please add U

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -335,5 +343,185 @@ bool operator!=(const iox::cxx::vector<T, CapacityLeft>& lhs, const iox::cxx::ve
...
+    }
+
+    inline uint64_t size() const
+    {
+        return 0;
```
elfenpiff, line 481: Please add U

elfenpiff (changes requested):
This pull request introduces a lot more code for a very exotic corner case. I would still recommend to just adjust the `m_data` storage to one for a vector of capacity 0 like
```cpp
alignas(alignof(T)) element_t m_data[(Capacity == 0 ) ? 1 : Capacity];
```
Then the capacity implementation would consist only of one adjusted line instead of 200+ additional lines.

Otherwise we have the following expensive problems:
1. To reach a 100% test coverage for all branches we have to test additional 200 lines of code.
2. We have to make sure that the API of the two vectors are in sync, this can become error prone.
3. The new vector specialization is in the end a new class which has to be fully documented and in sync with normal vector documentation wise.
4. We have to make sure that the behavior of these two vectors are identical, otherwise we get some weird surprises. At the moment this is not the case since a vector of capacity != 0 returns a non `nullptr` with `data()`. This is at the moment a flaw in that vector and should be adjusted to `nullptr`.

We optimize here for something where we do not know if its affecting us and premature optimization is the root of all evil.

If you would like to follow this road could you please:
1. Provide me with an argument why we have to do this optimization. At the moment there is no case in iceoryx where this would have negative affects. When we later discover that there is a case we could still implement it in this way but to follow clean code guidelines we should say YAGNI (You Aint Gonna Need It).
2. Move the declarations of `vector<T, 0>` into the header.
3. Provide a full doxygen documentation for this specialization.

elBoberido:
In not quite fond of specializing cxx::vector. I have an alternative, more generic proposal and will post that one in the issue itself to not capture the PR for that discussion.

elfenpiff:
I agree with @elBoberido and think that the vector specialization is the wrong approach here. I would continue with the following steps:
1. For now you can rely on the non standardized zero sized array which is supported by gcc and clang
2. A fast solution is to adjust the following two lines in `vector.hpp`
    ```cpp
    using element_t = uint8_t[(Capacity == 0 ) ? 1 : sizeof(T)];
    alignas(alignof(T)) element_t m_data[(Capacity == 0 ) ? 1 : Capacity];
    ```
    In this case you would only waste 9 Bytes, 8 Bytes for `m_size` and 1 Byte for `m_data`. This would be an efficient enough temporary solution.
3. The final solution would be using `uninitialized_array` proposal from @elBoberido  in the issue #488 . This could be reused also for the `list` and `forward_list`.

elBoberido (changes requested):
After formulating out the alternative proposal it became clear to me that I strongly object the specialization of the `cxx::vector` since it unnecessarily adds code which needs to be maintained and certified. This functionality can be achieved in a more generic and reusable way without specializing `cxx::vector` or any other container which should also support a capacity of 0. Of course there might be some adaptions necessary in some methods for a special handling if capacity is zero, but this can be done by a simple if-statement which will be compiled away and the number of method which need this additional handling should be minimal compared to a full specialization.

The alternative proposals refactors the storage into an own class which can be used in multiple other container which in the end does not only make them capable of a capacity of 0 but also reduces the amount of code we currently have.

iceoryx_utils/include/iceoryx_utils/cxx/vector.hpp
```diff
@@ -155,11 +162,18 @@ class vector
...
     ///         the middle of the vector every element is moved one place to the
     ///         left to ensure that the elements are stored contiguously
+    /// @return iterator to the next element, returns nullptr if position is out-of-range
     iterator erase(iterator position);
 
+    /// @brief removes an element at the given vector index. if this element is in
+    ///         the middle of the vector every element is moved one place to the
+    ///         left to ensure that the elements are stored contiguously
+    /// @return iterator to the next element, returns nullptr if index is out-of-range
+    iterator erase(uint64_t index);
```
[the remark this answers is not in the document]
sschnitzer: The method of std::vector is "iterator erase( const_iterator pos )". It does not have "noexcept", since an assignment operation of the underlying data type could potentially throw an exception. I propose to have "noexcept" either on both erase methods or on none. What do you think?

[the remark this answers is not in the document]
elfenpiff: We essentially write on every method `noexcept`. You are right, it is possible that the underlying assignment operator of a certain type throws an exception but in general we do not support exceptions of any kind. And with the noexcept we state it clearly to the public - so please add it to the method.
But maybe we should also enforce it even more with a `static_assert` in the vector to only allow types which are noexcept copy and movable. @elBoberido what do you think?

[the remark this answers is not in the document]
elBoberido: Good question. The big problem I see here is a third party class where you do not have the possibility to adapt it, but you know it cannot throw an exception. It's just the developer who forgot to add the keyword. I think it is sufficient to have it in the documentation.

sschnitzer:
@elfenpiff @elBoberido I just updated PR, please have a look.
- I preferred to put the new template class that holds m_size and is specialized for Capacity 0 at the end of the inheritance chain and called it `SizedContainer` instead of `vector_storage`, since it is also used for list and forward_list. The class holding the array is now called `SizedUninitializedArray`.
- I refactored also list and forward_list to use the new classes, which slightly reduced duplicate code there. Both classes now also support Capacity 0. 
- For the erase methods, I omitted noexcept and documented the reason.

iceoryx_utils/include/iceoryx_utils/cxx/vector.hpp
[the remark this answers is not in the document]
sschnitzer: Done (documented and without noexcept).

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -184,6 +190,7 @@ TEST_F(vector_test, EmplaceBackFailsWhenSpaceNotAvailable)
         EXPECT_THAT(sut.emplace_back(5), Eq(true));
     }
     EXPECT_THAT(sut.emplace_back(5), Eq(false));
+    EXPECT_THAT(sut0.emplace_back(5), Eq(false));
```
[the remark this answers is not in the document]
sschnitzer: Done

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -202,6 +209,7 @@ TEST_F(vector_test, PushBackFailsWhenSpaceNotAvailableLValue)
         EXPECT_THAT(sut.push_back(a), Eq(true));
     }
     EXPECT_THAT(sut.push_back(a), Eq(false));
+    EXPECT_THAT(sut0.push_back(a), Eq(false));
```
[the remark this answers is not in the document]
sschnitzer: Done.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -218,12 +226,15 @@ TEST_F(vector_test, PushBackFailsWhenSpaceNotAvailableRValue)
         EXPECT_THAT(sut.push_back(5), Eq(true));
     }
     EXPECT_THAT(sut.push_back(5), Eq(false));
+    EXPECT_THAT(sut0.push_back(5), Eq(false));
```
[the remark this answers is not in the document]
sschnitzer: Done.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -303,6 +319,12 @@ TEST_F(vector_test, CopyConstructorWithEmptyVector)
     EXPECT_THAT(copyCTor, Eq(0));
     EXPECT_THAT(sut2.size(), Eq(0));
     EXPECT_THAT(sut2.empty(), Eq(true));
+
+    vector<CTorTest, 0> sut0_1;
```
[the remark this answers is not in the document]
sschnitzer: Done.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
sschnitzer: I assume by "erase multiple elements" you mean `iterator erase(const_iterator first, const_iterator last)` from std::vector? This method is not yet implemented, but could be added of course.
The other points you mentioned are done.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
sschnitzer: Done.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
sschnitzer: Done

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "iceoryx_utils/cxx/sized_container.hpp"
+#include <cstdint>
+
+namespace iox
+{
+namespace cxx
+{
+template <typename T, uint64_t Capacity>
+class SizedUninitializedArray : public SizedContainer<T, Capacity>
```
elfenpiff, line 26: What was your intention when you created `SizedContainer`?
I do not fully understand your design and at the first glimpse it makes for me more sense to move the implementation of `SizedContainer` directly into the `SizedUninitializedArray`.

Additionally, I would overload `SizedUninitializedArray` for the zero use case. With this approach we would get rid of the `m_data` member completely! The trick with `element_t m_data[Capacity == 0 ? 1 : Capacity]` would be a good approach when we wanted to solve the problem directly in the vector but when we create an abstraction we should go for the specialization.

Something like this:

```cpp
class SizedUninitializedArray;

template <typename T>
class SizedUninitializedArray<T, 0> {
  public:
     uint64_t capacity() const noexcept {
           return 0U;
     }
     // and data(), max_size()...
};

class SizedUninitializedArray<T, Capacity> {
  public:
     uint64_t capacity() const noexcept {
           return Capacity;
     }
     // and data(), max_size()...
};
```

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2021 by Robert Bosch GmbH. All rights reserved.
```
elfenpiff, line 1: I would rename the class into `uninitialized_array` then it follows more the C++/our conventions.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -1034,3 +1189,9 @@ TEST_F(vector_test, PartiallyEqualVectorsWithDifferentCapacityAreNotEqual)
     EXPECT_FALSE(a == b);
     EXPECT_TRUE(a != b);
 }
+
+TEST_F(vector_test, MemoryEfficientImplementation)
```
elfenpiff, line 1193: Very nice, I like it. I think we should do this more often - creating tests with some efficiency requirements.

elfenpiff (changes requested):

iceoryx_utils/include/iceoryx_utils/cxx/sized_container.hpp
```diff
@@ -0,0 +1,63 @@
...
+{
+namespace cxx
+{
+template <typename T, uint64_t Capacity>
+class SizedContainer;
+
+template <typename T>
+class SizedContainer<T, 0U>;
+
+template <typename T, uint64_t Capacity>
```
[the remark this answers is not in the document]
sschnitzer: Agreed, this is not needed anymore.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -335,5 +343,185 @@ bool operator!=(const iox::cxx::vector<T, CapacityLeft>& lhs, const iox::cxx::ve
     return !(lhs == rhs);
 }
 
+namespace iox
+{
+namespace cxx
+{
+// Specialization for Capacity 0
+template <typename T>
```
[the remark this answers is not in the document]
sschnitzer: Done.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
elBoberido: the specialization is not necessary since an instance of that specialization would have the same size as the current implementation. We wouldn't gain anything but add code.

elfenpiff: @elBoberido argh, you are right ... I missed that we cut it to one byte.

sschnitzer: If we merge `SizedContainer` and `SizedUninitializedArray`, we would need to specialize this new templated class for Capacity 0. This would imply that the 4 methods of `SizedUninitializedArray` would be duplicated. The "trick" with m_data avoids this and does not increase memory consumption, since an empty struct also has size 1.
Do you prefer to merge `SizedContainer` into `SizedUninitializedArray`?

elfenpiff: @elBoberido but I would still move the `SizedContainer` impl into the `SizedUninitializedArray`.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
sschnitzer: Sure, can change that. 
I tried to follow the naming conventions described in CONTRIBUTING.md, which suggest to use camel case for class names.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
elBoberido: Like I already wrote in the `SizedContainer` class. Don`t inherit from it. It makes this building block less flexible.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
```diff
@@ -0,0 +1,65 @@
...
+    /// @brief returns the capacity, which was given via the template argument,
+    ///         i.e., the maximum number of elements it can hold (same as capacity())
+    uint64_t max_size() const noexcept;
+
+
+  protected:
+    using element_t = uint8_t[Capacity == 0 ? 1 : sizeof(T)];
+
+  private:
+    alignas(Capacity == 0 ? 1 : alignof(T)) element_t m_data[Capacity == 0 ? 1 : Capacity];
```
elBoberido, line 57: We have to carefully think about possible issues arising from this. I have a slight leaning towards having this aligned to `T`.
The `std::array` also keeps the alignment of `T` if the capacity is 0 and I would rather follow the STL and trust their judgement.

iceoryx_utils/include/iceoryx_utils/cxx/vector.hpp
```diff
@@ -69,75 +74,62 @@ class vector
 
     /// @brief returns an iterator to the first element of the vector,
     ///         if the vector is empty it returns the same iterator as
-    ///         end (the first iterator which is outside of the vector)
-    iterator begin();
+    ///         end (nullptr)
```
elBoberido, line 77: please don't state that it's a `nullptr`. It's just an implementian detail and users should always compare to `end()`. If they start comparing to a `nullptr` we are locked into this implementation.

elBoberido (changes requested):
The implementation has some flaws which needs to be fixed.

Just a friendly advice. At least in the beginning of your contributions please post a draft of your implementation in the relevant issue, just function signatures and inheritance chains. This can be reviewed and saves you in the end time. You are of course free to just implement the feature and create PRs but that might result in throwing away some of your code.

Also, the PR is a little large. It's not a big problem in this particular case since the changes are not that complex but please consider to have multiple PRs if it's easily possible. In cases like this it's easier for the reviewer to have only the implementation of the new classes, their tests and maybe the integration into one class as one PR and the integration into all other classes as a second PR.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
elBoberido: > @elBoberido but I would still move the `SizedContainer` impl into the `SizedUninitializedArray`.

@elfenpiff no, it doesn't belong into `SizedUninitializedArray`, neither should `SizedUninitializedArray` inherit from it. They are not related to each other. An array has a capacity, not a size to set at runtime. That's what a vector is or a storage used to build a vector but not an array.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
sschnitzer: @elBoberido I just tried `std::array<int, 0> x` with libstdc++ 7.3.0 and g++ 7.5, which gives me `sizeof(x)==1`, `alignof(x)==1`, but `alignof(int)==4`. So I could not reproduce this. Which versions do you use? 
The explicit alignas in the uninitialized_array is just needed because the uint8_t array is later cast to T. And accessing a wrongly aligned T has to be prevented. But an empty array is never cast to T anyways, we even do not hold enough memory for a single T (assuming sizeof(T)>1). Always aligning to T would potentially waste memory if Capacity is 0.

iceoryx_utils/include/iceoryx_utils/cxx/vector.hpp
sschnitzer: Clear. The implementation no longer returns a nullptr, but I forgot the comment...

iceoryx_utils/include/iceoryx_utils/cxx/forward_list.hpp
```diff
@@ -47,9 +48,9 @@ namespace cxx
 ///
 ///
 /// @param T type user data to be managed within list
-/// @param Capacity number of maximum list elements a client can push to the list. minimum value is '1'
+/// @param Capacity number of maximum list elements a client can push to the list.
 template <typename T, uint64_t Capacity>
-class forward_list
+class forward_list : public SizedUninitializedArray<T, Capacity>
```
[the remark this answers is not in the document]
sschnitzer: You are right, this should be implemented as composition instead of inheritance. So I would change to 
- `class UninitializedArray<T, Capacity>`
- `class SizedContainer<Capacity>`
- vector, list, and forward_list use private members of `UninitializedArray` and `SizedContainer`. Only applicable public methods are added (e.g., vector won't have set_size() anymore).
@elBoberido @elfenpiff Are you both fine with this proposal?

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
elBoberido: .@sschnitzer you are right, I must have hallucinated. Like I said, it's just a slight leaning towards aligning it to `T`, since it also affects the code surrounding it, but it's better to stick with the STL and set the alignment to 1.

Just one thing, since `element_t` is protected, it might be better to do the alignment on `element_t` instead of `m_data`.

iceoryx_utils/include/iceoryx_utils/cxx/sized_container.hpp
```diff
@@ -0,0 +1,63 @@
...
+class SizedContainer
```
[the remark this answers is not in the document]
sschnitzer: Using Composition instead of Inheritance will solve this, too. See previous comment on forward_list.hpp.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
sschnitzer: Doing the alignment on element_t is a good idea in case it is used in some other class. Will change that.

iceoryx_utils/include/iceoryx_utils/cxx/sized_uninitialized_array.hpp
sschnitzer: See https://github.com/eclipse-iceoryx/iceoryx/pull/491#discussion_r560183825.

iceoryx_utils/include/iceoryx_utils/cxx/forward_list.hpp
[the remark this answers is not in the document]
sschnitzer: @elBoberido @elfenpiff: Does my previous proposal solve the review findings? If yes, I would update the PR, accordingly.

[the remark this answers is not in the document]
elBoberido: @sschnitzer sorry, this totally slipped my attention.

I still think that a specialization for `UninitializedArray` for capacity of 0 is not necessary, since the final struct still has a size of 1. But if you want to do the work I won't stop you from doing it.
If you really want to save some space, then `SizeValue` must contain the `UninitializedArray` and then calling it `SizedContainer`, `SizedStorage`, `FixedSizeDynamicContainer` or `WhatChamaCallitStorage` makes sense again. The specialization of this container makes sense in order to have a member with only one byte, which results in a `cxx::vector` with a size of also only one byte.
Just to prevent misunderstandings. I proceeded on the assumption your suggestion was not to have `SizeValue` as member of `UninitializedArray`.

Also, as elfenpiff already pointed out. For the implementation of STL types we use the STL naming convention. I think we need to clarify this in the CONTRIBUTION.md. This is not part of the public API, but just an implementation detail. Since @elfenpiff implemented most of the STL container, do we haven an unwritten rule to also have this naming convention for internal constructs?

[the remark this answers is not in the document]
sschnitzer: There are two options remaining:
a) Keep  `UninitializedArray` and `SizeValue` separate. 
    - Pro: allows to have an UninitializedArray without a size, or a size of different data type (https://github.com/eclipse-iceoryx/iceoryx/pull/491#discussion_r560125879)
    - Pro: The array-related methods are implemented just once
b) Merge the two into something like `SizedUninitializedArray`
    - Pro: Saves 1 byte for Capacity 0, since a `SizeValue` is not a separate empty struct consuming 1 Byte
    - Pro: one class less (.hpp + .inl)

In case of a), I also would prefer the "trick":
```
using element_t = alignas(Capacity == 0 ? 1 : alignof(T)) uint8_t[Capacity == 0 ? 1 : sizeof(T)];
element_t m_data[Capacity == 0 ? 1 : Capacity];
```
In case of b), the trick is not needed. For capacity 0, the `end()` iterator would just return `this()`.

If you ask me, I prefer option b), since it is slightly more memory-efficient and the loss in modularity might be a case of YAGNI.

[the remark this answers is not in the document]
elBoberido: There is option c)
- have a `ContainerStorage` which inherits from `UninitializedArray` and has a `fillLevel` member if capacity is not 0
- this has all the methods from `UninitializedArray`
- the number of specialized methods is minimal <- this is the main reason I'm leaning towards this solution
- since this is not user facing, inheritance shouldn't create problems further down the road

Currently I can think only of the `cxx::vector` where this member is useful. If the `cxx::list` also uses it, it's a source for future optimization if we want to have the container as small as possible.
There are other container in our code base which could make use of the `UninitializedArray` but don't need the additional size member because they handle this stuff differently, e.g. the `SoFi`  or the `ObjectPool`. For these, the size member would be dead memory.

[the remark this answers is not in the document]
sschnitzer: Right, Option c) has the same Pro's as a) and additionally the first Pro from b). 
Btw., the `SoFi` and the `ObjectPool` also have `m_size`, so this could also be refactored to use ContainerStorage.
@elfenpiff Do you agree to proceed with Option c)?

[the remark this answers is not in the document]
elBoberido: @sschnitzer the `m_size` from the `SoFi` has a different meaning. A better name would be `m_currentMaxCapacity`. This was introduced to have a somewhat resizable `SoFi`. It just limits the the number of elements the `SoFi` can store, but doesn't tell anything about the current fill level.
With the `ObjectPool` you are right.

I also just noticed that @MatthiasKillat did something similar for his queue
https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/buffer.hpp
Maybe we can reuse that and put it into a more common place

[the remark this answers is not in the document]
MatthiasKillat: @elBoberido It should probably not be called buffer then, unitialized_array might be better, I have no other proposals. As far as I am not sure whether it can allocate 0 bytes out of the box. Using it with template parameter 0 will cause an internal array to have size 0, which is not portable. I think I used it (in a test) with capacity 0 though.

This could easily be adapted to allocate 1 byte internally if Capacity is 0, it is just compile time computation.
As for where it should be placed I do not know, currently it is just a helper structure for the lockfree queue.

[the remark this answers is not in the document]
sschnitzer: @elBoberido @elfenpiff I updated the PR according to Option c). Can you please have a look?

sschnitzer:
During development I found a bug in the existing vector::erase() method: It called the wrong destructor. The fix is part of this PR (https://github.com/eclipse-iceoryx/iceoryx/pull/491/commits/7692dfe17a5b766870e081ea8931807a09ef84cb and https://github.com/eclipse-iceoryx/iceoryx/pull/491/commits/97fbec5244fef213490014c420efe47bcbed36bd).

elfenpiff:
> During development I found a bug in the existing vector::erase() method: It called the wrong destructor. The fix is part of this PR (7692dfe and 97fbec5).

I looked at your bugfix and I think `vector::erase()` works fine.

Let's say you have a vector `myVector` with the contents { 1, 2, 3, 4, 5 }. If you would like to remove the third element the following should happening behind the scenes.
```
myVector.erase(myVector.begin() + 2);

1. move element 4 into element 3
2. move element 5 into element 4
3. call dtor of element 5
```

With your fix the following would happen:
```
myVector.erase(myVector.begin() + 2);

1. call dtor of element 3
2. move element 4 into element 3
```
This would cause undefined behavior since the dtor of element 3 was already called and therefore you are not allowed to perform operations on element 3!
You could fix this by performing a placement new with the move constructor and change the code from step 2. into

```
2. placement new at position of element 3 and calling move ctor with the value of element 4
3. move element 5 into element 4
```
We never called the dtor of element 5 at this step. This would cause a undefined behavior and a real bug when we would add another element to the vector with `push_back` for instance.

At this moment we have the following contents inside the vector {1, 2, 4, 5, ?} the question mark is an object in an undefined state but the size states 4 even so the element is present.
```
myVector.push_back(6);
```
Now we create the element 6 via placement new at the position of the element ? which formerly contained element 5. This means we hard override the memory of a still existing element which could lead to a memory leak, segmentation fault or even worse, undefined behavior. To fix this we have to do the following.


```
// we start with the vector {1, 2, 3, 4, 5}
myVector.erase(myVector.begin() + 2);
1. call dtor of element 3
2. placement new at position of element 3 and calling move ctor with the value of element 4
3. move element 5 into element 4
4. call dtor of element 5 at position X

myVector.push_back(6);
1. placement new at position X
```
We this approach we would call the dtor of element 3 and additionally element 5 which would cause some overhead and I am not sure if we would like to have this since element 3 has the responsibility to cleanup all resources when a move assignment happens.

sschnitzer:
@elfenpiff Thanks for the review of the bugfix. 
You basically brought up 3 points:
1. Is it allowed to access an element after dtor was called. While this works in this particular case, we should not rely on unspecified behavior. I agree that using the move constructor with inplace-new should be used instead. (std::vector just copies the memory, which is faster but also less flexible)
1.  The move assignment operator has the responsibility to cleanup resources. Are you sure? Is this defined by the C++ standard or rather good practice? There are also a couple of places in Iceoryx where this is not the case (whereas the dtor correctly cleans up), see [unix_domain_socket](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp#L113-L126),  [shared_memory](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/source/posix_wrapper/shared_memory_object/shared_memory.cpp#L102-L116), [memory_map](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/source/posix_wrapper/shared_memory_object/memory_map.cpp#L111-L122), and [GenericRAII](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/source/cxx/generic_raii.cpp#L40-L48).
1. An object (e.g., at index 5) that was moved somewhere else (e.g., to index 4) is still existing. I assumed it is not allowed to use an object after it was moved. While this might be recommended, you are correct that the C++ standard does not explicitly requires that. But shall we really assume this? If yes, why is it sufficient to call dtor of 5 after the move assignment, but calling the dtor of 4 after it got moved to 3 is not required?

elfenpiff:
@sschnitzer

2. I created a bug report since the classes which you pointed out are definitely working incorrectly and can produce resource leaks: #542 

Thank you for the hint!!!!

3. After move an object is in an undefined but destructable state according to the C++ standard. This means when you call methods and access member variables after move you have undefined behavior but calling the dtor works.
   Let's again assume you have vector { 1[a], 2[b], 3[c], 4[d], 5[e]}. The numbers are the position and in the brackets are the values [a] - [e]. If you delete position 3 the element [d] will be moved to position 3 and element [e] will be moved to position 4. Position 5 is now in an undefined state [?] but is still destructable therefore we have to destruct it to clean correctly. When you moved [d] to position 3 it has to clean up all resources which where managed by position 4 since the ownership has now position 3. And position 3 has to clean up all resources of [c] which it was managing since after the move it has lost the ownership of [c].

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -259,66 +271,77 @@ const T& vector<T, Capacity>::front() const noexcept
...
-            at(n) = std::move(at(n + 1u));
-        }
-        at(n).~T();
-        m_size--;
+        return end();
+    }
+    at(index).~T();
+    for (; n + 1u < this->size(); ++n)
+    {
+        new (&at(n)) T(std::move(at(n + 1u)));
```
elfenpiff, line 341: Here you would have undefined behavior. Let's assume we have a vector with the position[content] { 1[a], 2[b], 3[c], 4[d], 5[e] }.

We call again
```cpp
myVector.erase(2);
```

1. DTor of position 3 is called
2. Placement new at position 3 with move ctor of content [d] of position 4.
3. Placement new at position 4 with move ctor of content [e] of position 5. This is undefined behavior since you override the memory of an existing object without calling the dtor before.

elBoberido:
>     1. After move an object is in an undefined but destructable state according to the C++ standard.
Small nitpick. It must be in a defined but unspecified state. After a move a vector could be empty or still have it's original size but with a flag that the data is moved. For the vector itself it's well defined, but it's not specified to the outer world. To overcome the unspecified state you can e.g. assign a default vector and make it again defined and specified.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
sschnitzer: @elfenpiff I updated the PR and now also call the dtor of the moved objects.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -51,8 +52,8 @@ inline vector<T, Capacity>::vector(const uint64_t count)
                   << std::endl;
     }
 
-    m_size = std::min(count, Capacity);
-    for (uint64_t i = 0U; i < m_size; ++i)
```
elfenpiff, line 55: Could you please add `cxx::Expects(count <= Capacity);`

iceoryx_utils/include/iceoryx_utils/cxx/uninitialized_array.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+template <typename T, uint64_t Capacity>
+class uninitialized_array
+{
+  public:
+    using value_type = T;
+    using size_type = decltype(Capacity);
+
+    uninitialized_array() = default;
+
```
elfenpiff, line 33: This thing should be neither move nor copyable otherwise we will run into serious troubles.
Could you please delete copy and move ctor/assignment.

This is just a building block and the copy/move has to be implemented from the class which is using this as a member.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -261,66 +274,78 @@ const T& vector<T, Capacity>::front() const noexcept
...
+    /// If position < begin(), this shall return end(), which is guaranteed:
+    /// WLOG, let vector v have Capacity=(2^64-begin())/sizeof(T)-sizeof(m_size)
+    /// (note that iterators, index, and Capacity are all limited by 64 bit address space)
+    /// With position=0, position-begin()==(2^64-begin())/sizeof(T) > Capacity => returns end()
+    uint64_t index = static_cast<uint64_t>(position - begin());
+    return erase(index);
+}
+
+template <typename T, uint64_t Capacity>
+inline typename vector<T, Capacity>::iterator vector<T, Capacity>::erase(uint64_t index)
```
elfenpiff, line 334: Please add const for the parameter here and in the header.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -250,7 +263,7 @@ T& vector<T, Capacity>::front() noexcept
 }
 
 template <typename T, uint64_t Capacity>
-const T& vector<T, Capacity>::front() const noexcept
+inline typename vector<T, Capacity>::const_reference vector<T, Capacity>::front() const noexcept
```
elfenpiff, line 266: Why is this better? I would leave as it is expect the missing inline, this should be added. This goes for all aliases except `value_type` which is an API element for developer to get the underlying type of a vector.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
```diff
@@ -767,11 +898,37 @@ TEST_F(vector_test, EraseOfLastElementCallsDTorOnly)
...
+}
+
+TEST_F(vector_test, EraseByIndexOfLastElementCallsDTorOnly)
+{
+    vector<CTorTest, 5> sut1;
+    sut1.emplace_back(7);
+    sut1.emplace_back(8);
+    sut1.emplace_back(9);
+
+    sut1.erase((uint64_t)2);
```
elfenpiff, line 928: Please never use c-style casts. Here you should write `sut1.erase(2U);`

elfenpiff (changes requested):
Everything reviewed except the vector unit tests. This comes later.

iceoryx_utils/include/iceoryx_utils/cxx/uninitialized_array.hpp
```diff
@@ -0,0 +1,61 @@
...
+#ifndef ICEORYX_UTILS_CXX_UNINITIALIZED_ARRAY_HPP
+#define ICEORYX_UTILS_CXX_UNINITIALIZED_ARRAY_HPP
+
+#include <cstdint>
+
+namespace iox
+{
+namespace cxx
+{
+template <typename T, uint64_t Capacity>
```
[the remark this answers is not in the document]
sschnitzer: @elfenpiff An uninitialized_array for Capacity == 0 uses just 1 Byte, so overloading would not consume less memory but just add more code. See https://github.com/eclipse-iceoryx/iceoryx/issues/488#issuecomment-760158500 by @elBoberido.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
sschnitzer: Sure. Just want to mention that this changes the behavior in case of `count > Capacity` from a warning message to terminate(). For consistency reasons I would also apply the same change to `vector(const uint64_t count, const T& value)`.

iceoryx_utils/include/iceoryx_utils/cxx/uninitialized_array.hpp
sschnitzer: Agreed.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
sschnitzer: Ok.

iceoryx_utils/test/moduletests/test_cxx_vector.cpp
sschnitzer: Fine. I consequently also changed `erase((uint64_t)0U)` to `erase(static_cast<uint64_t>(0U))`.

iceoryx_utils/include/iceoryx_utils/cxx/uninitialized_array.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    /// @brief return the const pointer to the underlying array
+    /// @return const pointer to underlying array
+    const T* data() const noexcept;
+
+
+    /// @brief returns the capacity, which was given via the template argument,
+    ///         i.e., the maximum number of elements it can hold
+    uint64_t capacity() const noexcept;
+
```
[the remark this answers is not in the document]
sschnitzer: On a vector having Capacity==0, returning index 0 would be a bad idea. If you really want to be sure we should call terminate() in the vector. 
Proposal: Lets instead rename (in some future ticket) `Expects` to something like `Expect_or_Terminate` to make very clear that terminate() must not be removed.

iceoryx_utils/include/iceoryx_utils/cxx/list.hpp
```diff
@@ -30,7 +31,7 @@ namespace cxx
 /// @details Adjustments in the API were done to not use exceptions and serve the requirement of
 ///         a data structure movable over shared memory.
 ///         attempt to add elements to a full list will be ignored.
-///         Capacity must at least be 1, (unintended) negative initialization is rejected with compile assertion
+///         (Unintended) negative initialization is rejected with compile assertion
```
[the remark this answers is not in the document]
sschnitzer: Right, I update the comment.

iceoryx_utils/include/iceoryx_utils/cxx/forward_list.hpp
```diff
@@ -30,7 +31,7 @@ namespace cxx
 /// @details  Adjustments in the API were done to not use exceptions and serve the requirement of
 ///         a data structure movable over shared memory.
 ///         attempt to add elements to a full list will be ignored.
-///         Capacity must at least be 1, (unintended) negative initialization is rejected with compile assertion
+///         (Unintended) negative initialization is rejected with compile assertion
```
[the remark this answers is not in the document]
sschnitzer: Right, I update the comment.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
sschnitzer: Typedefs like `const_reference` or `const_iterator` are part of std::vector. To use them was also proposed by @elBoberido (see https://github.com/eclipse-iceoryx/iceoryx/issues/488#issuecomment-760158500). 
Do you propose to remove them from both, the vector.hpp and the vector.inl? Or keep them at vector.hpp but remove them at vector.inl?
I still think that keeping vector.hpp and vector.inl method signatures in sync with each other and also with stl::vector is the best approach.

iceoryx_utils/include/iceoryx_utils/cxx/container_storage.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+namespace iox
+{
+namespace cxx
+{
+template <typename T, uint64_t Capacity>
+class container_storage;
+
+template <typename T>
+class container_storage<T, 0U>;
```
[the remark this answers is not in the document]
sschnitzer: Ok. Definition for the specialization is now also in the .hpp file. Implementation (out-of-class) is in the .inl file.

elBoberido:
@sschnitzer it seems there were changes to the `cxx::vector` which are in conflict with this PR since this PR is now open for quite a long time.

Maybe it's best to split this PR up into two or three smaller ones. This makes it less likely to have merge conflicts and also makes it easier to review. My proposal would be
- create a `uninitialized_array` class and tests based on  https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/internal/concurrent/lockfree_queue/buffer.hpp
  - move `concurrent/buffer.hpp` to `cxx/uninitialized_array.hpp` and adjust classes which are using the buffer
  - extend it to fullfil your `uninitialized_array` proposal
- create `container_storage` and tests
  - maybe integrate it into one container
- integrate into remaining container where applicable

What do you think?

elfenpiff (commented):
@sschnitzer the last commit is from 18.03.2020. Is this PR dead and can be removed or can we expect further work on it?
Maybe it makes sense to close this PR for now, you finish the work or introduce a design document and we continue from there otherwise we have PRs in here which are older than 4 months which is getting more and more out of sync.

sschnitzer:
> @sschnitzer the last commit is from 18.03.2020. Is this PR dead and can be removed or can we expect further work on it?
> Maybe it makes sense to close this PR for now, you finish the work or introduce a design document and we continue from there otherwise we have PRs in here which are older than 4 months which is getting more and more out of sync.
@elfenpiff Thanks for the hint. At the [last developer meetup](https://github.com/eclipse-iceoryx/iceoryx/wiki/2021-04-15-Eclipse-iceoryx-developer-meetup) the conclusion was to restart this topic with a design document and smaller PRs. Consequently, no further work will happen in PR 491 and it can be closed.

sschnitzer:
As discussed in the [last developer meetup](https://github.com/eclipse-iceoryx/iceoryx/wiki/2021-04-15-Eclipse-iceoryx-developer-meetup) this topic will be continued using a design document and smaller PRs.

PR #501, opened 2021-01-16
Iox #190 fixed point types for duration internal representation

### LINKED ISSUE #190: nanoseconds in units::Duration are truncated

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer

This replaces the `long double` internal representation of `units::Duration` with `uint64_t` for seconds and `uint32_t` for nanoseconds.

1. [ ] Merge

## References

- Closes #190

### DIFF
--- a/iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
+++ b/iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
@@ -161,9 +161,13 @@ class Mepoo_IntegrationTest : public Test
 
     void PrintTiming(iox::units::Duration start)
     {
-        std::cerr << "RouDi startup took " << start.milliSeconds<int>() << " milliseconds, "
-                  << "(which is " << start.seconds<int>() << " seconds)"
-                  << "(which is " << start.minutes<int>() << " minutes)" << std::endl;
+        auto totalMillisconds = start.milliSeconds();
+        auto millisceonds = totalMillisconds % 1000U;
+        auto totalSeconds = totalMillisconds / 1000U;
+        auto seconds = totalSeconds % 60U;
+        auto minutes = totalSeconds / 60U;
+        std::cerr << "RouDi startup took " << minutes << " minutes " << seconds << " seconds and " << millisceonds
+                  << " milliseconds" << std::endl;
     }
 
     template <uint32_t size>
--- a/iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
+++ b/iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
@@ -162,11 +162,11 @@ class Mepoo_IntegrationTest : public Test
     void PrintTiming(iox::units::Duration start)
     {
         auto totalMillisconds = start.milliSeconds();
-        auto millisceonds = totalMillisconds % 1000U;
+        auto milliseconds = totalMillisconds % 1000U;
         auto totalSeconds = totalMillisconds / 1000U;
         auto seconds = totalSeconds % 60U;
         auto minutes = totalSeconds / 60U;
-        std::cerr << "RouDi startup took " << minutes << " minutes " << seconds << " seconds and " << millisceonds
+        std::cerr << "RouDi startup took " << minutes << " minutes " << seconds << " seconds and " << milliseconds
                   << " milliseconds" << std::endl;
     }
 
--- a/iceoryx_posh/test/moduletests/test_gw_gateway_discovery.cpp
+++ b/iceoryx_posh/test/moduletests/test_gw_gateway_discovery.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2019, 2021 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -14,14 +14,8 @@
 
 #include "test.hpp"
 
-#define private public
-#define protected public
-
 #include "iceoryx_posh/gateway/gateway_discovery.hpp"
 #include "iceoryx_posh/internal/popo/ports/base_port.hpp"
-
-#undef private
-#undef protected
 
 using namespace ::testing;
 using ::testing::Return;
@@ -30,6 +24,15 @@ using CaproMessage = iox::capro::CaproMessage;
 using BasePort = iox::popo::BasePort;
 using InterfacePort = iox::popo::InterfacePort;
 
+template <typename T>
+class GatewayDiscoveryAccess : public iox::gw::GatewayDiscovery<T>
+{
+  public:
+    GatewayDiscoveryAccess(T interfacePortImpl)
+        : iox::gw::GatewayDiscovery<T>(interfacePortImpl)
+    {
+    }
+};
 
 class InterfacePort_mock
 {
@@ -53,7 +56,8 @@ class GatewayDiscovery_test : public Test
 TEST_F(GatewayDiscovery_test, GetCaproMessage)
 {
     InterfacePort_mock interfacePortImplMock;
-    iox::gw::GatewayDiscovery<InterfacePort_mock> GatewayDiscovery(interfacePortImplMock);
+    iox::gw::GatewayDiscovery<InterfacePort_mock> GatewayDiscovery =
+        GatewayDiscoveryAccess<InterfacePort_mock>(interfacePortImplMock);
     CaproMessage msg;
     GatewayDiscovery.getCaproMessage(msg);
     EXPECT_EQ(iox::capro::CaproMessageType::ACK, msg.m_type);
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -291,21 +291,15 @@ class Duration
 
   private:
     template <typename T>
-    inline constexpr Duration
-    multiplySeconds(const uint64_t seconds,
-                    const std::enable_if_t<!std::is_floating_point<T>::value, T>& rhs) const noexcept;
+    inline constexpr Duration fromFloatingPointSeconds(const T floatingPointSeconds) const noexcept;
+
     template <typename T>
     inline constexpr Duration
-    multiplySeconds(const uint64_t seconds,
-                    const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const noexcept;
+    multiplyWith(const std::enable_if_t<!std::is_floating_point<T>::value, T>& rhs) const noexcept;
+
     template <typename T>
     inline constexpr Duration
-    multiplyNanoseconds(const uint32_t nanoseconds,
-                        const std::enable_if_t<!std::is_floating_point<T>::value, T>& rhs) const noexcept;
-    template <typename T>
-    inline constexpr Duration
-    multiplyNanoseconds(const uint32_t nanoseconds,
-                        const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const noexcept;
+    multiplyWith(const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const noexcept;
 
     static constexpr uint32_t SECS_PER_MINUTE{60U};
     static constexpr uint32_t SECS_PER_HOUR{3600U};
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -21,6 +21,7 @@
 #include <chrono>
 #include <cmath>
 #include <iostream>
+#include <numeric>
 
 namespace iox
 {
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -207,24 +207,24 @@ class Duration
 
     // BEGIN ARITHMETIC
 
-    /// @brief creates Duration object by adding right
+    /// @brief Creates Duration object by adding right
     /// @param[in] right is the second summand
     /// @return a new Duration object
     constexpr Duration operator+(const Duration& right) const noexcept;
 
-    /// @brief creates Duration object by subtracting right
+    /// @brief Creates Duration object by subtracting right
     /// @param[in] right is the subtrahend
     /// @return a new Duration object
     /// @attention since negative durations are not allowed, the duration will be capped to 0
     constexpr Duration operator-(const Duration& right) const noexcept;
 
-    /// @brief creates Duration object by multiplication
+    /// @brief Creates Duration object by multiplication
     /// @tparam T is an arithmetic type for the multiplicator
     /// @param[in] right is the multiplicator
     /// @return a new Duration object
     /// @attention since negative durations are not allowed, the duration will be capped to 0
-    /// @note there is no explicit division operator, if you want a fraction of of a Duration, either multiply with the
-    /// inverse of the divisor
+    /// @note There is no explicit division operator! This can be achieved by multiplication with the inverse of the
+    /// divisor.
     template <typename T>
     constexpr Duration operator*(const T& right) const noexcept;
 
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -38,43 +38,22 @@ class Duration;
 inline namespace duration_literals
 {
 /// @brief constructs a new Duration object in nanoseconds
-constexpr Duration operator"" _ns(long double);
-
-/// @brief constructs a new Duration object in nanoseconds
 constexpr Duration operator"" _ns(unsigned long long int); // PRQA S 48
-
-/// @brief constructs a new Duration object in microseconds
-constexpr Duration operator"" _us(long double);
 
 /// @brief constructs a new Duration object in microseconds
 constexpr Duration operator"" _us(unsigned long long int); // PRQA S 48
 
 /// @brief constructs a new Duration object in milliseconds
-constexpr Duration operator"" _ms(long double);
-
-/// @brief constructs a new Duration object in milliseconds
 constexpr Duration operator"" _ms(unsigned long long int); // PRQA S 48
-
-/// @brief constructs a new Duration object in seconds
-constexpr Duration operator"" _s(long double);
 
 /// @brief constructs a new Duration object in seconds
 constexpr Duration operator"" _s(unsigned long long int); // PRQA S 48
 
 /// @brief constructs a new Duration object in minutes
-constexpr Duration operator"" _m(long double);
-
-/// @brief constructs a new Duration object in minutes
 constexpr Duration operator"" _m(unsigned long long int); // PRQA S 48
 
 /// @brief constructs a new Duration object in hours
-constexpr Duration operator"" _h(long double);
-
-/// @brief constructs a new Duration object in hours
 constexpr Duration operator"" _h(unsigned long long int); // PRQA S 48
-
-/// @brief constructs a new Duration object in days
-constexpr Duration operator"" _d(long double);
 
 /// @brief constructs a new Duration object in days
 constexpr Duration operator"" _d(unsigned long long int); // PRQA S 48
@@ -126,17 +105,29 @@ class Duration
     /// @brief Assigns a std::chrono::milliseconds to an duration object
     Duration& operator=(const std::chrono::milliseconds& right);
 
-    /// @brief return true if durationInSeconds is larger or equal than right
+    /// @brief Equal to operator
+    /// @return true if duration equal to right
+    constexpr bool operator==(const Duration& right) const;
+
+    /// @brief Not equal to operator
+    /// @return true if duration not equal to right
+    constexpr bool operator!=(const Duration& right) const;
+
+    /// @brief Less than operator
+    /// @return true if duration is less than right
     constexpr bool operator<(const Duration& right) const;
 
-    /// @brief return true if durationInSeconds is larger or equal than right
+    /// @brief Less than or equal to operator
+    /// @return true if duration is less than or equal to right
+    constexpr bool operator<=(const Duration& right) const;
+
+    /// @brief Greater than operator
+    /// @return true if duration is greater than right
     constexpr bool operator>(const Duration& right) const;
 
-    /// @brief return true if durationInSeconds is larger or equal than right
+    /// @brief Greater than or equal to operator
+    /// @return true if duration is greater than or equal to right
     constexpr bool operator>=(const Duration& right) const;
-
-    /// @brief returns true if right is larger or equal than durationInSeconds
-    constexpr bool operator<=(const Duration& right) const;
 
     /// @brief creates Duration object by adding right and durationInSeconds
     constexpr Duration operator+(const Duration& right) const;
@@ -194,26 +185,18 @@ class Duration
     /// @brief converts time in a timespec c struct
     struct timespec timespec(const TimeSpecReference& reference = TimeSpecReference::None) const;
 
-    //@ brief Make operators accessible, that have to be defined outside the class
-    // template <typename T>
-    // friend constexpr Duration operator*(const T& left, const Duration& right);
-    // template <typename T>
-    // friend constexpr Duration operator/(const T& left, const Duration& right);
+    template <typename T>
+    friend constexpr Duration operator*(const T& left, const Duration& right);
+    template <typename T>
+    friend constexpr Duration operator/(const T& left, const Duration& right);
     friend std::ostream& operator<<(std::ostream& stream, const Duration& t);
-    friend constexpr Duration duration_literals::operator"" _ns(long double);
     friend constexpr Duration duration_literals::operator"" _ns(unsigned long long int); // PRQA S 48
-    friend constexpr Duration duration_literals::operator"" _us(long double);
     friend constexpr Duration duration_literals::operator"" _us(unsigned long long int); // PRQA S 48
-    friend constexpr Duration duration_literals::operator"" _ms(long double);
     friend constexpr Duration duration_literals::operator"" _ms(unsigned long long int); // PRQA S 48
-    friend constexpr Duration duration_literals::operator"" _s(long double);
-    friend constexpr Duration duration_literals::operator"" _s(unsigned long long int); // PRQA S 48
-    friend constexpr Duration duration_literals::operator"" _m(long double);
-    friend constexpr Duration duration_literals::operator"" _m(unsigned long long int); // PRQA S 48
-    friend constexpr Duration duration_literals::operator"" _h(long double);
-    friend constexpr Duration duration_literals::operator"" _h(unsigned long long int); // PRQA S 48
-    friend constexpr Duration duration_literals::operator"" _d(long double);
-    friend constexpr Duration duration_literals::operator"" _d(unsigned long long int); // PRQA S 48
+    friend constexpr Duration duration_literals::operator"" _s(unsigned long long int);  // PRQA S 48
+    friend constexpr Duration duration_literals::operator"" _m(unsigned long long int);  // PRQA S 48
+    friend constexpr Duration duration_literals::operator"" _h(unsigned long long int);  // PRQA S 48
+    friend constexpr Duration duration_literals::operator"" _d(unsigned long long int);  // PRQA S 48
 
   private:
     /// @brief constructor needs to be private to ensure a unit safe usage of duration
@@ -221,7 +204,7 @@ class Duration
     long double durationInSeconds{0.0};
 };
 
-/// @brief creates Duration object multplying T with durationInSeconds
+/// @brief creates Duration object multiplying T with durationInSeconds
 template <typename T>
 constexpr Duration operator*(const T& left, const Duration& right);
 
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -221,6 +221,7 @@ class Duration
     /// @attention since negative durations are not allowed, the duration will be capped to 0
     /// @note There is no explicit division operator! This can be achieved by multiplication with the inverse of the
     /// divisor.
+    /// @note Multiplication with NaN and +Inf results in a saturated max duration
     template <typename T>
     constexpr Duration operator*(const T& rhs) const noexcept;
 
@@ -305,26 +306,26 @@ class Duration
     /// @note this is protected to be able to use it in unit tests
     constexpr Duration(const SECONDS_TYPE seconds, const NANOSECONDS_TYPE nanoseconds) noexcept;
 
+    /// @note this is factory method is necessary to build with msvc due to issues calling a protected constexpr ctor
+    /// from public methods
     static constexpr Duration createDuration(const SECONDS_TYPE seconds, const NANOSECONDS_TYPE nanoseconds) noexcept;
 
-    inline static constexpr Duration max();
+    static constexpr Duration max() noexcept;
 
   private:
     template <typename T, typename String>
-    inline static constexpr unsigned long long int positiveValueOrClampToZero(const T value, const String fromMethod);
+    static constexpr unsigned long long int positiveValueOrClampToZero(const T value, const String fromMethod) noexcept;
 
     template <typename T>
-    inline constexpr Duration fromFloatingPointSeconds(const T floatingPointSeconds) const noexcept;
+    constexpr Duration fromFloatingPointSeconds(const T floatingPointSeconds) const noexcept;
     template <typename From, typename To>
-    inline constexpr bool wouldCastFromFloatingPointProbablyOverflow(const From floatingPoint) const noexcept;
+    constexpr bool wouldCastFromFloatingPointProbablyOverflow(const From floatingPoint) const noexcept;
 
     template <typename T>
-    inline constexpr Duration multiplyWith(const std::enable_if_t<!std::is_floating_point<T>::value, T>& rhs) const
-        noexcept;
+    constexpr Duration multiplyWith(const std::enable_if_t<!std::is_floating_point<T>::value, T>& rhs) const noexcept;
 
     template <typename T>
-    inline constexpr Duration multiplyWith(const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const
-        noexcept;
+    constexpr Duration multiplyWith(const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const noexcept;
 
   private:
     SECONDS_TYPE m_seconds{0U};
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -304,9 +304,22 @@ class Duration
     multiplyNanoseconds(const uint32_t nanoseconds,
                         const std::enable_if_t<std::is_floating_point<T>::value, T>& right) const;
 
+    static constexpr uint32_t SECS_PER_MINUTE{60U};
+    static constexpr uint32_t SECS_PER_HOUR{3600U};
+    static constexpr uint32_t HOURS_PER_DAY{24U};
+
+    static constexpr uint32_t MILLISECS_PER_SEC{1000U};
+    static constexpr uint32_t MICROSECS_PER_SEC{MILLISECS_PER_SEC * 1000U};
+
+    static constexpr uint32_t NANOSECS_PER_MICROSEC{1000U};
+    static constexpr uint32_t NANOSECS_PER_MILLISEC{NANOSECS_PER_MICROSEC * 1000U};
+    static constexpr uint32_t NANOSECS_PER_SEC{NANOSECS_PER_MILLISEC * 1000U};
+
+    static_assert(NANOSECS_PER_SEC == 1000U * MICROSECS_PER_SEC, "Mismatch in calculation for conversion constants!");
+
   private:
-    uint64_t m_seconds{0};
-    uint32_t m_nanoseconds{0};
+    uint64_t m_seconds{0U};
+    uint32_t m_nanoseconds{0U};
 };
 
 /// @brief creates Duration object by multiplying object T with a duration
--- a/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
@@ -80,7 +80,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as nanoseconds
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration nanoseconds(const T value) noexcept;
 
@@ -88,7 +88,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as microseconds
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration microseconds(const T value) noexcept;
 
@@ -96,7 +96,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as milliseconds
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration milliseconds(const T value) noexcept;
 
@@ -104,7 +104,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as seconds
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration seconds(const T value) noexcept;
 
@@ -112,7 +112,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as minutes
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration minutes(const T value) noexcept;
 
@@ -120,7 +120,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as hours
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration hours(const T value) noexcept;
 
@@ -128,7 +128,7 @@ class Duration
     /// @tparam T is an integer type for the value
     /// @param[in] value as days
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     template <typename T>
     static constexpr Duration days(const T value) noexcept;
 
@@ -151,18 +151,18 @@ class Duration
 
     /// @brief Construct a Duration object from std::chrono::milliseconds
     /// @param[in] value as milliseconds
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     constexpr explicit Duration(const std::chrono::milliseconds& value) noexcept;
 
     /// @brief Construct a Duration object from std::chrono::nanoseconds
     /// @param[in] value as nanoseconds
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     constexpr explicit Duration(const std::chrono::nanoseconds& value) noexcept;
 
     /// @brief Assigns a std::chrono::milliseconds to an duration object
     /// @param[in] rhs is the right hand side of the assignment
     /// @return a reference to the Duration object with the assigned millisecond value
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     Duration& operator=(const std::chrono::milliseconds& rhs) noexcept;
 
     // END CONSTRUCTORS AND ASSIGNMENT
@@ -211,17 +211,18 @@ class Duration
     /// @brief Creates Duration object by subtraction
     /// @param[in] rhs is the subtrahend
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
     constexpr Duration operator-(const Duration& rhs) const noexcept;
 
     /// @brief Creates Duration object by multiplication
     /// @tparam T is an arithmetic type for the multiplicator
     /// @param[in] rhs is the multiplicator
     /// @return a new Duration object
-    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @attention Since negative durations are not allowed, the duration will be clamped to 0
+    /// @note A duration of 0 will always result in 0, no matter if multiplied with NaN or +Inf
     /// @note There is no explicit division operator! This can be achieved by multiplication with the inverse of the
     /// divisor.
-    /// @note Multiplication with NaN and +Inf results in a saturated max duration
+    /// @note Multiplication of a non-zero duration with NaN and +Inf results in a saturated max duration
     template <typename T>
     constexpr Duration operator*(const T& rhs) const noexcept;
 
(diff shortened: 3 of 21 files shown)


iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,127 +37,206 @@ class Duration;
...
+    static constexpr Duration milliseconds(const T value);
+
+    /// @brief Constructs a new Duration object from seconds
+    /// @tparam T is an integer type for the value
+    /// @param[in] value as seconds
+    /// @return a new Duration object
+    /// @attention since negative durations are not allowed, the duration will be capped to 0
     template <typename T>
-    static constexpr Duration seconds(const T seconds);
+    static constexpr Duration seconds(const T value);
```
elBoberido, line 108: We have also a `seconds` getter method. I would like to rename the creation methods in a follow up PR and add a `from` prefix, e.g. `fromSeconds` and make the getter methods all lower case.

It would be different from `chrono`, but the alternative would be to have something like a `units_duration_cast`, which I wouldn't like that much all over the code base.

Any objections?

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -186,49 +265,59 @@ class Duration
...
+    multiplyNanoseconds(const uint32_t nanoseconds,
+                        const std::enable_if_t<!std::is_floating_point<T>::value, T>& right) const;
+    template <typename T>
+    inline constexpr Duration
+    multiplyNanoseconds(const uint32_t nanoseconds,
+                        const std::enable_if_t<std::is_floating_point<T>::value, T>& right) const;
+
+  private:
+    uint64_t m_seconds{0};
+    uint32_t m_nanoseconds{0};
```
bishibashiB, line 309: Next to the `double` implementation I found the inability to calculate _negative_ durations (like in received expected vs. actual).
In design you will have chosen for `unsigned` for reasons *)... But it came unexpected to me as there is no such limitation e.g. with `chrono`. (*No negative offsets for timers?)
Did I over-read the documentation explaining it (kindfully point me there, thanks~)?

elBoberido: This was just a straight port to integer types without changing the previous behavior. @elfenpiff once told me why he did it this way, but I forgot it in the meantime. Maybe he can enlighten you.
Regarding the documentation, that's unfortunately missing. This is something we have to improve the next months.

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
+
+TEST(Duration_test, CompareTwoEqualDurations)
 {
     auto time1 = 200_us;
-    auto time2 = 200.0_us;
-    EXPECT_EQ(time1.seconds<double>(), time2.seconds<double>());
+    auto time2 = 200000_ns;
+    EXPECT_TRUE(time1 == time2);
+    EXPECT_TRUE(time2 == time1);
+}
```
FerdinandSpitzschnueffler, line 502: Please add also a test where the equality operator returns false. A paranoid part of me fears that someone in the future might change the behavior of the operator so that it always returns true.

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
 
-    auto time3 = 0.21_ms;
-    EXPECT_LT(time1, time3);
-    EXPECT_GT(time3, time2);
+TEST(Duration_test, CompareTwoNonEqualDurations)
+{
+    auto time1 = 1_s + 200_us;
+    auto time2 = 1_ns;
+    EXPECT_TRUE(time1 != time2);
+    EXPECT_TRUE(time2 != time1);
```
FerdinandSpitzschnueffler, line 509: Please add also a test where the inequality operator returns false.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,47 +21,102 @@ namespace iox
...
 template <typename T>
-constexpr Duration Duration::nanoseconds(const T ns)
+constexpr Duration Duration::nanoseconds(const T value)
 {
-    return operator"" _ns(ns);
+    static_assert(std::numeric_limits<T>::is_integer, "only integer are supported");
+
+    if (value < 0)
+    {
+        return Duration{0U, 0U};
```
FerdinandSpitzschnueffler, line 30: I think a warning that the duration is set to 0 would be good. Same for the other c'tors.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -186,49 +265,59 @@ class Duration
...
+    inline constexpr Duration
+    multiplyNanoseconds(const uint32_t nanoseconds,
+                        const std::enable_if_t<!std::is_floating_point<T>::value, T>& right) const;
+    template <typename T>
+    inline constexpr Duration
+    multiplyNanoseconds(const uint32_t nanoseconds,
+                        const std::enable_if_t<std::is_floating_point<T>::value, T>& right) const;
+
+  private:
+    uint64_t m_seconds{0};
```
dkroenke, line 308: Are U correctly initialized? ;-)

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,47 +21,102 @@ namespace iox
...
+    return operator"" _d(value);
+}
+
+
+inline constexpr Duration::Duration(const uint64_t seconds, const uint32_t nanoseconds)
+    : m_seconds(seconds)
+    , m_nanoseconds(nanoseconds)
 {
-    return operator"" _d(days);
+    if (nanoseconds >= 1000000000U)
```
dkroenke, line 106: `1000000000U` is used quite often as conversion factor in the code, maybe it make sense to make a constexpr of it?
Proposal: `CONV_NANOSECONDS`

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,127 +37,206 @@ class Duration;
...
 
     /// @brief Construct a Duration object from itimerspec
+    /// @param[in] value as itimespec
+    /// @note only it_interval from the itimerspec is used
     constexpr explicit Duration(const struct itimerspec& value);
 
     /// @brief Construct a Duration object from std::chrono::milliseconds
+    /// @param[in] value as milliseconds
+    /// @attention since negative durations are not allowed, the duration will be capped to 0
     constexpr explicit Duration(const std::chrono::milliseconds& value);
```
elfenpiff, line 159: In a follow up pull request we should templatize this somehow so that all `std::chrono::` timetypes are allowed.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
elfenpiff: Duration is a measurement unit for the distance of two time points and it does not make sense to have a negative distance. It is like the distance of two points in space, negative distance is weird.

I think `std::chrono::duration` behaves identical when you call for instance `count`.

But maybe we have to rethink `duration`. For me `units::time` would make more sense somehow with a `units::duration` specialization. It would be very helpful for methods like this:
```cpp
mySemaphore.sleep_until(units::time pointInTIme);
mySemaphore.sleep_for(units::duration duration);
```

@elBoberido what do you think, would this be something for a follow up PR?

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,47 +21,102 @@ namespace iox
...
+constexpr Duration Duration::nanoseconds(const T value)
 {
-    return operator"" _ns(ns);
+    static_assert(std::numeric_limits<T>::is_integer, "only integer are supported");
+
+    if (value < 0)
+    {
+        return Duration{0U, 0U};
+    }
+    return operator"" _ns(value);
```
elfenpiff, line 32: Too much code,
```cpp
return (value < 0 ) ? Duration(0U, 0U) : operator""_ns(value);
```

Would spare us of the four additional lines. Could you please change that for this and all the other methods.

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
+    constexpr uint64_t SECONDS{13U};
+    constexpr uint64_t NANOSECONDS{GIGA};
+    constexpr uint64_t EXPECTED_DURATION_IN_NANOSECONDS{(SECONDS + 1U) * GIGA};
+
+    auto sut = Duration{SECONDS, NANOSECONDS};
+
+    EXPECT_THAT(sut.nanoSeconds<uint64_t>(), Eq(EXPECTED_DURATION_IN_NANOSECONDS));
+}
+
+TEST(Duration_test, ConstructDurationWithMoreNanosecondsThanOneSecond)
```
elfenpiff, line 49: I would add a test where you also construct  a duration object with 3 billion + x nanoseconds which should you then give you 3 seconds.
Additionally, when you think of:
- zero
- one
- many
- corner cases
- limits

You maybe should also construct a duration object with zero nanoseconds, both seconds and nanoseconds are zero, nanoseconds is `std::numeric_limits<uint32_t>::max` and where seconds and nanoseconds are both `std::numeric_limits<T>::max`

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -71,211 +126,298 @@ inline constexpr Duration::Duration(const struct itimerspec& value)
...
+    if (right < static_cast<T>(0))
+    {
+        return Duration{0U, 0U};
+    }
+
+    auto result = m_seconds / right;
+    double seconds{0.0};
+    double secondsFraction = modf(result, &seconds);
+    auto nanoseconds =
+        static_cast<uint64_t>(secondsFraction * 1000000000U) + static_cast<uint64_t>(m_nanoseconds / right);
```
elfenpiff, line 361: I am disappointed, so disappointed!!! ;)

What happens when `right == 0`??

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
+    EXPECT_THAT(2.95 * time, Eq(17_s + 110_ms));
 
-    Duration result(value);
-    EXPECT_EQ(result.microSeconds<uint64_t>(), 42 + 1000000 * 1337);
+    EXPECT_THAT(time * -1, Eq(0_s));
+    EXPECT_THAT(time * -1.0, Eq(0_s));
 }
 
-TEST(Duration_test, isZeroWhenConstructedFromNegativeChronoMilliSeconds)
+TEST(Duration_test, DivideDuration)
```
elfenpiff, line 691: We should test:
- division by zero!
- division with negative number
- division with 0.0000000001f and a large duration type so that you run into an overflow

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
     auto time4 = 10_s;
 
-    result = result / time4;
-    EXPECT_NEAR(result.seconds<double>(), 8.666666666666666, tenTimesEpsilon);
+    result = result - time4;
+    EXPECT_EQ(result.milliSeconds<uint64_t>(), 0U);
 }
 
-TEST(Duration_test, constructFromTimespec)
+TEST(Duration_test, MultiplyDuration)
```
elfenpiff, line 675: We should test:
- multiplication with negative number
- multiplication with 1000000000 and a large duration type so that you run into an overflow

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
+    EXPECT_EQ(result.milliSeconds<uint64_t>(), 5400U);
 
-    result = result * time4;
-    EXPECT_NEAR(result.seconds<double>(), 3.12, tenTimesEpsilon);
+    result = result + time3;
+    EXPECT_EQ(result.microSeconds<uint64_t>(), 6100000U);
 }
 
-TEST(Duration_test, divideDuration)
+TEST(Duration_test, SubtractDuration)
```
elfenpiff, line 655: Overflow test?

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,686 @@
...
+    EXPECT_FALSE(time3 >= time2);
+    EXPECT_FALSE(time4 >= time3);
 }
 
-TEST(Duration_test, multiplyDuration)
+// END COMPARISON TESTS
+
+// BEGIN ARITHMETIC TESTS
+
+TEST(Duration_test, AddDuration)
```
elfenpiff, line 640: Overflow test?

elfenpiff (changes requested):
Please add the mentioned test cases.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
elBoberido: yes, it's initialized with 64 bits set to 0 😉

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
mossmaurice: Seems like Rust's also [not allowing](https://docs.rs/chrono/0.4.6/chrono/struct.Duration.html) a negative duration:

> As standard library implementation is limited to non-negative values.

I'm still unsure on why `std::chrono::duration` can be negative. We might ping @ [HowardHinnant](https://github.com/HowardHinnant) the creator of `chrono` :-)

mossmaurice (commented):
Thanks for this pull request and the increased test coverage. I love seeing technical debt being addressed :rocket: 

However, when starting to think about it, I'm wondering if we're on the right path here.

Why are we reinventing the wheel here? Tried to summarize:

| | :heavy_plus_sign: | :heavy_minus_sign: |
|----|----|----|
|`units::Duration`| Conversion operator needed for `timeval` and `timespec` | Many additional LoC for test |
| | | More certification effort |
| `chrono::duration` | Lower complexity | No user-friendly conversion  |
| | Less certification effort | | 
| | Keeps it simple and stupid| | 

So my conclusion is, why not having the best of both world by wrapping `chrono::duration` and maybe later `chrono::time_point` as wished by @elfenpiff?

elfenpiff:
> Thanks for this pull request and the increased test coverage. I love seeing technical debt being addressed 
> 
> However, when starting to think about it, I'm wondering if we're on the right path here.
> 
> Why are we reinventing the wheel here? Tried to summarize:
> 
> 	
> `units::Duration`	Conversion operator for `timeval` and `timespec`	Many additional LoC for test
> More certification effort
> `chrono::duration`	Lower complexity	No user-friendly conversion
> Less certification effort	
> Keeps it simple and stupid	
> So my conclusion is, why not having the best of both world by wrapping `chrono::duration` and maybe later `chrono::time_point` as wished by @elfenpiff?

I would also prefer `std::chrono`. For the conversion to `timeval` and `timespec` we could write some custom conversion function but before we do that we have to evaluate the following points:
- Exceptions in `std::chrono`, where, why and how?
- Is heap used in `std::chrono`?
- The c++ `std::chrono` library is huge and we only require a small portion of it - is it possible to certify this subset or do we have to pay for the whole thing.

But I think that switching to `std::chrono` is something for the far future and at the moment I would stick to our approach.

mossmaurice:
> [snip]
> ...before we do that we have to evaluate the following points:
> 
>     * Exceptions in `std::chrono`, where, why and how?
> 
>     * Is heap used in `std::chrono`?
> 
>     * The c++ `std::chrono` library is huge and we only require a small portion of it - is it possible to certify this subset or do we have to pay for the whole thing.

Good point and important questions!

Maybe @MichaelOrlov or @sumanth-nirmal already can across this topic and can answer some of the questions.

@elBoberido Shall we maybe start with a design document, which briefly describes our requirements and the path that we wanna take, before merging this PR?

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
elBoberido: @elfenpiff I like the idea. The difference of two `time` objects would be a `duration`. It would need to be truncated to 0 for negative values. Maybe that's also the reason why `chrono::duration` supports negative values.

I think we should follow @mossmaurice suggestion and first fully explore if chrono is an option before we add more features.

@mossmaurice just for completeness. The link you added was to the `chrono` crate and that does allow negative durations. The standard library has `std::time::Duration` which does not allow negative durations, though.

mossmaurice: @elBoberido Ok, thanks for clarifying that. My Rust knowledge is still quite rusty 🤡

elBoberido: @mossmaurice time to change that and make Ferris the 🦀 happy 😁

elBoberido:
> @elBoberido Shall we maybe start with a design document, which briefly describes our requirements and the path that we wanna take, before merging this PR?

@mossmaurice I think a design document is a good idea but I wouldn't wait to merge this PR. It doesn't add new features but is just a refactoring and improves the current situation. Also, regardless if we choose `chrono` as internal representation, I would still keep the current `Duration` class as wrapper. I currently also don't have time to write this design document since I have other things to finish first. So if you don't mind I would continue with this PR and do the design document before new features are added.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
elBoberido: btw, what do you think of
```c++
#define 0 0U
```
or
```c++
constexpr unsigned U{0U};
```

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: there is now a print to inform the user of the clamping, therefore your proposal cannot be used anymore

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: Which division by zero? There is no `operator/` for `Duration` 😉

iceoryx_utils/test/moduletests/test_unit_duration.cpp
elBoberido: can't do, there is no `operator/` for `Duration` 😁

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +37,299 @@ class Duration;
...
+    /// @return a new Duration object
+    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    constexpr Duration operator-(const Duration& right) const noexcept;
+
+    /// @brief creates Duration object by multiplication
+    /// @tparam T is an arithmetic type for the multiplicator
+    /// @param[in] right is the multiplicator
+    /// @return a new Duration object
+    /// @attention since negative durations are not allowed, the duration will be capped to 0
+    /// @note there is no explicit division operator, if you want a fraction of of a Duration, either multiply with the
```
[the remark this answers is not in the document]
elBoberido: instead of skipping letters I'm now duplication words :D

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +37,299 @@ class Duration;
 
 inline namespace duration_literals
 {
-/// @brief constructs a new Duration object in nanoseconds
-constexpr Duration operator"" _ns(long double);
+/// @brief Constructs a new Duration object from nanoseconds
+constexpr Duration operator"" _ns(unsigned long long int) noexcept; // PRQA S 48
```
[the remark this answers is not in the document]
elBoberido: only `long double` and `unsigned long long int` are allowed for user defined literals

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +37,299 @@ class Duration;
...
 class Duration
 {
   public:
+    // BEGIN CREATION FROM STATIC FUNCTIONS
+
+    /// @brief Constructs a new Duration object from nanoseconds
+    /// @tparam T is an integer type for the value
+    /// @param[in] value as nanoseconds
+    /// @return a new Duration object
+    /// @attention since negative durations are not allowed, the duration will be capped to 0
```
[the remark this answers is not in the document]
elBoberido: How about "clamped to"?

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +37,299 @@ class Duration;
...
 
   private:
-    /// @brief constructor needs to be private to ensure a unit safe usage of duration
-    constexpr explicit Duration(const long double durationInSeconds);
-    long double durationInSeconds{0.0};
-};
+    template <typename T>
+    inline constexpr Duration
+    multiplySeconds(const uint64_t seconds,
+                    const std::enable_if_t<!std::is_floating_point<T>::value, T>& right) const noexcept;
```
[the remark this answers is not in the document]
elBoberido: there is already a check for arithmetic types in `operator*(const T& lhs, const Duration& rhs)`. These are some helper functions for integer and floating point multiplicators

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +37,299 @@ class Duration;
...
+    static constexpr uint32_t HOURS_PER_DAY{24U};
+
+    static constexpr uint32_t MILLISECS_PER_SEC{1000U};
+    static constexpr uint32_t MICROSECS_PER_SEC{MILLISECS_PER_SEC * 1000U};
+
+    static constexpr uint32_t NANOSECS_PER_MICROSEC{1000U};
+    static constexpr uint32_t NANOSECS_PER_MILLISEC{NANOSECS_PER_MICROSEC * 1000U};
+    static constexpr uint32_t NANOSECS_PER_SEC{NANOSECS_PER_MILLISEC * 1000U};
+
+    static_assert(NANOSECS_PER_SEC == 1000U * MICROSECS_PER_SEC, "Mismatch in calculation for conversion constants!");
```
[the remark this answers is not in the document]
elBoberido: yes, it's just my inner chicken

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,261 +21,410 @@ namespace iox
...
+    constexpr uint64_t MAX_SECONDS_BEFORE_OVERFLOW{std::numeric_limits<T>::max() / NANOSECS_PER_SEC};
+    constexpr uint64_t MAX_NANOSECONDS_BEFORE_OVERFLOW{std::numeric_limits<T>::max() % NANOSECS_PER_SEC};
+    constexpr Duration MAX_DURATION_BEFORE_OVERFLOW{MAX_SECONDS_BEFORE_OVERFLOW, MAX_NANOSECONDS_BEFORE_OVERFLOW};
+
+    if (*this > MAX_DURATION_BEFORE_OVERFLOW)
+    {
+        std::clog << __PRETTY_FUNCTION__ << ": Result of conversion would overflow, clamping to max value!"
+                  << std::endl;
+        return std::numeric_limits<T>::max();
+    }
```
elBoberido, line 176: @elfenpiff this is the price we are paying for the overflow detection. I think it's not too bad and could be disabled for release builds if necessary

elBoberido:
@mossmaurice a `chrono::duration` as member deletes the defaulted noexcept ctor
```c++
#include <chrono>
#include <iostream>
#include <type_traits>

struct Foo {
    Foo() noexcept = default;

    Foo(const Foo&) noexcept = default;
    Foo(Foo&&) noexcept = default;

    std::chrono::duration<uint64_t, std::nano> duration{0U};
};
int main() {
    std::cout << "Is nothrow default constructible: " << std::is_nothrow_default_constructible<Foo>::value << std::endl;
    std::cout << "Is nothrow move constructible   : " << std::is_nothrow_move_constructible<Foo>::value << std::endl;
}
```
output
```
Is nothrow default constructible: 0
Is nothrow move constructible   : 1
```

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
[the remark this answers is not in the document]
dkroenke: will be fixed to 0?

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
bishibashiB: Thanks all for giving inputs on this! Will follow the discussion interested~!

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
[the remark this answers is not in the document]
elBoberido: other possibilities would be
- bounded to 0
- limited to 0
- narrowed to 0

Maybe we need the help of a native speaker @ithier

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +38,298 @@ class Duration;
...
+    constexpr bool operator>=(const Duration& rhs) const noexcept;
+
+    // END COMPARISON
+
+    // BEGIN ARITHMETIC
+
+    /// @brief Creates Duration object by addition
+    /// @param[in] rhs is the second summand
+    /// @return a new Duration object
+    constexpr Duration operator+(const Duration& rhs) const noexcept;
```
elfenpiff, line 214: Oh ... what about `operator+=`, `operator-=` and `operator*=`?
These are my most favorite operators :D

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,261 +21,490 @@ namespace iox
 namespace units
 {
 template <typename T>
-constexpr Duration Duration::nanoseconds(const T ns)
+constexpr Duration Duration::nanoseconds(const T value) noexcept
 {
-    return operator"" _ns(ns);
+    static_assert(std::numeric_limits<T>::is_integer, "only integer are supported");
```
elfenpiff, line 26: Typo: integers

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,261 +21,490 @@ namespace iox
...
+
+    if (value < 0)
```
elfenpiff, line 28: Since this code is always the same I would create a free function to handle this error handling, otherwise we have a lot of code duplications.

What do you think of the following approach:
```cpp
template<typename T, typename OnSuccessFactory>
Duration createDurationFromValue(const T value, const OnSuccessFactory) {
    static_assert(std::numeric_limits<T>::is_integer, "only integer are supported");
    if (value < 0)
    {
        std::clog << __PRETTY_FUNCTION__ << ": Clampling negative value '" << value << "' to zero!" << std::endl;
        return Duration{0U, 0U};
    }
    return OnSuccessFactory(static_cast<unsigned long long int>(value));
}

template <typename T>
constexpr Duration Duration::nanoseconds(const T value) noexcept {
    return createDurationFromValue(value, [this](const unsigned long long int value){ return operator"" _ns(value)' });
}

```

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -21,261 +21,490 @@ namespace iox
...
-template <typename T>
-inline constexpr T Duration::nanoSeconds() const
+inline constexpr uint64_t Duration::nanoSeconds() const noexcept
 {
-    return static_cast<T>(durationInSeconds * 1000000000.0);
+    constexpr uint64_t MAX_SECONDS_BEFORE_OVERFLOW{std::numeric_limits<uint64_t>::max() / NANOSECS_PER_SEC};
+    constexpr uint64_t MAX_NANOSECONDS_BEFORE_OVERFLOW{std::numeric_limits<uint64_t>::max() % NANOSECS_PER_SEC};
+    constexpr Duration MAX_DURATION_BEFORE_OVERFLOW{MAX_SECONDS_BEFORE_OVERFLOW, MAX_NANOSECONDS_BEFORE_OVERFLOW};
+
+    if (*this > MAX_DURATION_BEFORE_OVERFLOW)
```
elfenpiff, line 168: Here I would maybe also write an internal free function to avoid code duplication.

```cpp
template<typename OnSuccessFunc>
uint64_t performConversion(const Duration & currentDuration, const Duration & maxDuration, const OnSuccessFunc & func) {
  if ( currentDuration > maxDuration ) {
        std::clog << __PRETTY_FUNCTION__ << ": Result of conversion would overflow, clamping to max value!"
                  << std::endl;
        return std::numeric_limits<uint64_t>::max();
  } 
  return OnSuccessFunc();
}

inline constexpr uint64_t Duration::nanoSeconds() const noexcept {
   // ...
   performConversion(*this, 
                      {MAX_SECONDS_BEFORE_OVERFLOW, MAX_NANOSECONDS_BEFORE_OVERFLOW}, 
                      [&]{ return m_seconds * NANOSECS_PER_SEC + m_nanoseconds; } );
}

```

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +38,298 @@ class Duration;
...
+    static constexpr Duration days(const T value) noexcept;
+
+    // END CREATION FROM STATIC FUNCTIONS
+
+    // BEGIN CONSTRUCTORS AND ASSIGNMENT
+
+    /// @brief Constructs a Duration from seconds and nanoseconds
+    /// @param[in] seconds portion of the duration
+    /// @param[in] nanoseconds portion of the duration
+    constexpr Duration(const uint64_t seconds, const uint32_t nanoseconds) noexcept;
```
elfenpiff, line 142: I would make this ctor private since it somehow defeats the purpose of the readability.

```cpp
// good
auto blubb = Duration::seconds(123);
// bad
auto bla = Duration(123, 123); // what is it?
```

If this is really needed I would recommend using the conversion ctor which takes `timeval` or `itimerspec`.

elfenpiff (changes requested):

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
elBoberido: That's a new feature, this PR just replaces the internal double representation with a fixed point one. Feel free to create a PR with your most favorite operator :P

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: use a similar approach without the lambda

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: In this case I'd rather live with minimal code duplication

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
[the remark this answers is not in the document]
MatthiasKillat: I do not really care too much about the wording, it just caught my attention while reading. I think clamp is popular in computer graphics but maybe otherwise unheard of. Do whatever.

iceoryx_posh/test/moduletests/test_gw_gateway_discovery.cpp
```diff
@@ -14,22 +15,25 @@
 
 #include "test.hpp"
 
-#define private public
-#define protected public
```
elBoberido, line 18: only 4 of the old sins left :)

elfenpiff: One of the hereditary sins of iceoryx...

iceoryx_utils/include/iceoryx_utils/internal/units/duration.hpp
```diff
@@ -37,200 +38,311 @@ class Duration;
...
-    friend constexpr Duration duration_literals::operator"" _d(long double);
-    friend constexpr Duration duration_literals::operator"" _d(unsigned long long int); // PRQA S 48
+    template <typename T>
+    constexpr Duration multiplyWith(const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const noexcept;
 
   private:
-    /// @brief constructor needs to be private to ensure a unit safe usage of duration
-    constexpr explicit Duration(const long double durationInSeconds);
-    long double durationInSeconds{0.0};
+    SECONDS_TYPE m_seconds{0U};
```
elfenpiff, line 331: I think we should use UpperCamelCase with a postfix _t for type aliases. UPPER_SNAKE_CASE are for compile time constants and enum values.

@elBoberido  And according to my faint and wrinkled memory you introduced also the string type aliases like `ProcessName_t` ... How could you forget this?! ;) 

I need to verify that it is really you, what do you say to this
```cpp
if ( bla )
   while ( fuu )
       doStuff();
```

Please answer with at least two Bobs and a Pöhnl!

iceoryx_posh/test/moduletests/test_gw_gateway_discovery.cpp
elBoberido: I added a task to the issue with the Windows port. I think two of them are quite trivial to fix, the other two might need some time.

elfenpiff (changes requested):
Please change all `using` type aliases to UpperCamelCase_t

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
 }
 
-inline constexpr Duration operator"" _ms(long double value)
+template <typename T>
+inline constexpr Duration Duration::fromFloatingPointSeconds(const T floatingPointSeconds) const noexcept
 {
-    return Duration{value / 1000.0};
+    static_assert(std::is_floating_point<T>::value, "only floating point is allowed");
+
+    if (std::isinf(floatingPointSeconds))
```
elfenpiff, line 382: Mmhhh. What if Duration == 0? Then you have 0 * infinity? This could become really really weird.

But we get in either case a warning!

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
 
-inline constexpr Duration operator"" _ms(unsigned long long int value) // PRQA S 48
+template <typename T>
+inline constexpr Duration
+Duration::multiplyWith(const std::enable_if_t<std::is_floating_point<T>::value, T>& rhs) const noexcept
 {
-    return Duration{static_cast<long double>(value) / 1000.0};
+    // operator*(...) takes care of negative values for rhs
+
+    if (std::isnan(rhs))
```
elfenpiff, line 408: When you assume that not a number is like infinity, what about the case when Duration == 0? In this case duration should not be set to max - after the operation it should still be zero.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: Well, I decided by a coin flip. Will change it to 0 if Duration is 0

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
 
-template <typename T>
-inline constexpr T Duration::microSeconds() const
+inline constexpr uint64_t Duration::microSeconds() const noexcept
 {
-    return static_cast<T>(durationInSeconds * 1000000.0);
+    constexpr Seconds_t MAX_SECONDS_BEFORE_OVERFLOW{std::numeric_limits<Seconds_t>::max() / MICROSECS_PER_SEC};
+    constexpr Nanoseconds_t MAX_NANOSECONDS_BEFORE_OVERFLOW{(std::numeric_limits<Seconds_t>::max() % MICROSECS_PER_SEC)
+                                                            * NANOSECS_PER_MICROSEC};
+    constexpr Duration MAX_DURATION_BEFORE_OVERFLOW =
```
MatthiasKillat, line 159: We could move the computation of the critical overflow boundary duration to a constexpr function only depending on the ratio (MICROSECS_PER_SEC etc.) Just a minor code duplication reduction though. Not really worth it for 3 uses.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
+
+    return m_seconds * NANOSECS_PER_SEC + m_nanoseconds;
 }
 
-template <typename T>
-inline constexpr T Duration::microSeconds() const
+inline constexpr uint64_t Duration::microSeconds() const noexcept
 {
-    return static_cast<T>(durationInSeconds * 1000000.0);
+    constexpr Seconds_t MAX_SECONDS_BEFORE_OVERFLOW{std::numeric_limits<Seconds_t>::max() / MICROSECS_PER_SEC};
```
MatthiasKillat, line 156: I am not sure the overflow duration is correct in general in this way, since it actually depends on the return type (uint64_t). If this is a type larger than Seconds_t the boundary would be to pessimistic I think. Returning max of uint64 is also generally not correct in those cases (it might actually fit into uint64_t and be less).

Currently it is fine since Seconds_t is uint64_t.

I would not think this is a bug for this reason. Moving on.

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: done

iceoryx_utils/source/units/duration.cpp
```diff
@@ -47,21 +55,26 @@ struct timespec Duration::timespec(const TimeSpecReference& reference) const
         }
         else
         {
-            int64_t timeInNanoSeconds = this->nanoSeconds<int64_t>();
-            int64_t remainingNanoSecondsTimeout = timeInNanoSeconds % NanoSecondsPerSecond;
-            int64_t sumOfNanoSeconds = remainingNanoSecondsTimeout + referenceTime.tv_nsec;
-            int64_t additionalSeconds = sumOfNanoSeconds / NanoSecondsPerSecond;
-            int64_t seconds = timeInNanoSeconds / NanoSecondsPerSecond + referenceTime.tv_sec + additionalSeconds;
-            int64_t nanoSeconds = sumOfNanoSeconds - additionalSeconds * NanoSecondsPerSecond;
+            auto targetTime = Duration(referenceTime) + *this;
```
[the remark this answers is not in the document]
elBoberido: yes, that's correct

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
+    return Duration(seconds, nanoseconds);
+}
+
+inline constexpr Duration Duration::max() noexcept
+{
+    return Duration{std::numeric_limits<Seconds_t>::max(), NANOSECS_PER_SEC - 1U};
+}
+
+template <typename T, typename String>
+inline constexpr unsigned long long int Duration::positiveValueOrClampToZero(const T value,
```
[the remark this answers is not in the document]
elBoberido: even if it takes time, it's just in the unhappy path and logging itself has some costs

iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
```diff
@@ -160,9 +160,13 @@ class Mepoo_IntegrationTest : public Test
 
     void PrintTiming(iox::units::Duration start)
     {
-        std::cerr << "RouDi startup took " << start.milliSeconds<int>() << " milliseconds, "
-                  << "(which is " << start.seconds<int>() << " seconds)"
-                  << "(which is " << start.minutes<int>() << " minutes)" << std::endl;
+        auto totalMillisconds = start.milliSeconds();
+        auto millisceonds = totalMillisconds % 1000U;
```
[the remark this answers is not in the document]
elBoberido: oh, that's refreshing. I did a pöhnl, not a bob :D

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: yes, elfenpiff also suggested it, but the reduction of code would be minimal and I had the feeling it looked uglier.

elfenpiff (dismissed):

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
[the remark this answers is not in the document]
elBoberido: Oh wait. You are right. It was in the unhappy path until there was a request to reduce code duplication :D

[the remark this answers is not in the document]
elBoberido: On the other hand. It's possible to use it in constexpr, so it shouldn't have any extra costs

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: you are right, it should be the return type

elfenpiff (dismissed):

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
+    constexpr uint64_t SECONDS_PER_FULL_BLOCK{LEAST_COMMON_MULTIPLE / NANOSECS_PER_SEC};
+
+    uint64_t multiplicatorHigh = static_cast<uint32_t>(multiplicator >> 32U);
+    auto nanosecondsFromHigh = m_nanoseconds * multiplicatorHigh;
+    auto fullBlocksOfSecondsOnly = nanosecondsFromHigh / ONE_FULL_BLOCK_OF_SECONDS_ONLY;
+    auto remainingBlockWithFullAndFractionalSeconds = nanosecondsFromHigh % ONE_FULL_BLOCK_OF_SECONDS_ONLY;
+
+    auto durationFromNanosecondsHigh = Duration{fullBlocksOfSecondsOnly * SECONDS_PER_FULL_BLOCK, 0U}
+                                       + Duration::nanoseconds(remainingBlockWithFullAndFractionalSeconds << 32);
+
```
MatthiasKillat, line 364: OK.
1) nanosecondsFromHigh fits into 64 bits.
2) fullBlocksOfSecondsOnly fits into 64 Bits, counts only some seconds and no leftover nanoseconds in multiples of ONE_FULL_BLOCK_OF_SECONDS_ONLY
3) remainingBlockWithFullAndFractionalSeconds fits into 64 bit and contains the remaining nanoseconds, may exceed 10⁹ i.e. contain some full seconds

4) fullBlocksOfSecondsOnly * SECONDS_PER_FULL_BLOCK should still fit into 64 bit
5) Can we guarantee that remainingBlockWithFullAndFractionalSeconds << 32 always fits into 64 bits as well?
Otherwise we might have a problem in certain edge cases.

I need to rethink this ...

If 5) causes no problem, it should be correct (operator+ deals with potential further overflows)

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,1919 @@
...
+
+    multiply(duration, -1, EXPECTED_DURATION);
+    multiply(duration, -1.0, EXPECTED_DURATION);
+}
+
+TEST(Duration_test, MultiplyDurationLessThanOneSecondResultsInMoreNanosecondsThan64BitCanRepresent)
+{
+    constexpr uint64_t MULTIPLICATOR{(1ULL << 32U) * 42U + 73U};
+    constexpr Duration DURATION = 473_ms + 578_us + 511_ns;
+    auto EXPECTED_RESULT = DurationFactory{85428177141U, 573034055U};
```
MatthiasKillat, line 1781: Hard to follow without calculator. Ideally would be computed (constexpr) from the multiplicator and the individual parts (nanoseconds and seconds) of Duration.

The point is, the nanoseconds overflow. :-)

MatthiasKillat (dismissed):

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
+    // the result;
+    // this cannot be done by dividing through NANOSECS_PER_SEC, since that one is base 1_000_000_000 and the result is
+    // base 2^32, therefore the least common multiple must be used to get blocks of full seconds represented as 2^32
+    // nanoseconds per bit;
+    // this can then safely be converted to seconds as well as nanoseconds without loosing precision
+
+    // least common multiple of 2^32 and NANOSECONDS_PER_SECOND
+    constexpr uint64_t LEAST_COMMON_MULTIPLE{8388608000000000};
+    static_assert(LEAST_COMMON_MULTIPLE % (1ULL << 32) == 0, "invalid multiple");
+    static_assert(LEAST_COMMON_MULTIPLE % NANOSECS_PER_SEC == 0, "invalid multiple");
```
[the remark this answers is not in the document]
elBoberido: for the following calculation it's not important to be the least common multiple, any common multiple which can be represented in 64 bits will do

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
+    // m_nanoseconds(uint32_t) would fit into 64 bits;
+    // to accurately determine the result, the calculation is split into a multiplication with the lower 32 bits of the
+    // multiplicator and another one with the upper 32 bits;
+
+    // this is the easy part with the lower 32 bits
+    uint64_t multiplicatorLow = static_cast<uint32_t>(multiplicator);
+    auto durationFromNanosecondsLow = Duration::nanoseconds(m_nanoseconds * multiplicatorLow);
+
+    // this is the complicated part with the upper 32 bits;
+    // the m_nanoseconds are multiplied with the upper 32 bits of the multiplicator shifted by 32 bit to the left, thus
```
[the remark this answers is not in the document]
elBoberido: fixed

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -20,262 +20,511 @@ namespace iox
...
+    // having again a multiplication of two 32 bit values whose result fits into a 64 bit variable;
+    // one bit of the result represents 2^32 nanoseconds;
```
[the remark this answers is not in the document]
elBoberido: clarified

iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
elBoberido: yes, it is the modulo of a number less than 2^32 therefore it can safely be shifted

iceoryx_utils/test/moduletests/test_unit_duration.cpp
elBoberido: yes, the problem is the nanoseconds overflow. It's easier to calculate it with a calculator which supports big numbers :)

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,1919 @@
...
-    result = result / time3;
-    EXPECT_NEAR(result.seconds<double>(), 86.666666666666666, tenTimesEpsilon);
+    EXPECT_THAT(sut1, Eq(DurationFactory::max()));
+    EXPECT_THAT(sut2, Eq(DurationFactory::max()));
+}
+
+TEST(Duration_test, AddDurationResultsInSaturationFromSeconds)
+{
+    auto duration1 =
+        DurationFactory{std::numeric_limits<DurationFactory::Seconds_t>::max() - 2U, NANOSECS_PER_SECOND - 1U};
```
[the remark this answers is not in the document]
elBoberido: fixed

iceoryx_utils/test/moduletests/test_unit_duration.cpp
```diff
@@ -18,218 +18,1919 @@
...
 
-TEST(Duration_test, convertDays)
+constexpr uint64_t MILLISECS_PER_SECOND = Duration::MILLISECS_PER_SEC;
+constexpr uint64_t MICROSECS_PER_SECONDS = Duration::MICROSECS_PER_SEC;
+
+constexpr uint64_t NANOSECS_PER_MICROSECOND = Duration::NANOSECS_PER_MICROSEC;
+constexpr uint64_t NANOSECS_PER_MILLISECOND = Duration::NANOSECS_PER_MILLISEC;
+constexpr uint64_t NANOSECS_PER_SECOND = Duration::NANOSECS_PER_SEC;
+
+struct DurationFactory : public Duration
```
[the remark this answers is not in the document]
elBoberido: renamed to `DurationAccessor`

MatthiasKillat (approved):

elfenpiff (approved):

PR #504, opened 2021-01-18
Iox #408 improve typed and untyped api

### LINKED ISSUE #408: Refine modern pub/sub API

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

1. [x] Merge

## References
- Does not close #408, will require follow-up work for cleanup, tests and remaining examples
- Closes n.a.

# Summary
Goal: keep the changes minimal but introduce the new API as discussed
Cleanup will follow in another PR directly after this one

- Changes the API of typed publisher and subscriber as discussed in #408 (variadic emplacement loan instead was discussed internally)
- Keeps the old API intact temporarily to not break all tests and examples before we agree on the final API (hence both APIs exist simultaneously for a short time)
- Changes the icedelivery example to use the 1.0 API, use this to discuss additional improvement potential
- Were overloading was not possible, the suffix _1_0 was introduced, this will change to the method without suffix when the original API is removed in the follow up
- TypedPublisher/Subscriber will be renamed (remove Typed prefix) in a follow up to keep the diff cleaner 
- tests will be adapted to the new API in the follow up
- documentation will be updated in a follow up to reflect the new API
- replacements and removals are marked with "iox-#408" in comments

Whether we want the BasePublisher and BaseSubscriber is debatable but the more I think about it the more I come to the conclusion that it is better to keep it (instead of using untyped as a base for typed and removing the current base). The reason is that this way the typed and untyped API can be changed independently in a cleaner way and it costs us not much (if anything) design-wise (in the worst case untyped just uses all base methods). BasePublisher/Subscriber will just have the common functionality.

### DIFF
--- a/iceoryx_examples/icedelivery/iox_publisher_typed.cpp
+++ b/iceoryx_examples/icedelivery/iox_publisher_typed.cpp
@@ -50,7 +50,7 @@ int main()
         // API Usage #1
         //  * Retrieve a typed sample from shared memory.
         //  * Sample can be held until ready to publish.
-        //
+        //  * Data is default constructed during loan
         auto result = publisher.loan_1_0();
         if (!result.has_error())
         {
@@ -66,11 +66,10 @@ int main()
             // Do something with error
         }
 
-        // API Usage #1
-        //  * Retrieve a typed sample from shared memory and construct data using the
-        //    arguments provided.
+        // API Usage #2
+        //  * Retrieve a typed sample from shared memory and construct data in-place
         //  * Sample can be held until ready to publish.
-        //
+        //  * Data is constructed with the aruments provided.
         result = publisher.loan_1_0(ct, ct, ct);
         if (!result.has_error())
         {
--- a/iceoryx_examples/icedelivery/iox_publisher_typed.cpp
+++ b/iceoryx_examples/icedelivery/iox_publisher_typed.cpp
@@ -66,7 +66,23 @@ int main()
             // Do something with error
         }
 
-        // API Usage #2
+        // API Usage #1
+        //  * Retrieve a typed sample from shared memory and construct data using the
+        //    arguments provided.
+        //  * Sample can be held until ready to publish.
+        //
+        result = publisher.loan_1_0(ct, ct, ct);
+        if (!result.has_error())
+        {
+            result.value().publish();
+        }
+        else
+        {
+            auto error = result.get_error();
+            // Do something with error
+        }
+
+        // API Usage #3
         //  * Retrieve a sample and provide the logic to immediately populate and publish it via a lambda.
         //
         publisher.loan_1_0()
@@ -82,13 +98,13 @@ int main()
             });
 
 
-        // API Usage #3
+        // API Usage #4
         //  * Basic copy-and-publish. Useful for smaller data types.
         //
         auto object = RadarObject(ct, ct, ct);
         publisher.publishCopyOf(object);
 
-        // API Usage #4
+        // API Usage #5
         //  * Provide a callable that will be used to populate the loaned sample.
         //  * The first argument of the callable must be T* and is the location that the callable should
         //      write its result to.
@@ -96,7 +112,7 @@ int main()
         publisher.publishResultOf(getRadarObject, ct);
         publisher.publishResultOf([&ct](RadarObject* object) { *object = RadarObject(ct, ct, ct); });
 
-        std::cout << "Sent five times value: " << ct << std::endl;
+        std::cout << "Sent six times value: " << ct << std::endl;
 
         std::this_thread::sleep_for(std::chrono::seconds(1));
     }
--- a/iceoryx_examples/icedelivery/iox_publisher_typed.cpp
+++ b/iceoryx_examples/icedelivery/iox_publisher_typed.cpp
@@ -39,8 +39,8 @@ int main()
 
     iox::runtime::PoshRuntime::initRuntime("iox-ex-publisher-typed");
 
-    iox::popo::TypedPublisher<RadarObject> typedPublisher({"Radar", "FrontLeft", "Object"});
-    typedPublisher.offer();
+    iox::popo::TypedPublisher<RadarObject> publisher({"Radar", "FrontLeft", "Object"});
+    publisher.offer();
 
     double ct = 0.0;
     while (!killswitch)
@@ -51,7 +51,7 @@ int main()
         //  * Retrieve a typed sample from shared memory.
         //  * Sample can be held until ready to publish.
         //
-        auto result = typedPublisher.loan();
+        auto result = publisher.loan_1_0();
         if (!result.has_error())
         {
             auto& sample = result.value();
@@ -69,7 +69,7 @@ int main()
         // API Usage #2
         //  * Retrieve a sample and provide the logic to immediately populate and publish it via a lambda.
         //
-        typedPublisher.loan()
+        publisher.loan_1_0()
             .and_then([&](auto& sample) {
                 auto object = sample.get();
                 // Do some stuff leading to eventually generating the data in the samples loaned memory...
@@ -86,15 +86,15 @@ int main()
         //  * Basic copy-and-publish. Useful for smaller data types.
         //
         auto object = RadarObject(ct, ct, ct);
-        typedPublisher.publishCopyOf(object);
+        publisher.publishCopyOf(object);
 
         // API Usage #4
         //  * Provide a callable that will be used to populate the loaned sample.
         //  * The first argument of the callable must be T* and is the location that the callable should
         //      write its result to.
         //
-        typedPublisher.publishResultOf(getRadarObject, ct);
-        typedPublisher.publishResultOf([&ct](RadarObject* object) { *object = RadarObject(ct, ct, ct); });
+        publisher.publishResultOf(getRadarObject, ct);
+        publisher.publishResultOf([&ct](RadarObject* object) { *object = RadarObject(ct, ct, ct); });
 
         std::cout << "Sent five times value: " << ct << std::endl;
 
--- a/iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
+++ b/iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
@@ -38,8 +38,8 @@ int main()
     iox::popo::PublisherOptions publisherOptions;
     publisherOptions.historyCapacity = 10U;
 
-    iox::popo::TypedPublisher<RadarObject> typedPublisher({"Radar", "FrontLeft", "Object"}, publisherOptions);
-    typedPublisher.offer();
+    iox::popo::TypedPublisher<RadarObject> publisher({"Radar", "FrontLeft", "Object"}, publisherOptions);
+    publisher.offer();
 
     double ct = 0.0;
     while (!killswitch)
@@ -47,13 +47,7 @@ int main()
         ++ct;
 
         // Retrieve a sample and provide the logic to immediately populate and publish it via a lambda.
-        typedPublisher.loan().and_then([&](auto& sample) {
-            auto object = sample.get();
-            // Do some stuff leading to eventually generating the data in the samples loaned memory...
-            *object = RadarObject(ct, ct, ct);
-            // ...then publish the sample
-            sample.publish();
-        });
+        publisher.loan_1_0(ct, ct, ct).and_then([&](auto& sample) { sample.publish(); });
 
         std::cout << "Sent value: " << ct << std::endl;
 
--- a/iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
+++ b/iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
@@ -46,8 +46,8 @@ int main()
     {
         ++ct;
 
-        // Retrieve a sample and provide the logic to immediately populate and publish it via a lambda.
-        publisher.loan_1_0(ct, ct, ct).and_then([&](auto& sample) { sample.publish(); });
+        // Retrieve a sample, construct it with the given arguments and publish it via a lambda.
+        publisher.loan_1_0(ct, ct, ct).and_then([](auto& sample) { sample.publish(); });
 
         std::cout << "Sent value: " << ct << std::endl;
 
--- a/iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
+++ b/iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
@@ -52,7 +52,7 @@ int main()
                     std::cout << "Got value: " << object->x << std::endl;
                 })
                 .if_empty([] { std::cout << std::endl; })
-                .or_else([](iox::popo::ChunkReceiveError) { std::cout << "Error receiving chunk." << std::endl; });
+                .or_else([](iox::popo::ChunkReceiveResult) { std::cout << "Error receiving chunk." << std::endl; });
         }
         else
         {
--- a/iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
+++ b/iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
@@ -50,14 +50,10 @@ int main()
             subscriber.take_1_0()
                 .and_then([](auto& sample) { std::cout << "Got value: " << sample->x << std::endl; })
                 .or_else([](auto& result) {
-                    // only has to be called if the alternative is of interest
-                    // (i.e. if nothing has to happen when no data is received and
-                    // a possible error alternative is not checked or_else is not needed)
-                    if (result == iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
-                    {
-                        std::cout << "No sample available" << std::endl;
-                    }
-                    else
+                    // only has to be called if the alternative is of interest,
+                    // i.e. if nothing has to happen when no data is received and
+                    // a possible error alternative is not checked or_else is not needed
+                    if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
                     {
                         std::cout << "Error receiving chunk." << std::endl;
                     }
--- a/iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
+++ b/iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
@@ -39,20 +39,29 @@ int main()
     // initialized subscriber
     iox::popo::SubscriberOptions subscriberOptions;
     subscriberOptions.queueCapacity = 10U;
-    iox::popo::TypedSubscriber<RadarObject> typedSubscriber({"Radar", "FrontLeft", "Object"}, subscriberOptions);
-    typedSubscriber.subscribe();
+    iox::popo::TypedSubscriber<RadarObject> subscriber({"Radar", "FrontLeft", "Object"}, subscriberOptions);
+    subscriber.subscribe();
 
     // run until interrupted by Ctrl-C
     while (!killswitch)
     {
-        if (typedSubscriber.getSubscriptionState() == iox::SubscribeState::SUBSCRIBED)
+        if (subscriber.getSubscriptionState() == iox::SubscribeState::SUBSCRIBED)
         {
-            typedSubscriber.take()
-                .and_then([](iox::popo::Sample<const RadarObject>& object) {
-                    std::cout << "Got value: " << object->x << std::endl;
-                })
-                .if_empty([] { std::cout << std::endl; })
-                .or_else([](iox::popo::ChunkReceiveResult) { std::cout << "Error receiving chunk." << std::endl; });
+            subscriber.take_1_0()
+                .and_then([](auto& sample) { std::cout << "Got value: " << sample->x << std::endl; })
+                .or_else([](auto& result) {
+                    // only has to be called if the alternative is of interest
+                    // (i.e. if nothing has to happen when no data is received and
+                    // a possible error alternative is not checked or_else is not needed)
+                    if (result == iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
+                    {
+                        std::cout << "No sample available" << std::endl;
+                    }
+                    else
+                    {
+                        std::cout << "Error receiving chunk." << std::endl;
+                    }
+                });
         }
         else
         {
@@ -62,7 +71,7 @@ int main()
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
     }
 
-    typedSubscriber.unsubscribe();
+    subscriber.unsubscribe();
 
     return (EXIT_SUCCESS);
 }
--- a/iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
+++ b/iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
@@ -39,21 +39,25 @@ int main()
     // initialized subscriber
     iox::popo::SubscriberOptions subscriberOptions;
     subscriberOptions.queueCapacity = 10U;
-    iox::popo::UntypedSubscriber untypedSubscriber({"Radar", "FrontLeft", "Object"}, subscriberOptions);
-    untypedSubscriber.subscribe();
+    iox::popo::UntypedSubscriber subscriber({"Radar", "FrontLeft", "Object"}, subscriberOptions);
+    subscriber.subscribe();
 
     // run until interrupted by Ctrl-C
     while (!killswitch)
     {
-        if (untypedSubscriber.getSubscriptionState() == iox::SubscribeState::SUBSCRIBED)
+        if (subscriber.getSubscriptionState() == iox::SubscribeState::SUBSCRIBED)
         {
-            untypedSubscriber.take()
-                .and_then([](iox::popo::Sample<const void>& sample) {
-                    auto object = static_cast<const RadarObject*>(sample.get());
+            subscriber.take_1_0()
+                .and_then([](const void* data) {
+                    auto object = static_cast<const RadarObject*>(data);
                     std::cout << "Got value: " << object->x << std::endl;
                 })
-                .if_empty([] { std::cout << std::endl; })
-                .or_else([](iox::popo::ChunkReceiveResult) { std::cout << "Error receiving chunk." << std::endl; });
+                .or_else([](auto& result) {
+                    if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
+                    {
+                        std::cout << "Error receiving chunk." << std::endl;
+                    }
+                });
         }
         else
         {
@@ -63,7 +67,7 @@ int main()
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
     }
 
-    untypedSubscriber.unsubscribe();
+    subscriber.unsubscribe();
 
     return (EXIT_SUCCESS);
 }
--- a/iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
+++ b/iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
@@ -48,9 +48,13 @@ int main()
         if (subscriber.getSubscriptionState() == iox::SubscribeState::SUBSCRIBED)
         {
             subscriber.take_1_0()
-                .and_then([](const void* data) {
+                .and_then([&](const void* data) {
                     auto object = static_cast<const RadarObject*>(data);
                     std::cout << "Got value: " << object->x << std::endl;
+
+                    // note that we explicitly have to release the data
+                    // and afterwards the pointer access is undefined behavior
+                    subscriber.releaseChunk(object);
                 })
                 .or_else([](auto& result) {
                     if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
--- a/iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
+++ b/iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
@@ -53,7 +53,7 @@ int main()
                     std::cout << "Got value: " << object->x << std::endl;
                 })
                 .if_empty([] { std::cout << std::endl; })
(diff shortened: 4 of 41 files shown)


MatthiasKillat:
I am not exceptionally happy with ChunkReceiveResult in the API as the alternative of the expected, also the renaming may have been a bit too early. It is what it is though...

I try to come up with of something better for the empty data handling (NO_CHUNKS_AVAILABLE). We may have to rename it to NO_DATA_AVAILABLE at least for the public API, the typed API for example should not really talk about chunks...

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -96,6 +109,16 @@ inline void Sample<T>::release() noexcept
     m_samplePtr.release();
 }
 
+template <typename T>
+template <typename... Args>
+inline void Sample<T>::emplace(Args&&... args) noexcept
```
MatthiasKillat, line 114: We should probably get rid of this, I left it here but there is currently no way to obtain an uninitialized sample as far as I am aware. What do you think @elBoberido

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -48,6 +48,21 @@ class Sample
     ///
     T* operator->() noexcept;
 
+    ///
+    /// @brief operator* Provide a reference to the encapsulated type.
+    /// @return A T& to the encapsulated type.
+    /// @details Only available for non void type T.
+    ///
+    template <typename S = T,
```
MatthiasKillat, line 56: We can probably eliminate this when we no longer instantiate Sample<void>. (Done for the current API). It is the only reason for this SFINAE magic right now.

iceoryx_posh/include/iceoryx_posh/popo/untyped_subscriber.hpp
```diff
@@ -48,13 +48,43 @@ class UntypedSubscriberImpl
...
+
+    // iox-#408
+    // the untyped API is supposed to deal with chunks, hence the renaming iox #408 remove comment
+    // calling it chunks looks inappropriate in the function names (use data instead of chunks?)...
+
+    ///
+    /// @brief hasChunks Check if chunks are available.
+    /// @return True if a new chunk is available.
+    ///
+    bool hasChunks() const noexcept;
```
MatthiasKillat, line 75: I was also thinking about better names for the functions in the untyped case, maybe hasData? hasChunks does not look right, but neither does hasSamples in the untyped case. I am open for suggestions but also do not consider it a major problem.

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
```diff
@@ -83,31 +83,49 @@ class BaseSubscriber
     ///
     void unsubscribe() noexcept;
 
+    // iox-#408 replace
     ///
     /// @brief hasData Check if sample is available.
     /// @return True if a new sample is available.
     ///
     bool hasSamples() const noexcept;
```
elBoberido, line 91: do you really want to completely replace this? This is also useful for the typed publisher

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_subscriber.inl
```diff
@@ -26,6 +26,35 @@ inline UntypedSubscriberImpl<base_subscriber_t>::UntypedSubscriberImpl(const cap
...
 
+template <template <typename, typename, typename> class base_subscriber_t>
+inline cxx::expected<const void*, ChunkReceiveResult> UntypedSubscriberImpl<base_subscriber_t>::take_1_0() noexcept
+{
+    auto result = BaseSubscriber::takeChunk();
+    if (result.has_error())
+    {
+        return cxx::error<ChunkReceiveResult>(result.get_error());
+    }
+    return cxx::success<const void*>(result.value()->payload());
```
elBoberido, line 37: payload should probably return a `const void*` when called from a `const ChunkHeader*`

iceoryx_posh/include/iceoryx_posh/internal/popo/typed_subscriber.inl
```diff
@@ -26,6 +26,19 @@ inline TypedSubscriber<T, base_subscriber_t>::TypedSubscriber(const capro::Servi
...
+template <typename T, template <typename, typename, typename> class base_subscriber_t>
+inline cxx::expected<Sample<const T>, ChunkReceiveResult> TypedSubscriber<T, base_subscriber_t>::take_1_0() noexcept
+{
+    auto result = BaseSubscriber::takeChunk();
+    if (result.has_error())
+    {
+        return cxx::error<ChunkReceiveResult>(result.get_error());
+    }
+    auto payloadPtr = reinterpret_cast<T*>(result.value()->payload());
+    auto samplePtr = cxx::unique_ptr<T>(reinterpret_cast<T*>(payloadPtr), BaseSubscriber::m_sampleDeleter);
```
elBoberido, line 38: you should be able to achieve the same with one `static_cast<T*>`

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
elBoberido: yes, remove it. It not necessary since loan now takes variadic arguments

iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
```diff
@@ -38,22 +38,16 @@ int main()
...
 
         // Retrieve a sample and provide the logic to immediately populate and publish it via a lambda.
-        typedPublisher.loan().and_then([&](auto& sample) {
-            auto object = sample.get();
-            // Do some stuff leading to eventually generating the data in the samples loaned memory...
-            *object = RadarObject(ct, ct, ct);
-            // ...then publish the sample
-            sample.publish();
-        });
+        publisher.loan_1_0(ct, ct, ct).and_then([&](auto& sample) { sample.publish(); });
```
elBoberido, line 50: this reads quite nice

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
```diff
@@ -39,20 +39,25 @@ int main()
...
+                .and_then([](auto& sample) { std::cout << "Got value: " << sample->x << std::endl; })
+                .or_else([](auto& result) {
+                    // only has to be called if the alternative is of interest,
+                    // i.e. if nothing has to happen when no data is received and
+                    // a possible error alternative is not checked or_else is not needed
+                    if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
+                    {
+                        std::cout << "Error receiving chunk." << std::endl;
+                    }
+                });
```
elBoberido, line 60: Maybe something for the future. What do you thing about a `or_else_if(E c, [](E& r))` and `or_else_if_not(E c, [](E& r))` which calls the lambda only if `c == r` or `c != r`. It might be helpful for such cases where you want to ignore some of the error cases or act only on a specific error case.

It might be dangerous though, since you could still call `or_else` afterwards and it would still be executed, which is semantically different than an `else if` statement. This could be solved if `or_else` would return a reference to itself, but a new `expected` when executed

iceoryx_examples/icedelivery/iox_subscriber_with_history.cpp
```diff
@@ -43,32 +43,30 @@ int main()
...
             do
             {
-                typedSubscriber.take()
-                    .and_then([](iox::popo::Sample<const RadarObject>& object) {
-                        std::cout << "Got value: " << object->x << std::endl;
-                    })
-                    .if_empty([&] { hasMoreSamples = false; });
+                subscriber.take_1_0()
+                    .and_then([](auto& object) { std::cout << "Got value: " << object->x << std::endl; })
+                    .or_else([&](auto&) { hasMoreSamples = false; });
```
elBoberido, line 61: This is semantically different from the old code. But I think more appropriate to prevent an infinite loop it take always returns an error.

iceoryx_posh/include/iceoryx_posh/popo/untyped_subscriber.hpp
elBoberido: since all the other methods also call it `Chunks` I would keep it for consistency reasons or rename everything to `*Data`

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
```diff
@@ -37,9 +37,11 @@ class UntypedPublisherImpl : public base_publisher_t
     using base_publisher_t::getServiceDescription;
     using base_publisher_t::getUid;
     using base_publisher_t::hasSubscribers;
-    using base_publisher_t::isOffered;
-    using base_publisher_t::loan;
-    using base_publisher_t::loanPreviousSample;
+    using base_publisher_t::isOffered; // iox-#408 better hasOffered ?
```
ithier, line 40: `hasOffered` is not better IMO

iceoryx_posh/include/iceoryx_posh/popo/untyped_subscriber.hpp
ithier: I understood that we wanted the users of the untyped API to actually be thinking in terms of memory chunks when reasoning about the logic.
See: https://github.com/eclipse-iceoryx/iceoryx/issues/408#issuecomment-757910091

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
elBoberido: `isOffering` or `doesOffer`?

iceoryx_examples/icedelivery/iox_publisher_typed.cpp
```diff
@@ -39,8 +39,8 @@ int main()
 
     iox::runtime::PoshRuntime::initRuntime("iox-ex-publisher-typed");
 
-    iox::popo::TypedPublisher<RadarObject> typedPublisher({"Radar", "FrontLeft", "Object"});
-    typedPublisher.offer();
+    iox::popo::TypedPublisher<RadarObject> publisher({"Radar", "FrontLeft", "Object"});
```
budrus, line 42: Renaming of `TypedPublisher` into `Publisher` and `UntypedPublisher` in `RawPublisher` will be done in the next PR, right?

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
```diff
@@ -39,20 +39,25 @@ int main()
...
-                .or_else([](iox::popo::ChunkReceiveError) { std::cout << "Error receiving chunk." << std::endl; });
+            subscriber.take_1_0()
+                .and_then([](auto& sample) { std::cout << "Got value: " << sample->x << std::endl; })
+                .or_else([](auto& result) {
+                    // only has to be called if the alternative is of interest,
+                    // i.e. if nothing has to happen when no data is received and
+                    // a possible error alternative is not checked or_else is not needed
+                    if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
+                    {
+                        std::cout << "Error receiving chunk." << std::endl;
```
budrus, line 58: `NO_CHUNK_AVAILABLE `is no error IMHO. You just polled for new samples and there is no one

iceoryx_examples/icedelivery/iox_subscriber_untyped.cpp
```diff
@@ -39,21 +39,25 @@ int main()
...
+                .and_then([](const void* data) {
+                    auto object = static_cast<const RadarObject*>(data);
                     std::cout << "Got value: " << object->x << std::endl;
                 })
-                .if_empty([] { std::cout << std::endl; })
-                .or_else([](iox::popo::ChunkReceiveError) { std::cout << "Error receiving chunk." << std::endl; });
+                .or_else([](auto& result) {
+                    if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
+                    {
+                        std::cout << "Error receiving chunk." << std::endl;
```
budrus, line 58: same same

iceoryx_posh/include/iceoryx_posh/internal/popo/base_subscriber.inl
```diff
@@ -108,6 +108,27 @@ BaseSubscriber<T, Subscriber, port_t>::take() noexcept
...
+    }
+    else
+    {
+        auto maybeHeader = result.value();
+        if (maybeHeader.has_value())
+        {
+            return cxx::success<const mepoo::ChunkHeader*>(maybeHeader.value());
+        }
+    }
+    // could move this to a tryGetChunk but then we should remove expected<optional<>> there in the call chain
```
budrus, line 128: would this be a runtime benefit @elBoberido ?

iceoryx_posh/include/iceoryx_posh/internal/popo/base_subscriber.inl
```diff
@@ -124,6 +145,13 @@ inline void BaseSubscriber<T, Subscriber, port_t>::invalidateTrigger(const uint6
     }
 }
 
+template <typename T, typename Subscriber, typename port_t>
+void BaseSubscriber<T, Subscriber, port_t>::releaseChunk(const void* payload) noexcept
+{
+    auto header = mepoo::ChunkHeader::fromPayload(payload);
+    m_port.get().releaseChunk(header);
```
budrus, line 152: Why do we need a get() call here and not in all the other methods?

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
```diff
@@ -83,31 +83,49 @@ class BaseSubscriber
...
     ///
-    cxx::expected<cxx::optional<Sample<const T>>, ChunkReceiveError> take() noexcept;
+    cxx::expected<cxx::optional<Sample<const T>>, ChunkReceiveResult> take() noexcept;
 
+    ///
+    /// @brief takeChunk Take the chunk from the top of the receive queue.
+    /// @return The header of the chunk taken.
+    /// @details No automatic cleaunp of the associated chunk is performed.
+    ///
+    cxx::expected<const mepoo::ChunkHeader*, ChunkReceiveResult> takeChunk() noexcept;
```
budrus, line 114: why do we have a `takeChunk` with `ChunkHeader*` and a `relaeseChunk` with `void*`. Should be symmetric here, or?

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
elfenpiff: Oh this reminds me of the old days: `pop_if`. But in this case I would vote against it. I played around with your idea and in my opinion the code is less readable then with an `if`-statement. Take a look:
```cpp
.or_else([](auto& result) {
                    if (result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
                    {
                        std::cout << "Error receiving chunk." << std::endl;
                    });
```
```cpp
.or_else_if( [](auto & result) { return result != iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE; }
            ,[](auto& result) { std::cout << "Error receiving chunk." << std::endl; });
```

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
elfenpiff: @budrus but it states if result is not equal to `NO_CHUNK_AVAILABLE` ;) or did I miss something

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -64,6 +64,19 @@ inline T* Sample<T>::operator->() noexcept
     return get();
 }
 
+template <typename T>
+template <typename S, typename>
+inline S& Sample<T>::operator*() noexcept
```
elfenpiff, line 69: Should we also introduce const overloads for `operator*`, `operator->`? Since it makes sense for the publisher to call maybe some kind of method with a const sample reference argument?

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -64,6 +64,19 @@ inline T* Sample<T>::operator->() noexcept
...
+{
+    return *get();
+}
+
+template <typename T>
+inline Sample<T>::operator bool()
```
elfenpiff, line 75: This method should be const.

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
budrus: Sorry, my fault

iceoryx_posh/include/iceoryx_posh/internal/popo/base_subscriber.inl
elBoberido: could be a minimal runtime benefit. But even if it wouldn't, it still makes sense to also change the `tryGetChunk` method since the nested optional is not used further down in the code

iceoryx_posh/include/iceoryx_posh/popo/untyped_subscriber.hpp
MatthiasKillat: I got the same impression @ithier , but I am open for proposals. When there are none, I will leave it at chunks.

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_subscriber.inl
MatthiasKillat: Then we need another overload for payload.

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
MatthiasKillat: I was thinking about having it called chunks in the base and untyped and samples in the typed class. Or just have both in the base and pull in the one needed in the derived class.

iceoryx_posh/include/iceoryx_posh/internal/popo/typed_subscriber.inl
MatthiasKillat: Yes. It is just a pointer. I prefer reinterpret_casts when casting void pointers to typed pointers, it reminds me that something low level is going on.

iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
MatthiasKillat: Yes it is short but the one thing that irks me is I do not naturally associate loaning with construction.

iceoryx_examples/icedelivery/iox_subscriber_with_history.cpp
MatthiasKillat: It is, but if there is an error there are also no samples...any proposal how it should behave?

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
MatthiasKillat: Just because the entity is a publisher, which either has or has not offered a topic. I can live with both but if we want to change this now is the time.

iceoryx_examples/icedelivery/iox_publisher_typed.cpp
MatthiasKillat: Yes, this change would propagate through all examples and many tests.

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
MatthiasKillat: The idea was that this is just the base and not exposed to the user. But releaseChunk in base at least should be symmetric, right. I will reorganize it somewhat.

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
MatthiasKillat: Can probably do this.

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
MatthiasKillat: Yes

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
elBoberido: @elfenpiff you misunderstood me. My idea was more like this
```c++
.or_else_if_not(iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE, [](auto& result) {
    std::cout << "Error receiving chunk." << std::endl; 
})
```

It's problematic that you still could use `or_else_if(...).or_else(...)` and both of them could be executed. So, a flag would be needed to remember if the `else` branch was already executed. This might still be a good idea, even without implementing `or_else_if` to catch a `.or_else().or_else()` chain.

iceoryx_examples/icedelivery/iox_subscriber_with_history.cpp
elBoberido: I already think this is a better solution. It also breaks the loop in case of an error.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
elBoberido: I find it more intuitive to think about the present, not the past. So I would prefer a `is` or `does` prefix. Depending o the point of view, one could say the service of the publisher `isOffered` or the publisher `isOffering/doesOffer` a service

iceoryx_posh/include/iceoryx_posh/internal/popo/typed_subscriber.inl
elBoberido: I was just asking because some guidelines restrict the usage of `reinterpret_cast` and then we would need to write exceptions which wouldn't be necessary when using `static_cast`

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_subscriber.inl
elBoberido: oh right, this comes from the `ChunkHeader`. I'll add those overload with my feature. I also misread this as a cast, but it's actually just creating an `expected`.

iceoryx_posh/include/iceoryx_posh/internal/popo/base_subscriber.inl
MatthiasKillat: If I do this now it will impact several tests. Maybe in the follow-up?

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
MatthiasKillat: @elBoberido I would consider it to be too confusing. I envisioned returning a ``ReceiveResult<T, ReceiveStatus>`` which inherits from ``expected<T, ReceiveStatus>`` and hence provides ``and_then`` etc. In addition it could have ``if_empty`` (which would internally check the Status) or what have you. This is also quite extension friendly for the receive use case, as it does not need to be too general. We could also have a slightly more specific ``if_status(Status, action)`` or general form ``if(condition, action)``. 

I already proposed this but nobody wanted to listen so...

MatthiasKillat:
@elBoberido @budrus What do you think about renaming ChunkReceiveResult to ReceiveResult or ReceiveStatus? I dislike that it mentions Chunks even in the typed API.

iceoryx_posh/include/iceoryx_posh/internal/popo/typed_subscriber.inl
budrus: I think I already saw code analyzer tools that complain that `reinterpret_cast` is evil. So if a `static_cast` will do, I would go for this

iceoryx_examples/icedelivery/iox_publisher_with_history.cpp
MatthiasKillat: Capture & is not needed here.

budrus:
> @elBoberido @budrus What do you think about renaming ChunkReceiveResult to ReceiveResult or ReceiveStatus? I dislike that it mentions Chunks even in the typed API.

@MatthiasKillat On the public API we renamed the methods to take.. Should we then have a `TakeResult`? If the same type is used in lower layers, we maybe rename get.. to take also there. On the public API we should be consistent and

iceoryx_posh/include/iceoryx_posh/internal/popo/typed_subscriber.inl
MatthiasKillat: Yes. At some places it will be needed though not here. I find it somewhat hides critical information, it is not comparable to say casting a Derived* to a Base* or a int32_t to an int64_t (which static_cast is mainly used for).

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_subscriber.inl
MatthiasKillat: Yes, I did not want to introduce changes to Chunkheader and will hence do nothing here.

iceoryx_examples/icedelivery/iox_subscriber_with_history.cpp
MatthiasKillat: One could distinguish between the error case and no data, but this is already done in the typed subscriber example.

iceoryx_posh/include/iceoryx_posh/internal/popo/base_subscriber.inl
MatthiasKillat: I'll mark it as some improvement but will take no action here.

elBoberido:
> > @elBoberido @budrus What do you think about renaming ChunkReceiveResult to ReceiveResult or ReceiveStatus? I dislike that it mentions Chunks even in the typed API.
> 
> @MatthiasKillat On the public API we renamed the methods to take.. Should we then have a `TakeResult`? If the same type is used in lower layers, we maybe rename get.. to take also there. On the public API we should be consistent and then `TakeResult` or `TakeStatus` would be fine for all API versions, or?

I thought it's `takeChunk` for the untyped API and just `take` for the typed API? Nevertheless, I wouldn't add `Result` or `Status` to the method name. That's just one possible ... well, result of the function. The other, more expected one is a sample/chunk.

Just `ReceiveResult` might be too generic. On the other hand, it's in the `popo` namespace, so potential name clashes/ambiguities are minimal.

MatthiasKillat:
@budrus Depends on were this enum is used and I would try to avoid having multiple enums that essentially distinguish the same cases. I tend more to *Status since for me the result is what I get back from the call and it is either usable data (chunk, sample) or some status information (avoid to call it error when no data is possible). 

You could say it is weird not to have this information in the success case as well, compared with many APIs, but it would just be an ``OK`` or something, so not needed.

elBoberido:
@MatthiasKillat don't take this serious, but I just thought about calling it `UnhappyTakeResult` or `UnhappyTakeStatus` since you only get it on the unhappy path 😁

MatthiasKillat:
Not helpful but you know this. It is just a naming matter, the call structure/signature of the methods are fine from my point of view so (I hope we can agree on this, if NOT, now is the time to provide counterproposals).

 I will just wait and change it do the final outcome when everything else has been resolved.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
ithier: > one could say the service of the publisher `isOffered`
This was my thinking, and I personally prefer it to "is offering".

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
MatthiasKillat: If desired it can be added later, it will not invalidate the current usage.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
MatthiasKillat: I leave it as is.

iceoryx_posh/include/iceoryx_posh/popo/base_subscriber.hpp
MatthiasKillat: It would be the best to convert only once and as late as possible(little efficiency gain).. It makes no sense to convert back and forth between header and payload and the low level API provides the header.

iceoryx_examples/icedelivery/iox_subscriber_typed.cpp
elBoberido: @MatthiasKillat it was just food for thought. I think with `and_then` and `or_else` currently just returning a reference it might easily lead to errors due to unintended double execution.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -96,9 +132,17 @@ class Sample<const T>
...
+              typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_same<S, void>::value, S>>
+    const S& operator*() noexcept;
+
     const T* get() noexcept;
     const mepoo::ChunkHeader* getHeader() noexcept;
 
+    template <typename... Args>
+    void emplace(Args&&... args) noexcept = delete;
+
```
elBoberido, line 145: I think this one can be removed. It is already removed from the `.inl`

elBoberido (approved):

budrus (approved):

Issue #350, opened 2020-11-10
New chunk available callback for the new C++ and C APIs 


budrus:
## Brief feature description

So far the new API allows for "normal" polling access and the usage in combination with a waitset. It shall also support the registration of callbacks similar to the old API

## Detailed information
The old API had the receive handler for this [see](https://github.com/eclipse/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/popo/subscriber.hpp#L79). But instead of a thread per subscriber we should go for a thread in the runtime that handles the callbacks (could even be extended to a configurable thread pool in future) 

- [x] callbacks available for modern C++ API

- [x] callbacks available for C API

## ToDo

- [x] Write design proposal with an usage example
- [x] `enableEvent` in WaitSet attachable classes should have a signature so that they can be reused by the ReactAl. Change the signature to: 
```cpp
/// @return invalidation callback used by the WaitSet and ReactAl
cxx::MethodCallback<void, uint64_t> enableEvent(
                     const EventTypeEnum event, 
                     iox::popo::TriggerHandle & triggerHandle);
```

- [x] Implement design proposal
- [x] Rename `ActiveCallSet` into `Listener`
- ~~[ ] Create use case based comparison table between WaitSet and callback proposal~~ -> #482
- [x] Update example and write a user documentation
- [x] Add hint that the Listener does not take ownership of the callback. (In doxygen and user docu)
- [x] Rename EventInfo into NotificationInfo
- [x] Update design document to be in sync with the final implementation
- ~~[ ] Update Waitset and Listener section in [overview.md](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/getting-started/overview.md)~~ -> #482 
- [x] Separate WaitSet State enums from Listener Event enums (`STATE_HAS_DATA`, `EVENT_DATA_RECEIVED`)
- [x] `hasTriggered`/`resetTrigger` refactoring in user trigger since it is automatically resetted

### Minor ToDo
- ~~[ ] `active_call_set.hpp` maybe merge indexmanager functionality into loffli~~ -> #680
- [x] Remove `doesContainEventVariable` from `TriggerHandle`.
- [x] Remove `setEventVariable` from `ChunkQueuePopper`.
- ~~[ ] Consider using the `index_queue` instead of `loffli` as `active_call_set` indexmanager~~ -> #680
- [x] do not iterate over array in `addEvent` ask event origin instead if event is already attached

[...]

PR #508, opened 2021-01-19
iox-#350 event based callbacks - design proposal

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

1. [ ] Merge

## References
#350


doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,171 @@
+## Name proposals
+
+ - ReactAl = React And Listen
```
elBoberido, line 3: With `React And Listen` the causality is violated and our universe might implode. This is more a `Listen and React` -> `LiAR`

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,171 @@
...
+
+    **Reason:**
+      - The reactal will iterate through this array to know who was triggering it 
+      - It is cache local so very fast to iterate.
+      - WaitSet approach are callbacks which have overhead implementation 
+        and performance wise.
+
+    **Alternative:**
+      - Create a new class or provide a template bool array size argument to the 
+        ConditionVariable to be more flexible and memory efficient.
```
mossmaurice, line 105: Why not extending the `ConditionVariable` classes like we are doing it with the chunk building blocks. I'm not happy with adapting the condition variable API. We should try to follow the STL and extend the STL implementation e.g. `NotifySpecific(id)`. IMHO I would name this derived class `EventSignaler`.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,171 @@
...
+which event is attached, like in the _WaitSet_.
+
+It should have the following behavior:
+
+ - Whenever an event occurs the corresponding callback should be called **once**
+    immediately.
+ - If an event occurs multiple times before the callback was called, the callback
+    should be called **once**.
+ - If an event occurs while the callback is being executed the callback should be 
+    called again **once**.
```
mossmaurice, line 22: Nicely pointed out :sunflower:  However, we should define a name for this behaviour. I'm quite font of the name `CatchupPolicy`. What you are describing is probably "ASAP" and "N_TO_1" behaviour. As the `WaitSet` is the sibling of ReactAl, we could add such a description there, too.

mossmaurice (commented):
Thanks for the write-up!

A general remark: Could we add a design template with this pull request? That would be helpful for other contributors.

I suggest a structure similar to what @elBoberido used:

```
# Name of design element

## Summary

## Terminology

## Design

### Considerations

### Solution

## Open issues


```

I'm happy to add it if you don't mind. E.g. I'm missing a bit the introduction for people not familar with this topic and the why in a few sentences. You could link to the terminology of the WaitSet or maybe create a common one?


**Open questions:**
* If it is similar to the `WaitSet` why not re-using more parts of it?
  * AFAICT you're only planning to use and extend the `ConditionVariableSignaler`
* Could the `ice_waitset_trigger.cpp` example be implemented with ReactAl?
  * My current understanding is that this example could be implemented with both. Hence, I have the feeling that we're violating KISS (keep it simple & stupid) here. The useser shouldn't have more than one solution per use-case

doc/design/eventdriven-callbacks.md
elfenpiff: - ASAP describes when you react (as soon as possible) and not how many times.
- N_TO_1 is when you channel N notifications to one recipient which is the Reactor Pattern and we have two solutions for it, the WaitSet and the upcoming "ReactAl"
- `CatchupPolicy` meant in the timer, if you were activated 5 times and didn't call the callback once then call the callback 5 times and then only every time you were activated.

We are following here the typical Reactive Programming approach which means an event causes a callback (once). We  could think of a name for this kind of behavior but it would most likely not end up in some C++ code since this behavior is tightly coupled to the ReactAl and changing the behavior via template/ctor arg would make the ReactAl even more complex as it is right now.

But if you have a nice fitting name I would welcome it very much!

doc/design/eventdriven-callbacks.md
elfenpiff: I added a reasoning for this approach.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+
+
+## Terminology
+
+ - **event driven** a one time reaction which is caused directly by an event.
+      Example: a new sample has been delivered to a subscriber.
+ - **state driven** a repeating reaction which is caused by a specific structural 
+      state until the state changes.
+      Example: a subscriber has stored samples which were not inspected by the user.
+ - **robust API** we use this term to describe an API which is nearly impossible
```
mossmaurice, line 21: I guess we should avoid the term "nearly". This can be misunderstood. What about "Keep the misuse potential to a minimum"?

doc/design/eventdriven-callbacks.md
mossmaurice: Ok, I think we talked past each other a bit here. I didn't mean to introduce an enum here, rather having some policies like the DDS QoS one. Then we could formally write down which reactive pattern fulfill which policy to make it easier for the user to choose and to have a clear differentiation for use-cases. Still searching for good names, though :-/

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+  myCallbackReactal.attachEvent(mySubscriber, iox::popo::SubscriberEvent::SUBSCRIBED, waitForSubscription);
+}
+
+```
+
+### Solution
+#### Condition Variable
+
+  - Create from `ConditionVariableWithOriginTracing` (yeah we need a better name) has `ConditionVariableData` 
+    as a member and adds: `std::atomic_bool m_triggeredBy[MAX_CALLBACKS];`
```
mossmaurice, line 117: :heart:

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+    **Reason:** it behaves quiete differently then the `ConditionVariable`. The `ConditionVariable` tells you:
+    someone has signalled you and you have to find out who it was. The `ConditionVariableWithOriginTracing`
+    would tell you: A, B and C have signalled you.
+
+    ```cpp
+    class ConditionVariableWithOriginTracingSignaler {
+      public:
+        // id = corresponds to the id in the ConditionVariable atomic_bool array to 
+        // identify the trigger origin easily
+        void notifyOne(const uint64_t id);
```
mossmaurice, line 151: NotifySpecifc?

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+int main() {
+  myCallbackReactal.attachEvent(mySubscriber, iox::popo::SubscriberEvent::SUBSCRIBED, waitForSubscription);
+}
+
+```
+
+### Solution
+#### Condition Variable
+
+  - Create from `ConditionVariableWithOriginTracing` (yeah we need a better name) has `ConditionVariableData`
```
mossmaurice, line 116: `EventSignaler` `EventWaiter`?

doc/design/eventdriven-callbacks.md
mossmaurice: Misunderstood it: id = who is notifying

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+      const cxx::function_ref<void(EventOrigin&)> & callback
+    ) noexcept;
+
+    template<typename EventOrigin, typename EventType>
+    void detachEvent( EventOrigin & origin, const EventType & eventType );
+};
+
+```
+ 
+  - Contains a thread which will wake up and iterate through ConditionVariable
```
mossmaurice, line 191: If this a user-thread does this has any implication on the real-time behaviour on QNX?

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+```cpp
+class SomeSubscriber {
+  //...
+  // index is the index to the bool inside of the condition variable, maybe it 
+  // is later hidden inside of some abstraction
+  //
+  // OutOfScopeCallback_t - callback which should be called when SomeSubscriber
+  //                        goes out of scope and has to deregister itself from 
+  //                        the reactal
+  void attachCallback(SubscriberEvent event, ConditionVariable & cv, ConditionVariable::index_t & index
```
mossmaurice, line 206: Update example?

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+      Exception: If a callback detaches itself it blocks until the callback is removed 
+                 and not until the callback is finished.
+    - Calling detach means that after that call the callback is no longer attached
+      even when it was not attached in the first place. Therefore it will always succeed.
+
+ - When the ReactAl goes out of scope it should detach itself from every class 
+     to which it was attached via a provided callback (like in the WaitSet).
+
+ - When the class which is attached to the ReactAl goes out of scope it should 
+    detach itself from the ReactAl via a provided callback (like in the WaitSet).
```
mossmaurice, line 80: Does it make sense to have a common base class for the WaitSet and ReactAl?

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+      public:
+        // id = corresponds to the id in the ConditionVariable atomic_bool array to 
+        // identify the trigger origin easily
+        void notifyOne(const uint64_t id); 
+    };
+    ```
+
+  - add class `ConditionVariableWithOriginTracingWaiter`
+  ```cpp
+  class ConditionVariableWithOriginTracingWaiter {
```
elBoberido, line 157: I have no idea how this, `ConditionVariableWithOriginTracingSignaler` and `ConditionVariableWithOriginTracing` are interacting with each other and with the `ReactAL`. Could you briefly describe it?

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+      // returns a list of ids which have notified the conditionVariable
+      cxx::vector<uint64_t> wait(); 
+      // sets the atomic bool to false in m_triggeredBy[id]
+      void resetWaitState(const uint64_t id) noexcept; 
+  };
+  ```
+
+We maybe introduce some abstraction for this. One thought is to introduce a 
+`SignalVector` where you can acquire a `Notifyier` which then signals the 
+`ConditionVariable` and with the correct id.
```
elBoberido, line 168: Please don't use `we`, `you`, `maybe`, etc. For open points please add a section at the bottom.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+  void attachCallback(SubscriberEvent event, ConditionVariable & cv, ConditionVariable::index_t & index
+                      const OutOfScopeCallback_t & callback);
+
+  void detachCallback(SubscriberEvent event);
+  //...
+};
+```
+
+The `ReactAl` will call the methods above to attach or detach a condition 
+variable to a class.
```
elBoberido, line 215: I guess the methods are not public and the `ReactAL` must be a friend to be able to access it.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
+## Problem Description
+
+We require a public API so that a developer can register callbacks for certain
+events like the arrival of a sample.
+
+## Name of the design element
```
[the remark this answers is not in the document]
elfenpiff: I know ... I thought, start with why and the template was added by @mossmaurice - he hijacked my PR ;)

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+      conditions, Heisenbugs and in general hard to debug problems.
+ - **robust API feature** a feature we only support to increase the robustness 
+      of the API and bugs caused by misuse.
+ - (Reactor pattern)[https://en.wikipedia.org/wiki/Reactor_pattern]
+ - (Condition Variable)[https://en.wikipedia.org/wiki/Monitor_(synchronization)#Condition_variables_2]
+ 
+## Design
+
+### General
+The usage should be similar to the _WaitSet_ with a key difference - it should
```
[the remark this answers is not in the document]
elfenpiff: Added to the issue and the table will be available when the feature is implemented since we know then exactly what we are dealing with.

doc/design/eventdriven-callbacks.md
elfenpiff: Added todo to the issue " Create use case based comparison table between WaitSet and callback proposal" where this will be handled when the proposal is implemented since then we exactly know how the ReactAl will behave.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+
+ - concurrently: attaching a callback at any time from anywhere. This means in particular
+    - Attaching a callback from within a callback.
+    - [robust] Updating a callback from within the same callback 
+      ```cpp
+      void sampleReceived2(iox::popo::UntypedSubscriber & subscriber) {}
+
+      void sampleReceived(iox::popo::UntypedSubscriber & subscriber) {
+        myCallbackReactal.attachEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, sampleReceived2);
+      }
```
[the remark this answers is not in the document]
elfenpiff: I documented the following: One callback per event of an object. Settings a callback when a callback is already attached overrides the callback with the new one.

The rational here is:
 - Overriding callbacks is no problem for us and allowing it per default reduces the error handling on our and the user side
 - It is exactly what my intuition expects when I call an `attachEvent` with a new callback.
 - We could deny it but this could also be a source of bugs due to a type or a c&p error. An error message could clear things somehow up, if the bug is tested. But as a compromise, how about a tracing message every time the callback is updated? Then the user can take a look at the logs and knows what's going on!

I would not use the `AttachmentPolicy` idea since for me it is intuitively clear that `attachEvent` assigns the callback. 

But maybe we should rename the method into `assignCallback` then it is more clear that a callback is assigned and it is like assigning a number.

```cpp
int a = 5; 
// and now assign 8
a = 8; // and everyone knows that the old value 5 is gone
```

doc/design/eventdriven-callbacks.md
elfenpiff: Maybe?! But I would not go into so much detail in the first design draft since this will the implementation later show. But I created a task to update the design document when the implementation is finished so that when we get a base class it is documented here.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,215 @@
...
+
+ - When the class which is attached to the ReactAl goes out of scope it should 
+    detach itself from the ReactAl via a provided callback (like in the WaitSet).
+
+### Usage Code Example
+```cpp
+ReactAl myCallbackReactal;
+iox::popo::UntypedSubscriber mySubscriber;
+
+void sampleReceived(iox::popo::UntypedSubscriber & subscriber) {
```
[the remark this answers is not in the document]
elfenpiff: Me like it!

doc/design/eventdriven-callbacks.md
elfenpiff: Sounds better but I am still not perfectly happy.

doc/design/eventdriven-callbacks.md
elfenpiff: Every instance which can signal an event has its own id and signals the event directly. There will most likely never be some kind of class which collects events and notifies a set of ids.

doc/design/eventdriven-callbacks.md
elfenpiff: I documented it in the beginning of the Condition Variable Solution section (right above)

doc/design/eventdriven-callbacks.md
elfenpiff: I think this is the wrong place to handle the thread-realtime-issue. But I will try to keep this in mind and create a more flexible design.

doc/design/eventdriven-callbacks.md
elfenpiff: Correct. I added it in the code example.

doc/design/eventdriven-callbacks.md
elfenpiff: Will be done in the final task when this design document is updated after the implementation is finished. At the moment it should provide you with the idea on how it could look like.

doc/design/eventdriven-callbacks.md
elfenpiff: Reactor makes real sense since we use the Reactor Pattern behind the scenes.

elfenpiff: What about uncreative name CallSet to be somehow consistent to WaitSet?!

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+void onSampleReceived(iox::popo::UntypedSubscriber & subscriber) {
+  subscriber.take().and_then([&](auto & sample){
+    std::cout << "received " << std::hex << sample->payload() << std::endl;
+  });
+}
+
+void onWaitForSubscription(iox::popo::UntypedSubscriber & subscriber) {
+  std::cout << "subscribed!\n";
+  myCallbackReactal.attachEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
+  myCallbackReactal.attachEvent(subscriber, iox::popo::SubscriberEvent::UNSUBSCRIBED, onWaitForSubscription);
```
FerdinandSpitzschnueffler, line 99: The callback should be ``onWaitForUnsubscribe``.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+      - Create the class with a template bool array size argument to be more flexible and memory efficient.
+
+    **Note:**
+      - Sadly, we cannot pursue this approach in the WaitSet since it is event 
+        and state based at the same time and this approach supports only event 
+        driven triggering.
+
+  - add class `ConditionVariableWithOriginTracingSignaler` (I know, better name)
+  
+    **Reason:** it behaves quiete differently then the `ConditionVariable`. The `ConditionVariable` tells you:
```
FerdinandSpitzschnueffler, line 162: typo: quite

doc/design/eventdriven-callbacks.md
ithier: Yeah, why is it capitalized like this (with the small L) ? It looks a bit weird to me.
Also, I keep reading it as "Rectal"...

doc/design/eventdriven-callbacks.md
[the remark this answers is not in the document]
mossmaurice: I agree. Assign makes it clearer for me compared to attach.

doc/design/eventdriven-callbacks.md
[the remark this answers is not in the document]
mossmaurice: IMHO the structure doesn't need to be very strict. It's just there to give other contributors the chance to start with a minimal template.

doc/design/eventdriven-callbacks.md
elBoberido: @elfenpiff I have another one your definitely like `event listening for execution` -> `ELFE`

doc/design/eventdriven-callbacks.md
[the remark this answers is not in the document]
elBoberido: +1 for assign

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+      myCallbackReactal.attachEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
+      ```
+ - One can attach at most one callback to a specific event of an object.
+    - Attaching a callback to an event where a callback has been already attached overrides
+      the existing callback with the new one.
+ - concurrently: detach a callback at any time from anywhere. This means in particular
+    - Detaching a callback from within a callback.
+    - [robust] When the detach call returns we guarantee that the callback is never called
+      again. E.g. blocks till the callback is removed, if the callback is concurrently 
+      running it will block until the callback is finished and removed.
```
MatthiasKillat, line 73: Hence we will have a deadlock in library code when the callback provided by the user is never returning. Probably unavoidable if we want this, we cannot really stop the callback. The alternative is to reject adding the new callback when the old callback is running (we could still block in user code than, this is entirely unavoidable with callbacks without preemption by the OS).

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+      Exception: If a callback detaches itself it blocks until the callback is removed
```
MatthiasKillat, line 74: So the callback will/cannot continue after it is removed? (In some cases it could not properly continue I think, e.g. when depends on the data it relies on).

The semantics are not entirely clear here.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+      running it will block until the callback is finished and removed.
+      Exception: If a callback detaches itself it blocks until the callback is removed 
+                 and not until the callback is finished.
+    - Calling detach means that after that call the callback is no longer attached
+      even when it was not attached in the first place. Therefore it will always succeed.
+
+ - When the ReactAl goes out of scope it should detach itself from every class 
+     to which it was attached via a provided callback (like in the WaitSet).
+
+ - When the class which is attached to the ReactAl goes out of scope it should
```
MatthiasKillat, line 82: The callback is probably user provided. Can this be automated to make it less error-prone?

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+
+void onSampleReceived(iox::popo::UntypedSubscriber & subscriber) {
+  subscriber.take().and_then([&](auto & sample){
+    std::cout << "received " << std::hex << sample->payload() << std::endl;
+  });
+}
+
+void onWaitForSubscription(iox::popo::UntypedSubscriber & subscriber) {
+  std::cout << "subscribed!\n";
+  myCallbackReactal.attachEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
```
MatthiasKillat, line 98: I find this reliance on those event enums unnecessarily limiting. I am fairly sure this can be suitably generalized. In any case, those events need more explanation on a conceptual level I think.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+  myCallbackReactal.detachEvent(subscriber, iox::popo::SubscriberEvent::UNSUBSCRIBED);
+}
+
+int main() {
+  myCallbackReactal.attachEvent(mySubscriber, iox::popo::SubscriberEvent::SUBSCRIBED, onWaitForSubscription);
+}
+```
+
+### Solution
+#### Condition Variable
```
MatthiasKillat, line 116: I always think the building block for this should be a Semaphore (with the ability to exist in shared memory if needed). It is not actually a condition variable in my book.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+
+  - Alternative names: `EventVariable`, `EventSignaler`, `EventWaiter`
+  
+  - **Problem:** The ReactAl would like to know by whom it was triggered. The WaitSet has a 
+                  big list of callbacks where it can ask the object if it triggered the WaitSet. This has
```
MatthiasKillat, line 121: Big, small, large, huge ... does not matter here, remove. It can query the object by whom it was triggered and this might be inefficient, period.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+        and performance wise.
+        * implementation, since we have to manage set and reset all the callbacks 
+        * performance, when we call a callback with every iteration, we have to first load the 
+            callback into the cpu cache (generates cache misses for all the other callback 
+            elements in the vector which is being iterated). In the callback a method in an 
+            object is called (the object has to be loaded again into the cpu cache, cache miss for 
+            every member which is not used by the method).
+
+    **Potential optimization:**
+      - Create an additional bool member `m_wasTriggered` which is true when at least
```
MatthiasKillat, line 148: Should be a flag in the array (e.g. first or last one) itself for cache efficiency.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+
+    **Note:**
+      - Sadly, we cannot pursue this approach in the WaitSet since it is event 
+        and state based at the same time and this approach supports only event 
+        driven triggering.
+
+  - add class `ConditionVariableWithOriginTracingSignaler` (I know, better name)
+  
+    **Reason:** it behaves quiete differently then the `ConditionVariable`. The `ConditionVariable` tells you:
+    someone has signalled you and you have to find out who it was. The `ConditionVariableWithOriginTracing`
```
MatthiasKillat, line 163: I would write something like this from another perspective and more formally. "The CV notifies an object which needs to find the source of this signal or something like this."

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+};
+```
+
+The `ReactAl` will call the methods above to attach or detach a condition 
+variable to a class. The methods should be private and the `ReactAl` must then 
+be declared as a friend.
+
+## Open Points
+
+ - Maybe introduce some abstraction Condition Variable tracing handling.
```
MatthiasKillat, line 242: I would try to generalize the trigger mechanism to accept very general objects which just need to provide a wait/notify mechanism. (As a template parameter Notifyable, most likely). There maybe the necessity to also rely on a factory method ``Notifyable::create(Args...&& args)`` or something similar for construction in specific cases.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+    void detachEvent( EventOrigin & origin, const EventType & eventType );
+};
+
+```
+ 
+  - Contains a thread which will wake up and iterate through ConditionVariable
+    bool array. If an entry is true, set it to false and then call the 
+    corresponding callback.
+
+#### Class which is attachable to ReactAl
```
MatthiasKillat, line 215: The restrictions on the user defined type using this pattern need to be clarified. What is the purpose of those entities and what must be done/provided by the client class internally and what can the user of the class do (this is shown somewhat in the usage above).

Still, from this alone I would have a hard time to write such a user defined class.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+
+
+## Terminology
+
+ - **event driven** a one time reaction which is caused directly by an event.
+      Example: a new sample has been delivered to a subscriber.
+ - **state driven** a repeating reaction which is caused by a specific structural 
+      state until the state changes.
+      Example: a subscriber has stored samples which were not inspected by the user.
+ - **robust API** we use this term to describe an API which is hard
```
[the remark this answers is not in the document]
elfenpiff: The idea behind this definition was to state that certain features are not introduced because they are nice but because they can make the daily developer live hard since they had to pay attention to a lot of restrictions. Sometimes they are maybe even not aware that the code is running inside of such a Reactor.

@MatthiasKillat 
Robust is the word which fits in this particular case best. But I am open for other suggestions! I know it is vaguely defined but again, I am again open for suggestions!

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+      of the API and bugs caused by misuse.
+ - (Reactor pattern)[https://en.wikipedia.org/wiki/Reactor_pattern]
+ - (Condition Variable)[https://en.wikipedia.org/wiki/Monitor_(synchronization)#Condition_variables_2]
+ 
+## Design
+
+### General
+The usage should be similar to the _WaitSet_ with a key difference - it should 
+be **event driven** and not a mixture of event and state driven, depending on
+which event is attached, like in the _WaitSet_.
```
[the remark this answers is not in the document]
elfenpiff: I extended the first sentence in the document (Problem Description) for a better description.

At the moment we cannot link a WaitSet document since we do not have one and I would suggest to wait with the design documentation till this part is finalized since the WaitSet has to maybe adapt as well.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,247 @@
...
+    should be called **once**.
+ - If an event occurs while the callback is being executed the callback should be 
+    called again **once**.
+
+The following use cases and behaviors should be implemented.
+
+ - The API and ReactAl should be robust this means that it should be very hard 
+    to use the API in the wrong way. Since ReactAl is working concurrently a 
+    wrong usage could lead to race conditions, extremely hard to debug bug reports 
+    (HeisenBugs) and can be frustrating to the user.
```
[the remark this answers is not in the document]
elfenpiff: I would like to keep this unnecessary wording to make the problems visible for the less experienced.

doc/design/eventdriven-callbacks.md
elfenpiff: I would say yes otherwise the API can get very error prone since the developer is not aware that the code may be running inside a Reactor.

doc/design/eventdriven-callbacks.md
elfenpiff: The callback is not user provided and it will be provided by the attached class.

doc/design/eventdriven-callbacks.md
elfenpiff: Restricting it to enums is making the implementation a little bit more easy. I think  YAGNI until we need a more general solution but at the moment there is not a single use case for it.

doc/design/eventdriven-callbacks.md
elfenpiff: Here I would say again YAGNI until the use case arises. To create a much more general approach would be nice but it also has some overhead since we have to create the ReactAL then with multiple template parameters, have to provide factory methods etc..

doc/design/eventdriven-callbacks.md
elfenpiff: I will add this in detail when the implementation is done. At the moment this document should provide you with a first glimpse on where the journey is going.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+      ```cpp
+      void onSampleReceived2(iox::popo::UntypedSubscriber & subscriber) {}
+
+      void onSampleReceived(iox::popo::UntypedSubscriber & subscriber) {
+        myCallbackReactAL.assignEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived2);
+      }
+
+      myCallbackReactAL.assignEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
+      ```
+ - One can assign at most one callback to a specific event of an object. The event is usually defined
```
elBoberido, line 70: So it's possible to do this?
```
myCallbackReactAL.assignEvent(subscriber1, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
myCallbackReactAL.assignEvent(subscriber2, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
```

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+      the existing callback with the new one.
+ - concurrently: detach a callback at any time from anywhere. This means in particular
+    - Detaching a callback from within a callback.
+    - [robust] When the detach call returns we guarantee that the callback is never called
+      again. E.g. blocks till the callback is removed, if the callback is concurrently 
+      running it will block until the callback is finished and removed.
+      Exception: If a callback detaches itself it blocks until the callback is removed 
+                 and not until the callback is finished. After the successful removal the 
+                 callback will continue and will not be called again.
+    - Calling detach means that after that call the callback is no longer assigned
```
elBoberido, line 82: This could be phrased a bit simpler
```
The callback is no longer assigned to the event after calling `detach`. The method will succeed even if it was not assigned in the first place.
```

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+      even when it was not assigned in the first place. Therefore it will always succeed.
+
+ - When the ReactAL goes out of scope it should detach itself from every class
```
elBoberido, line 85: Should detach or will detach? It's not the same

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+ReactAL myCallbackReactAL;
+iox::popo::UntypedSubscriber mySubscriber;
+
+void onSampleReceived(iox::popo::UntypedSubscriber & subscriber) {
+  subscriber.take().and_then([&](auto & sample){
+    std::cout << "received " << std::hex << sample->payload() << std::endl;
+  });
+}
+
+void onWaitForSubscription(iox::popo::UntypedSubscriber & subscriber) {
```
elBoberido, line 102: I think the name is wrong. This isn't executed while waiting for subscription, but on subscription.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+  std::cout << "subscribed!\n";
+  myCallbackReactAL.assignEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED, onSampleReceived);
+  myCallbackReactAL.assignEvent(subscriber, iox::popo::SubscriberEvent::UNSUBSCRIBED, onWaitForUnsubscription);
+  myCallbackReactAL.cancelEvent(subscriber, iox::popo::SubscriberEvent::SUBSCRIBED);
+}
+
+void onWaitForUnsubscribe(iox::popo::UntypedSubscriber & subscriber) {
```
elBoberido, line 109: same here

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+  std::cout << "unsubscribed from publisher\n";
+  myCallbackReactAL.assignEvent(subscriber, iox::popo::SubscriberEvent::SUBSCRIBED, onWaitForSubscription);
+  myCallbackReactAL.cancelEvent(subscriber, iox::popo::SubscriberEvent::HAS_SAMPLE_RECEIVED);
```
elBoberido, line 112: Not sure about `cancelEvent`. I would not assume that it's also detached. Why not just keeping `detach`? `retract` or `retrieve` might also work better that `cancel`.

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,271 @@
...
+#### Class Interactions
+
+  - **Creating ReactAL:** an `EventVariable` is created in the shared memory. 
+      The `ReactAL` uses the `EventWaiter` to wait for incoming events.
+  - **Attaching Subscriber Event (sampleReceived) to ReactAL:** The subscriber attaches the `EventSignaler`
+      appropriately and signals the `ReactAL` via the `EventSignaler` about the occurrence 
+      of the event (sampleReceived).
+  - **Signal an event from subscriber:** `EventSignaler.notify()` is called which sets an, to the subscriber 
+      assigned flag, to true and `EventWaiter` will wake up.
+  - **ReactAL reacting to events:** `EventWaiter.wait()` will provide a complete list of all `EventSignaler`
```
elBoberido, line 131: `EventListener.wait()`?

elBoberido (commented):
Could you clean this up a bit?
It's till not fully clear to me how all this classes interact with each other. One reason is that different names are used for the same classes. Please keep the same name throughout the document and suggest different names only at a single place.
Could you maybe also add a small ascii diagram for the interactions of the classes?

mossmaurice:
> Could you maybe also add a small ascii diagram for the interactions of the classes?

I think having a bit of ASCII or a PlantUML is a great idea.

doc/design/eventdriven-callbacks.md
elfenpiff: Sounds better!

doc/design/eventdriven-callbacks.md
elfenpiff: I thought assign & cancel and attach & detach. Detach did sound like the opposite of assign to me. But when it is clearer for the rest of the world I will keep detach.

elBoberido: maybe it's just me, but although it feels natural to say `cancel an assignment` it feels unnatural to me to use `cancel an event` with the meaning of not being assigned anymore.

elfenpiff:
@mossmaurice 

> If it is similar to the WaitSet why not re-using more parts of it?
> AFAICT you're only planning to use and extend the ConditionVariableSignaler

I will try to reuse as much as possible. But this design document is just a first draft. To design this in full detail will cost a lot of time and it will still not fit the realized implementation later.

> Could the ice_waitset_trigger.cpp example be implemented with ReactAl?
> My current understanding is that this example could be implemented with both. Hence, I have the feeling that we're violating KISS (keep it simple & stupid) here. The users shouldn't have more than one solution per use-case

 As you suggested I am reusing as much as possible and this leads to the nice side effect that the WaitSet and the ReactAL can be used in the same manner. In the end I think we are fully following KISS - two easily usable APIs which can be handled similarly

mossmaurice (approved):
Nice design document!

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,337 @@
...
+
+The `ReactAL` will call the methods above to enable or disable a specific event by 
+providing a condition variable. The methods should be private and the `ReactAL` must then 
+be declared as a friend.
+
+## Open Points
+
+ - Maybe introduce some abstraction Condition Variable tracing handling. 
+    One thought is to introduce a `SignalVector` where you can acquire a 
+    `Notifyier` which then signals the `ConditionVariable` and with the
```
elBoberido, line 334: typo `Notifier`

doc/design/eventdriven-callbacks.md
```diff
@@ -0,0 +1,337 @@
...
+  - **ReactAL reacting to events:** `EventListener.wait()` will provide a complete list of all `EventNotifier`
+      which were notifying the `EventListener` till the last `wait()` call. The ReactAL uses this list 
+      to call the corresponding callbacks.
+```cpp
+  auto activeCallbacks = m_eventListener.wait();
+  for(uint64_t i = 0U; i < activeCallbacks.size(); ++i)
+    if ( activeCallbacks[i] ) m_callbacks[i]();
+```
+
+#### Event Variable
```
elBoberido, line 204: This section is a bit cluttered. Could you try to reorganize it a bit?

elBoberido (commented):
The proposal looks sound. The document would benefit from some additional work, but from my point of view you could already start with the implementation.

elBoberido (approved):
I'm a bit hesitant if drafts should be in a parallel folder to `design` or just a `draft` subfolder

mossmaurice (dismissed):

elBoberido (approved):

mossmaurice (approved):

Issue #518, opened 2021-01-22
Refactoring of introspection


FerdinandSpitzschnueffler:
## Brief feature description

The introspection needs a refactoring regarding:
- how is data internally stored and accessible for introspection
- how is this data sent to the client
- which information should be presented and how
- introspection overhead should be minimal when the introspection is not used

## Detailed information

When the introspection was implemented multiple pub/sub scenarios have not been considered, e.g. when a subscriber is subscribed to multiple publishers only one connection is displayed. When one publisher stops sending data then the subscriber is shown as disconnected (while still receiving data from the other publisher). This has to be adapted.
Additionally, introspection tests are missing and the existing ones may have to be adapted once it is clear how data is internally stored and sent to the client.

Open questions:

- Can we replace std::map?
- Do we want to have a mechanism to turn off the data transfer to the introspection? Do we still need it after the refactoring?
- Do we want to stick to ncurses?
- Also consider #414 and #31

budrus:
>  When one publisher stops sending data then the subscriber is shown as disconnected (while still receiving data from the other publisher).

@FerdinandSpitzschnueffler . Is this because the intropspection processes the CaPro messages again to determine the connection state? And this "state machine" no does not fit to the mult-puhser one? I think we would have to change it in a way that state changes of subscribers are reported and not a redundant (and maybe other) processing of all CaPro messages is made

dkroenke:
> Do we want to stick to ncurses?

From my point of view we can stick to ncurses for the terminal output, but we need a solution for writing the introspection data into a parseable logfile for debugging purpose.

FerdinandSpitzschnueffler:
> > When one publisher stops sending data then the subscriber is shown as disconnected (while still receiving data from the other publisher).
> 
> @FerdinandSpitzschnueffler . Is this because the intropspection processes the CaPro messages again to determine the connection state? And this "state machine" no does not fit to the mult-puhser one? I think we would have to change it in a way that state changes of subscribers are reported and not a redundant (and maybe other) processing of all CaPro messages is made

@budrus Yes, when one publisher stops sending data then the connection state is set to disconnected without checking whether the subscriber is also subscribed to another publisher.

MatthiasKillat:
[...]

PR #519, opened 2021-01-22
Iox #349 add nodename to pub sub options

### LINKED ISSUE #349: Introduction of Runnables/Nodes for the new APIs
## Brief feature description

## Detailed information

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
in scope:
- node names as additional information in pub/sub options
- nodes visible in introspection
- add missing NodeData tests
- default nodeName = runtime name
- small refactoring of introspection so that connection to publisher is shown correctly (not working for multiple publishers, see #518)

out of scope:
- C API
- full refactoring of introspection (#518)
- additional introspection tests, because the introspection needs to be refactored

open questions for follow-up PR:
- Should pubs/subs be grouped by their node? This means:
Can we have multiple publishers with the same ServiceDescription and the same nodeName? In this PR they are distinguished by their unique port id. Do subscribers have to have the same nodeName as the publisher to be connected?
- Are the names sufficient for now or do we want additional information (like a device id)? Do we want to replace the nodeName by a NodeData pointer?

1. [ ] Merge

## References

- partly closes #349


iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -61,14 +61,11 @@ PortManager::PortManager(RouDiMemoryInterface* roudiMemoryInterface) noexcept
 
     popo::PublisherOptions options;
     options.historyCapacity = 1;
+    options.nodeName = INTROSPECTION_SERVICE_ID;
```
budrus, line 64: This is something like a built-in node the introspection publisher port belongs to? `INTROSPECTION_SERVICE_ID` is maybe a bad name. Would this be `INTROSPECTION_NODE`?

budrus:
With respect to your open questions @FerdinandSpitzschnueffler 

>     Can we have multiple publishers with the same ServiceDescription and the same nodeName? In this PR they are distinguished by their unique port id. 

For me it wouldn't make sense that there multiple publishers with the same `ServiceDescription` in one node. But if we do not have a problem with this we maybe shouldn't do a limitation. Can the introspection client handle this? I.e. Do I then see two identical entries? 

>     Do subscribers have to have the same nodeName as the publisher to be connected?
The normal case is that publishers and subscribers don't have the same node name. And this is perfectly fine as we do communication between nodes. Or did I get the question wrong? 

>     Are the names sufficient for now or do we want additional information (like a device id)? Do we want to replace the nodeName by a NodeData pointer?
For now it is sufficient. If we remove the `NodeData` (see other comment) the other question is answered too. Currently we do not have to create node objects for being able to set the node name for publishers and subscribers

budrus:
@FerdinandSpitzschnueffler In #518 you mention 

> When one publisher stops sending data then the subscriber is shown as disconnected (while still receiving data from the other publisher)

This means that we currently see wrong connection states in the introspection app? With single- or multi-publisher options we now have two different state machines in the subscribers (also two different classes).  The state machine in the introspection behaves like the old single-publisher one I guess. But the default setting in multi-publisher. As I already mentioned in #518 it would be better if the introspection reads the connection state and does not determine it again. But this should maybe be addresses in #518 

Having wrong information displayed in the introspection app is not nice either. I think we could fix it when evaluating the multi-pusher switch in the `getNextState` method

iceoryx_posh/source/roudi/roudi_process.cpp
```diff
@@ -581,18 +581,30 @@ void ProcessManager::addNodeForProcess(const ProcessName_t& processName, const N
     RouDiProcess* process = getProcessFromList(processName);
     if (nullptr != process)
     {
-        runtime::NodeData* node = m_portManager.acquireNodeData(processName, nodeName);
+        m_portManager.acquireNodeData(processName, nodeName)
```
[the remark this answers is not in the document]
budrus: For now I would keep it. I have no other idea how to setup a multi-process database with available nodes for rmw_iceoryx. We can discuss this once we touch rmw_iceoryx again

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -446,29 +439,105 @@ TEST_F(PortManager_test, ConditionVariablesDestroy)
...
+            });
+
+        auto nodeData = m_portManager->acquireNodeData("AnotherProcess", "AnotherNode");
+        EXPECT_THAT(nodeData.has_error(), Eq(true));
+        EXPECT_THAT(errorHandlerCalled, Eq(true));
+        EXPECT_THAT(nodeData.get_error(), Eq(PortPoolError::NODE_DATA_LIST_FULL));
+    }
+
+    // delete one and add one NodeData should be possible now
+    {
```
MatthiasKillat, line 471: In the light of recent discussion about test case granularity one could argue that this is another test case (i.e. was full, is not full anymore and now creation is possible again).

I am not necessarily of this opinion here, since it would mainly increase code duplication and test execution time (due to requiring another setup phase).

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -446,29 +439,105 @@ TEST_F(PortManager_test, ConditionVariablesDestroy)
...
+
+        auto nodeData = m_portManager->acquireNodeData(newProcessName, newNodeName);
+        EXPECT_THAT(nodeData.has_error(), Eq(false));
+        EXPECT_THAT(nodeData.value()->m_node, StrEq(newNodeName));
+    }
+}
+
+TEST_F(PortManager_test, NodeDataDestroy)
+{
+    iox::ProcessName_t process = "Humuhumunukunukuapua'a";
```
MatthiasKillat, line 485: Ever heard of uMgungundlovu?

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -446,29 +439,105 @@ TEST_F(PortManager_test, ConditionVariablesDestroy)
...
+    iox::NodeName_t node = "Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu";
```
MatthiasKillat, line 486: Introducing such a long name without necessity distracts from the purpose of the test....

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -446,29 +439,105 @@ TEST_F(PortManager_test, ConditionVariablesDestroy)
...
+        iox::NodeName_t newNodeName(iox::cxx::TruncateToCapacity, node + std::to_string(testi));
+        m_portManager->deletePortsOfProcess(newProcessName);
+
+        auto nodeData = m_portManager->acquireNodeData(newProcessName, newNodeName);
+        EXPECT_THAT(nodeData.has_error(), Eq(false));
+        EXPECT_THAT(nodeData.value()->m_node, StrEq(newNodeName));
+    }
+}
+
+TEST_F(PortManager_test, NodeDataDestroy)
```
MatthiasKillat, line 483: The purpose of this test seems to be to show that destroyed node data enables us to recycle the memory to create new node data. The test name could be improved to reflect this.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -446,29 +439,105 @@ TEST_F(PortManager_test, ConditionVariablesDestroy)
...
+        EXPECT_THAT(nodeData.has_error(), Eq(true));
+    }
+
+    // set the destroy flag and let the discovery loop take care
+    for (unsigned int i = 0U; i < iox::MAX_NODE_NUMBER; i++)
+    {
+        nodeContainer[i]->m_toBeDestroyed.store(true, std::memory_order_relaxed);
+    }
+    m_portManager->doDiscovery();
+    nodeContainer.clear(); // These pointers are dangling now
```
MatthiasKillat, line 509: Remove the comment or expand on this (dangling is only relevant if we try to access them, which we do not). We can always safely forget pointers no one intends to read from if someone else does the cleanup of the data they point to (doDiscovery in this case).

The comment is not helpful to explain this by itself.

tools/introspection/source/introspection_app.cpp
```diff
@@ -579,10 +590,11 @@ std::vector<ComposedSubscriberPortData> IntrospectionApp::composeSubscriberPortD
...
         {
-            subscriberPortData.push_back(
-                {port,
-                 (port.m_publisherIndex != -1) ? &portData->m_publisherList[static_cast<uint64_t>(port.m_publisherIndex)] : nullptr,
-                 subscriberPortChangingData->subscriberPortChangingDataList[i++]});
+            subscriberPortData.push_back({port,
+                                          (port.m_publisherIndex != -1)
+                                              ? &portData->m_publisherList[static_cast<uint64_t>(port.m_publisherIndex)]
+                                              : nullptr,
+                                          subscriberPortChangingData->subscriberPortChangingDataList[i++]});
```
MatthiasKillat, line 597: General question wrt. introspection changes: We are not sorting by node name in the introspection right now (for display purposes)? Was this not a (low prio) requirement?

As I see it, we just display the node name in addition now.

MatthiasKillat (changes requested):
Looks good in general. I consider some minor changes (names, std::move) to be required. Some optional suggestions and open questions remain as well.

MatthiasKillat:
@budrus @FerdinandSpitzschnueffler The introspection was not designed to be able to deal with multiple publishers. This is not straightforward to fix and out of scope here (we could create a separate issue).

The problem is that we currently only have the option to track whether a subscriber is connected to a **single** publisher (and by the current design this is the last publisher it connects to). As soon as this last publisher stops offering, the subscriber is considered disconnected, even if the other publisher is still sending.

There is probably a quickfix involving a counter on the side of the info for each subscriber, number of active connections.
It goes up by one whenever it connects and down by one if a publisher stops offering. 

The subscriber is disconnected whenever the counter is 0. Now, this does not magically create the option to show track all corresponding publishers by name (needs thoughtful rework of the data structures), therefore I do not believe it should be done here.

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, 2020 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
+// Copyright (c) 2019, 2021 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: As I understand it, the year of the creation and the year of the last modification are needed. But I'm not sure and hope we can get rid of the year in the copyright header. Let's see what #483 brings.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -446,29 +439,105 @@ TEST_F(PortManager_test, ConditionVariablesDestroy)
     }
 }
 
+TEST_F(PortManager_test, NodeDataOverflow)
+{
+    std::string process = "Process";
+    std::string node = "Node";
+
+    // first acquire all possible NodeData
+    for (unsigned int i = 0U; i < iox::MAX_NODE_NUMBER; i++)
```
elfenpiff, line 448: I would split up this test into multiple ones.
- AcquiringMaximumNumberOfNodesIsWorking
- AcquiringOneMoreThenMaximumNumberOfNodesFails
- DeletingNodeWorks (fill up to maximum, delete one and add one)

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
elfenpiff: @MatthiasKillat don't be a party killer ;)

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
```diff
@@ -204,7 +204,7 @@ TEST_F(ProcessIntrospection_test, addRemoveNode)
         const char NODE_2[] = "the_octagon";
```
elfenpiff, line 204: Please adjust year in the header.

elfenpiff (changes requested):
Please handle nullptr ctor arguments.

iceoryx_posh/source/roudi/roudi_process.cpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: We do not use ``NodeData`` at the moment but it could be useful if we want to store additional information (like a device id which could be necessary if we will have GPU support).

iceoryx_posh/test/moduletests/test_roudi_process_introspection.cpp
FerdinandSpitzschnueffler: I think this is not necessary since I just replaced the word ``runnable`` by ``node`` in some comments.

FerdinandSpitzschnueffler:
> With respect to your open questions @FerdinandSpitzschnueffler
> 
> > ```
> > Can we have multiple publishers with the same ServiceDescription and the same nodeName? In this PR they are distinguished by their unique port id. 
> > ```
> 
> For me it wouldn't make sense that there multiple publishers with the same `ServiceDescription` in one node. But if we do not have a problem with this we maybe shouldn't do a limitation. Can the introspection client handle this? I.e. Do I then see two identical entries?
> 

The introspection client can handle this - we see two identical entries when ServiceDescription and nodeName are equal for both publishers. The distinction by unique port ids was an easy first step, but I agree with you, this probably doesn't make that much sense. But changing this also means that in a multi-publisher scenario we force the user to add the node name (e.g. ice_multi_publisher). 
For avoiding publishers with the same ServiceDescription within the same node we have to check the node name when creating the publisher. I would try this in a follow-up PR. 
@budrus @MatthiasKillat @elfenpiff what do you think?

FerdinandSpitzschnueffler:
> > ```
> > Do subscribers have to have the same nodeName as the publisher to be connected?
> > ```
> 
> The normal case is that publishers and subscribers don't have the same node name. And this is perfectly fine as we do communication between nodes. Or did I get the question wrong?
> 

Agreed.

FerdinandSpitzschnueffler:
> @FerdinandSpitzschnueffler In #518 you mention
> 
> > When one publisher stops sending data then the subscriber is shown as disconnected (while still receiving data from the other publisher)
> 
> This means that we currently see wrong connection states in the introspection app? With single- or multi-publisher options we now have two different state machines in the subscribers (also two different classes). The state machine in the introspection behaves like the old single-publisher one I guess. But the default setting in multi-publisher. As I already mentioned in #518 it would be better if the introspection reads the connection state and does not determine it again. But this should maybe be addresses in #518
> 
> Having wrong information displayed in the introspection app is not nice either. I think we could fix it when evaluating the multi-pusher switch in the `getNextState` method

Fix in follow-up PR, nice refactoring in #518.

FerdinandSpitzschnueffler:
> @budrus @FerdinandSpitzschnueffler The introspection was not designed to be able to deal with multiple publishers. This is not straightforward to fix and out of scope here (we could create a separate issue).
> 
> The problem is that we currently only have the option to track whether a subscriber is connected to a **single** publisher (and by the current design this is the last publisher it connects to). As soon as this last publisher stops offering, the subscriber is considered disconnected, even if the other publisher is still sending.
> 
> There is probably a quickfix involving a counter on the side of the info for each subscriber, number of active connections.
> It goes up by one whenever it connects and down by one if a publisher stops offering.
> 
Will try fix quickfix in follow-up PR.
> The subscriber is disconnected whenever the counter is 0. Now, this does not magically create the option to show track all corresponding publishers by name (needs thoughtful rework of the data structures), therefore I do not believe it should be done here.

Could be part of #518.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
MatthiasKillat: You can call it whatever you want in a private repo but considering the fact that there will be a point where this should be certified I would advise against it. 

In any case, I will not change those names myself should the request arise in the future...

MatthiasKillat (approved):

elfenpiff (approved):

PR #527, opened 2021-01-29
Iox #454 enhancing coverage for class roudi and port manager in roudi folder

### LINKED ISSUE #454: Enhancing Tests in Roudi
## Brief feature description
Introduce Module tests for RouDi component.

## Detailed information
Increase the code coverage.

Classes in RouDi components which needs test cases to be added.
In file roudi.cpp, class 
- [ ] RouDi.

In file roudi_process.cpp class 
- [x] RouDiProcess
- [x] PortManager
- [ ] ProcessManager
- [ ] RoudiMemoryManager

In file roudi_lock.cpp class 
- [ ] RouDiLock

chiranjeevimaddi:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
Code contributed by @JimmyBch @manel94

1. [x] Merge

## References

- Partly Closes #454


iceoryx_posh/test/moduletests/test_roudi.cpp
```diff
@@ -0,0 +1,545 @@
+// Copyright (c) 2021 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
```
JimmyBch, line 1: Concerning this file, many EXPECT are missing. I am aware but I didn't have any idea how to really check the different tests. On my side, I think it is missing an errorHandler when number of elements is incorrect.

dkroenke: To which number of elements are you referring to?

iceoryx_posh/test/moduletests/test_roudi.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+using namespace iox::roudi;
+using namespace iox::runtime;
+
+namespace iox
+{
+namespace test
+{
+/// @req IOX_SWRS_112, IOX_SWRS_200
+/// @brief Test goal: "This test suite verifies that the BaseClass function is verified"
```
dkroenke, line 37: Very good! Please fill out the template.

iceoryx_posh/test/moduletests/test_roudi.cpp
```diff
@@ -0,0 +1,545 @@
...
+            << static_cast<cxx::Serialization>(iox::runtime::PortConfigInfo(11U, 22U, 33U)).toString();
+    auto cmd = runtime::stringToMqMessageType(message.getElementAtIndex(0).c_str());
+    std::string processName = message.getElementAtIndex(1);
+
+    m_roudiApp->processMessageTest(message, cmd, ProcessName_t(cxx::TruncateToCapacity, processName));
+
+    m_roudiApp->shutDownTest();
+}
+
+/// @note name of the Testcase shall describe the test case in detail to avoid additional comments
```
dkroenke, line 226: can be removed when the name is meaningful

dkroenke (commented):
Just a few comments, detailed review will follow....

iceoryx_posh/test/moduletests/test_roudi.cpp
JimmyBch: Sorry, I was not so explicit. In the procesManager method of class Roudi (roudi.cpp), if the message does not contain the correct number of elements depending on the MqMessageType received, an error is set. In my opinion, errohandler could be used there.

elfenpiff (changes requested):
Most of the tests are missing the verification of the behavior. Please add it.

JimmyBch:
> Most of the tests are missing the verification of the behavior. Please add it.

Yes, I let a comment about it above :)

iceoryx_posh/test/moduletests/test_roudi.cpp
JimmyBch: ErrorHandler is not used right now. But, I guess it could be a potential improvement of the existing code, correct ?

dkroenke: I guess you refer to the method `processMessage` right?
Yes the errorhandler should be used there, but only with level `Moderate` because RouDi can proceed and the request from the runtime is not processed.
There is even in one case the errorhandler used: https://github.com/eclipse-iceoryx/iceoryx/blob/f7fee3c237bfe8ae15da434aa78937dba0b0c8ec/iceoryx_posh/source/roudi/roudi.cpp#L230

@elBoberido does it make sense to have an error like `Error::kROUDI__WRONG_NUMBER_PARAMETERS_MQMESSAGE`? Or do you want to refactor that too?

elBoberido: I think that's not necessary on the roudi side. It just hast to be communicated back to the application since this is input from an application and therefore the error handler should also be called on the application side.

JimmyBch: Does it mean the method process manager should return something ?

elBoberido: This is something that will heavily be refactored with #332 in the future to give the application more information what's going on. As a side effect, this makes testing also easier.
Currently there is no response to the application if the number of parameter don't match. With #332 that will change and an error response will be sent back.
What's even worse, calling the error handler with `Error::kPORT_MANAGER__INTROSPECTION_MEMORY_MANAGER_UNAVAILABLE` is definitely wrong and should be removed. The application just provided us with the wrong number of parameter.
Currently the `RouDi` class is not really good testable.

iceoryx_posh/test/moduletests/test_roudi.cpp
```diff
@@ -0,0 +1,545 @@
...
+}
+
+TEST_F(RouDiBaseClass_test, ShutDownKillProcessInDestructorTrueIsSuccessfull)
+{
+    std::unique_ptr<IceOryxRouDiComponents> m_roudiComponents =
+        std::unique_ptr<IceOryxRouDiComponents>(new IceOryxRouDiComponents(iox::RouDiConfig_t().setDefaults()));
+    std::unique_ptr<RoudiClassTest> m_roudiApp = std::unique_ptr<RoudiClassTest>(
+        new RoudiClassTest(m_roudiComponents->m_rouDiMemoryManager,
+                           m_roudiComponents->m_portManager,
+                           RouDi::RoudiStartupParameters{roudi::MonitoringMode::ON, true}));
```
elBoberido, line 117: this is repeated for almost every test and should be moved to `SetUp` method. The two tests with the ctor could be tests withoud a fixture, so just `TEST`, then it's easily possible

iceoryx_posh/test/moduletests/test_roudi.cpp
```diff
@@ -0,0 +1,545 @@
...
+    const iox::NodeName_t m_nodeName{"testNode"};
+    message << mqMessageTypeToString(MqMessageType::REG) << "App" << std::to_string(getpid())
+            << std::to_string(posix::PosixUser::getUserOfCurrentProcess().getID()) << std::to_string(0)
+            << static_cast<cxx::Serialization>(version::VersionInfo::getCurrentVersion()).toString();
+
+    const auto successfullySent = m_runtime->sendRequestToRouDi(message, m_receiveBuffer);
+
+    EXPECT_TRUE(m_receiveBuffer.isValid());
+    EXPECT_TRUE(successfullySent);
+}
```
elBoberido, line 163: this is a `PoshRuntime` test

iceoryx_posh/test/moduletests/test_roudi.cpp
```diff
@@ -0,0 +1,545 @@
...
+    {
+        this->shutdown();
+    }
+
+    void processMessageTest(const runtime::MqMessage& message,
+                            const iox::runtime::MqMessageType& cmd,
+                            const ProcessName_t& processName)
+    {
+        this->processMessage(message, cmd, processName);
+    }
```
elBoberido, line 67: you can just use `using RouDi::startMqThread;`, etc

elBoberido (commented):
All the tests without an `EXPECT` give a false feeling of safety because the line coverage is high, but except not crashes, nothing is verified. The `RouDi` class is currently not easily testable and needs a refactoring, which will happen in the near future.

iceoryx_posh/test/moduletests/test_roudi.cpp
JimmyBch: That was my only way to be able to get an assertion in this test.

elBoberido: But that still makes it a runtime test. You basically combined the `SendRequestToRouDiValidMessage` and `ValidAppName` from `PoshRuntime_test`.
Like I already said, this class is currently not easy testable and needs refactoring before useful tests can be written. On the positive side, much of this is already tested by other means, e.g. the posh runtime tests which uses the RouDiEnvironment which uses this class.

JimmyBch: Sure, but what I did could serve as a base for when the class will be refactored. By using poshruntime, I am able to improve the coverage and add Assert. Sure, it will be change after refactoring but currently, it is the only thing I can do.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -562,3 +571,133 @@ TEST_F(PortManager_test, PortDestroy)
         }
     }
 }
+
+TEST_F(PortManager_test, CheckNodeOverflowErrorReturnsnullptrForAcquireNodeData)
```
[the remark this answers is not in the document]
manel94: @elBoberido   all tests related to node added by me would be removed since i found this related PR update the tests related to node :
https://github.com/eclipse-iceoryx/iceoryx/issues/349

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -562,3 +571,133 @@ TEST_F(PortManager_test, PortDestroy)
...
+
+    // first aquire all possible condition variables
+    for (uint32_t i = 0U; i < iox::MAX_NODE_NUMBER; i++)
+    {
+        auto newnodeName = nodeName + std::to_string(i);
+        auto nodeDataResult =
+            m_portManager->acquireNodeData(m_Process, iox::NodeName_t(iox::cxx::TruncateToCapacity, newnodeName));
+        EXPECT_NE(nodeDataResult, nullptr);
+        nodeContainer.push_back(nodeDataResult);
+    }
```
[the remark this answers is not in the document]
manel94: #349 already done on master

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -562,3 +571,133 @@ TEST_F(PortManager_test, PortDestroy)
...
+    // so now no one should be available
+    {
+        auto interp = m_portManager->acquireInterfacePortData(iox::capro::Interfaces::INTERNAL, m_Process);
+        EXPECT_EQ(interp, nullptr);
+    }
+
+    // set the destroy flag and let the discovery loop take care
+    for (uint32_t i = 0U; i < iox::MAX_INTERFACE_NUMBER; i++)
+    {
+        interfaceContainer[i]->m_toBeDestroyed.store(true, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
manel94: i don´t think so if  i < iox::MAX_INTERFACE_NUMBER no ? or i didn´t understand well

iceoryx_posh/test/moduletests/test_roudi.cpp
elBoberido: But this doesn't increase the coverage in a meaningful way. In this specific case there is nothing tested which isn't already tested in `PoshRuntime_test`. Other tests which don't have any assert give a false sense of safety because they give the illusion of increased coverage while they cannot fail, except when they crash the application but this is already ensured by other means.
This also makes refactoring harder because not only the code needs to be refactored but also the tests, which would be okay if the tests did something meaningful. The chances are good that they will be completely rewritten.

Please don't write tests where nothing can be asserted just to hunt a coverage number. In general, if nothing can be asserted either the class needs to be adjusted or no test should be written.

JimmyBch: I know that the tests without assert are kind of useless right now. But, I thought of using them after refactoring as the logic, from my side, should be the same.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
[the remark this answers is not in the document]
elBoberido: @manel94 yes, if the implementation you are testing is correct, but before this test verifies that, the implementation could return just one valid pointer and only nullptr for the remaining calls. In this scenario, the previous `EXPECT_NE` will fail but the test keeps running until the nullptr access will crash the test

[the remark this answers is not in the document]
manel94: okay, in understand now thank you :D

iceoryx_posh/test/moduletests/test_roudi.cpp
elBoberido: But that is not necessarily the case and it makes things harder to refactor without having any benefit right now.

JimmyBch: I see, i will just checkout the change so.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -562,3 +571,133 @@ TEST_F(PortManager_test, PortDestroy)
...
+    // so now no one should be available
+    {
+        auto nodeDataResult = m_portManager->acquireNodeData(m_Process, m_Node);
+        EXPECT_EQ(nodeDataResult, nullptr);
+    }
+
+    // set the destroy flag and let the discovery loop take care
+    for (uint32_t i = 0U; i < iox::MAX_NODE_NUMBER; i++)
+    {
+        nodeContainer[i]->m_toBeDestroyed.store(true, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
manel94: @elBoberido is it the same case in this two old tests in the same file:
DestroyNodeDataAndAddNewNodeDataSucceeds and ConditionVariablesDestroy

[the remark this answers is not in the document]
elBoberido: @manel94 yes, they have the same issue. A bad implementation can potentially crash the test. They should also have an `ASSERT_*` instead of an `EXPECT_*` in order to immediately stop the test case and prevent a nullptr access later on.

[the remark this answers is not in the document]
manel94: Okay i will update it also as a refactoring

chiranjeevimaddi:
working on fixing the build

manel94:
@elBoberido  i pushed updates :D

elBoberido (commented):
Looking at the existing test in this file makes it a bad example on how to write tests. It seems this file can count as one of our sins from the past.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +675,57 @@ TEST_F(PortManager_test, PortDestroy)
...
+    }
+
+    // so now no one should be available
+    {
+        auto interp = m_portManager->acquireInterfacePortData(iox::capro::Interfaces::INTERNAL, m_Process);
+        EXPECT_EQ(interp, nullptr);
+    }
+
+    // set the destroy flag and let the discovery loop take care
+    for (uint32_t i = 0U; i < iox::MAX_INTERFACE_NUMBER; i++)
```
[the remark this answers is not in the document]
manel94: @elBoberido  do you want to use this template in all function using container and flags?
i mean ConditionVariablesDestroy, DestroyNodeDataAndAddNewNodeDataSucceeds tests  also.
template <typename vector >
void setDestroyFlagAndClearContainer(vector& Container, int id, int iterations)
{
    for (int i = 0; i < iterations; ++i)
    {
        Container[i]->m_toBeDestroyed.store(true, std::memory_order_relaxed);
    }
    m_portManager->doDiscovery();
    Container.clear();
}
for range based for loop to simplify the code i didn´t understand what you mean exactly.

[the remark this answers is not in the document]
elBoberido: @manel94 yes everywhere where this code snippet is used. You can use the ranged base for loop to get rid of the `iterations` e.g. `for(auto item: container) {...}`. `id` should also not be necessary and `m_portManager->doDiscovery();` should not be part of the method.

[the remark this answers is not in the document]
manel94: okay @elBoberido , it will be added in the next update, could you please add remarks need to be refactoring  to all the test file port manager, since i will modify some other things not only my added tests. of course if it´s necessary :D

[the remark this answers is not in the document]
elBoberido: @manel94 it's not right clear to me what the question was 😅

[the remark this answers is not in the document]
manel94: Sorry for that @elBoberido,  i mean in this test file i think the old test cases should be refactoring for example test names, using template..., if you could please not just review my added changes but also the old one that help me to refactor all the test file. if it is possible of course.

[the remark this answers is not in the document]
elBoberido: It's mainly regarding the size of the tests. By skimming over the file I wondered if it's possible to consolidate the `arrangement` of some tests in some template methods. And then there is this big `PortDestroy` test, where I'm also not quite sure how to make that more manageable.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, 2021 by Robert Bosch GmbH, Apex.AI Inc. All rights reserved.
+// Copyright (c) 2019, 2021 by Robert Bosch GmbH. All rights reserved.
```
JimmyBch, line 1: I guess you should add back Apex company.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +709,53 @@ TEST_F(PortManager_test, PortDestroy)
...
+
+    PublisherPortUser publisher(
+        m_portManager
+            ->acquirePublisherPortData({1, 1, 1}, publisherOptions, m_Process, m_payloadMemoryManager, PortConfigInfo())
+            .value());
+
+    publisher.offer();
+    m_portManager->doDiscovery();
+
+    EXPECT_EQ(initialCount + 1 == serviceCounter->load(), true);
```
elBoberido, line 760: please have a look at my previous comment for this line ;)

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -312,23 +326,37 @@ TEST_F(PortManager_test, InterfacesOverflow)
             });
 
         auto interfacePointer = m_portManager->acquireInterfacePortData(iox::capro::Interfaces::INTERNAL, "itfPenguin");
-        EXPECT_THAT(interfacePointer, Eq(nullptr));
+        EXPECT_EQ(interfacePointer, nullptr);
         EXPECT_TRUE(errorHandlerCalled);
     }
+}
+
+TEST_F(PortManager_test, deleteInterfacePortfromMaximumNumberAndAddOneIsSuccessful)
```
elBoberido, line 334: Nice split of the tests. Just one nitpicking, The names of all the other test cases start with a capital letter. Could you please also make it upper case.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -312,23 +326,37 @@ TEST_F(PortManager_test, InterfacesOverflow)
...
+{
+    std::string itf = "itf";
+
+    for (unsigned int i = 0; i < iox::MAX_INTERFACE_NUMBER; i++)
+    {
+        auto newItfName = itf + std::to_string(i);
+        auto interp = m_portManager->acquireInterfacePortData(
+            iox::capro::Interfaces::INTERNAL, iox::ProcessName_t(iox::cxx::TruncateToCapacity, newItfName));
+        EXPECT_NE(interp, nullptr);
+    }
```
elBoberido, line 344: This is also repeated quite often. If you like, you could put this in separate method and just call that one. Additionally, if the ports are needed for later use, the method could have an optional lambda which takes the port and could be used like this
```c++
std::vector<iox::popo::InterfacePortData*> interfaceContainer;
exhaustInterfacePorts([&](auto port){
    interfaceContainer.push(port);
});
```

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -467,6 +503,8 @@ TEST_F(PortManager_test, AcquiringOneMoreThanMaximumNumberOfNodesFails)
         iox::NodeName_t newNodeName(iox::cxx::TruncateToCapacity, node + std::to_string(i));
         auto nodeData = m_portManager->acquireNodeData(newProcessName, newNodeName);
         ASSERT_THAT(nodeData.has_error(), Eq(false));
+        EXPECT_THAT(nodeData.value()->m_node, StrEq(newNodeName));
+        EXPECT_THAT(nodeData.value()->m_process, StrEq(newProcessName));
```
elBoberido, line 507: Please don't merge these two tests. One is for acquiring the maximum number of nodes and one for the correct overflow handling.
Please have a look at #547 for the reason.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -624,19 +646,38 @@ TEST_F(PortManager_test, PortDestroy)
             EXPECT_THAT(subscriber1.getSubscriptionState(), Eq(iox::SubscribeState::WAIT_FOR_OFFER));
         }
     }
+}
 
-    // re-create the ports of process p2
-    publisherData2 =
+TEST_F(PortManager_test, CleanupProcessP2ChangesStatesOfPortsInProcessP1)
```
elBoberido, line 651: After looking at the entire test it seems this is more an integration test than a unit test and I think this needs to be reverted, since the semantic of the test are changed if it just split. In the original test, the `publisher1` is offering the all the time and the `subscriber1` is subscribed all the time. So `subscriber1` will automatically be re-subscribed when `publisher2` offers the service. After the split, this is not the case anymore.

manel94: okay so no need to split the previous test?

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
manel94: Hi @elBoberido, Do you mean something like this:

template typename vector

iox::popo::InterfacePortData* acquireInterfacePortData(vector& Container, std::string itf)
{
    for (auto item : Container)
    {
        auto newItfName = itf + std::to_string(item);
        auto interp = m_portManager->acquireInterfacePortData(
            iox::capro::Interfaces::INTERNAL, iox::ProcessName_t(iox::cxx::TruncateToCapacity, newItfName));
        ASSERT_NE(interp, nullptr);
    }
    return interp;
}

and in the test:

{
std::vector<iox::popo::InterfacePortData*> interfaceContainer;
    std::string itf = "itf";

    // first aquire all possible condition variables
    auto interp = acquireInterfacePortData(interfaceContainer, itf);
    auto exhaustInterfacePorts([&](auto interp){
    interfaceContainer.push(interp);});
...
i didn´t know how the template could take the optional lambda in this case.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
elBoberido: no, it's a big one but if it is split, it wouldn't mimic the previous behavior with re-subscription

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
elBoberido: This would be without templates, just for the specific port, e.g.
```
void exhaustInterfacePorts(std::function<void(iox::popo::InterfacePortData*)> f = std::function<void(iox::popo::InterfacePortData*)>())
{
    for(...)
    {
        auto port = ...
        ASSERT ...
        if (f)
        {
            f(port);
        }
    }
}
```
This could then be used like I mentioned in the previous comment

manel94: If I understand this is the implementation needed:

           void exhaustInterfacePorts(std::string itf , PortManagerTester* m_portManager, 
           std::function<void(iox::popo::InterfacePortData*)> f = std::function<void(iox::popo::InterfacePortData*)>())
           {
                  for (unsigned int i = 0; i < iox::MAX_INTERFACE_NUMBER; i++)
                  {
                       auto newItfName = itf + std::to_string(i);
                       auto port = m_portManager->acquireInterfacePortData(
                        iox::capro::Interfaces::INTERNAL, iox::ProcessName_t(iox::cxx::TruncateToCapacity, newItfName));
                        ASSERT_NE(port, nullptr);
                            if (f)
                            {
                                f(port);
                             }
                       }

                }
and in the test it will be:
               {
                      std::vector<iox::popo::InterfacePortData*> interfaceContainer;
                      std::string itf = "itf";

                       // first aquire all possible condition variables
                      exhaustInterfacePorts(itf, m_portManager,[&](auto port){
                      interfaceContainer.push_back(port);});
...
If yes this two tests also need two similar methods no?
AcquireConditionVariablesDataAfterDestroyingPreviouslyAcquiredOnesIsSuccessful, AcquireNodeDataAfterDestroyingPreviouslyAcquiredOnesIsSuccessful,

elBoberido: `m_portManager` is already part of the fixture and doesn't need to be passed to the method.
The interface name could be passed, yes.
Yes, the same could be done for the other tests were a resource is exhausted and where it is done multiple times

manel94: elBoberido, i have another question, how to deal with this expects in test AcquiringMaximumNumberOfNodesWorks:     
        EXPECT_THAT(nodeData.value()->m_node, StrEq(newNodeName));
        EXPECT_THAT(nodeData.value()->m_process, StrEq(newProcessName));
should i add it in the method related to node , since AcquireNodeDataAfterDestroyingPreviouslyAcquiredOnesIsSuccessful didn´t use those expects?

elBoberido: @manel94 I'm quite busy at the moment and come back later, but regarding the process name issue you can check if it is possible to always ave variable names and use it that way. Alternatively you can create an enum, e.g.
```
enum class NamePostfix
{
    OFF,
    COUNTER,
};
```
This can be passed as parameter to `exhaustInterfacePorts`.
One idea might also be to let `exhaustInterfacePorts` return the number of successfully acquired ports. The assert cound then be done on that number instead of in `exhaustInterfacePorts`.

manel94: @elBoberido could you please explain more, i didn´t get you, i didn´t have any problem with exhaustInterfacePorts, my question is  how to deal with this test  AcquiringMaximumNumberOfNodesWorks to use exhaustNodePorts.

elBoberido: @manel94 that was meant for the issue with the static and variable names. There is another PR which extends the tests for this file and there the method is called `acquireMaxNumberOfEventVariables`, so it might be a good idea to follow that naming.
So this would become `acquireMaxNumberOfNodePorts`. The `ASSERT_*` would only be necessary for `EXPECT_THAT(nodeData.has_error(), Eq(false));` the other ones can be changed but are not necessary since it doesn't have an potential fatal effect later in the test.
Those checks could either kept in the `acquireMaxNumberOfNodePorts` test or done in the lambda in one specific test case. The lambda would then take three parameters instead of one

manel94:
requested changes are pushed :D

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -337,7 +408,7 @@ TEST_F(PortManager_test, ApplicationsOverflow)
         auto newAppName = app + std::to_string(i);
         auto appp =
```
elBoberido, line 409: please rename this to `appPort`

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -482,70 +551,48 @@ TEST_F(PortManager_test, AcquiringOneMoreThanMaximumNumberOfNodesFails)
...
     // delete one and add one NodeData should be possible now
     unsigned int i = 0U;
     iox::ProcessName_t newProcessName(iox::cxx::TruncateToCapacity, process + std::to_string(i));
     iox::NodeName_t newNodeName(iox::cxx::TruncateToCapacity, node + std::to_string(i));
+    // this is done because there is no removeNodeData method in the PortManager class
     m_portManager->deletePortsOfProcess(newProcessName);
 
-    nodeData = m_portManager->acquireNodeData(newProcessName, newNodeName);
+    auto nodeData = m_portManager->acquireNodeData(newProcessName, newNodeName);
     EXPECT_THAT(nodeData.has_error(), Eq(false));
```
elBoberido, line 570: this must be an `ASSERT_THAT` to abort the test in case there is no value to access

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -482,70 +551,48 @@ TEST_F(PortManager_test, AcquiringOneMoreThanMaximumNumberOfNodesFails)
...
-    for (unsigned int i = 0U; i < iox::MAX_NODE_NUMBER; i++)
-    {
-        auto nodeData = m_portManager->acquireNodeData(process, node);
-        EXPECT_THAT(nodeData.has_error(), Eq(false));
-        nodeContainer.push_back(nodeData.value());
-    }
+    exhaustNodePorts(node, process, [&](auto NodePort, auto newNodeName, auto newProcessName) {
+        nodeContainer.push_back(NodePort);
+        EXPECT_THAT(NodePort->m_node, StrEq(newNodeName));
+        EXPECT_THAT(NodePort->m_process, StrEq(newProcessName));
```
elBoberido, line 586: this checks can be removed, since it is already tested

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +714,46 @@ TEST_F(PortManager_test, PortDestroy)
         }
     }
 }
+
+TEST_F(PortManager_test, AcquireInterfacePortDataAfterDestroyingPreviouslyAcquiredOnesIsSuccessful)
```
elBoberido, line 718: could you please move this test to the other interface port tests

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +714,46 @@ TEST_F(PortManager_test, PortDestroy)
...
+    std::string itf = "itf";
+
+    // first aquire all possible interfaces
+    exhaustInterfacePorts(itf, [&](auto InterafcePort) { interfaceContainer.push_back(InterafcePort); });
+
+    // so now no one should be available
+    {
+        auto interp = m_portManager->acquireInterfacePortData(iox::capro::Interfaces::INTERNAL, m_Process);
+        EXPECT_EQ(interp, nullptr);
+    }
```
elBoberido, line 730: this is already checked by `AcquiringOneMoreThanMaximumNumberOfInterfacesFails` and can be removed

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +714,46 @@ TEST_F(PortManager_test, PortDestroy)
...
+    // set the destroy flag and let the discovery loop take care
+    setDestroyFlagAndClearContainer(interfaceContainer);
+    m_portManager->doDiscovery();
+
+    // so we should able to get some more now
+    for (uint32_t i = 0U; i < iox::MAX_INTERFACE_NUMBER; i++)
+    {
+        auto interp = m_portManager->acquireInterfacePortData(iox::capro::Interfaces::INTERNAL, m_Process);
+        EXPECT_NE(interp, nullptr);
+    }
```
elBoberido, line 741: can this be replace by `exhaustInterfacePorts`?

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +714,46 @@ TEST_F(PortManager_test, PortDestroy)
...
+}
+
+TEST_F(PortManager_test, OfferPublisherServiceUpdatesServiceRegistryChangeCounter)
+{
+    auto serviceCounter = m_portManager->serviceRegistryChangeCounter();
```
elBoberido, line 746: could you pleas `ASSERT` that `serviceCounter` in not a `nullptr`

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -668,3 +714,46 @@ TEST_F(PortManager_test, PortDestroy)
...
+    auto initialCount = serviceCounter->load();
+    PublisherOptions publisherOptions{1};
+
+    PublisherPortUser publisher(
+        m_portManager
+            ->acquirePublisherPortData({1, 1, 1}, publisherOptions, m_Process, m_payloadMemoryManager, PortConfigInfo())
+            .value());
```
elBoberido, line 753: could you please acquire the publisher port data separately and `ASSERT` that is has no error before using `vlaue()`

elBoberido (commented):
The code is mostly good. Most of the requests are regarding code style and most of that is due to the existing code.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -385,89 +506,69 @@ TEST_F(PortManager_test, ConditionVariablesOverflow)
                 errorHandlerCalled = true;
             });
 
-        auto conditionVariableDataResult = m_portManager->acquireConditionVariableData("AnotherToad");
-        EXPECT_TRUE(conditionVariableDataResult.has_error());
+        auto conditionVariablePort = m_portManager->acquireConditionVariableData("AnotherToad");
```
elBoberido, line 509: this is also not a port and our convention for a variable which takes a `cxx::expected` is to prepend `Result`, so this would become `conditionVariableResult`

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -476,107 +577,78 @@ TEST_F(PortManager_test, AcquiringOneMoreThanMaximumNumberOfNodesFails)
             errorHandlerCalled = true;
         });
 
-    auto nodeData = m_portManager->acquireNodeData("AnotherProcess", "AnotherNode");
-    EXPECT_THAT(nodeData.has_error(), Eq(true));
+    auto nodePort = m_portManager->acquireNodeData("AnotherProcess", "AnotherNode");
```
elBoberido, line 580: similar here, just `nodeResult` or `nodeDataResult`

elBoberido (commented):
almost there

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -119,8 +125,87 @@ class PortManager_test : public Test
...
+    acquireMaxNumberOfNodes(std::string nodeName,
+                            std::string processName,
+                            std::function<void(iox::runtime::NodeData*, iox::NodeName_t, iox::ProcessName_t)> f =
+                                std::function<void(iox::runtime::NodeData*, iox::NodeName_t, iox::ProcessName_t)>())
+    {
+        for (unsigned int i = 0U; i < iox::MAX_NODE_NUMBER; i++)
+        {
+            iox::ProcessName_t newProcessName(iox::cxx::TruncateToCapacity, processName + std::to_string(i));
+            iox::NodeName_t newNodeName(iox::cxx::TruncateToCapacity, nodeName + std::to_string(i));
+            auto nodePort = m_portManager->acquireNodeData(newProcessName, newNodeName);
```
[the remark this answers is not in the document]
mossmaurice: What about renaming it back to `SharedMemoryManager`? :smile: Just kidding :stuck_out_tongue: Isn't it sort of a factory for data part of the objects?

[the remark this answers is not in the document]
elBoberido: so something like a `SharedMemoryDataFactory` or `SharedMemoryMangementDataFactory` :smile:

elBoberido (approved):
Looks good. Thank's for the contribution

dkroenke (approved):

PR #530, opened 2021-01-29
Iox #496 improve coverage and refactor ServiceDescription moduletests

### LINKED ISSUE #496: Add unit test for iceoryx-posh
## For iceoryx is more testing needed to achieve good code coverage

add more unit/module tests for following files

- [x] capro/service_description.cpp 
- [ ] mepoo/shared_chunk.cpp 
- [ ] mepoo/ mem_pool.cpp 
- [ ] mepoo/memory_manager.cpp 
- [ ] mepoo/ mepoo_config.cpp 
- [ ] runtime/shared_memory_user.cpp 
- [ ] runtime/posh_runtime_single_process.cpp 
- [ ] runtime/posh_runtime.cpp 
- [x] runtime/node.cpp 
- [ ] gateway/gateway_base.cpp 
- [ ] gateway/gateway_config.cpp 
- [ ] gateway/toml_gateway_config_parser.cpp 
- [ ] roudi/port_pool.cpp 
- [ ] roudi/roudi_process.cpp 
- [ ] roudi/ roudi_config.cpp 
- [ ] roudi/port_manager.cpp 
- [ ] roudi/roudi_cmd_line_parser.cpp 
- [ ] roudi/roudi_cmd_line_parser_config_file_option.cpp 
- [ ] roudi/memory/memory_block.cpp 
- [ ] roudi/memory/roudi_memory_manager.cpp 
- [ ] roudi/memory/memory_provider.cpp 
- [ ] roudi/application/ iceoryx_roudi_app.cpp 

- [x] Define a default constructor ServiceDescription::ClassHash 

reference for code-coverage
https://codecov.io/gh/eclipse-iceoryx/iceoryx?branch=master

nihalchari:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
code contributed by @georgemathewjerrin

1. [x] Merge

## References

- Partly closes #496


iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,649 @@ using namespace iox::capro;
...
+
+TEST_F(ServiceDescription_test,
+       ServiceDescriptionIsInvalidWhen_m_hasServiceOnlyDescriptionIsFalseAndServieInstanceAndEventIDsAreValid)
+{
+    uint16_t testServiceID = 1U;
+    uint16_t testEventID = 1U;
+    uint16_t testInstanceID = 1U;
+    ServiceDescription serviceDescription1 = ServiceDescription(testServiceID, testEventID, testInstanceID);
+
+    EXPECT_TRUE(serviceDescription1.isValid());
```
dkroenke, line 651: I don't see any test where a ServiceDescription is valid with `m_hasServiceOnlyDescription = true`

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,649 @@ using namespace iox::capro;
...
-    EXPECT_THAT(csd.getServiceID(), Eq(1));
-    EXPECT_THAT(csd.getInstanceID(), Eq(2));
-    EXPECT_THAT(csd.getEventID(), Eq(3));
-    EXPECT_THAT((csd.getClassHash())[0], Eq(testHash[0]));
-    EXPECT_THAT((csd.getClassHash())[1], Eq(testHash[1]));
-    EXPECT_THAT((csd.getClassHash())[2], Eq(testHash[2]));
-    EXPECT_THAT((csd.getClassHash())[3], Eq(testHash[3]));
-    EXPECT_THAT(csd.hasServiceOnlyDescription(), Eq(true));
-    EXPECT_THAT(csd.isInternal(), Eq(true));
+                                                     true,
```
dkroenke, line 56: I see this as problem here because we can create a servicedescription through the serialization with the three friends and `hasServiceOnlyDescription()` returns true, which should not be the case.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,649 @@ using namespace iox::capro;
...
+    EXPECT_THAT(std::to_string(serviceDescription1.getServiceID()), serviceDescription1.getServiceIDString());
+    EXPECT_THAT(std::to_string(serviceDescription1.getEventID()), serviceDescription1.getEventIDString());
+    EXPECT_THAT(std::to_string(serviceDescription1.getInstanceID()), serviceDescription1.getInstanceIDString());
+}
+
+TEST_F(ServiceDescription_test, ServiceDescriptionDefaultCtorInitializesTheIDsAndStringsToZero)
+{
+    ServiceDescription serviceDescription1 = ServiceDescription();
+
+    EXPECT_EQ(serviceDescription1.getServiceID(), 0);
```
dkroenke, line 154: `getServiceID()` returns unsigned integer, same below.

dkroenke (changes requested):
See the comments

dkroenke:
@nihalchari the PR title is misleading because it implies that capro is refactored which is not the case here. Please rename it.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,649 @@ using namespace iox::capro;
...
 }
 
-TEST_F(ServiceDescription_test, operatorEq)
+TEST_F(ServiceDescription_test, ServiceDescriptionStringCtorWithOutOfBoundaryIntegerStringValuesSetTheIDsToInvalid)
 {
-    EXPECT_TRUE(csd1 == csd1Eq);
-    EXPECT_TRUE(csd2 == csd2Eq);
-    EXPECT_TRUE(csd3 == csd3Eq);
-    EXPECT_TRUE(csd4 == csd4Eq);
+    IdString_t testService("65536");
```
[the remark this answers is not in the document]
georgemathewjerrin: I didnt understand. testService is a string right ?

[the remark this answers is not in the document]
dkroenke: This is correct, but magic numbers are not good to understand. My advise is to initialize the ServiceDescription here with uint instead of a string. This make more clear that you do an out-of-bounds check here.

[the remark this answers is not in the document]
georgemathewjerrin: Yeah, true. But string is required to cover the branch of the constructor in lines 136, 142 and 148 of servicedescription.cpp. To cover those branch, I think the value has to be a string like "65536".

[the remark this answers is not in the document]
dkroenke: Ah yes, i was misled by the two possible cases. Nevertheless there should be an explicit out-of-bound test that use the uint Ctor and `iox::capro::AnyService + 1` to verify that too.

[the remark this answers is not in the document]
georgemathewjerrin: I'm getting overflow error while trying to build the test with iox::capro::AnyService + 1 as value to uint16_t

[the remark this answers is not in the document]
dkroenke: Uh yes, the Ctor takes uint16_t as type. Then the test makes of course no sense. Please leave a comment in this test then where the value 65536 comes from. You can do that with `/// @note` as doxygen comment.

iceoryx_posh/test/moduletests/test_capro_service.cpp
georgemathewjerrin: Actually , there are test cases for it. But my description wasn't sufficient enough. I have updated the description of those test cases.

iceoryx_posh/test/moduletests/test_capro_service.cpp
georgemathewjerrin: Sorry, I didn't get you. Should there be a refactoring of the existing code or the test ?

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,649 @@ using namespace iox::capro;
...
+    uint16_t testServiceID = 1U;
+    uint16_t testEventID = 2U;
+    uint16_t testInstanceID = 3U;
+    ServiceDescription serviceDescription1 = ServiceDescription(testServiceID, testEventID, testInstanceID);
+    ServiceDescription serviceDescription2 = ServiceDescription(testServiceID, testEventID, testInstanceID);
+
+    EXPECT_FALSE(serviceDescription1 != serviceDescription2);
+}
+
+TEST_F(ServiceDescription_test, ServiceDescriptionClassHashCtorCreatesClassHashWithValuesPassedToTheCtor)
```
dkroenke, line 371: There are some test cases for the ClassHash missing. For example the `[]` operator or the initialization of it.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,697 @@ using namespace iox::capro;
...
+       ServiceDescriptionIsInvalidWhen_m_hasServiceOnlyDescriptionIsTrueAndServiceStringIsInvalid)
+{
+    uint16_t testServiceID = iox::capro::AnyService;
+    uint16_t testEventID = iox::capro::AnyEvent;
+    uint16_t testInstanceID = iox::capro::AnyInstance;
+    testService = iox::capro::InvalidIDString;
+    testEvent = "validevent";
+    testInstance = "validinstance";
+    ServiceDescription::ClassHash testHash{1U, 2U, 3U, 4U};
+    auto serialObj = iox::cxx::Serialization::create(testService,
```
dkroenke, line 509: There no need to use the Serialization here, just use the normal Ctors of the ServiceDescription, they set for you the bool then rightly. The same with the tests below.
Generally the purpose of the Serialization is not to be an alternative Constructor. It is intended to send/receive the ServiceDescription over communication protocols which transfers strings like the MessageQueue.

iceoryx_posh/test/moduletests/test_capro_service.cpp
dkroenke: The problem is that it is possible to set `m_hasServiceOnlyDescription` wrongly.
What if you pass three valid IDs and set `m_hasServiceOnlyDescription` manually to true? Then we have a Servicedescription with a valid event ID but it is treated as reduced ServiceDescription with instance and service id only.

dkroenke (changes requested):

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,697 @@ using namespace iox::capro;
...
+TEST_F(ServiceDescription_test, TwoServiceDescriptionsWithDifferentInstanceStringsAreNotEqual)
+{
+    uint16_t testServiceID = 1U;
+    uint16_t testEventID = 2U;
+    uint16_t testInstanceID = 3U;
+    testService = "1";
+    testEvent = "2";
+    testInstance = "instance";
+    ServiceDescription::ClassHash testHash = {1U, 2U, 3U, 4U};
+    auto serialObj = iox::cxx::Serialization::create(testService,
```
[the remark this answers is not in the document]
georgemathewjerrin: Okay, I understand. But then, there shouldn't be a provision for the user to set it right ?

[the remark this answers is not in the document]
georgemathewjerrin: Also to cover every branch in the == operator, serialization is needed.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,697 @@ using namespace iox::capro;
...
+    uint16_t serviceID2 = 2U;
+    ServiceDescription description1 = ServiceDescription(serviceID1, iox::capro::AnyEvent, iox::capro::AnyInstance);
+    ServiceDescription description2 = ServiceDescription(serviceID2, iox::capro::AnyEvent, iox::capro::AnyInstance);
+
+    EXPECT_FALSE(iox::capro::serviceMatch(description1, description2));
+}
+
+TEST_F(ServiceDescription_test, ClassHashIndexOperatorReturnsTheValueInTheCorrespondingIndex)
+{
+    ServiceDescription::ClassHash testHash{1U, 2U, 3U, 4U};
```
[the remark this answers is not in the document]
georgemathewjerrin: I tested now. Since there are no default values, it assigns garbage values to the data array.

iceoryx_posh/test/moduletests/test_capro_service.cpp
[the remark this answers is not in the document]
dkroenke: > Okay, I understand. But then, there shouldn't be a provision for the user to set it right ?
> Also to cover every branch in the == operator, serialization is needed.

The thing is that for a correct transfer of the ServiceDescription, all member variables need to be set by the Serialization. Currently i tend more into the direction that the Serialization expects a correct ServiceDescription as a contract. A clean solution would it be to prevent potential misuse.

In the comparison operator of the ServiceDescription are only members compared which can be set by the normal Ctors. Why is here the serialization needed?

The problem i see here is that you construct a ServiceDescription which has set internally `m_instanceString = "instance"` and `m_instanceID = 3` which does happen when you go the normal way with the constructor.
The serialization is here used as "backdoor" to inject the internal variables directly. This is not forbidden because it's public API but the serialization should normally not be used for that.
Keep in mind that also developers which are new to iceoryx look at tests to get an idea how the classes are used. They could get the wrong impression that the serialization is an alternative constructor, which we not want.

Maybe it's a good idea to set a doxygen tag `/// @attention` to make that visible.

dkroenke (changes requested):

iceoryx_posh/test/moduletests/test_capro_service.cpp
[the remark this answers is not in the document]
georgemathewjerrin: From my understanding,  its not possible to specifically set separate values for ID and the corresponding string because the constructor converts the string to integer to store it in the ID and vice versa to store it in the corresponding string. Nevertheless, for the branch coverage that I mentioned, Invalid ID case can be used to cover the branch and as you said, serialization could be removed. I'm new to this project and don't know much about it so please correct me if I'm wrong . :)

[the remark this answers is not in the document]
dkroenke: > From my understanding, its not possible to specifically set separate values for ID and the corresponding string because the constructor converts the string to integer to store it in the ID and vice versa to store it in the corresponding string.

Yes correct, the "three friends" instance, service and event shall either be set as string or as unsiged integers. Mixed initialization is normally not allowed for the end-user. This is something which is not well documented and we need to improve on that.

iceoryx_posh/test/moduletests/test_capro_service.cpp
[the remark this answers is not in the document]
dkroenke: Looks like there is wrong behavior, very good! The patch is quite simple and you can directly add it to the PR.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +28,631 @@ using namespace iox::capro;
...
+
+TEST_F(ServiceDescription_test, LessThanOperatorReturnsFalseIfEventStringOfFirstServiceDescriptionIsLessThanSecond)
+{
+    ServiceDescription serviceDescription1("TestService", "TestInstance", "TestEvent1");
+    ServiceDescription serviceDescription2("TestService", "TestInstance", "TestEvent2");
+
+    EXPECT_FALSE(serviceDescription1 < serviceDescription2);
+}
+
+/// END SERVICEDESCRIPTION TESTS
```
dkroenke, line 658: An endline is missing

dkroenke:
You can take a look on https://github.com/eclipse-iceoryx/iceoryx/pull/547 for more infos and helpful hints about test practices in iceoryx.

georgemathewjerrin:
sd

@dkroenke https://github.com/nihalchari/iceoryx/blob/2a23de4488021f14531913f86e5c97066b3a282f/iceoryx_posh/source/capro/service_description.cpp#L38

On this line, shouldn't there be a boundary check for the index because the data array has only 4 values ? Most of the time, the operator[] is used inside a loop which limits the index ,but still, it isn't safe right ?

dkroenke:
> On this line, shouldn't there be a boundary check for the index because the data array has only 4 values ? Most of the time, the operator[] is used inside a loop which limits the index ,but still, it isn't safe right ?

Yes, it would be good to have a range based check in here to avoid Misuse by the end-user. Perfect would it be that the user is informed at compile-time about an error here.
We have for that purpose some helpers in the iceoryx utils included which do the job. I guess we can use the `cxx::range` here to  achieve that: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/helplets.hpp#L91 

A practical example of the quite similar `cxx::greater_or_equal` construct of the helplets can be seen here: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/source/mepoo/mem_pool.cpp#L28

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2019, 2021 by Robert Bosch GmbH. All rights reserved.
```
dkroenke, line 1: Can you please rework the fileheader according to the new guidelines in https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#header?

nihalchari: done

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
```diff
@@ -83,15 +84,17 @@ class ServiceDescription
...
-        uint32_t& operator[](const uint64_t index) noexcept;
-        const uint32_t& operator[](const uint64_t index) const noexcept;
-        bool operator==(const ClassHash& rhs) const noexcept;
-        bool operator!=(const ClassHash& rhs) const noexcept;
-
       private:
         static constexpr size_t CLASS_HASH_ELEMENT_COUNT{4u};
         uint32_t data[CLASS_HASH_ELEMENT_COUNT];
+
+      public:
```
elfenpiff, line 91: Why is it arranged in that way? In a struct everything is public until it is not stated as private. If you move this up then everything is public again.

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -33,12 +38,15 @@ ServiceDescription::ClassHash::ClassHash(const std::initializer_list<uint32_t>&
     }
 }
 
-uint32_t& ServiceDescription::ClassHash::operator[](const uint64_t index) noexcept
+uint32_t& ServiceDescription::ClassHash::operator[](
+    iox::cxx::range<uint64_t, 0U, CLASS_HASH_ELEMENT_COUNT - 1> index) noexcept
```
elfenpiff, line 42: Nice usage, I have totally forgotten that we have such cool constructs!

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +30,653 @@ using namespace iox::capro;
...
+{
+    ServiceDescription::ClassHash testHash{1U, 2U, 3U, 4U};
+
+    EXPECT_EQ(uint32_t(1), testHash[0]);
+    EXPECT_EQ(uint32_t(2), testHash[1]);
+    EXPECT_EQ(uint32_t(3), testHash[2]);
+    EXPECT_EQ(uint32_t(4), testHash[3]);
+}
+
+TEST_F(ServiceDescription_test, ClassHashIndexOperatorReturnsTheValueInTheCorrespondingIndex)
```
elfenpiff, line 62: This test is identical to the previous when, the only difference is that you use once EXPECT_EQ and once EXPECT_THAT.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +30,653 @@ using namespace iox::capro;
...
 
-TEST_F(ServiceDescription_test, operatorAssign)
+TEST_F(ServiceDescription_test, ServiceDescriptionStringCtorCreatesServiceDescriptionWithValuesPassedToTheCtor)
 {
-    ServiceDescription csdNew = csd1;
-    EXPECT_TRUE(csdNew == csd1);
+    testService = "1";
+    testInstance = "2";
+    testEvent = "3";
+    ServiceDescription::ClassHash testHash = {1U, 2U, 3U, 4U};
```
elfenpiff, line 290: You always use the number 1,2,3,4. This is alright but in rare cases, when we have for instance have a memory glitch and are reusing the old memory - caused by a bug - we would not discover this when we always use the same inputs. Therefore I would suggest to vary the input in every test a bit.
Add some additional integers to every number in the hash and you will be safe.

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -28,166 +30,653 @@ using namespace iox::capro;
...
+
+TEST_F(ServiceDescription_test, ServiceDescriptionStringCtorWithOutOfBoundaryIntegerStringValuesSetTheIDsToInvalid)
 {
-    ServiceDescription desc1(1u, 2u);
-    ServiceDescription desc2("bla", "fuh");
-    ServiceDescription desc3(1u, 2u, 3u);
-    ServiceDescription desc4("bla", "fuh", "dumbledoodoo");
+    /// @note The value 65536 is assigned to the strings to set the condition to out of boundary when string is
+    /// converted to uint_16t.
+    testService = "65536";
```
elfenpiff, line 337: Instead of a comment try to write more intentional code. for instance to make it clear that you would like to have a uint16_t overflow you could write
```cpp
testService = std::to_string(uint32_t(1) + std::numeric_limits<uint16_t>::max() );
```

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
nihalchari: Yeah, Actually I have used "CLASS_HASH_ELEMENT_COUNT" macro in declaration of operator[], with previous arrangement it was no able to find that macro.

dkroenke: The cleaner way is to put `CLASS_HASH_ELEMENT_COUNT` near to the other defined constexpr variables in namespace `iox::capro` at the top of the file. Then you don't need to reorder private/public defines and it can be used by the end-user for cleanly iterating over the ClassHash.

iceoryx_posh/test/moduletests/test_capro_service.cpp
georgemathewjerrin: I'm getting error if I'm trying to do this way.

no match for ‘operator=’ (operand types are ‘iox::capro::IdString_t {aka iox::cxx::string<100>}’ and ‘std::__cxx11::string {aka std::__cxx11::basic_string<char>}

dkroenke: You need to create a `cxx::string` from the `std::string`, here is described how to do this: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/string.hpp#L204
You can directly inject the std::to_string statement in the constructing call for the `cxx::string`

georgemathewjerrin: yeah, I just found it out :)

dkroenke (changes requested):
@nihalchari @marthtz the only thing left here are the correct copyright headers. Otherwise i think this PR is ok to merge. Please merge latest master into your branch for adding SPDX license statement.

dkroenke (approved):

marthtz (approved):

PR #532, opened 2021-01-29
Iox #349 introduce node in C API

### LINKED ISSUE #349: Introduction of Runnables/Nodes for the new APIs

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

1. [ ] Merge

## References

- partly closes #349


elBoberido (approved):

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -29,15 +29,17 @@ extern "C" {
...
+                       const char* const event,
+                       const uint64_t historyCapacity,
+                       const char* const nodeName)
 {
     new (self) cpp2c_Publisher();
     iox_pub_t me = reinterpret_cast<iox_pub_t>(self);
-    iox::popo::PublisherOptions options;
+    PublisherOptions options;
     options.historyCapacity = historyCapacity;
+    options.nodeName = NodeName_t(TruncateToCapacity, nodeName);
```
FerdinandSpitzschnueffler, line 42: I decided against introducing a pub/sub options struct similar to the PublisherOptions in the C++ API because in C we cannot ensure that the struct is initialized and therefore introduce undefined behavior when assigning to the C++ PublisherOptions.

elBoberido: While it's true that this can lead to undefined behavior, there is the same problem without using a struct
```c
const uint64_t historyCapacity;
const char nodeName[100];
iox_pub_storage_t publisherStorage;
iox_pub_t publisher = iox_pub_init(&publisherStorage, "Radar", "FrontLeft", "Object", historyCapacity, nodeName)
```

The C people have to deal with this stuff all the time and it's not uncommon to call an init function before a struct can be used. So the example above would look like this with a struct
```c
iox_pub_options_t publisherOptions;
iox_pub_init_options(publisherOptions);
publisherOptions.historyCapacity  = 10U;
strncpy(publisherOptions.nodeName, "iox-c-publisher-node");
iox_pub_t publisher = iox_pub_init(&publisherStorage, "Radar", "FrontLeft", "Object", &publisherOptions);
```
Using default options if a nullptr was passed to `iox_pub_init` instead of `publisherOptions`;
I think this can be done with the API cleanup issue, though.

iceoryx_examples/waitset_on_c/ice_c_waitset_publisher.c
```diff
@@ -35,9 +34,10 @@ void sending()
 {
     iox_runtime_init("iox-c-ex-waitset-publisher");
 
-    uint64_t historyRequest = 0U;
+    const uint64_t historyCapacity = 0U;
```
budrus, line 37: Nice

budrus (approved):

iceoryx_binding_c/source/c_publisher.cpp
elBoberido: For parameter only you would basically have to set them all because not all of them are null-able. With the struct you could decide that passing a `nullptr` is equal to a struct where only `init` was called.
`init` would also set the same defaults like in the C++ API

FerdinandSpitzschnueffler: @elBoberido you've convinced me. This will be done in #408.

Issue #542, opened 2021-02-04
Fix resource leak caused by move ctor/assignment


elfenpiff:
## Required information

All versions

**Observed result or behaviour:**
The classes:
 - [x] `unix_domain_socket` (fixed by https://github.com/eclipse-iceoryx/iceoryx/pull/550)
 - [x] `shared_memory`
 - [x] `memory_map`
 - [x] `GenericRAII`

are not cleaning up the resources while performing move. This could lead to a resource leak when for instance the following pseudo code is executed.

```cpp
SharedMemory mySharedMemory; // create a shared memory file A
SharedMemory anotherSharedMemory; // create a shared memory file B

mySharedMemory = std::move(anotherSharedMemory); // mySharedMemory takes ownership of B without releasing A

// at this point the shared memory file A is leaked

```

**Classes with correct move resource handling:**
- `Semaphore`
- `MessageQueue`
- `optional`
- `variant`

**Handle undefined state:**
- [x] `optional`
- [x] `expected`

**creation pattern:**
- [x] `shared_memory`
- [x] `shared_memory_object`
- [x] `memory_map`

**Fix incorrect move CTor/assigned in all classes derived from `DesignPattern::Creation`**
According to: https://isocpp.org/wiki/faq/assignment-operators every move ctor/assignment operator has to be implemented like
```cpp
Derived& Derived::operator= (Derived&& d)
{
  // self-assignment is not allowed in move assignment
  Base::operator= (std::move(d));
  // Do the rest of your assignment operator here...
  return *this;
}
```
- [x] `shared_memory`
- [x] `shared_memory_object`
- [x] `memory_map`
- [x] `semaphore`
- [x] `message_queue`
- [x] `unix_domain_socket`
- [x] Write a detailed documentation how to use the creation pattern correctly - especially the assignment operator part!

**Fix sighandler**
- [x] Write mini `sigaction` abstraction and put it into utils to remove the 5 lines of code.
- [x] Use `posix::SignalHandler` everywhere

### Details
* Consider [this](https://github.com/eclipse-iceoryx/iceoryx/pull/550#discussion_r573629792) discussion

elfenpiff:
@sschnitzer Thank you for finding this!

Issue #418, opened 2020-12-08
refactor smart_c to use contract by design and expected


elfenpiff:
## Brief feature description

The `SmartC` class should get the two methods `and_then` and `or_else` so that it can be used like the `optional` or `expected.

## Detailed information

* [ ] remove `getErrorString` from SmartC
* [ ] make all methods in SmartC private (even ctor since `makeSmartC` is only using this)
* [ ] consider removing the SmartC class and use only free functions
* [ ] It should be possible to use `SmartC` like
```cpp
cxx::makeSmartC(sigaction, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, SIGINT, &act, nullptr)
     .and_then([&](auto sigActionReturnValue){ /* do stuff */ })
     .or_else([&](auto sigActionErrorReturn){ /* report error */ });
```

* [ ] Find a nicer solution for this use case where ignoring `SomeErrorCode`:
```
        if (smartcCall.hasErrors())
        {
            // sad path
        }
        else if (connectCall.getErrNum() == SomeErrorCode)
        {
           // something inbetween happy and sad
        }
        else
        {
            // happy path
        }
```

PR #550, opened 2021-02-10
Iox #381 Fix flooding the command line and resource leak

### LINKED ISSUE #381: Integration of Unix Domain Sockets
## Brief feature description

Currently the application name has to start with a leading slash. This exposes an implementation details, since that name is used for the message queue name. Since there is already a Unix Domain Abstraction available we can switch now to UDS.

## Detailed information
The default IPC protocol for communication between RouDi and Runtime shall be switched to UDS.
Since the registration at RouDi will happen with with a different IPC mechanism than a message queue, e.g. unix domain sockets, this implementation detail should not be exposed to the user. If the application name is used as name for an IPC channel and this channel has certain requirements for its name, those should be taken care of in the IPC abstraction.

* [x] The check for the leading slash should be removed from the `PoshRuntime`.
* [x] Switch to UDS in `posh_types.hpp` and adapt existing classes and test to the new protocol
  * [x] Rename abstract interface from "MQ" to "IPC" e.g. in RouDi
  * [x] Temporarly disabled outdated and broken `MqInterface*` tests 
* [x] Refactor classes for `MqBase` and `MqInterfaces` to more generic names
  * [x] Split into three hpp/cpp
  * [x] Rename to `IpcInterfaceCreator`, `IpcInterfaceUser`, `IpcBase`, `IpcRuntimeInterface` and `IpcMessage`
  * [x] Delete `mqueue_MOCK`
  * [x] Re-write `iceoryx_posh/test/moduletests/test_mq_interface.cpp` to test only POSH specific functionality of the IpcInterface e.g. `getShmTopic()`, IpcChannel is already covered by `iceoryx_utils/test/moduletests/test_ipc_channel.cpp` (will be addressed in  #332)
* [x] Fix output when publisher or subscriber is started before RouDi, command line is flooded with errors

See also https://github.com/eclipse/iceoryx/pull/373/files#r525981221

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Remove noisy output in `UnixDomainSocket`
* Loop till the `IpcChannelSide::Server` is created, this allows users to be able to start publisher and subscribers before RouDi is started
* Fix two resource leaks by introducting `closeFd(..)`

1. [x] Merge

## References

- Partly closes #381 and #542

### DIFF
--- a/iceoryx_posh/source/runtime/message_queue_interface.cpp
+++ b/iceoryx_posh/source/runtime/message_queue_interface.cpp
@@ -158,9 +158,20 @@ bool MqBase::openMessageQueue(const posix::IpcChannelSide channelSide) noexcept
     m_mq.destroy();
 
     m_channelSide = channelSide;
-    IpcChannelType::create(
-        m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
-        .and_then([this](auto& mq) { this->m_mq = std::move(mq); });
+
+    posix::IpcChannelError receivedError{posix::IpcChannelError::UNDEFINED};
+    // In case the server is not present yet (ENOENT), we try again
+    do
+    {
+        IpcChannelType::create(
+            m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
+            .and_then([this, &receivedError](auto& mq) {
+                this->m_mq = std::move(mq);
+                receivedError = posix::IpcChannelError::UNDEFINED;
+            })
+            .or_else([&receivedError](posix::IpcChannelError& error) { receivedError = error; });
+
+    } while (receivedError == posix::IpcChannelError::NO_SUCH_CHANNEL);
 
     return m_mq.isInitialized();
 }
--- a/iceoryx_posh/source/runtime/message_queue_interface.cpp
+++ b/iceoryx_posh/source/runtime/message_queue_interface.cpp
@@ -158,20 +158,9 @@ bool MqBase::openMessageQueue(const posix::IpcChannelSide channelSide) noexcept
     m_mq.destroy();
 
     m_channelSide = channelSide;
-
-    posix::IpcChannelError receivedError{posix::IpcChannelError::UNDEFINED};
-    // In case the server is not present yet (ENOENT), we try again
-    do
-    {
-        IpcChannelType::create(
-            m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
-            .and_then([this, &receivedError](auto& mq) {
-                this->m_mq = std::move(mq);
-                receivedError = posix::IpcChannelError::UNDEFINED;
-            })
-            .or_else([&receivedError](posix::IpcChannelError& error) { receivedError = error; });
-
-    } while (receivedError == posix::IpcChannelError::NO_SUCH_CHANNEL);
+    IpcChannelType::create(
+        m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
+        .and_then([this](auto& mq) { this->m_mq = std::move(mq); });
 
     return m_mq.isInitialized();
 }
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
@@ -132,7 +132,7 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     /// @brief creates the unix domain socket
     /// @param mode blocking or non_blocking
     /// @return int with the socket file descriptor, IpcChannelError if error occured
-    cxx::expected<int32_t, IpcChannelError> createSocket(const IpcChannelMode mode) noexcept;
+    cxx::expected<IpcChannelError> initalizeSocket(const IpcChannelMode mode) noexcept;
 
     /// @brief create an IpcChannelError from the provides error code
     /// @return IpcChannelError if error occured
@@ -143,7 +143,7 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     /// @brief Tries to close the given file descriptor
     /// @param[in] file descriptor that shall be closed
     /// @return IpcChannelError if error occured
-    cxx::expected<IpcChannelError> closeFileDescriptor(const int32_t fileDescriptor) noexcept;
+    cxx::expected<IpcChannelError> closeFileDescriptor() noexcept;
 
   private:
     UdsName_t m_name;
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
@@ -141,8 +141,9 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     static bool isNameValid(const UdsName_t& name) noexcept;
 
     /// @brief Tries to close the given file descriptor
+    /// @param[in] file descriptor that shall be closed
     /// @return IpcChannelError if error occured
-    cxx::expected<IpcChannelError> closeFd(int32_t fileDescriptor);
+    cxx::expected<IpcChannelError> closeFileDescriptor(const int32_t fileDescriptor) noexcept;
 
   private:
     UdsName_t m_name;
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
@@ -46,7 +46,7 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     static constexpr int32_t ERROR_CODE = -1;
     static constexpr int32_t INVALID_FD = -1;
 
-    using UdsName_t = cxx::string<LONGEST_VALID_NAME>; 
+    using UdsName_t = cxx::string<LONGEST_VALID_NAME>;
 
     /// @brief for calling private constructor in create method
     friend class DesignPattern::Creation<UnixDomainSocket, IpcChannelError>;
@@ -72,8 +72,7 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     /// @param NoPathPrefix signalling that this method does not add a path prefix
     /// @param name of the unix domain socket to unlink
     /// @return true if the unix domain socket could be unlinked, false otherwise, IpcChannelError if error occured
-    static cxx::expected<bool, IpcChannelError> unlinkIfExists(const NoPathPrefix_t,
-                                                               const UdsName_t& name) noexcept;
+    static cxx::expected<bool, IpcChannelError> unlinkIfExists(const NoPathPrefix_t, const UdsName_t& name) noexcept;
 
     /// @brief close the unix domain socket.
     cxx::expected<IpcChannelError> destroy() noexcept;
@@ -140,6 +139,10 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     cxx::error<IpcChannelError> createErrorFromErrnum(const int32_t errnum) const noexcept;
 
     static bool isNameValid(const UdsName_t& name) noexcept;
+
+    /// @brief Tries to close the given filedescriptor
+    /// @return SmartC object
+    iox::cxx::SmartC<int(int), int, int> closeFd(int32_t fileDescriptor);
 
   private:
     UdsName_t m_name;
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
@@ -129,9 +129,9 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
                      const uint64_t maxMsgNumber = 10U) noexcept;
 
 
-    /// @brief creates the unix domain socket
+    /// @brief initializes the unix domain socket
     /// @param mode blocking or non_blocking
-    /// @return int with the socket file descriptor, IpcChannelError if error occured
+    /// @return IpcChannelError if error occured
     cxx::expected<IpcChannelError> initalizeSocket(const IpcChannelMode mode) noexcept;
 
     /// @brief create an IpcChannelError from the provides error code
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
@@ -140,8 +140,7 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
 
     static bool isNameValid(const UdsName_t& name) noexcept;
 
-    /// @brief Tries to close the given file descriptor
-    /// @param[in] file descriptor that shall be closed
+    /// @brief Tries to close the file descriptor
     /// @return IpcChannelError if error occured
     cxx::expected<IpcChannelError> closeFileDescriptor() noexcept;
 
--- a/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
@@ -140,9 +140,9 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
 
     static bool isNameValid(const UdsName_t& name) noexcept;
 
-    /// @brief Tries to close the given filedescriptor
-    /// @return SmartC object
-    iox::cxx::SmartC<int(int), int, int> closeFd(int32_t fileDescriptor);
+    /// @brief Tries to close the given file descriptor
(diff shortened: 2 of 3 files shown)


iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -114,6 +114,11 @@ UnixDomainSocket& UnixDomainSocket::operator=(UnixDomainSocket&& other) noexcept
 {
     if (this != &other)
     {
+        if (destroy().has_error())
```
mossmaurice, line 117: Writing tests for the move c'tor/move assignment operator is currently not possible due to the missing possibilty to check for an invalidate `expected`.

We need an `is_valid` method:
```
TYPED_TEST(IpcChannel_test, MoveAssignmentLeadsToInvalidation)
{
    auto clientResult =
        TestFixture::IpcChannelType::create(anotherGoodName, IpcChannelMode::BLOCKING, IpcChannelSide::SERVER);
    auto movedValue = std::move(clientResult);

    EXPECT_TRUE(clientResult.is_valid());
    EXPECT_FALSE(movedValue.is_valid());
}
```

mossmaurice: @elBoberido suggested either `has_value`, `is_moved`, `is_specified`, `is_valid`, `has_content`, `not_invalidated`. I like `is_valid ` the most.

What do you think @elfenpiff @budrus?

iceoryx_posh/source/runtime/message_queue_interface.cpp
```diff
@@ -158,9 +158,20 @@ bool MqBase::openMessageQueue(const posix::IpcChannelSide channelSide) noexcept
...
+    {
+        IpcChannelType::create(
+            m_interfaceName, posix::IpcChannelMode::BLOCKING, m_channelSide, m_maxMessageSize, m_maxMessages)
+            .and_then([this, &receivedError](auto& mq) {
+                this->m_mq = std::move(mq);
+                receivedError = posix::IpcChannelError::UNDEFINED;
+            })
+            .or_else([&receivedError](posix::IpcChannelError& error) { receivedError = error; });
+
+    } while (receivedError == posix::IpcChannelError::NO_SUCH_CHANNEL);
```
elBoberido, line 174: If you start more applications than cores are available before roudi is running, the whole system becomes unresponsive. Why not wait a few milliseconds? Why not use the previous behavior?

mossmaurice: Two valid points. While loops without a sleep burns CPU time and the loop is not needed as the state machine already handles it.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -370,29 +383,48 @@ cxx::expected<int32_t, IpcChannelError> UnixDomainSocket::createSocket(const Ipc
...
-            return cxx::success<int32_t>(sockfd);
+            auto closeCall = closeFd(sockfd);
+            if (closeCall.has_error())
+            {
+                return cxx::error<IpcChannelError>(closeCall.get_error());
+            }
+            return createErrorFromErrnum(connectCall.getErrNum());
         }
-        else
+        else if (connectCall.getErrNum() == ENOENT)
```
mossmaurice, line 415: This is an error/sad path but `hasErrors()` didn't return true, when taking this branch because `ENOENT` shall not be blamed as a error inside `SmartC`. This could be a good first issue, I've added it to #418

elBoberido: maybe we need something like silent errors in SmartC for stuff like this

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -370,29 +383,48 @@ cxx::expected<int32_t, IpcChannelError> UnixDomainSocket::createSocket(const Ipc
...
         {
+            auto closeCall = closeFd(sockfd);
```
elfenpiff, line 417: We have this code snippet at least 3 times in the code here:
```cpp
            auto closeCall = closeFd(sockfd);
            if (closeCall.has_error())
            {
                return cxx::error<IpcChannelError>(closeCall.get_error());
            }
```

I would suggest a lambda like
```cpp
auto callClose = [this](){
            auto closeCall = closeFd(sockfd);
            if (closeCall.has_error())
            {
                return cxx::error<IpcChannelError>(closeCall.get_error());
            }
};

///....

return callClose();
```

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -370,29 +383,48 @@ cxx::expected<int32_t, IpcChannelError> UnixDomainSocket::createSocket(const Ipc
...
-        if (!connectCall.hasErrors())
+        if (connectCall.hasErrors())
         {
-            return cxx::success<int32_t>(sockfd);
+            auto closeCall = closeFd(sockfd);
+            if (closeCall.has_error())
+            {
+                return cxx::error<IpcChannelError>(closeCall.get_error());
+            }
+            return createErrorFromErrnum(connectCall.getErrNum());
```
elfenpiff, line 413: This is weird. Let's say you have an error in your `connectCall` and caused by this you experience also an error maybe in the `close` call. Then we return only the closeCall error and not the connectCall error which led to this trouble. Here in this particular case I would recommend ignoring the error value of the closeCall since it is only a symptom and not the source and always return the `createErrorFromErrnum(connectCall.getErrNum());`

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -370,29 +383,48 @@ cxx::expected<int32_t, IpcChannelError> UnixDomainSocket::createSocket(const Ipc
...
         }
-        else
+        else if (connectCall.getErrNum() == ENOENT)
         {
+            auto closeCall = closeFd(sockfd);
+            if (closeCall.has_error())
+            {
+                return cxx::error<IpcChannelError>(closeCall.get_error());
```
elfenpiff, line 420: Please ignore the closeCall error and return the cause which led to this point: 
`return createErrorFromErrnum(connectCall.getErrNum());`

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -412,112 +444,90 @@ cxx::error<IpcChannelError> UnixDomainSocket::createErrorFromErrnum(const int32_
     {
     case EACCES:
     {
-        std::cerr << "permission to create unix domain socket denied \"" << m_name << "\"" << std::endl;
```
elfenpiff, line 415: Please leave this error messages in here since they are maybe becoming valuable later.
SmartC will return only a generic error message which explains for instance `EACCES` but does not provide context for the user - in this case that it was used for socket creation.

SmartC always provides you with a rough idea and is always good to provide an additional error message to give the user a more detailed context.

elfenpiff (changes requested):
1. closeFd without arguments
2. do not return closeFd errors when it hides the overlapping cause
3. Undelete the detailed error messages.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -141,6 +140,10 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
 
     static bool isNameValid(const UdsName_t& name) noexcept;
 
+    /// @brief Tries to close the given file descriptor
+    /// @return IpcChannelError if error occured
+    cxx::expected<IpcChannelError> closeFd(int32_t fileDescriptor);
```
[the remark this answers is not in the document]
mossmaurice: I'm disappointed, you missed the missing `noexcept` :P

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: Well, I came to find that quite noisy for the user. The user is notified via `expected` about `IpcChannelError` and hence the reason why the call failed and can decide to print an info on a higher level. I could agree to just remove the print for the `ENOENT`. This is what the whole pull request is about :monkey:

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: Yup, I agree, as close to the standard as possible:

Hits on CppRef:
* `has_value` 3
* `is_moved` 0
* `is_specified` 0
* `is_valid` 0
* `has_content` 0
* `not_invalidated` 0
* `has_undefined_state` 0

I'm quite fond of `is_undefined`

We can move this discussion to the follow-up pull request.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -155,29 +163,34 @@ cxx::expected<bool, IpcChannelError> UnixDomainSocket::unlinkIfExists(const NoPa
...
+    {
+        return createErrorFromErrnum(closeCall.getErrNum());
+    }
+}
+
+cxx::expected<IpcChannelError> UnixDomainSocket::destroy() noexcept
+{
+    if (m_isInitialized)
+    {
+        return closeFd(m_sockfd);
```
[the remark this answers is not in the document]
mossmaurice: Nope, inside `UnixDomainSocket::createSocket()`,  `closeFd()` is called with a different value. Using a default parameter instead as an alternative, is less readable for me, because I have to be aware of a default param.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: if the `expected` in in an undefined state after moving, we are violating the C++ standard which explicitly tells that the object must be in an unspecified but defined state ;)

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: You are right. Adapted.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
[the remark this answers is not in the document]
mossmaurice: Changed return value of `createSocket()` and rename it to `initalizeSocket()`

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: Move to #542

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -363,7 +361,7 @@ cxx::expected<int32_t, IpcChannelError> UnixDomainSocket::createSocket(const Ipc
         return createErrorFromErrnum(socketCall.getErrNum());
     }
 
-    int32_t sockfd = socketCall.getReturnValue();
+    m_sockfd = socketCall.getReturnValue();
```
elBoberido, line 364: this is a source for future bugs. We might end up with a dangling file descriptor which is not set to `INVALID_FD`.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -155,29 +161,34 @@ cxx::expected<bool, IpcChannelError> UnixDomainSocket::unlinkIfExists(const NoPa
...
 }
 
-cxx::expected<IpcChannelError> UnixDomainSocket::destroy() noexcept
+cxx::expected<IpcChannelError> UnixDomainSocket::closeFileDescriptor() noexcept
 {
-    if (m_sockfd != INVALID_FD)
-    {
-        auto closeCall = cxx::makeSmartC(
-            closePlatformFileHandle, cxx::ReturnMode::PRE_DEFINED_ERROR_CODE, {ERROR_CODE}, {}, m_sockfd);
+    auto closeCall =
```
elBoberido, line 166: what if `m_sockfd == INVALID_FD`?

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elfenpiff: @elBoberido I completely understand you but if you store it in a temporary variable and then use this variable to close the file descriptor again via `closeFd(tempSockFd)` we have there a bigger source for future bugs since `closeFd` needs to invalidate `m_sockfd` and has to set the whole object to `m_isInitialized = false`.

Then you would have to verify in `closeFd` if it is an arbitrary file descriptor or the socket file descriptor and you have to invalidate everything! 

In my opinion this is the cleanest approach and with this abstraction and sufficient tests we should be good to go.

Additionally, in the previous approach you must always assign the sockfd manually after a successful call and with the logic of a source for future bug you could argue that a future developer maybe forgets to assign this and then you would end up with an invalidated file descriptor.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: Fixed

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: I didn't say the previous approach was better, just that this might be improved if thinking a bit more about it

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -133,14 +132,18 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     /// @brief creates the unix domain socket
```
elfenpiff, line 132: Now it initializes the unix domain socket.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -133,14 +132,18 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
...
     /// @param mode blocking or non_blocking
     /// @return int with the socket file descriptor, IpcChannelError if error occured
```
elfenpiff, line 134: The return value description is out of sync.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -81,18 +81,16 @@ UnixDomainSocket::UnixDomainSocket(const NoPathPrefix_t,
     else
     {
         m_maxMessageSize = maxMsgSize;
-        auto createResult = createSocket(mode);
+        auto initalizeResult = initalizeSocket(mode);
```
elfenpiff, line 84: The code is correct but this would be a typical use case for the `and_then`/`or_else` pattern to reduce the amount of code and future errors.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -114,6 +112,11 @@ UnixDomainSocket& UnixDomainSocket::operator=(UnixDomainSocket&& other) noexcept
 {
     if (this != &other)
```
elfenpiff, line 113: We all implemented the move assignment in the wrong fashion. Please read: https://isocpp.org/wiki/faq/assignment-operators

and we need to add here the move assignment for the creation pattern. Additionally, we have to implement it as well (which is not yet being done but I work on it in a different PR). Could you please add here only the explicit call to the move assignment of the creation:
```cpp
DesignPattern::Creation<UnixDomainSocket, UnixDomainSocketError>::operator=(std::move(other));
```

elfenpiff (changes requested):

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: Imporant point, thanks for raising it! Fixed.

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: How about the following:
- keep the old name and return the file descriptor
- rename the current `closeFileDescriptor` to `closeAndUnlink`
- create a static `closeFileDescriptor` method which takes a file descriptor as parameter
- call that one in this method and in `closeAndUnlink`

With this, it should not be possible to end up with a dangling file descriptor.

elBoberido (commented):
Except the food for thought for `initializeSocket` this looks good

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
mossmaurice: For me it doesn't make sense to use a creational pattern aka `createSocket()` here. What you describe perfectly fit for something like a factory, where I say "Hey please bake that object for me". In our case it is different. `sockfd` is a member of the object, hence an integral part and nothing that gets created and is then attached to the object. Think of a car that already has tires and which get pumped up aka initalised at a certain point in the lifetime of the car. What do you think fits better `initializeTires()" or `createTires()`? With the current approach it should also not be possible to end up with dangling pointers.

elfenpiff (approved):

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: Well it's more like a `check if the tires are flat before mounting it on the car` analogy :)

elBoberido (approved):

iceoryx_utils/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: oh, and it's possible run twice through `closeFileDescriptor` if the first call fails. That wouldn't be the case if `m_sockfd` is not set in the first place.

PR #553, opened 2021-02-11
Iox #542 resource leak fix

elfenpiff:
## Pre-Review Checklist for the PR Author

1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## Post-review Checklist for the Eclipse Committer

1. [x] All checkboxes in the PR checklist are checked or crossed out
1. [x] Merge

## References

fixes partially #542

PR #554, opened 2021-02-11
Iox #542 Invalidate `rhs` if an `expected` is moved

### LINKED ISSUE #542: Fix resource leak caused by move ctor/assignment

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All checks have passed
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Invalidate `rhs` if an `expected` is moved
* ~~`is_initialized` is [debatable](https://github.com/eclipse-iceoryx/iceoryx/pull/550#discussion_r573629792), chose it to be consistent with other classes, see #553~~
* Added `INVALID_STATE` to all enum classes used as `ErrorTyoe`

1. [x] Merge

## References

- Partly closes #542


iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -82,6 +82,15 @@ TEST_F(expected_test, CreateError)
     EXPECT_THAT(sut.get_error().m_b, Eq(212));
 }
 
+TEST_F(expected_test, CreateValueAndMove)
```
elfenpiff, line 85: Please name the test like: What are you doing at what are you expecting from the public api or the behavior.
In this case `MoveAssignmentInvalidatesSource`

Additionally, you are missing some tests:
1. expected with value and error
    1. move ctor of expected with value invalidates source
    1. move ctor of expected with error invalidates source
    1. move assignment of expected with value invalidates source
    1. move assignment of expected with error invalidates source
1. expected with only error
    1. move ctor of expected with error invalidates source
    1. move assignment of expected with error invalidates source

elfenpiff (changes requested):
Add further tests:
1. expected with value and error
    1. move ctor of expected with value invalidates source
    1. move ctor of expected with error invalidates source
    1. move assignment of expected with value invalidates source
    1. move assignment of expected with error invalidates source
1. expected with only error
    1. move ctor of expected with error invalidates source
    1. move assignment of expected with error invalidates source

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -431,6 +433,10 @@ class expected<ValueType, ErrorType>
     template <typename... Targs>
     static expected create_error(Targs&&... args) noexcept;
 
+    /// @brief returns true if the expected is in a valid state either containing a value or an error
+    /// @return true if in valid state or false if in invalid state (e.g. after move)
+    bool is_initialized() const noexcept;
```
MatthiasKillat, line 438: I dislike this as now one could get the impression that we always need to check this before accessing the expected. I know other types also do it like this but there it is also bad. 
The convention is that it is the responsibility of the caller not to access any object that was moved from.
It can be silently invalidated (I also am not a fan of this, for performance and logic reasons) but it is important to not have a mandatory is_initialized check. 
I am fine with it in the public API, it can be used...but does not have to be.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -792,10 +798,9 @@ class expected<ValueType, ErrorType>
 
   private:
     expected(variant<ValueType, ErrorType>&& f_store, const bool hasError) noexcept;
-
-  private:
     variant<ValueType, ErrorType> m_store;
     bool m_hasError;
+    bool m_isInitialized{true};
```
MatthiasKillat, line 803: This is ok if it is always ensured that any ctor constructs a valid object or sets it to false otherwise. Otherwise we need to default it to false and set to true in ctor.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -92,6 +92,26 @@ inline expected<ValueType, ErrorType>::expected(error<ErrorType>&& errorValue) n
 {
 }
 
+template <typename ValueType, typename ErrorType>
+inline expected<ValueType, ErrorType>::expected(expected<ValueType, ErrorType>&& rhs) noexcept
+{
+    *this = std::move(rhs);
```
MatthiasKillat, line 98: I always see this and it is suboptimal in general: the move (copy) ctor is implemented in terms of the (copy) move in terms of the corresponding operator=. Looks fine on first sight from code reuse point of view etc. The problem is that this way in many cases a potentially costly default initialization happens before the assignment.

From my point of view it is often better to do initialize directly with the exception of classes with primitive type members only (not entirely the case here, due to m_store). 

Those inefficiencies add up in the long run.

Maybe consider this in future implementations.

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -175,6 +184,15 @@ TEST_F(expected_test, VoidCreateError)
     ASSERT_THAT(sut.get_error(), Eq(12.2f));
 }
 
+TEST_F(expected_test, VoidCreateErrorAndMove)
+{
+    auto sut = expected<float>::create_error(42.0f);
+    auto movedValue = std::move(sut);
+    EXPECT_FALSE(sut.is_initialized());
+    EXPECT_TRUE(movedValue.is_initialized());
```
MatthiasKillat, line 192: Needs to be an assert otherwise we risk an invalid access.

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -82,6 +82,15 @@ TEST_F(expected_test, CreateError)
     EXPECT_THAT(sut.get_error().m_b, Eq(212));
 }
 
+TEST_F(expected_test, CreateValueAndMove)
+{
+    auto sut = expected<int, Test>::create_value(73);
+    auto movedValue = std::move(sut);
+    EXPECT_FALSE(sut.is_initialized());
+    EXPECT_TRUE(movedValue.is_initialized());
```
MatthiasKillat, line 90: Needs to be an assert otherwise we risk an invalid access.

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -175,6 +184,15 @@ TEST_F(expected_test, VoidCreateError)
     ASSERT_THAT(sut.get_error(), Eq(12.2f));
 }
 
+TEST_F(expected_test, VoidCreateErrorAndMove)
```
MatthiasKillat, line 187: I do not understand the Void in the test name. Clarify or change the name.

MatthiasKillat (changes requested):

elfenpiff:
@mossmaurice we have a little problem with `and_then` and `or_else`.
Take a look at the code:
```cpp
cxx::expected<int, float> sut1(1), sut2(2.1f);
sut1 = std::move(sut2);

sut2
    .and_then([](auto p){std::cout << "value: " << p;})
    .or_else([](auto p){std::cout << "error: " << p;});
```

What should `sut2` call now since it is not initialized? The answer would be: nothing! Since either way it would otherwise lead to undefined behavior. Therefore, could you please add and test this behavior.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
elfenpiff: I agree with @MatthiasKillat .

And the more I think about the name `is_initialized` the more I dislike it. Since the user does not see right away when this method returns false they are going straight into the land of undefined behavior. When we perform a code review it has to shine so that we pay extra attention. What about the name `has_undefined_state` or `is_in_unsafe_undefined_state`?

Is there a way to hide this and make the expected still working without checking this?! I think this is possible by:
1. Adjusting the behavior of `operator bool()` and return true only if it is not in an unsafe undefined state and has a value
2. Add `bool has_value()` and `bool has_error()`. Then the user has to check the correct method when they would like to access the value/error - it is error prone! But if you would like to use it in the recommended safe way we have `and_then`/`or_else`.
3. `and_then` and `or_else` should never be called when the object is in an unsafe undefined state.

elfenpiff (changes requested):
1. Rename `is_initialized` into a more warning and dangerous name.
2. Try to adjust the behavior of the expected so that the user does not need to check this method. (See comment).
3. Adjust `and_then` / `or_else` (see comment)

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
mossmaurice: Well, well, I remember a certain young @elfenpiff writing these tests some years ago :stuck_out_tongue: Reworking existing ones and will work on the ones you mentioned.

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
mossmaurice: Fixed

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
mossmaurice: Fixed

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
mossmaurice: I started reworking the existing test (see ed2db2)

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
elfenpiff: Yeah yeah ... I was young, naive and didn't knew what I was doing. But then Hypnotoad came and showed me the road to enlightenment.
thatsnoneofmy

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
MatthiasKillat: At least two possbilities: the methods and_then, or_else etc. can either to do nothing when the object is not initialized (which leads to an internal check) or just have undefined behavior in this case (likely cashing). Since the second case can only occur for sources of move, it would be within how C++ works with moved objects, so no real surprises for the user.

has_value should be added regardless, to make this usage pattern more symmetric to optional.

Since we have ``operator bool``, is_initialized could be private (or removed) and the check added to ``operator bool``.

Note: I am completely fine with undefined behavior as long as it is in line with established C++ rules (i.e. cases where it occurs) and there is good reason for it (usually performance).

MatthiasKillat:
I do not think it is strictly necessary to make all m(void) ethods essentially do nothing silently for unintialized objects since this can only occur in move, after which no methods are supposed to be called on the object anyway except for potentially re-assignment (ensure that ``operator=`` can deal with this) and (implicitly) the destructor.

elBoberido:
> Notice that `has_value` and `has_error` both return false in the uninitialized case. The user must be aware that he cannot check for one and if the check is negative assume the getter for the other can be called, but this is fine.

This might be problematic since we often use `if (!foo.has_error) { foo.value() ...}`

elfenpiff:
> > Notice that `has_value` and `has_error` both return false in the uninitialized case. The user must be aware that he cannot check for one and if the check is negative assume the getter for the other can be called, but this is fine.
> 
> This might be problematic since we often use `if (!foo.has_error) { foo.value() ...}`

Another argument to always use the `and_then`/`or_else` pattern if possible. But in the end you are right, this is dangerous!

mossmaurice:
Today is another day, where I'm on the warpath with C++. Destructive moves [simplify things immensely](https://radekvit.medium.com/move-semantics-in-c-and-rust-the-case-for-destructive-moves-d816891c354b) (Kudos to @elBoberido for sharing that post). Well, regarding the discussion above. Using the `operator bool()` to retrieve a tri-state is a bad idea IMHO. Users could not differentiate between and error and an invalid state. Hence, I decided to make `has_undefined_state` public.

For the move behaviour three possible scenarios came to my mind:

1. Forbid move semantic for `expected`
2. In case the user calls e.g. `get_error()` or `value()` on an invalid object, return a default constructed object e.g. an empty string
3. In case the user calls e.g. `get_error()` or `value()` on an invalid object, call `std::terminate`. Fail early and make sure there is defined behaviour. This is used by other iceoryx utils patterns e.g. `vector`

No. 3 is far from perfect. It would be better to continue the program execution and let the user decide what to do with this specific error. However, it is not trivial to imitate e.g. a `cxx::nullopt` with `cxx::expected`, so for now I decided to go along with no. 3 Safety systems in the automotive industry usually get validated with endurance runs which run for a long time and many thousand kilometers on the street. If no `std::terminate` occurs during an endurance run, users and stakeholders can then justify statistically for a safe system.

elBoberido:
> Today is another day, where I'm on the warpath with C++. Destructive moves [simplify things immensely](https://radekvit.medium.com/move-semantics-in-c-and-rust-the-case-for-destructive-moves-d816891c354b) (Kudos to @elBoberido for sharing that post). Well, regarding the discussion above. Using the `operator bool()` to retrieve a tri-state is a bad idea IMHO. Users could not differentiate between and error and an invalid state. Hence, I decided to make `has_undefined_state` public.
> 
> For the move behaviour three possible scenarios came to my mind:
> 
>     1. Forbid move semantic for `expected`
> 
>     2. In case the user calls e.g. `get_error()` or `value()` on an invalid object, return a default constructed object e.g. an empty string
> 
>     3. In case the user calls e.g. `get_error()` or `value()` on an invalid object, call `std::terminate`. Fail early and make sure there is defined behaviour. This is used by other iceoryx utils patterns e.g. `vector`
> 
> 
> No. 3 is far from perfect. It would be better to continue the program execution and let the user decide what to do with this specific error. However, it is not trivial to imitate e.g. a `cxx::nullopt` with `cxx::expected`, so for now I decided to go along with no. 3 Safety systems in the automotive industry usually get validated with endurance runs which run for a long time and many thousand kilometers on the street. If no `std::terminate` occurs during an endurance run, users and stakeholders can then justify statistically for a safe system.

Endurance runs are not necessary. Accessing a moved value should be detected by MC/DC tests and should therefore considered to be not present. Please use `cxx::Expects` instead of just `std::terminate()`.

I'm also not quite sure if `has_undefined_state` has the right name or should be public
- after the move it must not be in an undefined state, else we violate the C++ standard. It must be defined but unspecified. This is an important distinction
- should the user really have to care about this? Using a moved object is not allowed, except by bringing it back into an defined state. This leads to extend a bad design on the user side even further

mossmaurice:
> Endurance runs are not necessary. Accessing a moved value should be detected by MC/DC tests and should therefore considered to be not present. 

You are right, MC/DC tests are very helpful to detected problems before going on the road. Additionally, they are a necessity when it comes to ISO26262. However, those tests are fairly limited and always rely on a certain assumption in a certain environment. With automated driving and robotics one has to deal with the open world problem. It is nearly impossible to predict all special situations that can occur in advance, e.g. there could be more objects in the field of view than anticipated and the WCET will be violated when analyzing them, which could lead to other nodes/runnables being in an inconsistent state. This is why the likes of Waymo, et al. are driving [billions of kilometers](https://en.wikipedia.org/wiki/Waymo#Road_testing). I've never seen a safety system going on the street without an extensive endurance run.

> Please use cxx::Expects instead of just std::terminate()

Good point, thanks! Subsequently, you are saying all `std::terminate` occurrences shall be replaced by  `cxx::Expects`? It calls `std::terminate()` in our current implementation. I'll add this point to #261 then.

> I'm also not quite sure if has_undefined_state has the right name or should be public

1. I'm happy to change the name, what about `has_unspecified_state`? @MatthiasKillat @elfenpiff 
2. It should definitely be public IMHO. I suppose this is what the STL would do. Giving the user the possibility to check all possible states of the object, hiding the tri-state e.g. behind a `operator bool()` is for me like lying to the user.

elBoberido:
> > Endurance runs are not necessary. Accessing a moved value should be detected by MC/DC tests and should therefore considered to be not present.
> 
> You are right, MC/DC tests are very helpful to detected problems before going on the road. Additionally, they are a necessity when it comes to ISO26262. However, those tests are fairly limited and always rely on a certain assumption in a certain environment. With automated driving and robotics one has to deal with the open world problem. It is nearly impossible to predict all special situations that can occur in advance, e.g. there could be more objects in the field of view than anticipated and the WCET will be violated when analyzing them, which could lead to other nodes/runnables being in an inconsistent state. This is why the likes of Waymo, et al. are driving [billions of kilometers](https://en.wikipedia.org/wiki/Waymo#Road_testing). I've never seen a safety system going on the street without an extensive endurance run.

But those endurance runs have a different purpose. I talked specifically about accessing a moved object. This must be detected by MC/DC. What you are talking about is which branch of the code will be executed under which condition. That's the open world problem. The logically wrong branch could be taken. But that doesn't mean that we didn't test that branch and already ensured there is no moved object. It's like using the GPS to drive from SF to LA and arrive in Vegas. Wrong branching. Whereas rushing down the Grand Canyon is going off road and should have been covered by MC/DC.

> 
> > Please use cxx::Expects instead of just std::terminate()
> 
> Good point, thanks! Subsequently, you are saying all `std::terminate` occurrences shall be replaced by `cxx::Expects`? It calls `std::terminate()` in our current implementation. I'll add this point to #261 then.

It's an open point in the error handling document. It states it should be evaluated if all `std::terminate` should be replaced with assert like constructs like `cxx::Expects`, `cxx::Ensures`, etc.

> > I'm also not quite sure if has_undefined_state has the right name or should be public
> 
>     1. I'm happy to change the name, what about `has_unspecified_state`? @MatthiasKillat @elfenpiff
> 
>     2. It should definitely be public IMHO. I suppose this is what the STL would do. Giving the user the possibility to check all possible states of the object, hiding the tri-state e.g. behind a `operator bool()` is for me like lying to the user.

I question the need for such a method and also wouldn't do it with `operator bool()`. Let's look how the STL handles this with the `std::vector` ... well, don't use a moved vector. Instead of introducing an explicit tri-state just define the access of a moved object invalid which will terminate the application. Even with the `has_undefined_state` method it is necessary to test the code because developers forget to use such methods. Without this method it is possible to reach full MC/DC coverage. With this method you either cannot reach full coverage because you have a correct program and therefore dead code or you reach full coverage which means you have an incorrect program which should be fixed.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -355,15 +291,13 @@ class expected<ErrorType>
...
-    [[deprecated]] expected& on_success(const cxx::function_ref<void()>& callable) noexcept;
     expected& and_then(const cxx::function_ref<void()>& callable) noexcept;
 
   private:
     expected(variant<ErrorType>&& store, const bool hasError) noexcept;
-
-  private:
     variant<ErrorType> m_store;
     bool m_hasError;
+    bool m_isInitialized{true};
```
elfenpiff, line 300: This member should be renamed to `m_hasUndefinedState` to be consistent with the `has_undefined_state` method and in the code the developer would see immediately what is going on in there.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -127,96 +153,119 @@ inline bool expected<ValueType, ErrorType>::has_error() const noexcept
...
 inline const ErrorType&& expected<ValueType, ErrorType>::get_error() const&& noexcept
 {
-    return std::move(*m_store.template get_at_index<1>());
+    if (!has_undefined_state())
+    {
+        return std::move(*m_store.template get_at_index<1>());
+    }
+    else
+    {
+        std::cerr << "get_error() called on invalid object" << std::endl;
```
elfenpiff, line 162: I would state it more precisely: `get_error() called on an expected in an undefined state`.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -127,96 +153,119 @@ inline bool expected<ValueType, ErrorType>::has_error() const noexcept
...
+        std::terminate();
+    }
 }
 
 template <typename ValueType, typename ErrorType>
     inline ErrorType& expected<ValueType, ErrorType>::get_error() & noexcept
 {
-    return *m_store.template get_at_index<1>();
+    if (!has_undefined_state())
```
elfenpiff, line 184: Code duplication! Please use here `return std::move(*this).get_error()`. In this case move is used to transform this from an lvalue into a rvalue and nothing is moved behind the scenes!

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -127,96 +153,119 @@ inline bool expected<ValueType, ErrorType>::has_error() const noexcept
...
+    {
+        std::cerr << "get_error() called on invalid object" << std::endl;
+        std::terminate();
+    }
 }
 
 template <typename ValueType, typename ErrorType>
     inline ErrorType&& expected<ValueType, ErrorType>::get_error() && noexcept
 {
-    return std::move(*m_store.template get_at_index<1>());
```
elfenpiff, line 136: Code duplication! Please write `return const_cast<ErrorType&&>(const_cast<decltype(*this) *>(this).get_error());`

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -127,96 +153,119 @@ inline bool expected<ValueType, ErrorType>::has_error() const noexcept
...
+        std::terminate();
+    }
 }
 
 template <typename ValueType, typename ErrorType>
 inline const ValueType& expected<ValueType, ErrorType>::value() const& noexcept
 {
-    return *m_store.template get_at_index<0>();
-}
+    if (!has_undefined_state())
```
elfenpiff, line 226: Code duplication with `value()`

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -127,96 +153,119 @@ inline bool expected<ValueType, ErrorType>::has_error() const noexcept
...
+        std::terminate();
+    }
+} // namespace cxx
 
 template <typename ValueType, typename ErrorType>
 inline const ValueType&& expected<ValueType, ErrorType>::value() const&& noexcept
 {
-    return std::move(*m_store.template get_at_index<0>());
-}
+    if (!has_undefined_state())
```
elfenpiff, line 240: Code duplication with `value()`.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -127,96 +153,119 @@ inline bool expected<ValueType, ErrorType>::has_error() const noexcept
...
+        std::cerr << "value() called on invalid object" << std::endl;
+        std::terminate();
+    }
+} // namespace iox
 
 template <typename ValueType, typename ErrorType>
     inline ValueType& expected<ValueType, ErrorType>::value() & noexcept
 {
-    return *m_store.template get_at_index<0>();
+    if (!has_undefined_state())
```
elfenpiff, line 254: Code duplication with `value()`.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -602,67 +581,63 @@ inline bool expected<ErrorType>::has_error() const noexcept
...
+        std::cerr << "get_error() called on invalid object" << std::endl;
+        std::terminate();
+    }
 }
 
 template <typename ErrorType>
 inline const ErrorType& expected<ErrorType>::get_error() const& noexcept
 {
-    return *m_store.template get_at_index<0>();
+    if (!has_undefined_state())
```
elfenpiff, line 598: Code duplication!

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -602,67 +581,63 @@ inline bool expected<ErrorType>::has_error() const noexcept
...
+        std::cerr << "get_error() called on invalid object" << std::endl;
+        std::terminate();
+    }
 }
 
 template <typename ErrorType>
 inline const ErrorType&& expected<ErrorType>::get_error() const&& noexcept
 {
-    return std::move(*m_store.template get_at_index<0>());
+    if (!has_undefined_state())
```
elfenpiff, line 612: Code duplication!

elfenpiff (changes requested):
Remove code duplication

mossmaurice:
> For the move behaviour three possible scenarios came to my mind:
> 
>     1. Forbid move semantic for `expected`
> 
>     2. In case the user calls e.g. `get_error()` or `value()` on an invalid object, return a default constructed object e.g. an empty string
> 
>     3. In case the user calls e.g. `get_error()` or `value()` on an invalid object, call `std::terminate`. Fail early and make sure there is defined behaviour. This is used by other iceoryx utils patterns e.g. `vector`

After some brainstorming sessions with @elBoberido and @elfenpiff, @elfenpiff came up with the idea to take path 1) and make the move c'tor and move assignment operator protected and then implement a method `expected & manualMove(expected && rhs, ErrorType errorValue)`. This forces the user to always provide a valid `ErrorType` which is then stored in the object which was moved from. IMHO this is the cleanest way. The unspecified state of the normal move is avoided and both objects are in a defined state.

elBoberido:
> > For the move behaviour three possible scenarios came to my mind:
> > ```
> > 1. Forbid move semantic for `expected`
> > 
> > 2. In case the user calls e.g. `get_error()` or `value()` on an invalid object, return a default constructed object e.g. an empty string
> > 
> > 3. In case the user calls e.g. `get_error()` or `value()` on an invalid object, call `std::terminate`. Fail early and make sure there is defined behaviour. This is used by other iceoryx utils patterns e.g. `vector`
> > ```
> 
> After some brainstorming sessions with @elBoberido and @elfenpiff, @elfenpiff came up with the idea to take path 1) and make the move c'tor and move assignment operator protected and then implement a method `expected & manualMove(expected && rhs, ErrorType errorValue)`. This forces the user to always provide a valid `ErrorType` which is then stored in the object which was moved from. IMHO this is the cleanest way. The unspecified state of the normal move is avoided and both objects are in a defined state.

After some thinking about this idea I think I'm again the party pooper. How about this scenario?
```
auto v = std::move(rhs.value());
```

elfenpiff:
> > > For the move behaviour three possible scenarios came to my mind:
> > > ```
> > > 1. Forbid move semantic for `expected`
> > > 
> > > 2. In case the user calls e.g. `get_error()` or `value()` on an invalid object, return a default constructed object e.g. an empty string
> > > 
> > > 3. In case the user calls e.g. `get_error()` or `value()` on an invalid object, call `std::terminate`. Fail early and make sure there is defined behaviour. This is used by other iceoryx utils patterns e.g. `vector`
> > > ```
> > 
> > 
> > After some brainstorming sessions with @elBoberido and @elfenpiff, @elfenpiff came up with the idea to take path 1) and make the move c'tor and move assignment operator protected and then implement a method `expected & manualMove(expected && rhs, ErrorType errorValue)`. This forces the user to always provide a valid `ErrorType` which is then stored in the object which was moved from. IMHO this is the cleanest way. The unspecified state of the normal move is avoided and both objects are in a defined state.
> 
> After some thinking about this idea I think I'm again the party pooper. How about this scenario?
> 
> ```
> auto v = std::move(rhs.value());
> ```

Hey @elBoberido aka. Mr party pooper ;). Here I would say that this is not the responsibility of the expected anymore. If we would like to solve this issue the value has to take care of the undefined state handling. For this we could pack the value into an optional and solve the matter.

elBoberido:
> > auto v = std::move(rhs.value());
> > ```
> 
> Hey @elBoberido aka. Mr party pooper ;). Here I would say that this is not the responsibility of the expected anymore. If we would like to solve this issue the value has to take care of the undefined state handling. For this we could pack the value into an optional and solve the matter.

I was just thinking about this because it has the same effect to the user as the undefined state of the `expected`. But you are right. We need to enforce that moving something you don't own, like when you only have a reference, is not allowed. For situations like this there should be a `T&& take()` method for the `optional` and a `T&& take(MyError)` method for the `expected`

mossmaurice:
Well, this topic is a brain twister. The path of implementing a `manualMove()` or `safeMove()` can't be taken as the move c'tor needs to be public when [returning](https://en.cppreference.com/w/cpp/language/return#Notes) a local variable inside a function by value. But wait it gets even better! Due to [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision) this move c'tor is then never called.. See this small [Godbolt snippet](https://godbolt.org/z/Tn7veo).

So while discussion this topic, a bug was discovered in the `variant` and fixed with #572. Unlike with the variant, we can't rely on the `ValueType` alone for safe move operation with the `expected`. Latest idea is to require all `ErrorTypes` to have a `static const T INVALID_STATE` e.g.
```
enum class TestError : uint16_t
{
    INVALID_STATE,
    ERROR1,
    ERROR2
};
```

This error can then be used to invalidate the `expected` when moving.

doc/design/error-handling.md
```diff
@@ -139,7 +139,9 @@ The error handler cannot be used in utils.
 
 Whether it is appropriate to use ``cxx::expected`` even if STL compatibility is broken by doing so depends on the circumstances and needs to be decided on a case-by-case basis. If the function has no STL counterpart ``cxx::expected`` can be used freely to communicate potential failure to the caller.
 
-It should be noted that since currently Expects and Ensures are active at release mode, prolific usage of these will incur a runtime cost. Since this is likely to change in the future, it is still advised to use them to document the developer's intentions.
+It should be noted that since currently `Expects` and `Ensures` are active at release mode, prolific usage of these will incur a runtime cost. Since this is likely to change in the future, it is still advised to use them to document the developer's intentions.
+
+Do not use `std::terminate` directly but use assert-like constructs such as ``Expects``, ``Ensures`` or ``assert`` or something else.
```
mossmaurice, line 144: Also added to #261

iceoryx_utils/include/iceoryx_utils/posix_wrapper/timer.hpp
```diff
@@ -37,6 +37,7 @@ namespace posix
 {
 enum class TimerError
 {
```
elfenpiff, line 39: Please update copyright header.

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -32,9 +32,9 @@ class MockCallables
 class expected_test : public Test
 {
   public:
```
elfenpiff, line 34: Please update copyright header.

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -300,38 +520,23 @@ TEST_F(expected_test, AndThenUnpacksOptionalWhenNonEmptyOptionalValue)
...
 {
-    auto sut = expected<iox::cxx::optional<int>, float>::create_error(42.42);
+    auto sut = expected<iox::cxx::optional<int>, TestError>::create_value(iox::cxx::nullopt);
     MockCallables mocks{};
-    EXPECT_CALL(mocks, onEmpty).Times(0);
+    EXPECT_CALL(mocks, onSuccess).Times(0);
 
-    sut.if_empty([&mocks]() { mocks.onEmpty(); });
+    sut.and_then([&mocks](int&) { mocks.onSuccess(); });
 }
```
elfenpiff, line 542: Very nice and clean test. I love them!

elfenpiff (changes requested):

iceoryx_utils/test/stresstests/benchmark_optional_and_expected/benchmark_optional_and_expected.cpp
```diff
@@ -19,8 +19,25 @@
 
 #include "benchmark.hpp"
```
elfenpiff, line 20: Please adjust copyright header.

elfenpiff (changes requested):

iceoryx_posh/test/moduletests/test_mepoo_typed_mempool.cpp
```diff
@@ -58,6 +58,22 @@ class alignas(32) TypedMemPool_test : public Test
     TypedMemPool<TestClass> sut;
 };
 
+namespace iox
+{
+namespace cxx
+{
+template <>
+struct ErrorTypeAdapter<cxx::variant<mepoo::TypedMemPoolError, posix::SemaphoreError>>
```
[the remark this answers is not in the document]
mossmaurice: The last two test cases in this file use it by calling `createObjectWithCreationPattern()`

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -243,99 +269,30 @@ class expected<ErrorType>
...
-    ///         std::cout << "error occured : " << result.get_error() << std::endl;
-    ///     })
-    /// @endcode
-    [[deprecated]] expected& on_error(const cxx::function_ref<void(expected&)>& callable) noexcept;
-
     /// @brief  if the expected does contain an error the given callable is called and
     ///         a reference to the ErrorType is given as an argument to the callable
     /// @param[in] callable callable which will be called if the expected contains an error
     /// @return const reference to the expected itself
+    /// @note if the expected contains an INVALID_STATE the callable is not called
```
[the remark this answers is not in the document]
mossmaurice: Agreed. `INVALID_STATE` is invalid but a well defined error state.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -772,7 +591,7 @@ class expected<ValueType, ErrorType>
     /// @endcode
     ///
     template <typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
-    const expected& if_empty(const cxx::function_ref<void()>& callable) const noexcept;
+    [[deprecated]] const expected& if_empty(const cxx::function_ref<void()>& callable) const noexcept;
```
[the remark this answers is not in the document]
mossmaurice: Let's first deprecate it and then remove it shortly before release v1.0.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -94,6 +96,26 @@ inline expected<ValueType, ErrorType>::expected(error<ErrorType>&& errorValue) n
...
+}
+
+template <typename ValueType, typename ErrorType>
+inline expected<ValueType, ErrorType>&
+expected<ValueType, ErrorType>::operator=(expected<ValueType, ErrorType>&& rhs) noexcept
+{
+    if (this != &rhs)
+    {
+        m_store = std::move(rhs.m_store);
+        m_hasError = rhs.m_hasError;
```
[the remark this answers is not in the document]
mossmaurice: Does this even matter for a `bool`? ;-) Well, ok the compiler shall decide.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -256,49 +225,11 @@ inline const ValueType& expected<ValueType, ErrorType>::operator*() const noexce
...
-{
-    return const_cast<expected*>(this)->on_error(callable);
-}
-
 template <typename ValueType, typename ErrorType>
 inline expected<ValueType, ErrorType>&
 expected<ValueType, ErrorType>::or_else(const cxx::function_ref<void(ErrorType&)>& callable) noexcept
 {
-    if (this->has_error())
+    if (has_error() && (*m_store.template get<ErrorType>() != ErrorTypeAdapter<ErrorType>::getInvalidState()))
```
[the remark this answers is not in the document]
mossmaurice: Removed

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -607,16 +472,10 @@ template <typename ErrorType>
...
-    return *m_store.template get_at_index<0>();
-}
-
-template <typename ErrorType>
-inline const ErrorType&& expected<ErrorType>::get_error() const&& noexcept
+template <typename ValueType, typename ErrorType>
+inline const ErrorType& expected<ValueType, ErrorType>::get_error() const& noexcept
 {
-    return std::move(*m_store.template get_at_index<0>());
+    return *m_store.template get_at_index<1>();
```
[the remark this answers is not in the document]
mossmaurice: This is the `git diff` playing tricks with you brain. 

Green:  const ErrorType& expected<**ValueType, ErrorType**>::get_error()
Red:  const ErrorType&& expected<**ErrorType**>::get_error()

Anyway good idea to define the indexes.

iceoryx_posh/source/roudi/memory/memory_provider.cpp
```diff
@@ -167,6 +167,8 @@ const char* MemoryProvider::getErrorString(const MemoryProviderError error)
 {
     switch (error)
     {
+    case MemoryProviderError::INVALID_STATE:
```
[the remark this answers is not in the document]
mossmaurice: From a test coverage perspective you are right. However, one has to keep the array an the `enum class` in sync and there could be an error here. Still I think the array approach is better. I'll not address it in this PR, though, but in a follow-up.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -26,6 +26,34 @@ namespace iox
 {
 namespace cxx
 {
+/// @brief Type trait which verifies whether the passed type T has INVALID_STATE
+///        Overload chosen when INVALID_STATE is not present
+template <typename T, typename = void>
+struct has_invalid_state : std::false_type
```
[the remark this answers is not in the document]
mossmaurice: Added the `nternal` namespace and moved all type traits there. However, I left them in the hpp as with those constructs declaration and defintion are nearly the same. We also don't have an inl for `iceoryx_utils/include/iceoryx_utils/cxx/type_traits.hpp`

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -26,6 +26,34 @@ namespace iox
...
+/// @brief Generic adapter to access INVALID_STATE member or value
+/// @note Works for enum classes and classes having a member INVALID_STATE
+template <typename T>
+struct ErrorTypeAdapter
+{
+    static_assert(has_invalid_state<T>::value,
+                  "T must have a INVALID_STATE. Alternatively write an ErrorTypeAdapter specialisation for your type");
+
+    static T getInvalidState()
+    {
```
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -16,6 +16,8 @@
 #ifndef IOX_UTILS_CXX_EXPECTED_INL
 #define IOX_UTILS_CXX_EXPECTED_INL
 
+#include "iceoryx_utils/cxx/helplets.hpp"
```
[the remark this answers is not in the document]
mossmaurice: We already agreed to disagree on this one :smiley:  When moving this include to the header it wouldn't be clear anymore that just the definition depends on it. Think of the a refactoring where the `expected` wouldn't be a template anymore and stored in a `cpp`. Then you would have to try out include by include to find out which one belongs to the `cpp`. For me the dependencies are clearly separated if the we differentiate between declaration and defintion.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -504,6 +350,25 @@ inline expected<ErrorType>::expected(const success<void>&) noexcept
...
+}
+
+template <typename ErrorType>
+inline expected<ErrorType>& expected<ErrorType>::operator=(expected<ErrorType>&& rhs) noexcept
+{
+    if (this != &rhs)
+    {
+        m_store = std::move(rhs.m_store);
+        m_hasError = rhs.m_hasError;
+        rhs.m_store = ErrorTypeAdapter<ErrorType>::getInvalidState();
```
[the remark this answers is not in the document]
mossmaurice: This leads to a recursive call of the move assign operator. Will leave it as it is.

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -94,6 +96,26 @@ inline expected<ValueType, ErrorType>::expected(error<ErrorType>&& errorValue) n
...
+
+template <typename ValueType, typename ErrorType>
+inline expected<ValueType, ErrorType>&
+expected<ValueType, ErrorType>::operator=(expected<ValueType, ErrorType>&& rhs) noexcept
+{
+    if (this != &rhs)
+    {
+        m_store = std::move(rhs.m_store);
+        m_hasError = rhs.m_hasError;
+        rhs.m_store.template emplace_at_index<1>(ErrorTypeAdapter<ErrorType>::getInvalidState());
```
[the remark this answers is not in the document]
mossmaurice: This leads to a recursive call of the move assign operator. Will leave it as it is

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -26,6 +26,44 @@ namespace iox
 {
 namespace cxx
 {
+namespace internal
+{
+/// @brief Type trait which verifies whether the passed type T has INVALID_STATE
+///        Overload chosen when INVALID_STATE is not present
+template <typename T, typename = void>
+struct has_invalid_state_member : std::false_type
```
elfenpiff, line 34: I would use in every code which is not strictly stl related our code style. Therefore could you please rename this struct in UpperCamelCase HasInvalidStateMember.
Additionally, could you please create here only a forward declaration of that type like:
```cpp
template<typename, typename>
struct HasInvalidStateMember;
```
and define the types in the inl file. From the point of template programming this would be the implementation which you defined here.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -26,6 +26,44 @@ namespace iox
...
+
+/// @brief Type trait which verifies whether the passed type T has INVALID_STATE
+///        Overload chosen when INVALID_STATE is present
+template <typename T>
+struct has_invalid_state_member<T, std::void_t<decltype(T::INVALID_STATE)>> : std::true_type
+{
+};
+
+template <typename... T>
+struct is_optional : std::false_type
```
elfenpiff, line 46: Could you please rename this into `IsOptional`, declare here only the forward declaration and move the template implementation into the inl file.

elfenpiff (changes requested):

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
mossmaurice: Done

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
mossmaurice: Done

iceoryx_posh/include/iceoryx_posh/internal/mepoo/shared_pointer.hpp
```diff
@@ -26,6 +27,7 @@ namespace mepoo
 {
 enum class SharedPointerError
 {
+    INVALID_STATE,
```
MatthiasKillat, line 30: Naming inconsistent with other enum values. Do we have an agreement whether we want all caps in enum classes? It is not consistent throughout the repository.

Should not be changed here due to the diff but we need to improve in this regard.

mossmaurice: From `CONTRIUTING.md`: "_..If in doubt please try to follow them as well as our unwritten conventions in the existing parts of the code base.."_

So yes, `enum classes` are nearly everywhere all caps.

elfenpiff (approved):

iceoryx_utils/include/iceoryx_utils/internal/cxx/expected.inl
```diff
@@ -440,32 +306,12 @@ expected<ValueType, ErrorType>::if_empty(const cxx::function_ref<void(void)>& ca
...
-    return const_cast<expected*>(this)->on_success(callable);
-}
-
 template <typename ValueType, typename ErrorType>
 template <typename T>
 inline expected<ValueType, ErrorType>::operator expected<T>() noexcept
 {
-    if (this->has_error())
-        return error<ErrorType>(this->get_error());
+    if (has_error())
```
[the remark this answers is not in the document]
elBoberido: the brainwashing worked

iceoryx_utils/test/moduletests/test_cxx_expected.cpp
```diff
@@ -50,247 +51,467 @@ class expected_test : public Test
...
+
+
+using TestClassAdapter = iox::cxx::ErrorTypeAdapter<expected_test::TestClass>;
+
+enum class TestError : uint8_t
+{
+    INVALID_STATE,
+    ERROR1,
+    ERROR2,
+    ERROR3
```
elBoberido, line 96: Where is the creativity gone? I expected at least one hypnotoad and one Chuck Norris

iceoryx_utils/include/iceoryx_utils/error_handling/error_handling.hpp
```diff
@@ -137,6 +137,7 @@ namespace iox
 
 enum class Error : uint32_t
 {
+    INVALID_STATE,
     ICEORYX_ERRORS(CREATE_ICEORYX_ERROR_ENUM)
 };
```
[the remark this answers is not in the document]
mossmaurice: Yes, in `posh_runtime.hpp:410`

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
```diff
@@ -719,7 +542,8 @@ class expected<ValueType, ErrorType>
     ///     })
     /// @endcode
     ///
-    template <typename Optional = ValueType, typename std::enable_if<is_optional<Optional>::value, int>::type = 0>
+    template <typename Optional = ValueType,
+              typename std::enable_if<internal::IsOptional<Optional>::value, int>::type = 0>
```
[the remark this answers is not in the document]
mossmaurice: For me personally, it's short enough.

iceoryx_utils/test/stresstests/benchmark_optional_and_expected/benchmark_optional_and_expected.cpp
```diff
@@ -19,8 +20,25 @@
...
+template <>
+struct ErrorTypeAdapter<uint64_t>
+{
+    static uint64_t getInvalidState()
+    {
+        return 0;
+    };
+};
+} // namespace cxx
+} // namespace iox
```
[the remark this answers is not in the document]
mossmaurice: Nope, unfortunately not :disappointed: Currently I've no other idea.

iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp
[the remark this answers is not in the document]
mossmaurice: Valid point. Added a todo in #556

iceoryx_utils/include/iceoryx_utils/error_handling/error_handling.hpp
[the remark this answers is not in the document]
elBoberido: could you please change that by creating a `PoshRuntimeError` enum?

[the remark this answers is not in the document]
mossmaurice: You're, right. Will be done in follow-up #578

elBoberido (approved):

PR #565, opened 2021-02-17
Iox #542 sigaction abstraction

### LINKED ISSUE #542: Fix resource leak caused by move ctor/assignment

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

1. [x] Merge

## References

- Partially closes #542

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/posix_wrapper/signal_handler.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/posix_wrapper/signal_handler.hpp
@@ -1,0 +1,61 @@
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+#ifndef IOX_UTILS_POSIX_WRAPPER_SIGNAL_HANDLER_HPP
+#define IOX_UTILS_POSIX_WRAPPER_SIGNAL_HANDLER_HPP
+
+#include <csignal>
+
+namespace iox
+{
+namespace posix
+{
+using SignalHandlerCallback_t = void (*)(int);
+
+enum class Signal : int
+{
+    BUS = SIGBUS,
+    INT = SIGINT,
+    TERM = SIGTERM
+    /// @attention never add SIGKILL or SIGSTOP into this list, they cannot be caught
+    ///            and sigaction returns the errno EINVAL
+};
+
+class SignalGuard
+{
+  public:
+    SignalGuard() noexcept = default;
+    SignalGuard(const Signal signal, const struct sigaction& previousAction) noexcept;
+    SignalGuard(const SignalGuard&) = delete;
+    SignalGuard(SignalGuard&& rhs) noexcept;
+    ~SignalGuard() noexcept;
+
+    SignalGuard& operator=(const SignalGuard& rhs) = delete;
+    SignalGuard& operator=(SignalGuard&& rhs) noexcept;
+
+  private:
+    void restorePreviousAction() noexcept;
+
+  private:
+    Signal m_signal;
+    struct sigaction m_previousAction;
+    bool m_doRestorePreviousAction{false};
+};
+
+SignalGuard registerSignalHandler(const Signal signal, const SignalHandlerCallback_t callback) noexcept;
+} // namespace posix
+} // namespace iox
+#endif
+
--- a/iceoryx_utils/include/iceoryx_utils/posix_wrapper/signal_handler.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/posix_wrapper/signal_handler.hpp
@@ -44,7 +44,7 @@ enum class Signal : int
 ///         after the call.
 /// @code
 ///    {
-///      auto sigHandler = registerSignalHandler(Signal::BUS, printErrorMessage);
+///      auto signalGuard = registerSignalHandler(Signal::BUS, printErrorMessage);
 ///      my_c_call_which_can_cause_SIGBUS();
 ///    }
 ///    // here we are out of scope and the signal action for Signal::BUS is restored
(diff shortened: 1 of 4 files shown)


iceoryx_utils/source/posix_wrapper/signal_handler.cpp
```diff
@@ -0,0 +1,108 @@
...
+                  << std::endl;
+        return SignalGuard();
+    }
+
+    action.sa_handler = callback;
+    action.sa_flags = 0;
+
+    struct sigaction previousAction;
+
+    if (cxx::makeSmartC(sigaction,
```
dkroenke, line 93: It should be considered that if you register a sigaction multiple times for the same signal () that only the last registered handler is used.

elfenpiff: This is why I wrote the @attention NEVER USE THIS CLASS AS A MEMBER in the doxygen docu. I also described the use case there as well and with this I think this is covered.

iceoryx_utils/include/iceoryx_utils/posix_wrapper/signal_handler.hpp
```diff
@@ -0,0 +1,80 @@
...
+/// @attention NEVER USE THIS CLASS AS A MEMBER VARIABLE! A class which should be used
+///            only in method/function scopes.
+/// @brief The SignalGuard is a class returned by registerSignalHandler. When it goes
+///         out of scope it restores the previous signal action. Typical use case:
+///         One would like to override the signal action in main() or some C posix
+///         makes it necessary to override the standard signal action before and
+///         after the call.
+/// @code
+///    {
+///      auto sigHandler = registerSignalHandler(Signal::BUS, printErrorMessage);
```
elBoberido, line 47: this might be confusing since you get the `SignalGuard` not the signal handler

iceoryx_utils/include/iceoryx_utils/posix_wrapper/signal_handler.hpp
```diff
@@ -0,0 +1,80 @@
...
+///      my_c_call_which_can_cause_SIGBUS();
+///    }
+///    // here we are out of scope and the signal action for Signal::BUS is restored
+/// @endcode
+class SignalGuard
+{
+  public:
+    SignalGuard() noexcept = default;
+    SignalGuard(const Signal signal, const struct sigaction& previousAction) noexcept;
```
elBoberido, line 56: Please add documentation.

Should this be public? Maybe it's less error prone if it can only be used by `registerSignalHandler´

iceoryx_utils/source/posix_wrapper/signal_handler.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        restorePreviousAction();
+
+        m_signal = std::move(rhs.m_signal);
+        m_previousAction = std::move(rhs.m_previousAction);
+        m_doRestorePreviousAction = std::move(rhs.m_doRestorePreviousAction);
+
+        rhs.m_doRestorePreviousAction = false;
+    }
+    return *this;
```
elBoberido, line 53: It might be surprising that the move restores the previous action. It is technically correct, but still surprising. I would delete the move assignment and allow only the move ctor.

elBoberido (commented):
I did not yet review the tests.
Regarding the code. Let's assume I want to register `SIGINT` and `SIGTERM` in a function. An because I don't like two variables I create a `cxx::vector` and do an `emplace`. In which order are they destroyed and which sigaction will be restored at last?

iceoryx_utils/source/posix_wrapper/signal_handler.cpp
dkroenke: I'm referring to the case that `registerSignalHandler` overwrite the previous set `sa_handler` silently when you call it multiple times. So either we state that clearly in the documentation or we need to prevent that.

elfenpiff:
> I did not yet review the tests.
> Regarding the code. Let's assume I want to register `SIGINT` and `SIGTERM` in a function. An because I don't like two variables I create a `cxx::vector` and do an `emplace`. In which order are they destroyed and which sigaction will be restored at last?

Great question. Since they are two different signals both will be restored to their original state!

dkroenke (dismissed):

elBoberido:
> > I did not yet review the tests.
> > Regarding the code. Let's assume I want to register `SIGINT` and `SIGTERM` in a function. An because I don't like two variables I create a `cxx::vector` and do an `emplace`. In which order are they destroyed and which sigaction will be restored at last?
> 
> Great question. Since they are two different signals both will be restored to their original state!

Of course :D
What I was curious about was in which order the `cxx::vector` destroys it's elements. Let's assume you register `SIGINT` twice and store it in a `cxx::vector`. Are they destroyed in reverse order of creation? It's a misuse in this case, but what about other classes stored in a `cxx::vector`? If it is not destroyed in reverse order, is it okay?

iceoryx_utils/test/moduletests/test_posix_signal_handler.cpp
```diff
@@ -0,0 +1,133 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "iceoryx_utils/posix_wrapper/signal_handler.hpp"
+#include "test.hpp"
+
+using namespace ::testing;
+using namespace iox::posix;
+
+int signalOfCallback1 = 0;
+int signalOfCallback2 = 0;
```
elBoberido, line 24: Shouldn't this be atomics? I'm not quite sure, but since it is set from a different context it might be necessary.

MatthiasKillat: I am also not quite sure. I think the kernel causes the signalHandler to be executed in the thread which receives the signal, so an atomic would not be necessary here. A quick search confirms this, but it is not a reliable source https://stackoverflow.com/questions/6949025/how-are-asynchronous-signal-handlers-executed-on-linux.

Would not hurt to make those atomic I guess, then we would be sure the updates will be visible here as well (after some cache syncs at least....).

MatthiasKillat:
@elBoberido  According to https://stackoverflow.com/questions/6169125/order-of-destruction-of-elements-of-an-stdvector which refers to the standard the elements are destroyed in reverse order.

I would consider this a misuse by the user, she needs to consider destruction order of the container when using such scope guards. (It might get much less predictable with structures using trees for internal representation, e.g. maps, the user needs to consider this)

TLDR: not the responsibility of this class.

dkroenke (approved):

MatthiasKillat (approved):

Issue #376, opened 2020-11-18
DDS Gateway Housekeeping


ithier:
## Brief feature description

The following maintenance tasks need to be done to keep the dds gateway code healthy:
- [x] Update CMake build to retrieve cyclonedds-cxx library from the [official github repository](https://github.com/eclipse-cyclonedds/cyclonedds-cxx)
- [x] Refactor gateways to use new publisher/subscriber APIs
- [ ] Bring test coverage up to _at least_ 70% 
- [ ] Utilize the improved waitset to make the gateways reactive
- [ ] make `DataWriter::write` return a `cxx::expected`
- [ ] add convertor from `mepoo::ChunkHeader` to `dds::IoxChunkDatagramHeader`
- Findings budrus test run:
- [ ] No documentation how to setup a config file 
- [ ] When building with the build script the cyclone libraries are in ...build/dependencies/install/lib64 (not documented but needed for setting  LD_LIBRARY_PATH
- [ ] The default config should include the icedelivery topic. This allows to have a first running example with icedelivery 
- [ ] "Terminated" endless loop when RouDi gets killed?
- [ ] Replace `Channel` concept with `cxx::pair` (needs to be implemented)

These should be completed before further extension of the gateway.

Some ToDo items from #360:
- [ ] Modify implementation of `iox::dds::CycloneDataReader::hasNewSamples()` so it can be made `const`
- [ ] Resolve mismatch of uint sizes of `nextSampleSize` in `CycloneDataReader::peekNextSize()`
- [ ] Add error handling to semaphore creation in Iceoryx2DDS `main.cpp`
- [ ] Simplify three nested lambdas in `void DDS2IceoryxGateway<channel_t, gateway_t>::forward()`. For example
   ```
   auto maybeNextSize = reader->peakNextSize();
   if(!maybeNextSize)
   {
       continue;
   }
   auto& size = maybeNextSize.value();
   ...
   ```
   and only use for for reader->takeNext the approach with the lambda.
   It would be more code, but also less complex

budrus:
@ithier maybe it would also make sense to use the waitset in the gateway once we are ready with #341. Then you can have it more event driven and not with cyclic loops?

ithier:
Definitely would. I'll add it to the list.

budrus:
[...]

PR #577, opened 2021-02-23
iox-#14 Prepare sample for custom chunk header

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

In order to add the custom chunk header to the typed publisher and subscriber, the `Sample` must also get an additional template parameter for the custom chunk header type. By default this must be `void`, to keep the current behavior. The `Sample` already had a specialization for `void T` which made it impossible to have a default type for the custom chunk header. This PR reworks the `Sample` to be able to use an additional template argument for the custom chunk header which defaults to `void`.

1. [x] Merge

## References

- Partly closes #14

### DIFF
--- a/iceoryx_dds/test/moduletests/test_iox_to_dds.cpp
+++ b/iceoryx_dds/test/moduletests/test_iox_to_dds.cpp
@@ -39,12 +39,12 @@ using ::testing::Return;
 using ::testing::SetArgPointee;
 
 // ======================================== Helpers ======================================== //
-using TestChannel = iox::gw::Channel<MockSubscriber<void>, MockDataWriter>;
+using TestChannel = iox::gw::Channel<MockSubscriber, MockDataWriter>;
 using TestGateway =
     iox::dds::Iceoryx2DDSGateway<TestChannel, MockGenericGateway<TestChannel, iox::popo::SubscriberOptions>>;
 
 // ======================================== Fixture ======================================== //
-class Iceoryx2DDSGatewayTest : public DDSGatewayTestFixture<MockSubscriber<void>, MockDataWriter>
+class Iceoryx2DDSGatewayTest : public DDSGatewayTestFixture<MockSubscriber, MockDataWriter>
 {
 };
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
@@ -133,9 +133,9 @@ inline void Sample<T>::publish() noexcept
 }
 
 template <typename T>
-inline void Sample<T>::release() noexcept
+inline T* Sample<T>::release() noexcept
 {
-    m_members.sampleUniquePtr.release();
+    return m_members.sampleUniquePtr.release();
 }
 
 } // namespace popo
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
@@ -137,13 +137,14 @@ inline const T* Sample<T>::operator->() const noexcept
 
 template <typename T>
 template <typename S, typename>
-inline T& Sample<T>::operator*() noexcept
+inline S& Sample<T>::operator*() noexcept
 {
     return *get();
 }
 
 template <typename T>
-inline const T& Sample<T>::operator*() const noexcept
+template <typename S, typename>
+inline const S& Sample<T>::operator*() const noexcept
 {
     return *get();
 }
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
@@ -25,37 +25,38 @@ namespace popo
 namespace internal
 {
 template <typename T>
-inline SamplePrivateData<T>::SamplePrivateData(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher) noexcept
-    : samplePtr(std::move(samplePtr))
+inline SamplePrivateData<T>::SamplePrivateData(cxx::unique_ptr<T>&& sampleUniquePtr,
+                                               PublisherInterface<T>& publisher) noexcept
+    : sampleUniquePtr(std::move(sampleUniquePtr))
     , publisherRef(publisher)
 {
 }
 
 template <typename T>
-inline SamplePrivateData<const T>::SamplePrivateData(cxx::unique_ptr<const T>&& samplePtr) noexcept
-    : samplePtr(std::move(samplePtr))
+inline SamplePrivateData<const T>::SamplePrivateData(cxx::unique_ptr<const T>&& sampleUniquePtr) noexcept
+    : sampleUniquePtr(std::move(sampleUniquePtr))
 {
 }
 } // namespace internal
 
 template <typename T>
 template <typename S, typename>
-inline Sample<T>::Sample(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher) noexcept
-    : m_members({std::move(samplePtr), publisher})
+inline Sample<T>::Sample(cxx::unique_ptr<T>&& sampleUniquePtr, PublisherInterface<T>& publisher) noexcept
+    : m_members({std::move(sampleUniquePtr), publisher})
 {
 }
 
 template <typename T>
 template <typename S, typename>
-inline Sample<T>::Sample(cxx::unique_ptr<T>&& samplePtr) noexcept
-    : m_members(std::move(samplePtr))
+inline Sample<T>::Sample(cxx::unique_ptr<T>&& sampleUniquePtr) noexcept
+    : m_members(std::move(sampleUniquePtr))
 {
 }
 
 template <typename T>
 inline Sample<T>::Sample(std::nullptr_t) noexcept
 {
-    m_members.samplePtr.reset();
+    m_members.sampleUniquePtr.reset();
 }
 
 template <typename T>
@@ -94,33 +95,33 @@ template <typename T>
 template <typename S, typename>
 inline T* Sample<T>::get() noexcept
 {
-    return m_members.samplePtr.get();
+    return m_members.sampleUniquePtr.get();
 }
 
 template <typename T>
 inline const T* Sample<T>::get() const noexcept
 {
-    return m_members.samplePtr.get();
+    return m_members.sampleUniquePtr.get();
 }
 
 template <typename T>
 template <typename S, typename>
 inline mepoo::ChunkHeader* Sample<T>::getHeader() noexcept
 {
-    return mepoo::ChunkHeader::fromPayload(m_members.samplePtr.get());
+    return mepoo::ChunkHeader::fromPayload(m_members.sampleUniquePtr.get());
 }
 
 template <typename T>
 inline const mepoo::ChunkHeader* Sample<T>::getHeader() const noexcept
 {
-    return mepoo::ChunkHeader::fromPayload(m_members.samplePtr.get());
+    return mepoo::ChunkHeader::fromPayload(m_members.sampleUniquePtr.get());
 }
 
 template <typename T>
 template <typename S, typename>
 inline void Sample<T>::publish() noexcept
 {
-    if (m_members.samplePtr)
+    if (m_members.sampleUniquePtr)
     {
         m_members.publisherRef.get().publish(std::move(*this));
     }
@@ -135,7 +136,7 @@ template <typename T>
 template <typename S, typename>
 inline void Sample<T>::release() noexcept
 {
-    m_members.samplePtr.reset();
+    m_members.sampleUniquePtr.reset();
 }
 
 } // namespace popo
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
@@ -143,8 +143,7 @@ inline S& Sample<T>::operator*() noexcept
 }
 
 template <typename T>
-template <typename S, typename>
-inline const S& Sample<T>::operator*() const noexcept
+inline const T& Sample<T>::operator*() const noexcept
 {
     return *get();
 }
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
@@ -133,10 +133,9 @@ inline void Sample<T>::publish() noexcept
 }
 
 template <typename T>
-template <typename S, typename>
 inline void Sample<T>::release() noexcept
 {
-    m_members.sampleUniquePtr.reset();
+    m_members.sampleUniquePtr.release();
 }
 
 } // namespace popo
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
@@ -1,4 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,11 +22,75 @@ namespace iox
 {
 namespace popo
 {
+namespace internal
+{
+// ============================== SamplePrivateData<T> ========================= //
+template <typename T>
+inline SamplePrivateData<T>::SamplePrivateData(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher)
+    : samplePtr(std::move(samplePtr))
+    , publisherRef(publisher)
+{
+}
+
+template <typename T>
+inline SamplePrivateData<T>& SamplePrivateData<T>::operator=(SamplePrivateData<T>&& rhs)
+{
+    if (this != &rhs)
+    {
+        publisherRef = rhs.publisherRef;
+        samplePtr = std::move(rhs.samplePtr);
+        rhs.samplePtr = nullptr;
+    }
+    return *this;
+}
+
+template <typename T>
+inline SamplePrivateData<T>::SamplePrivateData(SamplePrivateData<T>&& rhs)
+    : samplePtr(std::move(rhs.samplePtr))
+    , publisherRef(std::move(rhs.publisherRef))
+{
+    rhs.samplePtr = nullptr;
+}
+
+// ============================== SamplePrivateData<const T> ========================= //
+
+template <typename T>
+inline SamplePrivateData<const T>::SamplePrivateData(cxx::unique_ptr<const T>&& samplePtr) noexcept
+    : samplePtr(std::move(samplePtr))
+{
+}
+
+template <typename T>
+inline SamplePrivateData<const T>& SamplePrivateData<const T>::operator=(SamplePrivateData<const T>&& rhs)
+{
+    if (this != &rhs)
+    {
+        samplePtr = std::move(rhs.samplePtr);
+        rhs.samplePtr = nullptr;
+    }
+    return *this;
+}
+
+template <typename T>
+inline SamplePrivateData<const T>::SamplePrivateData(SamplePrivateData<const T>&& rhs)
+    : samplePtr(std::move(rhs.samplePtr))
+{
+    rhs.samplePtr = nullptr;
+}
+} // namespace internal
+
 // ============================== Sample<T> ========================= //
 template <typename T>
+template <typename S, typename>
 inline Sample<T>::Sample(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher)
-    : m_samplePtr(std::move(samplePtr))
-    , m_publisherRef(publisher)
+    : m_members({std::move(samplePtr), publisher})
+{
+}
+
+template <typename T>
+template <typename S, typename>
+inline Sample<T>::Sample(cxx::unique_ptr<T>&& samplePtr) noexcept
+    : m_members(std::move(samplePtr))
 {
 }
 
@@ -34,33 +99,31 @@ inline Sample<T>& Sample<T>::operator=(Sample<T>&& rhs)
 {
     if (this != &rhs)
     {
-        m_publisherRef = rhs.m_publisherRef;
-        m_samplePtr = std::move(rhs.m_samplePtr);
-        rhs.m_samplePtr = nullptr;
+        m_members = std::move(rhs.m_members);
     }
     return *this;
 }
 
 template <typename T>
 inline Sample<T>::Sample(Sample<T>&& rhs)
-    : m_publisherRef(rhs.m_publisherRef) // Need to initialize references in initializer list.
+    : m_members(std::move(rhs.m_members))
 {
-    *this = std::move(rhs);
 }
 
 template <typename T>
 inline Sample<T>::~Sample()
 {
-    m_samplePtr = nullptr;
+    m_members.samplePtr = nullptr;
 }
 
 template <typename T>
 inline Sample<T>::Sample(std::nullptr_t) noexcept
 {
-    m_samplePtr = nullptr; // The pointer will take care of cleaning up resources.
+    m_members.samplePtr = nullptr; // The pointer will take care of cleaning up resources.
 }
 
 template <typename T>
+template <typename S, typename>
 inline T* Sample<T>::operator->() noexcept
 {
     return get();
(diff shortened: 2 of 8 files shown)


mossmaurice (changes requested):
LGTM in general, some minor findings.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -27,6 +28,41 @@ namespace popo
...
+{
+    SamplePrivateData(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher);
+
+    SamplePrivateData(SamplePrivateData&& rhs);
+    SamplePrivateData& operator=(SamplePrivateData&& rhs);
+
+    SamplePrivateData(const SamplePrivateData&) = delete;
+    SamplePrivateData& operator=(const SamplePrivateData&) = delete;
+
+    cxx::unique_ptr<T> samplePtr{[](T* const) {}}; // Placeholder. This is overwritten on sample construction.
```
elfenpiff, line 45: Here I discovered multiple issues.
1. The `cxx::unique_ptr` should get a public member alias `using DeleteCallback = cxx::function_ref<void(T*)>;` and in this line you should use the default ctor of the `cxx::unique_ptr` (after you have handled point 2 ;) ).
2. The `cxx::unique_ptr` should get a default ctor where it is constructed with an empty `DeleteCallback` and behaves exactly like the stl counterpart. This is required since the `cxx::unique_ptr` is movable so that the source of the move operation is exactly in the same state as being default constructed - does not contain any pointer and has an unset `DeleteCallback`.
3. The `cxx::unique_ptr` has to always verify before calling the `DeleteCallback` that this callback is set, otherwise we call an unset `DeleteCallback` which is undefined behavior.
4. The `cxx::unique_ptr` constructor where you can set the `DeleteCallback` should always get a setted function_ref. We could do this in a straight forward way and just test the `DeleteCallback` if it is set in the ctor - but what do we do if its not set?! Maybe the solution is something like this:
```cpp
class always_valid_function_ref : public function_ref {
   public:
       always_valid_function_ref() =delete; // delete default ctor to ensure that it is set up
       operator function_ref(); // conversion operator to function_ref so that the user doesn't have to do a thing
};
```

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -27,6 +28,41 @@ namespace popo
...
+    std::reference_wrapper<PublisherInterface<T>> publisherRef;
+};
+
+/// @brief specialization of helper struct for sample for const T
+template <typename T>
+struct SamplePrivateData<const T>
+{
+    SamplePrivateData(cxx::unique_ptr<const T>&& samplePtr) noexcept;
+
+    SamplePrivateData(SamplePrivateData&& rhs);
```
elfenpiff, line 55: Please add noexcept.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -27,6 +28,41 @@ namespace popo
...
+    SamplePrivateData& operator=(SamplePrivateData&& rhs);
+
+    SamplePrivateData(const SamplePrivateData&) = delete;
+    SamplePrivateData& operator=(const SamplePrivateData&) = delete;
+
+    cxx::unique_ptr<const T> samplePtr{
```
elfenpiff, line 61: This is just weird. The `cxx::unique_ptr` requires a default ctor which should be used here.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -35,40 +71,57 @@ template <typename T>
...
     /// @return A T& to the encapsulated type.
-    /// @details Only available for non void type T.
+    /// @details Only available for non-const type T.
     ///
-    template <typename S = T,
-              typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_same<S, void>::value, S>>
+    template <
+        typename S = T,
+        typename =
+            std::enable_if_t<std::is_same<S, T>::value && !std::is_same<S, void>::value && !std::is_const<S>::value, S>>
```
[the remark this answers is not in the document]
elBoberido: not possible since it the template parameter of the method

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -35,40 +71,57 @@ template <typename T>
...
 
     ///
     /// @brief operator* Provide a reference to the encapsulated type.
     /// @return A T& to the encapsulated type.
-    /// @details Only available for non void type T.
+    /// @details Only available for non-const type T.
     ///
-    template <typename S = T,
-              typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_same<S, void>::value, S>>
+    template <
```
elfenpiff, line 116: This `enable_if` construct will cause ugly template errors. I think in this case we could also use `static_assert` and provide the user a nice to read error message.

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -21,175 +22,185 @@ namespace iox
 {
 namespace popo
 {
-// ============================== Sample<T> ========================= //
+namespace internal
+{
+// ============================== SamplePrivateData<T> ========================= //
```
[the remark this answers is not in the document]
elBoberido: It was already there and I thought it's nice to have a separator. How about a `// BEGIN Sample<T>` and `//END Sample<T>`? With this you get some code folding.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -81,69 +134,51 @@ class Sample
...
+    /// @details Only available for non-const type T.
     ///
+    template <typename S = T, typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_const<S>::value, S>>
     mepoo::ChunkHeader* getHeader() noexcept;
 
+    ///
+    /// @brief header Retrieve the header of the underlying memory chunk loaned to the sample.
+    /// @return The const ChunkHeader of the underlying memory chunk.
+    ///
+    const mepoo::ChunkHeader* getHeader() const noexcept;
```
elfenpiff, line 162: Theoretically a pointer can always be a nullptr. To promise the user that they always get a pointer to a valid header I would return here a reference like `const mepoo::ChunkHeader& getHeader() const noexcept;` Otherwise the question always arises, do I have to check that the pointer != nullptr when I access it.

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -21,175 +22,185 @@ namespace iox
...
 template <typename T>
-inline Sample<T>& Sample<T>::operator=(Sample<T>&& rhs)
+inline SamplePrivateData<T>& SamplePrivateData<T>::operator=(SamplePrivateData<T>&& rhs)
 {
     if (this != &rhs)
     {
-        m_publisherRef = rhs.m_publisherRef;
-        m_samplePtr = std::move(rhs.m_samplePtr);
-        rhs.m_samplePtr = nullptr;
+        publisherRef = rhs.publisherRef;
```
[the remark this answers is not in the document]
elBoberido: good question. c&p from the old code

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -81,69 +134,51 @@ class Sample
...
     ///
     const T* get() const noexcept;
 
     ///
     /// @brief header Retrieve the header of the underlying memory chunk loaned to the sample.
     /// @return The ChunkHeader of the underlying memory chunk.
+    /// @details Only available for non-const type T.
     ///
+    template <typename S = T, typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_const<S>::value, S>>
     mepoo::ChunkHeader* getHeader() noexcept;
```
elfenpiff, line 156: Could you return a reference here?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -81,69 +134,51 @@ class Sample
...
+    /// @return The const ChunkHeader of the underlying memory chunk.
+    ///
+    const mepoo::ChunkHeader* getHeader() const noexcept;
+
     ///
     /// @brief publish Publish the sample via the publisher from which it was loaned and automatically
     /// release ownership to it.
+    /// @details Only available for non-const type T.
     ///
+    template <typename S = T, typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_const<S>::value, S>>
```
elfenpiff, line 169: We require the rule: Always prefer `static_assert` over `enable_if`.
I think here we can again use `static_assert` to provide a nice error message for the user.
Maybe the rule could be: only use `enable_if` when you would like to do some template specialization.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -81,69 +134,51 @@ class Sample
...
     void publish() noexcept;
 
     ///
     /// @brief release Manually release ownership of the loaned memory chunk.
-    /// @details This prevents the sample from automatically releasing ownership on destruction.
+    /// @details This prevents the sample from automatically releasing ownership on destruction and is only available
+    /// for non-const type T.
     ///
+    template <typename S = T, typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_const<S>::value, S>>
```
elfenpiff, line 177: Is `static_assert` possible?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -81,69 +134,51 @@ class Sample
...
+    /// @return a const pointer to the encapsulated type.
     ///
     const T* get() const noexcept;
 
     ///
     /// @brief header Retrieve the header of the underlying memory chunk loaned to the sample.
     /// @return The ChunkHeader of the underlying memory chunk.
+    /// @details Only available for non-const type T.
     ///
+    template <typename S = T, typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_const<S>::value, S>>
```
elfenpiff, line 155: Is `static_assert` possible?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -81,69 +134,51 @@ class Sample
     operator bool() const;
 
     ///
-    /// @brief allocation Access to the memory chunk loaned to the sample.
-    /// @return
+    /// @brief allocation Access to the encapsulated type loaned to the sample.
+    /// @return a pointer to the encapsulated type.
+    /// @details Only available for non-const type T.
     ///
+    template <typename S = T, typename = std::enable_if_t<std::is_same<S, T>::value && !std::is_const<S>::value, S>>
```
elfenpiff, line 141: Is `static_assert` possible?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -27,6 +28,41 @@ namespace popo
...
 
+namespace internal
+{
+/// @brief helper struct for sample
+template <typename T>
+struct SamplePrivateData
+{
+    SamplePrivateData(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher);
+
+    SamplePrivateData(SamplePrivateData&& rhs);
```
elfenpiff, line 39: Please add `noexcept`.

elfenpiff (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -21,175 +22,185 @@ namespace iox
...
+        rhs.samplePtr = nullptr;
     }
     return *this;
 }
 
 template <typename T>
-inline Sample<T>::Sample(Sample<T>&& rhs)
-    : m_publisherRef(rhs.m_publisherRef) // Need to initialize references in initializer list.
+inline SamplePrivateData<T>::SamplePrivateData(SamplePrivateData<T>&& rhs)
+    : samplePtr(std::move(rhs.samplePtr))
```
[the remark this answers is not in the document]
elBoberido: actually two reasons
- `m_publisherRef` has to be set in the initializer list
- the sample is moved quite often and this is a nice optimization
In the end the code duplication is minimal and from my point of view worth to do it

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -21,175 +22,185 @@ namespace iox
...
+    return *this;
 }
 
 template <typename T>
-inline const T* Sample<T>::operator->() const noexcept
+inline SamplePrivateData<const T>::SamplePrivateData(SamplePrivateData<const T>&& rhs)
+    : samplePtr(std::move(rhs.samplePtr))
 {
-    return get();
+    rhs.samplePtr = nullptr;
```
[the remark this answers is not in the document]
elBoberido: same reason. The code duplication is minimal and the `unique_ptr` is not initialized twice

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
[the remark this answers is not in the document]
elfenpiff: I think you could write:
```cpp
template<typename S, typename T>
using myFuncyEnableIf = std::enable_if_t<std::is_same<S, T>::value && !std::is_same<S, void>::value && !std::is_const<S>::value, S>>;
```
and then use it in the method like
```cpp
template <
        typename S = T,
        typename = myFunkyEnablIf<S, T>
```

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: This is just to enable some methods only for the publisher, which can mutate the sample.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: will be done in a follow up PR

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: in this case the compiler would probably choose the non-const method in a non-const environment, which triggers the static_assert and makes it awkward to use the correct method

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: in this case the compiler would probably choose the non-const method in a non-const environment, which triggers the static_assert and makes it awkward to use the correct method

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: Previously this was only available for the sample when loaned from a publisher and probably also a remnant of the previous refactoring with the `void` sample. I would suggest to get rid of this method, since it makes the `Sample` explicitly null-able. I would suggest to use a `cxx::optional<Sample<T>>` if it is necessary to store a sample for later use and being able to release it.

@ithier what do you think?

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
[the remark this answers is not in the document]
elBoberido: solved the problem by just using the default move ctor and assignment operator

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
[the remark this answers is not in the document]
elBoberido: solved the problem by just using the default move ctor and assignment operator

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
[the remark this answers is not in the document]
elBoberido: solved the problem by just using the default move ctor and assignment operator

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -21,175 +22,185 @@ namespace iox
...
 {
-// ============================== Sample<T> ========================= //
+namespace internal
+{
+// ============================== SamplePrivateData<T> ========================= //
 template <typename T>
-inline Sample<T>::Sample(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher)
-    : m_samplePtr(std::move(samplePtr))
-    , m_publisherRef(publisher)
+inline SamplePrivateData<T>::SamplePrivateData(cxx::unique_ptr<T>&& samplePtr, PublisherInterface<T>& publisher)
```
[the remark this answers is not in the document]
elBoberido: done

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: initialized the pointer in the ctor initialization list

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: I removed the dummy initialization of the pointer and initialized it directly in the initializer list of the ctor. All the other issues are not strictly related to the `Sample` and should therefore be done in a separate issue. This PR is just a refactoring of the sample to be able to implement the custom header.

Shall I create an issue for the `unique_ptr` with your points?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: Yes, I also noticed this right a few minutes ago. It probably should not be public since it can lead to chunk leaks.sample
I will revert the change to not `reset` the pointer but `release` it. This Method should probably be private and the `Publisher` a friend of `Sample`.

iceoryx_dds/test/moduletests/test_iox_to_dds.cpp
```diff
@@ -39,12 +39,12 @@ using ::testing::Return;
 using ::testing::SetArgPointee;
 
 // ======================================== Helpers ======================================== //
-using TestChannel = iox::gw::Channel<MockSubscriber<void>, MockDataWriter>;
+using TestChannel = iox::gw::Channel<MockSubscriber, MockDataWriter>;
```
ithier, line 42: Completely unrelated thought - I think this `Channel` abstraction I introduced doesn't bring much extra value and its purpose would be easily fulfilled with something more generic like `std::pair`, if we ever end up porting this.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -27,123 +28,155 @@ namespace popo
...
     /// release ownership to it.
+    /// @details Only available for non-const type T.
     ///
+    template <typename S = T, typename = ForPublisherOnly<S, T>>
     void publish() noexcept;
 
     ///
-    /// @brief release Manually release ownership of the loaned memory chunk.
-    /// @details This prevents the sample from automatically releasing ownership on destruction.
+    /// @brief release Manually transfers ownership of the loaned memory chunk back to the mempool.
```
ithier, line 172: This comment is not quite right. Calling `releaseChunk()` would transfer the chunk back to the memory pool, which this method does not do.

elBoberido: That's exactly the reason I assumed this was wrong. It's the `release` semantic of the `unique_ptr` not of the `chunk`

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: Forget this. The publisher expects a sample and therefore `publish` cannot be in a base class.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elfenpiff: I slept over the used term here and I would recommend to follow the STL semantic because this is the common semantic in the C++ world. This means that `release` should be named `reset` since every C++ developer would expect that it behaves like that and the sample is somehow also a unique ptr.
I know that we also use `releaseChunk` but in this case I would break the semantics here since the name is also somehow different and it is not the name of another unique ptr like structure.

elBoberido: @elfenpiff I got it wrong. The `Sample::release` method should do exactly what `unique_ptr::release` does, releasing the ownership to the caller. After moving the untyped API away from `Sample<void>` and having the `Sample` only for the typed API, this should probably not be accessible by the user anymore, just by the publisher.

budrus: I don't get it. This `release()` does not return the resource like a unique_ptr. And @elBoberido changed the internal to do a reset of the unique_ptr. I think because a releae on the unique_ptr without taking care of the returned resource is a memory leak in our case. 
So from what it is doing now it is a reset. But why do we need this? If `publish()` is not called, the Sample d'tor will return the chunk to the mempool without publishing it. Is it an early release? But what is a `Sample` good for that has no chunk?

elBoberido: @budrus to clarify this a bit. I wrongly assumed that he `Sample::release` has the same semantic as `releaseChunk` and therefore called `unique_ptr::reset` instead of `unique_ptr::release` but this is wrong since the publisher needs to gain ownership of the chunk from the `Sample` and therefore needs to call `Sample::release` without invoking the custom deleter. Previously the `Sample::release` was also necessary for the untyped API, but this is not the case anymore. To make a long story short, `Sample::release` should be private and only be accessible by the `Publisher`, else this will lead to chunk leaks.

If someone want's to control the lifetime of the `Sample` manually, a container must be used, e.g. `cxx::optional<Sample<T>>`.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: will be done in a follow up PR, like the non-const variant

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
budrus: @elBoberido Got it. Then you found a way to create a leak and the solution is that only the publisher is allowed to call `release` which we should maybe rename to `reset`

iceoryx_dds/test/moduletests/test_iox_to_dds.cpp
elBoberido: Good point. Shall this be added to #376?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
budrus: @elBoberido You changed it to release the unique_ptr again. But the `Sample` `release` does not provide the resource. 
Should we 
* call reset and name the method in Sample also reset?
* call it release and return the resource and the publisher than calls release to get the ownership of the chunk?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: I'm a bit torn about the `publish` method. On the one hand, it's nice syntactical sugar. On the other hand, it complicates stuff and with the refactoring it will also show up in the auto-completion of the IDE. I have an idea which might help with the auto-completion, but I would rather spent my time on more important stuff. How about removing this feature for 1.0 and bringing it back afterwards?

@budrus @ithier what do you think?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: @budrus currently the publisher uses `Sample::get` to get a raw pointer to the chunk. But yes, it might be clearer what happens if `Sample::release` returns the raw pointer. Then the API would also be more similar to `unique_ptr::release`

budrus (approved):

ithier (approved):

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
budrus: @elBoberido What is the issue with the auto-completion. It suggests publish for the `Sample` on the Subscriber side? 
It seems to be a cool feature. We do not have much experience how people use the API but I already see that we use this feature when I look at [spontaneously written code](https://github.com/eclipse-iceoryx/iceoryx/issues/581#issuecomment-785789697). Your bad IDE would not be a good reason for me to remove this feature ;-) Or are there also other aspects?

elBoberido: @budrus it's not really an issue. If the method is selected it is marked as unavailable. I also checked what happens when this method is removed and the code looks a bit awkward with so many `publish` words in a row. So let's keep it.

elfenpiff (approved):

Issue #574, opened 2021-02-22
Duplicated service description entries within gateway config toml file should not result in duplication of service entries


shankar-in:
## Required information

**Operating system:**
Iceoryx Enviroments Linux / Ubuntu

**Compiler version:**
Normal Iceoryx build setup

**Observed result or behaviour:**
If the gateway config toml file has duplicated service descriptions the parser doesnt validates this and creates duplicated service description entries

**Expected result or behaviour:**
The parser should validate this and service entry dupication should not happen

**Conditions where it occurred / Performed steps:**
Ex:

    serviceEntry->insert("service", "service");
    serviceEntry->insert("instance", "instance");
    serviceEntry->insert("event", "event");

    serviceEntry->insert("service", "service");
    serviceEntry->insert("instance", "instance");
    serviceEntry->insert("event", "event");

elBoberido:
I'm not quite familiar with the gateway config but your example shows C++ code. Could you please clarify this a bit?

ithier:
Creating duplicate publishers/subscribers/data writers/data readers is probably the wrong behaviour. I can't think of any reason why they should be duplicated.
Possible solutions:
* skip over duplicate entries found in the configuration
* terminate application with error, prompting the user to fix the config

The fix should be relatively simple, relevant code is here:
* https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_dds/include/iceoryx_dds/internal/gateway/iox_to_dds.inl#L41
* https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_dds/include/iceoryx_dds/internal/gateway/dds_to_iox.inl#L36

Either change return type to be an expected and return an error on fail, or do the filtering in these methods.

shankar-in:
@elBoberido Yes. its a code snippet to write the data into the toml file. This can be found within iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp. 

@ithier yes agreed with your solution

dkroenke:
@ithier 
Maybe it make more sense to detect duplicate entries early in the parse() stage? I would make the solution not DDS-specific and apply this to the `TomlGatewayConfigParser` class which is intended used by all gateways. There is already a validate function available that can be extended to detect and handle the duplicate service entries:
https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/source/gateway/toml_gateway_config_parser.cpp#L78

[...]

PR #579, opened 2021-02-25
iox-#496 enhancing coverage for gateway Base and toml gateway config parser class for folder gateway

chiranjeevimaddi:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

1. [x] Merge

## References

- Closes -#496


iceoryx_posh/test/moduletests/test_gw_base.cpp
```diff
@@ -0,0 +1,95 @@
...
+    ~GatewayBaseTestDestructor() noexcept
+    {
+    }
+};
+
+TEST_F(GatewayBaseTestSuite, VerifyDestructorIsSuccessful)
+{
+    GatewayBaseTestDestructor base{iox::capro::Interfaces::INTERNAL};
+    iox::popo::InterfacePort* interfaceImpl = base.getInterfaceImpl();
+    base.~GatewayBaseTestDestructor();
```
elfenpiff, line 70: A very basic thing in C++ is: never call a destructor explicitly then any kind of weird thing can happen!

iceoryx_posh/test/moduletests/test_gw_base.cpp
```diff
@@ -0,0 +1,95 @@
...
+    iox::popo::InterfacePort* getInterfaceImpl()
+    {
+        return &m_interfaceImpl;
+    }
+    ~GatewayBaseTestDestructor() noexcept
+    {
+    }
+};
+
+TEST_F(GatewayBaseTestSuite, VerifyDestructorIsSuccessful)
```
elfenpiff, line 66: The naming of the test is not accurate. Please adjust to what you are testing here. For me it seems that you would like to verify that the interface port should be destroyed when the gateway goes out of scope.

iceoryx_posh/test/moduletests/test_gw_base.cpp
```diff
@@ -0,0 +1,95 @@
...
+using ::testing::_;
+
+using namespace iox::runtime;
+using iox::roudi::RouDiEnvironment;
+
+using namespace iox::gw;
+
+using iox::capro::IdString_t;
+
+class GatewayBaseTestSuite : public Test
```
elfenpiff, line 34: The `GatewayBase` class is a base class for all the gateway implementation. Therefore, we should use parameterized tests here, see: https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest.
The idea is that every class which inherits from `GatewayBase` should follow the Liskov Substitution Principle (https://en.wikipedia.org/wiki/Liskov_substitution_principle) and therefore should successfully run the same test then the base class.

iceoryx_posh/test/moduletests/test_gw_base.cpp
```diff
@@ -0,0 +1,95 @@
...
+    iox::runtime::PoshRuntime* m_senderRuntime{&iox::runtime::PoshRuntime::initRuntime("sender")};
+    iox::gw::GatewayBase m_base{iox::capro::Interfaces::INTERNAL};
+
+    void InterOpWait() const
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    }
+};
+
+class GatewayBaseTestDestructor : public GatewayBase
```
elfenpiff, line 50: Could you please move this class into `GatewayBaseSuite`

iceoryx_posh/test/moduletests/test_gw_base.cpp
elfenpiff, line 34: Please rename it into `GatewayBase_test` to follow our coding guidelines.

iceoryx_posh/test/moduletests/test_gw_base.cpp
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2021 by Robert Bosch GmbH. All rights reserved.
```
elfenpiff, line 1: Please rename the file into `test_gateway_gateway_base.cpp`.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,528 @@
...
 {
   public:
     void SetUp(){};
-    void TearDown(){};
+    void TearDown()
+    {
+        if (std::remove(TestFilePath.c_str()) != 0)
+        {
+            std::cerr << "Failed to remove temporary file '" << TestFilePath
+                      << "'. You'll have to remove it by yourself.";
```
elfenpiff, line 45: Please add `std::endl` at the end of `std::cerr`

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,528 @@
...
+        }
+    };
+
+    void CreateTmpTomlFile(std::shared_ptr<cpptoml::table> toml)
+    {
+        std::fstream fs(TestFilePath, std::fstream::out | std::fstream::trunc);
+        if (fs.std::fstream::is_open())
```
elfenpiff, line 52: What happens when this fails? I think you should add an `ASSERT_*` here.

elfenpiff (changes requested):
I reviewed `test_popo_toml_gateway.cpp` only partially but as soon as there is time I will continue.

iceoryx_posh/test/moduletests/test_gw_base.cpp
manel94: Do you mean i should rename this class to GatewayBaseSuite?

iceoryx_posh/test/moduletests/test_gw_base.cpp
manel94: @elfenpiff  as I understand you want to replace the class test by this class and make all the tests Test_P(*

```
class GatewayBase_test : public TestWithParam<iox::capro::Interfaces>
{
  public:
    void SetUp(){};
    void TearDown(){};

    RouDiEnvironment m_roudiEnv{iox::RouDiConfig_t().setDefaults()};
    iox::runtime::PoshRuntime* m_senderRuntime{&iox::runtime::PoshRuntime::initRuntime("sender")};
    iox::gw::GatewayBase m_base{iox::capro::Interfaces::INTERNAL};

    void InterOpWait() const
    {
        std::this_thread::sleep_for(std::chrono::milliseconds(200));
    }
};
```

elfenpiff: The first inheritance is correct but why would you like to inherit from `GatewayBaseTestDestructor`? 
I meant you should use it as a nested class.

iceoryx_posh/test/moduletests/test_gw_base.cpp
elfenpiff: No, I mean a nested class.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,528 @@
...
+
+TEST_F(TomlGatewayConfigParserTest,
+       NumbersUsedInServiceDescriptionInTomlConfigFileReturnNotEmptyConfigObjectAndNoParseError)
+{
+    auto toml = cpptoml::make_table();
+    auto serviceArray = cpptoml::make_table_array();
+
+    auto serviceEntryNumbers = cpptoml::make_table();
+    serviceEntryNumbers->insert("service", "Number1234567890");
+    serviceEntryNumbers->insert("instance", "Number1234567890");
```
dkroenke, line 345: would be good to have more variations here like ``123Number456` or `123456`

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,528 @@
...
+    auto result = TomlGatewayConfigParser::parse();
+    GatewayConfig config = result.value();
+
+    EXPECT_FALSE(result.has_error());
+    EXPECT_TRUE(config.m_configuredServices.empty());
+}
+
+TEST_F(TomlGatewayConfigParserTest, ParseWithEmptyPathReturnEmptyConfig)
+{
+    iox::roudi::ConfigFilePathString_t path = "";
```
dkroenke, line 260: This path is not empty because it creates a string with a null-terminator in it.

manel94: Yeah, what i would to test here is no provided path, i will refactor the test name
image

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,528 @@
...
+    iox::roudi::ConfigFilePathString_t Path =
+        iox::roudi::ConfigFilePathString_t(iox::cxx::TruncateToCapacity, TestFilePath);
+    auto result = TomlGatewayConfigParser::parse(Path);
+
+    ASSERT_TRUE(result.has_error());
+    EXPECT_EQ(TomlGatewayConfigParseError::INCOMPLETE_SERVICE_DESCRIPTION, result.get_error());
+}
+
+TEST_F(TomlGatewayConfigParserTest,
+       ParseWithoutEventNameInServiceDescriptionInTomlConfigFileReturnIncompleteServiceDescriptionError)
```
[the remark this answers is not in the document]
manel94: @dkroenke, But the test it´s passed and we have an INCOMPLETE SERVICE DESCRIPTION, is it a bug ?
image
 I would to mention also that it seems we have  another bug, It was about having duplicated entries in the gateway config file of iceoryx.

```
serviceEntry->insert("service", "service");

    serviceEntry->insert("instance", "instance");

    serviceEntry->insert("event", "event");

 

    serviceEntry->insert("service", "service");

    serviceEntry->insert("instance", "instance");

    serviceEntry->insert("event", "event");
```

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
dkroenke: The Problem is more that this test maybe goes into the right branch here so it is covered but it does no really verifies the behavior because the only indication for the empty path is a printed log message.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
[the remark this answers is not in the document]
dkroenke: Yes it is related to this issue: https://github.com/eclipse-iceoryx/iceoryx/issues/574 I will ask the question there.
For the double service entries it is imho ok to have this test here. This test should then be rewritten when the fix will be applied. The fix is quite easy, so we can maybe apply it here directly.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
manel94: @dkroenke, I understand , do you have any proposal ?

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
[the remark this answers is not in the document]
dkroenke: The reduced ServiceDescription is only needed for ServiceDiscovery, but in a gateway you normally want to do more than only SD with a service, therefore it is ok to enforce the full ServiceDescription here.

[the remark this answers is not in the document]
manel94: @dkroenke  what do you mean by  enforce the full ServiceDescription, could you explain more please? did you talk about Parse Without EventName In ServiceDescription ?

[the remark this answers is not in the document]
dkroenke: There are two variants of ServiceDescription, the normal one have InstanceID, ServiceID and EventID and is used in all our examples. If you want to do a ServiceDiscovery there is only the InstanceID and ServiceID sufficient, so you can also create a "reduced" ServiceDescription. If was just thinking out loud if it make sense that the TOML parser also accept the reduced version, but for now it ok to expect the full version in the parser.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
manel94: @dkroenke i propose here when the path size is equal to zero or we have an invalid path , an error handler should be returned, to be able to test this two cases, what do you think?

dkroenke: According to the guidelines in https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/design/error-handling.md the errorhandler should only be used in RouDi or the Runtime. The TomlGatewayParser is used in gateway executables so this is not an option.
An idea which came in my mind is that maybe we return a config in this case which indicates that something went wrong.
Basically there are two cases: 1. File path empty 2. config file could not be opened
If for example case 1 occur, then could maybe return a config with a service entry which has a ServiceDescription like this:
```
ServiceDescription("wrong_config_file_path", "wrong_config_file_path", "wrong_config_file_path")
```
Then we have an indication at loglevel and the returned config can be tested and tells the user that something went wrong.
@ithier Would it make sense to have that?

ithier: What's wrong with just returning `expected::error`, like it already does ?

ithier: If we want to inform the user of this, maybe we should change the behaviour to return `expected::error` in this case?

Then the caller - likely the main function of a gateway executable - could either terminate the program, or choose to proceed without a configuration.

manel94: yeah that was my suggestion

iceoryx_posh/test/moduletests/test_gw_base.cpp
marthtz: Looks like it's done.

iceoryx_posh/test/moduletests/test_gw_base.cpp
marthtz: Looks like it's done.

iceoryx_posh/test/moduletests/test_gw_base.cpp
marthtz: Looks like it's done.

iceoryx_posh/test/moduletests/test_gw_base.cpp
marthtz: Looks like it's done.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
marthtz: Looks like it's done.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
dkroenke: @ithier @manel94 i added this as open point to #574

dkroenke (approved):

iceoryx_posh/test/moduletests/test_gw_base.cpp
marthtz: Looks like it's done.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,532 @@
...
+    ASSERT_TRUE(result.has_error());
+    EXPECT_EQ(TomlGatewayConfigParseError::INCOMPLETE_CONFIGURATION, result.get_error());
+}
+
+TEST_F(TomlGatewayConfigParserTest, ParseWithoutParameterTakeDefaultPathReturnNoError)
+{
+    auto result = TomlGatewayConfigParser::parse();
+    GatewayConfig config = result.value();
+
+    EXPECT_FALSE(result.has_error());
```
marthtz, line 258: You're accessing value without checking for error first/if a value is present.
You should swap these lines and use `ASSERT_*`.
This is the case for all following tests below as well!

marthtz (changes requested):

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
marthtz, line 258: You're accessing value without checking for error first/if a value is present.
You should move the `EXPECT` before using result/value and change it to `ASSERT_*.`
This is the case for all following tests below as well!

marthtz (changes requested):

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,268 @@
...
+
+    iox::roudi::ConfigFilePathString_t Path =
+        iox::roudi::ConfigFilePathString_t(iox::cxx::TruncateToCapacity, TestFilePath);
+    auto result = TomlGatewayConfigParser::parse(Path);
+
+    ASSERT_TRUE(result.has_error());
+    EXPECT_EQ(TomlGatewayConfigParseError::INCOMPLETE_CONFIGURATION, result.get_error());
+}
+
+TEST_P(TomlGatewayConfigParser_Test, CheckCharactersUsedForServiceDescriptionInTomlConfigFile)
```
marthtz, line 258: Maybe add "Parse" to the name as well?

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -20,201 +20,268 @@
...
+const std::string TempPath = "/tmp";
+const std::string TestFilePath = TempPath + "/" + TestFile;
+#else
+const std::string TempPath = std::getenv("TEMP");
+const std::string TestFilePath = TempPath + "\\" + TestFile;
+#endif
 
 // ======================================== Fixture ======================================== //
-class TomlGatewayConfigParserTest : public Test
+class TomlGatewayConfigParser_Test : public TestWithParam<std::tuple<std::string, bool>>
```
marthtz, line 36: You could define your own type for `std::tuple<std::string, bool>` and just use the custom type in `::testing::Values` as well. Check out current master of this file!

iceoryx_posh/test/moduletests/test_gateway_gateway_base.cpp
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        ~GatewayBaseTestDestructor() noexcept
+        {
+        }
+    };
+};
+
+
+TEST_P(GatewayBase_test, InterfacePortWillBeDestroyedWhenGatewayGoesOutOfScope)
```
marthtz, line 74: Why are all gateway tests labeled as parameterized tests but don't use a parameter? They should use all kinds of interfaces as parameters, like `iox::capro::Interfaces::SOMEIP`, `iox::capro::Interfaces::INTERNAL` etc.

dkroenke:
@chiranjeevimaddi the GitHub Actions are currently facing issues with the Runners which could result in an endless loop. I canceled your jobs and let you know when you can trigger them again.

manel94:
those jobs could be triggered again?

dkroenke:
> those jobs could be triggered again?

I recommend to merge latest master in your branch to trigger the build again

elfenpiff (commented):
@chiranjeevimaddi Is the work here in this PR continued? The last commit was 12 days ago?

manel94:
> @chiranjeevimaddi Is the work here in this PR continued? The last commit was 12 days ago?

Yeah, Review is needed here

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -43,192 +56,256 @@ class TomlGatewayConfigParserTest : public TestWithParam<ParseErrorInputFile_t>
...
+                                          CheckCharactersValidity_t{std::make_tuple("Number1234567890", false)},
+                                          CheckCharactersValidity_t{std::make_tuple("Under_score_Characters", false)},
+                                          CheckCharactersValidity_t{std::make_tuple("_BeginsWithUnderscore", false)},
+                                          CheckCharactersValidity_t{std::make_tuple("Hyphen-InService", true)},
+                                          CheckCharactersValidity_t{std::make_tuple("1234567890", true)},
+                                          CheckCharactersValidity_t{std::make_tuple("這場考試_!*#:", true)}));
+
+#pragma GCC diagnostic pop
+
+TEST_P(TomlGatewayConfigParserSuiteTest, CheckCharactersUsedInServiceDescription)
```
elfenpiff, line 105: Could you please add the following test cases:
- Parse a valid config file with the maximum allowed number of configured services (MAX_GATEWAY_SERVICES)
- Parse a valid config file with more than the maximum allowed number of configured services
- Parse a valid config file where the same (service/instance/event) entry triplet is contained multiple times -> gatewayConfig should then have only one entry (every entry should be unique)

elfenpiff (changes requested):

elfenpiff:
@manel94 thanks for the hint I reviewed it and everything looks good so far, we only need some more tests and then this is good to merge.

dkroenke (dismissed):

JimmyBch:
@elfenpiff Do you mind to review one last time the PR and merged it if everything is ok for you ?

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -255,3 +332,88 @@ TEST_P(TomlGatewayConfigParserTest, ParseMalformedInputFileCausesError)
...
+    CreateTmpTomlFile(toml);
+
+    iox::roudi::ConfigFilePathString_t Path =
+    iox::roudi::ConfigFilePathString_t(iox::cxx::TruncateToCapacity, TestFilePath);
+    auto result = TomlGatewayConfigParser::parse(Path);
+    GatewayConfig config = result.value();
+
+    EXPECT_EQ(config.m_configuredServices.size(), iox::MAX_GATEWAY_SERVICES);
+    EXPECT_FALSE(result.has_error());
+    EXPECT_FALSE(config.m_configuredServices.empty());
```
elfenpiff, line 389: Could you please verify the contents here as well?

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
```diff
@@ -255,3 +332,88 @@ TEST_P(TomlGatewayConfigParserTest, ParseMalformedInputFileCausesError)
...
+        serviceEntry->insert("event", stringentry);
+        serviceArray->push_back(serviceEntry);
+    }
+
+    toml->insert("services", serviceArray);
+    CreateTmpTomlFile(toml);
+
+    iox::roudi::ConfigFilePathString_t Path =
+    iox::roudi::ConfigFilePathString_t(iox::cxx::TruncateToCapacity, TestFilePath);
+    auto result = TomlGatewayConfigParser::parse(Path);
```
elfenpiff, line 413: In this particular case, I would add the entry `MAXIMUM_NUMBER_OF_ENTRIES_EXCEEDED` into `TomlGatewayConfigParseError` enum and return it here. Otherwise, the user has no idea that two services are missing.

elfenpiff (changes requested):

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
manel94: @elfenpiff  what do you mean by return it here, you would like to check the exceeded number of entries in parse function no
?

elfenpiff: @manel94 I meant that `TomlGatewayConfigParserError` returns `MAXIMUM_NUMBER_OF_ENTRIES_EXCEEDED` when there are more entries in the config file present then the parser can handle.

iceoryx_posh/test/moduletests/test_popo_toml_gateway_config_parser.cpp
manel94: @elfenpiff Do  you mean :
 for (uint32_t i = 1U; i <= iox::MAX_GATEWAY_SERVICES; ++i)
    {
        std::string stringentry = "validservice"+ std::to_string(i);
        
        EXPECT_EQ(config.m_configuredServices.at(i).m_serviceDescription.getServiceIDString(), stringentry);
        EXPECT_EQ(config.m_configuredServices.at(i).m_serviceDescription.getInstanceIDString(), stringentry);
        EXPECT_EQ(config.m_configuredServices.at(i).m_serviceDescription.getEventIDString(), stringentry);
    }

elfenpiff: @manel94 yes. You could use a ranged-based loop here, this would make the code more readable.

manel94:
Could someone help with those failed checks?

dkroenke:
> Could someone help with those failed checks?
@manel94 
Please merge the latest master into your branch to resolve the issue.
Related: https://github.com/eclipse-iceoryx/iceoryx/issues/825

manel94:
Can someone help me why param tests failed on iceoryx windows?

manel94:
Hi, could someone help me on this error related to iceoryx-windows build,  "unknown file: error: SEH exception with code 0xc0000005 thrown in the test body."

elfenpiff:
> Hi, could someone help me on this error related to iceoryx-windows build, "unknown file: error: SEH exception with code 0xc0000005 thrown in the test body."

@manel94 usually this error comes up when you access an invalid memory location in windows. Could you try to build iceoryx in visual studio and debug the origin of this error?
If this not an option could you analyse one single test which causes such an SEH exception and look for an out of bounds access of an array? Or run one single test case in linux with valgrind?

manel94:
> > Hi, could someone help me on this error related to iceoryx-windows build, "unknown file: error: SEH exception with code 0xc0000005 thrown in the test body."
> 
> @manel94 usually this error comes up when you access an invalid memory location in windows. Could you try to build iceoryx in visual studio and debug the origin of this error?
> If this not an option could you analyse one single test which causes such an SEH exception and look for an out of bounds access of an array? Or run one single test case in linux with valgrind?

@elfenpiff  after debug i find that this windows build problem is the result of iox::config::TomlGatewayConfigParser::parse(iox::cxx::string<1024ul> const&) (toml_gateway_config_parser.cpp:31) called in all my added tests, but i can´t really fix it, any proposal?

manel94:
windows build problem fixed.

dkroenke (approved):

dkroenke:
@elfenpiff any further objections in this PR?

elfenpiff (approved):

PR #580, opened 2021-02-25
Iox #408 transition to 1 0 c api

### LINKED ISSUE #349: Introduction of Runnables/Nodes for the new APIs

### LINKED ISSUE #408: Refine modern pub/sub API

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

1. [x] Merge

## References

- Closes #408
- Closes #349 

## Summary
- Renaming in C-API as planned
- Adapted examples and tests as needed
- Added function to obtain chunk header in C-API as agreed on (not too helpful since it is a C++ structure...requires further considerations)
- optimized the expected of optional in the middle layer (ChunkReceiver, Port) on subscriber side (including test adaptation). This was done for consistency with the lower layer and efficiency.

## Open Points
- Provide the ability to use the chunk header in C
- Minimal tests for the header are currently done on publisher/subscriber side (1. I consider it needed there to check whether we can get the header, 2. the current minimal functionality seems not to warrant an individual test file, we cannot test much due to the limited functionality in C)
- Do we want a method to release chunks at the untyped publisher (e.g. when the publisher app decides not to publish the allocated sample for whatever reason)?
- proposal: deignated using payload_t = void* to make the interface more clear (whteher chunk header or payload is expected may be not clear enough in some cases). This cannot fully enforce type safety (especially in C) but may improve readability.


iceoryx_binding_c/include/iceoryx_binding_c/chunk.h
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
```
MatthiasKillat, line 1: Please not that I considered naming it chunkheader but 1) that then the prefixes for the functions would sound weird and 2) I imagine we could add additional functionality concerning chunks here. But we could still split later into e.g. chunk.h and chunk_header.h if we deem it necessary (depends on the functionality we want to add).

iceoryx_posh/test/moduletests/test_popo_base_subscriber.cpp
```diff
@@ -146,20 +145,6 @@ TEST_F(BaseSubscriberTest, ReceiveForwardsErrorsFromUnderlyingPort)
     // ===== Cleanup ===== //
 }
```
MatthiasKillat, line 147: Test not needed anymore. The propagation of the "error" type is tested elsewhere. Adapting this test would essentially replicate this as we now propagate NO_CHUNK_AVAILABLE.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +24,29 @@
 /// @brief publisher handle
 typedef struct cpp2c_Publisher* iox_pub_t;
 
+/// @brief options to be set for a publisher
```
elfenpiff, line 27: Please adjust the copyright header. One company per line.

iceoryx_binding_c/include/iceoryx_binding_c/enums.h
```diff
@@ -38,7 +38,7 @@ enum iox_SubscribeState
 enum iox_ChunkReceiveResult
 {
     ChunkReceiveResult_TOO_MANY_CHUNKS_HELD_IN_PARALLEL,
```
elfenpiff, line 40: Please adjust the copyright header. One company per line.

iceoryx_binding_c/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ add_library(${PROJECT_NAME}
     source/c_subscriber.cpp
     source/c_event_info.cpp
     source/c_wait_set.cpp
```
elfenpiff, line 55: Please adjust the copyright header. One company per line.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -34,21 +34,20 @@ iox_pub_t iox_pub_init(iox_pub_storage_t* self,
                        const char* const service,
                        const char* const instance,
                        const char* const event,
```
elfenpiff, line 36: Please adjust the copyright header. One company per line.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -34,21 +34,20 @@ iox_pub_t iox_pub_init(iox_pub_storage_t* self,
...
+                       const iox_publisher_options_t options)
 {
     new (self) cpp2c_Publisher();
     iox_pub_t me = reinterpret_cast<iox_pub_t>(self);
-    PublisherOptions options;
-    options.historyCapacity = historyCapacity;
-    options.nodeName = NodeName_t(TruncateToCapacity, nodeName);
+    PublisherOptions publisherOptions;
+    publisherOptions.historyCapacity = options.historyCapacity;
+    publisherOptions.nodeName = NodeName_t(TruncateToCapacity, options.nodeName);
```
elfenpiff, line 43: Maybe we should print a warning here when `options.nodeName` is longer then the `NodeName_t` capacity so that the user is at least informed and does not report a bug that his node name is truncated. Or we follow the array approach.

iceoryx_binding_c/source/c_subscriber.cpp
```diff
@@ -39,21 +40,19 @@ iox_sub_t iox_sub_init(iox_sub_storage_t* self,
...
     new (self) cpp2c_Subscriber();
     iox_sub_t me = reinterpret_cast<iox_sub_t>(self);
-    SubscriberOptions options;
-    options.queueCapacity = queueCapacity;
-    options.historyRequest = historyRequest;
-    options.nodeName = NodeName_t(TruncateToCapacity, nodeName);
+    SubscriberOptions subscriberOptions;
+    subscriberOptions.queueCapacity = options.queueCapacity;
+    subscriberOptions.historyRequest = options.historyRequest;
+    subscriberOptions.nodeName = NodeName_t(TruncateToCapacity, options.nodeName);
```
elfenpiff, line 50: Maybe we should print a warning here when options.nodeName is longer then the NodeName_t capacity so that the user is at least informed and does not report a bug that his node name is truncated.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +24,29 @@
...
+typedef struct
+{
+    // size of the history chunk queue
+    uint64_t historyCapacity;
+
+    // name of the node the publisher belongs to
+    const char* nodeName;
```
elBoberido, line 34: how about using a `char nodeName[100]`?

On the other hand one cannot assign `const char*` like `nodeName = "foo";` anymore.

elfenpiff (changes requested):
Please check the copyright header again for all files and adapt it if necessary.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +24,29 @@
...
+/// @param[in] options publisher options set by the user
 /// @return handle of the publisher
+/// @attention the options must be set to valid values
 iox_pub_t iox_pub_init(iox_pub_storage_t* self,
                        const char* const service,
                        const char* const instance,
                        const char* const event,
-                       const uint64_t historyCapacity,
-                       const char* const nodeName);
+                       const iox_publisher_options_t options);
```
elBoberido, line 49: I always thought one cannot pass structs to C functions or is it only regarding returning structs?

Nevertheless, if a pointer is passed, it could be interpreted as using the default options. This would somehow mimic the C++ behavior.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +24,29 @@
...
 
+/// @brief options to be set for a publisher
+typedef struct
+{
+    // size of the history chunk queue
+    uint64_t historyCapacity;
+
+    // name of the node the publisher belongs to
+    const char* nodeName;
+} iox_publisher_options_t;
```
elBoberido, line 35: please create a `iox_pub_init_options` method to initialize all members to default values. This enables us to add additional members to the struct without breaking the code of our users. At least not when used correctly.

iceoryx_examples/icedelivery_in_c/README.md
```diff
@@ -46,11 +46,13 @@ Let's take a look at the `receiving` function which comes with the
     The `subscriberStorage` is the place where the subscriber is stored in
     memory and `subscriber` is actually a pointer to that location.
     ```c
-    const uint64_t historyRequest = 10U;
-    const uint64_t queueCapacity = 5U;
-    const char* const nodeName = "iox-c-subscriber-node";
+    iox_subscriber_options_t options;
```
elBoberido, line 49: here the `iox_sub_init_options` should be called first

iceoryx_examples/waitset_in_c/ice_c_waitset_individual.c
```diff
@@ -61,16 +61,19 @@ int main()
...
-    const char* const nodeName1 = "iox-c-ex-waitset-individual-node1";
-    const char* const nodeName2 = "iox-c-ex-waitset-individual-node2";
+    iox_subscriber_options_t options1, options2;
+    options1.historyRequest = 1U;
+    options1.queueCapacity = 256U;
+    options1.nodeName = "iox-c-ex-waitset-individual-node1";
+    
+    options2.historyRequest = 1U;
+    options2.queueCapacity = 256U;
+    options2.nodeName = "iox-c-ex-waitset-individual-node2";
```
elBoberido, line 71: could `option1` be reused by just setting `nodeName` after `iox_sub_init`

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
elfenpiff: Yes but you can still use a one liner like `strncpy(nodeName, "foo",  100);`.
You are right this is not the most beautiful way but we are working here also in a safety context ...

iceoryx_binding_c/source/c_publisher.cpp
elBoberido: isn't `cxx::string` already printing a warning when the string is truncated?

elfenpiff: I think so but the warning is somehow out of context since the user does not know that we are using a `cxx::string` behind the scenes. It is better then nothing but from a user point of view I would wish to get as much information as possible so that I can fix the problem.

An error message like: `publisher node name to long, truncate it to: "truncatedNodeName"` is very informative and would me help a lot!

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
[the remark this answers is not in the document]
elBoberido: I also thought about this, but we pay with ease of use and make it potentially more error prone because `nodeName = "foo";`  is not possible anymore and `strncpy` or `strcpy` must be used. So I would opt in for the ease of use and catch too long names internally.

elBoberido: yes, but it's more error prone like `strncpy(myNodeName, options.nodeName, 100);` and then needing to debug the code because of this mistake that would not be possible if it would remain a `const char*`

iceoryx_binding_c/test/moduletests/test_cpp2c_enum_translation.cpp
```diff
@@ -38,6 +38,9 @@ TEST(cpp2c_enum_translation_test, SubscribeState)
 
 TEST(cpp2c_enum_translation_test, ChunkReceiveResult)
 {
+    EXPECT_EQ(cpp2c::ChunkReceiveResult(iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE),
```
FerdinandSpitzschnueffler, line 41: Please update the copyright header.

iceoryx_binding_c/source/c_publisher.cpp
dkroenke: The cxx string prints the input parameter when giving a warning that it will be truncated.
```bash
Constructor truncates the last 1 characters of TooLongNodeName, because the char array length is larger than the capacity.
```

iceoryx_binding_c/include/iceoryx_binding_c/chunk.h
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IOX_BINDING_C_CHUNK_H
+#define IOX_BINDING_C_CHUNK_H
+
+#include "iceoryx_binding_c/internal/c2cpp_binding.h"
+
+/// @note: currently we can only use this header safely with the corresponding
```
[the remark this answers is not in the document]
MatthiasKillat: Sure, I this is for discussion when/how we want to solve this and to raise awareness.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +24,29 @@
...
 /// @brief creates a publisher handle
 /// @param[in] self pointer to preallocated memory of size = sizeof(iox_pub_storage_t)
 /// @param[in] service serviceString
 /// @param[in] instance instanceString
 /// @param[in] event eventString
-/// @param[in] historyCapacity size of the history chunk queue
-/// @param[in] nodeName name of the node the publisher belongs to
+/// @param[in] options publisher options set by the user
 /// @return handle of the publisher
+/// @attention the options must be set to valid values
```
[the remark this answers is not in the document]
MatthiasKillat: Can do, but it is easier to forget this in case of structs than an int variable where you a least have to pass something which you hopefully initialized... It is easy to forget to initialize a struct member in C.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +24,29 @@
 /// @brief publisher handle
 typedef struct cpp2c_Publisher* iox_pub_t;
 
+/// @brief options to be set for a publisher
+typedef struct
```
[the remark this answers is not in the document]
MatthiasKillat: This is done because of C compilers which follow the standard very strictly to avoid having to write ``struct iox_publisher_options_t``. I would recommend a typedef for this reason, but in C++ no typedef is needed (makes no difference).

See https://stackoverflow.com/questions/612328/difference-between-struct-and-typedef-struct-in-c

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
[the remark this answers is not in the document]
MatthiasKillat: I am not in favor of arrays with a fixed length here because it is more cumbersome to use and does not really buy you anything regarding securing the code. Internally it is truncated anyway.

iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
```diff
@@ -24,22 +24,32 @@
 /// @brief Subscriber handle
 typedef struct cpp2c_Subscriber* iox_sub_t;
 
+/// @brief options to be set for a subscriber
+typedef struct
```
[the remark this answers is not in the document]
MatthiasKillat: Cf. ``iox_subscriber_options_t``.

iceoryx_binding_c/source/c_publisher.cpp
MatthiasKillat: Yes but we could argue this for a lot of parameters and would need a systematic approach on validation (is the historycapacity valid? Too large? Do we truncate or fail? How do we inform the user). I would not do a check here in addition to the cxx::string since it I consider the additional value low.

iceoryx_binding_c/source/c_subscriber.cpp
MatthiasKillat: cxx::string does this. As stated above I would not add the additional check.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
MatthiasKillat: Rejected. Usability suffers for not much (if any) gain.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
MatthiasKillat: Can do. We cannot force anyone to use it though.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
MatthiasKillat: You can pass it by value or as pointer.

But If I understand you right you want nullptr to indicate default options? (It is not clear from what you wrote).

iceoryx_examples/waitset_in_c/ice_c_waitset_individual.c
MatthiasKillat: Yes but I did not want to confuse the user.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
MatthiasKillat: Shroten to iox_pub_options_t

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -31,17 +31,28 @@ extern "C" {
...
                        const char* const event,
-                       const iox_pub_options_t options)
+                       const iox_pub_options_t* const options)
 {
     new (self) cpp2c_Publisher();
     iox_pub_t me = reinterpret_cast<iox_pub_t>(self);
     PublisherOptions publisherOptions;
-    publisherOptions.historyCapacity = options.historyCapacity;
-    publisherOptions.nodeName = NodeName_t(TruncateToCapacity, options.nodeName);
+    publisherOptions.historyCapacity = options->historyCapacity;
```
elBoberido, line 50: please check if `options` is a nullptr before accessing it

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -31,17 +31,28 @@ extern "C" {
 #include "iceoryx_binding_c/publisher.h"
 }
 
+void iox_pub_options_init(iox_pub_options_t* options)
+{
+    PublisherOptions publisherOptions;
+    options->historyCapacity = publisherOptions.historyCapacity;
```
elBoberido, line 37: please check if `options` is a nullptr before accessing it

elBoberido (commented):
documentation not yet updated, but I guess that's just because it's currently WIP

iceoryx_binding_c/source/c_publisher.cpp
MatthiasKillat: I thought about it and can do it but in the end I can get an invalid non-nullptr.

elBoberido: I think just printing a warning/error and continue. If it's a nullptr, it will crash later when the user access it on the caller site, but at least there is some output on the console.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
```diff
@@ -59,6 +59,14 @@ class UntypedPublisherImpl : public base_publisher_t
     ///
     void publish(const void* chunk) noexcept;
 
+    ///
+    /// @brief Releases the chunk provided by its payload pointer.
```
elBoberido, line 63: maybe add that the ownership is released and that the chunk must not be accessed afterwards

elBoberido (commented):
What do you think about adding a `sanityCheck` member to the option structs. They could be initialized to a specific value with the init function and `iox_pub_init` and `iox_sub_init` could check if the provided options have this values. It's not 100% safe, but at least we could catch some cases where the options are not initialized correctly

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +25,33 @@
...
+    // size of the history chunk queue
+    uint64_t historyCapacity;
+
+    // name of the node the publisher belongs to
+    // nullptr indicates that the default node name is used
+    const char* nodeName;
+} iox_pub_options_t;
+
+/// @brief initialize publisher options to default values
+/// @param[in] options pointer to options to be initialized
```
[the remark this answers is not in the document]
MatthiasKillat: Sanity check makes sense. But what shall happen if it is not set? An output on std::err? Termination? The apocalypse? ... Questions, questions...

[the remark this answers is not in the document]
MatthiasKillat: On second thought, the sanity check variable is problematic since it must be public and therefore can confuse the user? What shall he do with it? I do not think it is worth it, wasted effort for not much gain.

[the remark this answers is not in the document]
MatthiasKillat: Use LogWarn now for uninitialized options.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +25,46 @@
...
+{
+    // size of the history chunk queue
+    uint64_t historyCapacity;
+
+    // name of the node the publisher belongs to
+    // nullptr indicates that the default node name is used
+    const char* nodeName;
+
+    // this value will be set exclusively by iox_pub_options_init
+    // and is not supposed to modified otherwise
```
elBoberido, line 39: to **be** modified?

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_subscriber.inl
```diff
@@ -41,9 +41,9 @@ inline cxx::expected<const void*, ChunkReceiveResult> UntypedSubscriberImpl<base
 }
 
 template <typename base_subscriber_t>
-inline void UntypedSubscriberImpl<base_subscriber_t>::releaseChunk(const void* payload) noexcept
+inline void UntypedSubscriberImpl<base_subscriber_t>::releaseChunk(const void* chunk) noexcept
```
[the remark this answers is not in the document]
MatthiasKillat: I favor the alias Payload, yes. The renaming was done for consistency with the publisher...
The method name should stay ``releaseChunk`` I think but I guess it depends on all the other names in the API.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -30,25 +31,63 @@ extern "C" {
...
+    options->initCheck = PUBLISHER_OPTIONS_INIT_CHECK_CONSTANT;
+}
+
+bool iox_pub_options_is_initialized(const iox_pub_options_t* const options)
+{
+    if (options && options->initCheck == PUBLISHER_OPTIONS_INIT_CHECK_CONSTANT)
+    {
+        return true;
+    }
+    return false;
```
elBoberido, line 57: I've seen this in Walter E. Browns Meeting C++ 2019 talk "multiple ways to return true/false" ;)

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -30,25 +31,63 @@ extern "C" {
 #include "iceoryx_binding_c/publisher.h"
 }
 
+static uint64_t PUBLISHER_OPTIONS_INIT_CHECK_CONSTANT = 123454321;
```
elBoberido, line 34: I expected more creativity https://en.wikipedia.org/wiki/Hexspeak
How about `0xDECAFCOFFEE` or `0xDEADBEEF` or `0xBADFOOD`

iceoryx_binding_c/source/c_publisher.cpp
MatthiasKillat: Not exactly but yes, can be shortened. There are other points in the code where I opted not to use e.g. the ternary operator for readability.

iceoryx_binding_c/source/c_publisher.cpp
MatthiasKillat: IDC.

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_subscriber.inl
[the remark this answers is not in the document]
MatthiasKillat: Renamed function to release, no payload alias.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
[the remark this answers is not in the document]
MatthiasKillat: Sure, I would not terminate though but can add this as well. The reason is: if the user by accident or on purpose, sets all fields to a reasonable value, it will work.

We can of course not detect errors in ll cases, i.e. may have false negatives (but no false positives, so we would never terminate in a case were init was called unless the value was overwritten afterwards).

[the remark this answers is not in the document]
elBoberido: The Problem is not that it might not work right now but that it works because everything is set manually and the user ignores the warning. If we add another option in half a year, there might be a weird behavior and the connection of the weird behavior to the warning might not be there, because it worked for half a year. So this is just to be future proof and force to user to write correct code as soon as possible.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -30,25 +31,63 @@ extern "C" {
...
+    // use default options otherwise
+    if (options != nullptr)
+    {
+        if (!iox_pub_options_is_initialized(options))
+        {
+            std::cout << "******* terminate" << std::endl;
+            // note that they may have been initialized but the initCheck
+            // pattern overwritten afterwards, we cannot be sure but it is a misuse
+            LogError() << "publisher options may not have been initialized with iox_pub_init";
+            std::terminate();
```
elBoberido, line 75: `LogFatal` should be used here and I guess also the error handler according to the error handling document

MatthiasKillat: Done. But Note that we have some more terminates in the C Binding and I am not sure whether we want the error handling rules also for Bindings, Gateways and so on but only the core. 
E.g. in utils we also have a different approach (since it is a generic library).

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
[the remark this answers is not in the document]
MatthiasKillat: I understand this. Just wanted to make clear it is not entirely foolproof. And there are a lot of fools everywhere. ;-)

elBoberido (dismissed):

iceoryx_binding_c/test/moduletests/test_subscriber.cpp
```diff
@@ -183,22 +186,42 @@ TEST_F(iox_sub_test, receiveChunkWithContent)
...
 
-    ASSERT_EQ(iox_sub_get_chunk(m_sut, &chunk), ChunkReceiveResult_SUCCESS);
+    ASSERT_EQ(iox_sub_take_chunk(m_sut, &chunk), ChunkReceiveResult_SUCCESS);
     EXPECT_THAT(static_cast<const data_t*>(chunk)->value, Eq(1234));
 }
 
+TEST_F(iox_sub_test, chunkHeaderCanBeObtainedFromChunkAfterTake)
+{
+    this->Subscribe(&m_portPtr);
+    struct data_t
```
FerdinandSpitzschnueffler, line 196: This struct is not used.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -24,20 +25,46 @@
...
+    const char* nodeName;
+
+    // this value will be set exclusively by iox_pub_options_init
+    // and is not supposed to be modified otherwise
+    uint64_t initCheck;
+} iox_pub_options_t;
+
+/// @brief initialize publisher options to default values
+/// @param[in] options pointer to options to be initialized,
+///                    if it is a null pointer default options are used
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, mixed up.

iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
```diff
@@ -24,22 +25,50 @@
...
+{
+    // size of the history chunk queue
+    uint64_t queueCapacity;
```
dkroenke, line 32: The struct members should be default initialized with zero or nullptr. Applies also for all other structs in this PR.

dkroenke (changes requested):
The struct members should be default initialized with zero or nullptr. Applies also for all other structs in this PR.

iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
FerdinandSpitzschnueffler: Unfortunately not possible in C. This is why @MatthiasKillat implemented the *_options_init methods.

MatthiasKillat: Impossible in C.

FerdinandSpitzschnueffler (approved):

FerdinandSpitzschnueffler (approved):

dkroenke (approved):

elfenpiff (approved):

PR #597, opened 2021-03-09
Iox #495 bugfix gcc 5 compatibility

### LINKED ISSUE #495: fail build on ubuntu 16.04 with g++ 5.4.0
## Required information

**Operating system:**
PRETTY_NAME="Ubuntu 16.04.7 LTS"

**Compiler version:**
Target: x86_64-linux-gnu
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.12)

See attached log file [compiler_log_BH1SCW.txt](https://github.com/eclipse-iceoryx/iceoryx/files/6090535/compiler_log_BH1SCW.txt)

build with no error 

git commit id:
current master (original f1c4b31c54cc7fccb352fc3cdda562d9a22fe638)

---

Issues:
- [ ] Unsupported compiler command line option `-Wno-noexcept-type`
- [ ] Call to ostream from constexpr function; basically all `clog` calls in `duration.inl`; added in PR #501
   Error: `call to non-constexpr function ‘std::basic_ostream`

marthtz:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
Enable build on gcc 5, including QNX SDP 7.0
Remaining issue tracked in #607.

## References

- Partially closes #495
- Potential follow-up #609
- Remaining issue tracked in #607


iceoryx_utils/include/iceoryx_utils/internal/units/duration.inl
```diff
@@ -31,8 +31,8 @@ inline constexpr Duration::Duration(const Seconds_t seconds, const Nanoseconds_t
         auto additionalSeconds = nanoseconds / NANOSECS_PER_SEC;
         if (std::numeric_limits<Seconds_t>::max() - additionalSeconds < m_seconds)
         {
-            std::clog << __PRETTY_FUNCTION__
-                      << ": Applied values are out of range and would overflow, clamping to max value!" << std::endl;
+            // std::clog << __PRETTY_FUNCTION__
```
elfenpiff, line 34: Oh this is a deep and philosophical. Either we inform the user that something weird is going on but do not have `constexpr` compile time optimization on our hand or we get rid of all those warning messages and have compile time optimization again.

Since we have in every method defined behavior and the behavior is also so that no one should be surprised I tend to get rid of `clog` completely.

@MatthiasKillat what do you think?

marthtz: > I tend to get rid of `clog` completely.

Thumbs up from me!
If we still want to give active feedback we could just let it fail with an assert. This would force the user to handle overflows/negative values and not the duration class.

dkroenke: I would add a static_assert here for the debug mode and make in doxygen clear that values are clamped to maximum. The clog can be removed. Same for the other occurences with clog in constexpr functions.

elfenpiff: @dkroenke sadly you cannot use `static_assert` in this context since you require potential runtime variable knowledge. The best thing would be the usage of `cxx::Expects` and `cxx::Ensures` from `cxx/helplets.hpp`.
They will perform a `std::terminate` if the condition is violated with a detailed error message for the user.

dkroenke: @elfenpiff i agree, but i would still let it only terminate in debug mode because clamping a value is imho something which does not need to terminate in release mode

elfenpiff: @dkroenke I completely agree! But I would still use `cxx::Expects` and `cxx::Ensures` and modify there behavior to the one you suggested. If you are in debug mode you will be terminated here with a detailed error message and in release mode we continue.

But maybe we need to differentiate here a bit. If this is a behavior which is somehow allowed in Release mode and we terminate here only to provide a hint for the developer we maybe should add a parameter to expects and ensures stating that.

```cpp
cxx::Expects( myNumber > 0 ); // hard, calls std::terminate always
cxx::Expects( myNumber > 0, ALGORITHM_CAN_CONTINUE); // with this enum we state that without this 
// requirement fulfilled we can still continue working
```

This would be something for a different pull request but what do you think @dkroenke @elBoberido @marthtz @mossmaurice @budrus

mossmaurice: I agree with you, modifying `Expects` and `Ensures` depending on release or debug mode is a great solution as trade-off between safety and performance. However, I wouldn't make it more difficult by introducing `ALGORITHM_CAN_CONTINUE`. If utils/hoofs will be used in a broader context, I can predict what will happen: People will always use `ALGORITHM_CAN_CONTINUE`, basically ignoring it and then there will be reports of error in the field because someone forgot to remove it or was aware there even exists another version.. somehow [Hyrum's Law](https://www.hyrumslaw.com/). It's better to keep things simple and stupid.

marthtz: @elfenpiff unfortunately, we can't use `cxx::Expects`/`cxx::Ensures` either. They resolve to `iox::cxx::internal::Require` which is a non-constexpr function.
We also can't use `static_assert` as mentioned before, so I guess we're out of options and for now just remove all `clog`s?

marthtz: Guys, I created #607 to track the suggested changes to `cxx::Expects` and `cxx::Ensures`.
I'll replace all `clog`s now with `/// @todo #607 issue warning or fail`.

mossmaurice (approved):
LGTM, could you add a section QNX in `doc/website/getting-started/installation.md` and mention e.g. the compiler version, steps, et.c there?

marthtz:
> 
> 
> LGTM, could you add a section QNX in `doc/website/getting-started/installation.md` and mention e.g. the compiler version, steps, et.c there?

Added inf to use a toolchain file for QNX.

doc/website/getting-started/installation.md
```diff
@@ -54,6 +53,16 @@ sudo apt install gcc g++ cmake libacl1-dev libncurses5-dev pkg-config
 
 Additionally, there is an optional dependency to the MIT licensed [cpptoml](https://github.com/skystrife/cpptoml) library, which is used to parse a RouDi config file for the mempool config.
 
+### QNX
+
+QNX SDP 7.0 and 7.1 are supported (shipping with gcc 5.4 and gcc 8.3 respectively).
```
mossmaurice, line 58: Is it verfied that QCC/GCC 8.3 builds iceoryx? Otherwise I'd just mention 5.4.

We could add QNX 7.1 support for v1.x or even backport it v1.0 if needed.

doc/website/getting-started/installation.md
```diff
@@ -9,26 +9,25 @@ iceoryx_utils and iceoryx_posh are deployed as independent cmake packages. Posh
...
    - [msvc](https://visualstudio.microsoft.com/de/), part of Visual Studio 2019 or later
  - [libacl](http://download.savannah.gnu.org/releases/acl/), 2.2 or later. Only for Linux & QNX.
- - optional, [ncurses](https://invisible-island.net/ncurses/), 6.2 or later. Required by introspectiont tool.
+ - optional, [ncurses](https://invisible-island.net/ncurses/), 6.2 or later. Required by introspection tool.
 
 #### Optional, Cyclone DDS Gateway
-If you would like to use our Cyclone DDS Gateway you have to install 
-[Cyclone DDS](https://github.com/eclipse-cyclonedds/cyclonedds) first. Furthermore
-you have to install:
+If you would like to use our Cyclone DDS Gateway you have to install [Cyclone DDS](https://github.com/eclipse-cyclonedds/cyclonedds) first.<br>
```
mossmaurice, line 19: In general `<br>` shouldn't be necessary. Just add a extra new line.

marthtz: Are you sure? It doesn't work for me. The markdown guide mentions to use either `<br>` or two spaces.

doc/website/getting-started/installation.md
marthtz: Hoped to get confirmation here 😉 
7.1 removed!

dkroenke: @mossmaurice yes that is verified

mossmaurice (approved):

dkroenke (approved):

doc/website/getting-started/installation.md
```diff
@@ -54,6 +53,16 @@ sudo apt install gcc g++ cmake libacl1-dev libncurses5-dev pkg-config
...
 Additionally, there is an optional dependency to the MIT licensed [cpptoml](https://github.com/skystrife/cpptoml) library, which is used to parse a RouDi config file for the mempool config.
 
+### QNX
+
+QNX SDP 7.0 and 7.1 are supported (shipping with gcc 5.4 and gcc 8.3 respectively).
+
+Easiest way to build iceoryx on QNX is using the build script and providing a toolchain file.<br>
+Example:
+```
+./tools/iceoryx_build_test.sh -t /home/user/toolchains/qnx/qcc_x86_64_toolchain.cmake
```
elBoberido, line 63: could we add a toolchain file to iceoryx? Doesn't seem to be too complicated
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-qnx
https://cristianadam.eu/20181202/a-better-qnx-cmake-toolchain-file/

marthtz: +1
Also see #609.

PR #614, opened 2021-03-16
Iox #14 add payload alignment and custom header to memory manager

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

Another PR on the path to the custom chunk header. This is the preparation to add the custom header to the typed API. There is already a branch where the code from this PR is integrated. So after this one is merged, there will be immediately a PR for that one.

In this PR, the `ChunkHeader` ctor is extended to take parameters for the custom header and payload alignment. Additionally the `MemoryManager` is adjusted to calculate the correct size. Most of the changes are adjustments for the changes in `ChunkHeader` and `MemoryManager` and the corresponding tests.

## References

- Partially closes #14

### DIFF
--- a/iceoryx_binding_c/source/c_publisher.cpp
+++ b/iceoryx_binding_c/source/c_publisher.cpp
@@ -103,7 +103,7 @@ iox_AllocationResult iox_pub_loan_chunk(iox_pub_t const self, void** const chunk
 {
     auto result = PublisherPortUser(self->m_portData)
                       .tryAllocateChunk(payloadSize, CHUNK_DEFAULT_PAYLOAD_ALIGNMENT)
-                      .and_then([&](ChunkHeader* h) { *chunk = h->payload(); });
+                      .and_then([&chunk](ChunkHeader* h) { *chunk = h->payload(); });
     if (result.has_error())
     {
         return cpp2c::AllocationResult(result.get_error());
--- a/iceoryx_binding_c/source/c_publisher.cpp
+++ b/iceoryx_binding_c/source/c_publisher.cpp
@@ -101,9 +101,9 @@ void iox_pub_deinit(iox_pub_t const self)
 
 iox_AllocationResult iox_pub_loan_chunk(iox_pub_t const self, void** const chunk, const uint32_t payloadSize)
 {
-    auto result = PublisherPortUser(self->m_portData).tryAllocateChunk(payloadSize).and_then([&](ChunkHeader* h) {
-        *chunk = h->payload();
-    });
+    auto result = PublisherPortUser(self->m_portData)
+                      .tryAllocateChunk(payloadSize, CHUNK_DEFAULT_PAYLOAD_ALIGNMENT)
+                      .and_then([&](ChunkHeader* h) { *chunk = h->payload(); });
     if (result.has_error())
     {
         return cpp2c::AllocationResult(result.get_error());
@@ -155,4 +155,3 @@ iox_service_description_t iox_pub_get_service_description(iox_pub_t const self)
 {
     return TranslateServiceDescription(PublisherPortUser(self->m_portData).getCaProServiceDescription());
 }
-
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
@@ -41,44 +41,44 @@ class MemoryManager
     using MaxChunkPayloadSize_t = cxx::range<uint32_t, 1, std::numeric_limits<uint32_t>::max() - sizeof(ChunkHeader)>;
 
   public:
-    MemoryManager() = default;
+    MemoryManager() noexcept = default;
     MemoryManager(const MemoryManager&) = delete;
     MemoryManager(MemoryManager&&) = delete;
     MemoryManager& operator=(const MemoryManager&) = delete;
     MemoryManager& operator=(MemoryManager&&) = delete;
-    ~MemoryManager() = default;
+    ~MemoryManager() noexcept = default;
 
     void configureMemoryManager(const MePooConfig& f_mePooConfig,
                                 posix::Allocator* f_managementAllocator,
-                                posix::Allocator* f_payloadAllocator);
+                                posix::Allocator* f_payloadAllocator) noexcept;
 
     SharedChunk getChunk(const uint32_t payloadSize,
                          const uint32_t payloadAlignment,
                          const uint32_t customHeaderSize,
-                         const uint32_t customHeaderAlignment);
+                         const uint32_t customHeaderAlignment) noexcept;
 
-    uint32_t getNumberOfMemPools() const;
+    uint32_t getNumberOfMemPools() const noexcept;
 
-    MemPoolInfo getMemPoolInfo(uint32_t f_index) const;
+    MemPoolInfo getMemPoolInfo(uint32_t f_index) const noexcept;
 
     static uint32_t requiredChunkSize(const uint32_t payloadSize,
                                       const uint32_t payloadAlignment,
                                       const uint32_t customHeaderSize,
-                                      const uint32_t customHeaderAlignment);
+                                      const uint32_t customHeaderAlignment) noexcept;
 
-    static uint64_t requiredChunkMemorySize(const MePooConfig& f_mePooConfig);
-    static uint64_t requiredManagementMemorySize(const MePooConfig& f_mePooConfig);
-    static uint64_t requiredFullMemorySize(const MePooConfig& f_mePooConfig);
+    static uint64_t requiredChunkMemorySize(const MePooConfig& f_mePooConfig) noexcept;
+    static uint64_t requiredManagementMemorySize(const MePooConfig& f_mePooConfig) noexcept;
+    static uint64_t requiredFullMemorySize(const MePooConfig& f_mePooConfig) noexcept;
 
   private:
-    static uint32_t sizeWithChunkHeaderStruct(const MaxChunkPayloadSize_t size);
+    static uint32_t sizeWithChunkHeaderStruct(const MaxChunkPayloadSize_t size) noexcept;
 
-    void printMemPoolVector() const;
+    void printMemPoolVector() const noexcept;
     void addMemPool(posix::Allocator* f_managementAllocator,
                     posix::Allocator* f_payloadAllocator,
                     const cxx::greater_or_equal<uint32_t, MemPool::MEMORY_ALIGNMENT> f_payloadSize,
-                    const cxx::greater_or_equal<uint32_t, 1> f_numberOfChunks);
-    void generateChunkManagementPool(posix::Allocator* f_managementAllocator);
+                    const cxx::greater_or_equal<uint32_t, 1> f_numberOfChunks) noexcept;
+    void generateChunkManagementPool(posix::Allocator* f_managementAllocator) noexcept;
 
   private:
     bool m_denyAddMemPool{false};
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
@@ -61,6 +61,14 @@ class MemoryManager
 
     MemPoolInfo getMemPoolInfo(uint32_t f_index) const noexcept;
 
+    /// @brief calculates the required chunk size regarding the constraints on payload and custom header
+    /// @param[in] payloadSize is the size of the user payload without additional headers
+    /// @param[in] payloadAlignment is the alignment of the user payload
+    /// @param[in] customHeaderSize is the size of the custom user header; use iox::CHUNK_NO_CUSTOM_HEADER_SIZE to omit
+    /// a custom header
+    /// @param[in] customHeaderAlignment is the alignment of the custom user header; use
+    /// iox::CHUNK_NO_CUSTOM_HEADER_ALIGNMENT to omit a custom header
+    /// @return the minimum chunk size required to hold the user payload with the custom header
     static uint32_t requiredChunkSize(const uint32_t payloadSize,
                                       const uint32_t payloadAlignment,
                                       const uint32_t customHeaderSize,
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
@@ -1,4 +1,5 @@
-// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2019 - 2020 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -51,7 +52,10 @@ class MemoryManager
                                 posix::Allocator* f_managementAllocator,
                                 posix::Allocator* f_payloadAllocator);
 
-    SharedChunk getChunk(const MaxSize_t f_size);
+    SharedChunk getChunk(const MaxSize_t f_size,
+                         const uint32_t payloadAlignment,
+                         const uint32_t customHeaderSize,
+                         const uint32_t customHeaderAlignment);
 
     uint32_t getMempoolChunkSizeForPayloadSize(const uint32_t f_size) const;
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
@@ -38,7 +38,7 @@ struct MePooConfig;
 
 class MemoryManager
 {
-    using MaxSize_t = cxx::range<uint32_t, 1, std::numeric_limits<uint32_t>::max() - sizeof(ChunkHeader)>;
+    using MaxChunkPayloadSize_t = cxx::range<uint32_t, 1, std::numeric_limits<uint32_t>::max() - sizeof(ChunkHeader)>;
 
   public:
     MemoryManager() = default;
@@ -52,7 +52,7 @@ class MemoryManager
                                 posix::Allocator* f_managementAllocator,
                                 posix::Allocator* f_payloadAllocator);
 
-    SharedChunk getChunk(const MaxSize_t f_size,
+    SharedChunk getChunk(const uint32_t payloadSize,
                          const uint32_t payloadAlignment,
                          const uint32_t customHeaderSize,
                          const uint32_t customHeaderAlignment);
@@ -71,7 +71,7 @@ class MemoryManager
     static uint64_t requiredFullMemorySize(const MePooConfig& f_mePooConfig);
 
   private:
-    static uint32_t sizeWithChunkHeaderStruct(const MaxSize_t f_size);
+    static uint32_t sizeWithChunkHeaderStruct(const MaxChunkPayloadSize_t size);
 
     void printMemPoolVector() const;
     void addMemPool(posix::Allocator* f_managementAllocator,
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
@@ -57,19 +57,22 @@ class MemoryManager
                          const uint32_t customHeaderSize,
                          const uint32_t customHeaderAlignment);
 
-    uint32_t getMempoolChunkSizeForPayloadSize(const uint32_t f_size) const;
-
     uint32_t getNumberOfMemPools() const;
 
     MemPoolInfo getMemPoolInfo(uint32_t f_index) const;
 
-    static uint32_t sizeWithChunkHeaderStruct(const MaxSize_t f_size);
+    static uint32_t requiredChunkSize(const uint32_t payloadSize,
+                                      const uint32_t payloadAlignment,
+                                      const uint32_t customHeaderSize,
+                                      const uint32_t customHeaderAlignment);
 
     static uint64_t requiredChunkMemorySize(const MePooConfig& f_mePooConfig);
     static uint64_t requiredManagementMemorySize(const MePooConfig& f_mePooConfig);
     static uint64_t requiredFullMemorySize(const MePooConfig& f_mePooConfig);
 
   private:
+    static uint32_t sizeWithChunkHeaderStruct(const MaxSize_t f_size);
+
     void printMemPoolVector() const;
     void addMemPool(posix::Allocator* f_managementAllocator,
                     posix::Allocator* f_payloadAllocator,
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
@@ -1,4 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -41,7 +42,11 @@ inline typename ChunkSender<ChunkSenderDataType>::MemberType_t* ChunkSender<Chun
 
 template <typename ChunkSenderDataType>
 inline cxx::expected<mepoo::ChunkHeader*, AllocationError>
-ChunkSender<ChunkSenderDataType>::tryAllocate(const uint32_t payloadSize, const UniquePortId originId) noexcept
+ChunkSender<ChunkSenderDataType>::tryAllocate(const uint32_t payloadSize,
+                                              const UniquePortId originId,
+                                              const uint32_t payloadAlignment,
+                                              const uint32_t customHeaderSize,
+                                              const uint32_t customHeaderAlignment) noexcept
 {
     // use the chunk stored in m_lastChunk if there is one, there is no other owner and the new payload still fits in it
     const uint32_t neededChunkSize = getMembers()->m_memoryMgr->sizeWithChunkHeaderStruct(payloadSize);
@@ -63,7 +68,8 @@ ChunkSender<ChunkSenderDataType>::tryAllocate(const uint32_t payloadSize, const 
     {
         // BEGIN of critical section, chunk will be lost if process gets hard terminated in between
         // get a new chunk
-        mepoo::SharedChunk chunk = getMembers()->m_memoryMgr->getChunk(payloadSize);
+        mepoo::SharedChunk chunk =
+            getMembers()->m_memoryMgr->getChunk(payloadSize, payloadAlignment, customHeaderSize, customHeaderAlignment);
 
         if (chunk)
         {
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
@@ -49,7 +49,8 @@ ChunkSender<ChunkSenderDataType>::tryAllocate(const uint32_t payloadSize,
                                               const uint32_t customHeaderAlignment) noexcept
 {
     // use the chunk stored in m_lastChunk if there is one, there is no other owner and the new payload still fits in it
-    const uint32_t neededChunkSize = getMembers()->m_memoryMgr->sizeWithChunkHeaderStruct(payloadSize);
+    const uint32_t neededChunkSize = getMembers()->m_memoryMgr->requiredChunkSize(
+        payloadSize, payloadAlignment, customHeaderSize, customHeaderAlignment);
 
     if (getMembers()->m_lastChunk && getMembers()->m_lastChunk.hasNoOtherOwners()
         && (getMembers()->m_lastChunk.getChunkHeader()->chunkSize >= neededChunkSize))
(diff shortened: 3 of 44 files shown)


iceoryx_posh/include/iceoryx_posh/mepoo/chunk_header.hpp
```diff
@@ -25,18 +26,45 @@ namespace iox
...
+    /// @param[in] chunkSize is the size of the chunk the ChunkHeader is constructed
+    /// @param[in] payloadSize is the size of the payload
+    /// @param[in] payloadAlignment is the alignment of the payload
+    /// @param[in] customHeaderSize is the size of the custom header; if no custom header is used, use 0
+    /// @param[in] customHeaderAlignment is the alignment for the custom header; if no custom header is used, use 1
+    ChunkHeader(const uint32_t chunkSize,
+                const uint32_t payloadSize,
+                const uint32_t payloadAlignment,
+                const uint32_t customHeaderSize,
+                const uint32_t customHeaderAlignment) noexcept;
```
elBoberido, line 50: I think I will rename this to `userPayloadSize` ... `userHeaderAlignment` in a later PR. Then we can continue to configure our chunks like now with just the payload, which will be renamed to `chunkPayload`. With this we would have `chunk header` and `chunk payload` and the `chunk payload` could be split further into `user header` and `user payload`. Without `user header` and with the appropriate alignment `user payload` would have the same size as `chunk payload` which resembles the behavior before this changes.
```
+--------------+-------------------------------------------------------+
|              | chunk payload <- will be configured with the mempools |
| chunk header |+------------+-----------------------------------------+
| (fixed size) | user header | user payload                            |
+--------------+-------------+-----------------------------------------+
```

What do you think?

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -101,9 +101,9 @@ void iox_pub_deinit(iox_pub_t const self)
 
 iox_AllocationResult iox_pub_loan_chunk(iox_pub_t const self, void** const chunk, const uint32_t payloadSize)
 {
-    auto result = PublisherPortUser(self->m_portData).tryAllocateChunk(payloadSize).and_then([&](ChunkHeader* h) {
-        *chunk = h->payload();
-    });
+    auto result = PublisherPortUser(self->m_portData)
+                      .tryAllocateChunk(payloadSize, CHUNK_DEFAULT_PAYLOAD_ALIGNMENT)
+                      .and_then([&](ChunkHeader* h) { *chunk = h->payload(); });
```
elfenpiff, line 106: According to a very wise man (elBoberido the Great) you should only capture what you need.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
```diff
@@ -51,21 +52,27 @@ class MemoryManager
                                 posix::Allocator* f_managementAllocator,
                                 posix::Allocator* f_payloadAllocator);
 
-    SharedChunk getChunk(const MaxSize_t f_size);
-
-    uint32_t getMempoolChunkSizeForPayloadSize(const uint32_t f_size) const;
+    SharedChunk getChunk(const uint32_t payloadSize,
```
elfenpiff, line 55: Please add noexcept to all methods here.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
```diff
@@ -41,17 +42,26 @@ inline typename ChunkSender<ChunkSenderDataType>::MemberType_t* ChunkSender<Chun
...
+ChunkSender<ChunkSenderDataType>::tryAllocate(const uint32_t payloadSize,
+                                              const UniquePortId originId,
+                                              const uint32_t payloadAlignment,
+                                              const uint32_t customHeaderSize,
+                                              const uint32_t customHeaderAlignment) noexcept
 {
     // use the chunk stored in m_lastChunk if there is one, there is no other owner and the new payload still fits in it
-    const uint32_t neededChunkSize = getMembers()->m_memoryMgr->sizeWithChunkHeaderStruct(payloadSize);
+    const uint32_t neededChunkSize = getMembers()->m_memoryMgr->requiredChunkSize(
+        payloadSize, payloadAlignment, customHeaderSize, customHeaderAlignment);
```
elfenpiff, line 54: I think this whole code would be a little bit more understandable with
`auto & lastChunk = getMembers()->m_lastChunk;`
But it is not a must, just a suggestion.

iceoryx_posh/include/iceoryx_posh/mepoo/chunk_header.hpp
```diff
@@ -25,18 +26,45 @@ namespace iox
 {
 namespace mepoo
 {
+/// @brief Helper struct to use as default template parameter when no custom header is used
+struct NoCustomHeader
```
elfenpiff, line 30: To be more efficient at compile time I suggest the following approach:
```cpp
struct NoCustomHeader_t {};
constexpr NoCustomHeader_t NoCustomHeader{};
```

Then you can use it without always creating a new empty struct.

iceoryx_posh/source/mepoo/chunk_header.cpp
```diff
@@ -21,8 +22,74 @@ namespace iox
...
+    static_assert(alignof(ChunkHeader) >= 8U,
+                  "All the calculations expect the ChunkHeader alignment to be at least 8!");
+    cxx::Expects(customHeaderAlignment <= alignof(ChunkHeader)
+                 && "The alignment of the custom header must not exceed the alignment of the ChunkHeader!");
+
+    this->chunkSize = chunkSize;
+    this->payloadSize = payloadSize;
+
+    // have a look at »Payload Offset Calculation« in chunk_header.md for more details regarding the calculation
+    if (customHeaderSize == 0)
```
elfenpiff, line 40: Please add U

elfenpiff (commented):
Reviewed 20/43 files, more tomorrow

iceoryx_posh/include/iceoryx_posh/mepoo/chunk_header.hpp
elBoberido: the struct will never be instantiated since it's sole purpose is to live the life of a default template parameter

iceoryx_posh/source/mepoo/chunk_header.cpp
```diff
@@ -21,8 +22,74 @@ namespace iox
...
+            auto backOffset = reinterpret_cast<PayloadOffset_t*>(addressOfBackOffset);
+            *backOffset = payloadOffset;
+        }
+    }
+    else
+    {
+        // the most complex case with a custom header
+        auto addressOfChunkHeader = reinterpret_cast<uint64_t>(this);
+        uint64_t headerEndAddress = addressOfChunkHeader + sizeof(ChunkHeader) + customHeaderSize;
+        uint64_t potentialBackOffsetAddress =
```
elfenpiff, line 73: I think the name `unalignedBackOffsetAddress` is better fitting.

iceoryx_posh/source/mepoo/chunk_header.cpp
```diff
@@ -21,8 +22,74 @@ namespace iox
...
+            iox::cxx::align(headerEndAddress, static_cast<uint64_t>(alignof(PayloadOffset_t)));
+        uint64_t potentialPayloadAddress = potentialBackOffsetAddress + sizeof(PayloadOffset_t);
```
elfenpiff, line 75: What about `unalignedPayloadAddress`

iceoryx_posh/source/mepoo/chunk_header.cpp
```diff
@@ -21,8 +22,74 @@ namespace iox
...
+                     && "Payload offset too large for chunk!");
+        payloadOffset = static_cast<PayloadOffset_t>(offsetToPayload);
+
+        // this always works if the alignment of PayloadOffset_t and payloadAlignment are equal,
+        // if not there are two options:
+        //   - either the alignment of the PayloadOffset_t is larger than payloadAlignment
+        //     -> the payload is adjacent to the back-offset and therefore this also works
+        //   - or the alignment of the PayloadOffset_t is smaller than payloadAlignment
+        //     -> the back-offset can be put adjacent to to the Topic since the smaller alignment always fits
+        auto addressOfBackOffset = alignedPayloadAddress - sizeof(PayloadOffset_t);
```
elfenpiff, line 89: `BackOffsetAddress` to be consistent with the other names

iceoryx_posh/source/mepoo/memory_manager.cpp
```diff
@@ -155,21 +191,25 @@ void MemoryManager::configureMemoryManager(const MePooConfig& f_mePooConfig,
...
+                                    const uint32_t customHeaderSize,
+                                    const uint32_t customHeaderAlignment) noexcept
 {
     void* chunk{nullptr};
     MemPool* memPoolPointer{nullptr};
-    uint32_t adjustedSize = MemoryManager::sizeWithChunkHeaderStruct(f_size);
-    uint32_t totalSizeOfAquiredChunk = 0;
+    uint32_t requiredChunkSize =
+        MemoryManager::requiredChunkSize(payloadSize, payloadAlignment, customHeaderSize, customHeaderAlignment);
+    uint32_t aquiredChunkSize = 0;
```
elfenpiff, line 203: Add U

iceoryx_posh/source/mepoo/memory_manager.cpp
```diff
@@ -155,21 +191,25 @@ void MemoryManager::configureMemoryManager(const MePooConfig& f_mePooConfig,
...
 
     for (auto& memPool : m_memPoolVector)
     {
         uint32_t chunkSizeOfMemPool = memPool.getChunkSize();
-        if (chunkSizeOfMemPool >= adjustedSize)
+        if (chunkSizeOfMemPool >= requiredChunkSize)
```
elfenpiff, line 208: It has nothing to do with your pull request but I am realizing that this algorithm makes only then sense when the mempools are ordered from small to large otherwise it can happen that you acquire the 8 Mb chunk for your 10 byte data structure since it is simply the first in the list.

Is this somewhere ensured? Are we testing this?! 

Btw: :point_right: :nose:

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
```diff
@@ -108,26 +120,43 @@ TEST_F(ChunkHeader_test, FromPayloadFunctionCalledWithNullptrReturnsNullptr)
 
 TEST_F(ChunkHeader_test, usedChunkSizeIsSizeOfChunkHeaderWhenPayloadIsZero)
 {
-    ChunkHeader sut;
+    constexpr uint32_t CHUNK_SIZE{32};
```
elfenpiff, line 123: Gimme some U

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
```diff
@@ -108,26 +120,43 @@ TEST_F(ChunkHeader_test, FromPayloadFunctionCalledWithNullptrReturnsNullptr)
...
+
     sut.chunkSize = 2 * sizeof(ChunkHeader);
     EXPECT_THAT(sut.usedSizeOfChunk(), Eq(sizeof(ChunkHeader)));
 }
 
 TEST_F(ChunkHeader_test, usedChunkSizeIsSizeOfChunkHeaderPlusOneWhenPayloadIsOne)
 {
-    ChunkHeader sut;
-    sut.chunkSize = 2 * sizeof(ChunkHeader);
+    constexpr uint32_t CHUNK_SIZE{32};
```
elfenpiff, line 137: U

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
```diff
@@ -108,26 +120,43 @@ TEST_F(ChunkHeader_test, FromPayloadFunctionCalledWithNullptrReturnsNullptr)
...
+
+    sut.chunkSize = 2 * sizeof(ChunkHeader);
     EXPECT_THAT(sut.usedSizeOfChunk(), Eq(sizeof(ChunkHeader) + PAYLOAD_SIZE));
 }
 
 TEST_F(ChunkHeader_test, usedChunkSizeTerminatesWhenPayloadSizeExceedsChunkSize)
 {
-    ChunkHeader sut;
-    sut.chunkSize = 2 * sizeof(ChunkHeader);
+    constexpr uint32_t CHUNK_SIZE{32};
```
elfenpiff, line 151: U

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
```diff
@@ -274,16 +255,16 @@ TEST_F(MemoryManager_test, getChunkMultiMemPoolTooMuchChunks)
     std::vector<iox::mepoo::SharedChunk> chunkStore;
     for (size_t i = 0; i < ChunkCount; i++)
     {
-        chunkStore.push_back(sut->getChunk(32));
-        chunkStore.push_back(sut->getChunk(64));
-        chunkStore.push_back(sut->getChunk(128));
-        chunkStore.push_back(sut->getChunk(256));
+        chunkStore.push_back(sut->getChunk(32U, PAYLOAD_ALIGNMENT, CUSTOM_HEADER_SIZE, CUSTOM_HEADER_ALIGNMENT));
```
elfenpiff, line 258: What happens when you reverse the order here? I know it is not part of the pull request - it is just for fun :D

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
```diff
@@ -341,13 +322,224 @@ TEST_F(MemoryManager_test, freeChunkMultiMemPoolFullToEmptyToFull)
...
+    constexpr uint32_t CUSTOM_HEADER_SIZE{8U};
+    constexpr uint32_t CUSTOM_HEADER_ALIGNMENT{2 * alignof(iox::mepoo::ChunkHeader)};
+
+    EXPECT_DEATH(
+        {
+            iox::mepoo::MemoryManager::requiredChunkSize(
+                PAYLOAD_SIZE, PAYLOAD_ALIGNMENT, CUSTOM_HEADER_SIZE, CUSTOM_HEADER_ALIGNMENT);
+        },
+        ".*");
+}
```
elfenpiff, line 545: I think the following tests are missing:
- PayloadSize = 42 with an payload alignment of zero (same for the custom header)
- PayloadSize = 42 with an uint32_t max payload alignment (same for the custom header)
- PayloadSize = 0 with an uint32_t max payload alignment (same for custom header)
- Custom Header Size = 0 with a reasonable payload alignment

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
```diff
@@ -341,13 +322,224 @@ TEST_F(MemoryManager_test, freeChunkMultiMemPoolFullToEmptyToFull)
...
+    auto chunkSize = iox::mepoo::MemoryManager::requiredChunkSize(
+        PAYLOAD_SIZE, PAYLOAD_ALIGNMENT, CUSTOM_HEADER_SIZE, CUSTOM_HEADER_ALIGNMENT);
+
+    EXPECT_THAT(chunkSize, Eq(EXPECTED_SIZE));
+}
+
+TEST_F(MemoryManager_test, requiredChunkSizeWithPayloadAlignmentLessThanChunkHeaderAlignmentResultsInAdjacentPayload)
+{
+    constexpr uint32_t PAYLOAD_SIZE{42U};
+    constexpr uint32_t PAYLOAD_ALIGNMENT{alignof(iox::mepoo::ChunkHeader) / 2};
```
elfenpiff, line 371: You are testing here always with an alignment which is power of 2 but actually alignments like 29 and 31 (fun fact: prime twins) are also valid I think. Could you please add test for those lovely prime twins?

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
```diff
@@ -42,17 +42,22 @@ class ChunkQueue_testBase
...
+        ChunkHeader* chunkHeader = new (chunk) ChunkHeader(chunkMgmtPool.getChunkSize(),
+                                                           PAYLOAD_SIZE,
+                                                           iox::CHUNK_DEFAULT_PAYLOAD_ALIGNMENT,
+                                                           iox::CHUNK_NO_CUSTOM_HEADER_SIZE,
+                                                           iox::CHUNK_NO_CUSTOM_HEADER_ALIGNMENT);
         new (chunkMgmt) ChunkManagement{chunkHeader, &mempool, &chunkMgmtPool};
         return SharedChunk(chunkMgmt);
     }
 
+    static constexpr uint32_t PAYLOAD_SIZE{128};
```
elfenpiff, line 54: I think we need some U here

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
```diff
@@ -42,17 +42,22 @@ class ChunkQueue_testBase
...
     static constexpr size_t MEGABYTE = 1 << 20;
     static constexpr size_t MEMORY_SIZE = 4 * MEGABYTE;
     std::unique_ptr<char[]> memory{new char[MEMORY_SIZE]};
     iox::posix::Allocator allocator{memory.get(), MEMORY_SIZE};
-    MemPool mempool{128, 2 * iox::MAX_SUBSCRIBER_QUEUE_CAPACITY, &allocator, &allocator};
-    MemPool chunkMgmtPool{128, 2 * iox::MAX_SUBSCRIBER_QUEUE_CAPACITY, &allocator, &allocator};
+    MemPool mempool{PAYLOAD_SIZE, 2 * iox::MAX_SUBSCRIBER_QUEUE_CAPACITY, &allocator, &allocator};
```
elfenpiff, line 59: U and U

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -105,24 +110,28 @@ class ChunkSender_test : public Test
 
 TEST_F(ChunkSender_test, allocate_OneChunk)
 {
-    auto maybeChunkHeader = m_chunkSender.tryAllocate(sizeof(DummySample), iox::UniquePortId());
+    auto maybeChunkHeader = m_chunkSender.tryAllocate(
```
elfenpiff, line 113: In theory it could be possible that the custom header arguments are just ignored inside of `tryAllocate`. Therefore I would add a test here that verifies that `tryAllocate` with a custom header size != `iox::CHUNK_NO_CUSTOM_HEADER_SIZE` works and provides a chunk with the correct size/alignment

iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
```diff
@@ -202,26 +206,29 @@ TEST_F(PublisherPort_test,
 
 TEST_F(PublisherPort_test, allocatingAChunk)
 {
-    auto maybeChunkHeader = m_sutNoOfferOnCreateUserSide.tryAllocateChunk(10u);
+    auto maybeChunkHeader = m_sutNoOfferOnCreateUserSide.tryAllocateChunk(
```
elfenpiff, line 209: In theory it could be possible that the custom header arguments are just ignored inside of `tryAllocate`. Therefore I would add a test here that verifies that `tryAllocate` with a custom header size != `iox::CHUNK_NO_CUSTOM_HEADER_SIZE` works and provides a chunk with the correct size/alignment

elfenpiff (changes requested):

iceoryx_posh/source/mepoo/memory_manager.cpp
elBoberido: U know, every time one mentions the `U` a small little unicorn dies

iceoryx_posh/source/mepoo/memory_manager.cpp
elBoberido: AFAIK it's checked and not possible to have it ordered otherwise. But I will double check to be sure. It has been a long time since we implemented this.

iceoryx_posh/source/mepoo/chunk_header.cpp
elBoberido: well, it is perfectly aligned. The problem is that it must be adjacent to the payload and the payload might have a different alignment. The actual address can only be calculated later on, so this is only the potential address. Only after we know the aligned address of the payload we can recalculate the back offset address.

iceoryx_posh/source/mepoo/chunk_header.cpp
elBoberido: similar to the back offset, but in this case we don't know if it's aligned so in this case it would make sense to call it unaligned

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
elBoberido: and another unicorn died

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
elBoberido: number 4 :laughing:

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elBoberido: nope, alignments must always be power of 2

iceoryx_posh/test/moduletests/test_popo_chunk_queue.cpp
elBoberido: congratulations, you killed 6 unicorns :wink:

elBoberido: btw. in order to save some unicorns I propose to stick to the first exception of MISRA Rule 10.3
```
An essentially signed integer constant expression, with a rank no greater than signed int, may be assigned to an object of essentially unsigned type if its value can be represented in that type.
```
https://www.misra.org.uk/LinkClick.aspx?fileticket=-BRmxs8omoM%3d&tabid=57

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elfenpiff: Or we adjust the API a bit. That you do not provide the alignment directly as number, you provide which power of 2 your alignment has. Then this payload alignment could also be an uint8_t and we wouldn't require an assert which would mean less error handling, less code to test.

iceoryx_posh/source/mepoo/memory_manager.cpp
elBoberido: there is a test for this `addMemPoolWrongOrderAtLastElement` and the application is expected to die

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elBoberido: good question. It's not explicitly tested

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elBoberido: well, that would not be idiomatic. Everywhere else you can just use `alignof(T)` but in this case it would be `sqrt(alignof(T))`. I would hate the developer who enforces such an API on me. We could create a `cxx::align_t` and use that instead of the `uint32_t`. What do you think?

iceoryx_posh/source/mepoo/chunk_header.cpp
elBoberido: what do you think about `anticipatedBackOffsetAddress`?

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
```diff
@@ -108,26 +120,43 @@ TEST_F(ChunkHeader_test, FromPayloadFunctionCalledWithNullptrReturnsNullptr)
...
+                    iox::CHUNK_NO_CUSTOM_HEADER_SIZE,
+                    iox::CHUNK_NO_CUSTOM_HEADER_ALIGNMENT};
+
+    sut.chunkSize = 2 * sizeof(ChunkHeader);
     EXPECT_THAT(sut.usedSizeOfChunk(), Eq(sizeof(ChunkHeader) + PAYLOAD_SIZE));
 }
 
 TEST_F(ChunkHeader_test, usedChunkSizeTerminatesWhenPayloadSizeExceedsChunkSize)
 {
-    ChunkHeader sut;
```
elfenpiff, line 127: I think you require tests for the case:
- creating a chunk header with a custom header size where the custom header alignment does not exceed the chunk header alignment (death test)
- `fromPayload` function
    * acquire payload correct with custom chunk header (cch) & cch alignment <= header alignment
    * acquire payload correct with custom chunk header (cch) & cch alignment > header alignment
    * acquire payload when chunkheader end == aligned payload start
    * acquire payload when chunkheader end != aligned payload start (with backoffset)
    * acquire payload when chunkheader alignment < payload alignment
    * acquire payload when chunkheader alignment >= payload alignment
    * acquire payload when chunkheader alignment < payload alignment (with backoffset)
    * acquire payload when chunkheader alignment >= payload alignment (with backoffset)
- `usedSizeOfChunk` I think it requires the same cases as the payload tests
- `payload` requires also the same test cases ... maybe it makes sense to test always all 3 of them in the same test case
- alignment (maybe this can be handled by an alignment newtype?!)
    * what happens when the alignment is not power of 2?
    * what happen when the alignment is zero?

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elfenpiff: `cxx::align_t` sounds great!

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -103,32 +108,57 @@ class ChunkSender_test : public Test
...
 
     ASSERT_FALSE(chunk1.has_error());
     ASSERT_FALSE(chunk2.has_error());
     // must be different chunks
     EXPECT_THAT(*chunk1, Ne(*chunk2));
-    EXPECT_THAT(m_memoryManager.getMemPoolInfo(0).m_usedChunks, Eq(2u));
+    EXPECT_THAT(m_memoryManager.getMemPoolInfo(0).m_usedChunks, Eq(2U));
 }
 
 TEST_F(ChunkSender_test, allocate_Overflow)
```
elfenpiff, line 164: Maybe a test which allocates until overflow and then releases some samples to make space to allocate again would be nice.

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -183,53 +222,65 @@ TEST_F(ChunkSender_test, freeChunk)
         m_chunkSender.release(chunks[i]);
     }
 
-    EXPECT_THAT(m_memoryManager.getMemPoolInfo(0).m_usedChunks, Eq(0u));
+    EXPECT_THAT(m_memoryManager.getMemPoolInfo(0).m_usedChunks, Eq(0U));
 }
 
 TEST_F(ChunkSender_test, freeInvalidChunk)
```
elfenpiff, line 228: What about a double free test?

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elBoberido: Should be done now. Note, a custom header alignment larger than the alignment of `ChunkHeader` is not allowed

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
elBoberido: yes, they are not yet done

elBoberido: Should be done now. The power of 2 alignment checks will be done with the the `cxx::align_t` type in the next PR

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
elBoberido: After some thinking, I came to the conclusion that this is already done by `TEST_F(ChunkSender_test, freeChunk)`. This tests doesn't allocate a new chunk, but ensures that the chunks are freed

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
elBoberido: Shouldn't this be part of the tests for the `MemPool` and `UsedChunkList`? There is no way to check something with the `ChunkSender` except for not segfaulting.

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elBoberido: I added a test to check that the correct `MemPool` is used, even if empty. This should ensure that the order doesn't matter.

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
elBoberido: I've come to the conclusion to create `ChunkSettings` struct with those 4 parameter and also move the `requiredChunkSize` calculation to that struct. All the method would expect this struct instead of the 4 parameters. This makes it typesafe and hard to pass different arguments to the `ChunkHeader` ctor. Would like to do it in a follow up PR, since this PR is already quite large and currently this is only used internally

elfenpiff (approved):

elfenpiff (approved):

FerdinandSpitzschnueffler (approved):

PR #618, opened 2021-03-17
Iox #605 cleanup relocatable ptr

### LINKED ISSUE #605: Refactor Relocatable Pointer and Relative Pointer
## Brief feature description

The relocatable pointer (`iceoryx_utils/internal/relocatable_pointer`) implementation needs to be cleaned up:

 1. [x] add noexcept, const, replace `size_t` with `uint64_t` (done in #618 except making parameters const)
 2. [x] separate implementation and declaration for all classes (#618)
 3. [x] one class per header file (#618)
 4. [x] add subnamespace `iox::rp` where everything is stored (#618)
 5. [x] ~~restructure the unit tests / add missing tests~~ Moved to #1745
 6. [x] add doxygen documentation (#618)
 7. [x] remove includes of base_relative_ptr.hpp & base_relocatable_ptr.hpp
 9. [x] ~~consider removing ``ptr_t`` and ``const_ptr_t`` aliases and use ``void*`` directly (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597604809)~~
 10. [x] ~~add ``void* get() noexcept``, ``const void* get() const noexcept``, ``void* getBasePtr() noexcept`` and ``const void* getBasePtr() const noexcept`` to BaseRelativePointer~~ -> Not implemented as `SharedPointer` uses `get() const noexcept`
 11. [x] ~~consider changing return type of ``PointerRepository::unregisterPrt()`` to ``void`` (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597609248)~~ - Note: maybe not, it might be useful to have an indication whether the deregistration was successful
 12. [x] move duplicated code into a separate method (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597613171)
 13. [x] rename ``PointerRepository::registerPtr()`` methods if both are needed (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597613855) + consider returning a ``cxx::unique_ptr<id_t>``
 14. [x] check invalid id logic (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597617681 and https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597619467)
 15. [x] ~~add const methods to RelativePointer (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597620985, https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597621144 and https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597621263).~~ This might not be needed, compare with e.g. std::shared_pointer https://en.cppreference.com/w/cpp/memory/shared_ptr/operator* It should work similar to STL smart pointers.
 16. [x] invalidate ``other`` move c'tor and move assignment operator of BaseRelocatablePointer (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597626715 and https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597628041)
 17. [x] use composition instead of inheritance (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r599596822 and https://github.com/eclipse-iceoryx/iceoryx/issues/605#issuecomment-806045008) - if done, consider renaming Base...Pointer to Untyped...Pointer
 18. [x] ~~replace `RelativePointer` by `RelocatablePointer` where applicable, e.g. `Semaphore` and `LoFFLi`~~ Moved to #1745
 19. [x] Move `RelativePointer` to public include

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
The restructuring of the unit tests and adding missing tests will be done in a separate PR.

## References

- Partly closes #605


doc/conceptual-guide.md
```diff
@@ -142,7 +142,7 @@ Memory chunks are returned to the pool once all attached `SubscriberPort`s indic
 ### A Note on Pointers
 As already discussed, shared memory segments may be mapped to different memory areas in the virtual address space of a
 process.
-To deal with this, iceoryx utilizes specialized pointer types: the `iox::RelativePointer` and
+To deal with this, iceoryx utilizes specialized pointer types: the `iox::rp::RelativePointer` and
 the `iox::RelocatablePointer`.
```
MatthiasKillat, line 146: RelocatablePointer is also in the rp namespace as far as I can see.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/shared_chunk.hpp
```diff
@@ -35,7 +35,7 @@ class SharedChunk
   public:
     SharedChunk() = default;
     SharedChunk(ChunkManagement* const f_resource);
-    SharedChunk(const relative_ptr<ChunkManagement>& f_resource);
+    SharedChunk(const rp::RelativePointer<ChunkManagement>& f_resource);
```
MatthiasKillat, line 38: I liked the STL syntax more. Can stay like this but we should consider using snake case in utils. It also got more verbose for no good reason, always a negative in my book.

iceoryx_posh/test/moduletests/test_mepoo_segment_management.cpp
```diff
@@ -21,7 +21,7 @@
 #include "iceoryx_utils/error_handling/error_handling.hpp"
 #include "iceoryx_utils/internal/posix_wrapper/shared_memory_object.hpp"
 #include "iceoryx_utils/internal/posix_wrapper/shared_memory_object/allocator.hpp"
-#include "iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp"
+#include "iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp"
```
MatthiasKillat, line 24: We do not need to change it now, but now hat the base version is in separate header I would prefer we would never have to include it since it is an implementation detail.  relative.ptr.hpp alone should define the interface.
This would require more refactoring though to e.g. have all those static methods and constants in the rp namespace.

I would prefer this not to be the last refactoring but it is ok for now.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
```diff
@@ -21,45 +23,57 @@
...
+/// RelocatablePointer cannot be used in an atomic since the copy ctor is nontrivial.
 template <typename T>
 class atomic_relocatable_ptr
 {
   public:
     using offset_t = std::ptrdiff_t;
     static constexpr offset_t NULL_POINTER_OFFSET = std::numeric_limits<offset_t>::max();
 
-    atomic_relocatable_ptr(const T* ptr = nullptr);
+    /// @brief creates atomic_relocatable_ptr pointing to ptr
```
MatthiasKillat, line 40: It creates an atomic_relocatable_ptr that points to the same pointee as ptr **from ptr**, but not to ptr itself.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
```diff
@@ -21,45 +23,57 @@
...
     atomic_relocatable_ptr(const atomic_relocatable_ptr&) = delete;
     atomic_relocatable_ptr& operator=(const atomic_relocatable_ptr& other) = delete;
-
     atomic_relocatable_ptr(atomic_relocatable_ptr&& other) = delete;
     atomic_relocatable_ptr& operator=(atomic_relocatable_ptr&& other) = delete;
 
-    // minimal set of required operators, can be extended later
+    /// @note minimal set of required operators, can be extended later
 
+    /// @brief assign atomic_relocatable_ptr to point to ptr
```
MatthiasKillat, line 52: point to the same pointee as ptr

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+/// whole segment, such a segment would have to be registered on its own (and the original segment deregistered).
+class BaseRelativePointer
+{
+  public:
+    using id_t = uint64_t;
+    using ptr_t = void*;
+    using const_ptr_t = const void* const;
+    using offset_t = std::uintptr_t;
+
+    /// @brief constructs a BaseRelativePointer pointing to ptr in a segment identified by id
```
MatthiasKillat, line 59: "pointing to the same pointee as ptr" (may not be the best wording, but it is accurate)
This occurs multiple times.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+    /// @return reference to self
+    BaseRelativePointer& operator=(void* ptr) noexcept;
+
+    /// @brief move assignment
+    /// @param[in] other is the move origin
+    /// @return a reference to self
+    BaseRelativePointer& operator=(BaseRelativePointer&& other) noexcept;
+
+    /// @brief access to the underlying object
+    /// @return a poiter to the underlying object
```
MatthiasKillat, line 97: pointer

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+    /// @return the registered base pointer
+    ptr_t getBasePtr() const noexcept;
+
+    /// @brief registers a memory segment at ptr with size of a new id
+    /// @param[in] ptr starting address of the segment to be registered
+    /// @param[in] size is the size of the segment
+    /// @return id it was registered to
+    static id_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
+
+    /// @brief registers a memory segment at ptr with size of given id
```
MatthiasKillat, line 118: tries to register a memory segment with a given size starting at ptr to a given id

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+    static void unregisterAll() noexcept;
+
+    /// @brief get the offset from id and ptr
+    /// @param[in] id is the id of the segment and is used to get the bease pointer
+    /// @param[in] ptr is the pointer whose offset should be calculated
+    /// @return offset
+    static offset_t getOffset(const id_t id, const_ptr_t ptr) noexcept;
+
+    /// @brief get the pointer from id and offset ("inverse" to getOffset)
+    /// @param[in] id is the id of the segment and is used to get the bease pointer
```
MatthiasKillat, line 145: base

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relocatable_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+    template <typename T>
+    friend class RelocatablePointer;
+
+  public:
+    using offset_t = std::ptrdiff_t;
+
+    /// @brief default constructs a logical nullptr
+    BaseRelocatablePointer() noexcept;
+
+    /// @brief creates a relocatable pointer pointing to ptr
```
MatthiasKillat, line 54: same pointee as ptr (pointer and relocatable pointer are different entities but they are equivalent wrt. what they point to ...)

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -38,142 +42,71 @@ class PointerRepository
...
-        if (id <= MAX_ID && id >= MIN_ID)
-        {
-            return m_info[id].basePtr;
-        }
+    /// @brief unregisters all ids and also invalidates all relative pointers
+    void unregisterAll() noexcept;
 
-        // for id 0 nullptr is returned, meaning we will later interpret a relative pointer
-        // by casting the offset into a pointer (i.e. we measure relative to 0)
+    /// @brief gets the base pointer associated with id
```
MatthiasKillat, line 78: Maybe confusing but base pointer and and starting address of a segment are used interchangebly in the code I wrote a long time ago :-) Maybe a comment would be helpful to clarify this...

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -13,291 +14,95 @@
...
-    relative_ptr& operator=(ptr_t ptr)
-    {
-        m_id = searchId(ptr);
-        m_offset = computeOffset(ptr);
-
-        return *this;
-    }
-
+    /// @brief constructs a RelativePointer pointing to ptr in a segment identified by id
+    /// @param[in] ptr is the pointee
```
MatthiasKillat, line 37: I think I understand what you mean with this now: the RelativePointer points to ptr (it locates the ptr) which in turn points to the data. Therefore ptr is the pointee of a RelativePointer. It can certainly be seen like this.

On the other hand my view is: they logically both point to the same data (the pointee), and can be constrcuted from one another (here relativePointer from ptr). Neither are really pointees in my book if we see it this way.

What is more "natural" or intuitive I cannot say. We just need to be consistent. I prefer the second interpretation, it is more in line of how we think smart pointers like shared_ptr and so on.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -13,291 +14,95 @@
...
-    }
+    /// @brief converts the RelativePointer to a pointer of the type of the underlying object
+    /// @return a pointer of type T pointing to the underlying object
+    operator T*() const noexcept;
 
-    bool operator==(T* const ptr) const
-    {
-        return ptr == get();
-    }
+    /// @brief checks if the raw pointer is equal to ptr
```
MatthiasKillat, line 92: more precisely: we check whether the RelativePointer (this) and ptr point to the same pointee.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -13,188 +14,69 @@
...
-    }
-
-    operator T*() const
-    {
-        return reinterpret_cast<T*>(computeRawPtr());
-    }
+    /// @brief default constructs a logical nullptr
+    RelocatablePointer() noexcept;
+
+    /// @brief creates a RelocatablePointer pointing to ptr
```
MatthiasKillat, line 35: Depending again on the view of logical equivalence of pointers and what we want to consider the pointee.

MatthiasKillat (commented):
Some minor comment changes are needed. Refactoring to not use the base_relative_ptr.hpp header in the code should be deferred. Gcov also indicates lack of test coverage. his will be solved in a future issue from my point of view (functional refactoring and increasing test coverage). This issue here separates interfaces and improves documentation.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_types.hpp
```diff
@@ -26,10 +26,10 @@ namespace popo
 struct ChunkTuple
 {
     ChunkTuple() = default;
-    explicit ChunkTuple(relative_ptr<mepoo::ChunkManagement> f_chunk) noexcept;
+    explicit ChunkTuple(rp::RelativePointer<mepoo::ChunkManagement> f_chunk) noexcept;
```
mossmaurice, line 29: You could consider removing the `f_*` in the touched classes if that's not to much effort.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relocatable_ptr.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// @brief checks if this is not a logical nullptr
+    /// @return true if this is not a logical nullptr, otherwise false
+    operator bool() const noexcept;
+
+    /// @brief checks if this is a logical nullptr
+    /// @return true if this is a logical nullptr, otherwise false
+    bool operator!() const noexcept;
+
+    /// @brief access to the underlying object in shared memory
+    /// @return a poiter to the underlying object
```
MatthiasKillat, line 94: pointer

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -38,142 +42,71 @@ class PointerRepository
...
-    }
-
-    bool unregisterPtr(id_t id)
-    {
-        if (id <= MAX_ID && id >= MIN_ID)
-        {
-            if (m_info[id].basePtr != nullptr)
-            {
-                m_info[id].basePtr = nullptr;
+    /// @brief registers the start pointer of the segment in another application with a specific id
```
MatthiasKillat, line 57: Strictly speaking it does not have to be in another application, but it is used in this way and exists for this very reason. Leave as is for clarity of the use case.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -38,142 +42,71 @@ class PointerRepository
...
-        }
+    /// @brief registers the start pointer of a segment with a specific size
+    /// @param[in] ptr is the start pointer of the segment
+    /// @param[in] size is the size of the segment
+    /// @return the id that identifies the segment
+    id_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
 
-        return false;
-    }
+    /// @brief unregisters the id and also invalidates relative pointers using this id
```
MatthiasKillat, line 70: It does not really invalidate the relative pointers in use (it cannot, it does not know them). It would be better to have an @attention were we state that relative pointers corresponding to this id become unsafe to use (they can be considered invalid).

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -13,188 +14,69 @@
...
+    /// @brief dereferencing operator which returns a const reference to the pointee
+    /// @return a const reference to the pointee
+    const T& operator*() const noexcept;
+
+    /// @brief read-only access to the underlying object in shared memory
+    /// @return a const pointer to the underlying object
+    const T* operator->() const noexcept;
+
+    /// @brief returns a reference to the memory location of the underlying object + an offset
+    /// @param[in] index is the offset
```
MatthiasKillat, line 69: measured in sizeof(T), but this should be clear

MatthiasKillat (approved):

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -0,0 +1,157 @@
...
+        {
+            m_maxRegistered = id;
+        }
+        return true;
+    }
+    return false;
+}
+
+template <typename id_t, typename ptr_t, uint64_t CAPACITY>
+inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(const ptr_t ptr, uint64_t size) noexcept
```
elfenpiff, line 54: We should use RAII here as well and return a `cxx::unique_ptr<id_t>` which calls unregisterPtr when it goes out of scope.

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+}
+
+BaseRelocatablePointer::BaseRelocatablePointer(const BaseRelocatablePointer& other) noexcept
+    : m_offset(computeOffset(other.computeRawPtr()))
+{
+}
+
+BaseRelocatablePointer::BaseRelocatablePointer(BaseRelocatablePointer&& other) noexcept
+    : m_offset(computeOffset(other.computeRawPtr()))
```
elfenpiff, line 39: We should invalidate other here by writing `other.m_offset = NULL_POINTER_OFFSET;`

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+BaseRelocatablePointer& BaseRelocatablePointer::operator=(const void* rawPtr) noexcept
+{
+    m_offset = computeOffset(rawPtr);
+    return *this;
+}
+
+BaseRelocatablePointer& BaseRelocatablePointer::operator=(BaseRelocatablePointer&& other) noexcept
+{
+    m_offset = computeOffset(other.computeRawPtr());
```
elfenpiff, line 62: We have to invalidate other here, otherwise we may have undefined behavior. Please add `other.m_offset = NULL_POINTER_OFFSET;`

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+BaseRelocatablePointer::operator bool() const noexcept
+{
+    return m_offset != NULL_POINTER_OFFSET;
+}
+
+bool BaseRelocatablePointer::operator!() const noexcept
+{
+    return m_offset == NULL_POINTER_OFFSET;
```
elfenpiff, line 78: Maybe we add some more logic to this and therefore I would return here:
`return !(operator bool());`

elfenpiff (changes requested):
- There are a lot of issues in the UsedChunkList which should not be fixed in this PR but we have to create an issue for that.
- There are also a lot of issues remaining in the relocatable pointer which should be tracked in the issue and do not have to be managed in this PR.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -38,142 +42,72 @@ class PointerRepository
...
+    /// @return the id that identifies the segment
+    id_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
+
+    /// @brief unregisters the id
+    /// @param[in] id is the id to be unregistered
+    /// @return true if successful, otherwise false
+    /// @attention the relative pointers corresponding to this id become unsafe to use
+    bool unregisterPtr(id_t id) noexcept;
+
+    /// @brief unregisters all ids and also invalidates all relative pointers
```
FerdinandSpitzschnueffler, line 76: it doesn't invalidate the pinters

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
```diff
@@ -0,0 +1,115 @@
...
+BaseRelocatablePointer::BaseRelocatablePointer(const void* ptr) noexcept
+    : m_offset(computeOffset(ptr))
+{
+}
+
+BaseRelocatablePointer::BaseRelocatablePointer(const BaseRelocatablePointer& other) noexcept
+    : m_offset(computeOffset(other.computeRawPtr()))
+{
+}
+
```
FerdinandSpitzschnueffler, line 37: virtual d'tor for BaseRelocatablePointer + for every base class

elBoberido: nope, this lives in the shared memory, therefore virtual is not allowed

iceoryx_posh/test/moduletests/test_roudi_memory_provider.cpp
```diff
@@ -16,7 +16,7 @@
 
 #include "iceoryx_posh/roudi/memory/memory_provider.hpp"
 
-#include "iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp"
+#include "iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp"
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since I only updated the include path and the namespace - no changes in the implementation.

iceoryx_posh/test/moduletests/test_mepoo_shared_pointer.cpp
```diff
@@ -18,7 +18,7 @@
 #include "iceoryx_posh/internal/mepoo/shared_pointer.hpp"
 #include "iceoryx_posh/mepoo/chunk_header.hpp"
 #include "iceoryx_utils/internal/posix_wrapper/shared_memory_object/allocator.hpp"
-#include "iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp"
+#include "iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp"
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since I only updated the include path and the namespace - no changes of the implementation.

iceoryx_posh/test/moduletests/test_mepoo_segment_management.cpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since I only updated the include path and the namespace - no changes of the implementation.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/client_server_port_types.hpp
```diff
@@ -89,7 +89,7 @@ class RPCBaseHeader
     }
 
   protected:
-    relative_ptr<ClientChunkQueueData_t> m_clientQueueDataPtr;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since I only updated the namespace - no changes of the implementation.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender_data.hpp
```diff
@@ -38,7 +38,7 @@ struct ChunkSenderData : public ChunkDistributorDataType
 
     using ChunkDistributorData_t = ChunkDistributorDataType;
 
-    const relative_ptr<mepoo::MemoryManager> m_memoryMgr;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since I only updated the namespace - no changes of the implementation.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor_data.hpp
```diff
@@ -46,7 +46,8 @@ struct ChunkDistributorData : public LockingPolicy
 
     const uint64_t m_historyCapacity;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since It's only a renaming - no changes of the implementation.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -59,14 +59,13 @@ ChunkDistributor<ChunkDistributorDataType>::tryAddQueue(cxx::not_null<ChunkQueue
         if (getMembers()->m_queues.size() < getMembers()->m_queues.capacity())
         {
             // PRQA S 3804 1 # we checked the capacity, so pushing will be fine
-            getMembers()->m_queues.push_back(relative_ptr<ChunkQueueData_t>(queueToAdd));
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I don't think this is necessary here since It's only a renaming - no changes of the implementation.

iceoryx_posh/test/moduletests/test_mepoo_segment_management.cpp
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
```diff
@@ -21,45 +23,57 @@
...
+    /// @note minimal set of required operators, can be extended later
 
+    /// @brief assign atomic_relocatable_ptr to point to the same pointee as ptr
+    /// @param[in] ptr the pointer whose pointee shall be the same for this
+    /// @return reference to self
     atomic_relocatable_ptr& operator=(const T* ptr) noexcept;
 
+    /// @brief access to the underlying object in shared memory
+    /// @return a pointer to the underlying object
     T* operator->() const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
```diff
@@ -21,45 +23,57 @@
...
 class atomic_relocatable_ptr
 {
   public:
     using offset_t = std::ptrdiff_t;
     static constexpr offset_t NULL_POINTER_OFFSET = std::numeric_limits<offset_t>::max();
 
-    atomic_relocatable_ptr(const T* ptr = nullptr);
+    /// @brief creates an atomic_relocatable_ptr pointing to the same pointee as ptr
+    /// @param[in] ptr the pointer whose pointee shall be the same for this
+    atomic_relocatable_ptr(const T* ptr = nullptr) noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This is one open point of the issue.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
```diff
@@ -21,45 +23,57 @@
...
     operator T*() const noexcept;
 
   private:
     std::atomic<offset_t> m_offset{NULL_POINTER_OFFSET};
 
-    inline T* computeRawPtr() const;
+    inline T* computeRawPtr() const noexcept;
 
-    inline offset_t computeOffset(const void* ptr) const;
+    inline offset_t computeOffset(const void* ptr) const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This is one open point of the issue.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+  public:
+    using id_t = uint64_t;
+    using ptr_t = void*;
+    using const_ptr_t = const void* const;
+    using offset_t = std::uintptr_t;
+
+    /// @brief constructs a BaseRelativePointer pointing to the same pointee as ptr in a segment identified by id
+    /// @param[in] ptr the pointer whose pointee shall be the same for this
+    /// @param[in] id is the unique id of the segment
+    BaseRelativePointer(ptr_t ptr, id_t id) noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This is one open point of the issue.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+    BaseRelativePointer& operator=(void* ptr) noexcept;
+
+    /// @brief move assignment
+    /// @param[in] other is the move origin
+    /// @return a reference to self
+    BaseRelativePointer& operator=(BaseRelativePointer&& other) noexcept;
+
+    /// @brief access to the underlying object
+    /// @return a pointer to the underlying object
+    ptr_t get() const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
```diff
@@ -0,0 +1,184 @@
...
+    /// @return the id which identifies the segment
+    id_t getId() const noexcept;
+
+    /// @brief returns the offset
+    /// @return the offset
+    offset_t getOffset() const noexcept;
+
+    /// @brief get the base pointer associated with this' id
+    /// @return the registered base pointer
+    ptr_t getBasePtr() const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -38,142 +42,72 @@ class PointerRepository
...
-    }
+    /// @brief default constructor
+    PointerRepository() noexcept;
+
+    /// @brief registers the start pointer of the segment in another application with a specific id
+    /// @param[in] id identifies the segment
+    /// @param[in] ptr is the start pointer of the segment
+    /// @param[in] size is the size of the segment
+    /// @return true if the registration was successful, otherwise false
+    bool registerPtr(id_t id, ptr_t ptr, uint64_t size) noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This is one open point of the issue.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -38,142 +42,72 @@ class PointerRepository
...
+    /// @param[in] ptr is the start pointer of the segment
+    /// @param[in] size is the size of the segment
+    /// @return the id that identifies the segment
+    id_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
+
+    /// @brief unregisters the id
+    /// @param[in] id is the id to be unregistered
+    /// @return true if successful, otherwise false
+    /// @attention the relative pointers corresponding to this id become unsafe to use
+    bool unregisterPtr(id_t id) noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -0,0 +1,157 @@
...
+{
+    for (id_t id = 1U; id <= MAX_ID; ++id)
+    {
+        if (m_info[id].basePtr == nullptr)
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -0,0 +1,157 @@
...
+namespace rp
+{
+template <typename id_t, typename ptr_t, uint64_t CAPACITY>
+inline PointerRepository<id_t, ptr_t, CAPACITY>::PointerRepository() noexcept
+    : m_info(CAPACITY)
+{
+}
+
+template <typename id_t, typename ptr_t, uint64_t CAPACITY>
+inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t ptr, uint64_t size) noexcept
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -0,0 +1,157 @@
...
+        if (ptr >= m_info[id].basePtr && ptr <= m_info[id].endPtr)
+        {
+            return id;
+        }
+    }
+    /// @note implicitly interpret the pointer as a regular pointer if not found
+    /// by setting id to 0
+    /// rationale: test cases work without registered shared memory and require
+    /// this at the moment to avoid fundamental changes
+    return 0U;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -0,0 +1,157 @@
...
+    // return INVALID_ID;
+}
+
+template <typename id_t, typename ptr_t, uint64_t CAPACITY>
+inline bool PointerRepository<id_t, ptr_t, CAPACITY>::isValid(id_t id) const noexcept
+{
+    return id != INVALID_ID;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -13,291 +14,95 @@
...
-        m_id = searchId(ptr);
-        m_offset = computeOffset(ptr);
-
-        return *this;
-    }
-
+    /// @brief constructs a RelativePointer pointing to the same pointee as ptr in a segment identified by id
+    /// @param[in] ptr the pointer whose pointee shall be the same for this
+    /// @param[in] id is the unique id of the segment
+    RelativePointer(ptr_t ptr, id_t id) noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This is one open point of the issue.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -13,291 +14,95 @@
...
-    }
+    typename std::enable_if<!std::is_void<U>::value, const U&>::type operator*() const noexcept;
 
-    T* operator->() const
-    {
-        return get();
-    }
+    /// @brief read-only access to the underlying object
+    /// @return a const pointer to the underlying object
+    T* operator->() const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -13,291 +14,95 @@
...
 
-    T* get() const
-    {
-        return static_cast<T*>(computeRawPtr());
-    }
+    /// @brief access the underlying object
+    /// @return a pointer to the underlying object
+    T* get() const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
```diff
@@ -13,291 +14,95 @@
...
 
-    operator T*() const
-    {
-        return get();
-    }
+    /// @brief converts the RelativePointer to a pointer of the type of the underlying object
+    /// @return a pointer of type T pointing to the underlying object
+    operator T*() const noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
FerdinandSpitzschnueffler: added to issue

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
FerdinandSpitzschnueffler: added to issue

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -51,7 +51,7 @@ class UsedChunkList
             m_usedListHead = m_freeListHead;
 
             // store chunk mgmt ptr
-            m_data[m_usedListHead] = f_chunk.release();
+            m_data[m_usedListHead] = chunk.release();
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added #623

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -67,18 +67,18 @@ class UsedChunkList
     }
 
     // only from runtime context
-    bool remove(const mepoo::ChunkHeader* f_chunkHeader, mepoo::SharedChunk& f_chunk)
+    bool remove(const mepoo::ChunkHeader* chunkHeader, mepoo::SharedChunk& chunk)
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added to #623

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -172,7 +172,7 @@ class UsedChunkList
     uint32_t m_usedListHead{InvalidIndex};
     uint32_t m_freeListHead{0u};
     std::array<uint32_t, Size> m_list;
-    std::array<relative_ptr<mepoo::ChunkManagement>, Size> m_data;
+    std::array<rp::RelativePointer<mepoo::ChunkManagement>, Size> m_data;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: added #623

iceoryx_posh/include/iceoryx_posh/internal/mepoo/chunk_management.hpp
```diff
@@ -36,20 +36,20 @@ struct alignas(32) ChunkManagement
     using referenceCounterBase_t = uint64_t;
     using referenceCounter_t = std::atomic<referenceCounterBase_t>;
 
-    ChunkManagement(const cxx::not_null<base_t*> f_chunkHeader,
-                    const cxx::not_null<MemPool*> f_mempool,
-                    const cxx::not_null<MemPool*> f_chunkManagementPool)
-        : m_chunkHeader(f_chunkHeader)
-        , m_mempool(f_mempool)
-        , m_chunkManagementPool(f_chunkManagementPool)
+    ChunkManagement(const cxx::not_null<base_t*> chunkHeader,
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I added ``noexcept`` but in this PR I would like to keep the signature since changing it will require a refactoring of the c'tors of ``RelativePointer`` and ``BaseRelativePointer``. This should be done in a follow-up PR.

iceoryx_posh/test/moduletests/test_mepoo_shared_pointer.cpp
```diff
@@ -102,11 +102,11 @@ class SharedPointer_Test : public Test
 
     void SetUp() override
     {
-        iox::RelativePointer::registerPtr(memory, 4096);
+        iox::rp::BaseRelativePointer::registerPtr(memory, 4096);
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This test file includes many magic numbers and would need a lot of changes unrelated to the topic of this PR. Therefore I'd like to leave it as it is.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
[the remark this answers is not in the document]
MatthiasKillat: If we rework this in another PR we should consider making it lock-free or protecting the list by mutex (due to concurrent take and release operations on samples or raw pointers). Whether we then will use SharedChunk internally is another question, but I would not refactor this without thinking about the concurrency problem first.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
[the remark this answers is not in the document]
MatthiasKillat: std::array is ok to use as it does not allocate dynamic memory... we do not have to reinvent every wheel I think. Exception freedom I am almost sure that the operations performed cannot throw (i.e. accessing is unchecked as per STL philosophy)

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/relative_ptr.hpp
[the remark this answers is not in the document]
MatthiasKillat: This is debatable, as it constructs a new T*. You get call this function and it guarantees not to change the relative pointer itself. It probably should be const to not allow changing the data via the raw pointer then though. But there must be a const and non const version in this case since we need the possibility to get a non const pointer.
As it is now it is not critical to be "fixed".

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
MatthiasKillat: PointerRepository practically internal API (implementation detail) and optimized for speed (could still be improved in this regard but this would slow it down).
This behavior is also not what we want, otherwise I need to keep those ids around to avoid unregistering. Not cool.

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
MatthiasKillat: This is a philosophical question wrt. moving objects. The standard does not mandate it. Could be done in functional refactoring of the class.

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
MatthiasKillat: Move philosophy issue again. Depends on assumptions on how move is used, usually only explicit or from temporaries. In the latter case invalidation only incurs cost, in the former the user should be wise enough not to use the source since he triggered move explicitly.

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
MatthiasKillat: Virtual dtor also only if you use virtual functions to ensure the correct dtor is called. This is not an issue here since we do not use them.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.hpp
MatthiasKillat: Yes, it cannot (does not know them nor could it keep track of copies etc.). But they become unsafe to use, this is what is meant.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
```diff
@@ -21,45 +23,57 @@
...
-/// Rationale: the default relocatable_ptr cannot be used in an atomic since the copy ctor is nontrivial.
+namespace rp
+{
+/// @brief minimalistic relocatable pointer that can be written and read atomically and can be stored safely in shared
+/// memory.
+/// As the basic RelocatablePointer, it must point to something in the same shared memory segment as itself since the
+/// internally used offset must be an invariant different across adress spaces. Rationale: the default
+/// RelocatablePointer cannot be used in an atomic since the copy ctor is nontrivial.
 template <typename T>
 class atomic_relocatable_ptr
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, I saw this but we should consider to use snake case in our utility library in the future. This has the added benefit to more easily distinguish the calls from the STL/utils and posh at a quick glance.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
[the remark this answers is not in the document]
MatthiasKillat: This is not how it should behave. Please cf. https://en.cppreference.com/w/cpp/memory/shared_ptr/operator* for comparison

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/pointer_repository.inl
elfenpiff: @MatthiasKillat 
1. If you do not want to keep this ids around how do you keep track when cleaning up everything with `unregisterPtr`?
2. Is this on our hot path registering (and unregistering) memory? I thought this is only called on startup or whenever you map a new shared memory into your process?!

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
elfenpiff: Why do we inherit at all here? Wouldn't it be possible to use it as a composite then we would be completely on the safe side.
I agree with you but this can be easily error prone - not without reason Scott Meyers suggest to always have a virtual destructor when working with inheritance.

elfenpiff (approved):

iceoryx_utils/source/relocatable_pointer/base_relocatable_ptr.cpp
elBoberido: Can we maybe also rename the `BaseRelocatablePointer` to `UntypedRelocatablePointer` or `RelocatablePointerUntyped`?

Ideally this would be the same type and just an internal member would be specialized for void.

FerdinandSpitzschnueffler: Using composition instead of inheritance should be part of a refactoring of the relocatable pointer. Therefore I added this point to the issue. I'd prefer to do the renaming in this context as well.

elBoberido: I had another idea. Will continue the discussion in the issue.

MatthiasKillat: The inheritance can be avoided, that is correct. Historically the reason is hat I started with an untyped version (as @elBoberido observed), which is currently the base. For the sake of code reuse I inherited in the typed version but I would now solve it entirely with templates, neither composite nor inheritance I think. This is a larger refactoring which I could start after 1.0, depending on priorities.
But functionality-wise it is no priority since I am not aware of bugs. I will improve the generality and the design in the future.

elBoberido: @MatthiasKillat I made a proposal in the issue itself

iceoryx_posh/test/moduletests/test_mepoo_memory_manager.cpp
```diff
@@ -30,13 +30,13 @@ using PayloadOffset_t = iox::mepoo::ChunkHeader::PayloadOffset_t;
 class MemoryManager_test : public Test
 {
   public:
-    void SetUp()
+    void SetUp() override
```
MatthiasKillat, line 33: Just a note: I guess we do not state that we override the method in most tests and while nice, it should not be mandatory for tests. For other classes we should make it a rule to mark overidden functions to benefit from additional checking of the compiler.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/atomic_relocatable_ptr.hpp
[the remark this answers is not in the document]
MatthiasKillat: Should not be done. We can compute the offset for const and non const pointers in the same way, it makes no difference. To do so we need only the non const version.

iceoryx_utils/include/iceoryx_utils/internal/relocatable_pointer/base_relative_ptr.hpp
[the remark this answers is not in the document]
MatthiasKillat: There is no rule that const methods will always return pointers to const objects. It is customary to do this if a reference or pointer to something owned by the class is returned  (e.g. ``std::vector``) and ``operator[]``), but this is not the case here. The class does not own the pointee, for this reason I opted against it (as does ``shared_ptr<T>``).

MatthiasKillat (dismissed):

iceoryx_posh/include/iceoryx_posh/internal/roudi/process_manager.hpp
```diff
@@ -209,7 +209,9 @@ class ProcessManager : public ProcessManagerInterface
     PortManager& m_portManager;
     mepoo::SegmentManager<>* m_segmentManager{nullptr};
     mepoo::MemoryManager* m_introspectionMemoryManager{nullptr};
-    RelativePointer::id_t m_mgmtSegmentId{RelativePointer::NULL_POINTER_ID};
+    rp::BaseRelativePointer::id_t m_mgmtSegmentId{rp::BaseRelativePointer::NULL_POINTER_ID};
+    mutable std::mutex m_mutex;
```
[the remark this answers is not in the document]
elBoberido: I guess this is a wrong merge resolution with the merge from master

elBoberido (commented):
please remove the mutex from `ProcessManager` and the names of the relative/relocatable header and inl files should ideally be renamed (ptr -> pointer)

elBoberido (approved):

elBoberido (approved):

elfenpiff (approved):

PR #598, opened 2021-03-09
Iox #404 Extend timeout registration info and send termination request on destruction

mossmaurice:
## Pre-Review Checklist for the PR Author

1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/best-practice-for-testing.md

## Notes for Reviewer

* Addresses all points of #404 
* Refactor `ProcessManager`
  * Move it into its own `hpp` 
  * Replace `getProcessFromList` with `searchForProcessAndThen`
* Wrap `m_prcMgr` in `smart_lock` and get rid of internal mutex

### Todo

* Remove `RouDiLock` in follow-up pull request

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

- closes #404

PR #626, opened 2021-03-22
iox-#625 set file lock directory to /tmp/

### LINKED ISSUE #625: FileLock requires write access to root owned directory /var/lock/
## Required information

**Operating system:**
Gentoo, Archlinux

`/var/lock/` is only accessible by root but the `FileLock` requires access to it. We should switch the `FileLock` directory to `/tmp` since everyone is allowed to write data here.

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #625

### DIFF
--- a/iceoryx_utils/include/iceoryx_utils/posix_wrapper/file_lock.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/posix_wrapper/file_lock.hpp
@@ -45,7 +45,7 @@ enum class FileLockError
     INTERNAL_LOGIC_ERROR,
 };
 
-constexpr char PATH_PREFIX[] = "/var/lock/";
+constexpr char PATH_PREFIX[] = "/tmp/";
 
 /// @brief Posix file lock C++ wrapping class
 ///        Following RAII, the lock is acquired on creation and released on destruction. Releasing the locks works even


FerdinandSpitzschnueffler (approved):

elBoberido (approved):
just to unblock developers who are not using ubuntu but we need to check the impact of this on QNX

iceoryx_utils/include/iceoryx_utils/posix_wrapper/file_lock.hpp
```diff
@@ -45,7 +45,7 @@ enum class FileLockError
     INTERNAL_LOGIC_ERROR,
 };
 
-constexpr char PATH_PREFIX[] = "/var/lock/";
+constexpr char PATH_PREFIX[] = "/tmp/";
```
mossmaurice, line 48: According to the [file system hierarchy standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) defined by the Linux foundation, lock files [shall](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s09.html) be placed under `/var/locks`. Seems like your distribution decided to choose another location. Mac OS prefers [`/var/tmp/`](https://github.com/eclipse-iceoryx/iceoryx/pull/598/files#diff-65a31f3463e5bc0dce797aecc310342c1fbf2487fc602dcb8bd80019d5257640R48-R52), this will be available with #598 

QNX advises to use [`/var/locks`](http://www.qnx.com/developers/docs/qnxcar2/index.jsp?topic=%2Fcom.qnx.doc.neutrino.user_guide%2Ftopic%2Ffiles_var.html), but in our experience so far, it depends to the image, that is created.

Could you please add this with an `#ifdef` for common platforms? `installation.md#QNX` currently states that `/var/locks` is used, in case you change it please update it there.

mossmaurice (changes requested):

iceoryx_utils/include/iceoryx_utils/posix_wrapper/file_lock.hpp
elfenpiff: @mossmaurice sorry I read your comment after I merged the pull request :( ... but I will handle this comment nevertheless.

> According to the file system hierarchy standard defined by the Linux foundation, lock files shall be placed under /var/locks

This is correct but the access rights of `/var/lock` is in Gentoo, ArchLinux and OpenSuSe like the following:
```
drwxrwxr-x  5 root root    120 Mar 22 15:26 lock
# /var
```

This means the user which is starting RouDi has to be at least in the root group. On the other hand `/tmp/` is for all users always read and writable and since the applications and roudi has started as an arbitrary user it is the perfect spot for the lock file in QNX, Mac OS and Linux. If you insist in using `/var/lock` for the lock file we have to do the following things.

1. Write `#ifdef` for the platforms, instead of using `/tmp` which works for every platform (except Windows of course).
2. Add a chapter in the `installation.md` on howto add the user which is executing roudi to the lock management group of Linux, Mac OS (this will be a pain) and QNX.


If you do not object I would let the directory be `/tmp` and would add this to the `installation.md` since the file system hierarchy standard states the directory structure of a linux system and not where an application has to store the files. I admit it would be best to use the folders accordingly but it seems that on some linux systems we do not have access to `/var/lock` but we will always have access to `/tmp`.

By the way this also goes for QNX. The QNX boards of the past where also exactly in this way configured that everything was read only and we only had write access to `/tmp`.

PR #635, opened 2021-03-24
Iox #14 custom header for typed api

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

Finally, the custom header can be accessed by the typed API

## References

- Partly closes #14


iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
```diff
@@ -31,22 +31,31 @@ namespace popo
...
 {
     static_assert(!std::is_void<T>::value, "Type must not be void. Use the UntypedPublisher for void types.");
-    static_assert(std::is_default_constructible<T>::value, "The Publisher requires default-constructable types.");
+    static_assert(!std::is_void<H>::value, "Custom header must not be void.");
+
+    /// @todo clarify in the PR if this can be removed for T.
+    /// This restriction is not necessary anymore, since `loan` takes ctor arguments.
+    static_assert(std::is_default_constructible<T>::value, "The Publisher requires default-constructable data types.");
+    static_assert(std::is_default_constructible<H>::value,
+                  "The Publisher requires default-constructable custom header types.");
```
elBoberido, line 55: @MatthiasKillat I guess this can be removed, right?

iceoryx_posh/include/iceoryx_posh/internal/popo/publisher.inl
```diff
@@ -51,58 +51,59 @@ inline cxx::expected<AllocationError> Publisher<T, base_publisher_t>::publishRes
 
     return loanSample().and_then([&](auto& sample) {
         c(sample.get(), std::forward<ArgTypes>(args)...);
-        sample.publish();
+        publish(std::move(sample));
```
elfenpiff, line 54: Why are you using this approach instead of `sample.publish`. Is `sample.publish` going to be removed?

iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
```diff
@@ -31,22 +31,31 @@ namespace popo
...
+
+    static_assert(!std::is_const<T>::value, "The type must not be const.");
```
elfenpiff, line 57: Oh what about
1. when T or H is a reference?  (Here you send a sample with a reference to process local memory?!)
2. when T or H is a pointer? (Here you send a sample with a pointer to process local memory?!)

Could we get rid of the const check by using `std::remove_const<T>::type`. Then we can get rid of this `static_assert` and I think we can document this in the API and it wouldn't make the API more error prone ...

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -66,22 +67,29 @@ struct SamplePrivateData<const T>
 /// @brief The Sample class is a mutable abstraction over types which are written to loaned shared memory.
 /// These samples are publishable to the iceoryx system.
 ///
-template <typename T>
+template <typename T, typename H = typename cxx::add_const_conditionally<mepoo::NoCustomHeader, T>::type>
 class Sample
 {
+    static_assert(std::is_const<T>::value == std::is_const<H>::value,
```
elfenpiff, line 73: According to another `static_assert` they both have to be non const?! Maybe we should centralize this in some kind of concept or at least using alias?!

Or did I miss something?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -146,20 +148,33 @@ class Sample
...
     /// @return The const ChunkHeader of the underlying memory chunk.
     ///
     const mepoo::ChunkHeader* getHeader() const noexcept;
 
+    ///
+    /// @brief Retrieve the custom header of the underlying memory chunk loaned to the sample.
+    /// @return The custom header of the underlying memory chunk.
+    ///
+    template <typename R = H, typename = HasCustomHeader<R, H>>
+    R& getCustomHeader() noexcept;
```
elfenpiff, line 169: Do we also require a `using ConditionConstCustomHeader`?

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
```diff
@@ -146,20 +148,33 @@ class Sample
     ///
     const T* get() const noexcept;
 
+    using ConditionalConstChunkHeader_t = typename cxx::add_const_conditionally<mepoo::ChunkHeader, T>::type;
```
elfenpiff, line 151: Could you please explain in a short doxygen comment why we need a `ConditionalConstChunkHeader`?

iceoryx_posh/test/moduletests/test_popo_sample.cpp
```diff
@@ -83,3 +88,38 @@ TEST_F(SampleTest, PublishesSampleViaPublisherInterface)
     // ===== Verify ===== //
     // ===== Cleanup ===== //
 }
+
+TEST_F(SampleTest, getCutomHeaderFromNonConstTypeReturnCorrectAddress)
```
elfenpiff, line 92: Give me three ALL GLORY TO THE HYPNOTOAD

and then add an S to getCutom

iceoryx_posh/test/moduletests/test_popo_sample.cpp
```diff
@@ -83,3 +88,38 @@ TEST_F(SampleTest, PublishesSampleViaPublisherInterface)
...
+    // ===== Test ===== //
+    auto& customHeader = sut.getCustomHeader();
+
+    // ===== Verify ===== //
+    ASSERT_EQ(&customHeader, chunk.customHeader());
+
+    // ===== Cleanup ===== //
+}
+
+TEST_F(SampleTest, getCutomHeaderFromConstTypeReturnCorrectAddress)
```
elfenpiff, line 110: Name the 3 best Schwarzenegger Movies

and then add an S to getCutom

iceoryx_utils/include/iceoryx_utils/cxx/type_traits.hpp
```diff
@@ -24,6 +24,22 @@ namespace iox
...
+template <typename T, typename Condition>
+struct add_const_conditionally
+{
+    using type = T;
+};
+template <typename T, typename Condition>
+struct add_const_conditionally<T, const Condition>
+{
+    using type = const T;
+};
```
elfenpiff, line 41: Here you could add
```cpp
template<typename T, typename Condition>
using add_const_conditionally_t = typename add_const_conditionally<T, Condition>::type;
```

for better usability.

iceoryx_posh/include/iceoryx_posh/internal/popo/publisher.inl
elBoberido: I guess that's an oddment from an intermediate state when `sample.publish` didn't work. Will fix it

FerdinandSpitzschnueffler: same in line 63

iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
elBoberido: I'm not sure about this. In the end it's a wrong usage since the data cannot be altered if it's const. We could remove the const, but it might be surprising even if we document it.

We probably should also catch pointer and references since this is also a clear misuse.

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: No, since there is the static assert on line 73 which ensures that `T` and `H` are equal in their const qualifiers.

The `ConditionalConstChunkHeader_t` is necessary since the sample can be constructed from a publisher with non-const `T` or from a subscriber with `const T`. When it's constructed from the publisher it mutable, at least that's the current state. There is a task item to restrict the access to the `ChunkHeader`.

I'll add this as documentation to the `ConditionalConstChunkHeader_t`

iceoryx_posh/include/iceoryx_posh/popo/sample.hpp
elBoberido: You are referring to the `static_assert` in the publisher, right? That's correct. This assert just checks if the types are equal in their const qualifiers, so the assertion in the publisher is still valid.

iceoryx_posh/test/moduletests/test_popo_sample.cpp
```diff
@@ -83,3 +88,42 @@ TEST_F(SampleTest, PublishesSampleViaPublisherInterface)
...
+    // ===== Setup ===== //
+    ChunkMock<DummyData, DummyHeader> chunk;
+    iox::cxx::unique_ptr<DummyData> testSamplePtr{chunk.sample(), [](DummyData*) {}};
+    MockPublisherInterface<DummyData, DummyHeader> mockPublisherInterface{};
+
+    auto sut = iox::popo::Sample<DummyData, DummyHeader>(std::move(testSamplePtr), mockPublisherInterface);
+
+    // this line is actually not needed for the tests but if it is not present gmock raises a unused-function warning in
+    // gmock-spec-builders.h which breaks the build with -Werror
+    EXPECT_CALL(mockPublisherInterface, publishMock).Times(0);
```
elBoberido, line 103: oh boy, this was so much fun to find :laughing:

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
```diff
@@ -24,116 +24,125 @@ namespace popo
...
 {
     if (m_members.sampleUniquePtr)
     {
         m_members.publisherRef.get().publish(std::move(*this));
     }
-
     else
     {
-        /// @todo Notify caller of attempt to publish invalid chunk. Or something ?
+        LogWarn() << "Tried to publish empty sample!";
```
elBoberido, line 140: I was not quite sure if logging is enough or if it maybe should be an `LogError`. What do you think?

iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
elfenpiff: Okay I am fine the const handling like it is now.
But two additional `static_assert` for references and pointers should be implemented here.

iceoryx_posh/include/iceoryx_posh/internal/popo/sample.inl
elfenpiff: First I thought this can only happen when the developer tries to publish an already moved or published sample. In this case we could write "Trying to publish already published or moved sample!". But we could also construct an empty sample :( ... and this could also occur.

I would you should print a `LogError` message here and call additionally the errorHandler with a moderate warning since nothing gets corrupted here but it could indicate a kaputt algorithm.

elfenpiff (commented):
Please add a moderate errorHandler to sample publish with a `LogError` message

FerdinandSpitzschnueffler (approved):

elfenpiff (approved):

PR #639, opened 2021-03-24
Iox #623 use 8 byte type for chunk management pointer and add tests

### LINKED ISSUE #623: Refactoring of UsedChunkList
## Brief feature description

This whole list is error prone, uses the STL even though it does not have to use it (we should use here a ``cxx::vector``) and it is not tested.

## Detailed information

- [x] separate declaration and implementation
- [x] add unit tests
- [ ] use ``cxx::vector`` instead of ``std::array``
- [ ] store sharedChunks? (see https://github.com/eclipse-iceoryx/iceoryx/pull/618#discussion_r597591727)
- [ ] change signature of ``remove`` to ``mepoo::SharedChunk remove(const mepoo::ChunkHeader& chunkHeader)``
- [ ] decide if the `UsedChunkList` should behave like a `Set` and reject to take the same data a second time

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR moves the implementation of the `UsedChunkList` into an inl file and adds tests.

Additionally, with the introduction of the relative pointer, the `UsedChunkList` had a small defect since it relies on always valid pointers in the data array. Since the relativ pointer is 16 bytes the pointer could be invalid if the application died at the wrong time. This is also taken care of with this PR.

## References

- Partially closes #623

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
@@ -125,7 +125,7 @@ class UsedChunkList
     std::atomic_flag m_synchronizer = ATOMIC_FLAG_INIT;
     uint32_t m_usedListHead{INVALID_INDEX};
     uint32_t m_freeListHead{0u};
-    uint32_t m_listNodes[Capacity];
+    uint32_t m_listIndices[Capacity];
     DataElement_t m_listData[Capacity];
 };
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
@@ -42,6 +42,8 @@ namespace popo
 template <uint32_t Size>
 class UsedChunkList
 {
+    static_assert(Size > 0, "UsedChunkList Size must me larger than 0!");
+
   public:
     /// @brief Constructs a default UsedChunkList
     UsedChunkList() noexcept;
@@ -69,7 +71,7 @@ class UsedChunkList
     bool freeSpaceInList() const noexcept;
 
   private:
-    static constexpr uint32_t InvalidIndex{Size};
+    static constexpr uint32_t INVALID_INDEX{Size};
 
     // this shall be moved to the RelativePointer implementation
     struct RelativePointerData
@@ -112,7 +114,7 @@ class UsedChunkList
 
   private:
     std::atomic_flag m_synchronizer = ATOMIC_FLAG_INIT;
-    uint32_t m_usedListHead{InvalidIndex};
+    uint32_t m_usedListHead{INVALID_INDEX};
     uint32_t m_freeListHead{0u};
     uint32_t m_listNodes[Size];
     DataElement_t m_listData[Size];
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
@@ -109,13 +109,13 @@ class UsedChunkList
             return m_idAndOffset == LOGICAL_NULLPTR;
         }
 
-        static constexpr uint64_t MAX_ID{std::numeric_limits<uint16_t>::max()};
+        static constexpr uint16_t MAX_ID{std::numeric_limits<uint16_t>::max()};
         /// @note the id is 16 bit and the offset consumes the remaining 48 bits -> max offset is 2^48 - 1
         static constexpr uint64_t MAX_OFFSET{(1ULL << 48U) - 1U};
         static constexpr uint64_t LOGICAL_NULLPTR{std::numeric_limits<uint64_t>::max()};
 
       private:
-        uint64_t m_idAndOffset;
+        uint64_t m_idAndOffset{LOGICAL_NULLPTR};
     };
 
     using DataElement_t = RelativePointerData;
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
@@ -39,10 +39,10 @@ namespace popo
 ///        accessed. Additionally, the type stored is this array must be less or equal to 64 bit in order to write it
 ///        within one clock cycle to prevent torn writes, which would corrupt the list and could potentially crash
 ///        RouDi.
-template <uint32_t Size>
+template <uint32_t Capacity>
 class UsedChunkList
 {
-    static_assert(Size > 0, "UsedChunkList Size must me larger than 0!");
+    static_assert(Capacity > 0, "UsedChunkList Capacity must me larger than 0!");
 
   public:
     /// @brief Constructs a default UsedChunkList
@@ -71,7 +71,7 @@ class UsedChunkList
     bool freeSpaceInList() const noexcept;
 
   private:
-    static constexpr uint32_t INVALID_INDEX{Size};
+    static constexpr uint32_t INVALID_INDEX{Capacity};
 
     // this shall be moved to the RelativePointer implementation
     struct RelativePointerData
@@ -80,7 +80,7 @@ class UsedChunkList
             : segment(MAX_SEGMENT)
             , offset(MAX_OFFSET)
         {
-            static_assert(sizeof(RelativePointerData) <= 8U, "The RelativePointerData must not exceed 64 bit!");
+            static_assert(sizeof(RelativePointerData) <= 8U, "The RelativePointerData size must not exceed 64 bit!");
             static_assert(std::is_trivially_copyable<RelativePointerData>::value,
                           "The RelativePointerData must be trivially copyable!");
         }
@@ -116,8 +116,8 @@ class UsedChunkList
     std::atomic_flag m_synchronizer = ATOMIC_FLAG_INIT;
     uint32_t m_usedListHead{INVALID_INDEX};
     uint32_t m_freeListHead{0u};
-    uint32_t m_listNodes[Size];
-    DataElement_t m_listData[Size];
+    uint32_t m_listNodes[Capacity];
+    DataElement_t m_listData[Capacity];
 };
 
 } // namespace popo
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,7 +22,6 @@
 #include "iceoryx_posh/mepoo/chunk_header.hpp"
 #include "iceoryx_utils/internal/relocatable_pointer/relative_pointer.hpp"
 
-#include <array>
 #include <atomic>
 #include <cstdint>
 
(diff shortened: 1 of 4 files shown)


iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
-            return false;
-        }
-    }
+  private:
+    static constexpr uint32_t INVALID_INDEX{Capacity};
 
-    // only from runtime context
-    bool remove(const mepoo::ChunkHeader* f_chunkHeader, mepoo::SharedChunk& f_chunk)
+    // this shall be moved to the relative_ptr implementation
+    struct relativ_ptr_data
```
elBoberido, line 77: this should be consolidated with the relative pointer refactoring

iceoryx_posh/test/moduletests/test_popo_used_chunk_list.cpp
```diff
@@ -0,0 +1,300 @@
...
+    EXPECT_TRUE(sut.insert(getChunkFromMemoryManager()));
+}
+
+TEST_F(UsedChunkList_test, AddSameChunkTwiceWorks)
+{
+    auto chunk = getChunkFromMemoryManager();
+    sut.insert(chunk);
+
+    EXPECT_TRUE(sut.insert(chunk));
+}
```
elBoberido, line 96: This checks the current behavior. If we want a `Set` like behavior, this needs to be done in a follow up PR.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
-                // insert index to free list
-                m_list[current] = m_freeListHead;
-                m_freeListHead = current;
-
-                //@todo can we do this cheaper with a global fence in cleanup?
-                m_synchronizer.clear(std::memory_order_release);
-                return true;
-            }
-            previous = current;
+            static_assert(sizeof(relativ_ptr_data) <= 8U, "The relativ_ptr_data size must not exceed 64 bit!");
```
elBoberido, line 83: oh no, Windows doesn't like this

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
-            {
-                mepoo::SharedChunk{data};
-            }
+            cxx::Ensures(offset < MAX_OFFSET && "offset must not exceed MAX:OFFSET!");
         }
 
-        init(); // just to save us from the future self
-    }
+        uint16_t segment : 16;
+        uint64_t offset : 48;
```
MatthiasKillat, line 95: In theory the offset can be larger (depending on your max segment size however it will not be, I do not think we have segments larger than 2^48 Byte) so it should be ok. A problem is that those constants are not derived from constants defining those limits of #segments and max size of a segment.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
-            m_list[Size - 1u] = InvalidIndex; // just to save us from the future self
-        }
-        else
+        bool isNullptr() const noexcept
         {
-            m_list[0u] = InvalidIndex;
+            return segment == MAX_SEGMENT && offset == MAX_OFFSET;
         }
 
+        static constexpr uint64_t MAX_SEGMENT{std::numeric_limits<uint16_t>::max()};
```
MatthiasKillat, line 108: Those constants probably belong somewhere else where the shared memory parameters/bounds are defined. But ok for the time being I guess.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
-    bool freeSpaceInList()
-    {
-        return (m_freeListHead != InvalidIndex);
-    }
-
-
-  private:
-    static constexpr uint32_t InvalidIndex{Size};
+    using DataElement_t = RelativePointerData;
+    static constexpr DataElement_t DATA_ELEMENT_NULLPTR{};
```
MatthiasKillat, line 113: I guess this is the critical part since we need to be able to locate the chunks from app side and Roudi side for cleanup (if required). But with two IndexQueues and an array you would not have the restriction of a 64 Bit (pseudo) serialization of the relative pointer but could use them directly as before in the array I think.

Might cost a little performance and require more space I suppose.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
 
   private:
     std::atomic_flag m_synchronizer = ATOMIC_FLAG_INIT;
-    uint32_t m_usedListHead{InvalidIndex};
+    uint32_t m_usedListHead{INVALID_INDEX};
     uint32_t m_freeListHead{0u};
-    std::array<uint32_t, Size> m_list;
-    std::array<rp::RelativePointer<mepoo::ChunkManagement>, Size> m_data;
+    uint32_t m_listNodes[Capacity];
```
MatthiasKillat, line 119: I find the naming misleading. It is not a node, it is an index to the m_listData array. Its elements encode the next index in the (implicit) list(s).

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
dkroenke: I guess there are some implementation-defined bitfields in the size of structs: https://stackoverflow.com/questions/42314104/clang-and-msvc-report-different-sizeof-for-the-same-struct

I found out that it has on Windows size 16, maybe we do (an ugly) ifdef here or we need to find another way around.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
dkroenke: The renaming too?

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
```diff
@@ -0,0 +1,170 @@
...
+{
+namespace popo
+{
+template <uint32_t Capacity>
+constexpr typename UsedChunkList<Capacity>::DataElement_t UsedChunkList<Capacity>::DATA_ELEMENT_NULLPTR;
+
+template <uint32_t Capacity>
+UsedChunkList<Capacity>::UsedChunkList() noexcept
+{
+    static_assert(sizeof(DataElement_t) <= 8U, "The size of the data element type must not exceed 64 bit!");
```
dkroenke, line 31: Same Error with the sizeof in windows will occur here

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
```diff
@@ -0,0 +1,170 @@
...
+        auto segment = ptr.getId();
+        auto offset = ptr.getOffset();
+        cxx::Ensures(segment < DataElement_t::MAX_SEGMENT && "relative_ptr Id must fit into segment type!");
+        cxx::Ensures(offset < DataElement_t::MAX_OFFSET && "relative_ptr offset must fit into offset type!");
+        m_listData[m_usedListHead] = DataElement_t(static_cast<uint16_t>(segment), offset);
+
+        // set freeListHead to the next free entry
+        m_freeListHead = nextFree;
+
+        /// @todo can we do this cheaper with a global fence in cleanup?
```
MatthiasKillat, line 61: I we only use this from two contexts and can enforce that all writes of context 1 are made visible in context 2 (Roudi) when it needs to clean up then it should be sufficient to synchronize with a fence on cleanup. But this assumes that the values in the cache of the core where the application crashed on are still available. I do not know how this behaves across multiple applications, they might be cleaned up without being propagated to main memory or staying in cache.
The way it is should ensure at most one chunk (per list) can be leaked.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,107 @@
...
-        }
-        else
+        bool isNullptr() const noexcept
         {
-            m_list[0u] = InvalidIndex;
+            return segment == MAX_SEGMENT && offset == MAX_OFFSET;
         }
 
+        static constexpr uint64_t MAX_SEGMENT{std::numeric_limits<uint16_t>::max()};
+        static constexpr uint64_t MAX_OFFSET{(1ULL << 48) - 1};
```
dkroenke, line 109: Where does the offset 48 come from?

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+template <uint32_t Capacity>
+bool UsedChunkList<Capacity>::remove(const mepoo::ChunkHeader* chunkHeader, mepoo::SharedChunk& chunk) noexcept
+{
+    auto previous = INVALID_INDEX;
+
+    // go through usedList with stored chunks
+    for (auto current = m_usedListHead; current != INVALID_INDEX; current = m_listNodes[current])
```
MatthiasKillat, line 77: usedIndex might be more precise, it must be clear that we iterate over the used list

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
```diff
@@ -0,0 +1,170 @@
...
+    {
+        if (!m_listData[current].isNullptr())
```
MatthiasKillat, line 79: Can this even happen? Do we store logical nullptrs? There are no nullptr chunks. But ok, play it defensively even if it cannot happen. (If it cannot happen this is some kind of invariant I was talking about we should document to increase performance without losing comprehensibility).

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
```diff
@@ -0,0 +1,170 @@
...
+}
+
+template <uint32_t Capacity>
+void UsedChunkList<Capacity>::cleanup() noexcept
+{
+    m_synchronizer.test_and_set(std::memory_order_acquire);
+
+    for (auto& data : m_listData)
+    {
+        if (!data.isNullptr())
```
MatthiasKillat, line 121: here we need the check since it can be a logical nullptr for unused entries

iceoryx_posh/test/moduletests/test_popo_used_chunk_list.cpp
MatthiasKillat: I would opt against a set considering how it is used, since it would imply an error in the memory manager I think (it would have to provide a used chunk again). This adds unnecessary complexity and a performance loss for no gain. We must ensure that memory management works correctly.

As a side note: consider the pair malloc and free. Except for very simple sanity checks,free does not check for double free errors for good reason: the cost would simply be prohibitively high and there is no need if used correctly.

iceoryx_posh/test/moduletests/test_popo_used_chunk_list.cpp
```diff
@@ -0,0 +1,305 @@
...
+    sut.insert(chunk);
+
+    SharedChunk removedChunk;
+    EXPECT_TRUE(sut.remove(chunkHeader, removedChunk));
+    EXPECT_TRUE(removedChunk);
+
+    checkIfEmpty();
+}
+
+TEST_F(UsedChunkList_test, RemoveSameChunkAddedTwiceWorks)
```
MatthiasKillat, line 135: It should as it is implemented but we do not require this for our use case. Good to test anyway.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
elBoberido: since the segment is 16 bits wide, it's the remaining bits to not exceed 64 bits. Will make it clear in the code though.

iceoryx_posh/test/moduletests/test_popo_used_chunk_list.cpp
elBoberido: it's also no problem to insert the same data multiple times, one just hast to remove it as often as it was inserted

MatthiasKillat:
I think it would be possible to instead use just the data array and a hash on the pointer (for simplicity just % capacity for now, it will not be uniform/perfect since the pointers will not be but this can be improved). We would still use the 8 byte storage type for array elements.
We would even gain performance during removal, and lose some during insertion. Collision handling would be linear.
This means both can degenerate into linear search of half the array on average and the complete array in the worst case.

This is also the case now for removal though, hence why we gain something here. Insertion is O(1) now and would not be afterwards, even on average if we assume the hashtable is just large enough to hold all elements (otherwise we would trade time for space, but probably not a good idea here since the capacity will be relatively small and brute force insertion is viable I think).

cleanup would not change, iterate over the array and delete all valid entries.

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,117 @@
...
-    {
-        return (m_freeListHead != InvalidIndex);
-    }
+        static constexpr uint64_t MAX_ID{std::numeric_limits<uint16_t>::max()};
+        /// @note the id is 16 bit and the offset consumes the remaining 48 bits -> max offset is 2^48 - 1
+        static constexpr uint64_t MAX_OFFSET{(1ULL << 48U) - 1U};
+        static constexpr uint64_t LOGICAL_NULLPTR{std::numeric_limits<uint64_t>::max()};
 
+      private:
+        uint64_t m_idAndOffset;
```
dkroenke, line 119: Default value zero

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
```diff
@@ -21,161 +22,117 @@
...
         }
 
-        m_synchronizer.clear(std::memory_order_release);
-    }
-
-    bool freeSpaceInList()
-    {
-        return (m_freeListHead != InvalidIndex);
-    }
+        static constexpr uint64_t MAX_ID{std::numeric_limits<uint16_t>::max()};
```
dkroenke, line 113: Could MAX_ID maybe defined as uint16_t here? Then you have not to do the static_cast in `id()`

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
```diff
@@ -0,0 +1,170 @@
...
+{
+    static_assert(sizeof(DataElement_t) <= 8U, "The size of the data element type must not exceed 64 bit!");
+    static_assert(std::is_trivially_copyable<DataElement_t>::value,
+                  "The data element type must be trivially copyable!");
+
+    init();
+}
+
+template <uint32_t Capacity>
+bool UsedChunkList<Capacity>::insert(mepoo::SharedChunk chunk) noexcept
```
dkroenke, line 39: chunk can be const

elBoberido: no, since it will be released

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
elBoberido: I will change it to `m_listIndices`

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.inl
elBoberido: logical nullptr are stored, like in the `RelativePointer` I made it more clear by the documentation.

dkroenke (approved):

iceoryx_posh/include/iceoryx_posh/internal/popo/used_chunk_list.hpp
MatthiasKillat: Everything that is necessary for a lean design with names fitting the general codestyle (I hope we can adopt snake_case for utils classes, in line with STL).

iceoryx_posh/test/moduletests/test_popo_used_chunk_list.cpp
MatthiasKillat: Yes, but I just should not happen the way we use it and I do not like guarding against at significant costs (complexity, performance).

MatthiasKillat (approved):

PR #643, opened 2021-03-25
Iox #583 clean up doxygen in dds

### LINKED ISSUE #583: Clean-up Doxygen for all methods in headers
As discussed in the last iox dev meeting (2021-02-18) Doxygen comments for all methods in all headers need to be checked for completeness and cleaned-up if necessary.

It should look like this
```
/// @brief
/// @param[in]
///  ...
/// @return
```

Comment should have:
-  A brief description what's done
- Description of all incoming and return values
- Also check that the param[in] name matches the param name in declaration

Ex1:
```
/// @brief is the service still offered
/// @param[in] self handle of the publisher
/// @return true is the service is offered otherwise false
bool iox_pub_is_offered(iox_pub_t const self);
```

Ex2:
```
/// @brief Doing a thing
void doIt();
```

marthtz:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
Clean-up Doxygen in `iceoryx_dds`

## References

- Partially closes #583


iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -33,28 +33,68 @@ namespace dds
...
+
+    /// @brief move assignment is explicitly deleted
     CycloneDataReader& operator=(CycloneDataReader&&) = delete;
 
+    /// @brief Connect cylcone data reader to the underlying DDS network
     void connect() noexcept override;
 
+    /// @brief Get the size of the next sample if one is available
+    /// @return an optional pointer to the size of the next sample if one is available, cxx::nullopt_t if value is not
+    /// initialized
```
ithier, line 67: If we document overridden methods isn't this duplication ?

marthtz: Good point. So, what's our strategy here?

ithier: @dkroenke

dkroenke: @ithier @marthtz
I would advise to do the following:
The virtual methods in a base class need to be fully documented. Example here: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/roudi/memory/memory_block.hpp

For the overridden methods you can reference to the base class method, but only if that is fully documented. It should be explicitly mentioned that this is an implementation of an virtual method.
https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/roudi/memory/mempool_collection_memory_block.hpp

With that, doxygen generates a link to the base class method:
image

marthtz: Agreed, less duplication means less potential errors. However, do you want that applied to the `@brief` section only? So, the implementation should say `@brief Implementation of ` and the virtual one a full description. But what about `@param` and `@return`? Should they stay the same/duplicated or omitted in the implementation?
The example you referenced is slightly unclear/-clean for the return type (e.g. why `type T`, it's not a template?!):
```cpp
    /// @return the required memory as multiple of the alignment
    virtual uint64_t size() const noexcept = 0;
``` 
vs.
```cpp
    /// @return the size of type T
    uint64_t size() const noexcept override;
```

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -33,28 +33,68 @@ namespace dds
 class CycloneDataReader : public DataReader
 {
   public:
+    /// @brief delete default ctor
```
elfenpiff, line 36: In my opinion, it makes no sense to document deleted methods. They are not available for the user and why document something which is not present. 
But it would be helpful to document why they were deleted!

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -33,28 +33,68 @@ namespace dds
...
+    /// @param[in] instanceId ID of the instance of the service
+    /// @param[in] eventId ID of the event
     CycloneDataReader(const capro::IdString_t serviceId,
                       const capro::IdString_t instanceId,
                       const capro::IdString_t eventId) noexcept;
+
+    /// @brief Virtual destructor
     virtual ~CycloneDataReader();
 
+    /// @brief copy ctor is explicitly deleted
```
elfenpiff, line 50: Please remove the documentation for deleted methods and maybe replace them with an explanation on why they are deleted.
Please do this for all deleted methods.

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
elfenpiff: @dkroenke @ithier I would partially disagree. Isn't the idea of an overridden method that it can behave differently than the original one?! And when it behaves differently shouldn't it be documented?

In the example from @marthtz it is not required but think of policies for instances.

```cpp
struct LockPolicy {
    virtual void lock() = 0;
    virtual void unlock() = 0;
};

struct ThreadSafe : public LockPolicy {
    /// @brief locks a mutex
    void lock() override;
    /// @brief unlocks a mutex
    void unlock() override;
};

struct SingleThreaded : public LockPolicy {
    /// @brief does nothing
    void lock() override;
    /// @brief does nothing
    void unlock() override;
};
```

@dkroenke with your suggestion we would document the abstract base class fully (which can most likely make no statement on all the possible behaviors) and in the specialization, you would refer to the base class and the user then does not know if the lock is blocking or not.

To summarize. The rule should be: use common sense and when the implementation behaves differently for the user it has to be documented.

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_writer.hpp
```diff
@@ -32,21 +32,53 @@ namespace dds
 class CycloneDataWriter : public iox::dds::DataWriter
 {
   public:
+    /// @brief delete default ctor
```
elfenpiff, line 35: Please do not document deleted methods and add an explanation on why they are deleted.

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -33,28 +33,68 @@ namespace dds
...
+    /// @brief Connect cylcone data reader to the underlying DDS network
     void connect() noexcept override;
 
+    /// @brief Get the size of the next sample if one is available
+    /// @return an optional pointer to the size of the next sample if one is available, cxx::nullopt_t if value is not
+    /// initialized
     iox::cxx::optional<uint32_t> peekNextSize() override;
+
+    /// @brief Check if new samples are ready to take
+    /// @return bool, true if length of samples is not null, else false
```
elfenpiff, line 71: This sounds like the description of the implementation. What is the length of samples?! What about:
`returns true if there are samples available, otherwise false`

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -33,28 +33,68 @@ namespace dds
...
     bool hasSamples() override;
+
+    /// @brief Take the next available sample from the DDS data space.
+    /// @param[in] buffer Receive buffer in which sample will be stored.
+    /// @param[in] bufferSize Size of the provided buffer.
+    /// @return Error if unsuccessful.
```
elfenpiff, line 77: Maybe document the error better. Is it an error when there are no samples available or is it only an error when there is not enough space in `buffer`?

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -33,28 +33,68 @@ namespace dds
...
 
+    /// @brief Take up to a maximum number of samples from the DDS data space.
+    /// @param[in] buffer Receive buffer in which samples will be stored.
+    /// @param[in] bufferSize The size of the buffer (in bytes).
+    /// @param[in] maxSamples The maximum number of samples to request from the network.
+    /// @return Number of samples taken if successful. Number of samples will be in the range [0,maxSamples].
     iox::cxx::expected<uint64_t, DataReaderError>
     take(uint8_t* const buffer, const uint64_t& bufferSize, const iox::cxx::optional<uint64_t>& maxSamples) override;
 
+    /// @brief Get service id
```
elfenpiff, line 88: @dkroenke How is our policy when it comes to documenting tiny getter methods without side-effects.
In this case, you write the same text twice only with different words. Once in `@brief` and once in `@return`. Here I would also apply the common sense rule and just write something into the `@brief` tag.

iceoryx_dds/include/iceoryx_dds/dds/data_reader.hpp
```diff
@@ -38,66 +38,51 @@ enum class DataReaderError : uint8_t
...
     virtual iox::cxx::optional<uint32_t> peekNextSize() = 0;
 
-    ///
     /// @brief hasSamples Checks if new samples ready to take.
     /// @return True if new samples available.
-    ///
     virtual bool hasSamples() = 0;
 
-    ///
     /// @brief take Take the next available sample from the DDS data space.
```
elfenpiff, line 57: Could you please correct this line. `takeNext` takes the next ....

iceoryx_dds/include/iceoryx_dds/dds/data_reader.hpp
```diff
@@ -38,66 +38,51 @@ enum class DataReaderError : uint8_t
...
 
-    ///
-    /// @brief getEventId
+    /// @brief get event id
     /// @return The ID of the event producing the data
-    ///
     virtual capro::IdString_t getEventId() const noexcept = 0;
 
   protected:
+    /// @brief Protected default constructor
```
elfenpiff, line 85: You do not need to document methods that are not available for the public API.

iceoryx_dds/include/iceoryx_dds/dds/data_writer.hpp
```diff
@@ -27,44 +27,36 @@ namespace dds
...
 class DataWriter
 {
   public:
+    /// @brief Virtual default destructor
     virtual ~DataWriter() = default;
 
-    ///
-    /// @brief connect Connect the DataWriter to the underlying DDS network.
-    ///
+    /// @brief Virtual Function to connect the DataWriter to the underlying DDS network.
```
elfenpiff, line 38: C++ has no virtual functions only virtual methods.
To provide you some context: `DataWriter` is an abstract base class that enforces a certain interface like for instance that every class which inherits from `DataWriter` has to implement a `connect` method.

I would get rid of the `Virtual` keyword at all since this is something the user sees directly in the signature of the method.

iceoryx_dds/include/iceoryx_dds/dds/data_writer.hpp
```diff
@@ -27,44 +27,36 @@ namespace dds
 {
 ///
 /// @brief Abstraction for DDS Data Writers.
-/// Provides the minimum functionality required for posh-dds gateway implementations.
+/// @note Provides the minimum functionality required for posh-dds gateway implementations.
 ///
 class DataWriter
 {
   public:
+    /// @brief Virtual default destructor
```
elfenpiff, line 35: @dkroenke do we really have to document this. Comments like
`@brief default construct` or `@brief default destructor` deliver absolutely no value because it's exactly what the signature says since it is the default constructor or the default destructor. Here again, I would use common sense. The comment should be helpful for the developer and not clutter the code which can be counterproductive.

Otherwise, we end up with stuff like this:
```cpp
// assign zero to integer a
int a = 0;
```

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
```diff
@@ -37,14 +38,32 @@ class DDS2IceoryxGateway : public gateway_t
...
+    /// @brief Default constructor
     DDS2IceoryxGateway() noexcept;
+
+    /// @brief Set DDS to iceoryx gateway object
+    /// @param[in] channelFactory the channel factory to propagate data across
     DDS2IceoryxGateway(ChannelFactory channelFactory) noexcept;
+
+    /// @brief Load the provided gateway configuration
+    /// @param[in] config the gateway configuration to load
     void loadConfiguration(const config::GatewayConfig& config) noexcept;
```
elfenpiff, line 50: What happens when it fails to load the configuration?

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
```diff
@@ -34,14 +35,28 @@ template <typename channel_t = gw::Channel<popo::UntypedSubscriber, dds::data_wr
 class Iceoryx2DDSGateway : public gateway_t
 {
   public:
+    /// @brief Default constructor
     Iceoryx2DDSGateway() noexcept;
+
+    /// @brief Load the provided gateway configuration
+    /// @param[in] config The gateway configuration to load
     void loadConfiguration(const config::GatewayConfig& config) noexcept;
```
elfenpiff, line 43: What happens when it fails to load the configuration?

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
```diff
@@ -34,14 +35,28 @@ template <typename channel_t = gw::Channel<popo::UntypedSubscriber, dds::data_wr
...
+
+    /// @brief Discover messages coming from iceoryx.
+    /// @param[in] msg The discovery message.
     void discover(const capro::CaproMessage& msg) noexcept;
```
elfenpiff, line 47: What happens whet it fails to setup a channel?

elfenpiff (changes requested):

elfenpiff:
Please do not write commit messages like: `fix comments` - what comments, what did they criticize? this is not helpful. The code and the features should be traceable (in this case it's only docu so it's not critical but nevertheless.) 
The same goes for `fix review finding`.

Please state what you did and optionally why you did it in a commit message and do not write stuff like this. When we have to go over the history later to find a particular issue this does not help - then you can also just commit empty messages with a single commit.

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
marthtz: @elfenpiff yep, totally agree with you. It would have been too easy if this could be simplified but again it's a case by case decision!

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
marthtz: That's fine. We weren't sure about it so it was brought here.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: That's basically also a very general question. Should the description mention possible exit points?
In this example it sets up a new channel. Looks like `loadConfiguration` might not be a suitable name for this method.

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_writer.hpp
manel94: @elfenpiff  for explanation should we use @note, something like this should be enough :
 
         /// @note delete copy and move ctor and assignment since they are not needed

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
manel94: @elfenpiff, @marthtz, if we use only @brief, no doxygen warnings will be generated for return side?

elfenpiff:
@manel94 Too important things!

1. You are sharing Bosch internals here: Now everyone knows your internal user name (since you posted your home directory path). Should I delete your comment and you repost your findings without your home directory?!

2. This looks like your internal tooling is not comprehending C++/doxygen fully, maybe an update in the tool or your configuration will solve this issue on your side.

3. @marthtz Is your doxygen checker somehow open so that we can integrate it into our CI for the whole community?! Then we could also tackle the tool together and fix doxygen comments more efficiently.

marthtz:
>
> @marthtz Is your doxygen checker somehow open so that we can integrate it into our CI for the whole community?! 

I have a look.

>
> Then we could also tackle the tool together and fix doxygen comments more efficiently.

That's the way to go! Sounds good to me!

marthtz:
@elfenpiff There's no 3rd party doxygen checker. These warning are generated by the Doxygen tool but reported on stderr not stdcout! Doxygen version is 1.8.17, which is the default on Ubuntu 20.04.

@dkroenke @elfenpiff  We can decide if we want to live with these warnings for the sake of cleaner source code.

dkroenke:
> @elfenpiff There's no 3rd party doxygen checker. These warning are generated by the Doxygen tool but reported on stderr not stdcout! Doxygen version is 1.8.17, which is the default on Ubuntu 20.04.
> 
> @dkroenke @elfenpiff We can decide if we want to live with these warnings for the sake of cleaner source code.

@marthtz @elfenpiff You can see these warnings even in our CI when the documentation is generated, example: https://github.com/eclipse-iceoryx/iceoryx/pull/643/checks?check_run_id=2227351032#step:8:842
I agree to ignore warnings for documenting obvious code here because it does not add any value,

marthtz:
>  
> I agree to ignore warnings for documenting obvious code here because it does not add any value,

Awesome! I guess that was the last open and kind of blocking question. I'll go through everything again to be sure and then we can get this one finally closed!

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
marthtz: Only `@brief` fro getter/setter!

iceoryx_dds/include/iceoryx_dds/dds/data_reader.hpp
marthtz: Removed.

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
marthtz: There's no error evaluation of `setupChannel()` here.

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
marthtz: There's no check for a valid config here. It's currently used in a successful `expected`.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: There's no check for a valid config here. It's currently used in a successful expected. There's no error evaluation of `setupChannel()` here. Mentioned in comment.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
```diff
@@ -38,13 +37,30 @@ class DDS2IceoryxGateway : public gateway_t
 
   public:
     DDS2IceoryxGateway() noexcept;
+
+    /// @brief Set DDS to iceoryx gateway object
+    /// @param[in] channelFactory the channel factory to propagate data across
     DDS2IceoryxGateway(ChannelFactory channelFactory) noexcept;
+
+    /// @brief Load the provided gateway configuration; a channel is setup if required (no error check on channel setup)
```
elBoberido, line 45: please move the `(no error check on channel setup)` to a `@note no error ...`

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
```diff
@@ -38,13 +37,30 @@ class DDS2IceoryxGateway : public gateway_t
...
+    /// @param[in] channel the channel to propagate data across.
     void forward(const channel_t& channel) noexcept;
 
   private:
     void* m_reservedChunk = nullptr;
+
+    /// @brief Setup the channel for the given service
+    /// @param[in] service service description of the service to create a channel for
+    /// @param[in] publisherOptions the publisher options with historyCapacity and queueCapacity
+    /// @return an expected containing a copy of the added channel, otherwise an error
```
elBoberido, line 63: usually we don't document private methods with doxygen. It clutters the public API documentation

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
```diff
@@ -26,22 +27,33 @@ namespace iox
...
 /// @brief DDS Gateway implementation for the iceoryx to DDS direction.
-///
 template <typename channel_t = gw::Channel<popo::UntypedSubscriber, dds::data_writer_t>,
           typename gateway_t = gw::GatewayGeneric<channel_t>>
 class Iceoryx2DDSGateway : public gateway_t
 {
   public:
     Iceoryx2DDSGateway() noexcept;
+
+    /// @brief Load the provided gateway configuration; a channel is setup if required (no error check on channel setup)
```
elBoberido, line 38: please also move the `(no error ...` to a `@note` tag

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
```diff
@@ -26,22 +27,33 @@ namespace iox
...
+
+    /// @brief Forward data between the two terminals of the channel used by the implementation.
+    /// @param[in] channel The channel to propagate data across.
     void forward(const channel_t& channel) noexcept;
 
   private:
+    /// @brief Setup the channel for the given service
+    /// @param[in] service service description of the service to create a channel for
+    /// @param[in] subscriberOptions the Subscriber Options options with historyCapacity and queueCapacity
+    /// @return an expected containing a copy of the added channel, otherwise an error
```
elBoberido, line 54: this is also private and should not be documented with doxygen

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -27,34 +27,60 @@ namespace iox
...
+    /// initialized
     iox::cxx::optional<uint32_t> peekNextSize() override;
+
+    /// @brief Check if new samples are ready to take
+    /// @return true if there are samples available, otherwise false
     bool hasSamples() override;
+
+    /// @brief Take the next available sample from the DDS data space.
+    /// @param[in] buffer Receive buffer in which sample will be stored.
+    /// @param[in] bufferSize Size of the provided buffer.
```
elBoberido, line 65: please don't repeat yourself and have it just `@param[in] buffer to store the received sample` and `@param[in] bufferSize of provided buffer`

Please also adjust it at on the other methods which repeat the variable name

elBoberido (commented):
please rebase/merge master since there are changes which are conflicting with the PR, e.g. the `peekNextSize` method was replaced by `peekNexIoxChunkDatragramHeader`

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_writer.hpp
```diff
@@ -26,27 +26,49 @@ namespace iox
...
+
     virtual ~CycloneDataWriter();
+
     CycloneDataWriter(const CycloneDataWriter&) = delete;
     CycloneDataWriter& operator=(const CycloneDataWriter&) = delete;
-    // Required for vector
-    CycloneDataWriter(CycloneDataWriter&&) = default;
-    CycloneDataWriter& operator=(CycloneDataWriter&&) = default;
 
+    /// @brief Default move constructor to set the cyclone data writer
```
elfenpiff, line 48: This is just weird. Why documenting compiler-generated methods? 
The compiler generates a lot for us see this example:

```cpp
struct Fuu {
   int a;
};
```

In this case the compiler generates  methods so that the struct `Fuu` looks like:
```cpp
struct Fuu {
   Fuu();
   Fuu(Fuu&&);
   Fuu(const Fuu&);
   Fuu& operator=(Fuu&&);
   Fuu& operator=(const Fuu&);
   ~Fuu();

   int a;
};
```

And when we implement `Fuu` we try to follow the rule of zero, and if not possible, the rule of five. But in the rule of zero case we would have 6 methods which wouldn't be documented - and it makes sense!

So please remove the documentation from all compiler generated methods. Those methods are documented by the C++ standard which describes how they behave, what they do and shouldn't do.

iceoryx_dds/include/iceoryx_dds/dds/data_reader.hpp
```diff
@@ -38,63 +38,43 @@ enum class DataReaderError : uint8_t
...
-    ///
-    /// @brief peekNextSize Get the size of the next sample if one is available.
+    /// @brief Get the size of the next sample if one is available.
     /// @return The size of the next sample if one is available.
-    ///
     virtual iox::cxx::optional<uint32_t> peekNextSize() = 0;
 
-    ///
-    /// @brief hasSamples Checks if new samples ready to take.
+    /// @brief Checks if new samples ready to take.
```
elfenpiff, line 52: Typo: ... new samples are ready to ...

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
```diff
@@ -38,13 +37,30 @@ class DDS2IceoryxGateway : public gateway_t
...
+
+    /// @brief Set DDS to iceoryx gateway object
+    /// @param[in] channelFactory the channel factory to propagate data across
     DDS2IceoryxGateway(ChannelFactory channelFactory) noexcept;
+
+    /// @brief Load the provided gateway configuration; a channel is setup if required (no error check on channel setup)
+    /// @param[in] config the gateway configuration to load
     void loadConfiguration(const config::GatewayConfig& config) noexcept;
+
+    /// @brief Discover messages coming from DDS.
```
elfenpiff, line 49: Why is the  `CaproMessages` origin in DDS?
What discovery messages? I am somehow lost and do not understand the doxygen comment at all.

I think `CaproMessages` are used only internally by RouDi.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
```diff
@@ -38,13 +37,30 @@ class DDS2IceoryxGateway : public gateway_t
...
+    /// @param[in] msg the discovery message.
     void discover(const capro::CaproMessage& msg) noexcept;
+
+    /// @brief forward data between the two terminals of the channel used by the implementation.
```
elfenpiff, line 53: What are terminals? Where is it defined?

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
```diff
@@ -38,13 +37,30 @@ class DDS2IceoryxGateway : public gateway_t
 
   public:
     DDS2IceoryxGateway() noexcept;
```
elfenpiff, line 39: Missing documentation?

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
```diff
@@ -26,22 +27,33 @@ namespace iox
...
+
+    /// @brief Load the provided gateway configuration; a channel is setup if required (no error check on channel setup)
+    /// @param[in] config The gateway configuration to load
     void loadConfiguration(const config::GatewayConfig& config) noexcept;
+
+    /// @brief Discover messages coming from iceoryx.
+    /// @param[in] msg The discovery message.
     void discover(const capro::CaproMessage& msg) noexcept;
+
+    /// @brief Forward data between the two terminals of the channel used by the implementation.
```
elfenpiff, line 46: Again, what is a terminal and where is it defined?

elfenpiff (changes requested):
Please squash the commits!
Here is the documentation of 6 files touched and we already have 19 commits, three of them are "update copyright"

To reach the goal of traceability we need meaningful commit messages of a reasonable size. It does not make sense to clutter our history which makes it really hard for us in the future.

As a rule of thumb, when I implement something I split the work like this:
- implement 
- document (1 commit) 
- review which leads to another commit or two (when we only consider the documentation, except when there are API changes)

I think 3 - 6 commits are reasonable but nearly 20? When you take a look at the history it seems that the changes were never tested locally otherwise I cannot explain why there are two "fix warning" commits in here.

And please refrain from writing commit messages like "fix review finding". We would like to be traceable this means the commit message has to state what was done in a traceable manner - and here we are just wondering: What review finding? From who, what was there mentioned?

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: As discussed earlier it's now documented in the base class.

@ithier Can you answer the `CaproMessages` as you're the author?!

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: As discussed earlier it's now documented in the base class.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: done

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
marthtz: As discussed earlier it's now documented in the base class

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
```diff
@@ -27,34 +27,60 @@ namespace iox
...
     CycloneDataReader(const CycloneDataReader&) = delete;
     CycloneDataReader& operator=(const CycloneDataReader&) = delete;
     CycloneDataReader(CycloneDataReader&&) = delete;
     CycloneDataReader& operator=(CycloneDataReader&&) = delete;
 
+    /// @brief Connect cylcone data reader to the underlying DDS network
     void connect() noexcept override;
 
+    /// @brief Get the size of the next sample if one is available
+    /// @return an optional pointer to the size of the next sample if one is available, cxx::nullopt_t if value is not
```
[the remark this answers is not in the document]
marthtz: Move to base class and fixed.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: Move to base class and fixed.

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
marthtz: done

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
marthtz: Move to base class and fixed.

iceoryx_dds/include/iceoryx_dds/gateway/iox_to_dds.hpp
marthtz: done

iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_reader.hpp
marthtz: Move to base class and fixed.

elBoberido (commented):
looks good to me and can be merged once the question regarding the `CaproMessage` is answered

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
elBoberido: I'm not quite sure, but I think it is needed for service discovery and to suppress some services like introspection. At least in the `iox_to_dds.inl` it's used that way.

marthtz:
> 
> 
> looks good to me and can be merged once the question regarding the `CaproMessage` is answered

It took quite a while but I guess we clarified a couple of general questions here!

iceoryx_dds/include/iceoryx_dds/gateway/dds_to_iox.hpp
ithier: @marthtz @elBoberido Sorry I literally just saw this. I don't understand the question though. Where is the `CaproMessage` ?

elBoberido: @ithier it's below the line of the reviewed comment -> `void discover(const capro::CaproMessage& msg) noexcept;`

marthtz: @elBoberido @ithier @elfenpiff  Ok, So, is it fine for now or do you want me to change anything to finish this PR?

elBoberido (approved):
LGTM

dkroenke (approved):

PR #646, opened 2021-03-27
Iox #482 rework overview

### LINKED ISSUE #482: Extend documentation for 1.0 release and website
# Documentation epic

# Proposed outline
- index.md
- What is Eclipse iceoryx? (Explain the three pillars easy, safe, fast)
- getting-started
  - overview.md (Split?)
    - General intro (add overview pictures e.g. from [ROSCon slides](https://roscon.ros.org/2019/talks/roscon2019_truezerocopy.pdf), add section "limitations")
    - hello world/quick start: icedelivery c++ typed (without `and_then`?) `icehello`
    - User API (Move into examples)
      - Typed
      - Untyped  
  - installation.md (move colcon to advanced?, usual user just wants to use script and nothing more, link contributor installation guide, move iceoryx_posh options table to advanced, fix markdown no identation before code blocks)
  - examples (overview over examples)
    - Communication with polling for the user /for frameworks
    - Options
    - Listner 
    - Communication without polling
    - Debugging (iceperf, icecrystal)
- API-reference
- (API-cheatsheet.md)
- FAQ.md
- advanced
  - Contriubtor guides 
    - best-practice-for-testing.md
    - Modern C++ error handling (Optional, expected)
    - installation-guide-for-contributors.md 
  - iceoryx_utils.md
  - usage-guide.md

## Documents covering
- [x] Extend installation guide
  - [x] QNX7 with QCC 5.4.0
- [x] Introduction: motivation, goals and capabilities of iceoryx
  - [x] Which data types are allowed to be sent via iceoryx? What are the limitations (virtual, vtable, heap)?
  - [x] Clarify failure cases, e.g.
    - What happens if an application segfaults? If crashes in critical section memleak
    - Does the crash of one application affect the whole system
    - When an application crashes, what happens with its resources like memory chunks, UDS, etc.
  - [x] Update Waitset and Listener section in overview.md and have a use case based comparison 
- [x]  Add new examples
  - [x] README.md in the example 
  - [x] iceoptions 
  - [x] icehello
  - [x] iceensemble: Explain how C and C++ API work together (multiple publishers and subscribers)
  - [x] blocking publisher
  - [x] icecubetray: Add access control documentation (comes from #640)
  - [x] Record asciinema for all examples
- [x] Add example for static RouDi Config with custom mempools and one for the dynamic TOML config, e.b. by just a 
- [x] `@todo` in the code shall not be linked to issues for v1.0
- [x] Add Apex.OS to the framework section

### v2.0 Release (Moved to #743)

- [x] Improve utils documentation (in the long run: find more suitable name for utils, e.g. safe building blocks), see also #468
- [x] Describe all RouDi command line options?
- [x] Add info about release names (ice cream flavours)
- [x] add overview graphic introducing the main players (RouDi, Runtime/user apps, shared Memory, data packages/topics) in overview.md (see suggestions [here](https://github.com/eclipse-iceoryx/iceoryx/pull/646#discussion_r604950011) and [here](https://github.com/eclipse-iceoryx/iceoryx/pull/646#discussion_r604960591))
- [x] Add chunk header example
- [x] Add example sending complex data types
- [x] icedelivery:  Explain lifetime semantics of chunks/samples and user restrictions: loan needs to be followed with publish or release, MAX_CHUNKS_IN_USE on publisher and subscriber side etc., API is not thread-safe

Consolidate current documentation as far as possible (keep the scope of the issue confined to documentation).
Note: doxygen documentation rework is out of scope here and belongs to #468

This is later to be integrated into the website #136

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
In this PR I reworked the overview.md and fixed some links.

## References

- Partially closes #482


doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+result = iox::cxx::nullopt;
+```
+
+For a complete list of available functions see 
+[``optional.hpp``](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/optional.hpp).
+
+### Expected
+``iox::cxx::expected<T, E>`` generalizes ``iox::cxx::optional`` by admitting a value of another type ``E`` instead of
+no value at all, i.e. it contains either a value of type ``T`` or ``E``. In this way, ``expected`` is a special case of 
+the either monad. It is usually used to pass a value of type ``T`` or an error that may have occurred, i.e. ``E`` is the
```
dkroenke, line 66: i would set "either monad" in quotes and maybe a link can be added, like the other monads.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+Many parts of the iceoryx C++ API follow a functional programming approach and allow the user to specify functions 
+which handle the possible cases, e.g. what should happen when data is received.
+
+This is very flexible but requires using the monadic types ``cxx::expected`` and ``cxx::optional``, which we
+introduce in the following sections.
+
+### Optional
+
+The type ``iox::cxx::optional<T>`` is used to indicate that there may or may not be a value of a specific type ``T``
+available. This is essentially the maybe [monad](https://en.wikipedia.org/wiki/Monad_(functional_programming)) in
```
dkroenke, line 12: The maybe [monad](https://en.wikipedia.org/wiki/Monad_(functional_programming)) could be set in quotes... maybe

CONTRIBUTING.md
```diff
@@ -81,7 +81,8 @@ codebase follows these rules, things are work in progress.
 7) **Always use `iox::log::Logger`**, instead of `printf()`
 8) **Always use `iox::ErrorHandler()`**, instead of the direct STL calls
 
-See [error-handling.md](./doc/error-handling.md) for additional information about logging and error handling. 
+See [error-handling.md](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/design/error-handling.md) for additional 
+information about logging and error handling.
```
dkroenke, line 85: trailing withespace

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+```
+Notice that we get the value by reference, so if a copy is desired it has to be created explicitly in the 
+[lambda](https://en.wikipedia.org/wiki/Anonymous_function#C++_(since_C++11)) or function we pass.
+
+The optional can be initialized from a value directly
+```cpp
+optional<int> result = 73;
+result = 37;
+```
+If it is default initialized it is automatically set to its null value of type ``iox::cxx::nullopt_t``. This can be also
```
dkroenke, line 53: If the optional is default initialized, it is automatically set

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+    // do something with the value
+}
+else
+{
+    auto error = result.get_error();
+    // handle the error
+}
+```
+
+Should we need an error value we set
```
dkroenke, line 89: If we need an error value, we set

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+shared memory, which the applications use for exchanging payload data. Sometimes referred to as daemon, RouDi manages
+the shared memory and is responsible for the service discovery, i.e. enabling subscribers to find topics offered by 
+publishers. It also keeps track of all applications which have initialized a runtime and are hence able to use
+publishers or subscribers. To view the available command line options call `$ICEORYX_ROOT/build/iox-roudi --help`.
 
-##### 1. Loan and emplace
+## Runtime
 
-Before sending, we have to loan a chunk of memory to emplace our data.
+Each application which wants to use iceoryx has to instantiate its runtime, which essentially enables communication
```
dkroenke, line 121: Each application that wants

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-iox::popo::UntypedSubscriber subscriber({"Group", "Instance", "CounterTopic"});
-subscriber.subscribe();
-```
+This means the group and instance identifier can be ignored to create different ``ServiceDescription``s. They will be
+needed for advanced filtering functionality in the future.
 
-#### Receiving data
+The data type of the transmitted data can be any C++ class, struct or plain old data type as long as it satisfies the 
+following conditions:
+- no heap is used
```
dkroenke, line 160: maybe some prominent examples can be mentioned: `std::string` or `std::vector`.
On the other hand we should mention what is possible to use, like arrays filled with fixed-size data.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-Once we are done sending, we call ``stopOffer`` at the publisher.
+The easiest way to receive data is to periodically poll whether data is available. This is sufficient for simple use
+cases but inefficient in general, as it often leads to unnecessary latency and wake-ups without data.
 
-```cpp
-publisher.stopOffer();
-```
+The ``Waitset`` can be used to relinquish control (putting the thread to sleep) and wait for user defined ``events``
+to occur. Here an event is associated with a condition and occurs when this condition becomes true.
```
dkroenke, line 195: and wait that user defined ``events`` arise.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-```
-# If installed and available in PATH environment variable
-iox-roudi
-# If build from scratch with script in tools
-$ICEORYX_ROOT/build/posh/iox-roudi
-```
+We distinguish between the ``typed API`` and the ``untyped API``. In the typed API the underlying data type is made
+apparent by typed pointers or references to some data type T (often a template parameter). This allows working with
+the data in a C++ idiomatic and type-safe way and should be preferred whenever possible. This API is mainly used when
```
dkroenke, line 222: The typed API is mainly used

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+```cpp
+result = iox::cxx::error<E>(errorCode);
+```
+which assumes that ``E`` can be constructed from an ``errorCode``.
+
+We again can employ a functional approach like this
```
dkroenke, line 95: colon missing:

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+This means the group and instance identifier can be ignored to create different ``ServiceDescription``s. They will be
+needed for advanced filtering functionality in the future.
 
-#### Receiving data
+The data type of the transmitted data can be any C++ class, struct or plain old data type as long as it satisfies the 
+following conditions:
+- no heap is used
+- the data structure is entirely contained in shared memory - no pointers to process local memory, no references to 
+process local constructs, no dynamic allocators
+- the data structure must not internally use pointers/references
```
elfenpiff, line 163: Maybe add the hint that the data structure should be relocatable and therefore the data structure must not internally use pointers/references

elBoberido: I think we can add that the data must be trivially copyable. Maybe we should also check for this in the typed publisher/subscriber?

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-
 ```
+By calling ``take`` we get an ``expected`` and hence we may have to handle an error.
 
-By calling ``take`` we get a ``iox::cxx::expected<iox::optional<iox::popo::Sample<const CounterTopic>>>``. Since this
-may fail, we have to handle a potential error. If there is no error, we may still have not recived a sample, indicated by
-a ``nullopt``. Otherwise we can get the pointer of our data from the sample and process the received data.
+And that's it. We have created our first simple iceoryx example. 
+[Here](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/README.md) you can find further examples 
+which demonstrate how iceoryx can be used and describe our API in more detail.
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: nope, ``examples`` is plural

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-```cpp
-subscriber.unsubscribe();
-```
+For more information on how to use the Waitset see 
+[Waitset](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/waitset/README.md).
 
-Both will also be called in the respective destructor if needed (i.e. if the publisher is still offering or the
-subscriber is still subscribed).
+## Listener
+part of #350
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Will be added as part of #350

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: But then we would not be able to send for instance our cxx::vector or cxx::string and so on.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
FerdinandSpitzschnueffler: added quotation marks but not the link since it would be the same

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-A ``reinterpret_cast`` is used to interpret the data as a ``CounterTopic``. Note that a ``static_cast`` would also work here,
-but a ``reinterpret_cast`` is used to emphasize the data-agonstic nature of the data transmission itself. 
+## Subscriber
 
-As in the untyped case we also could use a loop to get all samples as long as they are available.
+Symmetrically a subscriber also corresponds to a topic and thus needs a service description to be constructed. As for
+publishers we distinguish between typed and untyped subscribers.
 
-A non-functional approach is also possible but more verbose.
+Once a subscriber is subscribed to some topic, it is able to receive data of the type tied to this topic. In the
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I'd prefer to leave it as it is.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-```cpp
-publisher.stopOffer();
-```
+The ``Waitset`` can be used to relinquish control (putting the thread to sleep) and wait for user defined ``events``
+to occur. Here an event is associated with a condition and occurs when this condition becomes true.
+Usually these events correspond to the availability of data at specific subscribers. This way we
+can immediately wake up when data is available and will avoid unnecessary wake-ups if no data is available.
 
-Similarly the subscriber can ``unsubscribe`` to stop receiving any data.
+It manages a set of triggers which can be activated to indicate that a corresponding event occurred which wakes
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I'd prefer to leave it as it is.

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: I'd prefer to leave it as it is.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +11,220 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-[conceptual guide](../../conceptual-guide.md) provides additional information about the _Shared Memory
-communication_ that lies at the heart of iceoryx.
+For further information how iceoryx can be used see the 
+[examples](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/README.md). The
+[conceptual guide](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/conceptual-guide.md) provides additional 
+information about the _Shared Memory communication_ that lies at the heart of iceoryx.
 
-We now briefly define the main entities of an iceoryx system before showing how they are created and used by the
-iceoryx API.
+### icehello example
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: An overview graphic would be cool, but I'm not sure if I can provide one until Thursday. I suggest to add the graphic within a separate PR so that we can merge this one until Thursday.

dkroenke (dismissed):

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
mossmaurice: Please add a todo with `<!-- ... -->` for the graphic

doc/design/fixedString.adoc
```diff
@@ -31,157 +31,3 @@ In the second step, the functionality has been extended:
 .step 3:
```
elfenpiff, line 31: Could we remove this file completely? A design document for the fixed string does not make sense in my opinion.
If we would require this that we would clutter the doc/design/ folder with a design document for every cxx container and construct.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+    // handle the case that there is no value
+}
+```
+A shorthand to get the value is
+```cpp
+auto value = *result;
+```
+
+!!! attention
+    Accessing the value if there is no value is undefined behavior, so it must be checked beforehand.
```
elfenpiff, line 37: No it is no undefined behavior. We call `cxx::Expects` beforehand and if it has no value the application will terminate.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+result = 37;
+```
+If the optional is default initialized, it is automatically set to its null value of type ``iox::cxx::nullopt_t``. 
+This can be also done directly by using the constant ``iox::cxx::nullopt``
+
+```cpp
+result = iox::cxx::nullopt;
+```
+
+For a complete list of available functions see
```
elfenpiff, line 60: Maybe we could also hint to a STL reference page since our optional behaves like the stl optional. The differences are, we do not throw exceptions and we do not have undefined behavior.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+
+```cpp
+result = iox::cxx::nullopt;
+```
+
+For a complete list of available functions see 
+[``optional.hpp``](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/optional.hpp).
+
+### Expected
+``iox::cxx::expected<T, E>`` generalizes ``iox::cxx::optional`` by admitting a value of another type ``E`` instead of
```
elfenpiff, line 64: The error type E has some requirements. It must contain a static member or enum value called `INVALID_STATE`. @mossmaurice you implemented this could you please provide here a hint what requirements `E` has?

You could also look at the expected unit tests since it is tested there.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,104 @@
...
+the 'either monad'. It is usually used to pass a value of type ``T`` or an error that may have occurred, i.e. ``E`` is the
+error type. For more information on how it is used for error handling see
+[error-handling.md](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/design/error-handling.md).
+
+Assume we have ``E`` as an error type, then we can create a value
+```cpp
+iox::cxx::expected<int, E> result(iox::cxx::success<int>(73));
+```
+
+and use the value or handle a potential error
```
elfenpiff, line 75: We should mention here also somewhere, when you move an expected the origin is set to the error value `E::INVALID_STATE` and will return always true in `.has_error()`.

Example:
```cpp
cxx::expected<int, E> result(iox::cxx::success<int>(1338));
cxx::expected<int, E> anotherResult = std::move(result);

if ( result.has_error() ) // is now true since it was moved
   result.get_error(); // returns E::INVALID_STATE
```

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-    sample->counter = 73;
+    sample->counter = 30;
     sample.publish();
 }
 else
 {
     // handle the error
 }
 ```
+Here ``result`` is an ``expected`` and hence we may get an error which we have to handle. This can happen if we try
```
elfenpiff, line 51: Please add the hint that you have to handle the error somehow since the expected class has the nodiscard keyword attached. Otherwise you will get a warning or and error when you build in strict mode.

If you do not want to handle the error you can use `IOX_DISCARD_RESULT(myMethodThatReturnsAnExpected())`

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+        //process the data
+    }
+    else
+    {
         //handle the error
     }
 }
-
 ```
+By calling ``take`` we get an ``expected`` and hence we may have to handle an error.
```
elfenpiff, line 83: We have to handle the error or explicitly discard it with `IOX_DISCARD_RESULT` (discarding is discouraged but can be helpful for prototypes).

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-```cpp
-iox::popo::UntypedPublisher publisher({"Group", "Instance", "CounterTopic"});
-publisher.offer();
-```
+## RouDi
 
-#### Sending data
+RouDi is an abbreviation for **Rou**ting and **Di**scovery. RouDi takes care of the
+communication setup but does not actually participate in the communication between the publisher and the subscriber.
+RouDi can be thought of as the switchboard operator of iceoryx. One of his other major tasks is the setup of the
```
elfenpiff, line 112: I though RouDi can be thought of as a dating site?! ;)

elfenpiff, line 112: One of its other major ...

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+To do so, the following lines of code are required
 ```cpp
-auto result = publisher.loan(sizeof(CounterTopic));
+iox::runtime::PoshRuntime::initRuntime("some_unique_application_name");
 ```
+## Creating service descriptions for topics
 
-Since the data type is not known to the publisher, we have to provide the size in bytes of the payload data we
-intend to send.
+A ``ServiceDescription`` in iceoryx represents the data to be transmitted and is uniquely identified by three string
```
elfenpiff, line 129: it represents a topic or channel under which publisher/subscribers can exchange data

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-publisher.loan(sizeof(CounterTopic))
-    .and_then([&](iox::popo::Sample<void>& sample) {
-        new (sample.get()) CounterTopic(73);
-        sample.publish();
-    })
-    .or_else([](iox::popo::AllocationError) {
-        /* handle the error */
-    });
-```
+Service and instance are like classes and objects in C++. So you always have a specific instance of a service during
```
elfenpiff, line 148: Service is related to instance like classes are related to objects in C++. Service describes an abstract topic and an instance is one instantiation of that abstraction. Like an object is an instantiated class.
Events are in this context like members of a class.

Example:

```cpp
class MyRadarService {
   public:
      bool hasObstacleDetected;
      float distanceToObstacle;
};

MyRadarService frontLeftRadarInstance;
std::cout << frontLeftRadarInstance.hasObstacleDetected << std::endl;
```

In the iceoryx world we would subscribe to the service ("MyRadarService", "frontLeftRadarInstance", "hasObstacleDetected") and would receive a sample whenever an obstacle were detected. Or we would subscribe to "distanceToObstacle" and would receive a constant stream of data which presents the distance to the obstacle.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-Receiving the data works in the same way as in the typed API, the main difference is the ``reinterpret_cast`` needed
-before accessing the data (since the subscriber has no knowledge of the underlying type).
+## Publisher
 
-```cpp
-while (keepRunning)
-{
-    // wait for new data (either sleep and wake up periodically or by notification from the waitset)
+A publisher is tied to a topic and needs a service description to be constructed. If it is typed one needs to
```
elfenpiff, line 173: If it is typed, one needs to

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+additionally specify the data type as a template parameter. Otherwise, the publisher is only aware of raw memory and 
+the user has to take care that it is interpreted correctly.
```
elfenpiff, line 175: has to ensure that it is

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-Once we are done sending, we call ``stopOffer`` at the publisher.
+The easiest way to receive data is to periodically poll whether data is available. This is sufficient for simple use
+cases but inefficient in general, as it often leads to unnecessary latency and wake-ups without data.
 
-```cpp
-publisher.stopOffer();
-```
+The ``Waitset`` can be used to relinquish control (putting the thread to sleep) and wait for user-defined ``events``
+to occur. Here an event is associated with a condition and occurs when this condition becomes true.
```
elfenpiff, line 200: Events are no longer associated with a condition those are states. I would just remove the sentence and keep it simple here and state that the user can wait for user-defined events and when they occur the user is informed.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,226 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+Usually, these events correspond to the availability of data at specific subscribers. This way we
+can immediately wake up when data is available and will avoid unnecessary wake-ups if no data is available.
 
-Similarly the subscriber can ``unsubscribe`` to stop receiving any data.
+It manages a set of triggers which can be activated to indicate that a corresponding event occurred which wakes
```
elfenpiff, line 204: Implementation detail! This is not relevant for the user. I would remove the trigger part and just state for the experts that the waitset is an implementation of the reactor pattern and is informed with a push strategy that one of the attached events occured at which it informs the user.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
mossmaurice:
```
    static_assert(internal::HasInvalidStateMember<T>::value,
                  "T must have an INVALID_STATE value/member. Alternatively write an ErrorTypeAdapter specialisation "
                  "for your type");
```

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
mossmaurice: Why are we doing this inside the `optional` and not inside the `expected`? This should be consistent!

doc/design/fixedString.adoc
elBoberido: we need maybe one where we describe our intention with the re-implementation of the STL container, what constraints our container have and how they deviate from the STL

FerdinandSpitzschnueffler: This already described [here](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#cxx).

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,122 @@
...
+
+The type ``iox::cxx::optional<T>`` is used to indicate that there may or may not be a value of a specific type ``T``
+available. This is essentially the 'maybe [monad](https://en.wikipedia.org/wiki/Monad_(functional_programming))' in 
+functional programming. Assuming we have some optional (usually the result of some computation)
+
+```cpp
+optional<int> result = someComputation();
+```
+
+we can check for its value using
```
mossmaurice, line 19:
```suggestion
We can check for its value using
```

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-For further information how iceoryx can be used see the [examples](../../../iceoryx_examples/README.md). The
-[conceptual guide](../../conceptual-guide.md) provides additional information about the _Shared Memory
-communication_ that lies at the heart of iceoryx.
+<!-- @todo add overview graphic -->
 
-We now briefly define the main entities of an iceoryx system before showing how they are created and used by the
-iceoryx API.
+For further information how iceoryx can be used, see the 
+[examples](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/README.md). The
```
mossmaurice, line 16:
```suggestion
[examples](../examples/). The
```
I wouldn't jump to GitHub here, this could be confusing for readers. There will be a link to the code at the end of each example.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 }
 else
 {
     // handle the error
 }
 ```
+Here ``result`` is an ``expected`` and hence we may get an error. This can happen if we try to loan too many samples 
+and exhaust memory. We have to handle this potential error since the expected class has the ``nodiscard`` keyword 
+attached. This means we get a warning (or an error when build in strict mode) when we don't handle it. We could also 
+explicitly discrad it with ``IOX_DISCARD_RESULT`` which is discouraged. If you want to know more about ``expected``,
```
mossmaurice, line 56: I'd move the sentence about `nodiscard` and `IOX_DISCARD_RESULT` to the "how-optional-and...md". IMHO this is not relevant at this point to go into details.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-
-We can shorten this somewhat by using a functional approach.
-
-##### 2. Functional approach
-
-```cpp
-while (keepRunning)
-{
-    // wait for new data (either sleep and wake up periodically or by notification from the waitset)
+And that's it. We have created our first simple iceoryx example.
```
mossmaurice, line 90: Add a link to the `icehello` example here? Will be available with #655

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+!!! note
+    Most of the STL types cannot be used, but we reimplemented some of them so that they meet the conditions above. 
+    You can find an overview [here](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#cxx).
 
-#### Receiving data
+## Publisher
 
-Receiving the data works in the same way as in the typed API, the main difference is the ``reinterpret_cast`` needed
-before accessing the data (since the subscriber has no knowledge of the underlying type).
+A publisher is tied to a topic and needs a service description to be constructed. If it is typed, one needs to
```
mossmaurice, line 192: `icehello` from #655 only uses the typed API. I would leave out the untyped or C API at this point and just answer the questions "What is a publisher?" "How is it related to the other building blocks in iceoryx that are depicted on the graphic?"

The difference between untyped, typed is introduced in the icedelivery example.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-neither publisher nor subscriber have any knowledge about the underlying type they send or receive. This means that
-the user is responsible to ensure the data is read correctly, i.e. there is no type safety guaranteed by the API
-itself.
+RouDi is an abbreviation for **Rou**ting and **Di**scovery. RouDi takes care of the
+communication setup but does not actually participate in the communication between the publisher and the subscriber.
+RouDi can be thought of as the switchboard operator of iceoryx. One of its other major tasks is the setup of the
+shared memory, which the applications use for exchanging payload data. Sometimes referred to as daemon, RouDi manages
+the shared memory and is responsible for the service discovery, i.e. enabling subscribers to find topics offered by 
+publishers. It also keeps track of all applications which have initialized a runtime and are hence able to use
+publishers or subscribers. To view the available command line options call `$ICEORYX_ROOT/build/iox-roudi --help`.
```
mossmaurice, line 122: I think here we should add a section "Shared Memory". I suppose this will be a red box where the data packages aka topics will live inside, like on the current [GIF](https://user-images.githubusercontent.com/8661268/74612998-b962bc80-510a-11ea-97f0-62f41c5d287b.gif).

The section restrictions could be a sub-section of "Shared Memory"

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-When creating an untyped publisher we do not need to specify a data type as template paraemter.
+Each application that wants to use iceoryx has to instantiate its runtime, which essentially enables communication
+with RouDi. Only one runtime object per user process is allowed.
 
+To do so, the following lines of code are required
 ```cpp
-iox::popo::UntypedPublisher publisher({"Group", "Instance", "CounterTopic"});
-publisher.offer();
+iox::runtime::PoshRuntime::initRuntime("some_unique_application_name");
```
mossmaurice, line 130: What does the runtime do? Why do we need to create it? Please mention here, that the runtime abstracts the communication to RouDi, it also maps the shared memory to the apps address space, etc.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,122 @@
...
+[``optional.hpp``](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/optional.hpp).
+The ``iox::cxx::optional`` behaves like the [``std::optional``](https://en.cppreference.com/w/cpp/utility/optional)
+except that it does not throw exceptions and has no undefined behavior.
+
+### Expected
+``iox::cxx::expected<T, E>`` generalizes ``iox::cxx::optional`` by admitting a value of another type ``E`` instead of
+no value at all, i.e. it contains either a value of type ``T`` or ``E``. In this way, ``expected`` is a special case of 
+the 'either monad'. It is usually used to pass a value of type ``T`` or an error that may have occurred, i.e. ``E`` is the
+error type. ``E`` must contain a static member or an enum value called ``INVALID_STATE``. Alternatively an 
+``ErrorTypeAdapter`` can be implemented.
```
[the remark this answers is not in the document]
elfenpiff: We have an `ErrorTypeAdapter` example?

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+- the data structure has to be relocatable and therefore must not internally use pointers/references
+- no virtual members
 
-```cpp
-iox::popo::UntypedSubscriber subscriber({"Group", "Instance", "CounterTopic"});
-subscriber.subscribe();
-```
+!!! note
+    Most of the STL types cannot be used, but we reimplemented some of them so that they meet the conditions above. 
+    You can find an overview [here](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#cxx).
```
mossmaurice, line 189: You should mention here, that RouDi and all Apps need to be built with the same compiler and the same compiler flags.

doc/website/getting-started/overview.md
elfenpiff: I would jump to GitHub here since most of the developers will use the GitHub documentation.
I have the feeling the thing with the website is somehow getting out of hand. Instead of optimizing markdown documents for a website, we should go the clean way and add markdown comments and fill them out with a bash script so that, when the website is generated everything looks clean.

This is getting weirder and uncleaner ...

Quality over quantity! The website is also not mandatory till Thursday but a clean docu is!

doc/website/getting-started/overview.md
elfenpiff: I disagree since it is mandatory to handle the error of an expected.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-            iox::popo::ChunkReceiveResult& error = result.get_error();
-            //handle the error
-        }
-    }
-```
+When multiple publishers have offered the same topic the subscriber will receive the data of all of them (but in
+indeterminate order between different publishers).
 
-### Shutdown
+## Waitset
```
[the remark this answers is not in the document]
elfenpiff: The overview should provide an overview. The WaitSet and Listener are constructs that should be contained in that overview otherwise it is not an overview.

What kind of ASCII art would you like to see here? A class diagram? I am a little bit puzzled.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-```cpp
-subscriber.unsubscribe();
-```
+For more information on how to use the Waitset see 
+[Waitset](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/waitset/README.md).
 
-Both will also be called in the respective destructor if needed (i.e. if the publisher is still offering or the
-subscriber is still subscribed).
+## Listener
```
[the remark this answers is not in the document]
elfenpiff: Why do we need a diagram for the relation here?
I would put this in the design document and not here.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
FerdinandSpitzschnueffler: This is not a new sentence.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-Now that we have applications capable of sending and receiving data, we can run the complete iceoryx system.
+The API is offered in two languages, C++ and C. Detailed information can be found in the 
+[C++ example](http://iceoryx.io/getting-started/examples/icedelivery.md) and 
+[C example](http://iceoryx.io/getting-started/examples/icedelivery_in_c.md).
 
-First we need to start RouDi.
+Many parts of the C++ API follow a functional programming approach which is less error-prone. This requires using 
+the monadic types ``cxx::expected`` and ``cxx::optional`` which are introduced 
+[here](http://iceoryx.io/advanced/usage-guide/how-optional-and-error-values-are-returned-in-iceoryx/).
```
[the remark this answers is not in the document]
elfenpiff: I would leave it as it is. Repetition is good since the reader will most likely not read it from beginning to the end. They will pick a topic of interest and read only this part. And in this case some redundant links are pretty nice.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+RouDi is an abbreviation for **Rou**ting and **Di**scovery. RouDi takes care of the
+communication setup but does not actually participate in the communication between the publisher and the subscriber.
+RouDi can be thought of as the switchboard operator of iceoryx. One of its other major tasks is the setup of the
+shared memory, which the applications use for exchanging payload data. Sometimes referred to as daemon, RouDi manages
+the shared memory and is responsible for the service discovery, i.e. enabling subscribers to find topics offered by 
+publishers. It also keeps track of all applications which have initialized a runtime and are hence able to use
+publishers or subscribers. To view the available command line options call `$ICEORYX_ROOT/build/iox-roudi --help`.
 
-#### Creating a publisher
+## Runtime
```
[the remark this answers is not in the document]
elfenpiff: I would suggest to not create any ASCII arts or graphics at all. This can become more confusing then helpful and it is also time consuming.

Shouldn't we wait until someone professional creates some nice diagrams for us?!

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: Nope sinsce this is not the icehello example you introduced meanwhile. I'll change the name.

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
mossmaurice: Sure, but we should have a rough idea or a draft available IMHO. We could add a link to it and comment it out `<!-- -->`

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
mossmaurice: What I have in mind is something like the [GIF](https://user-images.githubusercontent.com/8661268/74612998-b962bc80-510a-11ea-97f0-62f41c5d287b.gif) but only static.

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
mossmaurice: IMHO the purpose of the `overview.md` is to introduce the main building blocks and players that are available in iceoryx on a abstract level, not going into the details too much. Imagine you want to [explain](https://www.youtube.com/watch?v=R6zeikbTgVc) someone to iceoryx not yet familiar with it. "In iceoryx you have three building blocks. RouDi, a daemon, the user apps and shared memory. What happens is the that the apps register with RouDi and then map the shared memory into their address space. This part is done by the runtime, the user doesn't have to care about it. The runtime is a object / colored-bloc-in-the-graphic inside the user apps. Now you want to exchange data. For this, there are special building blocks called publisher and subscriber. They can exchange data packages which are called topics. Depicted in the graphic as postal parcel. Every package needs a so called service description. This defines the content of the parcel and is depicted as the color of the parcel. Green parcels depict our `RadarObject` and so on..." I imagine such a description just much more detailed and probably separated into the different sections. Afterwards, making a connection with the code snippet and diving into the details.

doc/website/getting-started/overview.md
mossmaurice: Sure this is a very important point. But is this really something every user needs to learn at this point in time? We're at line 56 here! I doubt that. We need to focus on the very basics!  I imagine most users will check out the `how-to-use-optional-expected.md` first and IMHO that's the right location to describe it. And besides that, the compiler will poke them in their face if they do it wrong and that's the right time to learn it ;)

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @mossmaurice As mentioned before, any graphic will be part of a follow-up PR since it might not be finished until tomorrow. 
Regarding the Waitset & Listener I agree with @elfenpiff

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: In my opinion it belongs to an overview.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,242 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-Note that since the received sample received is untyped (``iox::popo::Sample<const void>``), we cannot use
-``operator->`` to access the members of the underlying type but have to cast it to the correct type ``CounterTopic``
-manually. We know this type since it is uniquely identified by the topic we subscribed to (in our case ``CounterTopic``). 
-A ``reinterpret_cast`` is used to interpret the data as a ``CounterTopic``. Note that a ``static_cast`` would also work here,
-but a ``reinterpret_cast`` is used to emphasize the data-agonstic nature of the data transmission itself. 
+## Subscriber
 
-As in the untyped case we also could use a loop to get all samples as long as they are available.
+Symmetrically a subscriber also corresponds to a topic and thus needs a service description to be constructed. As for
+publishers we distinguish between typed and untyped subscribers.
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: In my opinion it belongs to an overview.

doc/website/getting-started/overview.md
mossmaurice: Ok, should I then remove the explanation in the icedelivery example or is ok from your side if we explain it again there?

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: I wouldn't add this section here. I think many potential iceoryx users will know what Shared memory is and therefore a definition here is not needed. For those who are interested there is a link to our conceptual-guide.md at the beginning of this file.

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
mossmaurice: If you are ok with it, I'd not rush here and take some more time to clarify these questions. Quality over quantity IMHO. Besides that, the final documentation is not needed for tomorrow's v1.0-rc1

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @mossmaurice I added your suggestions to the issue.

doc/website/getting-started/overview.md
mossmaurice: I would try to avoid to do such assumptions. We need to get every reader to the same level without prior knowledge. Users that already know what shared memory is, will just skip the section and jump to the next one. I'm happy to discuss the structure, though. However, shared memory is one of the basic elements of iceoryx, if not the main element. In my head it will be depicted by e.g. a red box in the graphic and therefore needs proper introduction.

doc/website/getting-started/overview.md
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @mossmaurice I added your suggestion to the issue.

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: added to example

doc/website/getting-started/overview.md
FerdinandSpitzschnueffler: It should be explained in the example too

CONTRIBUTING.md
```diff
@@ -81,7 +81,8 @@ codebase follows these rules, things are work in progress.
 7) **Always use `iox::log::Logger`**, instead of `printf()`
 8) **Always use `iox::ErrorHandler()`**, instead of the direct STL calls
```
MatthiasKillat, line 82: What are those direct STL calls and how can the error handler help me there? This is poorly explained and sounds like the error handler would be some substitute which it is not.

We use the error handler once a error that cannot or shall not be propagated via expected has occurred, as was (still is?) described in error-handling.md. STL uses a totally different strategy, namely exceptions. Using error the handler is not related to STL but to any calls where errors can occur (POSIX calls with return codes etc.).

doc/goals-non-goals.md
```diff
@@ -46,4 +46,4 @@ that slow self-made IPC to communicate with his sensors because his parents want
 makes it hard for him to react in real-time to dangerous incidents like flying coffee cups. When strolling through
 the interwebs on a lonely evening, he finds out about iceoryx: Free-to-use, high-performance data transfer with low
```
MatthiasKillat, line 47: Quite the compelling story. Interwebs ... I would keep it professional and not kiddie style.

doc/goals-non-goals.md
```diff
@@ -46,4 +46,4 @@ that slow self-made IPC to communicate with his sensors because his parents want
...
 runtime overhead, real-time support! Brilliant! Maybe even Robby's biggest wish for a network binding will come true,
-so he can stream his favorite [video](https://www.youtube.com/watch?v=g5NkgZXWl0w) even faster!
+so he can stream his favorite [video](https://www.youtube.com/watch?v=BfrloBS2MCo) even faster!
```
MatthiasKillat, line 49: I do not dare to click this, probably would regret it anyway.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
MatthiasKillat: The difference is also that it has no efficient move semantics which the STL optional potentially has. Not an interface difference but an important coniseration for use with large data. I guess accessing a nullopt (and hence an object that was moved from) now leads to defined termination instead of undefined behavior, otherwise we cannot claim this. We certainly do not use exceptions though.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,122 @@
...
+This can be also done directly by using the constant ``iox::cxx::nullopt``
+
+```cpp
+result = iox::cxx::nullopt;
+```
+
+For a complete list of available functions see 
+[``optional.hpp``](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/optional.hpp).
+The ``iox::cxx::optional`` behaves like the [``std::optional``](https://en.cppreference.com/w/cpp/utility/optional)
+except that it does not throw exceptions and has no undefined behavior.
```
MatthiasKillat, line 63: It has undefined behavior, see above. At least in the same sense as undefined behavior of dereferencing a nullptr goes (which will crash on all platforms known to me but in theory does not have to ... hence undefined). This is no different here. We do not throw exceptions though.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
MatthiasKillat: By the way this is a bad requirement and unnecessarily restrictive. Nothing to do with the overview though except that it needs to be stated somewhere for potential users.

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
```diff
@@ -0,0 +1,122 @@
...
+auto handleError = [](E& value) { /*handle the error*/ };
+result.and_then(handleValue).or_else(handleError);
+```
+
+There are more convenience functions such as ``value_or`` which provides the value or an alternative specified by the
+user. These can be found in 
+[``expected.hpp``](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_utils/include/iceoryx_utils/cxx/expected.hpp).
+
+Note that when we move an ``expected``, the origin is set to the error value ``E::INVALID_STATE`` and will always return true 
+in ``has_error()``:
```
MatthiasKillat, line 112: "when invoking has_error" or "has_error will always return true"

MatthiasKillat (commented):
Did not look at overview.md yet. Will take a look later.

doc/goals-non-goals.md
FerdinandSpitzschnueffler: ok :D removed it

doc/website/advanced/how-optional-and-error-values-are-returned-in-iceoryx.md
FerdinandSpitzschnueffler: When accessing a nullopt, we terminate. There is no undefined behavior. But I'm not sure if I got your point. Did I misunderstand something?

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-    }
-    // wait for new samples (either sleep or use the waitset)
-}
+iox::runtime::PoshRuntime::initRuntime("some_unique_application_name");
 ```
 
-Here we do not check whether we actually have data since we already know there is data available by calling `hasData`.
+The runtime is an object inside the user application that maps the shared memory into the user 
+application's address space. When an application terminates, the runtime cleans up all resources needed for the 
+communication with RouDi. If an application crashes, it also takes care about the clean-up. Due to our
```
elBoberido, line 148: oh, you handle the abnormal termination here. But it's RouDi who does the clean-up.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-}
+iox::runtime::PoshRuntime::initRuntime("some_unique_application_name");
 ```
 
-Here we do not check whether we actually have data since we already know there is data available by calling `hasData`.
+The runtime is an object inside the user application that maps the shared memory into the user 
+application's address space. When an application terminates, the runtime cleans up all resources needed for the 
+communication with RouDi. If an application crashes, it also takes care about the clean-up. Due to our 
+[lock-free constructs](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#concurrent), 
+a crash of one application does not affect the whole iceoryx system.
```
elBoberido, line 150: this is currently not quite right. Especially with the blocking publisher feature it's quite easy to crash RouDi and bring the whole system down. That's also possible without the blocking publisher, although a bit more difficult. If an application crashes when the `ChunkDistributor` holds the `last lock`™ in iceoryx, RouDi crashes when it tries to clean up the application resources.

I'm not sure if we need to state this here in this detail, but maybe we need to link to a `limitations.md`

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-neither publisher nor subscriber have any knowledge about the underlying type they send or receive. This means that
-the user is responsible to ensure the data is read correctly, i.e. there is no type safety guaranteed by the API
-itself.
+A ``ServiceDescription`` in iceoryx represents a topic under which publisher and subscribers can exchange data and is 
+uniquely identified by three string identifiers.
 
-#### Creating a publisher
+1. ``Group`` name
+2. ``Instance`` name
+3. ``Topic`` name
```
elBoberido, line 162: should we change this in `ServiceDescription` ctor and also the `AnyService`, `InvalidIDString`, etc. constants?

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+A triple consisting of such strings is called a ``ServiceDescription``. Two ``ServiceDescription``s are considered 
+matching if all these three strings are element-wise equal, i.e. group, instance and topic names are the same for both 
+of them. This means the group and instance identifier can be ignored to create different ``ServiceDescription``s. They 
+will be needed for advanced filtering functionality in the future.
 
-```cpp
-iox::popo::UntypedPublisher publisher({"Group", "Instance", "CounterTopic"});
-publisher.offer();
-```
+The service model of iceoryx is derived from AUTOSAR and is still used in the API with these names. The so called
```
elBoberido, line 169: okay, it's not gonna be changed for v1.0.0. I think it might make sense to add `... with these names(Service, Instance, Event).`, although it is in the table below.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-Since the data type is not known to the publisher, we have to provide the size in bytes of the payload data we
-intend to send.
+#### Restrictions
 
-If we successfully acquired a chunk, we can construct the data to be send using placement new and publish it.
-Notice that ``sample.get()`` returns a ``void*`` to the memory ``chunk`` were we can then place our data.
-Depending on what we want to send, we may also use ``memcpy`` to copy the data to the ``chunk``.
+The data type of the transmitted data can be any C++ class, struct or plain old data type as long as it satisfies the 
+following conditions:
+- no heap is used
```
mossmaurice, line 203: Please add and additional line break between line 202 / 203. Otherwise MkDocs does not detect is as an unordered list.

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
 
-Now that we have applications capable of sending and receiving data, we can run the complete iceoryx system.
+The API is offered in two languages, C++ and C. Detailed information can be found in the 
+[C++ example](../examples/icedelivery/) and 
+[C example](../examples/icedelivery_in_c/).
 
-First we need to start RouDi.
+Many parts of the C++ API follow a functional programming approach which is less error-prone. This requires using 
+the monadic types ``cxx::expected`` and ``cxx::optional`` which are introduced 
+[here](http://iceoryx.io/advanced/usage-guide/how-optional-and-error-values-are-returned-in-iceoryx/).
```
mossmaurice, line 287: Could you change this link to `../advanced/how-optional-and-error-values-are-returned-in-iceoryx.md`?

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-```
-# If installed and available in PATH environment variable
-iox-roudi
-# If build from scratch with script in tools
-$ICEORYX_ROOT/build/posh/iox-roudi
-```
+We distinguish between the ``typed API`` and the ``untyped API``. In the typed API, the underlying data type is made
+apparent by typed pointers or references to some data type T (often a template parameter). This allows working with
+the data in a C++ idiomatic and type-safe way and should be preferred whenever possible. The typed API is mainly used 
+when iceroryx is used stand-alone, i.e. not integrated into a third party framework.
```
mossmaurice, line 292: typo iceoryx

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
+the monadic types ``cxx::expected`` and ``cxx::optional`` which are introduced 
+[here](http://iceoryx.io/advanced/usage-guide/how-optional-and-error-values-are-returned-in-iceoryx/).
 
-```
-# If installed and available in PATH environment variable
-iox-roudi
-# If build from scratch with script in tools
-$ICEORYX_ROOT/build/posh/iox-roudi
-```
+We distinguish between the ``typed API`` and the ``untyped API``. In the typed API, the underlying data type is made
```
mossmaurice, line 289: With the C++ API we distinguish

mossmaurice (commented):
Great overview document!

Two wishes from my side:

1) Could you run a markdown linter over the files? There are e.g. some trailing white spaces.
2) Somehow the ping pong merges between `release_1.0` and `master` lead to the issue that some old commits are now showing up under the commit tab. I don't understand why this is happening. @dkroenke we should avoid ping pong merges in future and applying the delta to release and master branch separately and never join them.
@FerdinandSpitzschnueffler Could you please rebase your changes to the latest commit on master?

doc/website/getting-started/overview.md
elBoberido: I would at least tell that there is a last lock. This reads like we do not have locks at all so it would be quite embarrassing getting crash reports of RouDi at cleaning up a locked mutex. Something like
```
Due to our mostly lock-free inter-process mechanisms (only one last lock; we are working on it),  iceoryx-based communication is much more reliable compared to traditional mechanism using locking.
```
or
```
Due to our lock-free inter-process mechanisms*,  iceoryx-based communication is much more reliable compared to traditional mechanism using locking.
*) there is one last lock but we are working to remove it
```

doc/website/getting-started/overview.md
```diff
@@ -11,637 +10,290 @@ runtime and create _publishers_ and _subscribers_. The publishers send data of a
...
-
+Now we can use the publisher to send the data.
 ```cpp
 auto result = publisher.loan();
-if (!result.has_error())
+if(!result.has_error())
 {
     auto& sample = result.value();
-    sample->counter = 73;
+    sample->counter = 30;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: 'cause I'm not that old ;)

doc/website/getting-started/overview.md
```diff
@@ -302,346 +52,262 @@ else
...
-a ``nullopt``. Otherwise we can get the pointer of our data from the sample and process the received data.
+RouDi is an abbreviation for **Rou**ting and **Di**scovery. RouDi takes care of the communication setup but does not
+actually participate in the communication between the publisher and the subscriber. RouDi can be thought of as the
+switchboard operator of iceoryx. One of its other major tasks is the setup of the shared memory, which the
+applications use for exchanging payload data. Sometimes referred to as daemon, RouDi manages the shared memory and is
+responsible for the service discovery, i.e. enabling subscribers to find topics offered by publishers. It also keeps
+track of all applications which have initialized a runtime and are hence able to use publishers or subscribers.
 
-We can shorten this somewhat by using a functional approach.
+When an application terminates, RouDi cleans up all resources needed for the communication with RouDi. If an
```
elBoberido, line 133: we have to distinguish here between graceful and abnormal termination since there is the same sentence on line 117 `When the application terminates`

elBoberido (commented):
looks good. Just a few minor thing and please use a single »`« for inline code blocks

mossmaurice (approved):

elBoberido (approved):

PR #649, opened 2021-03-29
Iox #14 custom header for untyped C++ and C API

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR adds the custom header support to the C++ an C API and refactors the c++2c enum translation tests in order to discover missing enum values on the C side more easily.

## References

- Partly closes #14


iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
```diff
@@ -25,9 +25,13 @@ namespace iox
 {
 namespace popo
 {
-template <typename base_publisher_t = BasePublisher<>>
+template <typename H = mepoo::NoCustomHeader, typename base_publisher_t = BasePublisher<>>
 class UntypedPublisherImpl : public base_publisher_t
```
elBoberido, line 29: Instead of having this as template parameter, it would also be possible to do it like with the C API and have a method like.
```
template <typename H>
bool configureCustomHeader() noexcept;
```
Alternatively, it could also be put into the publisher options on the C++ side but then there are options which would be discarded with the typed API.

There will be an additional feature with a hook to access the custom header e.g. on sample delivery which could be a callback with the signature
```
void(const ChunkHeader*, H*)
```
With the current approach, this can be made typesafe.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
```diff
@@ -75,6 +81,9 @@ class UntypedPublisherImpl : public base_publisher_t
 
 using UntypedPublisher = UntypedPublisherImpl<>;
 
+template <typename H>
+using UntypedPublisherWithCustomHeader = UntypedPublisherImpl<H>;
```
elBoberido, line 85: Not sure if this is the best approach or if it's better to just have a
```
template <typename H>
using UntypedPublisher = UntypedPublisherImpl<H>;
```

Nevertheless, I think it might be a good idea to have something similar in the typed API to prevent the user from setting a different port.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -99,11 +99,31 @@ void iox_pub_deinit(iox_pub_t const self)
...
 
-iox_AllocationResult iox_pub_loan_chunk(iox_pub_t const self, void** const chunk, const uint32_t payloadSize)
+bool iox_pub_configure_custom_header(iox_pub_t const self,
+                                     const uint32_t customHeaderSize,
+                                     const uint32_t customHeaderAlignment)
 {
-    auto result = PublisherPortUser(self->m_portData)
-                      .tryAllocateChunk(payloadSize, CHUNK_DEFAULT_PAYLOAD_ALIGNMENT)
-                      .and_then([&chunk](ChunkHeader* h) { *chunk = h->payload(); });
+    auto settingsCheck = iox::mepoo::ChunkSettings::create(0U, 1U, customHeaderSize, customHeaderAlignment);
```
mossmaurice, line 106: Shouldn't `IOX_C_DEFAULT_USER_PAYLOAD_ALIGNMENT` be used here?

iceoryx_binding_c/source/cpp2c_enum_translation.cpp
```diff
@@ -63,6 +63,10 @@ iox_AllocationResult AllocationResult(const iox::popo::AllocationError value)
         return AllocationResult_RUNNING_OUT_OF_CHUNKS;
     case AllocationError::TOO_MANY_CHUNKS_ALLOCATED_IN_PARALLEL:
         return AllocationResult_TOO_MANY_CHUNKS_ALLOCATED_IN_PARALLEL;
+    case AllocationError::INVALID_CHUNK:
```
mossmaurice, line 66: How could they get out of sync? Shouldn't we have tests to check if they are in sync?

iceoryx_binding_c/test/moduletests/test_cpp2c_enum_translation.cpp
```diff
@@ -39,11 +73,29 @@ TEST(cpp2c_enum_translation_test, SubscribeState)
...
+        case iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE:
+            EXPECT_EQ(cpp2c::ChunkReceiveResult(chunkReceiveResult.cpp), chunkReceiveResult.c);
+            break;
+        case iox::popo::ChunkReceiveResult::TOO_MANY_CHUNKS_HELD_IN_PARALLEL:
+            EXPECT_EQ(cpp2c::ChunkReceiveResult(chunkReceiveResult.cpp), chunkReceiveResult.c);
+            break;
+        case iox::popo::ChunkReceiveResult::INVALID_STATE:
+            EXPECT_EQ(cpp2c::ChunkReceiveResult(chunkReceiveResult.cpp), chunkReceiveResult.c);
+            break;
+            // default intentionally left out in order to get a compiler warning if the enum gets extended and we forgot
```
mossmaurice, line 95: :+1:

iceoryx_binding_c/test/moduletests/test_publisher.cpp
```diff
@@ -158,24 +158,69 @@ TEST_F(iox_pub_test, noSubscribersAfterUnsubscribe)
     EXPECT_FALSE(iox_pub_has_subscribers(&m_sut));
 }
 
+TEST_F(iox_pub_test, configureCustomHeaderIsSuccessful)
```
mossmaurice, line 161: Capitalise all unit test cases?

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_publisher.inl
```diff
@@ -22,25 +22,27 @@ namespace iox
 {
 namespace popo
 {
-template <typename base_publisher_t>
-inline UntypedPublisherImpl<base_publisher_t>::UntypedPublisherImpl(const capro::ServiceDescription& service,
-                                                                    const PublisherOptions& publisherOptions)
+template <typename H, typename base_publisher_t>
```
mossmaurice, line 25: `HeaderType` or `Header_t`? I think we're really inconsistent with naming template params, makes reading harder :disappointed:

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
mossmaurice: Personally I like the template param approach. Easy and clean. I think it doesn't belong to the publisher options. They are described in the re-written examples as "QoS options". We should be aligned with DDS implementations here.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
```diff
@@ -38,22 +42,24 @@ class UntypedPublisherImpl : public base_publisher_t
...
+    /// @brief Get a chunk from loaned shared memory.
+    /// @param payloadSize The expected user payload size of the chunk.
+    /// @param payloadAlignment The expected user payload alignment of the chunk.
     /// @return A pointer to a chunk of memory with the requested size or
     ///         an AllocationError if no chunk could be loaned.
     /// @note An AllocationError occurs if no chunk is available in the shared memory.
     ///
-    cxx::expected<void*, AllocationError> loan(const uint32_t size) noexcept;
+    cxx::expected<void*, AllocationError> loan(const uint32_t payloadSize,
+                                               const uint32_t payloadAlignment = 8U) noexcept;
```
mossmaurice, line 53: Is the `8` not defined in some constexpr value already?

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
mossmaurice: I'd keep `UntypedPublisher` the template param will default in most cases.

elBoberido: That was also my first approach but then one has to use `UntypedPublisher<>` which might be cumbersome. But I guess it's a matter of taste and I will do what the majority wants

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_publisher.inl
elBoberido: It's just `H` to be consistent with the typed API. There was already a `T` template parameter and then I added the `H` one

iceoryx_binding_c/source/c_publisher.cpp
elBoberido: could be used, but it's not necessary since I want to check if the custom header arguments are correct, so any payload alignment will do and `1U` was less to type ;)

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
elBoberido: This was a placeholder in the beginning since I was not sure if this should be an optional parameter. I kept it that way but I'm still not sure if it should be mandatory to pass the payload alignment

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_publisher.inl
mossmaurice: Please change the template param in the typed API, too. Same goes for the `base_publisher_t`. I think in iceoryx most of the time we use `FooType`.

iceoryx_binding_c/test/moduletests/test_publisher.cpp
elBoberido: I would say that it's not part of this PR :)

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_publisher.inl
elBoberido: I can change `base_publisher_t` (in a separate PR together with the typed API) but would keep the `H`.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -99,11 +99,31 @@ void iox_pub_deinit(iox_pub_t const self)
...
+    if (settingsCheck.has_error())
+    {
+        return false;
+    }
+
```
MatthiasKillat, line 111: Logic seems a bit weird/redundant since in the true case settingsCheck is not used anymore. The way I understand it this is just to check validity of the arguments before setting them in the publisher.

iceoryx_binding_c/source/c_publisher.cpp
```diff
@@ -99,11 +99,31 @@ void iox_pub_deinit(iox_pub_t const self)
...
+}
+
+iox_AllocationResult iox_pub_loan_chunk(iox_pub_t const self,
+                                        void** const chunk,
+                                        const uint32_t payloadSize,
+                                        const uint32_t payloadAlignment)
+{
+    auto result =
+        PublisherPortUser(self->m_portData)
+            .tryAllocateChunk(payloadSize, payloadAlignment, self->m_customHeaderSize, self->m_customHeaderAlignment)
```
MatthiasKillat, line 125: Works, but probably is suboptimal performance wise to pass it every time to the port (maybe it is optimized, did not check). Ideally the port or the entity creating/allocating the chunk would know the size and alignment. Oh well, it is basically an aligned alloc (at least partly).
It is not likely to change often or at all over the lifetime of the publisher I think.

iceoryx_binding_c/source/cpp2c_enum_translation.cpp
```diff
@@ -76,6 +80,8 @@ iox_WaitSetResult WaitSetResult(const iox::popo::WaitSetError value)
         return WaitSetResult_WAIT_SET_FULL;
     case WaitSetError::EVENT_ALREADY_ATTACHED:
         return WaitSetResult_EVENT_ALREADY_ATTACHED;
+    case WaitSetError::PROVIDED_HAS_TRIGGERED_CALLBACK_IS_UNSET:
```
MatthiasKillat, line 83: Judging by the wording I do not know whether this can even happen (can we unset the callback during waiting?). Has probably not much to do with the chunk header though. Weird error name anyway.

iceoryx_binding_c/source/cpp2c_enum_translation.cpp
```diff
@@ -89,6 +95,8 @@ iox_ListenerResult ListenerResult(const iox::popo::ListenerError value)
         return ListenerResult_EVENT_ALREADY_ATTACHED;
     case ListenerError::LISTENER_FULL:
         return ListenerResult_LISTENER_FULL;
+    case ListenerError::EMPTY_INVALIDATION_CALLBACK:
```
MatthiasKillat, line 98: Similar kind of non existent callback error but worded differently. Consistency?

iceoryx_binding_c/test/moduletests/test_publisher.cpp
MatthiasKillat: I do like consistency but please prioritize the important things. This is not one of them, besides I prefer the lower case version.

iceoryx_examples/callbacks_in_c/ice_c_callbacks_publisher.c
```diff
@@ -51,7 +51,8 @@ void sending()
     {
         if (counter % 3 == 0)
         {
-            if (iox_pub_loan_chunk(publisherLeft, (void**)&chunk, sizeof(struct CounterTopic))
+            if (iox_pub_loan_chunk(
+                    publisherLeft, (void**)&chunk, sizeof(struct CounterTopic), IOX_C_DEFAULT_USER_PAYLOAD_ALIGNMENT)
```
MatthiasKillat, line 55: Great, the new argument will break my work were I use it I think...

iceoryx_posh/include/iceoryx_posh/internal/popo/untyped_publisher.inl
MatthiasKillat: I vote for short dummy variables in the inl, they do not have to be the same as in the hpp (readability). In fact, if they are not used even typename is enough. We need a convention for those _t suffixes but IMO they should not be used in template parameters.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
MatthiasKillat: This is not user API but I suppose the ser will only have to provide the template parameter if a non-default header is desired. Should be ok and the added type-safety is a plus.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
```diff
@@ -25,9 +25,13 @@ namespace iox
 {
 namespace popo
 {
-template <typename base_publisher_t = BasePublisher<>>
+template <typename H = mepoo::NoCustomHeader, typename base_publisher_t = BasePublisher<>>
 class UntypedPublisherImpl : public base_publisher_t
 {
+    static_assert(!std::is_const<H>::value, "The custom header must not be const.");
+    static_assert(!std::is_reference<H>::value, "The custom header must not be a reference.");
+    static_assert(!std::is_pointer<H>::value, "The custom header must not be a pointer.");
```
MatthiasKillat, line 33: I think it has to be a class? If so, is_class might be better and would make the pointer and reference checks obsolete I think. If we want to have e.g. primitive types as header we cannot use it though.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
MatthiasKillat: If we do not have a default alignment constant which we use elsewhere we can keep the 8U here. I find an optional argument ok but in the C API we will have to pass both. The other question is what is a reasonable default value? 8U targets 64 Bit architectures I guess and is fine.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
MatthiasKillat: We do not want to use ``UntypedPublisher<>`` which is the case that will occur mor often (most likely). So We need an alias for the custom header case. As for how it is to be named I have better proposal, and in doubt name it as it is (i.e. as you have).

First principles @elBoberido ,  simplicity:  "Of each particular thing, always ask what is it in itself."

So it is a HeadlessPublisher ...

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -75,13 +75,25 @@ iox_pub_t iox_pub_init(iox_pub_storage_t* self,
 /// @param[in] self the handle which should be removed
 void iox_pub_deinit(iox_pub_t const self);
 
+/// @brief configures the custom header size and alignment used with each iox_pub_loan_chunk call
+/// @param[in] customHeaderSize custom header size of the allocated chunk
+/// @param[in] customHeaderAlignment custom header alignment of the allocated chunk
+/// @return true if the parameters were valid, false otherwise
+bool iox_pub_configure_custom_header(iox_pub_t const self,
+                                     const uint32_t customHeaderSize,
+                                     const uint32_t customHeaderAlignment);
```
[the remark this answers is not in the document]
elBoberido: Well, it's now of after at the 2.0 release. The nice think when it's put on the options would be to not be able to change it after creation, which would make it similar to the C++ API

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
elBoberido: @MatthiasKillat I'm desperately waiting for better names. Both the Headless and Headed Untyped Publisher should be related with from the names

doc/design/chunk_header.md
```diff
@@ -273,7 +273,8 @@ class MyChunkHeaderHook : public ChunkHeaderHook
 auto customHeaderHook = MyChunkHeaderHook<MyHeader>();
 auto pub = iox::popo::Publisher<MyPayload>(serviceDescription, customHeaderHook);
 ```
-        - it has to be decided if it's a good idea to give the user easy write access to the `ChunkHeader` or just to the `CustomHeader`
+        - alternatively, instead of the ChunkHeaderHook class, the publisher could have a method `registerDeliveryHook(std::function<void(ChunkHeader&)>)`
+        - allow the user only read access to the `ChunkHeader` and write access to the `CustomHeader`
```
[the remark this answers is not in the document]
elBoberido: an option would be plain C function pointer but then we would make it harder to use. I also thought about `cxx::function`, which would replace the `std::function` when it's ready. The API with the std::function would be marked as deprecated.

iceoryx_binding_c/include/iceoryx_binding_c/internal/cpp2c_publisher.hpp
```diff
@@ -30,6 +31,8 @@ struct cpp2c_Publisher
     cpp2c_Publisher& operator=(cpp2c_Publisher&& rhs) noexcept;
 
     iox::popo::PublisherPortData* m_portData{nullptr};
+    uint32_t m_customHeaderSize{iox::CHUNK_NO_CUSTOM_HEADER_SIZE};
+    uint32_t m_customHeaderAlignment{iox::CHUNK_NO_CUSTOM_HEADER_ALIGNMENT};
```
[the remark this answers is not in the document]
elBoberido: No, this as just default values in case no custom header is used, i.e. 0 for size and 1 for alignment. There is always the restriction with alignment as power of two and size as multiple of alignment

iceoryx_binding_c/source/c_publisher.cpp
elBoberido: yes, the actual check happens when the user calls `iox_pub_loan_chunk` but since configuring the custom header with this function might lead to wrong parameters, with this check it is ensured that with `iox_pub_loan_chunk` only the payload parameter can cause an error

iceoryx_binding_c/source/c_publisher.cpp
elBoberido: good point, but that's not user facing API and could easily changed afterwards

iceoryx_binding_c/source/cpp2c_enum_translation.cpp
elBoberido: yes, I just noticed that some enum values were missing in the translation and updated them

iceoryx_examples/callbacks_in_c/ice_c_callbacks_publisher.c
elBoberido: of course it will. I'm open for suggestions. Shall I create a second function like `iox_pub_loan_aligned_chunk`?

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
elBoberido: At the moment it would be possible. Don't know it the restriction would harm, but since the code is already there, I would keep it as is.

iceoryx_binding_c/include/iceoryx_binding_c/internal/cpp2c_publisher.hpp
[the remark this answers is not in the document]
MatthiasKillat: alignment 1 is unrestricted. Can be any multiple of 1. Considering it must be an integer value ... it can be any ;-)

iceoryx_binding_c/source/c_publisher.cpp
MatthiasKillat: Sure, low prio and user API ans functionality is the main focus. Just an observation, nothing to fix.

iceoryx_examples/callbacks_in_c/ice_c_callbacks_publisher.c
MatthiasKillat: I do not think so, calling this without alignment will lead to subtle errors, better to force the user to do so.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
MatthiasKillat: Sure. Just asking what the header can be configured to be. With what you have it can be any non-const value type which makes sense.

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
mossmaurice: Regarding naming, I don't have a better idea other than `UntypedPublisherWithCustomHeader`. That's what's inside, so it's fine.

@elBoberido Just realized, we need an example for the custom header feature! Have you planned to write one till Thursday?

elBoberido: @mossmaurice not really, but I talked with @budrus and we decided to add it after the RC. Important would be to have the stable API until the release

iceoryx_examples/callbacks_in_c/ice_c_callbacks_publisher.c
elBoberido: There wouldn't be much code duplication, since `iox_pub_loan_chunk` would call `iox_pub_loan_aligned_chunk`. The more difficult question would be if we call it `iox_pub_loan_aligned_chunk` or `iox_pub_loan_chunk_with_alignment`?

elBoberido: @MatthiasKillat  elfenpiff had a third option. He suggested to use a variadic function. What do you think about that. His second choice would be `iox_pub_loan_aligned_chunk`

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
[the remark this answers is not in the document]
elBoberido: moved it to the options to make it similar to the C++ API where it also cannot be changed after the construction of the publisher

iceoryx_examples/callbacks_in_c/ice_c_callbacks_publisher.c
elBoberido: forget the variadic function. We can not determine if the optional parameter is set. Therefore it is `iox_pub_loan_aligned_chunk`

iceoryx_posh/include/iceoryx_posh/popo/untyped_publisher.hpp
elBoberido: Kkay, I think the naming is the last blocker for the merge. Please make proposals

MatthiasKillat: I have none, CustomPublisher e.g. would not be sufficient to determine what can be customized (but be significantly shorter). The fact that it is untyped would also be implicit from use. I would say keep it like this for now.

MatthiasKillat (approved):

FerdinandSpitzschnueffler (approved):

PR #655, opened 2021-03-29
Iox #482 Implement iceoptions, icehello, iceensemble and re-write icedelivery example

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Create icehello
* Create iceoptions example
* Replace ice_multi_publisher with iceensemble example
* `s/-ex/-cpp` in binary and runtime names

### Todo
* icecubetray example will be added in a follow-up PR

## References

- Closes partly #482


iceoryx_examples/icedelivery/README.md
```diff
@@ -63,38 +50,40 @@ struct RadarObject
...
-
 Now comes the work mode. Data needs to be created. But hang on.. we need memory first! Let's reserve a memory chunk
 which fits our RadarObject struct
+
 ```cpp
-auto result = untypedPublisher.loan(sizeof(RadarObject));
+auto result = publisher.loan(sizeof(RadarObject));
 ```
 
+#### #1 Loan and emplace
```
elfenpiff, line 80: This is not the way we endorse! We suggest to use always the functional approach and offer the traditional way only as an alternative. This should be reflected here so that the functional approach comes first and this appraoch is shown as an alternative.

iceoryx_examples/icedelivery/README.md
```diff
@@ -105,48 +94,54 @@ else
 }
 ```
 
-Or try the functional way (#2) by concatenating `and_then` and `or_else`. Read it like a story in a book: "Loan memory
-and then if it succeeds, fill it with some data or else if it fails, handle the error"
+#### #2 Functional approach with loaning
```
elfenpiff, line 97: This is the standard approach therefore I would call it: "loan and publish"

iceoryx_examples/icedelivery/README.md
elfenpiff, line 80: Since this is the not endorsed alternative appraoch I would call it: alternative loan and publish

iceoryx_examples/icedelivery/README.md
```diff
@@ -105,48 +94,54 @@ else
...
+
+Or try the functional way (#2) by concatenating `and_then` and `or_else`. Well, that's a bit of a
+[lambda](https://en.wikipedia.org/wiki/Anonymous_function#C++_(since_C++11)) jungle. Read it like a story in a book:
+"Loan memory and then if it succeeds, fill it with some data or else if it fails, handle the error"
```
elfenpiff, line 101: This is great!

iceoryx_examples/icedelivery/README.md
```diff
@@ -157,82 +152,72 @@ captured with the signal handler and stops the loop.
...
-    {
-        std::cout << "Error receiving chunk." << std::endl;
-    });
-```
-
-Well, that's a bit of a [lambda](https://en.wikipedia.org/wiki/Anonymous_function#C++_(since_C++11)) jungle. Let's
-translate it into a story again: "Take the data and then if this succeeds, work with the sample, if the sample is empty
-do something different, or else if an error occured, print the string 'Error receiving chunk.'" Of course you don't
-need to take care about all cases, but it is advised to do so.
+Let's translate it into a story again: "Take the data and then if this succeeds, work with the sample, or else if an
```
elfenpiff, line 202: I love your stories! I think they will help a lot of people. 

Maybe you should place them even more prominently and introduce `and_then` / `or_else` before jumping into this code here with your splendid story!

iceoryx_examples/icedelivery/README.md
```diff
@@ -242,25 +227,62 @@ The typed publisher application is an example for a high-level user API and does
...
 
 When it comes to the runtime, things are the same as in the untyped publisher. However, a typed publisher object is
-created
+created and the transmitted data type is provided as template parameter:
+
 ```cpp
 iox::popo::Publisher<RadarObject> publisher({"Radar", "FrontLeft", "Object"});
 ```
 
+The topic type must be default- and copy-constructible when the typed API is used.
```
elfenpiff, line 242: - The topic type does not need to be default constructible. If you have a non default ctor you can provide all arguments to `publisher.load(arg1, arg2 ...)` and they will be forwarded to the constructor (like emplace in the optional).

- The topic type must be trivially copyable: This type traits has to return true: https://en.cppreference.com/w/cpp/types/is_trivially_copyable.
- As an additionaly description you could state that this means that the data is memcpy-able!

iceoryx_examples/icedelivery/README.md
```diff
@@ -242,25 +227,62 @@ The typed publisher application is an example for a high-level user API and does
...
+
 A similar while-loop is used to send the data to the subscriber. In contrast to the untyped publisher the typed one
-offers two additional possibilities
+offers three additional possibilities.
+
+#### #2 Functional approach with loaning
```
elfenpiff, line 247: Wrong header: Classic approach with loaning since we have a `if` `else` in the code. 
But the order is correct, please leave it has an alternative and second approach and endorse the `and_then`/`or_else` first!

iceoryx_examples/icedelivery/README.md
```diff
@@ -242,25 +227,62 @@ The typed publisher application is an example for a high-level user API and does
...
+{
+    // ..
+}
+```
+
+One might wonder what the type of the variable `sample` is? It is `iox::popo::Sample<RadarObject>`. This class behaves
+similar to a [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr) and makes sure that the ownership
+handling is done automatically and memory is freed when going out of scope on subscriber side.
+
+#### #4 Publish by copy
```
elfenpiff, line 271: Where is `#1` and `#3`?

iceoryx_examples/icedelivery/README.md
```diff
@@ -242,25 +227,62 @@ The typed publisher application is an example for a high-level user API and does
...
 auto object = RadarObject(ct, ct, ct);
 publisher.publishCopyOf(object).or_else([](iox::popo::AllocationError) {
     // Do something with error.
 });
 ```
-This should only be used for small data types, as otherwise copies can lead to a larger runtime.
+
+#### #5 Publish the result of a computation
+
+Usage #5 can be useful if you have a callable e.g. a function or [functor](https://en.wikipedia.org/wiki/Function_object#In_C_and_C++) should be always called
```
elfenpiff, line 286: The required signature of the functor should be stated and explained here! `void(SampleType*)`. Additionally, I would explain roughly how it works:
1. publisher loans sample from shared memory
2. if loaning was successful a pointer to the sample is provided to the functor
3. if loaning was not successful the functor is not called and an error is returned to the user

iceoryx_examples/icedelivery/iox_publisher_untyped.cpp
```diff
@@ -58,17 +58,16 @@ int main()
             RadarObject* data = new (chunk) RadarObject(ct, ct, ct);
             iox::cxx::Expects(chunk == data);
 
-            data->x = 1.0;
-            data->y = 2.0;
-            data->z = 3.0;
+            data->x = ct;
+            data->y = ct;
```
elfenpiff, line 62: could we differ the values somehow, maybe like this:
```cpp
data->x = ct;
data->y = ct * ct;
data->z = ct + 2.5;
```

so that you can distinct them in the output.

iceoryx_examples/iceensemble/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+A common use case is that often we have multiple sensors sending data of the same type, e.g. LIDAR data, and a subscriber is interested in the data of all those sensors.
+
+This example demonstrates
+
+1. how you can run multiple publisher applications publishing on the same topic (n:m communication)
+2. how to communicate between C and C++
+
+!!! warning
+    If you have built iceoryx with the CMake flag `ONE_TO_MANY_ONLY` this example will not run
```
elfenpiff, line 13: Is this by default on or do we have to explicitly activate this? Could you please mention this also in the markdown file.

iceoryx_examples/iceensemble/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Alternatively you can use the provided [tmux](https://en.wikipedia.org/wiki/Tmux) script.
+
+```sh
+./run_iceensemble.sh
+```
+
+## Expected output
+
+<!-- add asciiema link here -->
```
elfenpiff, line 40: Where is the asciinema?

iceoryx_examples/iceensemble/README.md
```diff
@@ -0,0 +1,40 @@
...
+./build/iceoryx_examples/iceoptions/iox-subscriber-with-options
+```
+
+Alternatively you can use the provided [tmux](https://en.wikipedia.org/wiki/Tmux) script.
+
+```sh
+./run_iceensemble.sh
+```
+
+## Expected output
```
elfenpiff, line 38: Could you please correct all `Expected output` to `Expected Output`. Use this maybe: https://capitalizemytitle.com/

iceoryx_examples/icehello/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+project(example_icehello)
+
+include(GNUInstallDirs)
+
+find_package(iceoryx_posh CONFIG REQUIRED)
+find_package(iceoryx_utils CONFIG REQUIRED)
+
+get_target_property(ICEORYX_CXX_STANDARD iceoryx_posh::iceoryx_posh CXX_STANDARD)
+if ( NOT ICEORYX_CXX_STANDARD )
+    include(IceoryxPlatform)
```
elfenpiff, line 28: We have to always include this otherwise the `${ICEORYX_WARNINGS}` are empty and we have to fix warnings again in examples.

So could you please remove the `if` around `include(IceoryxPlatform)`

elfenpiff: Take a look at: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/icedelivery/CMakeLists.txt in master as a reference

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+Next up, we include the publisher and the runtime:
+
+```cpp
+#include "iceoryx_posh/popo/publisher.hpp"
+#include "iceoryx_posh/runtime/posh_runtime.hpp"
+```
+
+For the communication with the daemon RouDi, a runtime object is created. We use a unique string for identifying our application:
```
elfenpiff, line 49: We create a runtime object to communicate with the RouDi daemon.

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+```cpp
+constexpr char APP_NAME[] = "iox-cpp-publisher-helloworld";
+iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+```
+
+Up next, we create a publisher instance for our charming struct, notice the topic type is passed as a template parameter:
```
elfenpiff, line 56: charmin struct. Notice, the topic type...

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+```cpp
+iox::popo::Publisher<RadarObject> publisher({"Radar", "FrontLeft", "Object"});
+```
+
+The three strings passed as parameter to the constructor of `iox::popo::Publisher` are of the type
```
elfenpiff, line 62: The three strings that are passed as parameters to the constructor of `iox::popo::Publisher` form our `capro::ServiceDescription`.

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+```
+
+Next, the subscriber object is created, again passing the topic type `RadarObject` as template parameter:
+
+```cpp
+iox::popo::Subscriber<RadarObject> subscriber({"Radar", "FrontLeft", "Object"});
+```
+
+Publisher and subscriber will only be connected if they both use exactly the same three strings.
```
elfenpiff, line 140: exactly the three strings - our service description. (so that the reader gets used to the term)

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+Inside the `while` loop, we take the topic from shared memory and print it if taking was successfull:
```
elfenpiff, line 142: we take the sample from the shared memory and print it if we acquired it successfully.

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+{
+    std::cout << "No chunk available." << std::endl;
+}
+else
+{
+    std::cout << "Error receiving chunk." << std::endl;
+}
+```
+
+The subscriber application runs ten times faster than the publisher and should therefore never miss any samples:
```
elfenpiff, line 165: This requires some more explanation! Maybe refer here to more complex examples and say that this also depends on the configuration of the subscriber (history size etc.)

iceoryx_examples/icehello/iox_subscriber_helloworld.cpp
```diff
@@ -14,50 +14,58 @@
...
-int main()
-{
-    auto signalIntGuard = iox::posix::registerSignalHandler(iox::posix::Signal::INT, sigHandler);
-    auto signalTermGuard = iox::posix::registerSignalHandler(iox::posix::Signal::TERM, sigHandler);
-    iox::runtime::PoshRuntime::initRuntime("iox-subscriber");
+            std::cout << APP_NAME << " got value: " << takeResult.value()->x << std::endl;
+        }
+        else
+        {
+            if (takeResult.get_error() == iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
```
elfenpiff, line 57: Aren't we here flooded with "No chunk available" messages since the subscriber is 10 times faster than the publisher?

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
+# iceoptions
+
+## Introduction
+
+This example demonstrates what kind of quality of service options can be configured on the publisher and subscriber
+side. The options can be used for both, typed and untyped API flavours.
+
+## Expected output
+
+<!-- Add asciinema link here -->
```
elfenpiff, line 10: Either add the asciinema here or get rid of this section otherwise it looks like we have forgot to add the documentation here and that this file is either out of date or incomplete

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Code walkthrough
+
+### Publisher
+
+In order to configure a publisher, we have to supply a struct of the type `iox::popo::PublisherOptions` as a second parameter.
+
+`historyCapacity` will enable subscribers to read the last n topics e.g. in case they are started later than the publisher:
```
elfenpiff, line 18: last n samples.

Could you please reread the whole documentation and look for the word topic and replace it with sample when it makes sense.

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
...
+```cpp
+publisherOptions.nodeName = "Pub_Node_With_Options";
+```
+
+### Subscriber
+
+To configure a subscriber, we have to supply a struct of the type `iox::popo::SubscriberOptions` as a second parameter.
+
+The `queueCapacity` parameter specifies how many samples the queue of the subscriber object can hold. If the queue
+would encounter an overflow, the oldest sample is released to create space for the newest one, which is then stored.
```
elfenpiff, line 41: just mention here that the queue behaves like a circular buffer.

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```cpp
+subscriberOptions.queueCapacity = 10U;
+```
+
+`historyRequest` will enable a subscriber to receive the last n topics on subscription e.g. in case it was started later than the publisher. The publisher needs to have its `historyCapacity` enabled, too.
```
elfenpiff, line 47: last n samples on subscription

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
...
+`historyCapacity` will enable subscribers to read the last n topics e.g. in case they are started later than the publisher:
+
+```cpp
+publisherOptions.historyCapacity = 10U;
+```
+
+Topics are automatically offered on creation of a publisher, if you want to disable that feature and control the offering yourself, do:
+
+```cpp
+publisherOptions.offerOnCreate = false;
```
elfenpiff, line 27: Here you should also mention that you then have to call `publisher.offer`

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```cpp
+subscriberOptions.historyRequest = 5U;
+```
+
+Topics are automatically subscribed on creation, if you want to disable that feature and control the subscription
+yourself, set `subscribeOnCreate` appropriately:
+
+```cpp
+subscriberOptions.subscribeOnCreate = false;
```
elfenpiff, line 57: Here you should also mention that you then have to call `subscriber.subscribe`

elfenpiff (dismissed):
- Please look for all occurrence of the word `topic` and replace it with `sample` when it makes sense
- Please use grammarly or another spell checker for the markdown files

iceoryx_examples/callbacks/README.md
```diff
@@ -164,3 +164,7 @@ Afterwards we reset both caches to start fresh again.
         rightCache.reset();
     }
 ```
+
+<center>
+[Check out callbacks on GitHub :fontawesome-brands-github:](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_examples/callbacks){ .md-button }
```
FerdinandSpitzschnueffler, line 169: I like the idea to have a button to jump from the website to the code on github. But I don't like that you can see this raw line when you are on the github readme...

iceoryx_examples/callbacks_in_c/ice_c_callbacks_publisher.c
```diff
@@ -34,7 +34,7 @@ static void sigHandler(int f_sig)
 
 void sending()
 {
-    iox_runtime_init("iox-ex-callbacks-publisher");
+    iox_runtime_init("iox-cpp-callbacks-publisher");
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: The runtime name of the subscriber already has the right format.

iceoryx_examples/icecrystal/Readme.md
```diff
@@ -18,9 +18,9 @@ iox-roudi
 # If build from scratch with script in tools
 $ICEORYX_ROOT/build/iox-roudi
 
-build/iceoryx_examples/icedelivery/iox-ex-publisher-untyped
+build/iceoryx_examples/icedelivery/iox-cpp-publisher
 
-build/iceoryx_examples/icedelivery/iox-ex-subscriber-untyped
+build/iceoryx_examples/icedelivery/iox-cpp-subscriber
```
FerdinandSpitzschnueffler, line 23: The expected output probably has changed since the runtime names and the introspection changed. Please add a new screenshot.

FerdinandSpitzschnueffler (commented):
Reviewed the first 16 files, will continue later.

iceoryx_examples/icedelivery/README.md
```diff
@@ -157,82 +152,72 @@ captured with the signal handler and stops the loop.
 How can the subscriber application receive the data the publisher application just transmitted?
 
 Similar to the publisher we need to include the runtime and the subscriber as well as the topic data header:
```
FerdinandSpitzschnueffler, line 154: + signal handler

iceoryx_examples/callbacks/README.md
mossmaurice: I know what you mean. Same goes for the admonitions `!!! warnig`. However, I'd like to stick with the current approach for two reasons:
1) Most readers will probably read it in an IDE, there it's very readable, also other things like formulars with [MathJax](https://squidfunk.github.io/mkdocs-material/reference/mathjax/)
2) The GitHub markdown viewer will not parse these items correctly, with admonitions this is a feature, they strike the eye of the reader and get attention. With the button it does not disrupt the reading flow as it is placed at the very end.
3) Having a single source out weights IMHO all other points

iceoryx_examples/icedelivery/README.md
```diff
@@ -242,25 +227,62 @@ The typed publisher application is an example for a high-level user API and does
 described before. In this summary, just the differences to the prior publisher application are described.
```
FerdinandSpitzschnueffler, line 227: as the untyped publisher described before

iceoryx_examples/icedelivery/README.md
FerdinandSpitzschnueffler: @elfenpiff I agree with you at point one. Regarding your second point: if we force that we would not be able to send for instance the cxx::vector

iceoryx_examples/icedelivery/README.md
```diff
@@ -272,27 +294,24 @@ publisher.publishResultOf([&ct](RadarObject* object) { *object = RadarObject(ct,
...
 
-If you have a callable e.g. a function should be always called, this approach could be a good solution for you.
-
-Another difference compared to the untyped publisher, is the easier handling of `iox::popo::Sample`. There is no need
-for any casts with the typed publisher, as the type of the stored data is know. One can directly access the data with
-the `operator->()`.
-
 ### Subscriber application (typed)
 
 As with the typed publisher application there is an different include compared to the untyped subscriber:
```
FerdinandSpitzschnueffler, line 299: a different ...

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+For exiting on Ctrl-C, we register a `signalHandler`, that flips `bool killswitch`:
+
+```cpp
+#include "iceoryx_utils/posix_wrapper/signal_handler.hpp"
+// snip
+auto signalIntGuard = iox::posix::registerSignalHandler(iox::posix::Signal::INT, sigHandler);
+auto signalTermGuard = iox::posix::registerSignalHandler(iox::posix::Signal::TERM, sigHandler);
+```
+
+In order to send our topic we loan some some shared memory, inside the `while` loop:
```
FerdinandSpitzschnueffler, line 77: some of the somes is redundant ;)

iceoryx_examples/icehello/README.md
```diff
@@ -0,0 +1,173 @@
...
+    std::cout << APP_NAME << " got value: " << takeResult.value()->x << std::endl;
+}
+```
+
+In case an error occurred during taking, we need to handle it:
+
+```cpp
+if (takeResult.get_error() == iox::popo::ChunkReceiveResult::NO_CHUNK_AVAILABLE)
+{
+    std::cout << "No chunk available." << std::endl;
```
FerdinandSpitzschnueffler, line 157: Maybe we should refrain from printing that message. Since the subscriber application runs much faster than the publisher application, we will be flooded with this message.

iceoryx_examples/icehello/topic_data.hpp
```diff
@@ -13,23 +13,25 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-
-#ifndef IOX_EXAMPLES_MULTI_PUBLISHER_TOPIC_DATA_HPP
-#define IOX_EXAMPLES_MULTI_PUBLISHER_TOPIC_DATA_HPP
+#ifndef IOX_EXAMPLES_ICEHELLO_TOPIC_DATA_HPP
+#define IOX_EXAMPLES_ICEHELLO_TOPIC_DATA_HPP
 
 #include <cstdint>
```
FerdinandSpitzschnueffler, line 19: this include can be removed too

iceoryx_examples/iceoptions/README.md
```diff
@@ -0,0 +1,68 @@
+# iceoptions
+
+## Introduction
+
+This example demonstrates what kind of quality of service options can be configured on the publisher and subscriber
+side. The options can be used for both, typed and untyped API flavours.
```
FerdinandSpitzschnueffler, line 6: and they are available for the C++ and C API

iceoryx_examples/iceoptions/topic_data.hpp
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+#ifndef IOX_EXAMPLES_ICEOPTIONS_TOPIC_DATA_HPP
+#define IOX_EXAMPLES_ICEOPTIONS_TOPIC_DATA_HPP
+
+#include <cstdint>
```
FerdinandSpitzschnueffler, line 19: I think this include can be removed.

iceoryx_integrationtest/iceoryx_integrationtest/test_icehello_example.py
```diff
@@ -26,7 +26,7 @@
 
 import pytest
 
-# @brief Test goal: "Integrationtest for the multi publisher example of iceoryx"
+# @brief Test goal: "Integrationtest for the icehelloexample of iceoryx"
```
FerdinandSpitzschnueffler, line 29: icehello example

FerdinandSpitzschnueffler (approved):

FerdinandSpitzschnueffler (changes requested):
Please update the instructions in the iceensemble readme

iceoryx_examples/icedelivery/README.md
mossmaurice: Done

iceoryx_examples/icedelivery/README.md
mossmaurice: Thanks! I would add this to the new `expected` / `optional` document that @FerdinandSpitzschnueffler created.

iceoryx_examples/icedelivery/README.md
mossmaurice: Removed

iceoryx_examples/icedelivery/README.md
mossmaurice: These are already introduced in the untyped section, but I can add it again to make it clearer.

mossmaurice: Reworked by only using `and_then` / `or_else`

iceoryx_examples/icedelivery/README.md
mossmaurice: Done

iceoryx_examples/icedelivery/iox_publisher_untyped.cpp
mossmaurice: I'd keep things simple here and not distract the users. `RadarObject` is used to show a complex data type can be transmitted, instead of just transmitting `int`'s.

iceoryx_examples/iceensemble/README.md
mossmaurice: Added

iceoryx_examples/iceensemble/README.md
mossmaurice: Will be added once all example are finished. Most likely beginning of next week.

iceoryx_examples/icehello/CMakeLists.txt
mossmaurice: Done

iceoryx_examples/icehello/iox_subscriber_helloworld.cpp
mossmaurice: I quite liked this ouput when running icehello and iceensemble. The users learns here that this approach is a polling one. And not the best approach ;)

iceoryx_examples/iceoptions/README.md
mossmaurice: Will be done, once all examples are finalized.

iceoryx_examples/icehello/README.md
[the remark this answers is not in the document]
mossmaurice: Couldn't find a typo or grammar issue in this sentence. What do you meant?

iceoryx_examples/icehello/README.md
mossmaurice: :bee: :bee:

iceoryx_examples/icehello/README.md
mossmaurice: Actually, this was intentional to show that polling is not the best approach. Every 10th line a sample is received. I've extended the icehello readme in order to link to the other approaches, configuring history size or using the waitset.

iceoryx_examples/icedelivery/README.md
```diff
@@ -156,143 +119,182 @@ captured with the signal handler and stops the loop.
...
-iox::popo::Subscriber<RadarObject> subscriber({"Radar", "FrontLeft", "Object"}, subscriberOptions);
+iox::popo::Subscriber<RadarObject> subscriber({"Radar", "FrontLeft", "Object"});
 ```
 
 Everything else is nearly the same. However, there is one crucial difference which makes the `Subscriber` typed.
 
 Compare this line from the `UntypedSubscriber`
+
 ```cpp
 .and_then([](iox::popo::Sample<const RadarObject>& object)
```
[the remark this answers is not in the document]
mossmaurice: Oh dear, we need [geoffrey](https://github.com/elBoberido/geoffrey)

FerdinandSpitzschnueffler (dismissed):

dkroenke (approved):

FerdinandSpitzschnueffler (approved):

PR #702, opened 2021-04-09
Iox #482 Fix access right feature part1 and add corresponding example and integration test

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Fix permissions of UNIX domain sockets and shared memory segments
  * All apps and RouDi need to be in the same group
* Replace `cxx::MemoryManager` with `posix::User` inside `Process`
  *  When creating a publisher, check the access and notify user app in case no writable shared memory segment was found
* Add icecubetray example and corresponding integration test
* Add test for segment manager, creating more than one writer segment is not allowed

### Todo for v2.0 release (Moved to #722)

* Expected behaviour is that no data will ever be received from subscribers if the user has no read access to the shared memory segment to which the publisher is writing
  * This needs a runtime check in the discovery loop
  * Hence, add `posix::User` to `BasePort`
* Move `SegmentMapping` class to its own inl (segment_manager.hpp)
 

## References

- Closes partly #482


iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
```diff
@@ -0,0 +1,95 @@
...
+    # Assign users to group and disable login
+    sudo usermod -L perception -a -G privileged
+    sudo usermod -L infotainment -a -G unprivileged
+    sudo usermod -L perception -a -G iceoryx
+    sudo usermod -L infotainment -a -G iceoryx
+    sudo usermod -L roudi -a -G iceoryx
+    sudo usermod -L notallowed -a -G iceoryx
+
+    # Allow RouDi to send SIGKILL to other apps
+    sudo setcap cap_kill=ep $WORKSPACE/build/iceoryx_examples/icecubetray/iox-cpp-roudi-static-segments
```
mossmaurice, line 63: @marthtz That's what I've mentioned earlier. Please check if [this capability](https://man7.org/linux/man-pages/man7/capabilities.7.html) is feasible. IMHO it can only be a workaround. QNX has a [similar way](http://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.lib_ref/topic/p/procmgr_ability.html) of dealing with capabilites.

mossmaurice:
@budrus For now, I propse to leave out subscriber fix as described under `Todo for v2.0 release`. This will lead to a segfault in the subscriber when `take`ing chunks from a shared memory segment without read access.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+### RouDi and Apps
+
+#### Working setup
+
+RouDi is built with two static shared memory segments _infotainment_ and _privileged_. The access rights of the segments are configured as depicted in the graphic above.
+
+The `roudiConfig` is composed of a memory pool config called `mepooConfig`. When the segement is created, one needs to
+specific the reader group (first string), writer group (second string) as well as the `mepooConfig` (last parameter).
```
dkroenke, line 70: A short sentence would be good that the assignment of access rights is based on usergroups and not on explicit users.
Maybe an additional hint that all users in the readergroup are allowed to read but not write and the writergroup can do both.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+/// Create an Entry for a new Shared Memory Segment from the MempoolConfig and add it to the RouDiConfig
+roudiConfig.m_sharedMemorySegments.push_back({"unprivileged", "privileged", mepooConfig});
+roudiConfig.m_sharedMemorySegments.push_back({"infotainment", "infotainment", mepooConfig});
+```
+
+The radar app is started with the user _perception_ and is sending data into the _privileged_ shared memory segment.
+
+The display app is started with the user _infotainment_. It reads the topic `{"Radar", "FrontLeft", "Object"}` from the privileged segement and forwards it as a slighty modified topic `{"Radar", "HMI-Display", "Object"}`. Because the user _infotainment_ only has write access to the infotainment segment, the data is written to this segment.
+
+!!! hint
```
dkroenke, line 90: An additional hint would be good that the configuration can be done with TOML too.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+total 60268
+drwxrwxrwt  2 root  root         100 Apr  7 19:54 .
+drwxr-xr-x  6 root  root         460 Apr  6 15:53 ..
+-rw-rw----  1 roudi iceoryx 61383328 Apr  7 19:24 iceoryx_mgmt
+-rw-rw----+ 1 roudi iceoryx   160000 Apr  7 19:24 infotainment
+-rw-rw----+ 1 roudi iceoryx   160000 Apr  7 19:24 privileged
+```
+
+!!! note
+    Note the shared memory managment segment is always available for everyone to **read** and **write**
```
dkroenke, line 107: I guess you are reffereing the `iceoryx_mgmt` or?

elfenpiff, line 107: You demonstrated with your example output that this is not the case:
```
-rw-rw----  1 roudi iceoryx 61383328 Apr  7 19:24 iceoryx_mgmt
 ^  ^  ^
 |  |  |------- other (or everyone else, no rights)
 |  |---------- group (has read and write rights but no executable right)
 |------------- user (has read and write rights but no executable right)
```

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Not-working setup
+
+The cheeky app is started with the user _notallowed_. It has neither write nor read access to any shared memory segment. Hence, RouDi will print a warning in this case.
```
elfenpiff, line 111: Why has the user notallowed no read/write access (explain it that we have group based ACLs and the user is not member of any of those groups).

elfenpiff, line 111: The application will segfault whenever the application tries to write data into a memory segment to which it has no access rights (or only read access).

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+Despite having no read access, subscribers can still be created. In this case no data will ever arrive.
```
elfenpiff, line 113: Are you sure? I thought data will arrive but whenever you try to read the data (and derefence the pointer) you will access either a not mapped or mapped without read access part of memory which will cause the application to segfault.

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
```diff
@@ -0,0 +1,97 @@
...
+    sudo groupadd -f iceoryx
+
+    # Create users w/o homedir
+    sudo useradd -M perception
+    sudo useradd -M infotainment
+    sudo useradd -M roudi
+    sudo useradd -M notallowed
+
+    # Assign users to group and disable login
+    sudo usermod -L perception -a -G privileged
```
elfenpiff, line 59: You do not need to write usermod here you can pack everything in one command which makes this part less error prone. Just write:
```sh
usermod -M -G iceoryx,unprivileged -s /sbin/nologin infotainment
```

Some more infos:
1. every user is per default locked in a unix system. If you would like to unlock it you have to set a password with passwd which sets the field in /etc/shadow and unlocks the user
2. to deny the user to be used other then to run a specific application (deny login) we use `-s /sbin/nologin`. Otherwise you could always do stuff like: `su - infotainment` even when the user is locked, which we do not want.
3. After `-G` you can add a comma separated list of groups.

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
```diff
@@ -0,0 +1,97 @@
...
+    $tmux kill-server
+
+    $tmux has-session -t $SESSION
+    if [ $? -eq 0 ]; then
+        echo "Session $SESSION already exists. Attaching to session."
+        $tmux attach -t $SESSION
+        exit 0;
+    fi
+
+    command -v tmux >/dev/null 2>&1 || { echo >&2 "tmux is not installed but required. Trying to install it..."; sudo apt-get install tmux; }
```
elfenpiff, line 80: shouldn't we use $tmux here?

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
```diff
@@ -0,0 +1,97 @@
...
+
+# ***NOTE***
+# This shell script is for Linux-based operating systems only.
+# If you're using e.g QNX refer to the manual on how to set up groups, users and permissions
+
+WORKSPACE=$(git rev-parse --show-toplevel)
+CONFIG="OFF"
+RUN="OFF"
+SESSION=icecubetray
+tmux="tmux -2 -q"
```
elfenpiff, line 27: Shouldn't be every shell variable in capital letters?

iceoryx_examples/icecubetray/iox_cheeky_app.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+    // initialize runtime
+    iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+
+    // When starting this app with the user 'notallowed'
+
+    // 1) Subscribers can be created without any readable shared memory segment, in this case no data will ever arrive
+    iox::popo::Subscriber<RadarObject> subscriber({"Radar", "FrontLeft", "Object"});
+
+    // 2) The publisher object can't be initalised correctly because 'notallowed' does not have write access
```
elfenpiff, line 35: Please describe what happens here. How does the user know that it couldn't be initialized correctly as far as I know the publisher object is always in a valid and initialized state.

iceoryx_examples/icecubetray/roudi_main_static_segements.cpp
```diff
@@ -0,0 +1,50 @@
...
+    iox::RouDiConfig_t roudiConfig;
+
+    // Create Mempool Config
+    iox::mepoo::MePooConfig mepooConfig;
+
+    // We only send very small data, just one mempool per segment
+    mepooConfig.addMemPool({128, 1000});
+
+    /// Create an entry for a new shared memory segment from the mempooConfig and add it to the roudiConfig
+    roudiConfig.m_sharedMemorySegments.push_back({"unprivileged", "privileged", mepooConfig});
```
elfenpiff, line 44: Please mention via comment that the first argument is the reader group and the second is the writer group.

iceoryx_examples/icecubetray/topic_data.hpp
```diff
@@ -0,0 +1,37 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+#ifndef IOX_EXAMPLES_ICECUBETRAY_TOPIC_DATA_HPP
+#define IOX_EXAMPLES_ICECUBETRAY_TOPIC_DATA_HPP
+
+#include <cstdint>
+
+struct RadarObject
+{
+    RadarObject() noexcept
```
elfenpiff, line 23: Why do we need the default ctor? I think it should be removed.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
```diff
@@ -78,21 +79,21 @@ class SegmentManager
 
     struct SegmentUserInformation
     {
-        MemoryManager* m_memoryManager;
+        cxx::optional<MemoryManager*> m_memoryManager;
```
elfenpiff, line 82: I completely agree with your intention here but this solution causes even more problems. Since now you have two possible ways to invalidate memory manager:
1. cxx::nullopt
2. valid optional which contains a nullptr.

I recommand to use here a `cxx::optional<std::reference_wrapper<MemoryManager>>` since a reference can never be null and we can be sure that it always contains a valid value. Otherwise we have to double check everything with this update: one check for optional contains a value and one check for optional value is not nullptr.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
```diff
@@ -78,21 +79,21 @@ class SegmentManager
...
         uint64_t m_segmentID;
     };
 
     using SegmentMappingContainer = cxx::vector<SegmentMapping, MAX_SHM_SEGMENTS>;
 
-    SegmentMappingContainer getSegmentMappings(posix::PosixUser f_user);
-    SegmentUserInformation getSegmentInformationForUser(posix::PosixUser f_user);
+    SegmentMappingContainer getSegmentMappings(posix::PosixUser user) noexcept;
```
elfenpiff, line 88: Please use as argument `const posix::PosixUser & user`

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
```diff
@@ -78,21 +79,21 @@ class SegmentManager
...
+    SegmentUserInformation getSegmentInformationWithWriteAccessForUser(posix::PosixUser user) noexcept;
```
elfenpiff, line 89: Please use as argument `const posix::PosixUser & user`

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
```diff
@@ -57,7 +58,7 @@ class SegmentManager
                        uint64_t size,
                        bool isWritable,
                        uint64_t segmentId,
-                       const iox::mepoo::MemoryInfo& memoryInfo = iox::mepoo::MemoryInfo())
+                       const iox::mepoo::MemoryInfo& memoryInfo = iox::mepoo::MemoryInfo()) noexcept
```
elfenpiff, line 61: Please move implementation in inl file.

iceoryx_posh/include/iceoryx_posh/internal/roudi/process_manager.hpp
```diff
@@ -146,20 +146,17 @@ class ProcessManager : public ProcessManagerInterface
...
     /// @param [in] isMonitored indicates if the process should be monitored for being alive
     /// @param [in] transmissionTimestamp is an ID for the application to check for the expected response
-    /// @param [in] dataSegmentId is an identifier for the shm data segment
     /// @param [in] sessionId is an ID generated by RouDi to prevent sending outdated IPC channel transmission
     /// @param [in] versionInfo Version of iceoryx used
     /// @return Returns if the process could be added successfully.
     bool addProcess(const RuntimeName_t& name,
                     const uint32_t pid,
-                    cxx::not_null<mepoo::MemoryManager* const> payloadDataSegmentMemoryManager,
+                    posix::PosixUser user,
```
elfenpiff, line 157: Please use `const posix::PosixUser & user`

iceoryx_posh/source/roudi/process.cpp
```diff
@@ -27,16 +27,14 @@ namespace roudi
 {
 Process::Process(const RuntimeName_t& name,
                  const uint32_t pid,
-                 mepoo::MemoryManager& payloadDataSegmentMemoryManager,
+                 posix::PosixUser user,
```
elfenpiff, line 30: Please use `const posix::PosixUser & user`

iceoryx_posh/source/roudi/process_manager.cpp
```diff
@@ -220,37 +214,22 @@ bool ProcessManager::registerProcess(const RuntimeName_t& name,
...
+            returnValue = addProcess(name, pid, user, isMonitored, transmissionTimestamp, sessionId, versionInfo);
         });
 
     return returnValue;
 }
 
 bool ProcessManager::addProcess(const RuntimeName_t& name,
                                 const uint32_t pid,
-                                cxx::not_null<mepoo::MemoryManager* const> payloadDataSegmentMemoryManager,
+                                posix::PosixUser user,
```
elfenpiff, line 230: Please use `const posix::PosixUser & user`

iceoryx_posh/source/roudi/process_manager.cpp
```diff
@@ -516,8 +495,22 @@ void ProcessManager::addPublisherForProcess(const RuntimeName_t& name,
     searchForProcessAndThen(
         name,
         [&](Process& process) { // create a PublisherPort
+            auto segmentInfo = m_segmentManager->getSegmentInformationWithWriteAccessForUser(process.getUser());
```
elBoberido, line 498: Have you thought about having a `process.getSegmentInfoWithWriteAccess` instead of `process.getUser` followed with the call to get the information? This has the advantage that the segment info must not be queried again for each publisher of the process.

elBoberido (commented):
I didn't do a thorough review, just looked at the code for obtaining the segment info

iceoryx_examples/icecubetray/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
mossmaurice: No hard feelings on the name. But I'll only change it if we come up with a cooler name :P

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
+# icecubetray
+
+## Introduction
+
+This example demonstrates how access rights can be applied to shared memory segments on Linux-based operating systems.
```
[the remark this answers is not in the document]
mossmaurice: That's not correct. Although, the feature works on QNX as mentioned further down. Setting the capabilities on QNX is [different](http://www.qnx.com/developers/docs/7.1/index.html#com.qnx.doc.neutrino.lib_ref/topic/p/procmgr_ability.html) than on [Linux](https://man7.org/linux/man-pages/man7/capabilities.7.html). Hence, the script only works on Linux systems now. We can adapt that to work on QNX for v2.0.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+!!! hint
+    The access right feature is only supported on Linux and QNX
+
+## Expected output
+
+<!-- Add asciinema link here -->
+
+## Code walkthrough
+
+The user _roudi_ does not need root access rights. However, it needs _CAP\_KILL_ capability or similar rights on
```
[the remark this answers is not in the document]
mossmaurice: Please try to finish the paragraph before writing a comment.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+other POSIX operating system. RouDi needs to be able to send a _SIGKILL_ signal to the apps in case RouDi is shutdown.
```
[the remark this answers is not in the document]
mossmaurice: I don't like long sentences. Swapped them to make it clearer

iceoryx_examples/icecubetray/README.md
mossmaurice: Corrected.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+```cpp
+iox::RouDiConfig_t roudiConfig;
+
+// Create Mempool Config
+iox::mepoo::MePooConfig mepooConfig;
+
+// We only send very small data, just one mempool per segment
+mepooConfig.addMemPool({128, 1000});
+
+/// Create an Entry for a new Shared Memory Segment from the MempoolConfig and add it to the RouDiConfig
```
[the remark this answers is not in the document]
mossmaurice: This is mentioned above

iceoryx_examples/icecubetray/README.md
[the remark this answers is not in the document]
elfenpiff: I do not understand? Which script? I thought we have a C++ example here?

iceoryx_examples/icecubetray/README.md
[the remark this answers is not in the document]
elfenpiff: I would still add this as a code comment since most people will read the code first and will turn to the docu for further questions.

iceoryx_examples/icecubetray/CMakeLists.txt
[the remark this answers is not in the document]
elfenpiff: Since we already have the attitude quality over quantity I would also extend this for quality over coolness.
I know it's hard! 
But if we confuse a new community member with cool misleading names no one is helped. We want to make it as easy as possible for everyone so that they can focus on their cool stuff as fast as possible!

iceoryx_examples/icecubetray/README.md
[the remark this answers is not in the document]
mossmaurice: I'm talking about the [shell script](https://github.com/eclipse-iceoryx/iceoryx/pull/702/files#r610736394), which includes command that can't be run on QNX.
We should not mention something here, which is not tested! I follow the approach, if it is not tested, it's broken. This is why I've opted-out QNX, saying in theory it should work, but it is not tested.

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
marthtz: Does RouDi need this capability in general and all the time or just the special case of `iox-cpp-roudi-static-segments` or is just related to this example?
Haven't reviewed the full PR yet.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+| Users        | privileged group | unprivileged group | infotainment group |   iceoryx group    |
+|--------------|:----------------:|:------------------:|:------------------:|:------------------:|
+| perception   |        X         |                    |                    |         X          |
+| infotainment |                  |         X          |         X          |         X          |
+| roudi        |                  |                    |                    |         X          |
+| notallowed   |                  |                    |                    |         X          |
+
+### Overview over the Apps and Shared Memory Segments
+
+```
```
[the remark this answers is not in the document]
mossmaurice: You're right, I've adapted the picture.

> If you do not have write access and write into such a segment you will encounter a segmentation fault (for instance when Display App would like to write into Privileged Shared).

That's the behaviour that we want avoid in future. Segfaults shall never be the expected behaviour in reliable software. If no write access is given and the user creates a publisher object, the users gets an error message and program execution stops. In case of a subscriber without read access there will be just not data arriving. The subscriber behaviour will be implemented on a separate PR for v2.0

iceoryx_examples/icecubetray/README.md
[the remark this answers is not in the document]
mossmaurice: True, I've swapped code and explanation.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+mepooConfig.addMemPool({128, 1000});
+
+/// Create an Entry for a new Shared Memory Segment from the MempoolConfig and add it to the RouDiConfig
+roudiConfig.m_sharedMemorySegments.push_back({"unprivileged", "privileged", mepooConfig});
+roudiConfig.m_sharedMemorySegments.push_back({"infotainment", "infotainment", mepooConfig});
+```
+
+The radar app is started with the user _perception_ and is sending data into the _privileged_ shared memory segment.
+
+The display app is started with the user _infotainment_. It reads the topic `{"Radar", "FrontLeft", "Object"}` from the privileged segement and forwards it as a slighty modified topic `{"Radar", "HMI-Display", "Object"}`. Because the user _infotainment_ only has write access to the infotainment segment, the data is written to this segment.
```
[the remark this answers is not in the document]
mossmaurice: Added

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+!!! hint
+    It's advised to create per writer group only one shared memory segement (e.g. not two segements with `w: infotainment`).
+    In this case it wouldn't be possible to control which segment will be used.
```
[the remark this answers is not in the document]
mossmaurice: Currently, there is no possibility to tell a publisher "You go write to segment A" and another one "you go write to segment B". Once a writable segment is found in the list [only this segment](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.inl#L135-L139) is mapped to the application address space.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The shared memory segments can be found under `/dev/shm`
```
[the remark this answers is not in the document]
mossmaurice: As described in the beginning of the readme, this document is only for Linux-based operating systems. We can extend for QNX with the v2.0 release.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+roudiConfig.m_sharedMemorySegments.push_back({"unprivileged", "privileged", mepooConfig});
+roudiConfig.m_sharedMemorySegments.push_back({"infotainment", "infotainment", mepooConfig});
+```
+
+The radar app is started with the user _perception_ and is sending data into the _privileged_ shared memory segment.
+
+The display app is started with the user _infotainment_. It reads the topic `{"Radar", "FrontLeft", "Object"}` from the privileged segement and forwards it as a slighty modified topic `{"Radar", "HMI-Display", "Object"}`. Because the user _infotainment_ only has write access to the infotainment segment, the data is written to this segment.
+
+!!! hint
+    It's advised to create per writer group only one shared memory segement (e.g. not two segements with `w: infotainment`).
```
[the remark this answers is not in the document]
mossmaurice: Having two segments with the same writerGroup is not supported. See [this](https://github.com/eclipse-iceoryx/iceoryx/blob/abe95adc35746743351645a2864649c7eea392cd/iceoryx_posh/test/moduletests/test_mepoo_segment_management.cpp#L175) test.

iceoryx_examples/icecubetray/README.md
```diff
@@ -0,0 +1,127 @@
...
+    In this case it wouldn't be possible to control which segment will be used.
+
+The shared memory segments can be found under `/dev/shm`
+
+```
+moss@reynholm:~$ ls -al /dev/shm
```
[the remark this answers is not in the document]
mossmaurice: Replaced

iceoryx_examples/icecubetray/README.md
mossmaurice: Clarified

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
mossmaurice: No, this is a general requirement that was introduced with your pull request #333

iceoryx_examples/icecubetray/README.md
mossmaurice: That's not correct. If no write access right is found and a publisher is created the error handler is called with [`POSH__RUNTIME_NO_WRITABLE_SHM_SEGMENT`](https://github.com/eclipse-iceoryx/iceoryx/blob/abe95adc35746743351645a2864649c7eea392cd/iceoryx_posh/source/runtime/posh_runtime.cpp#L217-L221). In case a subscriber is created without read access, it currently segfaults. This will be fixed with a follow-up PR for v2.0 as this requires changes in the discovery loop.

iceoryx_examples/icecubetray/README.md
mossmaurice: See https://github.com/eclipse-iceoryx/iceoryx/pull/702#discussion_r612443112

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
elBoberido: To be more precise, it is needed so that RouDi can shutdown all registered apps if they belong to a different user-id than RouDi itself. So it's a general requirement but if it is missing it won't have any impact if everything is run with the same user-id.

iceoryx_examples/icecubetray/config_and_run_icecubetray.sh
mossmaurice: Thanks for the explanation! Added

iceoryx_examples/icecubetray/topic_data.hpp
mossmaurice: `iox::popo::Publisher::loan()` requires a default c'tor as it is default constructing it. The `RadarObject` should be consistent throughout all examples.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
mossmaurice: Will do this in the follow-up as I would like to do some more refactoring.

iceoryx_posh/source/roudi/process_manager.cpp
mossmaurice: This would be a possible optimisation. For me in this context it feels more natural to associate a `roudi::Process` with a `posix::User` instead of associating it with a `SegmentUserInformation`. Furthermore, I think the performance difference is neglectable. I would follow our normal mantra here: Make it work, make it beautiful, make it fast. Something for the next PR.

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
mossmaurice: Good point, done.

iceoryx_examples/icecubetray/CMakeLists.txt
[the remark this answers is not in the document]
mossmaurice: I agree, for sure we can be cool & provide quality at the same time.

I guess we need a vote on the name :smiley: 

* iceaccessrights
* icerights
* iceaccess
* icecubetray
* iceacl
* icecontrolaccess
* icecreamaccess
* controlaccesstoicecream

@budrus @elBoberido @marthtz @ithier Any other ideas? What do you think?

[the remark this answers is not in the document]
elfenpiff: how about ice_access_control? This is much more readable! (variation of the third from the bottom)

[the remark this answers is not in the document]
elBoberido: @mossmaurice you didn't mark your suggestions with emojis so I had to choose the one from @elfenpiff which isn't ambiguous with any emoji ;)

dkroenke (approved):

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.hpp
elfenpiff: To satisfy this checkbox: All open points are addressed and tracked via issues. Could you please link the issue where this open point is tracked.

mossmaurice: Already tracked in #722

elfenpiff (approved):

PR #716, opened 2021-04-13
Iox #482 add example for complex data types

### LINKED ISSUE #482: Extend documentation for 1.0 release and website

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
This PR introduces an example for sending/receiving our cxx constructs. The README.md includes only the outline and will be finished once the example code is approved.

## References

- Partially closes #743

### DIFF
--- a/iceoryx_examples/complexdata/README.md
+++ b/iceoryx_examples/complexdata/README.md
@@ -8,10 +8,9 @@ contained in the shared memory and must not internally use pointers or reference
 reimplemented some [constructs](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#cxx). This example shows how
 to send/receive a iox::cxx::vector and how to send/receive a complex data structure containing some of our STL container surrogates.
 
-<!--## Expected Output-->
+## Expected Output
 
-<!-- @todo Add expected output with asciinema recording before v2.0-->
-<!-- @todo multiple examples described in here, expected output should be in front of every example -->
+[![asciicast](https://asciinema.org/a/410662.svg)](https://asciinema.org/a/410662)
 
 ## Code Walkthrough
 
--- a/iceoryx_examples/complexdata/README.md
+++ b/iceoryx_examples/complexdata/README.md
@@ -23,6 +23,7 @@ our requirements.
 In this example we want our publisher to send a vector containing double. Since we cannot use dynamic memory, we use the 
 `iox::cxx::vector` with a capacity of 5.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_vector.cpp][create publisher]-->
 ```cpp
 iox::popo::Publisher<iox::cxx::vector<double, 5>> publisher({"Radar", "FrontRight", "VectorData"});
 ```
@@ -31,9 +32,12 @@ We use a while-loop similar to the one described in the
 [icedelivery example](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_examples/icedelivery) to send the
 vector to the subscriber. After successfully loaning memory we append elements to the vector until it's full.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_vector.cpp][vector emplace_back]-->
 ```cpp
 for (uint64_t i = 0U; i < sample->capacity(); ++i)
 {
+    // we can omit the check of the return value since the loop doesn't exceed the capacity of the
+    // vector
     sample->emplace_back(static_cast<double>(ct + i));
 }
 ```
@@ -47,12 +51,13 @@ since the for-loop doesn't exceed the capacity of the vector.
 Our subscriber application iterates over the received vector to print its entries to the console. Note that the `separator` is only
 used for a easy to read output.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_subscriber_vector.cpp][vector output]-->
 ```cpp
 for (const auto& entry : *sample)
 {
     s << separator << entry;
+    separator = ", ";
 }
-std::cout << s.str();
 ```
 
 ### Publisher application sending a complex data structure
@@ -61,6 +66,7 @@ In this example our publisher will send a more complex data structure. It contai
 in iceoryx. A list of all reimplemented containers can be found
 [here](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#cxx).
 
+<!--[geoffrey][iceoryx_examples/complexdata/topic_data.hpp][complexdata type]-->
 ```cpp
 struct ComplexDataType
 {
@@ -80,6 +86,7 @@ We use again a while-loop to loan memory, add data to our containers and send it
 all used insertion methods return a bool that indicates whether the insertion was successful. It will fail when a container is already
 full. To handle the return value we introduce a helper function.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_complexdata.cpp][handle return val]-->
 ```cpp
 void handleInsertionReturnVal(const bool success)
 {
@@ -94,6 +101,7 @@ void handleInsertionReturnVal(const bool success)
 Now let's add some data to our containers. For the lists we use the `push_front` methods which can be used similar to the 
 corresponding STL methods.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_complexdata.cpp][fill lists]-->
 ```cpp
 // forward_list<string<10>, 5>
 handleInsertionReturnVal(sample->stringForwardList.push_front("world"));
@@ -113,8 +121,8 @@ handleInsertionReturnVal(sample->optionalList.push_front(nullopt));
 
 Now we fill the stack
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_complexdata.cpp][fill stack]-->
 ```cpp
-// stack<float, 5>
 for (uint64_t i = 0U; i < sample->floatStack.capacity(); ++i)
 {
     handleInsertionReturnVal(sample->floatStack.push(static_cast<float>(ct * i)));
@@ -123,15 +131,15 @@ for (uint64_t i = 0U; i < sample->floatStack.capacity(); ++i)
 
 and assign a greeting to the string.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_complexdata.cpp][assign string]-->
 ```cpp
-// string<20>
 sample->someString = "hello iceoryx";
 ```
 
 For the vectors we use the `emplace_back` method, which can be used similar to corresponding `std::vector` method.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_publisher_complexdata.cpp][fill vectors]-->
 ```cpp
-// vector<double, 5>
 for (uint64_t i = 0U; i < sample->doubleVector.capacity(); ++i)
 {
     handleInsertionReturnVal(sample->doubleVector.emplace_back(static_cast<double>(ct + i)));
@@ -149,21 +157,25 @@ With `in_place_index` the passed object is constructed in-place at the given ind
 The subscriber application just prints the received data to the console. For the `optionalList` we have to check whether the
 `optional` contains a value. As in the first example, the `separator` is used for a clear output.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp][read optional list]-->
 ```cpp
 for (const auto& entry : sample->optionalList)
 {
     (entry.has_value()) ? s << separator << entry.value() : s << separator << "optional is empty";
+    separator = ", ";
 }
 ```
 
 To print the elements of the `floatStack`, we pop elements until the stack is empty.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp][read stack]-->
 ```cpp
 auto stackCopy = sample->floatStack;
 while (stackCopy.size() > 0U)
 {
     auto result = stackCopy.pop();
     s << separator << result.value();
+    separator = ", ";
 }
 ```
 
@@ -175,6 +187,7 @@ To print the elements of the `variantVector` we iterate over the vector entries 
 variant via its index. We use the not STL compliant `get_at_index` method which returns a pointer to the type stored at the
 index. If the variant does not contain any type, `index()` will return an `INVALID_VARIANT_INDEX`.
 
+<!--[geoffrey][iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp][read variant vector]-->
 ```cpp
 for (const auto& i : sample->variantVector)
 {
@@ -182,20 +195,17 @@ for (const auto& i : sample->variantVector)
     {
     case 0:
         s << separator << *i.template get_at_index<0>();
-        separator = ", ";
         break;
     case 1:
         s << separator << *i.template get_at_index<1>();
-        separator = ", ";
         break;
     case INVALID_VARIANT_INDEX:
         s << separator << "variant does not contain a type";
-        separator = ", ";
         break;
     default:
         s << separator << "this is a new type";
-        separator = ", ";
     }
+    separator = ", ";
 }
 ```
 
--- a/iceoryx_examples/complexdata/README.md
+++ b/iceoryx_examples/complexdata/README.md
@@ -77,7 +77,7 @@ struct ComplexDataType
 Contrary to the STL containers, the iceoryx containers have a static size, i.e. you have to provide the capacity (= max. size).
 
 We use again a while-loop to loan memory, add data to our containers and send it to the subscriber. Since we must not throw exceptions
-all used inerstion methods return a bool that indicates whether the insertion was successful. It will fail when a container is already
+all used insertion methods return a bool that indicates whether the insertion was successful. It will fail when a container is already
 full. To handle the return value we introduce a helper function.
 
 ```cpp
@@ -156,7 +156,7 @@ for (const auto& entry : sample->optionalList)
 }
 ```
 
-To print the elements of the `floatStack`, we pop elements until it's empty.
+To print the elements of the `floatStack`, we pop elements until the stack is empty.
 
 ```cpp
 auto stackCopy = sample->floatStack;
@@ -173,7 +173,7 @@ when the stack contains some.
 
 To print the elements of the `variantVector` we iterate over the vector entries and access the alternative that is held by the
 variant via its index. We use the not STL compliant `get_at_index` method which returns a pointer to the type stored at the
-index. If the variant does not contain any type `index()` will return an `INVALID_VARIANT_INDEX`.
+index. If the variant does not contain any type, `index()` will return an `INVALID_VARIANT_INDEX`.
 
 ```cpp
 for (const auto& i : sample->variantVector)
--- a/iceoryx_examples/complexdata/README.md
+++ b/iceoryx_examples/complexdata/README.md
@@ -2,24 +2,62 @@
 
 ## Introduction
 
-- to implement zero-copy data transfer we use a shared memory approach
-- this requires that every data structure needs to be stored in the shared memory and that
-  they do not use the heap (link to restrictions chapter in overview.md)
-- most of the STL types cannot be used, but we reimplemented some (link to utils readme/cxx)
-- this example shows a) how to send/receive a iox::cxx::vector b) how to send/receive a
-  complex data structure containing some of our STL container surrogates
+To implement zero-copy data transfer we use a shared memory approach. This requires that every data structure needs to be entirely
+contained in the shared memory and must not internally use pointers or references. The complete list of restrictions can be found
+[here](https://iceoryx.io/latest/getting-started/overview/#restrictions). Therefore, most of the STL types cannot be used, but we
+reimplemented some [constructs](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_utils#cxx). This example shows how
+to send/receive a iox::cxx::vector and how to send/receive a complex data structure containing some of our STL container surrogates.
 
-## Expected Output
+<!--## Expected Output-->
 
-<!-- @todo Add expected output with asciinema recording before v1.0-->
+<!-- @todo Add expected output with asciinema recording before v2.0-->
 <!-- @todo multiple examples described in here, expected output should be in front of every example -->
 
 ## Code Walkthrough
 
-### Publisher application sending a iox::cxx::vector
+introduction
 
-### Subscriber application receiving a iox::cxx::vector
+### Publisher application sending a `iox::cxx::vector`
+
+In this example we want our publisher to send a vector containing double. Since we cannot use dynamic memory, we use the 
+`iox::cxx::vector` with a capacity of 5.
+
+```cpp
+iox::popo::Publisher<iox::cxx::vector<double, 5>> publisher({"Radar", "FrontRight", "VectorData"});
+```
+
+We use a while-loop similar to the one described in the 
+[icedelivery example](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_examples/icedelivery) to send the
+vector to the subscriber. After successfully loaning memory we append elements to the vector until it's full.
+
+```cpp
+for (uint64_t i = 0U; i < sample->capacity(); ++i)
+{
+    sample->emplace_back(static_cast<double>(ct + i));
+}
+```
+
+The only difference here to the `std::vector` is that `emplace_back` returns a bool - true if the appending was successful,
+false otherwise. `emplace_back` fails when the vector is already full. In our case, we can omit the check of the return value
+since the for-loop doesn't exceed the capacity of the vector.
+
+### Subscriber application receiving a `iox::cxx::vector`
+
+Our subscriber application iterates over the received vector to print its entries to the console. Note that the `separator` is only
+used for a easy to read output.
+
+```cpp
+for (const auto& entry : *sample)
+{
+    s << separator << entry;
+}
+std::cout << s.str();
+```
 
 ### Publisher application sending a complex data structure
 
 ### Subscriber application receiving a complex data structure
+
+<center>
+[Check out complexdata on GitHub :fontawesome-brands-github:](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_examples/complexdata){ .md-button }
+</center>
--- a/iceoryx_examples/complexdata/README.md
+++ b/iceoryx_examples/complexdata/README.md
@@ -15,7 +15,8 @@ to send/receive a iox::cxx::vector and how to send/receive a complex data struct
 
 ## Code Walkthrough
 
-introduction
+The following examples demonstrate how to send/receive the STL containers that were reimplemented in iceoryx so that they meet
+our requirements.
 
 ### Publisher application sending a `iox::cxx::vector`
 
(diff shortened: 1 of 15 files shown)


iceoryx_examples/complexdata/iox_publisher_vector.cpp
```diff
@@ -0,0 +1,72 @@
...
+    while (!killswitch)
+    {
+        publisher.loan()
+            .and_then([&](auto& sample) {
+                for (uint64_t i = 0U; i < sample->capacity(); ++i)
+                {
+                    bool success = sample->emplace_back(static_cast<double>(ct + i));
+                    if (!success)
+                    {
+                        std::cerr << "Failed to insert element." << std::endl;
```
elfenpiff, line 55: If this happens I think we should `break;` after the error message and send the stuff that was emplaced

elfenpiff (approved):

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "topic_data.hpp"
+
+#include "iceoryx_posh/popo/publisher.hpp"
+#include "iceoryx_posh/runtime/posh_runtime.hpp"
+#include "iceoryx_utils/posix_wrapper/signal_handler.hpp"
+
+bool killswitch = false;
```
elBoberido, line 23: please make this atomic

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+// push_front (list), push (stack), emplace_back (vector) return a bool - true if the insertion succeeded, false
+// otherwise
+// to keep the example clear, this helper function handles the return values
+void handle(const bool hasSuccess)
+{
+    if (!hasSuccess)
+    {
+        std::cerr << "Failed to insert element." << std::endl;
+        std::terminate();
```
elBoberido, line 40: please use `std::exit(EXIT_FAILURE)`

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    // register sigHandler
+    auto signalIntGuard = iox::posix::registerSignalHandler(iox::posix::Signal::INT, sigHandler);
+    auto signalTermGuard = iox::posix::registerSignalHandler(iox::posix::Signal::TERM, sigHandler);
+
+    // initialize runtime
+    iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+
+    // initialize publisher
+    iox::popo::Publisher<ComplexDataType> publisher({"Radar", "FrontLeft", "Object"});
```
elBoberido, line 54: ouch, you stirred up a hornet's nest. This is wrong and should also be checked in the other examples. icedelivery and icehello already have the same ServiceDescription for topics with a different data type.

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    uint64_t ct = 0;
+    // run until interrupted by Ctrl-C
+    while (!killswitch)
+    {
+        ++ct;
+        publisher.loan()
+            .and_then([&](auto& sample) {
+                // forward_list<string<10>, 5>
+                handle(sample->stringForwardList.push_front("world"));
```
elBoberido, line 64: `handle` is quite generic. Could you maybe change this to something more descriptive?

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+                for (uint64_t i = 0U; i < sample->doubleVector.capacity(); ++i)
+                {
+                    handle(sample->doubleVector.emplace_back(static_cast<double>(ct + i)));
+                }
+                // vector<variant<string<10>, double>, 10>;
+                handle(sample->variantVector.emplace_back(in_place_index<0>(), "seven"));
+                handle(sample->variantVector.emplace_back(in_place_index<1>(), 8.0));
+                handle(sample->variantVector.emplace_back(in_place_index<0>(), "nine"));
+
+                sample.publish();
```
elBoberido, line 90: a hypnotoad or brain slug is missing

iceoryx_examples/complexdata/iox_publisher_vector.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    // register sigHandler
+    auto signalIntGuard = iox::posix::registerSignalHandler(iox::posix::Signal::INT, sigHandler);
+    auto signalTermGuard = iox::posix::registerSignalHandler(iox::posix::Signal::TERM, sigHandler);
+
+    // initialize runtime
+    iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+
+    // initialize publisher
+    iox::popo::Publisher<iox::cxx::vector<double, 5>> publisher({"Radar", "FrontLeft", "Object"});
```
elBoberido, line 42: this will result in pink elefants

iceoryx_examples/complexdata/iox_publisher_vector.cpp
elBoberido: in this case I would omit the check since the loop doesn't exceed the capacity of the vector and this is unnecessary noise.

But we might think about having
```c++
cxx::vector<uint8_t, 42> foo;

uint8_t counter = 0;
foo.fill_to_capacity([&] {
    return counter++;
});

foo.for_each([](auto& element){
    std::cout << element << std::endl;
});
```

iceoryx_examples/complexdata/iox_publisher_vector.cpp
```diff
@@ -0,0 +1,72 @@
...
+                    {
+                        std::cerr << "Failed to insert element." << std::endl;
+                    }
+                }
+
+                sample.publish();
+            })
+            .or_else([](auto& error) {
+                // do something with error
+                std::cerr << "Unable to loan sample, error code: " << static_cast<uint64_t>(error) << std::endl;
```
elBoberido, line 63: this reminds me that we need to implement the stream operator for `AllocationError` and `ChunkReceiveResult`

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,119 @@
...
+        subscriber.take()
+            .and_then([](auto& sample) {
+                std::stringstream s;
+                s << APP_NAME << " got values:";
+
+                s << std::endl << "stringForwardList: ";
+                for (auto i = sample->stringForwardList.begin(); i != sample->stringForwardList.end(); ++i)
+                {
+                    s << *i << " ";
+                }
```
elBoberido, line 57: just a small hint, if you omit the white-space after `stringForwardList: ` and make `s << " " << *i;` there is no superfluous white-space at the end of the line :)

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,119 @@
...
+                s << std::endl << "integerList: ";
+                for (auto i = sample->integerList.begin(); i != sample->integerList.end(); ++i)
+                {
+                    s << *i << " ";
+                }
+
+                s << std::endl << "optionalList: ";
+                for (auto i = sample->optionalList.begin(); i != sample->optionalList.end(); ++i)
+                {
+                    (i->has_value()) ? s << i->value() << " " : s << "optional is empty, ";
```
elBoberido, line 68: `, ` is not necessary

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,119 @@
...
+                }
+
+                s << std::endl << "floatStack: ";
+                auto stackCopy = sample->floatStack;
+                for (uint64_t i = stackCopy.capacity(); i > 0U; i--)
+                {
+                    auto result = stackCopy.pop();
+                    (result.has_value()) ? s << result.value() << " " : s << "stack is empty";
```
elBoberido, line 76: the stack is not full, but you don't know if it's empty

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,119 @@
...
+                s << APP_NAME << " got values:";
+
+                s << std::endl << "stringForwardList: ";
+                for (auto i = sample->stringForwardList.begin(); i != sample->stringForwardList.end(); ++i)
+                {
+                    s << *i << " ";
+                }
+
+                s << std::endl << "integerList: ";
+                for (auto i = sample->integerList.begin(); i != sample->integerList.end(); ++i)
```
elBoberido, line 60: would it be possible to use a range-based for-loop?

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,119 @@
...
+                }
+
+                s << std::endl << "someString: ";
+                s << sample->someString;
+
+                s << std::endl << "doubleVector: ";
+                for (uint64_t i = 0U; i < sample->doubleVector.size(); ++i)
+                {
+                    s << sample->doubleVector[i] << " ";
+                }
```
elBoberido, line 86: it would be nice to have this as coma separated output ... without a superfluous comma at the end of course ;)

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,119 @@
...
+                for (uint64_t i = 0U; i < sample->variantVector.size(); ++i)
+                {
+                    if (sample->variantVector[i].index() == 0)
+                    {
+                        s << *sample->variantVector[i].template get_at_index<0>() << " ";
+                    }
+                    else if (sample->variantVector[i].index() == 1)
+                    {
+                        s << *sample->variantVector[i].template get_at_index<1>() << " ";
+                    }
```
elBoberido, line 98: please either add an `else` branch or change to a switch statement with a default case

iceoryx_examples/complexdata/README.md
```diff
@@ -0,0 +1,25 @@
+# complexdata
+
+## Introduction
+
+- to implement zero-copy data transfer we use a shared memory approach
+- this requires that every data structure needs to be stored in the shared memory and that
+  they do not use the heap (link to restrictions chapter in overview.md)
+- most of the STL types cannot be used, but we reimplemented some (link to utils readme/cxx)
+- this example shows a) how to send/receive a iox::cxx::vector b) how to send/receive a
+  complex data structure containing some of our STL container surrogates
```
MatthiasKillat, line 10: The requirement is a bit stronger.
1. Cannot use dynamic memory (I would not write heap, because it is misleading and may have other meanings).
2. Cannot use pointers internally (even if no dynamic memory is used)
3. Cannot reference external data (by index or pointer, since if you do so the external data must be present at the destination as well which adds a lot of complications)

iceoryx_examples/complexdata/topic_data.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "iceoryx_utils/cxx/forward_list.hpp"
+#include "iceoryx_utils/cxx/list.hpp"
+#include "iceoryx_utils/cxx/optional.hpp"
+#include "iceoryx_utils/cxx/stack.hpp"
+#include "iceoryx_utils/cxx/string.hpp"
+#include "iceoryx_utils/cxx/variant.hpp"
+#include "iceoryx_utils/cxx/vector.hpp"
+
+using namespace iox::cxx;
```
MatthiasKillat, line 28: ok here in this small example, but usually not a good idea in head files.

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "topic_data.hpp"
+
+#include "iceoryx_posh/popo/publisher.hpp"
+#include "iceoryx_posh/runtime/posh_runtime.hpp"
+#include "iceoryx_utils/posix_wrapper/signal_handler.hpp"
+
+std::atomic_bool killswitch{false};
```
MatthiasKillat, line 23: I would recommend using it shutdown or similar in all examples, but of course is inconsequential.

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+static void sigHandler(int f_sig [[gnu::unused]])
+{
+    // caught SIGINT or SIGTERM, now exit gracefully
+    killswitch = true;
+}
+
+// push_front (list), push (stack), emplace_back (vector) return a bool - true if the insertion succeeded, false
+// otherwise
+// to keep the example clear, this helper function handles the return values
+void handleInsertionReturnVal(const bool hasSuccess)
```
MatthiasKillat, line 35: better isSuccessful or success

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    // register sigHandler
+    auto signalIntGuard = iox::posix::registerSignalHandler(iox::posix::Signal::INT, sigHandler);
+    auto signalTermGuard = iox::posix::registerSignalHandler(iox::posix::Signal::TERM, sigHandler);
+
+    // initialize runtime
+    iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+
+    // initialize publisher
+    iox::popo::Publisher<ComplexDataType> publisher({"Group", "Instance", "Topic"});
```
MatthiasKillat, line 54: Just putting it here, but I think **Partition** might be a more fitting name than Group (more important in the description), but we need to think about this anyway concept-wise. Especially since this could be used like in DDS or network partitioning.

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+        ++ct;
+        publisher.loan()
+            .and_then([&](auto& sample) {
+                // forward_list<string<10>, 5>
+                handleInsertionReturnVal(sample->stringForwardList.push_front("world"));
+                handleInsertionReturnVal(sample->stringForwardList.push_front("hello"));
+                // list<uint64_t, 10>;
+                handleInsertionReturnVal(sample->integerList.push_front(ct));
+                handleInsertionReturnVal(sample->integerList.push_front(ct * 2));
+                handleInsertionReturnVal(sample->integerList.push_front(ct + 4));
```
MatthiasKillat, line 69: Would remove one of those inserts, does not test or demonstrate anything new. (i.e. is just noise)

iceoryx_examples/complexdata/iox_publisher_vector.cpp
MatthiasKillat: This is one of the cases where we even know that the emplacement cannot fail anyway (barring a a wrong implementation of vector or loan which does not construct a sample with an empty vector, of course). So we should avoid unnecessary checks. 
Also take a look at https://en.cppreference.com/w/cpp/algorithm/iota for similar purposes (since we should even have random access iterators here it would apply). But it would have to operate on a non empty range, which we will not have initially here.

iceoryx_examples/complexdata/iox_publisher_vector.cpp
```diff
@@ -0,0 +1,73 @@
...
+            });
+        ++ct;
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
MatthiasKillat, line 68: I would decrease the frequency somewhat (will flood the console too much).

iceoryx_examples/complexdata/topic_data.hpp
FerdinandSpitzschnueffler: I decided to use the namespace for better readability, especially in the (not yet finished) readme.md

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
FerdinandSpitzschnueffler: To be consistent with most of the other examples, I would keep the name.

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
FerdinandSpitzschnueffler: For now, I would prefer to be in sync with our overview.md where it is `Group`. But yes, we should think about a better naming, although I'm not sure if we should use the DDS naming.

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
FerdinandSpitzschnueffler: Well, I didn't want to add more extra code only used for the output which is not necessary for understanding how to send complex data. But no hard feelings about that. I can add it.

FerdinandSpitzschnueffler: Oh, maybe I got you wrong. Can you explain in more detail what you mean with `coma separated`? ;)

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
elBoberido: could you be a bit more specific with the ServiceDescription, e.g. using `ComplexDataTopic` as event id to prevent accidental matching ServiceDescription with different data

iceoryx_examples/complexdata/iox_publisher_vector.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+    // register sigHandler
+    auto signalIntGuard = iox::posix::registerSignalHandler(iox::posix::Signal::INT, sigHandler);
+    auto signalTermGuard = iox::posix::registerSignalHandler(iox::posix::Signal::TERM, sigHandler);
+
+    // initialize runtime
+    iox::runtime::PoshRuntime::initRuntime(APP_NAME);
+
+    // initialize publisher
+    iox::popo::Publisher<iox::cxx::vector<double, 5>> publisher({"Radar", "FrontRight", "Object"});
```
elBoberido, line 42: This is still quite near to our examples with a different data type and my inner chicken feels that this might become a problem. Could you please change this even more, like having `VectorData` as event id?

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
elBoberido: no, you got it right. I meant a comma separated list, but it seems you already did it in an elegant way ... I guess I was in a coma when I wrote that :)

iceoryx_examples/complexdata/iox_publisher_vector.cpp
elfenpiff: @elBoberido we already have and `.for_each` at our hands.
```cpp
std::for_each(foo.begin(), foo.end(), [](auto e){std::cout << e << std::endl;});
```

Maybe we need a cxx version for it so that it is possible to do this for the whole range whenever it has `.end` and `.begin` as methods like:
```cpp
cxx::for_each(foo, [](auto e){std::cout << e << std::endl;});
```
This would be more like the C++ approach advertised in the STL.


The second problem I would solve like this:
```cpp
cxx::vector<uint8_t, 42> foo(42);
std::generate(foo.begin(), foo.end(), [counter=0]() mutable { return counter++; });
```
Maybe we also need a cxx version of that which would become more readable;
```cpp
cxx::vector<uint8_t, 42> foo(42);
cxx::generate(foo, [counter=0]() mutable { return counter++; });
```

But in the case that the elements are not default constructable we would have to pursue your approach since the containers have different method names to insert another element.

FerdinandSpitzschnueffler: removed the check

elBoberido: you're example for generate won't work since `foo.begin() == foo.end()`

elfenpiff: @elBoberido, see the first line:
```cpp
cxx::vector<uint8_t, 42> foo(42); // this ctor creates 42 elements with the default ctor
```
This is a full vector and `foo.begin() != foo.end()`.

dkroenke (commented):
LGTM, just a nitpick from my side.

dkroenke (dismissed):

iceoryx_examples/complexdata/iox_publisher_vector.cpp
elBoberido: @elfenpiff in this case you call the default ctor first, just to afterwards immediately replace the values. Not very efficient ;)

iceoryx_examples/complexdata/iox_publisher_complexdata.cpp
elBoberido: I agree with Matthias. We should promote love, not killing ;)
This should be done in a separate PR, though ... maybe when geoffrey takes care of the syncing ;)

iceoryx_examples/complexdata/iox_subscriber_complexdata.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+                s << std::endl << "variantVector:";
+                separator = " ";
+                for (const auto& i : sample->variantVector)
+                {
+                    switch (i.index())
+                    {
+                    case 0:
+                        s << separator << *i.template get_at_index<0>();
+                        separator = ", ";
```
elBoberido, line 105: this could be moved out of the switch statement

iceoryx_examples/complexdata/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+Our subscriber application iterates over the received vector to print its entries to the console. Note that the `separator` is only
+used for a easy to read output.
+
+```cpp
+for (const auto& entry : *sample)
+{
+    s << separator << entry;
+}
+std::cout << s.str();
```
elBoberido, line 55: this is outdated, please update

iceoryx_examples/complexdata/README.md
```diff
@@ -0,0 +1,204 @@
...
+### Subscriber application receiving a complex data structure
+
+The subscriber application just prints the received data to the console. For the `optionalList` we have to check whether the
+`optional` contains a value. As in the first example, the `separator` is used for a clear output.
+
+```cpp
+for (const auto& entry : sample->optionalList)
+{
+    (entry.has_value()) ? s << separator << entry.value() : s << separator << "optional is empty";
+}
```
elBoberido, line 156: this is also outdated

iceoryx_examples/complexdata/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+To print the elements of the `floatStack`, we pop elements until the stack is empty.
+
+```cpp
+auto stackCopy = sample->floatStack;
+while (stackCopy.size() > 0U)
+{
+    auto result = stackCopy.pop();
+    s << separator << result.value();
+}
```
elBoberido, line 167: same here

iceoryx_examples/complexdata/README.md
FerdinandSpitzschnueffler: oh my geoffrey... ;)

elBoberido: you can already use goeffrey if you want. I use it for the user-header example and it helps a lot. There is now also a readme in the geoffrey repo which explains how to use it :)

iceoryx_integrationtest/iceoryx_integrationtest/test_complexdata_example.py
```diff
@@ -0,0 +1,99 @@
...
+        process_list[1],
+        process_list[2],
+        process_list[3],
+        roudi_process,
+        launch_testing.actions.ReadyToTest()
+    ]), {'iox-cpp-publisher-vector': process_list[0], 'iox-cpp-subscriber-vector': process_list[1],
+         'iox-cpp-publisher-complexdata': process_list[2], 'iox-cpp-subscriber-complexdata': process_list[3],
+         'roudi_process': roudi_process}
+
+# These tests will run concurrently with the dut process. After this test is done,
```
MatthiasKillat, line 74: sut? or maybe it is a dud ...

MatthiasKillat (approved):

elBoberido (approved):

iceoryx_integrationtest/iceoryx_integrationtest/test_complexdata_example.py
FerdinandSpitzschnueffler: it's the abbreviation for device under test

iceoryx_examples/complexdata/topic_data.hpp
MatthiasKillat: Sure, as I said. Would this be a library header I would reject this argument.

MatthiasKillat (dismissed):

MatthiasKillat (approved):

dkroenke (approved):

PR #748, opened 2021-04-20
iox-#325 introduce DualAccessTransactionTray

### LINKED ISSUE #325: Deadlock in shared mem mutex when apps are killed

elBoberido:

1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## Notes for Reviewer

This is just a draft to discuss the API of the `DualAccessTransactionTray` which might replace the mutex in the `ChunkDistributor`. A better name is also welcome.

## References

- Partly closes #325

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
@@ -256,12 +256,7 @@ inline void ChunkDistributor<ChunkDistributorDataType>::clearHistory() noexcept
 template <typename ChunkDistributorDataType>
 inline void ChunkDistributor<ChunkDistributorDataType>::cleanup() noexcept
 {
-    if (getMembers()->tryLock())
-    {
-        clearHistory();
-        getMembers()->unlock();
-    }
-    else
+    if (!getMembers()->tryLock())
     {
         /// @todo currently we have a deadlock / mutex destroy vulnerability if the ThreadSafePolicy is used
         /// and a sending application dies when having the lock for sending. If the RouDi daemon wants to
@@ -271,6 +266,10 @@ inline void ChunkDistributor<ChunkDistributorDataType>::cleanup() noexcept
                      nullptr,
                      ErrorLevel::FATAL);
     }
+
+    getMembers()->unlock();
+    // the mutex is locked in this method
+    clearHistory();
 }
 
 } // namespace popo
--- a/iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
@@ -1,0 +1,79 @@
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IOX_UTILS_CONCURRENT_DUAL_ACCESS_TRANSACTION_TRAY_HPP
+#define IOX_UTILS_CONCURRENT_DUAL_ACCESS_TRANSACTION_TRAY_HPP
+
+#include "iceoryx_utils/posix_wrapper/semaphore.hpp"
+
+#include <atomic>
+
+namespace iox
+{
+namespace concurrent
+{
+class DualAccessTransactionTray
+{
+  private:
+    /// @note not public to prevent the user from passing AccessToken::NONE to a method
+    enum class AccessToken
+    {
+        NONE,
+        LEFT,
+        RIGHT
+    };
+
+  public:
+    static constexpr AccessToken LEFT{AccessToken::LEFT};
+    static constexpr AccessToken RIGHT{AccessToken::RIGHT};
+
+    class AccessGuard
+    {
+      public:
+        AccessGuard(DualAccessTransactionTray& transactionTray, AccessToken accessToken);
+        ~AccessGuard();
+
+        AccessGuard(const AccessGuard&) = delete;
+        AccessGuard(AccessGuard&&) = delete;
+
+        AccessGuard& operator=(const AccessGuard&) = delete;
+        AccessGuard& operator=(AccessGuard&&) = delete;
+
+      private:
+        DualAccessTransactionTray& m_transactionTray;
+        AccessToken m_accessToken{AccessToken::NONE};
+    };
+
+    /// @brief Clean up a potential locking when LEFT or RIGHT terminated abnormally.
+    /// @param[in] tokenToCleanup is the DualAccessTransactionTray::LEFT or DualAccessTransactionTray::RIGHT token
+    /// @attention this should only be called if the thread with `tokenToCleanup` is not running anymore else the
+    /// invariants are broken and you might observe pink elephants and dragons
+    void cleanupAndSyncMemory(AccessToken tokenToCleanup);
+
+  private:
+    void acquireExclusiveAccess(AccessToken tokenToAcquireAccess);
+    void releaseExclusiveAccess(AccessToken tokenToBeReleased);
+
+    std::atomic<AccessToken> m_accessToken{AccessToken::NONE};
+    posix::Semaphore m_waitingLineLeft{posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
+    posix::Semaphore m_waitingLineRight{
+        posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
+};
+
+} // namespace concurrent
+} // namespace iox
+
+#endif // IOX_UTILS_CONCURRENT_DUAL_ACCESS_TRANSACTION_TRAY_HPP
--- a/iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
@@ -43,7 +43,7 @@ class DualAccessTransactionTray
     class AccessGuard
     {
       public:
-        AccessGuard(DualAccessTransactionTray& transactionTray, AccessToken accessToken);
+        AccessGuard(DualAccessTransactionTray& transactionTray, const AccessToken accessToken) noexcept;
         ~AccessGuard();
 
         AccessGuard(const AccessGuard&) = delete;
@@ -61,11 +61,11 @@ class DualAccessTransactionTray
     /// @param[in] tokenToCleanup is the DualAccessTransactionTray::LEFT or DualAccessTransactionTray::RIGHT token
     /// @attention this should only be called if the thread with `tokenToCleanup` is not running anymore else the
     /// invariants are broken and you might observe pink elephants and dragons
-    void cleanupAndSyncMemory(AccessToken tokenToCleanup);
+    void cleanupAndSyncMemory(const AccessToken tokenToCleanup);
 
   private:
-    void acquireExclusiveAccess(AccessToken tokenToAcquireAccess);
-    void releaseExclusiveAccess(AccessToken tokenToBeReleased);
+    void acquireExclusiveAccess(const AccessToken tokenToAcquireAccess);
+    void releaseExclusiveAccess(const AccessToken tokenToBeReleased);
 
     std::atomic<AccessToken> m_accessToken{AccessToken::NONE};
     posix::Semaphore m_waitingLineLeft{posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
--- a/iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
+++ b/iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
@@ -57,11 +57,12 @@ class DualAccessTransactionTray
         AccessToken m_accessToken{AccessToken::NONE};
     };
 
-    /// @brief Clean up a potential locking when LEFT or RIGHT terminated abnormally.
-    /// @param[in] tokenToCleanup is the DualAccessTransactionTray::LEFT or DualAccessTransactionTray::RIGHT token
-    /// @attention this should only be called if the thread with `tokenToCleanup` is not running anymore else the
+    /// @brief Revokes the lock from an absent participant when LEFT or RIGHT terminated abnormally.
+    /// @param[in] absentPaticipantToken is the DualAccessTransactionTray::LEFT or DualAccessTransactionTray::RIGHT
+    /// token
+    /// @attention this should only be called if the thread with `absentPaticipantToken` is not running anymore else the
     /// invariants are broken and you might observe pink elephants and dragons
-    void cleanupAndSyncMemory(const AccessToken tokenToCleanup);
+    void revokeLockFromAbsentParticipant(const AccessToken absentPaticipantToken);
 
   private:
     void acquireExclusiveAccess(const AccessToken tokenToAcquireAccess);
--- a/iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
+++ b/iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
@@ -33,17 +33,14 @@ DualAccessTransactionTray::AccessGuard::~AccessGuard()
     m_transactionTray.releaseExclusiveAccess(m_accessToken);
 }
 
-void DualAccessTransactionTray::cleanupAndSyncMemory(const AccessToken tokenToCleanup)
+void DualAccessTransactionTray::revokeLockFromAbsentParticipant(const AccessToken absentPaticipantToken)
 {
-    releaseExclusiveAccess(tokenToCleanup);
-    // just to be sure the memory is synchronized
-    AccessToken expected = tokenToCleanup;
-    m_accessToken.compare_exchange_strong(expected, AccessToken::NONE, std::memory_order_acq_rel);
+    releaseExclusiveAccess(absentPaticipantToken);
 }
 
 void DualAccessTransactionTray::acquireExclusiveAccess(const AccessToken tokenToAcquireAccess)
 {
-    auto existingToken = m_accessToken.exchange(tokenToAcquireAccess, std::memory_order_acquire);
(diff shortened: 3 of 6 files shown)


iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
```diff
@@ -0,0 +1,79 @@
...
+        NONE,
+        LEFT,
+        RIGHT
+    };
+
+  public:
+    static constexpr AccessToken LEFT{AccessToken::LEFT};
+    static constexpr AccessToken RIGHT{AccessToken::RIGHT};
+
+    class AccessGuard
```
elBoberido, line 43: I modeled this after the `std::lock_guard` but I'm not really convinced about this. Currently I'm leaning more towards something like
```c++
DualAccessTransactionTray transactionTray;
transactionTray.acquireExclusiveAccess(DualAccessTransactionTray::LEFT, [&] {
    // do stuff
});
```

iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
```diff
@@ -0,0 +1,79 @@
...
+    void cleanupAndSyncMemory(AccessToken tokenToCleanup);
+
+  private:
+    void acquireExclusiveAccess(AccessToken tokenToAcquireAccess);
+    void releaseExclusiveAccess(AccessToken tokenToBeReleased);
+
+    std::atomic<AccessToken> m_accessToken{AccessToken::NONE};
+    posix::Semaphore m_waitingLineLeft{posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
+    posix::Semaphore m_waitingLineRight{
+        posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
```
elBoberido, line 73: The semaphores are tricky. Are we sure it's allowed to move them?
The mutex in the `ChunkDistributor`, which this is supposed to replace does just a `std::terminate` if creation fails.

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
```diff
@@ -0,0 +1,95 @@
...
+    AccessToken expected = tokenToCleanup;
+    m_accessToken.compare_exchange_strong(expected, AccessToken::NONE, std::memory_order_acq_rel);
+}
+
+void DualAccessTransactionTray::acquireExclusiveAccess(AccessToken tokenToAcquireAccess)
+{
+    auto existingToken = m_accessToken.exchange(tokenToAcquireAccess, std::memory_order_acquire);
+    if (existingToken == tokenToAcquireAccess)
+    {
+        // TODO return expected DOUBLE_ACQUIRE_BROKEN_INVARIANT
```
elBoberido, line 48: with the `AccessGuard` approach, this will also be cumbersome. Another reason to use a lambda instead

iceoryx_utils/test/moduletests/test_concurrent_dual_access_transaction_tray.cpp
```diff
@@ -0,0 +1,174 @@
...
+    for (uint64_t i = 0; i < NUMBER_OF_LOOPS; ++i)
+    {
+        DualAccessTransactionTray::AccessGuard guard(transactionTray, DualAccessTransactionTray::RIGHT);
+        ++counter;
+    }
+
+    auto finish = std::chrono::system_clock::now();
+    auto averageTimeInNs =
+        std::chrono::duration_cast<std::chrono::nanoseconds>(finish - now).count() / (NUMBER_OF_LOOPS * 2);
+    std::cout << "Average locking time: " << averageTimeInNs / 1000. << "µs" << std::endl;
```
elBoberido, line 62: from my measurements
|  | Mutex  | DualAccessTransactionTray|
--- | --- | ---
|no contention|16ns|25ns|
|contention|85ns|1980ns|
With contention the `DualAccessTransactionTray` is much slower, but there is essentially no contention in the `ChunkDistributor` since this can only happen when a subscriber queue is added or removed. So I would argue that the pros outweighs the cons.

iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
```diff
@@ -0,0 +1,79 @@
...
+    /// @attention this should only be called if the thread with `tokenToCleanup` is not running anymore else the
+    /// invariants are broken and you might observe pink elephants and dragons
+    void cleanupAndSyncMemory(AccessToken tokenToCleanup);
+
+  private:
+    void acquireExclusiveAccess(AccessToken tokenToAcquireAccess);
+    void releaseExclusiveAccess(AccessToken tokenToBeReleased);
+
+    std::atomic<AccessToken> m_accessToken{AccessToken::NONE};
+    posix::Semaphore m_waitingLineLeft{posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
```
elfenpiff, line 71: I think you have to move the initialization into an immediately invoked lambda so that you can handle the error case when the semaphore could not be created. Something like:
```cpp
posix::Semaphore m_waitingLineLeft{[]{ 
    std::move(posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U)\
       .or_else([](auto){std::cerr << "skynet arises, prepare for termination\n"; std::terminate(); }).value());
}()};
```

elfenpiff (commented):
4 of 6

iceoryx_utils/include/iceoryx_utils/internal/concurrent/dual_access_transaction_tray.hpp
MatthiasKillat: As we internally already discussed moving when created by sem_open is ok. When created by sem_init we cannot do this, since our `move` leads to a copy of a `sem_t` struct, which is not safe (especially when someone is waiting). Besides the behavior will be weird for the user. The underlying sem_t is most likely not allowed to be copied and the new STL semaphores are als also not movable, surely for good reason.

So no, in general the semaphore is not movable as it is implemented.

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
MatthiasKillat: The Lambda approach is superior. The access request will be blocking either way until access is obtained.

MatthiasKillat, line 48: Only one thread is supposed to have access, either via the LEFT or RIGHT token, correct? In this case it makes sense, it is like calling lock of a mutex twice without unlocking. Otherwise we could argue that e.g. LEFT access was already granted and we would return immediately without error. (I do not fully know the use case.)

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
```diff
@@ -0,0 +1,95 @@
...
+        std::terminate();
+    }
+    else if (existingToken != AccessToken::NONE)
+    {
+        if (tokenToAcquireAccess == AccessToken::LEFT)
+        {
+            if (m_waitingLineLeft.wait().has_error())
+            {
+                // TODO return expected ERROR_WHILE_WAITING_FOR_SEMAPHORE
```
MatthiasKillat, line 57: What can we do if the wait causes an error? I think the access request fails in this case and the user has to handle it. In any case the logic to wait looks ok.

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+void DualAccessTransactionTray::releaseExclusiveAccess(AccessToken tokenToBeReleased)
+{
+    AccessToken expected = tokenToBeReleased;
+    auto casSuccessful = m_accessToken.compare_exchange_strong(expected, AccessToken::NONE, std::memory_order_release);
+    if (!casSuccessful)
+    {
+        if (expected == AccessToken::NONE)
```
MatthiasKillat, line 76: The user cannot pass NONE, due to private enum. So this can only happen due to double release or releasing something not acquired before. Similar to unlocking a non-locked mutex, which is undefined.

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
```diff
@@ -0,0 +1,95 @@
...
+        {
+            // TODO return expected RELEASE_ON_NONE_BROKEN_INVARIANT
+            std::terminate();
+        }
+        else if (tokenToBeReleased == AccessToken::LEFT)
+        {
+            // post can result in either EINVAL or EOVERFLOW; neither of those can be triggered by this code
+            IOX_DISCARD_RESULT(m_waitingLineRight.post());
```
MatthiasKillat, line 84: We post even though no one might be waiting? This is dangerous. Consider a Sequence of acq LEFT, rel LEFT, acq RIGHT, rel RIGHT without someone waiting. Afterwards a LEFT and RIGHT request could arrive concurrently and both would gain access (both semaphore counters are 1) or some other problematic state I think. 

The compare exchanges may protect against this, I need to fully think this through, but the potential interleaves are non-trivial. In any case, there might be a problem but I would like to find a convincing argument why there is none (or a fix if there is).

iceoryx_utils/test/moduletests/test_concurrent_dual_access_transaction_tray.cpp
MatthiasKillat: Hard to point towards the reason. There are probably possible small improvements  wrt. semaphore and CAS usage. Also the expected may cause slowdown compared to native semaphores.

In any case, do we want this kind of output in a test? These are also not moduletests, but stresstests, i.e. belong into another target.

MatthiasKillat (commented):
I do not fully understand the use case yet. It looks somewhat like implementing two in a class mutexes using two semaphores. Now implementing mutexes with semaphores is fairly standard, and likely what is happening if a mutex would be used directly. 

What we have here is something like a tristate mutex, would you agree? (with states LEFT_LOCKED, RIGHT_LOCKED, UNLOCKED) I think it is possible to get rid of one semaphore, but I need to think about this. This will probably not help much in terms of performance but remedy one potential problem and simplify the code a little.

iceoryx_utils/test/moduletests/test_concurrent_dual_access_transaction_tray.cpp
elBoberido: this is just temporarily and will either be removed or moved to a stresstest

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -256,12 +256,7 @@ inline void ChunkDistributor<ChunkDistributorDataType>::clearHistory() noexcept
 template <typename ChunkDistributorDataType>
 inline void ChunkDistributor<ChunkDistributorDataType>::cleanup() noexcept
 {
-    if (getMembers()->tryLock())
-    {
-        clearHistory();
-        getMembers()->unlock();
-    }
-    else
+    if (!getMembers()->tryLock())
```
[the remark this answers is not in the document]
elBoberido: This is currently used to detect if the mutex is locked. It will be obsolete once the semaphore based locking will be used

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -271,6 +266,10 @@ inline void ChunkDistributor<ChunkDistributorDataType>::cleanup() noexcept
                      nullptr,
                      ErrorLevel::FATAL);
     }
+
+    getMembers()->unlock();
```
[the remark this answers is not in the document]
elBoberido: The currently used mutex has the recursive flag set and this call was the only reason. I just wanted to check if it could be removed and it is possible since this code is only executed once the application died and therefore it cannot get the lock again. The method called after `getMembers()->unlock();` will lock the mutex before the data is accessed but the locking in this method was just to detect if the application held the lock when it died.

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
elBoberido: yes, that's right

iceoryx_utils/source/concurrent/dual_access_transaction_tray.cpp
elBoberido: With the current implementation two are needed else we experience kind of an ABA problem. Let's assume we have only one semaphore, LEFT has the lock and RIGHT is waiting on the semaphore. When LEFT does a sem_post and immediately tries to get the lock, the `exchange` will tell LEFT that the lock is held by RIGHT and must therefore do a sem_wait. It can happen that RIGHT was not yet scheduled and the semaphore count is still 1, which means that LEFT will continue its execution and won't do a sem_post at the end, leaving RIGHT in a deadlock.
With an alternative implementation this problem might be solved differently.

elBoberido:
After some thinking, there is additional work necessary before this class could be used in the `ChunkDistributor`.

On the one hand, the vector with the history needs to be replaced by a ring buffer which can safely be accessed by RouDi even if the application terminated while mutating this buffer. It could be done in a similar fashion to the `UsedChunkList`.

On the other hand, there is still a deadlock in RouDi since the `ChunkDistributor` is periodically accessed from the `monitorAndDiscoveryUpdate` thread. If the application held the lock while dying and the discovery update is running before the monitoring detected the absence of the process, the discovery will be in a deadlock, waiting for the application to release the lock. Since this is done from the `ProcessManager`, which is guarded by a `smartLock`, the message processing is essentially also deadlocked. The solution for this problem would be to have a `tryLock` on the `ChunkDistributor` from the discovery update. This has the disadvantage that some subscriber might experience a delay when being subscribed/unsubscribed, but the whole system remains responsive.

@budrus @MatthiasKillat @elfenpiff what do you think?

elfenpiff:
> After some thinking, there is additional work necessary before this class could be used in the `ChunkDistributor`.
> 
> On the one hand, the vector with the history needs to be replaced by a ring buffer which can safely be accessed by RouDi even if the application terminated while mutating this buffer. It could be done in a similar fashion to the `UsedChunkList`.
> 
> On the other hand, there is still a deadlock in RouDi since the `ChunkDistributor` is periodically accessed from the `monitorAndDiscoveryUpdate` thread. If the application held the lock while dying and the discovery update is running before the monitoring detected the absence of the process, the discovery will be in a deadlock, waiting for the application to release the lock. Since this is done from the `ProcessManager`, which is guarded by a `smartLock`, the message processing is essentially also deadlocked. The solution for this problem would be to have a `tryLock` on the `ChunkDistributor` from the discovery update. This has the disadvantage that some subscriber might experience a delay when being subscribed/unsubscribed, but the whole system remains responsive.
> 
> @budrus @MatthiasKillat @elfenpiff what do you think?

My spider-sense tells me that this sounds like a refactoring where we will discover additional unpleasant surprises.

1. replace history vector with ring buffer - agreed
2. For the rest I think we need some kind of brainstorming session. My feeling is that we have to implement it lock-free in one way or another.

elBoberido:
@elfenpiff I have the same feeling. The mutex in the `ChunkDistributor` is just the tip of the iceberg. It's not that easy to do a `tryLock` in the discovery loop since it would be needed to cache the capro message if the lock could not be obtained. Alternatively a busy wait could be done with a `tryLockOrRevokeLockFromAbsentParticipant` method could be introduce, but that smells fishy. I fear we either need a lock-free mechanism or a bigger refactoring

MatthiasKillat:
@elboberido I think we need a lock-free caching structure for this. Maybe we can discuss the exact requirements again, but I think I have some idea I could sketch. I will do so in a few days.

From my point of view the requirements are.
1. Concurrent write access by one party only (add/remove)
2. Concurrent read access by multiple parties (random access and iteration)
3. efficient add/remove operations
4. iterate and read operations
5. operations must be non-blocking on the reader side at least, but ideally completely lock-free
6. ... anything else?

elfenpiff:
> @elBoberido I think we need a lock-free caching structure for this. Maybe we can discuss the exact requirements again, but I think I have some idea I could sketch. I will do so in a few days.

@MatthiasKillat I think I already figured this out and wrote something. Here is the code

```cpp
template <typename T>
struct Cache {
    static_assert(std::is_trivially_copyable<T>::value,
                  "T needs to be trivially copyable");
    void write(const T& value) {
        memcpy(&cache[(abaCounter.load() + 1) % 2], &value);
        abaCounter.fetch_add(1, std::memory_order_release);
    }

    T read() {
        uint64_t currentAba = abaCounter.load(std::memory_order_acquire);
        do {
            T data;
            memcpy(&data, cache[currentAba % 2]);
        } while (abaCounter.compare_exchange_strong(currentAba, currentAba,
                                                    std::memory_order_relaxed,
                                                    std::memory_order_acquire));
    }

    T cache[2];
    std::atomic<uint64_t> abaCounter{0};
};
```

I think this fulfills the requirements:
- one writer
- arbitrary number of readers
- arbitrary data type (as long as it is trivally copyable)

@MatthiasKillat what do you think? Is this something to work with or do you see some improvements - if so feel free to use or discard this.

MatthiasKillat:
@elfenpiff This will not work in itself, due to ring-buffer semantics (we do not want this). I also question how this will deal with iteration and also have some more optimizations in mind. I am not sure whether I have the requirements right, but the proposal does not satisfy all of them. Will post details later.

But yes, the easy way is to restrict T to be trivially copyable, and this is sufficient for our case. More precisely (and general), the copy must only be non-disruptive in a sense that it does cause no harm in reading while it is modified (but the modification is detected afterwards). A copy ctor operating on memory that we always own may also satisfy this, depending on what it does.

I believe it is solvable for general T, but less efficient. An educated guess is also that we may not even need `compare_exchange` operations (which are fairly costly but no worse than mutexes generally).

My other optimizations are wrt. access efficiency.

MatthiasKillat:
After closer inspection of the use case, we need to distinguish the following. 

1) For the history cache (which currently uses a vector, inefficiently due to front removal) we can use ringbuffer semantics. We still need (almost) random access reads though and they should not be protected by the abaCounter (too coarse and pessimistic) but a different counter.

2) For The chunks in use cache we need data structures that allow arbitrary removal (i.e. no ringbuffer). We can also create this kind of cache structure. I am not sure whether @elBoberido already had something for this problem.

All of this will in itself not be sufficient to eliminate the lock in the ChunkDistributor, since  adding the queues (currently) has to be protected. It may eliminate some user-side locking though (i.e. the user cannot cause deadlocks anymore), not entirely sure though (would have to take a deeper look).

elBoberido:
@elfenpiff I think we should not copy since there is almost no contention and this would be a performance hit for each sample to deliver on each subscriber. We can have an asymmetric lock were only RouDi can lock and the application will block. RouDi would also detect in an application is in the critical section while locking and potentially let the application do some work for RouDi.

@MatthiasKillat there is already the `UsedChunkList` which is able to store the chunks in a way which prevents torn writes and enables RouDi to clean up even if the list is corrupt. We need something similar for the history but as ring buffer, like you already mentioned. This is not a big problem, though.

elBoberido:
I close this PR since it's more or less a dead end and would requite further changes in iceoryx

Issue #539, opened 2021-02-03
RouDi modularisation


mossmaurice:
## Brief feature description

The current modules and classes in RouDi are entangled and sometimes depend on each other. For safety certification a clear structure and concept is needed.

## Detailed information

* [ ] Replace `m_run` in `CmdLineArgs_t`, also see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/523#discussion_r569483881)

PR #750, opened 2021-04-21
Iox #261 Replace std::terminate with Expects/Ensures

### LINKED ISSUE #261: Clean-up std::terminate usage

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Replace `std::terminate` with `Expects`/`Ensures` or `errorHandler()`
* Resolve some cyclic dependencies
  * Move `makeScopedStatic()` to its own hpp / inl

## References

- Closes #261


iceoryx_posh/source/roudi/process_manager.cpp
```diff
@@ -49,23 +49,23 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
     if (!maybeSegmentManager.has_value())
     {
         LogFatal() << "Invalid state! Could not obtain SegmentManager!";
-        std::terminate();
+        errorHandler(Error::kROUDI__PRECONDITIONS_FOR_PROCESS_MANAGER_NOT_FULFILLED, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 52: I thought every error enum should be used only once but here we use the same error enum at three places. 
The reason is that if we encounter an error value later in some log we exactly know where it originated, in this case however we would have to look at three possibilities.

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -125,17 +125,17 @@ const RuntimeName_t& PoshRuntime::verifyInstanceName(cxx::optional<const Runtime
     if (!name.has_value())
     {
         LogError() << "Cannot initialize runtime. Application name has not been specified!";
-        std::terminate();
+        errorHandler(Error::kPOSH__RUNTIME_NO_NAME_PROVIDED, nullptr, ErrorLevel::FATAL);
```
elfenpiff, line 128: Please use every error enum only once and introduce some additional error enums here.

iceoryx_utils/include/iceoryx_utils/cxx/helplets.hpp
```diff
@@ -17,15 +17,14 @@
...
-#include "iceoryx_utils/cxx/generic_raii.hpp"
-#include "iceoryx_utils/platform/platform_correction.hpp"
-
 #include <assert.h>
 #include <cstdint>
 #include <iostream>
 #include <limits>
 #include <type_traits>
 
+#include "iceoryx_utils/platform/platform_correction.hpp"
```
elfenpiff, line 26: Nice! Thanks, the platform correction has to be always the last include to cleanup the mess which other platforms produce.

iceoryx_utils/include/iceoryx_utils/cxx/make_scoped_static.hpp
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
```
elfenpiff, line 1: This is wrong. It was created today.

I know the first implementation was in 2019 in the helplets.hpp file but thanks to git the whole history is preserved and correct. Here we have some implicit assumptions which just lead directly into a weird and complex world.

elfenpiff, line 1: Could you rename this file into `scoped_static.hpp` and maybe create an alias `using ScopedStatic = GenericRAII`.
The function `make_scoped_static` is then the factory to create a `ScopedStatic`.

iceoryx_utils/include/iceoryx_utils/cxx/make_scoped_static.hpp
```diff
@@ -0,0 +1,41 @@
...
+#ifndef IOX_UTILS_CXX_MAKE_SCOPED_STATIC_HPP
+#define IOX_UTILS_CXX_MAKE_SCOPED_STATIC_HPP
+
+#include "iceoryx_utils/cxx/generic_raii.hpp"
+
+namespace iox
+{
+namespace cxx
+{
+/// @todo better name
```
elfenpiff, line 26: I suggest `ScopyDopyCleanerThingyAllHailHypnotoadWeLoveUnicornsWhichTasteLikeScotchDuringAWarmWinterbreezeButWhyDoYouCare`

iceoryx_utils/include/iceoryx_utils/internal/cxx/function_ref.inl
```diff
@@ -60,12 +60,8 @@ function_ref<ReturnType(ArgTypes...)>::operator=(function_ref<ReturnType(ArgType
...
 inline ReturnType function_ref<ReturnType(ArgTypes...)>::operator()(ArgTypes... args) const noexcept
 {
-    if (!m_pointerToCallable)
-    {
-        // Callable was called without user having assigned one beforehand
-        std::cerr << "Empty function_ref invoked" << std::endl;
-        std::terminate();
-    }
+    // Expect that a callable was assigned beforehand
+    cxx::Expects((m_pointerToCallable != nullptr) && "Empty function_ref invoked");
```
elfenpiff, line 64: In `method_callback.hpp` we adjusted the signature to return the `ReturnType` packed inside an expected like:
```cpp
template <typename... MethodArguments>
expected<ReturnValue, MethodCallbackError> operator()(MethodArguments&&... args)
```

This error handling is much cleaner since the user is informed and this error is actually something where we can recover from easily. So let the user decide what to do.

If we go the way with the `Expects` the user has to be aware of this and has to read the documentation carefully - which most of the time does not happen. But if we return an expected, the user knows that something can go south, since the compiler enforces an error handling here.
This can reduce unexpected failures in user code.

iceoryx_utils/include/iceoryx_utils/internal/cxx/forward_list.inl
```diff
@@ -324,15 +324,15 @@ template <typename T, uint64_t Capacity>
 inline T& forward_list<T, Capacity>::front() noexcept
 {
     auto iter = begin();
-    handleInvalidElement(iter.m_iterListNodeIdx);
+    IOX_DISCARD_RESULT(handleInvalidElement(iter.m_iterListNodeIdx));
```
elfenpiff, line 327: Could you please remove the `handleInvalidElement` and all the calls and replace it with
```cpp
cxx::Expects(isValidElementIdx(idx));
```
This is what the method does, we do not need to overcomplicate things here...

iceoryx_utils/include/iceoryx_utils/internal/cxx/list.inl
```diff
@@ -323,31 +323,31 @@ template <typename T, uint64_t Capacity>
 inline T& list<T, Capacity>::front() noexcept
 {
     auto iter = begin();
-    handleInvalidElement(iter.m_iterListNodeIdx);
+    IOX_DISCARD_RESULT(handleInvalidElement(iter.m_iterListNodeIdx));
```
elfenpiff, line 326: Could you please remove the `handleInvalidElement` and all the calls and replace it with
```cpp
cxx::Expects(isValidElementIdx(idx));
```
This is what the method does, we do not need to overcomplicate things here...

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -291,44 +288,32 @@ inline const T& vector<T, Capacity>::operator[](const uint64_t index) const noex
...
 
 template <typename T, uint64_t Capacity>
 inline const T& vector<T, Capacity>::front() const noexcept
 {
-    if (empty())
-    {
-        std::cerr << "Attempting to access the front of an empty vector!" << std::endl;
-        std::terminate();
-    }
+    // No empty vector shall be accessed
```
elfenpiff, line 299: I think we should use const cast here and return the const casted value of the non const `front()` method to avoid code duplication. Otherwise the doubled error handling can become easily out of sync.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -291,44 +288,32 @@ inline const T& vector<T, Capacity>::operator[](const uint64_t index) const noex
...
 
 template <typename T, uint64_t Capacity>
 inline const T& vector<T, Capacity>::back() const noexcept
 {
-    if (empty())
-    {
-        std::cerr << "Attempting to access the back of an empty vector!" << std::endl;
-        std::terminate();
-    }
+    // No empty vector shall be accessed
```
elfenpiff, line 315: Please return the const casted return value of the non const `back()` method to avoid code duplication.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/mutex.hpp
```diff
@@ -56,7 +56,7 @@ class mutex
 {
   public:
     /// @brief the construction of the mutex can fail which will lead to a call
-    ///         to std::terminate which is alright for the moment since we are
+    ///         to programm termination which is alright for the moment since we are
```
elfenpiff, line 59: will lead to a call to program termination ... sounds somehow weird. Could you please rephrase it a bit
Also typo: program.

iceoryx_utils/source/posix_wrapper/access_control.cpp
```diff
@@ -79,20 +79,12 @@ AccessController::smartAclPointer_t AccessController::createACL(const int32_t f_
     auto aclInitCall = cxx::makeSmartC(
         acl_init, cxx::ReturnMode::PRE_DEFINED_ERROR_CODE, {static_cast<acl_t>(nullptr)}, {}, f_numEntries);
 
-    if (aclInitCall.hasErrors())
-    {
-        std::cerr << "Error: Could not allocate new ACL." << std::endl;
-        std::terminate();
-    }
+    cxx::Ensures(!aclInitCall.hasErrors() && " Could not allocate new ACL");
```
elfenpiff, line 82: Here it makes more sense to return a `cxx::expected` and let the user handle the error.
Everytime we hide some kind of termination inside an `cxx::Ensures`, `cxx::Expects` we increase the likelyhood that some unexpected termination happens due to a recoverable error.

But if we return an expected the compiler enforces error handling to the user and the user has to make sure that all requirements are met.
So we do not hide the error, we let the user face it.

iceoryx_utils/source/posix_wrapper/shared_memory_object/allocator.cpp
```diff
@@ -37,13 +37,9 @@ void* Allocator::allocate(const uint64_t size, const uint64_t alignment) noexcep
...
     cxx::Expects(size > 0);
 
-    if (m_allocationFinalized)
-    {
-        std::cerr << "allocate() call after finalizeAllocation()! You are not allowed to acquire shared memory chunks "
-                     "anymore!"
-                  << std::endl;
-        std::terminate();
-    }
+    cxx::Expects(
```
elfenpiff, line 40: Here we should also return an `cxx::expected` instead of using `Expects` and `Ensures`.

This error is easily recoverable and the user can handle it if they are informed.

elfenpiff (changes requested):
Please try to avoid the usage of `cxx::Expects` and `cxx::Ensures` as much as possible. I think they should only be used when you encounter an error from which you can not recover or where it is impossible for you to continue.

Please try to use `cxx::expected` as often as possible. The reasons are:
- the compiler will force the user to perform some error handling. When you use `cxx::Expects`/`cxx::Ensures` the user will most likely be not aware that the program can terminate under certain conditions.
- If you use `cxx::expected` the error is baked into the signature and documented
- When you use `cxx::Expects/Ensures` you have to document this in a doxygen note or somewhere in the brief section. This can be easily be overlooked, by us when we refactor the code, by the user when they use the code. 

The last point is demonstrated well in this PR, since in most of the doxygen docu it is nowhere mentioned that the methods which are using `cxx::Expects`/`cxx::Ensures` can be lead to fatal process termination.

iceoryx_posh/source/roudi/process_manager.cpp
mossmaurice: This was done intentionally to show there is still some work todo with our [RouDi/POSH split](https://github.com/eclipse-iceoryx/iceoryx/issues/539). Added `@todo` and moved to bottom.

iceoryx_utils/include/iceoryx_utils/cxx/make_scoped_static.hpp
mossmaurice: Sorry but copyright does not rely on the Git history. This file was copied and pasted and not written by me. As a result all copyright information has to be copied to new file as well. It's done like this in every FOSS project I've worked on.

mossmaurice: Renamed the file.

@elBoberido IIUC, `makeScopedStatic` goes hand-in-hand with `GenericRAII`? Generalising would mean our `ScopedStatic` would require taking one callable, which for me is mainly `GenericRAII`.

iceoryx_utils/include/iceoryx_utils/internal/cxx/forward_list.inl
mossmaurice: Done

FYI @bishibashiB

mossmaurice: You're right, done.

@bishibashiB FYI, feel free to push some changes or provide other ideas.

iceoryx_utils/include/iceoryx_utils/internal/cxx/function_ref.inl
mossmaurice: I definitely see your point. We should try to use `expected` wherever possible. However, IMHO it's more important to be close to the `function_ref` proposal of the standard committee. I would try to avoid adding more dependencies like the `expected` here but rather try to be aligned with the STL. This has the usability downsides that you mentioned, but for the sake of compatibility we should strive for plain and simple building blocks in utils/hoofs.

Added a hint in doxygen.

iceoryx_posh/source/roudi/process_manager.cpp
elfenpiff: Sorry for opening this conversation again but I do not understand this. We intentionally create technical debt and dismiss our error handling concept (with unique error enums) so that we can add it to an issue and clean it up again?
Wouldn't it be easier to use three different error enums here right away and show that there is some work todo in the issue itself? Why do we have to show that there is work to do with bad code style?

iceoryx_posh/source/runtime/posh_runtime.cpp
elfenpiff: It is still used twice, see line 133.

iceoryx_utils/include/iceoryx_utils/cxx/make_scoped_static.hpp
elfenpiff: @mossmaurice I strongly disagree. @elBoberido , @dkroenke , @budrus @FerdinandSpitzschnueffler @MatthiasKillat  what do you think?

The reason is the following. Let's assume you are right then everytime we do some refactoring and move a function into a new file we have to trace the whole function. This means when it was first written in file A, then moved to file B, then to file C and with the new refactoring to file D we have to trace the git history of the function through the files. In the new file we have then to add all the people who ever touched a piece of the function in the copyright header.

It is getting even worse when we would like to split up the function in the next refactoring in to different files. Then we have to trace every line!!! With every file we create and move code we have trace all the lines where they did come from! This is absolutely horrible.

Please do not close or resolve the conversation until we all agreed on how we proceed in such cases.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
```diff
@@ -268,11 +268,8 @@ inline T& vector<T, Capacity>::at(const uint64_t index) noexcept
 template <typename T, uint64_t Capacity>
 inline const T& vector<T, Capacity>::at(const uint64_t index) const noexcept
 {
-    if (index + 1u > m_size)
-    {
-        std::cerr << "out of bounds access, current size is " << m_size << " but given index is " << index << std::endl;
-        std::terminate();
-    }
+    // No out of bound accesses shall be possible
+    cxx::Expects(!(index + 1u > m_size) && "Out of bounds access");
```
elBoberido, line 272: Negations makes it harder to reason about complex conditions. Why not just `cxx::Expects(index < m_size && "...");`? This has the additional benefit to be more performant

iceoryx_utils/source/posix_wrapper/shared_memory_object/allocator.cpp
elBoberido: In this case I would keep the `Expects` since the user clearly did something wrong. by not providing enough memory to the allocator in the first place

iceoryx_utils/include/iceoryx_utils/cxx/make_scoped_static.hpp
elBoberido: I'm not sure what you mean? If you suggest to remove the dependency to GenericRAII, go for it.

I would also not create the `using ScopedStatic = GenericRAII` alias since it would indicate that the return value is a `ScopedStatic`, which it is not. The scoped static is the `cxx::optional<T>` or any other container which has a `emplace` and `destroy` method. A better name is welcome.

A potential solution would be
```c++
template <typename T>
class ScopedStaticGuard {
  public:
    ~ScopedStaticGuard() { m_meory.reset(); }
  private:
    friend .... makeScopedStatic
    ScopedStaticGuard(T& memory) : m_memory(memory) {}

    T& m_memory;
};

template <typename T, typename... CTorArgs>
static ScopedStaticGuard makeScopedStatic(T& memory, CTorArgs&&... ctorArgs) {
    memory.emplace(std::forward<CTorArgs>(ctorArgs)...);
    return ScopedStaticGuard<T>{memory};
}
```

iceoryx_posh/source/roudi/process_manager.cpp
```diff
@@ -45,29 +45,37 @@ ProcessManager::ProcessManager(RouDiMemoryInterface& roudiMemoryInterface,
...
     auto maybeMgmtSegmentId = m_roudiMemoryInterface.mgmtMemoryProvider()->segmentId();
     if (!maybeMgmtSegmentId.has_value())
     {
         LogFatal() << "Invalid state! Could not obtain SegmentId for iceoryx management segment!";
-        std::terminate();
+        fatalError = true;
     }
     m_mgmtSegmentId = maybeMgmtSegmentId.value();
+
+    if (fatalError)
```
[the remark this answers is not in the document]
mossmaurice: I see your point, but IMHO the maintenance effort is not much higher, hence I'll stick with the current approach as this will be refactored relatively soon anyway.

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -135,8 +135,19 @@ TEST_F(PoshRuntime_test, LeadingSlashAppName)
 {
     const iox::RuntimeName_t invalidAppName = "/miau";
 
-    EXPECT_DEATH({ PoshRuntime::initRuntime(invalidAppName); },
-                 "Cannot initialize runtime. Please remove leading slash from Application name /miau");
+    auto errorHandlerCalled{false};
+    iox::Error receivedError{iox::Error::kNO_ERROR};
+    auto errorHandlerGuard = iox::ErrorHandler::SetTemporaryErrorHandler(
```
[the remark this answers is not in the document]
mossmaurice: You are right. I removed this test case from the fixture like in the test case below. IMHO that's better than a death test because we check for the actual error.

iceoryx_utils/include/iceoryx_utils/internal/cxx/vector.inl
mossmaurice: Done

iceoryx_utils/source/posix_wrapper/access_control.cpp
mossmaurice: I agree with you in this case! Adapted the code accordingly.

iceoryx_utils/source/posix_wrapper/shared_memory_object/allocator.cpp
mossmaurice: I'm torn between here :grimacing: In the end it's about the contract with the user. As this is such a simple class and the user will only use c'tor and the call `allocate()` I'm inclined to just add doxygen hint about the behaviour instead of adding a extra dependency by using `expected.`

iceoryx_utils/include/iceoryx_utils/cxx/make_scoped_static.hpp
mossmaurice: Ok, we could consider this. But that's something for another pull request.

dkroenke: @elfenpiff @mossmaurice 
Well, looking at the Eclipse Handbook could maybe clarify this: https://www.eclipse.org/projects/handbook/#legaldoc-faq

> How do file headers change when content is refactored?

> When you split a file during a refactoring exercise, any new files that you create must automatically inherit the header of the original. If even fragment of text from a single line of the original content survives the refactoring (whether generated, rote, or otherwise), then the original copyright statement and author information should be retained.

> Even a refactoring exercise that does not change behaviour, may add intellectual property. When refactoring adds content, copyright statements should be updated to reflect the addition. In concrete terms, you can either add the new copyright holders to the existing copyright statement or append "and others" (especially when the list of copyright holders gets long).

mossmaurice: Ok, thanks @dkroenke for clarifying this. If we should get into a situation with dozens of copyright holder we might switch to "and others". We can raise awareness about this topic in the next dev meetup.

elfenpiff: @mossmaurice @dkroenke How do we ensure the correctness of the copyright headers as a reviewer? Do we have to retrace the code lines and functions over the total git history? Or anyone else?

At the moment we can handle this somehow but when there will be more this is the ultimate time killer.

iceoryx_utils/include/iceoryx_utils/cxx/vector.hpp
```diff
@@ -29,6 +31,9 @@ namespace cxx
 /// @brief  C++11 compatible vector implementation. We needed to do some
 ///         adjustments in the API since we do not use exceptions and we require
 ///         a data structure which can be located fully in the shared memory.
+///
+/// @note   Out of bounds access or accessing an empty vector can lead to a program termination!
```
elfenpiff, line 35: Wrong place. Please add this note to all methods which have such behavior.

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/mutex.hpp
```diff
@@ -52,12 +52,14 @@ namespace posix
 ///         }
 ///
 ///     }
+///
+/// @note   Errors in c'tor or d'tor can lead to a program termination!
```
elfenpiff, line 56: I think it would be better to add this note to the ctor as well as the dtor.

elfenpiff (commented):
Please place the termination comments directly into the function/method documentation and then you get my thumbs up :)

iceoryx_posh/source/roudi/process_manager.cpp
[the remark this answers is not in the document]
elBoberido: it's your decision, but keep in mind, nothing is more definitive than the temporary ;)

iceoryx_utils/source/posix_wrapper/access_control.cpp
```diff
@@ -34,12 +33,18 @@ bool AccessController::writePermissionsToFile(const int32_t f_fileDescriptor) co
...
+    {
+        std::cerr << "Error: Creating ACL failed." << std::endl;
+        return false;
+    }
 
     // add acl entries
     for (const auto& entry : m_permissions)
     {
-        if (!createACLEntry(workingACL.get(), entry))
+        if (!createACLEntry(maybeWorkingACL.value().get(), entry))
```
elBoberido, line 47: please add a `auto& workingACL = maybeWorkingACL.value();` after the `if (maybeWorkingACL.has_error())` statement and use that instead of always calling `maybeWorkingACL.value();`

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/mutex.hpp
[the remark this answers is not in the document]
mossmaurice: Makes sense, edited.

iceoryx_utils/source/posix_wrapper/access_control.cpp
mossmaurice: That's a valid point, changed.

elBoberido (dismissed):

iceoryx_utils/source/posix_wrapper/mutex.cpp
```diff
@@ -61,25 +61,15 @@ mutex::mutex(bool f_isRecursive)
         !cxx::makeSmartC(pthread_mutexattr_destroy, cxx::ReturnMode::PRE_DEFINED_SUCCESS_CODE, {0}, {}, &attr)
```
elBoberido, line 61: please update copyright

iceoryx_utils/source/posix_wrapper/access_control.cpp
```diff
@@ -15,13 +15,12 @@
 // SPDX-License-Identifier: Apache-2.0
```
elBoberido, line 15: please update copyright

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/mutex.hpp
```diff
@@ -52,14 +52,16 @@ namespace posix
 ///         }
```
elBoberido, line 52: please update copyright

iceoryx_utils/include/iceoryx_utils/internal/posix_wrapper/access_control.hpp
```diff
@@ -16,7 +16,7 @@
 #ifndef IOX_UTILS_POSIX_WRAPPER_ACCESS_CONTROL_HPP
```
elBoberido, line 16: please update copyright

iceoryx_utils/include/iceoryx_utils/internal/cxx/list.inl
```diff
@@ -323,31 +323,31 @@ template <typename T, uint64_t Capacity>
 inline T& list<T, Capacity>::front() noexcept
```
elBoberido, line 323: please update copyright

iceoryx_utils/include/iceoryx_utils/internal/cxx/function_ref.inl
```diff
@@ -60,12 +60,8 @@ function_ref<ReturnType(ArgTypes...)>::operator=(function_ref<ReturnType(ArgType
 template <class ReturnType, class... ArgTypes>
```
elBoberido, line 60: please update copyright

iceoryx_utils/include/iceoryx_utils/internal/cxx/forward_list.inl
```diff
@@ -324,15 +324,15 @@ template <typename T, uint64_t Capacity>
 inline T& forward_list<T, Capacity>::front() noexcept
```
elBoberido, line 324: please update copyright

iceoryx_utils/include/iceoryx_utils/cxx/list.hpp
```diff
@@ -17,6 +17,8 @@
 #ifndef IOX_UTILS_CXX_LIST_HPP
```
elBoberido, line 17: please update copyright

iceoryx_utils/include/iceoryx_utils/cxx/forward_list.hpp
```diff
@@ -17,6 +17,8 @@
 #ifndef IOX_UTILS_CXX_FORWARD_LIST_HPP
```
elBoberido, line 17: please update copyright

iceoryx_posh/source/mepoo/mepoo_config.cpp
```diff
@@ -35,8 +35,8 @@ void MePooConfig::addMemPool(MePooConfig::Entry f_entry) noexcept
     }
```
elBoberido, line 35: please update copyright

elBoberido (approved):

elfenpiff (approved):

PR #754, opened 2021-04-28
iox-#337 Introduce `PeriodicTimer`

shankar-in:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes **https://github.com/eclipse-iceoryx/iceoryx/issues/337**


shankar-in:
The disabled test which was part of [#243 ](https://github.com/eclipse-iceoryx/iceoryx/issues/243) is also enabled in this PR.

shankar-in:
Looking currently on the failures of windows and macos build.

elBoberido:
> Looking currently on the failures of windows and macos build.

Regarding the Windows issue. If you need to compile locally, this is a howto
- if you need a Windows Image, Microsoft provides free Developer Images at https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/
- download and install cmake from https://cmake.org/download/ (set option to add cmake to the system PATH for all users)
- download and install git from https://gitforwindows.org/
- run VisualStudioInstaller and install `Desktop development with C++` Workload via the `Modify` button on `Visual Studio Community 2019`
- either `VS Code` or `Developer Command Prompt` can be used to build iceoryx with `cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON` followed by `cmake --build build` (maybe one or two restarts are required to let cmake find the compiler)
- alternatively `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat` can be executed in any shell

I need to put this to the CONTRIBUTION.md at some time

shankar-in:
> > Looking currently on the failures of windows and macos build.
> 
> Regarding the Windows issue. If you need to compile locally, this is a howto
> 
>     * if you need a Windows Image, Microsoft provides free Developer Images at https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/
> 
>     * download and install cmake from https://cmake.org/download/ (set option to add cmake to the system PATH for all users)
> 
>     * download and install git from https://gitforwindows.org/
> 
>     * run VisualStudioInstaller and install `Desktop development with C++` Workload via the `Modify` button on `Visual Studio Community 2019`
> 
>     * either `VS Code` or `Developer Command Prompt` can be used to build iceoryx with `cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON` followed by `cmake --build build` (maybe one or two restarts are required to let cmake find the compiler)
> 
>     * alternatively `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat` can be executed in any shell
> 
> 
> I need to put this to the CONTRIBUTION.md at some time

Thanks @elBoberido My host machine has windows. So let me try to install the CMake to reproduce the issue locally. Yes having this in the Readme helps!

dkroenke:
The build log gives also already useful infos about the failure:
```
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(174,1): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(162,1): error C2146: syntax error: missing ';' before identifier 'EXPECTED_DELAY' [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(162,1): error C2065: 'EXPECTED_DELAY': undeclared identifier [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(162,1): error C2660: 'iox::utils::testing::verifyTimingTestResult': function does not take 5 arguments [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\testing\include\iceoryx_utils/testing/timing_test.hpp(118,13): message : see declaration of 'iox::utils::testing::verifyTimingTestResult' [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(176,1): error C2065: 'uint': undeclared identifier [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(176,1): error C2146: syntax error: missing ';' before identifier 'EXPECTED_DELAY' [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(176,1): error C2065: 'EXPECTED_DELAY': undeclared identifier [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
D:\a\iceoryx\iceoryx\iceoryx_utils\test\moduletests\test_cxx_timer_periodic.cpp(176,1): error C2660: 'iox::utils::testing::verifyTimingTestResult': function does not take 5 arguments [D:\a\iceoryx\iceoryx\build\utils\test\utils_moduletests.vcxproj]
```

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+/// After the wait if the timer is enabled and to be activated with delay it returns TICK_DELAY.
+/// After the wait if the timer is enabled and to be activated with delay breaching the delay threshold it returns
+/// TICK_THRESHOLD_DELAY.
+enum class TimerEvent : uint8_t
+{
+    STOP,
+    TICK,
+    TICK_DELAY,
+    TICK_THRESHOLD_DELAY
+};
```
elBoberido, line 52: I'm not sure this is a good approach. It makes the Timer cumbersome to use. The `TICK_THRESHOLD_DELAY` is potentially an error and should be in the error path of the expected.

I would stick to just `STOP` and `TICK` and keep this simple. For the `TICK_DELAY` and `TICK_THRESHOLD_DELAY` I would recommend to make that part of the `CatchupPolicy` like
```cpp
enum class CatchupPolicy
{
    IMMEDIATE_TICK,
    SKIP_TO_NEXT_TICK,
    ERROR
};
```
and 
```cpp
enum class ErrorCause{
    TICK_EXCEEDED_TIME_LIMIT,
    OTHER_ERRORS
};

struct TimerError {
    ErrorCause cause;
    Duration tickDelay;
};
```

This could be either be a ctor parameter or a parameter to the `wait` method.
```cpp
cxx::expected<TimerEvent, TimerError> wait(CatchupPolicy p) noexcept;
```
with the following usage
```cpp
Timer t{100_ms};
t.start();
t.wait(CatchupPolicy::ERROR).or_else([] (auto& err) {
    if (err.cause == ErrorCause::TICK_EXCEEDED_TIME_LIMIT && err.tickDelay > 5_ms) {...}
});
```

Maybe it's also a good idea to completely get rid of the `TimerEvent` and handle the `STOP` case also with the `TimerError`, which would need to be renamed in this case.

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+    /// @brief This function returns the current time.
+    /// @return The system clock real time is returned.
+    const iox::units::Duration now() const noexcept;
+
+    /// @brief Briefly waits for the timer interval. This is achieved by trying to do a timed wait for the interval
+    /// duration on the already acquired binary semaphore.
+    /// @return TICK if the timer is active, STOP if the timer is stopped already, TICK_DELAY if there is delay and
+    /// TICK_THRESHOLD_DELAY if the delay is more than the given delay threshold.
+    cxx::expected<iox::cxx::TimerEvent, posix::SemaphoreError> wait() noexcept;
```
elBoberido, line 103: please don't use the `SemaphoreError`. It's an implementation detail and should not be exposed to the user

elBoberido (commented):
Only partially rewieved. Please add a link to the issue in the PR description

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
elBoberido: well, on the other hand a
```cpp
bool keepRunning = true;
while (keepRunning) {
    t.wait(CatchupPolicy::IMMEDIATE_TICK).and_then([&](TimerEvent e) {
        if(e == TimerEvent::STOP) {
            keepRunning = false;
            return;
        }
        // do stuff
    });
}
```
is also okay to use

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+#ifndef IOX_UTILS_PERIODIC_TIMER_HPP
+#define IOX_UTILS_PERIODIC_TIMER_HPP
+
+#include "iceoryx_utils/cxx/expected.hpp"
+#include "iceoryx_utils/internal/units/duration.hpp"
+#include "iceoryx_utils/posix_wrapper/semaphore.hpp"
+#include <thread>
+
+using namespace iox::units;
+using namespace iox::units::duration_literals;
```
mossmaurice, line 26: `using namespace` in *.hpp's leads to [namespace pollution](https://stackoverflow.com/questions/22903542/what-is-namespace-pollution) and higher build times. Better call `fromSeconds` et al:

```
  private:
    iox::units::Duration m_interval{units::Duration::fromSeconds(0)};
    iox::units::Duration m_timeForNextActivation{units::Duration::fromSeconds(0)};
```

`using namespace iox::units::duration_literals` is perfectly fine to use in `timer.cpp` in a local scope.

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+    /// binary semaphore.
+    /// @param[in] interval The new duration until the timer sleeps and wakes up for execution
+    void start(const units::Duration interval) noexcept;
+
+    /// @brief Stops the timer. This also releases the acquired binary semaphore.
+    void stop() noexcept;
+
+    /// @brief This function returns the current time.
+    /// @return The system clock real time is returned.
+    const iox::units::Duration now() const noexcept;
```
mossmaurice, line 97: This can be a static method.

iceoryx_utils/source/cxx/timer.cpp
```diff
@@ -0,0 +1,98 @@
...
+    {
+        auto stopResult = m_waitSemaphore.post();
+        cxx::Expects(!stopResult.has_error());
+    }
+}
+
+const iox::units::Duration Timer::now() const noexcept
+{
+    struct timespec ts;
+    clock_gettime(CLOCK_REALTIME, &ts);
```
mossmaurice, line 56: For easier error handling please use `SmartC` for all C calls, see old `posix::timer`

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+enum class TimerEvent : uint8_t
+{
+    STOP,
+    TICK,
+    TICK_DELAY,
+    TICK_THRESHOLD_DELAY
+};
+
+
+/// @brief This class offers periodic timer functionality. This timer is started immediately upon construction.
```
mossmaurice, line 55: Nice documentation!

iceoryx_utils/source/cxx/timer.cpp
```diff
@@ -0,0 +1,98 @@
...
+    m_interval = interval;
+    start();
+}
+
+void Timer::stop() noexcept
+{
+    if (*(m_waitSemaphore.getValue()) == static_cast<int>(posix::SemaphoreWaitState::TIMEOUT))
+    {
+        auto stopResult = m_waitSemaphore.post();
+        cxx::Expects(!stopResult.has_error());
```
mossmaurice, line 49: `Ensures` and `Expects` have exactly the same implementation. However, `Expects` is [preferred](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ri-ensures) for pre-conditions and `Ensures` for post-conditions.

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+namespace cxx
+{
+/// @brief This enum class offers the timer state events.
+/// If the timer is active the TimerEvent has ENABLED as the state.
+/// If tht timer is inactive the TimerEvent has DIABLED as the state.
+enum class TimerState : uint8_t
+{
+    DISABLED,
+    ENABLED
+};
```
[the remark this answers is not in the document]
shankar-in: Okay added the docs to the enum value line

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IOX_UTILS_PERIODIC_TIMER_HPP
```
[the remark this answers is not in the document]
shankar-in: Okay I renamed to `periodic_timer` and moved to the `posix` namespace

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include "iceoryx_utils/internal/units/duration.hpp"
+#include "iceoryx_utils/posix_wrapper/semaphore.hpp"
+#include <thread>
+
+using namespace iox::units;
+using namespace iox::units::duration_literals;
+
+namespace iox
+{
+namespace cxx
```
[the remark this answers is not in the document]
shankar-in: Okay. Moved this to the `POSIX_WRAPPER`.

iceoryx_utils/CMakeLists.txt
```diff
@@ -149,6 +149,7 @@ add_library(iceoryx_utils
     source/concurrent/active_object.cpp
     source/concurrent/loffli.cpp
     source/cxx/deadline_timer.cpp
+    source/cxx/timer.cpp
```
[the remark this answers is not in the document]
shankar-in: Yes thats the plan . This PR just introduces the new `periodic_timer` and the next one will integrate the new `periodic_timer` to the `periodic_task`. I have moved the `timer` classes to `posix` namespace and renamed to `periodic_timer`.

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
shankar-in: @elBoberido Actually this concept was proposed by @bishibashiB. Bernhard could you pls take a look the proposal by Bob and suggest?

iceoryx_utils/test/moduletests/test_posix_timer.cpp
```diff
@@ -481,7 +481,7 @@ TIMING_TEST_F(Timer_test, CatchUpPolicySkipToNextBeatCallsLessCallbacksThanASAPT
 /// Unit tests which segfaults (issue #243). If the segfault is fixed this unit test has
 /// to be adjusted but for the moment it seems that it causes the segfault
 /// reliable.
```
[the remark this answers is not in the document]
shankar-in: Okay, Removed the older comments which doent have significance after the #243

iceoryx_utils/source/cxx/timer.cpp
shankar-in: Okay, Replaced with Ensures.

iceoryx_utils/source/cxx/timer.cpp
shankar-in: Okay made use of the `smartC`

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
bishibashiB: Nice @elBoberido 😁 for bringing in a `policy` config 'handshake'. 
First I understand that using an (optional) threshold to differ 'critical' from 'less-critical' delays should be continued?!
Then regarding the policy / error handling IMO you have the perfect overview to bed it into iceoryx codebase: ... as long as the user can read the timer 'return' ...   
Policy via wait()-parameter will hold more freedom to the user (than c'tor) I guess.
@shankar-in you also want to add something to it?

shankar-in: Yes. I do agree that we don't introduce more complexity rather provide easy API for configuration of the timers. 

I understand that we pass the `policy` as a parameter to `wait()` as suggested by @bishibashiB. This also allows to modify the policy from the user side whenever needed. 

So `IMMEDIATE_TICK` precisely tries to call the next activation time and if the execution takes it returns error. The error returned will also have the duration of time of delay.

The `SKIP_TO_NEXT_TICK` in case of delay will get activated in the next available slot until then it blocks.

@elBoberido  Is my understanding right?

elBoberido: @bishibashiB this is just loud thinking since it was similar in the old `Timer`. Maybe it would have been a good idea to create a design document before implementing it, then we could have discussed the API and feature scope in more detail. Maybe it's still a good idea to create a design document alongside this PR.

@shankar-in not quite. The idea is to have the non-error path of the `expected` handle only the happy path and maybe the stop path. The happy path would be when there is no delay or if you don't care about the delay. `IMMEDIATE_TICK` and `SKIP_TO_NEXT_TICK` would be such cases. The former would mean to immediately return with a `TICK` if the target time is already missed and the latter to skip the current tick and wait for the next scheduled one. `CatchupPolicy::ERROR` would mean an immediate return from `wait` if the target time is missed, but in this case in the error path of the `expected`. Instead of just returning an error, the delay from the target time would also be returned and it's up to the user if that is still acceptable. In the future, ff we notice that the use case with the delay is used quite often, we could introduce something like a `RelaxedTimer`/`FuzzyTimer` which would be just a small wrapper around the `Timer` from this PR and make the handling more convenient. Basically just hiding the `if (err.cause == ErrorCause::TICK_EXCEEDED_TIME_LIMIT && err.tickDelay > 5_ms) {...}` in a method which would call `Timer::wait(...)` beforehand.

elBoberido: since a picture is more worth than thousand words, some ASCII art :)
```
Scheduled Ticks               |          |          |          |          |          |
IMMEDIATE_TICK Execution      [-------]  [----]     [-------------][-------][-----]
SKIP_TO_NEXT_TICK Execution   [-------]  [----]     [-------------]       [-------]  [-----]
ERROR Execution               [-------]  [----]     [-------------] <- return with error and let the user decide if the delay is acceptable
```

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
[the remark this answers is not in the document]
elBoberido: Please don't rename it to `periodic_timer`. I know from other timer implementations that they have a `oneShot` functionality. This could easily be added to the `Timer` itself, but not to a new class which wraps the `Timer`.

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
shankar-in: So @bishibashiB I will be reworking on the `delay` and this will be handled as @elBoberido suggested with the `CatchupPolicy`

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
shankar-in: Okay marked this as static

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
```diff
@@ -0,0 +1,115 @@
...
+    /// duration on the already acquired binary semaphore.
+    /// @return TICK if the timer is active, STOP if the timer is stopped already, TICK_DELAY if there is delay and
+    /// TICK_THRESHOLD_DELAY if the delay is more than the given delay threshold.
+    cxx::expected<iox::cxx::TimerEvent, posix::SemaphoreError> wait() noexcept;
+
+  private:
+    iox::units::Duration m_interval{0_ms};
+    iox::units::Duration m_timeForNextActivation{0_ms};
+    iox::units::Duration m_delayThreshold{0_ms};
+    posix::Semaphore m_waitSemaphore{posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U).value()};
```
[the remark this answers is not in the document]
shankar-in: Thanks. Added the clause to handle the error while creating the semaphore.

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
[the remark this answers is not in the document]
shankar-in: @mossmaurice wants to have the file named as `periodic_timer`?

iceoryx_utils/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,146 @@
...
+namespace posix
+{
+PeriodicTimer::PeriodicTimer(const iox::units::Duration interval, const iox::units::Duration delayThreshold) noexcept
+    : m_interval(interval)
+    , m_delayThreshold(delayThreshold)
+{
+    m_waitSemaphore =
+        std::move(posix::Semaphore::create(posix::CreateUnnamedSharedMemorySemaphore, 0U)
+                      .or_else([](posix::SemaphoreError&) {
+                          errorHandler(Error::kROUDI_APP__FAILED_TO_CREATE_SEMAPHORE, nullptr, ErrorLevel::FATAL);
```
elBoberido, line 32: according to the error-handling.md the error handler shall not be used in utils/hoofs. Please use `cxx::Ensures(false && "Could not create Semaphore for PeriodicTimer!")`

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
shankar-in: @elBoberido I have a clarification in SKIP_TO_NEXT policy. I guess we also need to handle if the execution time took very long than the interval set.

+--------------------------+--------------+--------------+
|                          | Case 1       | Case 2       |
+--------------------------+--------------+--------------+
| Initial Time             | 10           | 10           |
+--------------------------+--------------+--------------+
| Catch-up Policy          | SKIP_TO_NEXT | SKIP_TO_NEXT |
+--------------------------+--------------+--------------+
| Interval                 | 50           | 50           |
+--------------------------+--------------+--------------+
|                          |              |              |
+--------------------------+--------------+--------------+
| 1st Cycle Execution Time | 70           | 150          |
+--------------------------+--------------+--------------+
| Expected Activation      | 60           | 60           |
+--------------------------+--------------+--------------+
| Time After Execution     | 80           | 160          |
+--------------------------+--------------+--------------+
| Difference for Activation | 20           | 100          |
+--------------------------+--------------+--------------+
| Wait()                   | 30           | 0            |
+--------------------------+--------------+--------------+
| Next Activation          | 110          | 160          |
+--------------------------+--------------+--------------+

elBoberido: @shankar-in could you please wrap your ASCII art in markdown code blocks to get a monospace font?

shankar-in: `
|                          	| Case 1       	| Case 2       	|
|--------------------------	|--------------	|--------------	|
| Initial Time             	| 10           	| 10           	|
| Catch-up Policy          	| SKIP_TO_NEXT 	| SKIP_TO_NEXT 	|
| Interval                 	| 50           	| 50           	|
|                          	|              	|              	|
| 1st Cycle Execution Time 	| 70           	| 150          	|
| Expected Activation      	| 60           	| 60           	|
| Time After Execution     	| 80           	| 160          	|
| Difference of Activation 	| 20           	| 100          	|
| Wait()                   	| 30           	| 0            	|
| Next Activation          	| 110          	| 160          	|`

@elBoberido Thanks for the hint.

elBoberido: @shankar-in 
I guess you would have something like this
```
Scheduled Ticks               |     a    |     b    |     c    |     d    |     e    |
IMMEDIATE_TICK Execution      [---a---]  [------------b-----------][-c-][-d--][-e--] [---]
SKIP_TO_NEXT_TICK Execution   [---a---]  [------------b-----------]       [---e---]  [-----]
ERROR Execution               [---a---]  [------------b-----------] <- return with error and let the user decide if the delay is acceptable
```

shankar-in: @elBoberido  Yes. I am doing in that. Thank you.

iceoryx_utils/source/posix_wrapper/periodic_timer.cpp
shankar-in: Okay. Replaced with the cxx::Ensures.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+#include <chrono>
+#include <thread>
+
+using namespace ::testing;
+
+using namespace iox::units;
+using namespace iox::units::duration_literals;
+
+using Timer = iox::posix::PeriodicTimer;
```
mossmaurice, line 31: Could you please wrap the whole test in an anonymous namespace like [here](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/aspice_swe3_4/example/iceoryx_component/test/moduletests/test_example_base_class.cpp#L23)? We had many troubles with clashing symbols leading to weird errors lately.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,222 @@
...
+    auto currentSystemTime = iox::units::Duration(ts);
+
+    bool result = timeNow.toMilliseconds() == currentSystemTime.toMilliseconds() ? true : false;
+
+    TIMING_TEST_EXPECT_TRUE(result);
+});
+
+TIMING_TEST_F(PeriodicTimer_test, periodicityWithoutExecutionTimeTest, Repeat(5), [&] {
+    Timer sut(INTERVAL);
+    const int RANGE_APPROX = 2; // 2ms arppoximation. This may be lost in execution.
```
mossmaurice, line 140: Nitpick
```suggestion
    const int RANGE_APPROX = 2; // 2ms approximation. This may be lost in execution.
```

mossmaurice (commented):
Two small findings, otherwise LGTM. Let's get this merged!

iceoryx_utils/include/iceoryx_utils/cxx/timer.hpp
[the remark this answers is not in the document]
mossmaurice: What about the following:

1) For now we stick with `periodic_timer` and get this PR merged
2) We remove the old `posix::timer` and rename the new one to something we all like

In the beginning of this thread I was just noticing the difference between include guard and file name. I would be fine with `periodic_timer` as a final name.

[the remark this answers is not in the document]
shankar-in: @mossmaurice Yes I agree with the above plan.

Initially I named this file as `timer` and after which i modified this to `periodic_timer` as per your feedback.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
shankar-in: :) Thanks. This typo is corrected. Since its also present in 3 other locations I didnt apply the commit suggestion.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    return cxx::success<units::Duration>(ts);
+}
+
+cxx::expected<iox::posix::WaitResult, TimerErrorCause> PeriodicTimer::wait(TimerCatchupPolicy policy) noexcept
+{
+    // To check if the TIMER is active (if the sempahore is acquired)
+    if (*(m_waitSemaphore.getValue()) == static_cast<int>(posix::SemaphoreWaitState::TIMEOUT))
+    {
+        if (now().value() > m_timeForNextActivation)
```
elfenpiff, line 77: `now` returns an expected which can contain an error and here it is accessed without checking if it contains a value.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+        {
+            if (policy == TimerCatchupPolicy::IMMEDIATE_TICK)
+            {
+                m_timeForNextActivation = now().value();
```
elfenpiff, line 81: Again, accessing a value without checking it if its valid. Please use the `and_then`/`or_else` pattern wherever possible to avoid such mistakes.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                m_waitResult.state = iox::posix::TimerState::TICK;
+                return cxx::success<iox::posix::WaitResult>(m_waitResult);
+            }
+            else if (policy == TimerCatchupPolicy::SKIP_TO_NEXT_TICK)
+            {
+                auto delay = now().value() - m_timeForNextActivation; // calculate the time delay
```
elfenpiff, line 87: Accessing expected value without checking if an error is contained.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                    auto totalSlotToBeSkipped =
+                        delay.toMilliseconds() / m_interval.toMilliseconds(); // calculate the total slots missed
+                    m_timeForNextActivation =
+                        m_timeForNextActivation + m_interval * totalSlotToBeSkipped; // Skip to the next activation
+                }
+                else
+                {
+                    m_timeForNextActivation = m_timeForNextActivation + m_interval; // Skip to the next activation
+                }
+                auto timeDiff = m_timeForNextActivation - now().value(); // calculate remaining time for activation
```
elfenpiff, line 99: Accessing expected value without checking if an error is contained.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                }
+                else
+                {
+                    m_waitResult.state = iox::posix::TimerState::TICK;
+                    return cxx::success<iox::posix::WaitResult>(m_waitResult);
+                }
+            }
+            else
+            {
+                auto timeDiff = now().value() - m_timeForNextActivation; // Calculate the time delay
```
elfenpiff, line 113: Accessing expected value without checking if an error is contained.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                m_waitResult.state = iox::posix::TimerState::DELAY;
+                m_waitResult.timeDelay = timeDiff;
+                return cxx::success<iox::posix::WaitResult>(m_waitResult);
+            }
+        }
+        else
+        {
+            auto actualWaitDuration = m_timeForNextActivation - now().value();
```
elfenpiff, line 121: Accessing expected value without checking if an error is contained.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+cxx::expected<iox::posix::WaitResult, TimerErrorCause> PeriodicTimer::wait(TimerCatchupPolicy policy) noexcept
+{
+    // To check if the TIMER is active (if the sempahore is acquired)
+    if (*(m_waitSemaphore.getValue()) == static_cast<int>(posix::SemaphoreWaitState::TIMEOUT))
+    {
+        if (now().value() > m_timeForNextActivation)
+        {
+            if (policy == TimerCatchupPolicy::IMMEDIATE_TICK)
```
elfenpiff, line 79: Instead of if please use `switch case` since the compiler warns us then if some cases are missing.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void PeriodicTimer::start(const iox::units::Duration interval) noexcept
+{
+    m_interval = interval;
+    start();
+}
+
+void PeriodicTimer::stop() noexcept
+{
+    if (*(m_waitSemaphore.getValue()) == static_cast<int>(posix::SemaphoreWaitState::TIMEOUT))
```
elfenpiff, line 51: I am sorry to say this but this makes no sense. `getValue` returns the current value of the semaphore for instance 5 and casting  `posix::SemaphoreWaitState::TIMEOUT` to an integer provides you with an arbitrary number since there are no values defined for that enum.

Before you continue with this PR please
- read `man sem_overview` (for a technical overview of semaphores)
- read the API of our semaphore
- get familiar with C++ enum classes.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+    m_waitSemaphore = std::move(timerSemaphore.value());
+    start();
+}
+
+void PeriodicTimer::start() noexcept
+{
+    stop();
+    auto waitResult = m_waitSemaphore.timedWait(m_interval);
+    cxx::Ensures(!waitResult.has_error());
+    m_timeForNextActivation = now().value() + m_interval;
```
elfenpiff, line 40: Accessing expected value without check if a value is contained.
Please familiarize yourself with the iceoryx error handling concept before continuing this PR.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+    m_interval = interval;
+    start();
+}
+
+void PeriodicTimer::stop() noexcept
+{
+    if (*(m_waitSemaphore.getValue()) == static_cast<int>(posix::SemaphoreWaitState::TIMEOUT))
+    {
+        auto stopResult = m_waitSemaphore.post();
```
elfenpiff, line 53: You can pack this in one line like:
`cxx::Ensures(!m_waitSemaphore.post().has_error());`

elfenpiff (changes requested):

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+const Duration PeriodicTimer_test::INTERVAL{50_ms};
+
+TEST_F(PeriodicTimer_test, TimerAutoStartTest)
+{
+    Timer sut(0_s);
+
+    auto timerState = sut.wait(iox::posix::TimerCatchupPolicy::IMMEDIATE_TICK);
+
+    ASSERT_FALSE(timerState.has_error());
```
elfenpiff, line 60: Please use `ASSERT` for cases that the test cannot proceed if this is not fulfilled. For instance a case like:
```cpp
ASSERT_FALSE(vector.empty());
vector[0]; // would lead to segfault if vector would be empty.
```

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+TEST_F(PeriodicTimer_test, ZeroINTERVALTest)
+{
+    Timer sut(0_s);
+
+    auto timerState = sut.wait(iox::posix::TimerCatchupPolicy::IMMEDIATE_TICK);
+    bool result = timerState.value().state == iox::posix::TimerState::TICK ? true : false;
```
elfenpiff, line 68: Please familiarize yourself with gtest. You can use also constructs like `ASSERT_EQ`, `ASSERT_THAT` and the same goes for `EXPECT`.

Please use here something like `EXPECT_EQ(timerState.value(), iox::posix::TimerState::TICK)`;

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,225 @@
...
+    auto timerState = sut.wait(iox::posix::TimerCatchupPolicy::IMMEDIATE_TICK);
+
+    ASSERT_FALSE(timerState.has_error());
+}
+
+TEST_F(PeriodicTimer_test, ZeroINTERVALTest)
+{
+    Timer sut(0_s);
+
+    auto timerState = sut.wait(iox::posix::TimerCatchupPolicy::IMMEDIATE_TICK);
```
elfenpiff, line 67: Before accessing values of an expect please `ASSERT` that it really contains a value.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,225 @@
...
+    bool result = timerState.value().state == iox::posix::TimerState::TICK ? true : false;
+
+    ASSERT_TRUE(result);
+}
+
+TIMING_TEST_F(PeriodicTimer_test, DurationINTERVALTest, Repeat(5), [&] {
+    Timer sut(INTERVAL);
+
+    auto timeBeforeWait = sut.now().value();
```
elfenpiff, line 76: Accessing values of expected without checking if it really contains a value.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
```diff
@@ -0,0 +1,225 @@
...
+    auto timerState = sut.wait(iox::posix::TimerCatchupPolicy::IMMEDIATE_TICK);
+    auto timeAfterWait = sut.now().value();
```
elfenpiff, line 78: Accessing values of expected without checking if it really contains a value.

elfenpiff (changes requested):
Please cleanup the unit test for all occurrences where you access an expected value before checking if it contains a value.

Use `ASSERT_` and `EXPECT_` in a fitting manner.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
shankar-in: Yes modified to use the `EXPECT_EQ`

iceoryx_hoofs/test/moduletests/test_posix_periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes applied this change.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Sorry abt mishandling of the enum classes as normal enums. Modified this.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Yes. modified to check if the expected has errors.

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
shankar-in: Okay replaced `if` blocks with `switch`

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,188 @@
...
+                }
+            }
+
+            case iox::posix::TimerCatchupPolicy::HOLD_ON_DELAY:
+            {
+                auto timeDiff = currentTime.value() - m_timeForNextActivation; // Calculate the time delay
+                m_waitResult.state = iox::posix::TimerState::DELAY;
+                m_waitResult.timeDelay = timeDiff;
+                return cxx::success<iox::posix::WaitResult>(m_waitResult);
+            }
```
elBoberido, line 127: I think this should not return in the happy path, like explained in https://github.com/eclipse-iceoryx/iceoryx/pull/754#discussion_r622086949

@mossmaurice @elfenpiff what do you think?

elBoberido (commented):
Tests not yet reviewed

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,188 @@
...
+{
+    m_interval = interval;
+    start();
+}
+
+void PeriodicTimer::stop() noexcept
+{
+    auto semValue = m_waitSemaphore.getValue();
+    cxx::Ensures(!semValue.has_error());
+    if (semValue.value() == iox::posix::SEM_ACQUIRED)
```
elBoberido, line 54: Oh, btw. What do you think about naming this `WOULD_BLOCK`?

iceoryx_hoofs/source/posix_wrapper/periodic_timer.cpp
```diff
@@ -0,0 +1,188 @@
...
+                return cxx::success<iox::posix::WaitResult>(m_waitResult);
+            }
+
+            case iox::posix::TimerCatchupPolicy::SKIP_TO_NEXT_TICK:
+            {
+                auto delay = currentTime.value() - m_timeForNextActivation; // calculate the time delay
+                if (delay > m_interval)
+                {
+                    auto totalSlotToBeSkipped =
+                        delay.toMilliseconds() / m_interval.toMilliseconds(); // calculate the total slots missed
```
elfenpiff, line 100: I think it must be `auto totalSlotToBeSkipped = delay.toMilliseconds() / m_interval.toMilliseconds() + 1;`

Let's say delay is 1.5 and interval is 1. Then `totalSlotToBeSkipped == 1` and `m_timeForNextActivation += m_interval * 1` which would be a timepoint in the past.
When you then call `m_waitSemaphore.timedWait` you would immediately return and would have always `IMMEDIATE_TICK` behavior.

Maybe I misunderstood this here, but if not please fix it and add a tests. The best way of fixing this would be a test which shows that this is `IMMEDIATE_TICK` behavior and is red until this is fixed.

dkroenke:
@shankar-in Due to inactivity, we close this Pull-Request for now. There are several findings still pending.
Feel free to reopen if you want to continue.

FYI @elBoberido @elfenpiff

elBoberido:
@shankar-in with the spring cleaning, we remove the current implementation of the `PosixTimer` with #1376. You can of course continue with the clean re-implementation as you please.

shankaranarayanan-c:
> @shankar-in with the spring cleaning, we remove the current implementation of the `PosixTimer` with #1376. You can of course continue with the clean re-implementation as you please.

Thanks @elBoberido I would be likely taking over the left over work from my personal account as my interest.

elBoberido:
@shankar-in I'm closing this PR for now to reduce the number of open PRs. Feel free to open it again once you continue working on this. I'm looking forward to your contribution :)

shankaranarayanan-c:
> @shankar-in I'm closing this PR for now to reduce the number of open PRs. Feel free to open it again once you continue working on this. I'm looking forward to your contribution :)

Thanks @elBoberido I have been delaying this work for a long time. I will restart it soon.

Issue #779, opened 2021-05-07
Analyse remaining mutexes and remove any RouDi deadlock possibilities.


elfenpiff:
## Brief feature description

Since RouDi is our central daemon and the whole system is corrupted when it fails we can assume for now that RouDi with its inner logic never fails.
But we have to ensure that RouDi is not corrupted by any external instance like a publisher application which holds a lock and dies which would lead to a deadlock in RouDi. 

This means RouDi is allowed to block any arbitrary process including itself but no one is allowed to block RouDi.

1. When the publisher sends data it locks a mutex in `chunk_distributor.inl` which can be also locked by RouDi during the cleanup process. If the publisher dies while it holds the lock RouDi will deadlock.
2. The process manager is used in `roudi.cpp` with a smart lock. Is there a RouDi deadlock possibility.

## Detailed information

E.g. a list of pros and cons of different considerations and how iceoryx and its user could benefit from it

elBoberido:
Small addendum. It's quite improbable to get the deadlock since a subscriber would need to subscribe/unsubscribe after the publisher crashed when holding the lock and before RouDi cleans up the resources of the crashed application. It's much more probable that RouDi crashes when trying to cleanup the locked mutex.

PR #780, opened 2021-05-07
Iox #779 design document for subscription

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Relates to #779

### DIFF
--- a/doc/design/draft/subscription.md
+++ b/doc/design/draft/subscription.md
@@ -12,7 +12,7 @@ massively slow down (caused by deadlock discovery) or crash RouDi.
 
 ## Terminology
 
-- **ChunkHistory** the last N samples which the publishers of a service have
+- **ChunkHistory** the last N samples which the publishers of a topic have
    published
 - **HistorySize** the number of samples the subscriber will receive at most on
    subscription
@@ -35,7 +35,7 @@ application which holds a lock and dies which would lead to a deadlock in RouDi.
 
 ### Subscribing without concurrent send
 
-Let's assume a subscriber S would like to subscribe to a service which has two
+Let's assume a subscriber S would like to subscribe to a topic which has two
 publishers P1 and P2. Furthermore, we assume that P1 and P2 have send the
 following samples:
 
@@ -47,7 +47,7 @@ following samples:
 | P1        | 31        | D      | | P2        | 40        | I      |
 | P1        | 41        | E      | | P2        | 42        | J      |
 
-Let's assume the subscriber subscribes to the service.
+Let's assume the subscriber subscribes to the topic.
 
 #### Subscription with _HistorySize_ 0
 
@@ -101,7 +101,7 @@ requests always a _HistorySize_ of 2.
 All the listed cases should behave identical if there is more then one
 concurrent publish in progress.
 
-#### Publisher before _DeliveryTime_
+#### Subscription before _DeliveryTime_
 
 The subscriber cache contains two samples.
 
@@ -112,7 +112,7 @@ The subscriber cache contains two samples.
 
 The sample K will be delivered after successful subscription.
 
-#### Publisher after _DeliveryTime_
+#### Subscription after _DeliveryTime_
 
 The subscriber cache contains two samples.
 
@@ -125,9 +125,9 @@ The sample K will not be delivered by the publisher since the _DeliveryTime_
 is defined as exactly the timepoint after the sample is placed into the ChunkHistory
 and the subscriber cache is based on that ChunkHistory.
 
-### Subscribing After Publisher Stops Offering
+### Subscription After Publisher Stops Offering
 
-In this case we assume that publisher P1 has stopped offering the service
+In this case we assume that publisher P1 has stopped offering the topic
 before the subscription was requested.
 
 #### Subscription with _HistorySize_ 4
@@ -152,12 +152,47 @@ The subscriber cache contains the four samples ordered after _DeliveryTime_.
     and lock-free manner.
 - A publisher delivers a sample by writing that sample into the shared ring-buffer.
 - Every subscriber owns an index to their own read position of the ring-buffer.
-- On subscription the subscriber will acquire a const reference to the ring-buffer.
+- On subscription the subscriber will acquire a const relocatable pointer to the ring-buffer.
 - When subscriber acquires a sample with take, the shared-chunk is copied from
     the index position and the index is incremented.
 - When the publisher delivers a sample it will verify all subscriber indices.
    If a subscriber `index > ring-buffer-head - subscriberHistorySize` the index
    is incremented.
+
+#### Alternative Ring Buffer Idea
+
+The history ring buffer interface could look like:
+```cpp
+template<typename T, uint64_t Capacity>
+class RingBuffer {
+  public:
+    struct ReadResult {
+      uint64_t currentPushCounter
+      cxx::vector<T, Capacity> contents;
+    };
+
+    ReadResult read(const uint64_t numberOfSamples, const uint64_t lastPushCounter);
+
+    // another idea would to state readOldest or readNewest
+    ReadResult readOldest(const uint64_t numberOfSamples, const uint64_t lastPushCounter);
+    ReadResult readNewest(const uint64_t numberOfSamples, const uint64_t lastPushCounter);
+    // idea end
+
+    cxx::optional<T> push(const T & value);
+};
+```
+
+The publisher would use it as before and just calls `push` for every sample it delivers.
+With every push the `uint64_t` push counter is incremented (it is something similar to
+an aba counter).
+When some subscriber would like to read something it calls `read` and provides the
+number of samples it would like to read and the last known push counter. Thanks to the
+last known push counter the RingBuffer knows exactly which samples weren't read yet and
+it can return the oldest `numberOfSamples` which are closest to the `lastPushCounter`.
+Additionally, it would return the `pushCounter` of the next sample which comes directly
+after the last sample the reader just read.
+
+In the first approach we can ensure the thread safety with `smart_lock`.
 
 #### Open Points
 
@@ -168,7 +203,7 @@ The subscriber cache contains the four samples ordered after _DeliveryTime_.
   the size has to be defined via some service subscription options.
 - Who owns the ring-buffer? Since it is shared between multiple publishers and
   a subscriber should be able to read samples after all publisher have stopped
-  offering the service, the ring-buffer should be owned by the service?
+  offering the topic, the ring-buffer should be owned by the topic?
 
 ### Considerations
 
--- a/doc/design/draft/subscription.md
+++ b/doc/design/draft/subscription.md
@@ -55,7 +55,7 @@ The subscriber will not be provided with samples from the history.
 
 #### Subscription with _HistorySize_ 1
 
-The subscriber cache contains the one sample ordered after _DeliveryTime_.
+The subscriber cache contains the one sample ordered by _DeliveryTime_.
 
 | Publisher | Timepoint | Sample |
 |:---------:|:---------:|:------:|
@@ -63,7 +63,7 @@ The subscriber cache contains the one sample ordered after _DeliveryTime_.
 
 #### Subscription with _HistorySize_ 6
 
-The subscriber cache contains the six samples ordered after _DeliveryTime_.
+The subscriber cache contains the six samples ordered by _DeliveryTime_.
 
 | Publisher | Timepoint | Sample |
 |:---------:|:---------:|:------:|
@@ -121,8 +121,9 @@ The subscriber cache contains two samples.
 | P1        | 51        | K      |
 | P2        | 42        | J      |
 
-The sample K will not be delivered by the publisher since it is already
-present in the cache.
+The sample K will not be delivered by the publisher since the _DeliveryTime_
+is defined as exactly the timepoint after the sample is placed into the ChunkHistory
+and the subscriber cache is based on that ChunkHistory.
 
 ### Subscribing After Publisher Stops Offering
 
@@ -139,6 +140,8 @@ The subscriber cache contains the four samples ordered after _DeliveryTime_.
 | P1        | 41        | E      |
 | P2        | 40        | I      |
 | P1        | 31        | D      |
+
+
 
 ## Design
 
@@ -159,6 +162,8 @@ The subscriber cache contains the four samples ordered after _DeliveryTime_.
 #### Open Points
 
 - Many chunks will not be released even when no susbcriber is using them.
+  - This might be solvable by having the ring-buffer size as the maximum of
+    history size and subscriber queue capacity?!
 - ring-buffer size? Since the ring-buffer is shared between multiple publishers
   the size has to be defined via some service subscription options.
 - Who owns the ring-buffer? Since it is shared between multiple publishers and
--- a/doc/design/subscription.md
+++ b/doc/design/subscription.md
@@ -124,7 +124,46 @@ The subscriber cache contains two samples.
 The sample K will not be delivered by the publisher since it is already
 present in the cache.
 
+### Subscribing After Publisher Stops Offering
+
+In this case we assume that publisher P1 has stopped offering the service
+before the subscription was requested.
+
+#### Subscription with _HistorySize_ 4
+
+The subscriber cache contains the four samples ordered after _DeliveryTime_.
+
+| Publisher | Timepoint | Sample |
+|:---------:|:---------:|:------:|
+| P2        | 42        | J      |
+| P1        | 41        | E      |
+| P2        | 40        | I      |
+| P1        | 31        | D      |
+
 ## Design
+
+### Idea 1: One Ring Buffer Which Contains The Samples
+
+- All publishers are sharing one random access lock-free ring-buffer.
+- The ring-buffer is filled with shared-chunks which can be copied in a threadsafe
+    and lock-free manner.
+- A publisher delivers a sample by writing that sample into the shared ring-buffer.
+- Every subscriber owns an index to their own read position of the ring-buffer.
+- On subscription the subscriber will acquire a const reference to the ring-buffer.
+- When subscriber acquires a sample with take, the shared-chunk is copied from
+    the index position and the index is incremented.
+- When the publisher delivers a sample it will verify all subscriber indices.
+   If a subscriber `index > ring-buffer-head - subscriberHistorySize` the index
+   is incremented.
+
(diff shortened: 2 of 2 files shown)


elBoberido (commented):
I don't think this should be placed in `design` but rather in `design/draft` or `design/proposals` since it might take some time to implement his, if ever.

doc/design/subscription.md
```diff
@@ -0,0 +1,175 @@
...
+present in the cache.
+
+### Subscribing After Publisher Stops Offering
+
+In this case we assume that publisher P1 has stopped offering the service
+before the subscription was requested.
+
+#### Subscription with _HistorySize_ 4
+
+The subscriber cache contains the four samples ordered after _DeliveryTime_.
```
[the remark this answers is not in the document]
elfenpiff: Yes, the subscriber is not subscribing to a publisher it is subscribing to a service. And the user requested the last 4 samples send in the name of that service.

Let's say you request a history of size two and the last four samples would be A, B, C and D. And now the publisher stops offering (maybe even not willingly, maybe something unexpected happened), and now you receive a history with holes like A and D.
This is quite error prone and not robust and as a user I would really wish to get the last 4 samples from the service and not depending on what publisher did send what.

Also the name history size would then be misleading since it is not the actually history ... it is again the history which was written by the winners. Again! ... ;)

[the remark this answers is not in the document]
MatthiasKillat: Yes, it probably has to be done at the reader side or at service level. Not only because offering may be stopped, but also because we could have multiple publishers for the same topic. Which data shall we receive then? A cache at the reader side would handle this but require needless replication for all readers. This is also bad since in the order of elements will differ between readers due to the nature of concurrent publishing.

A centralized cache for the topic would be the solution, but then we would a multi-writer multi-reader cache (the MPMC FIFO is not sufficient as pop is destructive). For now I can only offer a relatively easy to implement single writer multi reader cache as in https://github.com/eclipse-iceoryx/iceoryx/issues/857 which would solve the problem if the cache exists per problem (as it currently is).

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+The goal is to craft and implement a design which is unable to deadlock,
+massively slow down (caused by deadlock discovery) or crash RouDi.
+
+## Terminology
+
+- **ChunkHistory** the last N samples which the publishers of a service have
+   published
+- **HistorySize** the number of samples the subscriber will receive at most on
+   subscription
+- **DeliveryTime** the point in time where the publisher writes the samples in
```
MatthiasKillat, line 19: It is important to note that time points are only relevant if they are monotonic and, depending on clock synchronization, may be meaningless across process boundaries.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+
+Let's assume the subscriber subscribes to the service.
+
+#### Subscription with _HistorySize_ 0
+
+The subscriber will not be provided with samples from the history.
+
+#### Subscription with _HistorySize_ 1
+
+The subscriber cache contains the one sample ordered after _DeliveryTime_.
```
MatthiasKillat, line 58: ordered by?

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+
+| Publisher | Timepoint | Sample |
+|:---------:|:---------:|:------:|
+| P2        | 42        | J      |
+
+#### Subscription with _HistorySize_ 6
+
+The subscriber cache contains the six samples ordered after _DeliveryTime_.
```
MatthiasKillat, line 66: We cannot reason about order by delivery time unless the clocks are synchronized. Essentially the MPMC queue used to deliver the data will generate (define) the ordering, but it is not necessarily the one defined by time points. 

The only guarantee we have is that it respects the partial orders of both individual publishers.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+
+The subscriber cache contains two samples.
+
+| Publisher | Timepoint | Sample |
+|:---------:|:---------:|:------:|
+| P1        | 51        | K      |
+| P2        | 42        | J      |
+
+The sample K will not be delivered by the publisher since the _DeliveryTime_
+is defined as exactly the timepoint after the sample is placed into the ChunkHistory
```
MatthiasKillat, line 125: There is no such timepoint **exactly** after some observable event happens. In the distributed setting something happens before something else without meaningful timepoints or the order of two events is indeterminate (and not observed). We can define timepoints for the time some event became observable, but care must be taken.

There is no consistent way to define the point after something happens, unless we have a global synchronized logical clock (then it may be define as the next tick).

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+| P2        | 40        | I      |
+| P1        | 31        | D      |
+
+
+
+## Design
+
+### Idea 1: One Ring Buffer Which Contains The Samples
+
+- All publishers are sharing one random access lock-free ring-buffer.
```
MatthiasKillat, line 150: This would essentially require a muli write, multi read lock-free buffer. Possible, but more difficult then the single-write multi read buffer I propose in https://github.com/eclipse-iceoryx/iceoryx/issues/857. This is also quite a bit further away from our current archtitecture since we need such a buffer per service/topic and not per publisher then. 

I propose we start in small steps and properly implement the buffering in our current approach - buffering on publisher side. Then we can move to the more complicated solution where we buffer across publishers in a centralized manner per topic.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+- The ring-buffer is filled with shared-chunks which can be copied in a threadsafe
+    and lock-free manner.
+- A publisher delivers a sample by writing that sample into the shared ring-buffer.
+- Every subscriber owns an index to their own read position of the ring-buffer.
```
MatthiasKillat, line 154: Probably not necessary if we extend the idea I have in mind for  #857, but we will see this in a prototype implementation. Any subscriber can read randomly from the cache. It is the concurrent writing (concurrency between writes) which complicates things, I need to explore that.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+- On subscription the subscriber will acquire a const reference to the ring-buffer.
+- When subscriber acquires a sample with take, the shared-chunk is copied from
+    the index position and the index is incremented.
+- When the publisher delivers a sample it will verify all subscriber indices.
```
MatthiasKillat, line 158: I would see the responsibility here at subscriber side to know which data it wants to extract from the cache (again the overflow behavior and other issues).
Ideally each data item gets a sequence number when it is placed in the cache (this also generats an ordering for all subscribers then!) and the subscriber can verify whether it has consecutive data. However, losing data is inevitable with any restricted capacity buffer approach. That is why I consider storing read indices for individual subscribers meaningless.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+   If a subscriber `index > ring-buffer-head - subscriberHistorySize` the index
+   is incremented.
+
+#### Open Points
+
+- Many chunks will not be released even when no susbcriber is using them.
+  - This might be solvable by having the ring-buffer size as the maximum of
+    history size and subscriber queue capacity?!
```
MatthiasKillat, line 166: Solved by FIFO behavior, but yes, the latest cache size elements will be kept. This is in the nature of things and inevitable. This is also the case now with the history kept at a publisher, so no regression.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+- ring-buffer size? Since the ring-buffer is shared between multiple publishers
```
MatthiasKillat, line 167: Difficult. Theoretically it must have the size of the maximum of any currently requested history or offering publisher. This is not known in a totally dynamic world and will in addition require the buffer to always be at some maximum size (wasting shared memory if chunks are cached) or require it to dynamically grow (algorithmically difficult in a lock-free setting). 
If it is defined by the service options it would be better since then it is somewhat agreed on by all parties offering the service? Hard to say what is reasonable, usable and implementable.

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+  the size has to be defined via some service subscription options.
+- Who owns the ring-buffer? Since it is shared between multiple publishers and
+  a subscriber should be able to read samples after all publisher have stopped
+  offering the service, the ring-buffer should be owned by the service?
```
MatthiasKillat, line 171: Yes, or in particular it is associated with it by Roudi and managed. It would require considerable changes in the way Roudi manages services and its metadata.

This is why I am in favor of getting the caching right in small steps, beginning on publisher level. I beleive the insights from that can be transferred to the more general case as well.

doc/design/draft/subscription.md
MatthiasKillat: Addendum: the point of caching is to provide samples to late joining subscribers. Hence they are not known before they subscribe and we cannot keep any index specifically for them for this reason. Once they join, the get the requested samples from the cache (if any). After this we could keep an index for a subscriber, but then we would use this structure instead of the MPMC queue, which would be vastly more difficult since it would require non-destructive reading for multiple readers. I do not know how to solve this and I presume it is harder than the MPMC queue (it is more general).

MatthiasKillat:
I agree that there are many problems, especially related to multiple publishers. I think they are solvable but difficult to solve in one single step, particularly with lock-free structures. I have the feeling that the most general needed lock-free structure needed for centralized (per service) caching would be a lock-free multi-writer multi-reader cache with ringbuffer semantics. This is presumably much harder to achieve than the MPMC queue.

In comparison a single writer multi reader cache is way easier and could serve as a stepping stone, still solve parts of a problem (proper caching on publisher side, partially last-lock). It would also not need to interfere with the current architecture.

doc/design/draft/subscription.md
elfenpiff: @MatthiasKillat here I had a simple `uint64_t` number in mind which will increase with every delivery and is stored in the shared memory. 

This would be a simplistic and efficient way to sort the samples after there delivery time - we do not require a real time format for that, this would fit all the requirements.
But in the end it feels like a sequence number - maybe we just define the sequence number that way and say that it corresponds with the delivery time.

doc/design/draft/subscription.md
elfenpiff: See comment: https://github.com/eclipse-iceoryx/iceoryx/pull/780/files#r654320386

Can be solved very easily with an incrementing number stored in the service description inside of the shared memory. We do not need clock synchronization.

doc/design/draft/subscription.md
elfenpiff: @MatthiasKillat If we put the sample into the ChunkHistory within a mutex and set the timestamp in that mutex scope as well we should have the exact timepoint. The global synchronized logical clock would be the incremented number in comment: https://github.com/eclipse-iceoryx/iceoryx/pull/780/files#r654320386

doc/design/draft/subscription.md
elfenpiff: @MatthiasKillat  you are right, but here we just wanted to collect ideas how the final solution could look like since we need something where we can converge to when we start with small steps.

elfenpiff:
> I agree that there are many problems, especially related to multiple publishers. I think they are solvable but difficult to solve in one single step, particularly with lock-free structures. I have the feeling that the most general lock-free structure needed for centralized (per service) caching would be a lock-free multi-writer multi-reader cache with ringbuffer semantics. This is presumably much harder to achieve than the MPMC queue.
> 
> In comparison a single writer multi reader cache is way easier and could serve as a stepping stone, still solves parts of a problem (proper caching on publisher side, partially last-lock). It would also not need to interfere with the current architecture.

I completely agree but I think I missed to mention the idea behind this pull request - we just wanted to collect some ideas and theoretical solutions so that we have a starting point. We have to go multiple steps but first we have to find out what we can actually realize since some steps require a MPMC random access ring buffer.
If we are unable to implement that we have to explore different approaches.

dkroenke (approved):

elBoberido:
I would vote to move this into discussions instead of merging it. There are still many open points which needs to be addressed and I fear those will be forgotten once this is merged. From the intention of the PR, this should have been a discussion in the first place

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+### Subscribing with concurrent send
+
+Let's assume publisher P1 would like to deliver the sample K at timepoint
+52 while the subscriber requested subsription. Furthermore, the subscriber
+requests always a _HistorySize_ of 2.
+
+All the listed cases should behave identical if there is more then one
+concurrent publish in progress.
+
+#### Publisher before _DeliveryTime_
```
budrus, line 104: Hmm, should you write "Subscription before _DeliveryTime_" ?

doc/design/draft/subscription.md
budrus: Should work if you have a mutex like today or some other kind of atomic operations that define a "happened before"

doc/design/draft/subscription.md
```diff
@@ -0,0 +1,180 @@
...
+    and lock-free manner.
+- A publisher delivers a sample by writing that sample into the shared ring-buffer.
+- Every subscriber owns an index to their own read position of the ring-buffer.
+- On subscription the subscriber will acquire a const reference to the ring-buffer.
+- When subscriber acquires a sample with take, the shared-chunk is copied from
+    the index position and the index is incremented.
+- When the publisher delivers a sample it will verify all subscriber indices.
+   If a subscriber `index > ring-buffer-head - subscriberHistorySize` the index
+   is incremented.
+
```
budrus, line 161: @elfenpiff So the subscribers would not have own queues anymore? Is your ringbuffer the history for all subscribers and with the different subscriber read indices also the queue for all subscribers?

budrus:
@elfenpiff @MatthiasKillat @elBoberido Let me try to summarize to get an old man on your page again

- History feature with multiple publishers is kind of broken 
- Consistently updating the history and the subscriber list in the publisher is still one of our hardest problem to get lock-free and is currently protected by a mutex
-  We could fix the history feature by having one history per topic(service). But then the lock would also affect all the publishers, or?  Would this be possible if we ignore the lock for now?
- The proposed design would be a complete new approach as the history and all the subscriber queues would end up in one MPMC ring-buffer. This would be super lightweight but also super complex as it has to handle all the history and queue settings with overflow etc. 
- What would be a possible next step that improves the status quo but doesn't need to solve this lock-free ring-buffer challenge?

elfenpiff:
@budrus 

> * History feature with multiple publishers is kind of broken

It is broken in a multi publisher case. You will loose some parts of the history when a publisher calls `stopOffer` before you are subscribed. Furthermore, you get only the history of one random publisher and with that you may not get the last N samples but some older ones.

> * Consistently updating the history and the subscriber list in the publisher is still one of our hardest problem to get lock-free and is currently protected by a mutex

Here the idea could be that the subscriber list is not part of the publisher anymore it is part of the service.

> * We could fix the history feature by having one history per topic(service). But then the lock would also affect all the publishers, or?  Would this be possible if we ignore the lock for now?

Here an idea was to have a lock free queue for every service and where the subscriber can acquire more then just one sample with `pop`. Maybe a method like `read(int numberOfElements)` where the subscriber can read the amount of samples it requires from the history.

> * The proposed design would be a complete new approach as the history and all the subscriber queues would end up in one MPMC ring-buffer. This would be super lightweight but also super complex as it has to handle all the history and queue settings with overflow etc.

This is correct. In the final implementation we would have a random access MPMC ring-buffer. Here it would make sense that the service itself has a maximum history size and the subscriber can reduce it by just reading less from that commonly shared ringbuffer.

> * What would be a possible next step that improves the status quo but doesn't need to solve this lock-free ring-buffer challenge?

For the moment I would say we use the simplest approach and use a mutex. Furthermore, we extend our understanding of a service. A service then could be a internal constructs which contains exactly one history container for all publishers and a list which contains all subscribers subscribed to that service. A todo list could then look like this:

1. Introduce topics/services in roudi as extra class which is created when a publisher calls offer and it does not already exist.
2. Move the history container from the publishers into that topic/service class. The publisher just holds a relocatable pointer to the history and calls push for every sample it delivers.
3. Create a ring buffer (not threadsafe) which offers an overflowing `push` like sofi and `read` of N samples without removing them. We use the `smart_lock` to make it threadsafe.
4. Remove the delivery queue from the subscribers. Every subscriber holds a relocatable pointer to the history container (3) and reads the samples it requires. The subscriber list in the publishers can be removed since the publisher just delivers one sample once to the shared MPMC ring buffer.

In this first approach I would use the containers we already have in combination with the `smart_lock` to make them threadsafe. When this is done we can create lock free approaches step by step.
Maybe we can extend our current lock-free MPMC ring buffer to support random access or to read N samples in queue order withour removing them.

elfenpiff:
@budrus @elBoberido @MatthiasKillat 

The history ring buffer interface could look like:
```cpp
template<typename T, uint64_t Capacity>
class RingBuffer {
  public:
    struct ReadResult {
      uint64_t currentPushCounter
      cxx::vector<T, Capacity> contents;
    };

    ReadResult read(const uint64_t numberOfSamples, const uint64_t lastPushCounter);
    cxx::optional<T> push(const T & value);
};
```

The publisher would use it as before and just calls `push` for every sample it delivers. With every push the `uint64_t` push counter is incremented (it is something similar to an aba counter).
When some subscriber would like to read something it calls `read` and provides the number of samples it would like to read and the last known push counter. Thanks to the last known push counter the RingBuffer knows exactly which samples weren't read yet and it can return the oldest `numberOfSamples` which are closest to the `lastPushCounter`. Additionally, it would return the `pushCounter` of the next sample which comes directly after the last sample the reader just read.

In the first approach we can ensure the thread safety with `smart_lock`.

doc/design/draft/subscription.md
elfenpiff: Correct!

elBoberido:
@elfenpiff can we close this. It already smells ;)

elfenpiff:
@elBoberido I would prefer to merge this into the `doc/design/draft` folder. In my opinion it has still some value and may become handy when we start refactoring iceoryx this year.

But I have no strong feelings about this.

elBoberido:
@elfenpiff I think there is still too much discussion ongoing and `doc/design/draft` is more for stuff which is settled and anyone could start with implementation.

Issue #415, opened 2020-12-08
service registry as a built-in topic


budrus:
## Brief feature description

The discovery requests from user side are currently transferred via the message queue connection. This has some drawbacks

- The message queue is not only used for registration and creation of resources during startup but also during runtime
- If an application polls the discovery information with a high frequency the message queue becomes a bottleneck 
- We have a message size limitation on the message queue or unix domain socket, i.e. we need some transport protocol to split the response to several messages if there are a lot services. Currently it is a limitation as we do not have such a protocol 
- The `findService()` is very close to AUTOSAR ara::com and maybe cannot be used in a reasonable way in other frameworks 
- If someone wants to use ara::com service discovery callbacks, having only a polling based interface is a big drawback 

## Detailed information

The proposal is to have a service registry that provides it's content as a built-in topic, similar to the introspection topics. Then a subscriber could be created to get the information either in a polling way or event-driven by using the subscriber in combination with a waitset. The service registry publisher would use a history with size 1. Then every subscriber gets the current state on subscribe if desired. By providing the full CaPro ID information and not only service and instance IDs, this built-in topic could also be used for non AUTOSAR use cases. With this approach having a large service registry is fine as we don't have to send the content over the message queue and it's distributed with iceoryx' finest true zero-copy. I guess we can then also get rid of `getServiceRegistryChangeCounter()` which smells like workaround

## Todo
- [x] Create design decision document
- [x] Introduce poor mans `cxx::map` (Will be done in #829)
- [x] Refactor `ServiceDescription`
  - [x] Remove the two argument c'tor in `ServiceDescription`
  - [x] Remove three integers in `ServiceDescription`
  - [x] ~~Forbid creating Pub/Sub with `ServiceDescription(AnyServiceString, AnyInstanceString, "foo")` and forbid adding such a invalid `ServiceDescription` to the registry~~
  - [x] ~~Add [ServiceDescriptionBuilder](https://github.com/eclipse-iceoryx/iceoryx/pull/1047#discussion_r792652416)~~ Moved to #1141
- [x] Refactor `ServiceRegistry` to use the whole `ServiceDescription`
- [x] Error handling in `ServiceDescription(const cxx::Serialization& serial)` like `client_options.cpp`
- [x] Create example for new user-facing `DiscoveryInfo` class and move `getServiceRegistryChangeCounter`, `findService`, ~~`offerService` and `stopOfferService`~~ to this class 
  - [x] Extend `icediscovery`
[...]

Issue #611, opened 2021-03-15
Refactor Shared Memory Object Creation


mossmaurice:
## Brief feature description

Both `PoshRuntime` and `ProcessManager` don't follow our main design principles (RAII, clear responsibilty, etc.). Hence, a refactoring is necessary.

## Detailed information

- [ ] Create design document
- [ ] Agree on design
- [ ] Implement design

elBoberido:
Renaming `ProcessName_t` to `PoshRuntimeName_t` will be an API breaking change after 1.0. If you want to do it, you need to do it before the release.

PR #784, opened 2021-05-10
Iox #415 Add design decision document for service discovery

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Add design decision document on service discovery
  * Add sequence diagram for alternative b and d
  * Add component diagram for alternative b and d

## References

- Closes partly #415


doc/website/getting-started/overview.md
```diff
@@ -183,11 +183,12 @@ The service model of iceoryx is derived from AUTOSAR and is still used in the AP
...
-|---------|---------|------------------|------------------------|
-| ROS2    | Type    | Namespace::Topic | -                      |
-| AUTOSAR | Service | Instance         | Event                  |
-| DDS     | -       | -                | /Group/Instance/Topic  |
+|                                                                                   | Group   | Instance         | Topic                  |
+|-----------------------------------------------------------------------------------|---------|------------------|------------------------|
+| [rmw_iceoryx](https://github.com/ros2/rmw_iceoryx/)                               | Type    | Namespace::Topic | -                      |
+| AUTOSAR                                                                           | Service | Instance         | Event                  |
+| [DDS Gateway](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_dds) | -       | -                | /Group/Instance/Topic  |
+| [Cyclone DDS](https://github.com/ros2/rmw_cyclonedds/pull/303/files)              | -       | Type Name        | Topic Name             |
```
mossmaurice, line 191: @budrus @FerdinandSpitzschnueffler 

IMHO we should unify the mapping of DDS gateway, Cyclone and rmw as they are basically all just DDS. However, I'm unsure about the consequences if we adapt the mapping e.g. in rmw. @budrus Can you elaborate those?

mossmaurice: @jeising I just had a look at `rmw_iceoryx` and found out that you created [the mapping](https://github.com/ros2/rmw_iceoryx/blob/foxy/rmw_iceoryx_cpp/src/internal/iceoryx_name_conversion.cpp#L102-L147). Do you have details on why you chose this mapping? I'm currently in favor to go for the mapping that Cyclone DDS choose (Instance = Type Name, Event = Topic Name).

Side question: Do you fancy a hackathon in the next weeks updating rmw_iceoryx with iceoryx v1.0? :sunglasses: There is already a [PR](https://github.com/ros2/rmw_iceoryx/pull/42) that we could finish off.

jeising: Let's have a look into what would be a good mapping. I guess there were multiple options. It should allow to map native iceoryx msgs as well as native ros msgs. Looking forward to the hackathon!

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Re-use building blocks
+
+Con:
+
+* Not needed by all users per default
+* Polling leads to overhead
+
+### Decision
+
+* Sychronous, one-shot RPC (Polling): Alternative A
```
elfenpiff, line 163: How did you come to this decision because I do not agree?
I would use for both push/pull alternative A which would spare us a lot of code.

The user just must be able to attach the service registry to a listener/waitset and we are fine, which can maybe abstracted away in the posh runtime.

doc/website/getting-started/overview.md
```diff
@@ -183,11 +183,12 @@ The service model of iceoryx is derived from AUTOSAR and is still used in the AP
 
 The following table gives an overview of the different terminologies and the current mapping:
 
-|         | Group   | Instance         | Topic                  |
-|---------|---------|------------------|------------------------|
-| ROS 2   | Type    | Namespace::Topic | -                      |
-| AUTOSAR | Service | Instance         | Event                  |
-| DDS     | -       | -                | /Group/Instance/Topic  |
+|                                                                                   | Group   | Instance         | Topic                  |
```
elfenpiff, line 186: How is this file related to the design decision document? Doesn't it violate our traceability rule when we change files which are unrelated to the issue?!

iceoryx_examples/callbacks/README.md
```diff
@@ -197,21 +197,22 @@ Afterward, we reset both caches to start fresh again.
 ### Additional context data for callbacks (ice_callbacks_listener_as_class_member.cpp)
```
elfenpiff, line 197: How is this file related to the design decision document? Doesn't it violate our traceability rule when we change files which are unrelated to the issue?!

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/interface_port_data.hpp
```diff
@@ -31,6 +31,7 @@ struct InterfacePortData : public BasePortData
     InterfacePortData() = default;
     InterfacePortData(const RuntimeName_t& runtimeName, const capro::Interfaces interface) noexcept;
```
elfenpiff, line 32: How is this file related to the design decision document? Doesn't it violate our traceability rule when we change files which are unrelated to the issue?!

iceoryx_posh/source/popo/ports/interface_port.cpp
```diff
@@ -38,6 +38,7 @@ void InterfacePort::dispatchCaProMessage(const capro::CaproMessage& caProMessage
         // information loss for this interface port
```
elfenpiff, line 38: How is this file related to the design decision document? Doesn't it violate our traceability rule when we change files which are unrelated to the issue?!

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/interface_port_data.hpp
```diff
@@ -31,6 +31,7 @@ struct InterfacePortData : public BasePortData
...
 
+    /// @todo put condition variable data here
```
elfenpiff, line 34: Please no todos in the code.

iceoryx_posh/source/popo/ports/interface_port.cpp
```diff
@@ -38,6 +38,7 @@ void InterfacePort::dispatchCaProMessage(const capro::CaproMessage& caProMessage
...
         errorHandler(Error::kPOSH__INTERFACEPORT_CAPRO_MESSAGE_DISMISSED, nullptr, ErrorLevel::SEVERE);
     }
+    /// @todo notify the condition variable here
```
elfenpiff, line 41: Please no todos in the code.

elfenpiff (changes requested):

elfenpiff (commented):
Could you please create a separate issue and adjust the commit iox number for those commit messages which are touching files which are totally unrelated to this issue.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Data transfer over shared memory
+  * Basically putting Service Registry to shared memory so everyone can access it
+
+Pro:
+
+* Fast data transfer
+
+Contra:
+
+* More complex than todays solution
```
elfenpiff, line 115: I disagree, in the end it would be much simpler than todays solution. Since this requires much less code, loops etc. It would also simplify and reduce the user code since the interfaceports would be gone.

iceoryx_examples/callbacks/README.md
mossmaurice: Just install a markdown linter and everything will be fine :stuck_out_tongue: 

IMHO trailing white spaces can be removed as a drive-by-refactoring. I can also move this to another issue.

You'll get a free hug :hugs: if you install a markdown linter!

elfenpiff: I did it already some weeks ago. Can I get a :cookie: instead?

doc/design/diagrams/service-discovery-overview.puml
```diff
@@ -0,0 +1,38 @@
+@startuml
+title Alternative B: iceoryx-based ECU forwarding CaPro messages to Gateway and PoshRuntime
+
+package "iceoryx ECU" {
```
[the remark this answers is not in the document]
mossmaurice: Used the term "iceoryx machine"

doc/design/diagrams/service-discovery-overview.puml
```diff
@@ -0,0 +1,38 @@
...
+    component RouDi {
+        artifact discoveryLoop [
+        PortManager::sendToAllMatchingInterfacePorts()
+        ]
+        artifact server [
```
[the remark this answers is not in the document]
mossmaurice: I took the naming and the API from #27. Please ping @elBoberido about the naming for the request/ response feature.

doc/design/diagrams/service-discovery-sequence-diagram.puml
```diff
@@ -0,0 +1,33 @@
+@startuml
+title Alternative B: Event-based notification between RouDi and PoshRuntime
```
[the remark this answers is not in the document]
mossmaurice: Nope, Alternative A does not have any diagrams.

`doc/design/diagrams/service-discovery-overview.puml`: Contains both polling and event-based notification
`doc/design/diagrams/service-discovery-sequence-diagram.puml`: Only contains event-based notification

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Polling and event notification approach shall both be possible
+
+### Considerations and use-cases
+
+#### AUTOSAR Adaptive
+
+* findService() call as implemented in Almond v1.0
+* ara::com service discovery callbacks
+
+1. `StartFindService()`: Register callback, std::function interface
```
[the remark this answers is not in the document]
mossmaurice: It has this signature: `std::function<void(ServiceContainer<T>,ServiceHandle)>`

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+* Static discovery shall be supported
+  * Having RouDi read in a config file, which has
+    * All the topics listed
+    * Wiring pub/sub information
+  * During init phase pub/sub are connected
+* Create abstract interface for `ServiceRegistry`
+  * Add new class to `StaticServiceRegistry::StaticServiceRegistry(std::map<CaproIdString_t, instance_t> map)`
+    * No notification-based callback possibilty
+    * Just static, polling with `FindService()` possible
```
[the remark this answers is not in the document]
mossmaurice: After a discussion with @budrus we came to the conclusion, that the interface of the `PoshRuntime` should be the same for both static and dynamic discovery. In the static discovery case, `FindService()` would always return the same information and event-based notification would be called just **once** and return the static registry. No user code or binding code should be adapted if you e.g. start developing with dynamic discovery and then before SOP switch to static discovery. We need to properly document the feature so users know how it behaves.

What will be split is RouDi and some of its classes, see #539

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+* `dds_find_topic_locally` function is available for users
+
+#### Static discovery
+
+* Static discovery shall be supported
+  * Having RouDi read in a config file, which has
+    * All the topics listed
+    * Wiring pub/sub information
+  * During init phase pub/sub are connected
+* Create abstract interface for `ServiceRegistry`
```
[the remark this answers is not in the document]
mossmaurice: Neat idea, I like it!

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+  * Topic key
+  * QoS settings
+* `dds_find_topic_locally` function is available for users
+
+#### Static discovery
+
+* Static discovery shall be supported
+  * Having RouDi read in a config file, which has
+    * All the topics listed
+    * Wiring pub/sub information
```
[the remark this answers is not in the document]
mossmaurice: > Does RouDi create all the publishers/subscribers during initialization and then just provides the fitting subscriber/publisher to the application?

Without having thought too much about the details: Yep, what you describe is what I have in my head regarding the static shared memory feature.

> Or are the services just registered in the service registry and not really present? If so what happens when the subscriber assumes that a publisher is present but it is not and therefore cannot subscribe to it?

What you describe here is the scenario where an application is not yet running but the static discovery returns information about pub/subs of this runtime, right? Instead of creating a new interface / API here, I would opt for a `{stop}offer` and `{un}subscribe` that are no-ops in the static case. Same [reason](https://github.com/eclipse-iceoryx/iceoryx/pull/784#discussion_r633514476) as before. I would say it's not misuse, but instead a feature, which allows the user a smooth transition from the dynamic world to a static world.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+Stick with IPC channel
+
+Pro:
+
+* No code changes needed
+
+Con:
+
+* No userspace solution
+  * IPC channel uses OS functions
```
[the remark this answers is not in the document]
mossmaurice: Sure our shared memory classes use OS functions, but not during runtime. Context switches between user- and kernel space. Clarified.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Upper frame limit may lead to segregation into several frames
```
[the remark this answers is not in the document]
mossmaurice: Well, generally speaking for network transmission protocol et al.: Splitting up a large message into several small ones, will slow things down (aka decreasing the MTU). Regarding your copy point: In my understanding a RPC call can never be 1:n, but will always be 1:1, that's the nature of an RPC call.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+#### Event-based notification
+
+##### Alternative A: Shared memory + IPC channel
+
+* Send ID's + ABA-counter/ServiceDiscoveryChangeCounter over IPC channel
+* Data transfer over shared memory
+  * Basically putting Service Registry to shared memory so everyone can access it
```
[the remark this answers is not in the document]
mossmaurice: IMHO read only should be enough.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+Pro:
+
+* Fast data transfer
+
+Contra:
+
+* More complex than todays solution
+* In an optimal case IPC channel communication shall only be used for startup
```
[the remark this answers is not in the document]
mossmaurice: Which comment do you mean? Could you please link it. 

IMHO the IPC channel shall only be used to establish the connection with RouDi and after that all communication should be shared memory based in a perfect world. That's what the two of us also discussed sometime ago in #611 
Being pragmatic the `CREATE_{SUBSCRIBER,PUBLISHER,..}` messages will stay around for quite some time, as I don't see them becoming a bottleneck anytime soon.

doc/design/service-discovery.md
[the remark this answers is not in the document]
mossmaurice: Added

iceoryx_examples/callbacks/README.md
mossmaurice: @mossmaurice passed @elfenpiff :cookie: and a :cocktail:

mossmaurice: Great you volunteered to create a PR that fixes all the trailing white-spaces in `.md` :smiley:

doc/website/getting-started/overview.md
mossmaurice: Nope, this is actually the first step before any implementation can start. We need to consolidate the different mappings as ROS 2, rmw_iceoryx and the DDS gateway in the end all use the same technology and I see no reason to treat them differently. The changes so far only update what is currently implemented.

iceoryx_posh/source/popo/ports/interface_port.cpp
mossmaurice: This is a draft PR and meant to show you what kind of code changes are planned

iceoryx_posh/source/popo/ports/interface_port.cpp
mossmaurice: See [this](https://github.com/eclipse-iceoryx/iceoryx/pull/784#discussion_r633843244) comment.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/interface_port_data.hpp
mossmaurice: See [this](https://github.com/eclipse-iceoryx/iceoryx/pull/784#discussion_r633843244) comment.

doc/website/getting-started/overview.md
budrus: @mossmaurice The mapping in rmw_iceoryx is based on a mapping from Autosar Adaptive to ROS. I.e. how to convert a service/instance/event into a topic/namespace/type and vice versa. We had questions like "what will be shown as type in ros_topc list if we have a publisher that was created with ara::com API" This is different from what we do in Cyclone where we just do a mapping of DDS topic and type to the three CaPro IDs without any consideration of an AUTOSAR world. If you just adapt rmw to the another simple mapping these mixing of worlds is no more considered. I think @jeising wanted to have the possibility to debug ara based applications with ROS tools

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+Pro:
+
+* Re-use building blocks to avoid code duplication
+* Dogfooding
+* Fast due to shared memory
+
+Con:
+
+* Shared memory approach might have overhead compared for a simple RPC call
```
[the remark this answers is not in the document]
budrus: @elfenpiff @mossmaurice @elBoberido: How about this
1. We introduce a Discovery class for the public API 
2. This is a wrapper around a subscriber on a service registry topic with queue size 1
3. The class has a get() method that will pop the latest element from the queue, copies the data in an internal member and returns it
4. If you call the get() method and there is no new element in the queue, there is no change and you get the last service registry content again
5. You can add this class to a waitset or listener like a normal subscriber
6. We don't need a lot of chunks as all the queues will always hold the latest one and if it is consumed, a copy is made 
7. We have the same Discovery class with the same information for polling and event-based access

doc/design/service-discovery.md
budrus, line 107: How to deal with concurrency and data consistency?

mossmaurice: Thread-safe and lock-free construct required. Added under "con".

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+* Extend `dispatchCaproMessage()` with condition variable and notification mechanism
+
+##### Alternative C: Custom thread
+
+* Create a custom thread in `PoshRuntime`, which polls for specific discovery information using request/response
+
+Pro:
+
+* More flexible
+* Re-use building blocks
```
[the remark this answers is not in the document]
mossmaurice: I meant the req/resp building block here. Adapted.

doc/design/service-discovery.md
mossmaurice: No final decision was done yet, I'll add this section once we've finished the discussion.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,261 @@
...
+* Similar to AUTOSAR Adaptive
+
+#### Cyclone DDS and ROS 2
+
+* Cyclone DDS calls this feature [topic discovery](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/topic_discovery.md). See also the [feature request](https://github.com/eclipse-cyclonedds/cyclonedds/issues/42).
+
+* Based on SDEP (simple endpoint discovery protocol)
+* `dds_ktopic`
+  * stores pointer to `ddsi_topic_definition`
+* Has built-in DCPSTopic topic and contains
```
mossmaurice, line 52: @dennis-adlink @eboasson

IIUC the `DCPSTopic` contains only the delta and not the full registry with all currently available topics?

doc/design/service-discovery.md
mossmaurice: I disagree, too. Writing a custom thread-safe and lock-free building block bespoke just for this use-case is complex task. In the other alternatives we would use well established patterns req/resp + Listner or `InterfacePort`s.

doc/design/service-discovery.md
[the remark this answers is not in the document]
mossmaurice: I like the idea! I adapted alternative D accordingly and added some code examples. Another big plus from my point of view: why was `findService()` added to `PoshRuntime`? Because there was no other place to put it! Refactoring this out into a separate class, makes sense for a saner responsibility.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,228 @@
...
+
+##### Alternative A: Shared memory + IPC channel
+
+* Send ID's + ABA-counter/ServiceDiscoveryChangeCounter over IPC channel
+* Data transfer over shared memory
+  * Basically putting Service Registry to shared memory so everyone can access it
+
+Pro:
+
+* Fast data transfer
```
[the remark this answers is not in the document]
mossmaurice: In this alternative, signaling a change in the service registry would go over the IPC channel. Hence, no need to attach to waitset/listner. Have a look at alternative D for approach similar to what you describe here.

doc/design/service-discovery.md
dennis-adlink: Yes, that's correct. For each locally created topic or discovered remote topic (if enabled), a `DCPSTopic` sample is created (or, more precisely, a sample is lazily created based on the local topic administration when reading `DCPSTopic` data). The `DCPSTopic` sample contains meta-data for that single topic (see DDS spec 2.2.5 for the full list of fields). On deletion of a topic (possibly remote via SEDP), the instance for that topic is disposed.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,304 @@
...
+    });
+}
+
+myListner.attachEvent(caproSubscriber, DATA_RECEIVED, createNotificationCallback(onDiscoveryUpdateCallback));
+```
+
+#### Event-based notification: Alternative D
+
+```cpp
+class DiscoveryInfo
```
elfenpiff, line 225: I favor this alternative but there remains two important open question: How do we ensure that there are enough chunks available?

Availability of chunks:
1. The user provides a custom mempool config where samples with the required size are not available at all. Or when they are available there are not enough of them.
2. The users publisher is using the same samples for user data and we are unable to acquire a chunk to send the service registry update.

Access rights:
The mempool must be inside a shared memory segment which is readable by all. What happens when the user configures a custom segment config where a certain application has no read access and is unable to retrieve the samples.


One solution for the first implementation could be to create a mempool inside the management segment which is used only for this particular publisher which sends the service registry.
Could you please include a design draft for this approach (or another) to solve the above mentioned problems.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,304 @@
...
+* Upper frame limit may lead to segregation into several frames
+
+##### Alternative B: Request/ response feature usage
+
+* Re-use [request/ response feature](https://github.com/eclipse-iceoryx/iceoryx/issues/27)
+
+Pro:
+
+* Re-use building blocks to avoid code duplication
+* Dogfooding
```
MatthiasKillat, line 99: I think I know what you mean but it is still confusing, would rephrase it.

doc/design/service-discovery.md
MatthiasKillat: Yes, I think in this case the memory used for this kind of topic must be separate from all other shared memory, i.e. no other publisher uses this. We need to move in the direction of customizing the memory allocator per publisher anyway I think, and as a default it may behave as it does now for all publishers.

There is possibly no perfect way to ensure sufficient memory is available, even with this approach. I suspect discovery workload will spike on startup (many services offered one after the other leads to frequent changes)  and require the most load. The memory must be chosen to anticipate this workload. The publisher could use the blocking send publish feature to ensure nothing gets lost, at the expense of a delay in the information being provided to all subscribing runtimes.

I do not think transmitting the whole registry every time is efficient and the way to go for larger systems, it must be done incrementally (i.e. we transmit deltas only).

doc/design/service-discovery.md
mossmaurice: Added final decision

doc/design/service-discovery.md
mossmaurice: Important point! I added a node, that all communication shall be done over the `iceoryx_managment` segment, so that the user can not interfere with it through a wrong configuration. For me this also applies to the introspection publisher, so I think I'll refactor them alongside to publish in the `iceoryx_mgmt` segment.

elfenpiff: @mossmaurice sounds like a good plan.

doc/design/service-discovery.md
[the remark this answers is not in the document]
MatthiasKillat: Roudi should still react as before on Offer/StopOffer requests I think. StopOffer should not really be used by applications in the static case (except for the publisher dtor in the shutdown phase of the aplication). If you do so, the application is not designed for static discovery. Changing the behavior to no op does not change much, the application also just stop publishing data (you cannot force it to send something, even if it looks/is connected).

Unsubscribe is not really an issue, if the application does not want the data it is free to unsubscribe. The point of  static discovery is in my opinion that the system is designed in a way that the available services do not change (after a startup phase until a shutdown phase). Whether you subscribe to them or not is a separate question.

The main problem is if an app does not follow this contract and stops offering a previously offered services mid-run. Then the static discovery information does not get updated and users trying to subscribe will just not find the service (and hence receive no data). In the static discovery case Roudi or some monitor must ensure that we react to this scenario (e.g. print a warning, terminate, something in-between).

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+
+* Close to DDS mechanism
+* `InterfacePort` already available
+* Synergies with gateway mechanisms
+* Gateways can benefit from event notification mechanism
+  * No need for a polling-loop anymore
+
+Con:
+
+* Overhead, `Listner` will wake up on every `CaproMessage`
```
MatthiasKillat, line 143: Listener

MatthiasKillat, line 143: I suspect even in the dynamic case there is little change after a startup phase, otherwise you have a highly artificial systems where services come and go often (unlikely). Hence I think this overhead is tolerable.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+* Bookkeeping at user-side, just delta of service registry is transferred
```
MatthiasKillat, line 144: Sending the delta only is not sufficient, since we need to deal with late joiners which do not receive all the updated delta information. A mechanism to send the complete registry is still needed. This is used to correct the delta information,  which is not as simple as one may think. I sketched the details how this could work locally in some file, I may append them here but the style is more formal.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+  * A safety-certified middleware based on iceoryx would not use the dynamic discovery feature
+
+Remark:
+
+* Need to filter out your own `CaproMessageType::FIND` requests
+* Extend `dispatchCaproMessage()` with condition variable and notification mechanism
+
+##### Alternative C: Custom thread
+
+* Implement a custom thread in `PoshRuntime`, which polls for specific discovery information using request/response
```
MatthiasKillat, line 157: This mainly makes sense if one does not want to use the listener and built-in topics. Otherwise the listener has this thread already and we just need to react on updates in its callback. I am not sure the user trigger callbacks are powerful to do so, but this can be corrected. They are able at least able to operate on some object or take an opaque void pointer, which is sufficient.

MatthiasKillat: Snapshoting is tricky in its own right. The registry container should not be protected by mutex, since contention would be quite high with a large number of applications (which will simultaneously try to get the snapshot if something has changed). Instead it must be able to be modified (insertions/deletion) by Roudi while those snapshots take place.

Not all apps will see the same snapshot, but this is ok as long as it is eventually consistent.

This is possible to do in a lock-free way, we may already have the right data structure, otherwise I could provide one I have already experimented with in another problem.

MatthiasKillat: My conclusion is that the effort to implement variant C is not reasonable and variant D is easier to implement. The potentially small performance does not justify not reusing the listener.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+Create a new publisher in RouDi which sends a `ServiceRegistryTopic`. This publisher would be used both to signal a
+change in the service registry and to transmit the service discovery registry. The complete old service registry
+(saved locally) would be compared to the new service registry in a new class, extending the public user API.
+
+Pro:
+
+* Simple and consistent user API for both event-based and sychronous requests
+  * Filtering for `findService` could be done inside the new class
+* No need for a change counter as ABA problem solved by POSH mechanisms
+* Not just delta of service registry is transmitted, but complete service registry info
```
MatthiasKillat, line 180: From my point of view this can even be improved later to also base the current cregistry state on incremental updates in addition. In some way it is a state estimation problem with the known registry being the state and some information (full state or updates) arriving which can be used to reestimate the state.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+        get();
+
+        return filterFor(serviceDescription);
+    }
+
+    // @todo check if still needed
+    bool offerService(const capro::ServiceDescription& serviceDescription) noexcept;
+    void stopOfferService(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    ServiceRegistryTopic get() noexcept
```
MatthiasKillat, line 250: As noted this also covers the one shot complete registry state update on demand.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+        m_subscriber.take().and_then([](auto& ServiceRegistryTopic){
+            // Update our local copy of the service registry
+            m_lastServiceRegistry = ServiceRegistryTopic;
+        }).or_else([](){
+            // No change in service registry
+        });
+        return m_lastServiceRegistry;
+    }
+
+    ServiceRegistryTopic getDelta() noexcept
```
MatthiasKillat, line 261: Computing the delta efficiently is important if it is used often. This may not be easy/possible with strings though but the general idea to compute the delta is useful/needed. We can also cache the result and use it if the state (`m_lastRegistry`) did not change, i.e. we did not receive any update, we do not need to recompute the delta.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+    {
+        auto old = m_lastServiceRegistry;
+        get();
+        return old - m_lastServiceRegistry;
```
MatthiasKillat, line 265: Ok, the - is just symbolic. But what is a difference here? Well, for starters we have  symmetirc difference of sets, which measures the total changes between both sets. But we may be more interested in a more detailed picture of computing
new/old and old/new, i.e. newly offered services and services where offering was stopped.

We could do so in two separate functions.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    ServiceRegistryTopic getDelta(const capro::ServiceDescription& serviceDescription) noexcept;
+
+   filterFor(const capro::ServiceDescription& serviceDescription) noexcept;
+  private:
+    Subscriber<ServiceRegistryTopic> m_subscriber{"ServiceDiscovery", "GlobalInstance", "ServiceRegistryTopic"};
```
MatthiasKillat, line 272: In an extension we could also provide incremental updates with an additional subscriber here to reduce the need f always having to send the full state. This does not need to be done in a first version but the possibility exists.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+    ServiceRegistryTopic m_lastServiceRegistry;
+}
+
+// Class should be attachable to listner
```
MatthiasKillat, line 276: With these combined the user can create an active class which handles the state update automatically when new data is received. The callback must contain the required state update on receiving the new registry.

While the user can do so on its own, it may be useful to define a `DynamicDiscovery` (or similar) class which encapsulates it for he default use case. If the user wants do perfom somemthing specific he needs to either write its own version or  provide a different callback.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+    ServiceRegistryTopic getDelta() noexcept
+    {
+        auto old = m_lastServiceRegistry;
+        get();
+        return old - m_lastServiceRegistry;
+    }
+
+    ServiceRegistryTopic getDelta(const capro::ServiceDescription& serviceDescription) noexcept;
+
+   filterFor(const capro::ServiceDescription& serviceDescription) noexcept;
```
MatthiasKillat, line 270: Depending on the size of the registry, efficient filtering may require a map like structure which also ideally needs to be able to be transmitted via shared memory (avoid serialization)

doc/design/service-discovery.md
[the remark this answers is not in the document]
mossmaurice: > StopOffer should not really be used by applications in the static case (except for the publisher dtor in the shutdown phase of the aplication). If you do so, the application is not designed for static discovery. Changing the behavior to no op does not change much, the application could also just stop publishing data (you cannot force it to send something, even if it looks/is connected).

Well, it's complicated. We need to provide users with a smooth transition from a dynamic world into a static world. So I agree, `stopOffer` should not be used but will likely for simplicity just become a no-op.

I'll added a link to this discussion in #539, we can continue there.

doc/design/service-discovery.md
mossmaurice: I agree, that's why we opted for alternative D.

doc/design/service-discovery.md
mossmaurice: Still need to put some brainpower into how to do that efficiently. Let's iterate a couple of times over it.

doc/design/service-discovery.md
mossmaurice: > While the user can do so on its own, it may be useful to define a DynamicDiscovery (or similar) class which encapsulates it for he default use case.
Will think about it, although I have the feeling this should probably be done in the binding code (e.g. AUTOSAR binding or rmw)

elBoberido (commented):
I think proposal A is superior to D but how about proposal E:
Since proposal D already involves copying, why the detour over a pub-sub and not just copy from an array in shared memory
- lets have an array with all the serialized service descriptions
```
struct Entry {
    std::atomic<uint64_t> generation{0};
    char serializedServiceDescription[400];
    uint16_t serializedSize{0};
};
```
- with each new service, RouDi updates an entry by
  - setting `generation` to `INVALID`
  - setting `serializedServiceDescription` and `serializedSize`
  - setting `generation` to previous value + 1
  - firing condition variables who where registered for a discovery notification
- application registers via IPC for a discovery notification and gets a condition variable
  - waits on condition variable to fire (this could be done by `WaitSet` or `Listener`)
  - application runs through the array
  - reads `generation` and `serializedSize`
  - memcopies `serializedServiceDescription` if `generation` is not `INVALID`
  - reads `generation` and dismisses the copied service description if `generation` changed

Although this has some lock-free code, it's a simple code and easy to reason about. It's fast and does not have any memory issues. It could reuse the keep alive thread to not increase the amount of threads in use by the runtime.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+* Fast data transport
+* Close to DDS mechanism
+* Could replace `InterfacePort`s
+
+Con:
+
+* New publisher in RouDi needed
+* Partially overlapping features with `InterfacePort`s
+  * Could the `InterfacePort`s be replace by alternative D?
+
```
elfenpiff, line 190: After thinking about this design I have some further cons for all the approaches which are using a variation of subscriber & publisher ports (or @mossmaurice eat your own dogfood ;) ). The reason is always that every app would like to do some service discovery.

- The number of subscribers has to be increased by the number of maximum allowed applications. This means instead of 256 subscribers we require 526 subscribers (since we allow 300 apps).
- The number of subscribers per publisher has to be increased to be at least equal to the maximum number of applications (from 256 to 300)
- Because of @elBoberido scenario from above: https://github.com/eclipse-iceoryx/iceoryx/pull/784/files#r638915043 the `IOX_MAX_CHUNKS_ALLOCATED_PER_PUBLISHER_SIMULTANEOUSLY` must be increased from 8 to maximum allowed applications (300) otherwise the publisher may kill the whole roudi when some application is interrupted and there are no more chunks.
- We require consistency checks for our configuration. Lets say someone sets iceoryx up with 300 apps but only 200 subscribers per publisher. Then only 200 apps can do some service discovery. In the end all the mentions from above have to be verified if the setup is feasible.
- We have to increase the maximum number of condition variable by the maximum number of applications. Since the first 300 will be reserved by the service discovery. This can slow down the WaitSet, when it is used with states and not events, if the user uses the default template arguments. 
- We have to ensure the availability of condition variables. This means when we configure 500 and the users acquires too much then new application are unable to acquire new condition variables which are required for the Listener/Waitset during service discovery. One solution would be another condition variable pool only for service discovery.
- We require a separate memory pool in the shared memory which is reserved solely for the service discovery.

In summary you can say those cons lead to a massive resource overhead (a subscriber alone is 16kb large I think) and implementation overhead. Yes, we can reuse code but at a high price since we have to add checks, resource separations etc to guarantee the availability of resources. The additional condition variable pool and mem pool will also be a lot of additional code which has to be tested as well. And this will be a lot of more code then we save.

Here an additional con for alternative D:
- We have some kind of a priority inversion problem. Let's say we have a high priority application and a low priority application. If the low priority is interrupted during its copy then all the high prio applications cannot receive service registry updates until the low priority application has finished its copy work. This has to be avoided.


With this said I would conclude that reusing our ports for service discovery are out of the question since the cons greatly outweigh the pros.

elfenpiff (commented):
> I think proposal A is superior to D but how about proposal E:
> Since proposal D already involves copying, why the detour over a pub-sub and not just copy from an array in shared memory
> 
> * lets have an array with all the serialized service descriptions
> 
> ```
> struct Entry {
>     std::atomic<uint64_t> generation{0};
>     char serializedServiceDescription[400];
>     uint16_t serializedSize{0};
> };
> ```
> 
> * with each new service, RouDi updates an entry by
>   
>   * setting `generation` to `INVALID`
>   * setting `serializedServiceDescription` and `serializedSize`
>   * setting `generation` to previous value + 1
>   * firing condition variables who where registered for a discovery notification
> * application registers via IPC for a discovery notification and gets a condition variable
>   
>   * waits on condition variable to fire (this could be done by `WaitSet` or `Listener`)
>   * application runs through the array
>   * reads `generation` and `serializedSize`
>   * memcopies `serializedServiceDescription` if `generation` is not `INVALID`
>   * reads `generation` and dismisses the copied service description if `generation` changed
> 
> Although this has some lock-free code, it's a simple code and easy to reason about. It's fast and does not have any memory issues. It could reuse the keep alive thread to not increase the amount of threads in use by the runtime.

@elBoberido we have still one problem here. We have to massively increase the number of condition variables. But this could be solved by a posix condition variable (a real one and not the semaphore trick we did there).
A real posix condition variable is inter process capable (I know it since I use them as semaphore alternative in mac OS) and can notify all with one call so multiple applications can wait on one single condition variable.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
+# Service Discovery
+
+## Summary and problem description
+
+Service discovery over IPC channel e.g. message queue or UNIX domain socket is not performant since larger data is
+transferred, which can lead to transmission of several frames. If lots of services are discovered at high-frequency
+e.g. at startup the IPC channel can become a bottleneck. Furthermore, a lesson learned from the iceoryx development
+so far is, that the IPC channel should just be used for the earliest communication during startup and not for the
+creation of objects inside the shared memory (see [this issue](https://github.com/eclipse-iceoryx/iceoryx/issues/611)).
```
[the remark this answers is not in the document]
MatthiasKillat: I agree.
The lock-free structures (queues and management of chunks for publishers and subscribers) are an issue but the queue should be decently fast even under contention. I assume most of the time is spend in calls on different abstraction layers for the message passing and managing the chunk cache.

This can be avoided with a more direct approach (all of them are IPC in a way) as @elBoberido said.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,316 @@
...
+* Not just delta of service registry is transmitted, but complete service registry info
+* Fast data transport
+* Close to DDS mechanism
+* Could replace `InterfacePort`s
+
+Con:
+
+* New publisher in RouDi needed
+* Partially overlapping features with `InterfacePort`s
+  * Could the `InterfacePort`s be replace by alternative D?
```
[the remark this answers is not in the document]
MatthiasKillat: Seconded. Ideally in a no port approach Roudi manages the state (services) in shared memory which can applications only have read access to. This state is as large as it needs to be (max number of services bounds it) or could be growing as long as memory is available. Applications have their own local state which does not reside in shared memory but is initialized e.. when the runtime starts (or service discovery is requested for the first time at the latest).

Apps will read from Roudis state under certain conditions in a lock-free way and update their own state accordingly.

This way the required memory is entirely static (assuming a non growing state approach for now).

MatthiasKillat:
@elfenpiff 
Regarding the number of semaphores/condition variables.

1. A condition variable internally (usually) uses one or more semaphores or a similar signaling mechanism. I do not know how exactly POSIX condition variables are usually implemented, but usually you need a semaphore for each waiting entity (process/thread) The semaphore cannot be shared easily since then a wake-all will not work properly (we need a wake all here).

2. As the minimum we need one semaphore per application (runtime) , but we would already have this with the listener as well (the condition variable it internally uses is not really a condition variable but a semaphore with some additional functionality).
So this is not worse than any of the presented approaches and I think unavoidable without falling back to periodic polling.

3. If we agree tat we use one semaphore (or listener, at a higher level) per application for this, we just need to use it to notify the app that something happened, in our case the known services changed. In particular I think we need another abstraction around a semaphore which I have yet to think about (need some time). 

Regarding the memory usage.

The state exchange I described in some comment above makes the required memory completely static and is compatible with the idea @elBoberido presented. I will avoid the full quote.
``
struct Entry {
    std::atomic<uint64_t> generation{0};
    char serializedServiceDescription[400];
    uint16_t serializedSize{0};
};
``
Whenever Roudi signals the state change, the application can react and read from the state to update its own state.

The representation is still matter of discussion, but the important part is we have single writer/multi reader system here and the reads and writes can be performed lock-free without corrupting the state or partial reads.

Regarding data representation
I would distinguish between the exchange format and the lookup format. The less time we spend on brute-force iterations, the better. (how large can this array be, it probably can be sparse and so on). The reading applications do not have to store the entries read in an array themselves, the can use some kind of map or another structure which allows fast look-ups. If only the app accesses this structure via the runtime it does not have to be lock-free (the worst scenario is the app locking itself out if its listening thread is not scheduled). A lock-free solution is also conceivable, but a map (with a lock-free search tree) would be quite hard here.

Optimization - Incremental updates.
The app does not necessarily need to be informed that it has to read the array, but it could be informed that it has to check a specific entry, avoiding iteration in the update. There is more to this though to ensure the state exchange is **eventually consistent** (which is the notion we need IMO) but this can certainly be done and does not have to be done in a first implementation which would check the whole array every time.

There are of course details that need to be thought about (such as state representation, i.e. the services) , but this is some kind of general state exchange protocol bypasses the ports and avoids most of the drawbacks. It is also important I think to hide most of this in the runtime and do not give the user an interface which allows her to cause an application deadlock.

doc/design/service-discovery.md
mossmaurice: > This means instead of 256 subscribers we require 526 subscribers (since we allow 300 apps)

That's a valid point!

> We require consistency checks for our configuration

This will be required anyway, no matter which alternative we choose.

> We have some kind of a priority inversion problem.

That's not the case with our SoFi or MPMC queue. RouDi (highest prio) will always be able to write data.

doc/design/service-discovery.md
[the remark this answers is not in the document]
elBoberido: That was my question. If this runs in a separate Listener/WaitSet than the regular pub-sub communication, then we need an additional condition variable. But this is also the case with my proposal.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,421 @@
+# Service Discovery
+
+## Summary and problem description
+
+Service discovery over IPC channel e.g. message queue or UNIX domain socket is not performant since larger data is
+transferred, which can lead to transmission of several frames. If lots of services are discovered at high-frequency
+e.g. at startup the IPC channel can become a bottleneck. Furthermore, a lesson learned from the iceoryx development
```
elBoberido, line 7: Could you please remove the line starting with `Furthermore` or rephrase it that this is only necessary for safety environments where all the resources are created on startup anyway, e.g. by placing all the objects in a `StatusPort` only the respective app has access to. This would be sent to the application when it registers at RouDi. In a dynamic non safety scenario this is not possible and we should continue to use the IPC channel.

doc/design/service-discovery.md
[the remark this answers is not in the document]
elBoberido: So we can use the `StatusPort` for the heartbeat?

In a safety environment you create all the resources at startup and currently only the heartbeat is sent via IPC channel. Please rephrase this to `In a safety environment IPC channel ...`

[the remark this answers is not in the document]
mossmaurice: > So we can use the StatusPort for the heartbeat?

After thinking about it again, I don't see any reason why not.

doc/design/service-discovery.md
mossmaurice: Rephrased. I guess you're right when it comes to the objects in shared memory in the safety case. However, we still need to be able to send specific pointers to the `PoshRuntime`. IMHO we should have one "Sned me the pointer"-mechanism for both the safety and the dynamic case for simplicity. #611 discusses this in more detail. @elfenpiff and myself were pondering about a factory class that lives in shared memory to be able to utilise RAII.

elBoberido (approved):

budrus (approved):

doc/design/diagrams/service-discovery/sequence-diagram-alternative-d.puml
```diff
@@ -0,0 +1,38 @@
...
+participant RouDi
+participant UserAppA
+participant UserAppB
+
+activate UserAppA
+
+UserAppA->UserAppA: DiscoveryInfo::DiscoveryInfo()
+UserAppA->RouDi:  getMiddlewareSubscriber() / CREATE_Subscriber
+
+RouDi->UserAppA: CREATE_INTERFACE_ACK
```
elfenpiff, line 13: Shouldn't this be `CREATE_SUBSCRIBER_ACK`?

doc/design/service-discovery.md
```diff
@@ -0,0 +1,421 @@
...
+
+#### SOME/IP-SD
+
+* Similar to AUTOSAR Adaptive
+
+#### Cyclone DDS and ROS 2
+
+* Cyclone DDS calls this feature [topic discovery](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/dev/topic_discovery.md). See also the [feature request](https://github.com/eclipse-cyclonedds/cyclonedds/issues/42).
+
+* Based on SDEP (simple endpoint discovery protocol)
```
elfenpiff, line 49: Shouldn't this be `SEDP` otherwise the letter ordering does not match the name. If so could you please adapt all occurrences.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,421 @@
...
+* Fast data transport by using shared memory
+* Could replace `InterfacePort`s
+* Not just delta of service registry is transmitted, but complete service registry info
+* Simple and consistent user API for both event-based and sychronous requests
+  * Filtering for `findService` could be done inside the new class
+
+Con:
+
+* New class, which needs RouDi infrastructure integration (e.g. requesting this port)
+* Starvation possible if the writer updates too often and the reader never finished its read (should be taken care by
```
elfenpiff, line 224: You can also add that this problem is commonly accepted since every lock-free algorithm has this problem.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,421 @@
...
+* Re-use the `StatusPort` for introspection 2.0, heartbeat/keepalive and to send static configurations
+* Fast data transport by using shared memory
+* Could replace `InterfacePort`s
+* Not just delta of service registry is transmitted, but complete service registry info
+* Simple and consistent user API for both event-based and sychronous requests
+  * Filtering for `findService` could be done inside the new class
+
+Con:
+
+* New class, which needs RouDi infrastructure integration (e.g. requesting this port)
```
elfenpiff, line 223: A con is also that the data needs to be copied.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,421 @@
...
+    using uint64_t UPDATING {-1};
+    std::atomic<uint64_t> generation{INVALID};
+}
+template <typename T>
+using StatusPortData = Sample<T, StatusPortHeader>; // pretty neat, eh? We use a chunk to store the StatusPortData -> this means the number of ports and data sizes can be configured by the user without recompiling RouDi; could potentially also be done for the other ports
+
+template <typename T>
+StatusPortWriter {
+  public:
+    void write(cxx::function_ref<void(T&)> f);
```
elfenpiff, line 328: If the `StatusPortWriter` writes the data into the status port, why requires the lambda a non const `T` reference then? The writer only reads the data and writes it into the port?
If its the other way around the name would be somehow misleading or I am just confused.

doc/design/service-discovery.md
```diff
@@ -0,0 +1,421 @@
...
+        get();
+
+        return filterFor(serviceDescription);
+    }
+
+    // @todo check if still needed
+    bool offerService(const capro::ServiceDescription& serviceDescription) noexcept;
+    void stopOfferService(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    ServiceRegistryTopic get() noexcept
```
elfenpiff, line 360: If the `statusPortReader` supports the callback based approach we should also make this available for the public interface since it could become handy.
Something like: `void getCallback(const cxx::function<const ServiceRegistryTopic&> f);` then the user can provide directly a free function which handles it. Something similar to `Publisher::publishResultOf`.

And `getCallback` could then be reused by all the `get` impls here.

doc/design/service-discovery.md
mossmaurice: Code snippet is from @elBoberido my understanding: The writes the data via his lambda so it can't be const.

doc/design/service-discovery.md
mossmaurice: I'm unsure here. Usually I prefer to have one way only how the user can access an API and not several slightly different ones (Hyrums Law). I'll keep it in mind for the implementation.

elfenpiff: Here I would disagree since it would make the life of the user easier and we also did go exactly this way with the subscriber and the publisher.

budrus (approved):

elfenpiff (approved):

doc/design/service-discovery.md
elBoberido: The `T&` is the reference to th data wrapped by the `StatusPort` and the method would be used like this
```cpp
StatusPortWriter<uint64_t> status;
status.write([] (auto& s) {
    s = 42;
});
```

elBoberido (approved):

PR #797, opened 2021-05-17
iox-#33 posix mutex support in windows

### LINKED ISSUE #33: Basic Windows 10 support

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #33

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -7,7 +7,8 @@ CheckOptions:
   - { key: readability-identifier-naming.MethodCase,              value: camelBack }
   - { key: readability-identifier-naming.FunctionCase,            value: camelBack }
   - { key: readability-identifier-naming.NamespaceCase,           value: lower_case }
-  - { key: readability-identifier-naming.MemberPrefix,            value: m_ }
+  - { key: readability-identifier-naming.PrivateMemberPrefix,     value: m_ }
+  - { key: readability-identifier-naming.ProtectedMemberPrefix,   value: m_ }
   - { key: readability-identifier-naming.MemberCase,              value: camelBack }
   - { key: readability-identifier-naming.ConstexprVariableCase,   value: UPPER_CASE }
   - { key: readability-identifier-naming.EnumCase,                value: UPPER_CASE }
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -108,10 +108,7 @@ jobs:
       # we have to exclude the tests explicitly until everyone is running
       - name: Run tests, excluding timing_tests
         run: |
-          build\hoofs\test\Debug\hoofs_moduletests.exe \
-            --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
+          build\hoofs\test\Debug\hoofs_moduletests.exe --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
           build\hoofs\test\Debug\hoofs_integrationtests.exe
-          build\binding_c\test\Debug\binding_c_moduletests.exe \
-            --gtest_filter="-Chunk*:iox_listener*:iox_node*:iox_pub*:BindingC_Runtime*:iox_sub*:iox_ws*:iox_types*:*TimingTest*"
-          build\posh\test\Debug\posh_moduletests.exe \
-            --gtest_filter="-PoshRuntime*:IceoryxRoudiApp*:IceoryxRoudiMemoryManager*:PortManager*:ProcessIntrospection*:ProcessManager*:PosixShmMemoryProvider*:Process_test*:ParseAllMalformedInput*:*TimingTest*"
+          build\binding_c\test\Debug\binding_c_moduletests.exe --gtest_filter="-Chunk*:iox_listener*:iox_node*:iox_pub*:BindingC_Runtime*:iox_sub*:iox_ws*:iox_types*:*TimingTest*"
+          build\posh\test\Debug\posh_moduletests.exe --gtest_filter="-PoshRuntime*:IceoryxRoudiApp*:IceoryxRoudiMemoryManager*:PortManager*:ProcessIntrospection*:ProcessManager*:PosixShmMemoryProvider*:Process_test*:ParseAllMalformedInput*:*TimingTest*"
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -103,3 +103,12 @@ jobs:
       - name: Build sources
         run: |
           cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON && cmake --build build
+
+      # until the windows support is fully implemented and we can use the windows cmake targets
+      # we have to exclude the tests explicitly until everyone is running
+      - name: Run tests, excluding timing_tests
+        run: |
+          build\hoofs\test\Debug\hoofs_moduletests.exe --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
+          build\hoofs\test\Debug\hoofs_integrationtests.exe
+          build\binding_c\test\Debug\binding_c_moduletests.exe --gtest_filter="-Chunk*:iox_listener*:iox_node*:iox_pub*:BindingC_Runtime*:iox_sub*:iox_ws*:iox_types*:*TimingTest*"
+          build\posh\test\Debug\posh_moduletests.exe --gtest_filter="-PoshRuntime*:IceoryxRoudiApp*:IceoryxRoudiMemoryManager*:PortManager*:ProcessIntrospection*:ProcessManager*:PosixShmMemoryProvider*:Process_test*:ParseAllMalformedInput*:*TimingTest*"
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Build sources
         run: |
-          cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON && cmake --build build
+          cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON -DCMAKE_CXX_FLAGS="/MP" && cmake --build build
 
       # until the windows support is fully implemented and we can use the windows cmake targets
       # we have to exclude the tests explicitly until everyone is running
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -108,7 +108,10 @@ jobs:
       # we have to exclude the tests explicitly until everyone is running
       - name: Run tests, excluding timing_tests
         run: |
-          build\hoofs\test\Debug\hoofs_moduletests.exe --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
+          build\hoofs\test\Debug\hoofs_moduletests.exe \
+            --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
           build\hoofs\test\Debug\hoofs_integrationtests.exe
-          build\binding_c\test\Debug\binding_c_moduletests.exe --gtest_filter="-Chunk*:iox_listener*:iox_node*:iox_pub*:BindingC_Runtime*:iox_sub*:iox_ws*:iox_types*:*TimingTest*"
-          build\posh\test\Debug\posh_moduletests.exe --gtest_filter="-PoshRuntime*:IceoryxRoudiApp*:IceoryxRoudiMemoryManager*:PortManager*:ProcessIntrospection*:ProcessManager*:PosixShmMemoryProvider*:Process_test*:ParseAllMalformedInput*:*TimingTest*"
+          build\binding_c\test\Debug\binding_c_moduletests.exe \
+            --gtest_filter="-Chunk*:iox_listener*:iox_node*:iox_pub*:BindingC_Runtime*:iox_sub*:iox_ws*:iox_types*:*TimingTest*"
+          build\posh\test\Debug\posh_moduletests.exe \
+            --gtest_filter="-PoshRuntime*:IceoryxRoudiApp*:IceoryxRoudiMemoryManager*:PortManager*:ProcessIntrospection*:ProcessManager*:PosixShmMemoryProvider*:Process_test*:ParseAllMalformedInput*:*TimingTest*"
--- a/iceoryx_hoofs/cmake/IceoryxPlatform.cmake
+++ b/iceoryx_hoofs/cmake/IceoryxPlatform.cmake
@@ -34,7 +34,7 @@ elseif(APPLE)
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+    add_definitions(/MP)
     set(ICEORYX_WARNINGS PRIVATE ${ICEORYX_WARNINGS} /W0) # TODO iox-#33 set to /W1
     # todo: '/O2' and '/RTC1' (set by default) options are incompatible,
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/iceoryx_hoofs/cmake/IceoryxPlatform.cmake
+++ b/iceoryx_hoofs/cmake/IceoryxPlatform.cmake
@@ -28,12 +28,13 @@ if(LINUX)
 elseif(QNX)
     set(ICEORYX_CXX_STANDARD 14)
 elseif(WIN32)
-    set(ICEORYX_CXX_STANDARD 14)
+    set(ICEORYX_CXX_STANDARD 17)
 elseif(APPLE)
     set(ICEORYX_CXX_STANDARD 17)
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
     set(ICEORYX_WARNINGS PRIVATE ${ICEORYX_WARNINGS} /W0) # TODO iox-#33 set to /W1
     # todo: '/O2' and '/RTC1' (set by default) options are incompatible,
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/iceoryx_hoofs/cmake/IceoryxPlatform.cmake
+++ b/iceoryx_hoofs/cmake/IceoryxPlatform.cmake
@@ -34,7 +34,6 @@ elseif(APPLE)
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-    add_definitions(/MP)
     set(ICEORYX_WARNINGS PRIVATE ${ICEORYX_WARNINGS} /W0) # TODO iox-#33 set to /W1
     # todo: '/O2' and '/RTC1' (set by default) options are incompatible,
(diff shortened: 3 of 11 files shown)


.github/workflows/build-test.yml
```diff
@@ -103,3 +103,12 @@ jobs:
       - name: Build sources
         run: |
           cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON && cmake --build build
+
+      # until the windows support is fully implemented and we can use the windows cmake targets
+      # we have to exclude the tests explicitly until everyone is running
+      - name: Run tests, excluding timing_tests
+        run: |
+          build\hoofs\test\Debug\hoofs_moduletests.exe --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
```
dkroenke, line 111: You can use here multiline strings with `\` as separator to make it more readable:
```bash
build\hoofs\test\Debug\hoofs_moduletests.exe \
--gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
```

.github/workflows/build-test.yml
```diff
@@ -103,3 +103,12 @@ jobs:
       - name: Build sources
         run: |
           cmake -Bbuild -Hiceoryx_meta -DBUILD_TEST=ON -DINTROSPECTION=OFF -DBINDING_C=ON -DEXAMPLES=ON && cmake --build build
```
dkroenke, line 105: You can add here `-DCMAKE_CXX_FLAGS="/MP"` to build windows with multiple CPU cores

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -134,4 +134,39 @@ TEST_F(Mutex_test, TryLockWithRecursiveMutexReturnsFalseWhenMutexLockedInOtherTh
     sutRecursive.unlock();
     lockThread.join();
 }
+
+TEST_F(Mutex_test, LockedMutexBlocks)
+{
+    std::atomic_bool isLockFinished{false};
+    sutNonRecursive.lock();
+
+    std::thread lockThread([&] {
```
dkroenke, line 143: Would be good to have some timeout in this thread to avoid a deadlock in these tests. Maybe we can use the `timedwait` from the semaphore here?

.github/workflows/build-test.yml
elfenpiff: I already added this cxx flag in iceoryx_hoofs/cmake/IceoryxPlatform.cmake, line 37

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
elfenpiff: I will release our Watchdog :)

elfenpiff: There is already a watchdog setup in the `SetUp` method of the test (line 37) which will terminate the test if it deadlocks.

iceoryx_hoofs/platform/win/source/semaphore.cpp
```diff
@@ -76,7 +80,8 @@ int iox_sem_timedwait(iox_sem_t* sem, const struct timespec* abs_timeout)
     }
 
     time_t epochCurrentTimeDiffInSeconds = abs_timeout->tv_sec - tv.tv_sec;
-    long milliseconds = epochCurrentTimeDiffInSeconds * 1000 + ((abs_timeout->tv_nsec / 1000) - tv.tv_usec) / 1000;
+    long milliseconds = epochCurrentTimeDiffInSeconds * 1000 + (abs_timeout->tv_nsec / 1000000 - tv.tv_usec / 1000);
```
elBoberido, line 83: What was the reason for this change? The previous one seem to be more accurate?

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -134,4 +134,39 @@ TEST_F(Mutex_test, TryLockWithRecursiveMutexReturnsFalseWhenMutexLockedInOtherTh
...
+        sutNonRecursive.lock();
+        isLockFinished.store(true);
+        sutNonRecursive.unlock();
```
elBoberido, line 146: why is this set to true under the lock?

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -134,4 +134,39 @@ TEST_F(Mutex_test, TryLockWithRecursiveMutexReturnsFalseWhenMutexLockedInOtherTh
...
+    });
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
elBoberido, line 149: Please be more creative than just adding big sleeps. Our CI is already slow.

A possible solution would be to
- lock the mutex in the main thread
- start a thread and signal the main thread to be ready
  - get the current time
  - get the lock
  - measure the time it takes to get the lock
  - check if the duration is larger than the sleep time in the main thread
- in the main thread wait for the ready signal from the lockThread
- sleep a few milliseconds (5 to 10)
- release the lock
- join the threads

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -134,4 +134,39 @@ TEST_F(Mutex_test, TryLockWithRecursiveMutexReturnsFalseWhenMutexLockedInOtherTh
...
+
+    EXPECT_THAT(isLockFinished.load(), Eq(true));
+}
+
+TEST_F(Mutex_test, TryLockWithRecursiveMutexWhenMutexLocked)
+{
+    std::atomic_bool isTryLockSuccessful{true};
+    sutRecursive.lock();
+
+    std::thread lockThread([&] { isTryLockSuccessful = sutRecursive.try_lock(); });
```
elBoberido, line 163: the same could be done here

iceoryx_hoofs/cmake/IceoryxPlatform.cmake
```diff
@@ -28,12 +28,13 @@ if(LINUX)
...
     set(ICEORYX_CXX_STANDARD 14)
 elseif(WIN32)
-    set(ICEORYX_CXX_STANDARD 14)
+    set(ICEORYX_CXX_STANDARD 17)
 elseif(APPLE)
     set(ICEORYX_CXX_STANDARD 17)
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
```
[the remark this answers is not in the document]
dkroenke: You are right, this will do the job:
```cmake
add_compile_options(/MP)
```

.github/workflows/build-test.yml
```diff
@@ -103,3 +103,15 @@ jobs:
...
+
+      # until the windows support is fully implemented and we can use the windows cmake targets
+      # we have to exclude the tests explicitly until everyone is running
+      - name: Run tests, excluding timing_tests
+        run: |
+          build\hoofs\test\Debug\hoofs_moduletests.exe \
+            --gtest_filter="-SignalHandler*:relativeptr*:Timer*:base_relative*:DeadlineTimer*:Thread*:Duration*:*TimingTest*"
```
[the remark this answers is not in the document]
elfenpiff: No that is not the problem, the problem is that in windows you cannot do this linebreak with \

iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/pthread.hpp
```diff
@@ -17,78 +17,46 @@
 #ifndef IOX_HOOFS_WIN_PLATFORM_PTHREAD_HPP
 #define IOX_HOOFS_WIN_PLATFORM_PTHREAD_HPP
 
+#include "iceoryx_hoofs/platform/win32_errorHandling.hpp"
+#include "iceoryx_hoofs/platform/windows.hpp"
+
+
```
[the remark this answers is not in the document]
elfenpiff: but I like empty lines ;)

iceoryx_hoofs/platform/win/source/semaphore.cpp
elBoberido: Well, that's not quite correct. In your example the resulting time would be 9.101ms and 9.001001ms but with the new calculation you get 10ms.  Is this intended?
How about
```
auto nanosecondsLeft = abs_timeout->tv_nsec;
auto nanosecondsRight = tv.tv_usec * 1000;
auto rounding = nanosecondsLeft >= nanosecondsRight ? 500000 : -500000;
(nanosecondsLeft - nanosecondsRight + rounding) / 1000000
```
This would correctly round up and down depending if the actual value is above or below .5 and if the result is positive or negative

elfenpiff: The 10ms are intended, then my example wasn't so good after all.

iceoryx_hoofs/source/posix_wrapper/mutex.cpp
```diff
@@ -74,7 +74,8 @@ bool mutex::unlock()
 
 bool mutex::try_lock()
 {
-    return !posixCall(pthread_mutex_trylock)(&m_handle).successReturnValue(0).evaluate().has_error();
+    auto result = posixCall(pthread_mutex_trylock)(&m_handle).successReturnValue(0).evaluateWithIgnoredErrnos(EBUSY);
+    return !(!result.has_error() && result->errnum == EBUSY) && !result.has_error();
```
elBoberido, line 78: The `!(!result.has_error() && result->errnum == EBUSY)` part is hard to graps. Would you mind introducing a variable like
```cpp
auto isEBUSY = !result.has_error() && result->errnum == EBUSY;
return !isEBUSY && !result.has_error();
```

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -46,40 +46,62 @@ class Mutex_test : public Test
         }
     }
 
-    iox::posix::mutex sutNonRecursive{false};
-    iox::posix::mutex sutRecursive{true};
+    void signalThreadReady()
+    {
+        m_waitForThread.store(false);
```
elBoberido, line 51: this could be `memory_order_relaxed`

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -46,40 +46,62 @@ class Mutex_test : public Test
...
+    }
+
+    void waitForThread()
+    {
+        while (m_waitForThread.load())
```
elBoberido, line 56: same here ... and could you add `yield` to give the other thread a chance to run?

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -46,40 +46,62 @@ class Mutex_test : public Test
...
+    }
+
+    template <typename T>
+    int64_t getDuration(const T& start, const T& end)
+    {
+        return std::chrono::duration_cast<std::chrono::nanoseconds>(end - start).count();
+    }
+
+    static constexpr int64_t WAIT_IN_MS = 100;
+    static constexpr int64_t WAIT_IN_NS = WAIT_IN_MS * 1000000;
```
elBoberido, line 68: This is only used in one test. Could it be moved there?
Also, 100ms are quite much. If the mutex is not locked, it takes essentially no time between the two measurements

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -46,40 +46,62 @@ class Mutex_test : public Test
...
+
+    std::atomic_bool m_waitForThread{true};
+    iox::posix::mutex m_sutNonRecursive{false};
+    iox::posix::mutex m_sutRecursive{true};
     iox::units::Duration m_watchdogTimeout = 5_s;
```
elBoberido, line 73: Why the `m_` prefix? This are all public members which should not have the prefix

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -100,38 +122,63 @@ TEST_F(Mutex_test, CallingDestructorOnLockedMutexLeadsToTermination)
...
-        sutNonRecursive.lock();
-        isLockFinished.store(true);
-        sutNonRecursive.unlock();
+        isTryLockSuccessful = mutex.try_lock();
+        test->signalThreadReady();
     });
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(500));
-    EXPECT_THAT(isLockFinished.load(), Eq(false));
+    test->waitForThread();
```
elBoberido, line 135: this can be replaced  by `lockThread.join();`

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -100,38 +122,63 @@ TEST_F(Mutex_test, CallingDestructorOnLockedMutexLeadsToTermination)
...
+void lockedMutexBlocks(Mutex_test* test, iox::posix::mutex& mutex)
+{
+    mutex.lock();
 
-    std::thread lockThread([&] { isTryLockSuccessful = sutRecursive.try_lock(); });
+    std::thread lockThread([&] {
+        test->signalThreadReady();
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(500));
+        auto start = std::chrono::high_resolution_clock::now();
```
elBoberido, line 159: The usage of `high_resolution_clock` should be avoided https://en.cppreference.com/w/cpp/chrono/high_resolution_clock
Please use `steady_clock` instead

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -100,38 +122,63 @@ TEST_F(Mutex_test, CallingDestructorOnLockedMutexLeadsToTermination)
...
+        mutex.unlock();
+        auto end = std::chrono::high_resolution_clock::now();
+        EXPECT_THAT(test->getDuration(start, end), Gt(test->WAIT_IN_NS));
+    });
 
-    EXPECT_THAT(isTryLockSuccessful.load(), Eq(false));
+    test->waitForThread();
 
-    sutRecursive.unlock();
+    std::this_thread::sleep_for(std::chrono::milliseconds(test->WAIT_IN_MS));
```
elBoberido, line 168: Please use `std::chrono::milliseconds` for the `WAIT_IN_MS`/`WAIT_DURATION` then you can do a plain `EXPECT_THAT(end - start, Gg(test->WAIT_DURATION));`

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -100,38 +122,63 @@ TEST_F(Mutex_test, CallingDestructorOnLockedMutexLeadsToTermination)
...
+    mutex.lock();
 
-    std::thread lockThread([&] { isTryLockSuccessful = sutRecursive.try_lock(); });
+    std::thread lockThread([&] {
+        test->signalThreadReady();
 
-    std::this_thread::sleep_for(std::chrono::milliseconds(500));
+        auto start = std::chrono::high_resolution_clock::now();
+        mutex.lock();
+        mutex.unlock();
```
elBoberido, line 161: I just hope this doesn't get optimized out in release builds

iceoryx_hoofs/test/moduletests/test_posix_posix_call.cpp
```diff
@@ -473,3 +479,39 @@ TEST_F(PosixCall_test, CallingFunctionWithMultipleFailureReturnValuesWhereFailur
...
 }
+
+TEST_F(PosixCall_test, ErrnoIsSetFromReturnValueWhenFunctionHandlesErrnosInReturnValue_GoodCase)
+{
+    internal::CaptureStderr();
+
+    constexpr int RETURN_VALUE = 41;
+
+    iox::posix::posixCall(returnValueIsErrno)(RETURN_VALUE)
+        .failureReturnValue(RETURN_VALUE - 1)
```
elBoberido, line 490: Just a question, should this also be done for `successReturnValue`?

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
elfenpiff: You are right but in my clang-tidy prints here a warning and then I let this thing fix it by clang-tidy since it is somehow our coding guideline.
I need to look for some clang-tidy rule where this rule only applies to non public members.

iceoryx_hoofs/test/moduletests/test_posix_posix_call.cpp
elfenpiff: Yes of course, thanks!

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
elBoberido: we already use this style for publisher/subscriber options. Hopefully it's not to hard to find a solution

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -100,38 +120,61 @@ TEST_F(Mutex_test, CallingDestructorOnLockedMutexLeadsToTermination)
...
+        EXPECT_TRUE(mutex.lock());
+        EXPECT_TRUE(mutex.unlock());
+        auto end = std::chrono::steady_clock::now();
+        EXPECT_THAT(end - start, Gt(WAIT_IN_MS));
+    });
+
+    test->waitForThread();
 
-    sutRecursive.unlock();
+    std::this_thread::sleep_for(std::chrono::milliseconds(WAIT_IN_MS));
```
elBoberido, line 164: this is superfluous, it's already `chrono::milliseconds`

elBoberido (approved):

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
```diff
@@ -34,7 +34,7 @@ class Mutex_test : public Test
     void SetUp() override
     {
         internal::CaptureStderr();
-        m_deadlockWatchdog.watchAndActOnFailure([] { std::terminate(); });
+        deadlockWatchdog.watchAndActOnFailure([] { std::terminate(); });
```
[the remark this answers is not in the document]
elBoberido: assert just return and does not unblock the deadlock

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
MatthiasKillat: It will terminate the whole executable as far as I can see. There may be no other way, but consider trying what I wrote in the comment at line 37.

.clang-tidy
```diff
@@ -7,7 +7,8 @@ CheckOptions:
   - { key: readability-identifier-naming.MethodCase,              value: camelBack }
   - { key: readability-identifier-naming.FunctionCase,            value: camelBack }
   - { key: readability-identifier-naming.NamespaceCase,           value: lower_case }
-  - { key: readability-identifier-naming.MemberPrefix,            value: m_ }
+  - { key: readability-identifier-naming.PrivateMemberPrefix,     value: m_ }
```
dkroenke, line 10: To which clang-tidy version refers this change?

dkroenke (approved):

.clang-tidy
elfenpiff: @dkroenke This should be available in clang-tidy 10, see: https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-identifier-naming.html

At the moment I would not yet enforce this clang-tidy file since needs to mature a little while longer. But I strongly suggest that we all integrate clang-tidy into our IDE so that this file can mature faster.

iceoryx_hoofs/test/moduletests/test_posix_mutex.cpp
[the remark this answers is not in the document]
MatthiasKillat: Right.We could in theory still do this in another way by essentially creating a NO_DEATH test which under normal conditions expects not to terminate but can terminated like here due to timeout. I did something like this some  time ago (not with timeout) in the C binding with an invalid options test case, Gtest has the necessary macro infrastructure. 

This would work (I think) because those are forked and if we terminate the forked test it would fail and a blocked mutex is of no concern since thre process is terminated anyway.

But it is not worth it to do so, I just think it is possible in principle.

Issue #208, opened 2020-07-24
Additional functionality for `cxx::string`


elfenpiff:
## Brief feature description

We require additional methods so that `cxx::string` is usable like `std::string`

## Detailed information
 * add method `cxx::string::clear` to clear the string
 
 * add method `cxx::string::operator[]()` to access a specific character of the string

 * assignment of `cxx::string` of different capacity
    ```
    iox::cxx::string<100> a;
    iox::cxx::string<29> b;
    a = b; // this should work

    iox::cxx::string<50> c(b); // this should work
    iox::cxx::string<50> c(a); // this should not work
    ```

  * `operator+=` and `operator+` for easy concatenation
    ```
    iox::cxx::string<100> a("hello");
    iox::cxx::string<100> b("world);

    a += b; // this should work
    iox::cxx::string<100> c = a + "world"; // this should work
    ```

  * `substr`
    ```
    iox::cxx::string<100> a("hello world");
    auto helloString = a.substr(0, 5); // should return iox::cxx::string<100> which contains "hello"
    auto worldString = a.substr(6); // should return iox::cxx::string<100> which contains "world"
    ```

  * `find`, `find_first_of`, `find_last_of` optionally

## Tasks

- [x] `find`, `find_first_of`, `find_last_of`
- [x] `operator+=`, `operator+`
- [x] assignment of `cxx::string` of different capacity
- [x] comparision operator between `cxx::string` and string literal or `char[]`
- [x] `operator[]`
- [x] `cxx::string::clear`
- [x] `insert`
- [ ] `append` + `concatenate` of simple chars
- [x] check if it's possible to have a `constexpr cxx::string`

## Refactoring tasks

- ~~[ ] check if it makes sense to define `using Self = string<Capacity>` in the header and use it within methods~~ - only two occurrences
- ~~[ ] in tests: replace `/// @note` with `// BEGIN tests for method xyz` to enable code folding on some IDEs~~ - seems to work only for Eclipse IDE and KDevelop, therefore postponed

elBoberido:
@FerdinandSpitzschnueffler, does the c'tor automatically infer the length of string literals or do we have to use `TruncateToCapacity`?
It would be nice to be able to write `cxx::string<10> minion("Larry");` instead of `cxx::string<10> minion(TruncateToCapacity, "Larry");`

FerdinandSpitzschnueffler:
[...]

PR #815, opened 2021-05-26
Iox #33 implement tests for unix domain socket

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
This PR introduces unit tests for the unix domain sockets and abstracted the posix calls further. All socket related calls have an `iox_` prefix, this was required since in windows some calls cannot be mapped one to one.

## References

- #33

### DIFF
--- a/iceoryx_examples/iceperf/uds.cpp
+++ b/iceoryx_examples/iceperf/uds.cpp
@@ -90,7 +90,7 @@ void UDS::initFollower() noexcept
 void UDS::init() noexcept
 {
     // init subscriber
-    iox::posix::posixCall(socket)(AF_LOCAL, SOCK_DGRAM, 0)
+    iox::posix::posixCall(iox_socket)(AF_LOCAL, SOCK_DGRAM, 0)
         .failureReturnValue(ERROR_CODE)
         .evaluate()
         .and_then([this](auto& r) { m_sockfdSubscriber = r.value; })
@@ -99,7 +99,7 @@ void UDS::init() noexcept
             exit(1);
         });
 
-    iox::posix::posixCall(bind)(
+    iox::posix::posixCall(iox_bind)(
         m_sockfdSubscriber, reinterpret_cast<struct sockaddr*>(&m_sockAddrSubscriber), sizeof(m_sockAddrSubscriber))
         .failureReturnValue(ERROR_CODE)
         .evaluate()
@@ -109,7 +109,7 @@ void UDS::init() noexcept
         });
 
     // init publisher
-    iox::posix::posixCall(socket)(AF_LOCAL, SOCK_DGRAM, 0)
+    iox::posix::posixCall(iox_socket)(AF_LOCAL, SOCK_DGRAM, 0)
         .failureReturnValue(ERROR_CODE)
         .evaluate()
         .and_then([this](auto& r) { m_sockfdPublisher = r.value; })
@@ -125,12 +125,12 @@ void UDS::waitForLeader() noexcept
     constexpr bool TRY_TO_SEND{true};
     while (TRY_TO_SEND)
     {
-        auto sendCall = iox::posix::posixCall(sendto)(m_sockfdPublisher,
-                                                      nullptr,
-                                                      0,
-                                                      0,
-                                                      reinterpret_cast<struct sockaddr*>(&m_sockAddrPublisher),
-                                                      sizeof(m_sockAddrPublisher))
+        auto sendCall = iox::posix::posixCall(iox_sendto)(m_sockfdPublisher,
+                                                          nullptr,
+                                                          0,
+                                                          0,
+                                                          reinterpret_cast<struct sockaddr*>(&m_sockAddrPublisher),
+                                                          sizeof(m_sockAddrPublisher))
                             .failureReturnValue(ERROR_CODE)
                             .ignoreErrnos(ENOENT)
                             .evaluate()
@@ -172,7 +172,7 @@ void UDS::shutdown() noexcept
 
     if (m_sockfdSubscriber != INVALID_FD)
     {
-        iox::posix::posixCall(iox_close)(m_sockfdSubscriber)
+        iox::posix::posixCall(iox_closesocket)(m_sockfdSubscriber)
             .failureReturnValue(ERROR_CODE)
             .evaluate()
             .or_else([](auto& r) {
@@ -236,12 +236,12 @@ void UDS::send(const char* buffer, uint32_t length) noexcept
     // only return from this loop when the message could be send successfully
     // if the OS socket message buffer if full, retry until it is free'd by
     // the OS and the message could be send
-    while (iox::posix::posixCall(sendto)(m_sockfdPublisher,
-                                         buffer,
-                                         length,
-                                         0,
-                                         reinterpret_cast<struct sockaddr*>(&m_sockAddrPublisher),
-                                         sizeof(m_sockAddrPublisher))
+    while (iox::posix::posixCall(iox_sendto)(m_sockfdPublisher,
+                                             buffer,
+                                             length,
+                                             0,
+                                             reinterpret_cast<struct sockaddr*>(&m_sockAddrPublisher),
+                                             sizeof(m_sockAddrPublisher))
                .failureReturnValue(ERROR_CODE)
                .ignoreErrnos(ENOBUFS)
                .evaluate()
@@ -257,7 +257,7 @@ void UDS::send(const char* buffer, uint32_t length) noexcept
 
 void UDS::receive(char* buffer) noexcept
 {
-    iox::posix::posixCall(recvfrom)(m_sockfdSubscriber, buffer, MAX_MESSAGE_SIZE, 0, nullptr, nullptr)
+    iox::posix::posixCall(iox_recvfrom)(m_sockfdSubscriber, buffer, MAX_MESSAGE_SIZE, 0, nullptr, nullptr)
         .failureReturnValue(ERROR_CODE)
         .evaluate()
         .or_else([](auto& r) {
--- a/iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
+++ b/iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
@@ -31,6 +31,7 @@ namespace iox
 namespace posix
 {
 constexpr char UnixDomainSocket::PATH_PREFIX[];
+constexpr uint64_t UnixDomainSocket::MAX_MESSAGE_SIZE;
 
 UnixDomainSocket::UnixDomainSocket() noexcept
 {
--- a/iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
+++ b/iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
@@ -39,7 +39,6 @@ UnixDomainSocket::UnixDomainSocket() noexcept
 }
 
 UnixDomainSocket::UnixDomainSocket(const IpcChannelName_t& name,
-                                   const IpcChannelMode mode,
                                    const IpcChannelSide channelSide,
                                    const size_t maxMsgSize,
                                    const uint64_t maxMsgNumber) noexcept
@@ -54,7 +53,6 @@ UnixDomainSocket::UnixDomainSocket(const IpcChannelName_t& name,
             }
             return UdsName_t(PATH_PREFIX).append(iox::cxx::TruncateToCapacity, name);
         }(),
-        mode,
         channelSide,
         maxMsgSize,
         maxMsgNumber)
@@ -63,7 +61,6 @@ UnixDomainSocket::UnixDomainSocket(const IpcChannelName_t& name,
 
 UnixDomainSocket::UnixDomainSocket(const NoPathPrefix_t,
                                    const UdsName_t& name,
-                                   const IpcChannelMode mode,
                                    const IpcChannelSide channelSide,
                                    const size_t maxMsgSize,
                                    const uint64_t maxMsgNumber IOX_MAYBE_UNUSED) noexcept
@@ -85,12 +82,10 @@ UnixDomainSocket::UnixDomainSocket(const NoPathPrefix_t,
     else
     {
         m_maxMessageSize = maxMsgSize;
-        initalizeSocket(mode)
-            .and_then([this]() { this->m_isInitialized = true; })
-            .or_else([this](IpcChannelError& error) {
-                this->m_isInitialized = false;
-                this->m_errorValue = error;
-            });
+        initalizeSocket().and_then([this]() { this->m_isInitialized = true; }).or_else([this](IpcChannelError& error) {
+            this->m_isInitialized = false;
+            this->m_errorValue = error;
+        });
     }
 }
 
@@ -302,7 +297,7 @@ UnixDomainSocket::timedReceive(const units::Duration& timeout) const noexcept
 }
 
 
-cxx::expected<IpcChannelError> UnixDomainSocket::initalizeSocket(const IpcChannelMode mode) noexcept
+cxx::expected<IpcChannelError> UnixDomainSocket::initalizeSocket() noexcept
 {
     // initialize the sockAddr data structure with the provided name
     memset(&m_sockAddr, 0, sizeof(m_sockAddr));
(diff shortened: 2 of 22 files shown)


iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -204,7 +209,7 @@ cxx::expected<IpcChannelError> UnixDomainSocket::timedSend(const std::string& ms
                                                            const units::Duration& timeout) const noexcept
 {
     constexpr uint64_t NULL_TERMINATOR_SIZE = 1U;
-    if (msg.size() > m_maxMessageSize + NULL_TERMINATOR_SIZE)
+    if (msg.size() >= m_maxMessageSize + NULL_TERMINATOR_SIZE)
```
elBoberido, line 212: Hmm, shouldn't this be just `msg.size > m_maxMessageSize`?

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -523,7 +522,7 @@ IpcChannelError UnixDomainSocket::convertErrnoToIpcChannelError(const int32_t er
 
 bool UnixDomainSocket::isNameValid(const UdsName_t& name) noexcept
 {
-    return !(name.empty() || name.size() < SHORTEST_VALID_NAME || name.size() > LONGEST_VALID_NAME);
```
elBoberido, line 526: Why is this removed?

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -78,36 +125,229 @@ class UnixDomainSocket_test : public Test
 const size_t UnixDomainSocket_test::MaxMsgSize = UnixDomainSocket::MAX_MESSAGE_SIZE;
 constexpr uint64_t UnixDomainSocket_test::MaxMsgNumber;
 
-TEST_F(UnixDomainSocket_test, NonBlockingModeNotSupported)
+TEST_F(UnixDomainSocket_test, UnlinkEmptySocketNameLeadsToInvalidChannelNameError)
```
elBoberido, line 128: Aren't this tests for all IPC channel implementations like the message queue, in order to ensure similar behavior?

iceoryx_examples/iceperf/uds.cpp
```diff
@@ -90,7 +90,7 @@ void UDS::initFollower() noexcept
 void UDS::init() noexcept
 {
     // init subscriber
-    iox::posix::posixCall(socket)(AF_LOCAL, SOCK_DGRAM, 0)
```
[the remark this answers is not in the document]
elfenpiff: If we do not use this abstraction then this example will not run in windows. In general I would agree but here it is just a parameter forwarding - a teeny tiny overhead. Could you live with it or do you think that this would effect the outcome of the benchmark in a measurable way?

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -133,6 +128,16 @@ UnixDomainSocket& UnixDomainSocket::operator=(UnixDomainSocket&& other) noexcept
...
+    {
+        return cxx::error<IpcChannelError>(IpcChannelError::INVALID_CHANNEL_NAME);
+    }
+
+    if (UdsName_t().capacity() < name.size() + UdsName_t(PATH_PREFIX).size())
+    {
+        return cxx::error<IpcChannelError>(IpcChannelError::INVALID_CHANNEL_NAME);
+    }
+
     return unlinkIfExists(NoPathPrefix, UdsName_t(PATH_PREFIX).append(iox::cxx::TruncateToCapacity, name));
```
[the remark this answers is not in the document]
elfenpiff: This is correct and it goes for the user. The `PATH_PREFIX` is set in the header and per default always applied but if the user would like to create an uds in its current folder they can user `unlinkIfExists(NoPathPrefix, "mySocketFile");`

This here is just the implementation avoids code duplication.

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
elfenpiff: You are right, but I also have to reduce the `MAX_MESSAGE_SIZE` by the terminator size since it is always added.

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
elfenpiff: There is only the requirement that you provide some kind of filename, therefore the name is not allowed to be empty. The `UdsName_t` is an alias for `cxx::string<LONGEST_VALID_NAME>` therefore the greater than check is done via the type.

A socket file name with one letter is allowed. Therefore we can reduce the check to `name.empty()`.

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
elfenpiff: The unlink checks are checks for the ipc channel but were not really done there. Therefore, I created them here specific for the unix domain socket where I created sockets etc.

The unix domain socket also offers features which are not used by the IPC channel and I wanted to clearly test them in a unit test.

iceoryx_examples/iceperf/uds.cpp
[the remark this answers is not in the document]
elBoberido: Well, I don't mind. You have to ask budrus.

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,282 @@ class UnixDomainSocket_test : public Test
...
+    for (uint64_t i = 0U; i < UnixDomainSocket::LONGEST_VALID_NAME - strlen(UnixDomainSocket::PATH_PREFIX) + 1; ++i)
+    {
+        longSocketName.append(cxx::TruncateToCapacity, "o");
+    }
+    auto ret = UnixDomainSocket::unlinkIfExists(longSocketName);
+    ASSERT_TRUE(ret.has_error());
+    EXPECT_THAT(ret.get_error(), Eq(IpcChannelError::INVALID_CHANNEL_NAME));
+}
+
+TEST_F(UnixDomainSocket_test, UnlinkExistingSocketLeadsIsSuccessful)
```
FerdinandSpitzschnueffler, line 154: rename to UnlinkExistingSocketIsSuccessful

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: So `"/"` is a valid name ;)

elfenpiff: Of course ... ;) I think in Windows it actually is maybe?!.
But I can extend the test that only the following characters are allowed `[a-zA-Z0-9.,-]` but it wasn't there in the first place...

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,282 @@ class UnixDomainSocket_test : public Test
...
+    EXPECT_THAT(ret.get_error(), Eq(IpcChannelError::INVALID_CHANNEL_NAME));
+}
+
+TEST_F(UnixDomainSocket_test, UnlinkTooLongSocketNameWithPathPrefixLeadsToInvalidChannelNameError)
+{
+    UnixDomainSocket::UdsName_t longSocketName;
+    for (uint64_t i = 0U; i < UnixDomainSocket::LONGEST_VALID_NAME - strlen(UnixDomainSocket::PATH_PREFIX) + 1; ++i)
+    {
+        longSocketName.append(cxx::TruncateToCapacity, "o");
+    }
```
elBoberido, line 148: maybe we need a `string::insert` method

FerdinandSpitzschnueffler (dismissed):

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,282 @@ class UnixDomainSocket_test : public Test
...
 
-TEST_F(UnixDomainSocket_test, SendingOnServerLeadsToError)
+void sendOnServerLeadsToError(const sendCall_t& send)
 {
     cxx::string<10> message{"Foo"};
-    Duration duration{1_ms};
-    auto result = server.timedSend(message, duration);
+    auto result = send(message);
     EXPECT_TRUE(result.has_error());
     ASSERT_THAT(result.get_error(), Eq(IpcChannelError::INTERNAL_LOGIC_ERROR));
```
elBoberido, line 178: This does not work as intended. The `ASSERT_THAT` returns from this function but on the caller side the execution is continued.
This will terminate the test process.
```cpp
void foo() {
    ASSERT_TRUE(false);
}

TEST(Foo, foo) {
    foo();
    auto n = 1/0;
}
```
Currently this function is not used in a harmful way, though

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,282 @@ class UnixDomainSocket_test : public Test
...
 }
 
-TEST_F(UnixDomainSocket_test, ReceivingOnClientLeadsToError)
+TEST_F(UnixDomainSocket_test, TimedSendOnServerLeadsToError)
```
elBoberido, line 181: Can you add the reason why it leads to an error?

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,282 @@ class UnixDomainSocket_test : public Test
...
+    sendOnServerLeadsToError([&](auto& msg) { return server.send(msg); });
+}
+
+void successfulSendAndReceive(const std::string& message, const sendCall_t& send, const receiveCall_t& receive)
+{
+    ASSERT_FALSE(send(message).has_error());
+
+    auto receivedMessage = receive();
+    ASSERT_FALSE(receivedMessage.has_error());
+    EXPECT_EQ(message, *receivedMessage);
```
elBoberido, line 197: Theoretically we do not yet know if our abstraction is using the sockets. It's possible to use a global variable for data transfer. Can we have one test using the POSIX API for send and receive directly?

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,282 @@ class UnixDomainSocket_test : public Test
...
+
+void receivingOnClientLeadsToError(const receiveCall_t& receive)
+{
+    auto result = receive();
     EXPECT_TRUE(result.has_error());
     ASSERT_THAT(result.get_error(), Eq(IpcChannelError::INTERNAL_LOGIC_ERROR));
 }
-} // namespace
+
+TEST_F(UnixDomainSocket_test, ReceivingOnClientLeadsToErrorWithReceive)
```
elBoberido, line 313: please add the cause of the error

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: You touched it ;)

That check is not sufficient since you can have `/tmp/mysocket`. The previous check was also not sufficient since this should also not be allowed `/////`. Maybe we just need to check if the last character isn't a path separator?

elBoberido: ... oh, and we need tests for this

iceoryx_examples/iceperf/uds.cpp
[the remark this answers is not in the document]
elfenpiff: @budrus said its alright!

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -274,7 +278,18 @@ UnixDomainSocket::timedReceive(const units::Duration& timeout) const noexcept
...
+#if defined(__APPLE__)
+    if (tv.tv_sec != 0 || tv.tv_usec != 0)
+    {
+        std::cerr
+            << "socket: \"" << m_name
+            << "\", timedReceive with a timeout != 0 is not supported on MacOS. timedReceive will behave like receive "
+               "instead."
+            << std::endl;
+    }
+#endif
```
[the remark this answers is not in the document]
elfenpiff: You are completely right, but this would be a more complex task since the timeout is not directly provided to the function, it is set in a call before via `setsockopt`.
I created an issue for this: https://github.com/eclipse-iceoryx/iceoryx/issues/823

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
elfenpiff: @FerdinandSpitzschnueffler yes we do. Also it would be nice if we could `concatenate`, `append` simple char's.

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
elfenpiff: You are right but here an `EXPECT_THAT` suffices.

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
elfenpiff: You are right, this would make send ... (for all posix wrapper :( I see the dawn of a new issue)

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
FerdinandSpitzschnueffler: Added these methods to https://github.com/eclipse-iceoryx/iceoryx/issues/208
Feel free to add more.

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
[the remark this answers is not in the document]
elBoberido: can this warning be printed in the `iox_setsockopt` implementation in the mac platform?

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
elBoberido: It has to be done once, though. When we know that e.g. send works, we can use that to test timedReceive, etc.

elBoberido (commented):
Can you please improve the naming on the two tests I mentioned in my last PR? Except that, this looks good

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
elBoberido: It seems this will be fixed with the next PR

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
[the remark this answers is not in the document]
elfenpiff: That was my idea. The warning stays but is then printed via platforms.

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
elfenpiff: Is already fixed with tests n stuff in the following PR: https://github.com/eclipse-iceoryx/iceoryx/pull/820

iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -69,45 +75,289 @@ class UnixDomainSocket_test : public Test
...
     cxx::string<10> message{"Foo"};
-    Duration duration{1_ms};
-    auto result = server.timedSend(message, duration);
+    auto result = send(message);
     EXPECT_TRUE(result.has_error());
-    ASSERT_THAT(result.get_error(), Eq(IpcChannelError::INTERNAL_LOGIC_ERROR));
+    EXPECT_THAT(result.get_error(), Eq(IpcChannelError::INTERNAL_LOGIC_ERROR));
+}
+
+TEST_F(UnixDomainSocket_test, TimedSendOnServerLeadsToError)
```
elBoberido, line 186: Oh, how about naming this `ServerCanOnlyReceiveThereforeTimedSendLeadsToError` and similar with the other 3 tests?

elBoberido (approved):
If you want to keep the names, you can merge this

FerdinandSpitzschnueffler (approved):

PR #819, opened 2021-05-27
Iox #27 request response design document

### LINKED ISSUE #27: Request/Response communication with iceoryx

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This is a design document for request-response. Initially I wanted to create a small PoC first, but since the progress is quite slow I came to the conclusion to discuss this now rather than later.

## References

- Partly closes #27


doc/design/request_response_communication.md
```diff
@@ -0,0 +1,102 @@
...
+
+![typed API](diagrams/request_response_typed_api.svg)
+
+#### Untyped API
+
+![untyped API](diagrams/request_response_untyped_api.svg)
+
+#### Client Port
+
+![client port](diagrams/request_response_client_port.svg)
```
mossmaurice, line 57: Could you move the `puml` and `svg` files into a folder called `request_response`? I did the same in #784 Otherwise this folder will become quite large soon ;-)

doc/design/diagrams/request_response_server_port.puml
```diff
@@ -0,0 +1,37 @@
...
+  stopOffer(): void
+  isOffered(): bool
+  hasClients(): bool
+  getRequest(): expected<const RequestHeader*, ChunkReceiveResult>
+  releaseRequest(response Header: const RequestHeader*): void
+  hasNewRequests(): void
+  hasLostRequestsSinceLastCall(): bool
+  allocateResponse(userPayloadSize: uint32_t, userPayloadAlignment: uint32_t): expected<ResponseHeader*, AllocationError>
+  sendResponse(requestHeader: ResponseHeader): void
+  freeResponset(requestHeader: ResponseHeader): void
```
FerdinandSpitzschnueffler, line 22: typo: `freeResponse`

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,102 @@
...
+
+![client and server options](diagrams/request_response_options.svg)
+
+The client and server options can be used to control certain aspects of the clients and servers.
+Beside setting the capacity of the queues and defining whether a client should be connected and a server offering on creation,
+the behaviour for a slow client and server can be defined.
+A client can ask a server to block if its response queue is full.
+The server will obey if the corresponding `ClientTooSlowPolicy` is set to `WAIT_FOR_CLIENT`.
+If the options don't match, the client will not be connected to the server.
+
```
FerdinandSpitzschnueffler, line 96: And the server can ask the client to block if the request queue is full? Therefore the `ServerTooSlowPolicy`?

doc/design/diagrams/puml_github_workflow.svg
```diff
@@ -1,6 +1,6 @@
...
+link Iceoryx_Contributor to 1. Create Diagram in PlanUML Editor--><path d="M338.57,176.53 C403.77,151.07 507.36,112.16 599,84.5 C629.38,75.33 662.79,66.68 693.36,59.31 " fill="none" id="Iceoryx_Contributor-to-1. Create Diagram in PlanUML Editor" style="stroke:#A80036;stroke-width:1.0;"/><polygon fill="#A80036" points="698.24,58.14,688.558,56.3346,693.3761,59.2987,690.412,64.1168,698.24,58.14" style="stroke:#A80036;stroke-width:1.0;"/><!--MD5=[66be488fc23afe3f548d0b8c15b84b21]
```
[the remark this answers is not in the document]
elBoberido: hahaha, this is generated but I will fix the source

doc/design/diagrams/request_response_server_port.puml
elBoberido: and I also have a C&P error at `requestHeader: ResponseHeader`

doc/design/request_response_communication.md
elBoberido: The Server doesn't ask the client, since its all decided on the client side but the option is called `ServerTooSlowPolicy`

doc/design/diagrams/request_response_options.puml
```diff
@@ -0,0 +1,34 @@
...
+}
+
+enum ClientTooSlowPolicy {
+    WAIT_FOR_CLIENT,
+    DISCARD_OLDEST_DATA
+}
+
+enum ServerTooSlowPolicy {
+    WAIT_FOR_SERVER,
+    DISCARD_OLDEST_DATA
```
elBoberido, line 31: For publish-subscribe we already have `QueueFullPolicy` and `SubscriberTooSlowPolicy`. Since they are also used in the `ChunkSender` it's cumbersome to always do the conversion. I would suggest to combine all this enums into
```
enum QueueFullPolicy {
    BLOCK_PARTICIPANT, // or BLOCK_WRITER
    DISCARD_OLDEST_DATA
}
```
and
```
enum ParticipantTooSlowPolicy { // or ReaderTooSlowPolicy
    WAIT_FOR_PARTICIPANT, // or WAIT_FOR_READER
    DISCARD_OLDEST_DATA
}
```
The corresponding member would still be called `subscriberTooSlowPolicy`. What do you think?

elBoberido: Okay, it's now
```cpp
enum QueueFullPolicy {
    BLOCK_PRODUCER,
    DISCARD_OLDEST_DATA
};

enum ConsumerTooSlowPolicy {
    WAIT_FOR_CONSUMER,
    DISCARD_OLDEST_DATA
};
```

doc/design/diagrams/request_response/typed_api.puml
```diff
@@ -0,0 +1,100 @@
+@startuml
```
elfenpiff, line 1: For Larry I wrote myself also some minimalistic client server thingy. Here I think we orient us to much on the existing publisher and subscriber port which are in this context a little bit cumbersome to use.

Here I drafted a code snippet on how I as a user would like to use the typed API.
```cpp
// Client side
// usage 1, wait directly for response of a request
Client<int, int> client;
auto response = client.sendCopyOf(123).waitForResponse();

// usage 2, acquire response from request handle
Client<int, int> client;
auto request = client.loan();
*request = 123;
request.send();
if ( request.hasResponse() ) // do stuff
// or wait here for a response
auto response request.waitForResponse();


// Server side.
// usage 1 respond via request (introduce answer method)
Server<int, int> server;
server.take().and_then([&](auto & request){
    request.answerWithCopyOf(*request + 5);
});

// usage 2 request comes with an already loan chunk for the response
Server<int, int> server;
auto request = server.take();
*request.getResponse().getDataPtr() = *request + 5;
*request.getResponse().send();
```

Maybe it is out of scope of this design document but for the public API (typed) I would start with a (pseude) code example to see how it look and feels. And I would optimize the typed API for easy usability - if the user would like to have very fine grained control over everything they can use the untyped API.

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+#### Untyped API
+
+![untyped API](diagrams/request_response/untyped_api.svg)
+
+Similar to the the typed API, `loan` takes a pointer to a `RequestHeader` to populate the `ResponseHeader` with the correct settings.
+
+#### Client Port
+
+![client port](diagrams/request_response/client_port.svg)
```
elfenpiff, line 62: I think the `loan` requires here also a size argument for the request?

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+![typed API](diagrams/request_response/typed_api.svg)
+
+Since the `Response` is tied to a specific `Request`, the `loan` method takes a `Request` to populate the `Response` with the correct settings.
+Alternatively, a `Request` could have a `createResponse` method which returns a `Response` with the correct settings.
+
+#### Untyped API
+
+![untyped API](diagrams/request_response/untyped_api.svg)
+
+Similar to the the typed API, `loan` takes a pointer to a `RequestHeader` to populate the `ResponseHeader` with the correct settings.
```
elfenpiff, line 58: I think here we need an additional argument to set the response size. The untyped api may sends different response sizes to the same request?

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+    ...
+    cxx::expected<uint64_t, Error> deliverToPort(UniquePortId uniquePortId, uint64_t lastKnownQueueIndex);
+    ...
+};
+```
+
+The sequence ID is used to match a response to a specific request if multiple requests are invoke asynchronously.
+Depending on the client and server options, a request might be dropped or a server could have a worker pool
+which results in sending the responses in a different order than the request were received.
+The sequence ID must be set by the user and also checked by the user on response.
```
elfenpiff, line 98: Do you mean with user the `ClientPortUser` or `ServerPortUser` or the end user of that API?

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+
+The `RequestHeader` has also the option to specify a message as fire and forget, which means it won't get a response to this request.
```
elfenpiff, line 100: What would be a typical use case for this? And why would generating a response in this use case be cumbersome or unhandy?

doc/design/request_response_communication.md
elBoberido: yes, C&P error from the typed API

doc/design/request_response_communication.md
elBoberido: The user of the API. This is necessary to map to the ROS 2 services.

doc/design/request_response_communication.md
elBoberido: Sending a status like keep alive or similar where the client does not care about a response

elfenpiff: And why not use a publisher for that? A fire and forget request seems to me a typical publisher use case.

elfenpiff (commented):
@elBoberido how do you ensure that an iceoryx subscriber does not subscribe to a server? Or a iceoryx publisher gets samples from a client?

tools/generate_plantuml_svg.sh
```diff
@@ -43,4 +45,17 @@ then
     exit
 fi
 
-java -jar $PLANTUML_DIR/plantuml.jar -nometadata -tsvg $EXPORT_DIR/*
+
+# set number of cores for building
+if [[ "$OSTYPE" == "linux-gnu"* ]] || [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
```
dkroenke, line 50: Just out of curiosity: How does it take to generate the svg files?

elBoberido: How long? About 3 seconds with 12 threads for 12 svg

elBoberido:
@elfenpiff the same way a subscriber doesn't subscribe to another subscriber and a publisher doesn't publish to a different publisher. Only matching ports are connected. Remember the `handlePublisherPorts`/`handleSubscriberPorts` in the PortManager?

doc/design/request_response_communication.md
elBoberido: Because there can be many subscriber for one publisher but only one server for a client. Furthermore, we need this to support bridges to other technologies who also have fire and forget RPCs

doc/design/diagrams/request_response/typed_api.puml
elBoberido: I think there will be one or two iterations over the API once we write a few examples.

While `client.send().wait()` looks nice, I think that's what we created the `WaitSet`/`Listener` for. So these should be used for blocking calls.

The `request.hasResponse` is also not always feasible since it is possible to send multiple requests and get out of order responses. This has also to be taken into account.

The `answerWithCopy`, etc. convenience methods can of course be added. I omitted them because they are not crucial for a first implementation.

The `request.getResponse().send()` only works for non fire and forget request, but could be done by `getResponse` returning an optional. I also thought about this as alternative to `loan` and kind of like it. The response already needs to know the server in order for the sample deleter, so this might work.

@Indra5196 what do you think about the suggestions of elfenpiff?

tools/generate_plantuml_svg.sh
elBoberido: Just in case you want hard numbers ;)
1 thread
```
Executed in    2,98 secs   fish           external 
   usr time    7,39 secs  621,00 micros    7,39 secs 
   sys time    0,23 secs  148,00 micros    0,23 secs
```
12 threads
```
Executed in    1,76 secs   fish           external 
   usr time   10,34 secs  250,00 micros   10,34 secs 
   sys time    0,40 secs   59,00 micros    0,40 secs
```

doc/design/diagrams/request_response/client_and_server_options.puml
```diff
@@ -0,0 +1,29 @@
...
+
+enum QueueFullPolicy {
+  BLOCK_PRODUCER,
+  DISCARD_OLDEST_DATA
+}
+
+enum ConsumerTooSlowPolicy {
+  WAIT_FOR_CONSUMER,
+  DISCARD_OLDEST_DATA
+}
```
elBoberido, line 27: btw, this would replace the current `QueueFullPolicy` and `SubscriberTooSlowPolicy`

doc/design/diagrams/request_response/typed_api.puml
Indra5196: Hi @elBoberido @elfenpiff 
Yup. Sounds good for a start!!

elfenpiff:
> @elfenpiff the same way a subscriber doesn't subscribe to another subscriber and a publisher doesn't publish to a different publisher. Only matching ports are connected. Remember the `handlePublisherPorts`/`handleSubscriberPorts` in the PortManager?

I see a problem arise on the horizon. Let's say someone uses our new magnificent service discovery, discovers a service and wants to subscribe to it - but it does not work since it is a server where a subscriber cannot connect to.

This means we have a service registry where a colorful mixture of publish/subscribe services and client/server services are contained which have nothing to do with each other. We can create a publisher with the same service description as a server and they both can have subscriber and clients which are completely independent of each other.

I do not like this and I bet we receive then some issues like: unable to connect to service  or publisher.

On simple solution would be to have two separate service registries one called ServiceRegistry (Client/Service) and another PublisherRegistry (Publish/Subscribe) since they are actually completely different. But this shouldn't be a big issue when we plan it right upfront since we can reuse the service registry for this use case.

@budrus @mossmaurice

doc/design/diagrams/request_response/typed_api.puml
```diff
@@ -0,0 +1,98 @@
...
+  get(): T*
+  get() const: const T*
+  getChunkHeader(): ChunkHeader*
+  getChunkHeader() const: const ChunkHeader*
+}
+
+note right of SmartChunk: ForProducerOnly and ForConsumerOnly type traits\nrestrict access to some ctors and methods,\nlike no non-const access to payload
+
+interface PublisherInterface<T, H> {
+  pubish(sample: Sample<T, H>): void
```
mossmaurice, line 53: Nitpick: typo

mossmaurice:
> I see a problem arise on the horizon. Let's say someone uses our new magnificent service discovery, discovers a service and wants to subscribe to it - but it does not work since it is a server where a subscriber cannot connect to.

Good point!

> This means we have a service registry where a colorful mixture of publish/subscribe services and client/server services are contained which have nothing to do with each other. We can create a publisher with the same service description as a server and they both can have subscriber and clients which are completely independent of each other. It is like having two distinct services with an identical service identifier (which sounds somehow bad).

I agree, that's quite bad. Why should all request/response server/clients be discoverable? I can only think of use-cases where a user would like to know the available RPC calls in the system aka available `Response` 's from `ServerPort`'s. Hence, a service discovery would need special treatment.

> I do not like this and I bet we receive then some issues like: unable to connect to service or publisher.
> 
> On simple solution would be to have two separate service registries one called ServiceRegistry (Client/Service) and another PublisherRegistry (Publish/Subscribe) since they are actually completely different. But this shouldn't be a big issue when we plan it right upfront since we can reuse the service registry for this use case.

Do we actually need two separate `ServiceRegistry`'s? From my point of view a `ServiceRegistry` could be something abstract that handles both publish/subscribe and reqest/response. In ROS 2 the request/response mechanism [is called](https://docs.ros.org/en/foxy/Tutorials/Services/Understanding-ROS2-Services.html) "services". Currently my feeling is that we should extend the current `ServiceRegistry` with special handling for request/response e.g. only storing the possible requests of servers. Services, then could be both, either a publish/subscribe data topic or an available RPC call.

Side note: ROS 2 also has [actions](https://docs.ros.org/en/foxy/Tutorials/Understanding-ROS2-Actions.html).

elBoberido:
It is not yet decided how to integrate this into the service registry. It is left open especially because it is currently refactored. But it is actually a non-issue to have both of them in the same service registry. Just add a flag if it's a pub-sub or req-res service. Users could then request all services or just a specific subset.

doc/design/diagrams/request_response/typed_api.puml
```diff
@@ -0,0 +1,98 @@
...
+}
+
+class Publisher {
+}
+}
+
+class SmartChunk<TransmissionInterface, T, H> {
+  SmartChunk(chunkUniquePtr: unique_ptr<T>&&, sender: TransmissionInterface&)
+  SmartChunk(chunkUniquePtr: unique_ptr<const T>&&)
+  operator->(): T*
```
elfenpiff, line 39: We had once discussed that every container/data type which may contain something or contains either the one value or the other uses additionally the `and_then`/`or_else` pattern. Data types like `optional`, `expected`, `unique_ptr` and of course our `SmartChunk` here.

I would extend this interface with the following methods:
```cpp
SmartChunk& and_then(const cxx::function_ref<void(T&)> & callback) noexcept;
SmartChunk& and_then(const cxx::function_ref<void(const T&)> & callback) const noexcept;
SmartChunk& or_else(const cxx::function_ref<void()> & callback) const noexcept;
```

doc/design/diagrams/request_response/typed_api.puml
elBoberido: We might also restrict the the server options.

@Indra5196 do we need to support fire and forget in parallel to response? I mean, does a server need to do fire and forget and response at the same time?

doc/design/diagrams/request_response/typed_api.puml
elBoberido: yup, I meant `punish` ;)

doc/design/diagrams/request_response/typed_api.puml
elBoberido: I would decline this request. If you have a null-able type use `cxx::optional<T>` instead. We should not reinvent null-able types all over the place since that essentially means you always have to do a null check. If it is wrapped in an optional you clearly state your intention.

elBoberido: Oh, and I'm aware that due to to C++ move semantic we are left with undefined but specified objects. But this objects are usually confided in a local scope where it is easy to reason about the movement and it can also easily be tested that moved objects are not used after the move operation.

doc/design/diagrams/request_response/typed_api.puml
```diff
@@ -0,0 +1,98 @@
...
+  operator->() const: const T*
+  operator*(): T&
+  operator*() const: const T&
+  operator bool() const
+  get(): T*
+  get() const: const T*
+  getChunkHeader(): ChunkHeader*
```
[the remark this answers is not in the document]
elBoberido: Agreed. But this changes the current API of the `Sample` and we need to decide if we want to put the burden on porting to our users. I think `get` is not used much but `getChunkHeader` is. @MatthiasKillat @budrus what do you think?

elBoberido:
This does no yet have to be merged, but if nobody disagrees with the refactoring of the current `Sample` into `SmartChunk`, `Sample`, `Request` and `Response`, we could start with the refactoring.

doc/design/diagrams/request_response/typed_api.puml
Indra5196: Like in the case where we are handling multiple clients, we might need to do this. Will this require locking of the request handler on the server side?

doc/design/diagrams/request_response/client_and_server_options.puml
elBoberido: @elfenpiff @MatthiasKillat @mossmaurice @dkroenke @FerdinandSpitzschnueffler to continue with the implementation I would refactor the stuff above first. This changes the public API, though. Are you okay with this part of the design?

elBoberido: @elfenpiff @MatthiasKillat @mossmaurice @dkroenke @FerdinandSpitzschnueffler to continue with the implementation I would refactor the stuff above first. This changes the public API, though. Are you okay with this part of the design?

elBoberido:
@elfenpiff @dkroenke @mossmaurice @FerdinandSpitzschnueffler @MatthiasKillat could you please check if this could be merged? Maybe as draft? Since some parts are already worked on, it's quite cumbersome to keep this up to date. It would be easier if this could be done in the respective branches.

Open points are introspection and service discovery, but that's not part of this design

FerdinandSpitzschnueffler (approved):

doc/design/diagrams/puml_github_workflow.svg
```diff
@@ -1,6 +1,6 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" contentScriptType="application/ecmascript" contentStyleType="text/css" height="405px" preserveAspectRatio="none" style="width:1007px;height:405px;background:#EEEBDC;" version="1.1" viewBox="0 0 1007 405" width="1007px" zoomAndPan="magnify"><defs><filter height="300%" id="f18ey02a78uejz" width="300%" x="-1" y="-1"><feGaussianBlur result="blurOut" stdDeviation="2.0"/><feColorMatrix in="blurOut" result="blurOut2" type="matrix" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 .4 0"/><feOffset dx="4.0" dy="4.0" in="blurOut2" result="blurOut3"/><feBlend in="SourceGraphic" in2="blurOut3" mode="normal"/></filter></defs><g><!--MD5=[81da3b3e20e7b00bc69cac0acd70a916]
```
MatthiasKillat, line 1: We need to think about whether we want to actually store svg files in git. Small changes in the file used to generate it lead to large changes in the svg. I do not know whether we plan to have many svg files in the repository and if so, if they are mostly static it is less problematic from a storage point of view.

dkroenke: It is recommended to use the script in the tools folder for generating the vectorgraphics: https://github.com/eclipse-iceoryx/iceoryx/blob/master/tools/generate_plantuml_svg.sh
With that, a defined version of PlantUML is used to create the svg and therefore reduce the diff in the svg files.
Personally i think that the diff in this file is not that large (only the size and some coordinates differ) and Github will be able to tackle that.

elBoberido: The files are also quite small and improve the comprehension alot

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+
+![simple class diagram](diagrams/request_response/overview_class.svg)
+
+The `Client` and `Server` are reusing the `ChunkSender` and `ChunkReceiver` building blocks. The `Client` uses a `ChunkSender` to send requests and a `ChunkReceiver` to get the responses while the `Server` uses a `ChunkReceiver` to get the requests and a `ChunkSender` to send the responses.
+
+#### Typed API
+
+![typed API](diagrams/request_response/typed_api.svg)
+
+Since the `Response` is tied to a specific `Request`, the `loan` method takes a `Request` to populate the `Response` with the correct settings.
```
MatthiasKillat, line 51: I could think about some slightly different alternative mechanism as follows if the response type is known to the client, this would work more like a future (and since we are using asynchronous RPC this may be useful). Both methods do not really exclude each other.
1) Client has created request data D (arguments for the RPC call basically).
2) Client wraps this in some kind of Request-Response structure RR, which already contains room for the response and enough information for the server about what to do with the arguments (i.e. the call information)
3) Client sends RR to Server, but holds on to a reference (chunk) of it
4) Server extracts the arguments from RR and computes the response (or rejects the request, which is also some kind of response), populates the response area of RR and marks the response as ready
4) Client waits until the response is ready, by observing RR (actively or passively) - non-blocking, it can do other things in the meantime
5) Client extracts actual response information once the response is ready and releases RR

Note that this would not directly work outside of shared memory communication, where the response as to be actually send somewhere. API-wise it would still be possible though, just with sending the response under behind the scenes.

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+interface RouDi needs to connect the client to the server once the server offers its service and to cleanup the port resources.
+
+It must be ensured that only one server with a given `ServiceDescription` can run at a time.
+
+#### Request/Response Header
+
+![rpc header](diagrams/request_response/request_response_header.svg)
+
+Since request and response need to encode different meta-information, we also need different header for the messages.
+The common data is aggregated in `RpcBaseHeader` which contains a `UniquePortId` to the port owning the `ClientChunkQueueData_t` and a sequence ID.
```
MatthiasKillat, line 83: Note that by coupling request and response as in my future based proposal (comment above), the sequence ID would probably be unnecessary (the client essentially knows where to find the response, the server knows where to write it). If it issued 3 requests and gets notified of one response, it has to look however which request is ready (in a naive implementation).

But yes, for out of order responses a sequence ID is necessary in the way described here.

doc/design/request_response_communication.md
```diff
@@ -0,0 +1,118 @@
...
+```cpp
+class ChunkDistributor {
+    ...
+    cxx::expected<uint64_t, Error> deliverToPort(UniquePortId uniquePortId, uint64_t lastKnownQueueIndex);
+    ...
+};
+```
+
+The sequence ID is used to match a response to a specific request if multiple requests are invoke asynchronously.
+Depending on the client and server options, a request might be dropped or a server could have a worker pool
```
MatthiasKillat, line 96: How do we communicate a rejected request? What are the reasons for rejecting requests? E.g.

1.) busy - some queue full or otherwise resource starved
2.) access rights?
3.) wrong arguments?
Especially in a resource starved case, we may not be able to communicate failure to serve the request though (should be uncommon, however).

MatthiasKillat (approved):
Overall concept is sound and implementable in my opinion. It is quite similar to publish-subscribe communication already used, which should make it easier to implement. Details will surely change or be added during implementation.

One can think about some additional future based concept later, once we have evaluated the basic version in some real use-cases. This would be just for extended usability and maybe some performance optimization (less SHM allocations, improved bookkeeping of request-response pairs).

doc/design/diagrams/request_response/client_port.puml
```diff
@@ -0,0 +1,36 @@
...
+  connectionState: atomic<ConnectionState>
+}
+
+class ClientPortUser {
+  connect(): void
+  disconnect(): void
+  getConnectionState(): ConnectionState
+  allocateRequest(userPayloadSize: uint32_t, userPayloadAlignment: uint32_t): expected<RequestHeader*, AllocationError>
+  sendRequest(requestHeader: RequestHeader*): void
+  freeRequest(requestHeader: RequestHeader*): void
```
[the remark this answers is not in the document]
elBoberido: no, this is the counterpart to `allocateRequest` it's the same like `freeChunk` with pub-sub

doc/design/diagrams/request_response/client_port.puml
```diff
@@ -0,0 +1,36 @@
...
+  getResponse(): expected<const ResponseHeader*, ChunkReceiveResult>
+  releaseResponse(responseHeader: const ResponseHeader*): void
+  hasNewResponses(): void
+  hasLostResponsesSinceLastCall(): bool
```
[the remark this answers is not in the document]
elBoberido: It depends on how the client and server are configured. There are options to block everything if the queues are full. We have to think about what to use as default. I think blocking a client is a sane default, but not for the server.

doc/design/diagrams/service-discovery/overview-alternative-b.svg
```diff
@@ -1,17 +1,17 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" contentScriptType="application/ecmascript" contentStyleType="text/css" height="977px" preserveAspectRatio="none" style="width:957px;height:977px;" version="1.1" viewBox="0 0 957 977" width="957px" zoomAndPan="magnify"><defs><filter height="300%" id="fpizquwr6qwv8" width="300%" x="-1" y="-1"><feGaussianBlur result="blurOut" stdDeviation="2.0"/><feColorMatrix in="blurOut" result="blurOut2" type="matrix" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 .4 0"/><feOffset dx="4.0" dy="4.0" in="blurOut2" result="blurOut3"/><feBlend in="SourceGraphic" in2="blurOut3" mode="normal"/></filter></defs><g><text fill="#000000" font-family="sans-serif" font-size="18" lengthAdjust="spacing" textLength="782" x="81" y="16.708">Alternative B: iceoryx-based ECU forwarding CaPro messages to Gateway and User App</text><!--MD5=[45a30aaa2590b6845ebacc5fda1fa0db]
```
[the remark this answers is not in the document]
elBoberido: This was created with a different version of plantuml. Now there is a script which ensures that the same plantuml is used and therefore this won't happen in the future

doc/design/request_response_communication.md
elBoberido: Basically it is just a full queue with a `DISCARD_OLDEST_DATA` policy. Everything else should be handled by the server which means it could also response with a `REJECTED` message.

PR #834, opened 2021-06-04
Iox #33 implement ipc mutex semaphore

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes **TBD**

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
@@ -36,12 +36,15 @@ enum class AccessMode : uint64_t
 };
 static constexpr const char* ACCESS_MODE_STRING[] = {"AccessMode::READ_ONLY", "AccessMode::READ_WRITE"};
 
-enum class OwnerShip : uint64_t
+enum class Policy : uint64_t
 {
-    MINE = 0U,
-    OPEN_EXISTING_SHM = 1U
+    EXCLUSIVE_CREATE = 0U,
+    PURGE_AND_CREATE = 1U,
+    CREATE_OR_OPEN = 2U,
+    OPEN = 3U
 };
-static constexpr const char* OWNERSHIP_STRING[] = {"OwnerShip::MINE", "OwnerShip::OPEN_EXISTING_SHM"};
+static constexpr const char* POLICY_STRING[] = {
+    "Policy::EXCLUSIVE_CREATE", "Policy::PURGE_AND_CREATE", "Policy::CREATE_OR_OPEN", "Policy::OPEN"};
 
 enum class SharedMemoryError
 {
@@ -76,28 +79,31 @@ class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryEr
     ~SharedMemory() noexcept;
 
     int32_t getHandle() const noexcept;
+    bool hasOwnership() const noexcept;
+
+    static bool unlinkIfExist(const Name_t& name) noexcept;
 
     friend class DesignPattern::Creation<SharedMemory, SharedMemoryError>;
 
   private:
     SharedMemory(const Name_t& name,
                  const AccessMode accessMode,
-                 const OwnerShip ownerShip,
+                 const Policy policy,
                  const mode_t permissions,
                  const uint64_t size) noexcept;
 
-    bool open(const int oflags, const mode_t permissions, const uint64_t size) noexcept;
+    bool open(const AccessMode accessMode, const Policy policy, const mode_t permissions, const uint64_t size) noexcept;
     bool unlink() noexcept;
     bool close() noexcept;
     void destroy() noexcept;
     void reset() noexcept;
+    static int getOflagsFor(const AccessMode accessMode, const Policy policy) noexcept;
 
     SharedMemoryError errnoToEnum(const int32_t errnum) const noexcept;
 
-
     Name_t m_name;
-    OwnerShip m_ownerShip;
     int m_handle{-1};
+    bool m_hasOwnership{false};
 };
 } // namespace posix
 } // namespace iox
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
@@ -70,6 +70,7 @@ class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryEr
 {
   public:
     static constexpr uint64_t NAME_SIZE = 128U;
+    static constexpr int INVALID_HANDLE = -1;
     using Name_t = cxx::string<NAME_SIZE>;
 
     SharedMemory(const SharedMemory&) = delete;
@@ -102,7 +103,7 @@ class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryEr
     SharedMemoryError errnoToEnum(const int32_t errnum) const noexcept;
 
     Name_t m_name;
-    int m_handle{-1};
+    int m_handle{INVALID_HANDLE};
     bool m_hasOwnership{false};
 };
 } // namespace posix
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
@@ -37,7 +37,7 @@ enum class AccessMode : uint64_t
 static constexpr const char* ACCESS_MODE_STRING[] = {"AccessMode::READ_ONLY", "AccessMode::READ_WRITE"};
 
 /// @brief describes how the shared memory is opened or created
-enum class Policy : uint64_t
+enum class OpenMode : uint64_t
 {
     /// @brief creates the shared memory, if it exists already the construction will fail
     EXCLUSIVE_CREATE = 0U,
@@ -46,10 +46,10 @@ enum class Policy : uint64_t
     /// @brief creates the shared memory, if it does not exist otherwise it opens it
     OPEN_OR_CREATE = 2U,
     /// @brief opens the shared memory, if it does not exist it will fail
-    OPEN = 3U
+    OPEN_EXISTING = 3U
 };
-static constexpr const char* POLICY_STRING[] = {
-    "Policy::EXCLUSIVE_CREATE", "Policy::PURGE_AND_CREATE", "Policy::CREATE_OR_OPEN", "Policy::OPEN"};
+static constexpr const char* OPEN_MODE_STRING[] = {
+    "OpenMode::EXCLUSIVE_CREATE", "OpenMode::PURGE_AND_CREATE", "OpenMode::CREATE_OR_OPEN", "OpenMode::OPEN_EXISTING"};
 
 enum class SharedMemoryError
 {
@@ -87,25 +87,26 @@ class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryEr
     int32_t getHandle() const noexcept;
     bool hasOwnership() const noexcept;
 
-    static bool unlinkIfExist(const Name_t& name) noexcept;
+    static cxx::expected<bool, SharedMemoryError> unlinkIfExist(const Name_t& name) noexcept;
 
     friend class DesignPattern::Creation<SharedMemory, SharedMemoryError>;
 
   private:
     SharedMemory(const Name_t& name,
                  const AccessMode accessMode,
-                 const Policy policy,
+                 const OpenMode openMode,
                  const mode_t permissions,
                  const uint64_t size) noexcept;
 
-    bool open(const AccessMode accessMode, const Policy policy, const mode_t permissions, const uint64_t size) noexcept;
+    bool
+    open(const AccessMode accessMode, const OpenMode openMode, const mode_t permissions, const uint64_t size) noexcept;
     bool unlink() noexcept;
     bool close() noexcept;
     void destroy() noexcept;
     void reset() noexcept;
-    static int getOflagsFor(const AccessMode accessMode, const Policy policy) noexcept;
+    static int getOflagsFor(const AccessMode accessMode, const OpenMode openMode) noexcept;
 
-    SharedMemoryError errnoToEnum(const int32_t errnum) const noexcept;
+    static SharedMemoryError errnoToEnum(const int32_t errnum) noexcept;
 
     Name_t m_name;
     int m_handle{INVALID_HANDLE};
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
@@ -74,7 +74,7 @@ enum class SharedMemoryError
 class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryError>
 {
   public:
-    static constexpr uint64_t NAME_SIZE = 128U;
+    static constexpr uint64_t NAME_SIZE = platform::IOX_MAX_SHM_NAME_LENGTH;
     static constexpr int INVALID_HANDLE = -1;
     using Name_t = cxx::string<NAME_SIZE>;
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
@@ -49,7 +49,7 @@ enum class OpenMode : uint64_t
     OPEN_EXISTING = 3U
 };
 static constexpr const char* OPEN_MODE_STRING[] = {
-    "OpenMode::EXCLUSIVE_CREATE", "OpenMode::PURGE_AND_CREATE", "OpenMode::CREATE_OR_OPEN", "OpenMode::OPEN_EXISTING"};
+    "OpenMode::EXCLUSIVE_CREATE", "OpenMode::PURGE_AND_CREATE", "OpenMode::OPEN_OR_CREATE", "OpenMode::OPEN_EXISTING"};
 
 enum class SharedMemoryError
 {
@@ -71,6 +71,10 @@ enum class SharedMemoryError
     UNKNOWN_ERROR
 };
 
+/// @brief Creates a bare metal shared memory object with the posix functions
+///        shm_open, shm_unlink etc.
+///        It must be used in combination with MemoryMap (or manual mmap calls)
+//         to gain access to the created/opened shared memory
 class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryError>
 {
   public:
@@ -84,14 +88,31 @@ class SharedMemory : public DesignPattern::Creation<SharedMemory, SharedMemoryEr
     SharedMemory& operator=(SharedMemory&&) noexcept;
     ~SharedMemory() noexcept;
 
+    /// @brief returns the file handle of the shared memory
     int32_t getHandle() const noexcept;
+
+    /// @brief this class has the ownership of the shared memory when the shared
+    ///        memory was created by this class. This is the case when this class
+    ///        was successful created with EXCLUSIVE_CREATE, PURGE_AND_CREATE or OPEN_OR_CREATE
+    ///        and the shared memory was created. If an already available shared memory
+    ///        is opened then this class does not have the ownership.
     bool hasOwnership() const noexcept;
 
+    /// @brief removes shared memory with a given name from the system
+    /// @param[in] name name of the shared memory
+    /// @return true if the shared memory was removed, false if the shared memory did not exist and
+    ///         SharedMemoryError when the underlying shm_unlink call failed.
     static cxx::expected<bool, SharedMemoryError> unlinkIfExist(const Name_t& name) noexcept;
 
     friend class DesignPattern::Creation<SharedMemory, SharedMemoryError>;
 
   private:
+    /// @brief constructs or opens existing shared memory
+    /// @param[in] name the name of the shared memory, must start with a leading /
+    /// @param[in] accessMode defines if the shared memory is mapped read only or with read write rights
+    /// @param[in] openMode states how the shared memory is created/opened
+    /// @param[in] permissions the permissions the shared memory should have
+    /// @param[in] size the size in bytes of the shared memory
     SharedMemory(const Name_t& name,
                  const AccessMode accessMode,
                  const OpenMode openMode,
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
@@ -36,11 +36,16 @@ enum class AccessMode : uint64_t
 };
 static constexpr const char* ACCESS_MODE_STRING[] = {"AccessMode::READ_ONLY", "AccessMode::READ_WRITE"};
 
+/// @brief describes how the shared memory is opened or created
 enum class Policy : uint64_t
 {
+    /// @brief creates the shared memory, if it exists already the construction will fail
     EXCLUSIVE_CREATE = 0U,
+    /// @brief creates the shared memory, if it exists it will be deleted and recreated
     PURGE_AND_CREATE = 1U,
-    CREATE_OR_OPEN = 2U,
+    /// @brief creates the shared memory, if it does not exist otherwise it opens it
+    OPEN_OR_CREATE = 2U,
+    /// @brief opens the shared memory, if it does not exist it will fail
     OPEN = 3U
 };
 static constexpr const char* POLICY_STRING[] = {
--- a/iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/handle_translator.hpp
+++ b/iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/handle_translator.hpp
@@ -19,6 +19,12 @@
 #include "iceoryx_hoofs/platform/windows.hpp"
 #include <vector>
 
+/// @brief In windows file handles have the type HANDLE (void*) in linux it is
+///        usually an int. To establish the portability we keep track of the
+///        windows handles and assign them a unique int so that they can be used
+///        in a platform independent manner.
+///        This class translates a windows handle of type HANDLE to its linux
+///        file handle int pendant.
 class HandleTranslator
 {
   public:
--- a/iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/handle_translator.hpp
+++ b/iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/handle_translator.hpp
@@ -1,4 +1,3 @@
-// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,21 +13,27 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-#ifndef IOX_HOOFS_WIN_PLATFORM_UNISTD_HPP
-#define IOX_HOOFS_WIN_PLATFORM_UNISTD_HPP
+#ifndef IOX_HOOFS_WIN_PLATFORM_HANDLE_TRANSLATOR_HPP
+#define IOX_HOOFS_WIN_PLATFORM_HANDLE_TRANSLATOR_HPP
 
-#include "iceoryx_hoofs/platform/types.hpp"
 #include "iceoryx_hoofs/platform/windows.hpp"
-
-#include <io.h>
 #include <vector>
 
-#define _SC_PAGESIZE 1
-#define STDERR_FILENO 2
+class HandleTranslator
+{
+  public:
+    static HandleTranslator& getInstance() noexcept;
+    HANDLE get(const int handle) const noexcept;
+    int add(HANDLE handle) noexcept;
+    void remove(int handle) noexcept;
 
+  private:
+    HandleTranslator() noexcept = default;
+    struct handle_t
+    {
+        HANDLE windowsHandle;
+    };
+    std::vector<handle_t> m_handleList;
+};
 
-int ftruncate(int fildes, off_t length);
-long sysconf(int name);
-int iox_close(int fd);
-
-#endif // IOX_HOOFS_WIN_PLATFORM_UNISTD_HPP
+#endif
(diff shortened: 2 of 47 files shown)


iceoryx_hoofs/test/moduletests/test_unix_domain_sockets.cpp
```diff
@@ -193,99 +196,159 @@ TEST_F(UnixDomainSocket_test, SendOnServerLeadsToError)
...
+         "he knows the answer to the ultimate question of life, the universe, and everything - just look deep into his "
+         "eyes"},
+        [&](auto& msg) { return client.timedSend(msg, 1_ms); },
+        [&]() { return server.receive(); });
+}
+
+TEST_F(UnixDomainSocket_test, SuccessfulCommunicationOfMultipleMessagesWithTimedSendAndTimedReceive)
+{
+    successfulSendAndReceive(
+        {"hypnotoad was part of the german pop band Modern Talking and produced songs like",
```
mossmaurice, line 334: :man_dancing:

iceoryx_hoofs/source/posix_wrapper/shared_memory_object.cpp
```diff
@@ -79,15 +79,15 @@ SharedMemoryObject::SharedMemoryObject(const SharedMemory::Name_t& name,
         std::cerr << "Unable to create a shared memory object with the following properties [ name = " << name
                   << ", sizeInBytes = " << memorySizeInBytes
                   << ", access mode = " << ACCESS_MODE_STRING[static_cast<uint64_t>(accessMode)]
-                  << ", ownership = " << OWNERSHIP_STRING[static_cast<uint64_t>(ownerShip)]
+                  << ", ownership = " << POLICY_STRING[static_cast<uint64_t>(policy)]
```
FerdinandSpitzschnueffler, line 82: Should `ownership` be replaced by `policy`? It doesn't seem to be the most suitable word for the possible values of `policy`...

FerdinandSpitzschnueffler: Same for line 105

iceoryx_hoofs/test/moduletests/test_posix_file_lock.cpp
```diff
@@ -63,15 +63,13 @@ TEST_F(FileLock_test, EmptyNameLeadsToError)
...
 }
 
 TEST_F(FileLock_test, MaxStringWorks)
 {
-    const FileLock::FileName_t maxString{
-        "OeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPalo"
-        "emaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlancaOeLaPaloemaBlanc"
-        "aOeLaPaloemaBlancaOeLaPaloemaB"};
+    const FileLock::FileName_t maxString(iox::cxx::TruncateToCapacity,
+                                         std::string(FileLock::FileName_t().capacity(), 'x'));
```
elBoberido, line 72: oh nö, över the mountains I fly is no more?

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -81,6 +76,14 @@ SharedMemory::~SharedMemory() noexcept
     destroy();
 }
 
+int SharedMemory::getOflagsFor(const AccessMode accessMode, const Policy policy) noexcept
+{
+    int oflags = 0;
+    oflags |= (accessMode == AccessMode::READ_ONLY) ? O_RDONLY : O_RDWR;
+    oflags |= (policy != Policy::OPEN) ? O_CREAT | O_EXCL : 0;
```
elBoberido, line 83: Is this correct? Shouldn't `O_EXCL` only be set if the policy is `EXCLUSIVE_CREATE`?

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -125,46 +128,64 @@ int32_t SharedMemory::getHandle() const noexcept
...
 
-        if (posixCall(iox_shm_open)(m_name.c_str(), oflags, permissions)
-                .failureReturnValue(-1)
-                .evaluate()
-                .and_then([this](auto& r) { m_handle = r.value; })
-                .or_else([this](auto& r) { m_errorValue = this->errnoToEnum(r.errnum); })
-                .has_error())
+        auto result = posixCall(iox_shm_open)(m_name.c_str(), getOflagsFor(accessMode, policy), permissions)
+                          .failureReturnValue(INVALID_HANDLE)
+                          .suppressErrorMessagesForErrnos((policy == Policy::CREATE_OR_OPEN) ? EEXIST : 0)
```
elBoberido, line 160: This shouldn't be necessary if `getOflagsFor` sets `O_EXCL` correctly

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -125,46 +128,64 @@ int32_t SharedMemory::getHandle() const noexcept
...
+                          .evaluate();
+        if (result.has_error())
         {
+            // if it was not possible to create the shm exclusively someone else has the
+            // ownership and we just try to open it
+            if (policy == Policy::CREATE_OR_OPEN && result.get_error().errnum == EEXIST)
```
elBoberido, line 166: Same here, if `getOflagsFor` sets `O_EXCL` correctly, this should not be necessary

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -176,19 +197,22 @@ bool SharedMemory::open(const int oflags, const mode_t permissions, const uint64
     return true;
 }
 
+bool SharedMemory::unlinkIfExist(const Name_t& name) noexcept
+{
+    return !posixCall(shm_unlink)(name.c_str())
+                .failureReturnValue(INVALID_HANDLE)
+                .suppressErrorMessagesForErrnos(ENOENT)
```
elBoberido, line 204: Shouldn't this be `.ignoreErrnos(ENOENT)`? After all the method has an `IfExists` suffix

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -19,34 +19,53 @@
...
+constexpr const char SharedMemory_Test::SUT_SHM_NAME[];
+
 TEST_F(SharedMemory_Test, CTorWithValidArguments)
 {
     auto sut = iox::posix::SharedMemory::create("/ignatz",
                                                 iox::posix::AccessMode::READ_WRITE,
-                                                iox::posix::OwnerShip::MINE,
+                                                iox::posix::Policy::PURGE_AND_CREATE,
                                                 S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH,
                                                 128);
```
elBoberido, line 70: Since you created `createSharedMemory` wouldn't it make sense to use it here and in the following tests?

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -100,11 +120,116 @@ TEST_F(SharedMemory_Test, MoveCTorWithValidValues)
...
+    EXPECT_TRUE(iox::posix::SharedMemory::unlinkIfExist(SUT_SHM_NAME));
+}
+
+TEST_F(SharedMemory_Test, OpenWorksWhenShmExist)
+{
+    ASSERT_FALSE(
+        iox::posix::posixCall(iox_shm_open)(SUT_SHM_NAME, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP)
+            .failureReturnValue(SharedMemory::INVALID_HANDLE)
+            .evaluate()
+            .has_error());
```
elBoberido, line 218: This is used quite often and could be moved to a `shmOpen` method or similar

elBoberido (commented):
platform files not yet reviewed ... something for tomorrow

iceoryx_hoofs/platform/win/source/handle_translator.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+HANDLE HandleTranslator::get(const int handle) const noexcept
+{
+    return m_handleList[static_cast<size_t>(handle)];
+}
+
+int HandleTranslator::add(HANDLE handle) noexcept
+{
+    for (int64_t limit = m_handleList.size(), k = 0; k < limit; ++k)
```
elBoberido, line 32: It's a pity we don't get the index in a range based for loop. But why this unusual for-loop parameters and not just `int64_t i = 0; i < m_handleList.size(); ++i`

iceoryx_hoofs/platform/win/source/semaphore.cpp
```diff
@@ -16,18 +16,53 @@
...
+    {
+        return newHandle;
+    }
+
+    newHandle =
+        Win32Call(OpenSemaphoreA, SEMAPHORE_ALL_ACCESS, false, generateSemaphoreName(sem->uniqueId).c_str()).value;
+    if (newHandle == nullptr)
+    {
+        fprintf(stderr,
+                "interprocess semaphore %s is corrupted - segmentation fault immanent\n",
```
elBoberido, line 48: typo? immanent -> imminent?

iceoryx_hoofs/platform/win/source/pthread.cpp
```diff
@@ -86,72 +89,98 @@ int pthread_mutex_destroy(pthread_mutex_t* mutex)
...
 
-        if (mutex->handle == NULL)
-        {
-            return EINVAL;
-        }
+    newHandle = Win32Call(OpenMutexA, MUTEX_ALL_ACCESS, false, generateMutexName(mutex->uniqueId).c_str()).value;
+    if (newHandle == nullptr)
+    {
+        fprintf(stderr,
+                "interprocess mutex %s is corrupted - segmentation fault immanent\n",
```
elBoberido, line 119: Same here, typo?

iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/handle_translator.hpp
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+#ifndef IOX_HOOFS_WIN_PLATFORM_HANDLE_TRANSLATOR_HPP
+#define IOX_HOOFS_WIN_PLATFORM_HANDLE_TRANSLATOR_HPP
+
+#include "iceoryx_hoofs/platform/windows.hpp"
+#include <vector>
+
+class HandleTranslator
```
dkroenke, line 22: Some explanation of the purpose of this class would be good.
Same for the classes below.

iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/platform_settings.hpp
```diff
@@ -34,7 +34,10 @@ constexpr const char IOX_PATH_SEPARATORS[] = "\\/";
 // so that the code with stub implementation at least compiles
 constexpr uint64_t IOX_UDS_SOCKET_MAX_MESSAGE_SIZE = 1024U;
 constexpr char IOX_UDS_SOCKET_PATH_PREFIX[] = "";
+constexpr const char IOX_LOCK_FILE_PATH_PREFIX[] = "C:\\Windows\\Temp\\";
```
dkroenke, line 37: Normally `C:\Windows\temp` is used to store temporary data from the operating system or system processes. I think that this is not our usecase. Application-related data is normally stored under `C:\Users\user\AppData\Local\Temp`.
Now the question is how to retrieve this path correctly. One possible way is to look at the ``%TEMP%` environment variable or maybe use [GetTempPath2A](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2a).

It's not needed to fix in this PR because temp under windows will work but it is at least an issue worth.
- [ ] fix path for temp files under windows or create an issue to follow up

dkroenke (changes requested):
Please add some explanation to the newly added classes

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -100,11 +120,116 @@ TEST_F(SharedMemory_Test, MoveCTorWithValidValues)
...
                                                 S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH,
                                                 128);
-    EXPECT_THAT(sut->getHandle(), Ne(-1));
+    ASSERT_FALSE(sut.has_error());
+    EXPECT_THAT(sut->getHandle(), Ne(SharedMemory::INVALID_HANDLE));
+}
+
+TEST_F(SharedMemory_Test, UnlinkNonExistingShmFails)
+{
+    EXPECT_FALSE(iox::posix::SharedMemory::unlinkIfExist("/look_over_there_a_dead_pidgin"));
```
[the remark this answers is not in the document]
elBoberido: To be even more precise "look, there's a dead seagull flying"

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
```diff
@@ -36,12 +36,15 @@ enum class AccessMode : uint64_t
 };
 static constexpr const char* ACCESS_MODE_STRING[] = {"AccessMode::READ_ONLY", "AccessMode::READ_WRITE"};
 
-enum class OwnerShip : uint64_t
+enum class Policy : uint64_t
```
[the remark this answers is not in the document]
elfenpiff: I consulted @ithier and he suggested `Policy`. @elBoberido would you accept this? `OpenOption::OPEN` sounds just weird and `Mode` is already taken.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
```diff
@@ -36,12 +36,15 @@ enum class AccessMode : uint64_t
...
 {
-    MINE = 0U,
-    OPEN_EXISTING_SHM = 1U
+    EXCLUSIVE_CREATE = 0U,
+    PURGE_AND_CREATE = 1U,
+    CREATE_OR_OPEN = 2U,
+    OPEN = 3U
```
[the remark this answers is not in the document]
elfenpiff: The policy `PURGE_AND_CREATE` is used by roudi for all the shared memory partitions. `OPEN` is used by every application when mapping shared memory. `CREATE_OR_OPEN` is used by the named pipes and I added `EXCLUSIVE_CREATE` for completeness and it is actually no real overhead.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -125,46 +128,64 @@ int32_t SharedMemory::getHandle() const noexcept
...
-    cxx::Expects(static_cast<int64_t>(size) <= std::numeric_limits<int64_t>::max());
+    return m_hasOwnership;
+}
 
+bool SharedMemory::open(const AccessMode accessMode,
+                        const Policy policy,
+                        const mode_t permissions,
+                        const uint64_t size) noexcept
+{
+    cxx::Expects(static_cast<int64_t>(size) <= std::numeric_limits<int64_t>::max());
```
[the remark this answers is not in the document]
elfenpiff: Actually it was already there but I will adjust it.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
elfenpiff: Actually no.
1. `PURGE_AND_CREATE` deletes the shared memory beforehand with `shm_unlink` and therefore we can assume that it does not exist. If its still existing either `shm_unlink` failed for unknown reasons or someone else swooped in and created a shared memory partition with that name before we could.
2. `OPEN_OR_CREATE` here we take ownership and my algorithm was like: create exclusively (and we have the ownership and have to clean up on destruction) or create fails since it already exists and we open it and do not have the ownership.

For `OPEN_OR_CREATE` I could also reverse the approach and open first and when it fails create the shared memory but the code looked a bit more cumbersome and therefore I decided for this approach.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
elfenpiff: No I want it to fail for `CREATE_OR_OPEN` when it exists. I need to know if it existed and if the shm was just opened. If we created it we have the ownership and are responsible to clean it up in the dtor but not when we just opened the shared memory then its someone else responsibility.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
elfenpiff: No, this is required because of the ownership. `shm_open` does not tell me if the shm was opened because it was still there or because it was created.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
elfenpiff: No the `IpcChannel` contract is to return true if the channel was removed otherwise false.

iceoryx_hoofs/platform/win/source/handle_translator.cpp
elfenpiff: It is my habit and this is useful when working with a `std::vector` where it is possible that with every iteration the size is determined again. But I admit this is not required in the `cxx::vector` case since we store the size there.

iceoryx_hoofs/platform/win/include/iceoryx_hoofs/platform/platform_settings.hpp
elfenpiff: Added as task to issue #33

iceoryx_hoofs/platform/win/source/handle_translator.cpp
elBoberido: It's the same with `std::vector` and other container, at least in release builds.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -125,46 +128,65 @@ int32_t SharedMemory::getHandle() const noexcept
...
+                          .evaluate();
+        if (result.has_error())
         {
+            // if it was not possible to create the shm exclusively someone else has the
+            // ownership and we just try to open it
+            if (policy == Policy::OPEN_OR_CREATE && result.get_error().errnum == EEXIST)
+            {
+                result = posixCall(iox_shm_open)(m_name.c_str(), getOflagsFor(accessMode, Policy::OPEN), permissions)
+                             .failureReturnValue(INVALID_HANDLE)
+                             .evaluate();
```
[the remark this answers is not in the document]
elfenpiff: I wouldn't say that this is an error since the whole object is then in an invalid state. But you are right we should be extra cautious and I change this.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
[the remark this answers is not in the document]
elBoberido: Well, I'm just pointing out you are inventing new terminology. The STL has [openmode](https://en.cppreference.com/w/cpp/io/basic_filebuf/open), Boost has [open_mode](https://www.boost.org/doc/libs/1_76_0/boost/log/utility/open_mode.hpp) for logging, Qt has [OpenMode](https://doc.qt.io/qt-5/qiodevice.html#OpenModeFlag-enum), Rust has [OpenOptions](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html) and Python has just [mode](https://docs.python.org/3/library/functions.html#open). They mostly also incorporate `AccessMode` into that parameter.

By just having the signature `bool open(const AccessMode accessMode, const Policy policy, const mode_t permissions, const uint64_t size) noexcept;` I would have no clue what `Policy policy` means, contrary to the other parameter. Although, it wouldn't be much different with `Ownership`.

But this is not only about opening files, but specifically about shared memory which has certain implications on ownership for cleanup. Let's have a look at boost [shared_memory_object](https://www.boost.org/doc/libs/1_76_0/doc/html/boost/interprocess/shared_memory_object.html) and Qt's [QSharedMemory](https://doc.qt.io/qt-5/qsharedmemory.html). Boost uses different ctors and Qt uses different methods. I kind of like the approach of Qt with `attach` and `create` which could also be used to indicate ownership, although I think they are using some memory to store meta information like this. For us, `attach` and `create` would clearly show the ownership situation. With `create` maybe having an `OnExisting` enum with `PURGE` and `ERROR`. The current `OPEN_OR_CREATE` would done by `create` followed by `attach`, maybe in an `attach_or_create` convenience method. Not sure how usable such an API would be and definitely nothing for this PR.

To make a long story short, please consider to not introduce new terminology ... oh and it could be `OpenOption::OPEN_ONLY` or `OpenMode::OPEN_ONLY` like boost does with `open_mode` ;)

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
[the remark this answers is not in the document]
elBoberido: yes, but it's not documented what the tags mean ;)

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
elBoberido: That's true for `unlink` but this is called `unlikIfExists` which, for me, implies a different semantic like "I only want to handle a failure if it is still there after this call but don't care if it wasn't there in the first place". Maybe instead of a binary return value, an expected would be necessary?

Oh, and please always add doxygen documentation for new code. You don't need to add it to all the methods which are already there, since this is out of scope for the PR but the new code should have a documentation.

elBoberido: Okay, now I understand. This class wasn't used for an `IpcChannel` implementation and therefore the method is new. The `IpcChannel` implementations have a `cxx::expected<bool, Error> unlinkIfExists(const Name_t& name)` signature though.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/shared_memory.hpp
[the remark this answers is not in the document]
elfenpiff: After the first paragraph you already convinced me. So you wrote the next three paragraphs for nothing muahahaha ;)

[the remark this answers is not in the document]
elBoberido: Doh :smile:

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
elfenpiff: returning an expected now

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -19,77 +19,81 @@
...
                                                 iox::posix::AccessMode::READ_WRITE,
-                                                iox::posix::OwnerShip::MINE,
+                                                openMode,
                                                 S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH,
                                                 128);
+    }
+
+    bool createRawSharedMemory(const iox::posix::SharedMemory::Name_t& name)
+    {
+        return !iox::posix::posixCall(iox_shm_open)(
```
elBoberido, line 63: Let's assume I am a chicken, should we call `unlinkIfExists` first to ensure this call will always succeed?

I also got confused by the names of the two `create...` functions since they are quite similar. Should `createSharedMemory` maybe be renamed to `createSut` or similar?

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -100,11 +104,105 @@ TEST_F(SharedMemory_Test, MoveCTorWithValidValues)
...
+    ASSERT_TRUE(createRawSharedMemory(SUT_SHM_NAME));
+    auto sut = createSharedMemory(SUT_SHM_NAME, OpenMode::PURGE_AND_CREATE);
+    ASSERT_FALSE(sut.has_error());
+    EXPECT_TRUE(sut->hasOwnership());
+    EXPECT_THAT(sut->getHandle(), Ne(SharedMemory::INVALID_HANDLE));
+}
+
+TEST_F(SharedMemory_Test, CreateOrOpenCreatesShmWhenShmDoesNotExist)
+{
+    auto sut = createSharedMemory(SUT_SHM_NAME, OpenMode::OPEN_OR_CREATE);
```
elBoberido, line 167: I guess here should also `unlinkIfExists` be called beforehand ... or maybe in `Setup` for `SUT_SHM_NAME`

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -100,11 +104,105 @@ TEST_F(SharedMemory_Test, MoveCTorWithValidValues)
...
+    ASSERT_TRUE(createRawSharedMemory(SUT_SHM_NAME));
+    {
+        auto sut = createSharedMemory(SUT_SHM_NAME, OpenMode::OPEN_OR_CREATE);
+        ASSERT_FALSE(sut.has_error());
+        EXPECT_FALSE(sut->hasOwnership());
+        EXPECT_THAT(sut->getHandle(), Ne(SharedMemory::INVALID_HANDLE));
+    }
+    auto result = iox::posix::SharedMemory::unlinkIfExist(SUT_SHM_NAME);
+    ASSERT_FALSE(result.has_error());
+    EXPECT_TRUE(*result);
```
elBoberido, line 184: Are this checks necessary here? Shouldn't this be an own test case if not already tested?

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
```diff
@@ -100,11 +104,105 @@ TEST_F(SharedMemory_Test, MoveCTorWithValidValues)
...
+    ASSERT_TRUE(createRawSharedMemory(SUT_SHM_NAME));
+    {
+        auto sut = createSharedMemory(SUT_SHM_NAME, OpenMode::OPEN_EXISTING);
+        ASSERT_FALSE(sut.has_error());
+        EXPECT_FALSE(sut->hasOwnership());
+        EXPECT_THAT(sut->getHandle(), Ne(SharedMemory::INVALID_HANDLE));
+    }
+    auto result = iox::posix::SharedMemory::unlinkIfExist(SUT_SHM_NAME);
+    ASSERT_FALSE(result.has_error());
+    EXPECT_TRUE(*result);
```
elBoberido, line 198: same here

iceoryx_hoofs/source/posix_wrapper/named_pipe.cpp
```diff
@@ -89,7 +89,7 @@ NamedPipe::NamedPipe(const IpcChannelName_t& name,
         // m_messages. when we add the alignment it is guaranteed that enough memory should be available.
         sizeof(MessageQueue_t) + alignof(MessageQueue_t),
         AccessMode::READ_WRITE,
-        (channelSide == IpcChannelSide::SERVER) ? OwnerShip::MINE : OwnerShip::OPEN_EXISTING_SHM,
+        (channelSide == IpcChannelSide::SERVER) ? OpenMode::OPEN_OR_CREATE : OpenMode::OPEN_EXISTING,
```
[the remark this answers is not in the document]
elfenpiff: This is correct but it is absolutely no problem to support multiple readers and writes for named pipes and therefore I wanted to use this advantage right from the beginning. I felt that restricting this to single server multiple clients would gain nothing.

The restriction would require more code (for the error handling when a named pipe already exists and you would like to create a server) and would have less flexibility and this seemed just weird.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -81,6 +76,14 @@ SharedMemory::~SharedMemory() noexcept
     destroy();
 }
 
+int SharedMemory::getOflagsFor(const AccessMode accessMode, const OpenMode openMode) noexcept
+{
+    int oflags = 0;
+    oflags |= (accessMode == AccessMode::READ_ONLY) ? O_RDONLY : O_RDWR;
+    oflags |= (openMode != OpenMode::OPEN_EXISTING) ? O_CREAT | O_EXCL : 0;
```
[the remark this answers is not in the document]
elfenpiff: Deal! :D

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
elfenpiff: Explicitly no since some tests require that the shared memory already exist and this will fail like `ExclusiveCreateFailsWhenShmExists`

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
elfenpiff: I wanted to test all side effects and make sure that `OPEN_OR_CREATE` does not delete the shared memory if it does not own it.

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
elfenpiff: Here I also wanted to make sure that the shm is not removed when it is not owned.

iceoryx_hoofs/test/moduletests/test_shared_memory.cpp
elBoberido: Okay, that's a good point. Could you move this into a function and give it a proper name so that ignorant devs like me don't refactor this out at some point.

iceoryx_hoofs/source/posix_wrapper/named_pipe.cpp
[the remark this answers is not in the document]
elBoberido: How does the shm behave i unlink is called even if there are other applications which have it still mapped into address space? Can they sill use the shm or do they crash?

Oh, and with this approach it is also possible to "steal" messages from the server without him even noticing it. The message queue has the same problem though, not sure about the unix domain socket.

elBoberido (commented):
almost there

iceoryx_hoofs/source/posix_wrapper/named_pipe.cpp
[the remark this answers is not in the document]
elfenpiff: >  Can they sill use the shm or do they crash?

They will crash and burn. Yeah this is maybe dangerous to use but when we restrict this I would restrict this on a higher level - for instance in a real IpcChannel abstraction. The posix wrappers do not restrict functionality they provide the developer with what is possible (with a safe API) so that they can decide howto handle it with logic or restrictions.

> "steal" messages?

Actually we can solve this problem easily with ACLs. Then only the trustworthy applications have read/write access. Also every client can steal messages from a named pipe since the server and the client requires write access.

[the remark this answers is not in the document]
elBoberido: That's a good point. It might be quite difficult to prevent stealing with the NamedPipe. I guess it should not be used in environments where this is a problem.

elBoberido (dismissed):

FerdinandSpitzschnueffler (approved):

elBoberido (approved):

PR #852, opened 2021-06-16
Iox #415 Refactor Service Description Part1: Remove integers values

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
This is the first part of a series of pull requests regarding the refactoring of `ServiceDescription`, which paves the way for a new `ServiceRegistry`.

**Done:**
* Remove the legacy integers inside `ServiceDescription` so the user always deals only with the three strings

**Future topics (not part of #415):**
* Map strings to integers underneath if we should run into runtime problems (dictionary approach)

## References

- Closes partly #415

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -108,9 +108,11 @@ class ServiceDescription
     ~ServiceDescription() = default;
 
     /// @brief construction of the capro service description using integers to create a service service description
+    /// @todo remove
     ServiceDescription(uint16_t f_serviceID, uint16_t f_instanceID) noexcept;
 
     /// @brief construction of the capro service description using fixed strings to create a service service description
+    /// @todo remove
     ServiceDescription(const IdString_t& f_service, const IdString_t& f_instance) noexcept;
 
     /// @brief construction of the capro service description using integers to create an event service description
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2019, 2021 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -29,16 +29,13 @@ namespace iox
 {
 namespace capro
 {
-static constexpr uint16_t InvalidID = 0u;
-static const IdString_t InvalidIDString{"0"};
-static constexpr uint16_t AnyService = 0xFFFFu;
-static constexpr uint16_t AnyInstance = 0xFFFFu;
-static constexpr uint16_t AnyEvent = 0xFFFFu;
+/// @todo remove
+static const IdString_t InvalidString{"INVALID"};
 static constexpr char AnyServiceString[]{"65535"};
 static constexpr char AnyInstanceString[]{"65535"};
 static constexpr char AnyEventString[]{"65535"};
 static constexpr int32_t MAX_NUMBER_OF_CHARS = 64;
-static constexpr size_t CLASS_HASH_ELEMENT_COUNT{4u};
+static constexpr size_t CLASS_HASH_ELEMENT_COUNT{4U};
 
 /// @brief Describes from which interface the service is coming from
 enum class Interfaces : uint16_t
@@ -103,20 +100,13 @@ class ServiceDescription
 
     /// @brief default C'tor
     ServiceDescription() noexcept;
-    ServiceDescription(const ServiceDescription&) = default;
-    ServiceDescription(ServiceDescription&&) = default;
-    ~ServiceDescription() = default;
-
-    /// @brief construction of the capro service description using integers to create a service service description
-    /// @todo remove
-    ServiceDescription(uint16_t f_serviceID, uint16_t f_instanceID) noexcept;
+    ServiceDescription(const ServiceDescription&) noexcept = default;
+    ServiceDescription(ServiceDescription&&) noexcept = default;
+    ~ServiceDescription() noexcept = default;
 
     /// @brief construction of the capro service description using fixed strings to create a service service description
     /// @todo remove
     ServiceDescription(const IdString_t& f_service, const IdString_t& f_instance) noexcept;
-
-    /// @brief construction of the capro service description using integers to create an event service description
-    ServiceDescription(uint16_t f_serviceID, uint16_t f_eventID, uint16_t f_instanceID) noexcept;
 
     /// @brief construction of the capro service description using fixed strings to create an event service description
     ServiceDescription(const IdString_t& f_service,
@@ -153,6 +143,7 @@ class ServiceDescription
     /// @brief Returns the scope of a ServiceDescription
     Scope getScope() noexcept;
 
+    /// @todo remove any*string from doxygen?
     ///@brief Returns true for valid ServiceDescription
     /// false for ServiceDescription that contains either of InvalidID/InvalidIDString  AnyService/AnyServiceString,
     /// AnyInstance/AnyInstanceString, AnyEvent/AnyEventString.
@@ -160,9 +151,6 @@ class ServiceDescription
 
     ///@{
     /// Getters for the integer and string IDs
-    uint16_t getInstanceID() const noexcept;
-    uint16_t getServiceID() const noexcept;
-    uint16_t getEventID() const noexcept;
     IdString_t getServiceIDString() const noexcept;
     IdString_t getInstanceIDString() const noexcept;
     IdString_t getEventIDString() const noexcept;
@@ -177,12 +165,6 @@ class ServiceDescription
     Interfaces getSourceInterface() const noexcept;
 
   private:
-    /// @brief 16-Bit service ID
-    uint16_t m_serviceID;
-    /// @brief 16-Bit event ID
-    uint16_t m_eventID;
-    /// @brief 16-Bit instance ID
-    uint16_t m_instanceID;
     /// @brief string representation of the service
     IdString_t m_serviceString;
     /// @brief string representation of the instance
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -163,11 +163,11 @@ class ServiceDescription
 
   private:
     /// @brief string representation of the service
-    IdString_t m_serviceString;
+    IdString_t m_serviceString{InvalidString};
     /// @brief string representation of the instance
-    IdString_t m_instanceString;
+    IdString_t m_instanceString{InvalidString};
     /// @brief string representation of the event
-    IdString_t m_eventString;
+    IdString_t m_eventString{InvalidString};
 
     bool m_hasServiceOnlyDescription = false;
     /// @brief 128-Bit class hash (32-Bit * 4)
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -29,7 +29,6 @@ namespace iox
 {
 namespace capro
 {
-/// @todo remove
 static const IdString_t InvalidString{"INVALID"};
 static constexpr char AnyServiceString[]{"65535"};
 static constexpr char AnyInstanceString[]{"65535"};
@@ -76,9 +75,7 @@ constexpr char ScopeTypeString[][MAX_NUMBER_OF_CHARS] = {"WORLDWIDE", "INTERNAL"
 
 /// @brief class for the identification of a communication event including information on the service, the service
 /// instance and the event id.
-/// In order to support different communication protocols, two types of members exist: integer and string identifiers.
-/// If string IDs are used, the integers are initialized to an invalid number. A class object can be
-/// serialized/deserialized, so it is possible to send the information e.g. over a IPC channel.
+/// A class object can be serialized/deserialized, so it is possible to send the information e.g. over a IPC channel.
 class ServiceDescription
 {
   public:
@@ -115,8 +112,8 @@ class ServiceDescription
                        ClassHash m_classHash = {0u, 0u, 0u, 0u},
                        Interfaces interfaceSource = Interfaces::INTERNAL) noexcept;
 
-    /// @brief compare operator. If wildcards AnyService, AnyInstance or AnyEvent are used as integer IDs, the
-    /// corresponding member comparisons are skipped. Otherwise, both the integer and the string members are compared.
+    /// @brief compare operator. If wildcards AnyServiceString, AnyInstanceString or AnyEventString are used, the
+    /// corresponding member comparisons are skipped.
     bool operator==(const ServiceDescription& rhs) const;
 
     /// @brief negation of compare operator.
@@ -143,7 +140,6 @@ class ServiceDescription
     /// @brief Returns the scope of a ServiceDescription
     Scope getScope() noexcept;
 
-    /// @todo remove any*string from doxygen?
     ///@brief Returns true for valid ServiceDescription
     /// false for ServiceDescription that contains either of InvalidID/InvalidIDString  AnyService/AnyServiceString,
     /// AnyInstance/AnyInstanceString, AnyEvent/AnyEventString.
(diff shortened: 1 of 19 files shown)


iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port_data.hpp
```diff
@@ -49,7 +49,7 @@ struct BasePortData
     BasePortData& operator=(BasePortData&&) = delete;
     ~BasePortData() = default;
 
-    capro::ServiceDescription m_serviceDescription;
+    capro::ServiceDescription m_serviceDescription{capro::InvalidString, capro::InvalidString, capro::InvalidString};
```
elBoberido, line 52: Please initialize the `m_serviceString` etc members in the `service_description.hpp ` with this values instead of doing it everywhere it is used

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
```diff
@@ -29,16 +30,12 @@ namespace iox
 {
 namespace capro
 {
-static constexpr uint16_t InvalidID = 0u;
-static const IdString_t InvalidIDString{"0"};
-static constexpr uint16_t AnyService = 0xFFFFu;
-static constexpr uint16_t AnyInstance = 0xFFFFu;
-static constexpr uint16_t AnyEvent = 0xFFFFu;
+static const IdString_t InvalidString{"INVALID"};
```
elBoberido, line 33: Could this maybe be something more like `#$%&@_INVALID_@&%$#` or something similar which is quite unlike to be accidentally used by a user?

Oh, and why not just `InvalidID`? There is now only one of these constexpr and it's more or less an implementation detail that it is a string.

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
```diff
@@ -29,16 +30,12 @@ namespace iox
...
 static constexpr char AnyServiceString[]{"65535"};
 static constexpr char AnyInstanceString[]{"65535"};
 static constexpr char AnyEventString[]{"65535"};
```
elBoberido, line 36: Please make these a `static constexpr const char foo[]`, then `TruncateToCapacity` is not necessary when converted to an `IdString`. Maybe the this should even be an `IdString` in the first place. I guess with C++14 the `cxx::string` ctor can be made constexpr, at least for string literals ... oh and why is `AnyServiceString` `"65535"`? I guess it's a remnant of the `uint16_t`. Could this maybe also be changed to something like `*` or `#$%&@_*_@&%$#`?

Could the `String` suffix be removed here? There is now only a ctor for `IdString` so there is no ambiguity. Or maybe change it to a `ID` suffix which is shorter.

iceoryx_posh/source/popo/ports/interface_port_data.cpp
```diff
@@ -23,7 +23,7 @@ namespace popo
 {
 InterfacePortData::InterfacePortData(const RuntimeName_t& runtimeName, const capro::Interfaces interface) noexcept
     : BasePortData(capro::ServiceDescription(
-                       capro::InvalidIDString, capro::InvalidIDString, capro::InvalidIDString, {0, 0, 0, 0}, interface),
+                       capro::InvalidString, capro::InvalidString, capro::InvalidString, {0, 0, 0, 0}, interface),
                    runtimeName,
```
elBoberido, line 27: The default ctor should take care or this

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -365,146 +267,120 @@ TEST_F(ServiceDescription_test, ServiceDescriptionCtorWithServiceStringAndInstan
 
 TEST_F(ServiceDescription_test, TwoServiceDescriptionsWithAnyServiceAnyInstanceAnyEventIDsAreEqual)
 {
-    uint16_t testServiceID = iox::capro::AnyService;
-    uint16_t testEventID = iox::capro::AnyEvent;
-    uint16_t testInstanceID = iox::capro::AnyInstance;
-    ServiceDescription serviceDescription1 = ServiceDescription(testServiceID, testEventID, testInstanceID);
-    ServiceDescription serviceDescription2 = ServiceDescription(testServiceID, testEventID, testInstanceID);
+    IdString_t testService = iox::capro::AnyServiceString;
```
elBoberido, line 270: Just a personal preference, but I found `testServiceID` a bit less ambiguous

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -365,146 +267,120 @@ TEST_F(ServiceDescription_test, ServiceDescriptionCtorWithServiceStringAndInstan
...
-    uint16_t testEventID1 = 2U;
-    uint16_t testInstanceID1 = 3U;
-    uint16_t testServiceID2 = iox::capro::AnyService;
-    uint16_t testEventID2 = iox::capro::AnyEvent;
-    uint16_t testInstanceID2 = iox::capro::AnyInstance;
-    ServiceDescription serviceDescription1 = ServiceDescription(testServiceID1, testEventID1, testInstanceID1);
-    ServiceDescription serviceDescription2 = ServiceDescription(testServiceID2, testEventID2, testInstanceID2);
+    IdString_t testService1 = "1";
+    IdString_t testEvent1 = "2";
+    IdString_t testInstance1 = "3";
```
elBoberido, line 284: We need more `hypnotoads`, `brain slugs` and `rock stars`

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
mossmaurice: Those are removed in #854 Happy to change the naming.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/base_port_data.hpp
mossmaurice: Done

iceoryx_posh/test/moduletests/test_capro_service.cpp
mossmaurice: Feel free to push an additional commit :laughing:

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
MatthiasKillat: Ideally we want none of these as actual string constants. Do we need to construct it with those? If no (and I think we should not allow it), then we should handle invalidation internally. If we even need invalid ServiceDescriptions, having invalid objects of any kind is almost always a code smell.

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
```diff
@@ -103,28 +98,23 @@ class ServiceDescription
...
     /// @brief construction of the capro service description using fixed strings to create an event service description
     ServiceDescription(const IdString_t& f_service,
                        const IdString_t& f_instance,
                        const IdString_t& f_event,
                        ClassHash m_classHash = {0u, 0u, 0u, 0u},
                        Interfaces interfaceSource = Interfaces::INTERNAL) noexcept;
 
-    /// @brief compare operator. If wildcards AnyService, AnyInstance or AnyEvent are used as integer IDs, the
-    /// corresponding member comparisons are skipped. Otherwise, both the integer and the string members are compared.
+    /// @brief compare operator. If wildcards AnyServiceString, AnyInstanceString or AnyEventString are used, the
```
MatthiasKillat, line 116: I believe wildcards do not belong in a **concrete** service description. They belong into a search pattern of a search. As soon as wildcards are in play, we have an implicit collection of such descriptions (those matching the pattern). Having those is bad design.

This would also make the comparison operator much more lightweight.

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -141,46 +125,32 @@ ServiceDescription::ServiceDescription(const IdString_t& f_service,
...
-          && cxx::convert::fromString(m_eventString.c_str(), m_eventID)))
-    {
-        m_eventID = InvalidID;
-    }
 }
 
 bool ServiceDescription::operator==(const ServiceDescription& rhs) const
 {
-    if ((m_serviceID != AnyService) && (rhs.m_serviceID != AnyService))
+    if ((m_serviceString != IdString_t(cxx::TruncateToCapacity, AnyServiceString))
```
MatthiasKillat, line 132: Imagining that this logic is always executed I when I search for those gives me the chills. Inefficient, we need to get rid of those wildcards. We can perform the wildcard search differently, as I will propose with the prefix tree approach.

Wildcards are an artifact of the search algorithm, not of the description itself. They should have no bearing on `operator==`. Equality should be simply point-wise (i.e. member-wise) equality here. Even that is inefficient if they are strings but there is no real way around this for now.

iceoryx_posh/source/popo/ports/interface_port_data.cpp
mossmaurice: You mean a c'tor just like `capro::ServiceDescription(Interfaces interfaceSource)`? Seems like `InterfacePort`'s don't need the `ServiceDescription` of the `BasePort`. If you're ok with it, I would like to keep it like that for now as the `InterfacePort`'s will be removed anyway as part of #415.

iceoryx_posh/test/integrationtests/test_roudi_findservice.cpp
```diff
@@ -71,8 +71,8 @@ TEST_F(RoudiFindService_test, OfferServiceWithAnyServiceIdStringDescriptionFails
 
 TEST_F(RoudiFindService_test, OfferServiceWithAnyServiceIdDescriptionFails)
 {
-    auto isServiceOffered =
-        senderRuntime->offerService(iox::capro::ServiceDescription(iox::capro::AnyService, iox::capro::AnyInstance));
+    auto isServiceOffered = senderRuntime->offerService(
+        iox::capro::ServiceDescription(iox::capro::AnyServiceString, iox::capro::AnyInstanceString));
```
MatthiasKillat, line 75: In my opinion this shows that this design is not intuitive at all:  Reading this it appears we are not offering a concrete service, but one with wildcards. But this makes no sense...

We need a separate pattern matching class or algorithm for those AnyStrings, ideally we can get rid of them in the class logic.

I know that this is not part of the PR, but we really should consider this in a redesign.

iceoryx_posh/test/moduletests/test_capro_service.cpp
MatthiasKillat: Considering it is not even really an ID in the strict sense (it is a partial key to identify the service)  it does not matter and shorter is better here.

MatthiasKillat (commented):
Removing the duplication with the integer values internally is good, but having the whole AnyString in this class is bad and needs to be considered. This belongs into search logic, not the class itself, for efficiency and design reasons (separation of concerns).

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -365,146 +267,120 @@ TEST_F(ServiceDescription_test, ServiceDescriptionCtorWithServiceStringAndInstan
...
+    IdString_t testEvent2 = "4";
+    ServiceDescription serviceDescription1 = ServiceDescription(testService1, testEvent1, testInstance1);
+    ServiceDescription serviceDescription2 = ServiceDescription(testService1, testEvent2, testInstance1);
 
     EXPECT_FALSE(serviceDescription1 == serviceDescription2);
 }
 
-TEST_F(ServiceDescription_test, TwoServiceDescriptionsWithDifferentInstanceStringsAreNotEqual)
+TEST_F(ServiceDescription_test, TwoServiceDescriptionsWithDifferentButValidInstancesAreNotEqual)
```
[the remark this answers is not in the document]
mossmaurice: Jep, I've already removed some of them in #854, let's continue the discussion there.

mossmaurice:
> but having the whole AnyString in this class is bad and needs to be considered. This belongs into search logic, not the class itself, for efficiency and design reasons (separation of concerns).

This is being addressed in #854. I've made a few remarks there on why I chose the implementation solution and what would be possible alternatives.

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
mossmaurice: That's implemented in #854, let's move the discussion over there.

iceoryx_posh/source/capro/service_description.cpp
mossmaurice: See [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/852#discussion_r652905346)

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -141,46 +125,32 @@ ServiceDescription::ServiceDescription(const IdString_t& f_service,
...
+        && (rhs.m_serviceString != IdString_t(cxx::TruncateToCapacity, AnyServiceString)))
```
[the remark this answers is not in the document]
mossmaurice: Mostly gone with #854. However, as noted in the [description](https://github.com/eclipse-iceoryx/iceoryx/pull/852#issue-671421972), we might at some point optimise the string usage inside the `ServiceDescription` (e.g. by mapping it to integers without the user noticing)

iceoryx_posh/test/integrationtests/test_roudi_findservice.cpp
mossmaurice: This test case is removed in #854. Also see [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/852#discussion_r652905346)

iceoryx_posh/source/popo/ports/interface_port_data.cpp
elBoberido: No, I mean a `ServiceDescription foo;` should set the members to the `capro::InvalidString`, etc.

Then this would become `BasePortData(ServiceDescription(), runtimeName, ...` or even `BasePortData({}, runtimeName, ...`

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
```diff
@@ -78,8 +79,9 @@ GatewayGeneric<channel_t, gateway_t>::addChannel(const capro::ServiceDescription
                                                  const IceoryxPubSubOptions& options) noexcept
 {
     // Filter out wildcard services
-    if (service.getServiceID() == capro::AnyService || service.getInstanceID() == capro::AnyInstance
-        || service.getEventID() == capro::AnyEvent)
+    if (service.getServiceIDString() == capro::IdString_t(cxx::TruncateToCapacity, capro::AnyServiceString)
```
elfenpiff, line 82: This logic should be part of the service description. Could you please create a method called `doesContainWildcards()` and call this method here. Maybe you could come up with an even better name...

elfenpiff (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
mossmaurice: I came to the conclusion that we need to differentiate between searching in the `ServiceRegistry` and creating `ServiceDescription`'s. Creating a `ServiceDescription` with a wildcard does not make sense semantically from my point of view. How can a `ServiceDescription` represent all of them with a wildcard? And even worse if you compare them, the equality operator would state them to be equal! They are not. They should always be well defined. This is why I removed them from `ServiceDescription` and added a possibility to search for Wildcard in #854. We can adapt the DDS gateway in a follow-up PR. Let's move this discussion over to #854. 

Also, see [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/852#discussion_r652853192) from @MatthiasKillat

iceoryx_posh/source/popo/ports/interface_port_data.cpp
elBoberido: oh, I overlooked the `interface` parameter since it is hidden due to (insert some swearing regarding this artificial UI limitation to 800px or whatever even it there is so much space left and right) github UI

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -104,32 +103,17 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& f_serial) noexc
...
 {
 }
 
-ServiceDescription::ServiceDescription(uint16_t f_serviceID, uint16_t f_instanceID) noexcept
-    : ServiceDescription(f_serviceID, InvalidID, f_instanceID)
-{
-    m_hasServiceOnlyDescription = true;
-}
-
+/// @todo remove
```
elBoberido, line 110: please always add an issue number to a todo

mossmaurice: I agree with you. Next time I'll add those. However, in this case all `@todo`'s are already removed in either #853 or #854 so there's no need to track them. If I add them now, this will result in merge conflicts.

elBoberido (approved):

MatthiasKillat (approved):

Issue #859, opened 2021-06-18
Implement a prefix tree for fast string search


MatthiasKillat:
## Brief feature description

Prefix Tree (Trie)
Allows fast searching of strings, to be used in Service Registry.
String search is then performed in time proportional to the string to be searched.

## General Constraints
- relocatable - we want to use it in shared memory
- not thread safe - too much effort / runtime cost
- keys are always strings
- values are generic
- multiple values for a key (hence it is not really a unique key)
- self contained memory, static memory

## Operations
1. Insert key, value  O(#key + #values to key)
2. Find all values to a key O(#key + #result)
3. Remove all values to a key O(#key + #values removed)
4. Remove a specific value to a key O(#key + #values to key)

- # denotes the size of the specific quantity, i.e. #key the length of the key 

## Detailed information

TBD

- string type for API - internally we should use `char*` for efficiency

elfenpiff:
@MatthiasKillat why not implement a `cxx::map` which is more generic and can be used in more places of iceoryx?

MatthiasKillat:
@elfenpiff I will do that, but a good map is more complicated and does not sufficiently solve the problems ahead. It fits the problem better and apart from static memory and some interface convenience I already have it implemented (still a way to go, but doable).

The prefix tree is actually very useful also for introspection (where we also currently use a map) and other applications, I would say. It is useful whenever we map strings (theoretically any kind of array with equality-comparable symbols like characters, but I will not generalize this for now) to something else, like ids or any object that describes the data to the key. It locates this data much faster than a map (in theory and I will measure it also in practice against `std::map`).

As for use cases, there are many more then we may think. In ROS or other communication standards (AUTSOAR), user ids are often defined as strings on user level and this cannot be changed (without changing the standard). Working with those ids internally is problematic from a storage and lookup point of view.

But you can use a trie as some kind of very fast string lookup mechanism, which maps to internal ids (such systemwide unique numbers) or the data itself. It can be used to build a system-wide entity-dictionary for fast lookup like this. The service registry is exactly this, but only for services. It also comes with prefix search for free, unlike maps (find all entities that begin with a certain prefix).

[...]

PR #854, opened 2021-06-16
Iox #415 Refactor Service Description Part3: Restrict wildcard usage to searching only

mossmaurice:
## Pre-Review Checklist for the PR Author

1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/best-practice-for-testing.md

This is the third part of a series of pull requests regarding the refactoring of `ServiceDescription`, which paves the way for a new `ServiceRegistry`.

**Done:**
* Add `isValid()` checks in `PortManager` and friends to make sure no invalid `ServiceDescription` is saved into the `ServiceRegistry`
  * Define empty strings as invalid
* Allow wildcards only for searching the `ServiceRegistry` to have well-defined `ServiceDescription`'s

**To be done in follow-up PR as part of #415:**
* Move the`Any_t` from `posh_runtime.hpp` to the new user-facing search discovery header
* [x] Implement `template<class T> cxx::variant(T&& t)` to have a easier user API for searching the `ServiceRegistry` (Done in #864)

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

[...]

PR #853, opened 2021-06-16
Iox #415 Refactor Service Description Part2: Remove c'tor with two args

mossmaurice:
## Pre-Review Checklist for the PR Author

1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/best-practice-for-testing.md

## Notes for Reviewer

> Please make sure to start the review with #852 before reviewing this pull request!

This is the second part of a series of pull requests regarding the refactoring of `ServiceDescription`, which paves the way for a new `ServiceRegistry`.

**Done:**
* Remove c'tor of `ServiceDescription` with two args

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

- Closes partly #415

Issue #830, opened 2021-06-01
Create a set data structure for the iceoryx hoofs library


MatthiasKillat:
## Brief feature description

We need a data structure which admits efficient management of values which satisfies the requirements

- replacement for `std::set` in its core functionality: insertion, finding and removing of values
- no duplicate values
- no full interface `std::set` interface, we start with a custom interface
- efficient operations, O(log(n)) in the long run
- memory static (by bounded capacity) and no exceptions 
- relocatable if required by the user (to be able to send it via iceoryx)

## Detailed information

 - a relocatable set will incur overhead, but we can have two implementations differing on only a internal template type to have 
 - can be implemented (with some inefficiency) in terms of the map in #829 
 - if not implemented in terms of #829 will need a specific tree implementation applicable only for values (instead of key value pairs), note that reusing a tree implementation for both by specialized `operator<` for pairs is not recommended and causes inefficiencies and hard to grasp code (it appears)
 - the interface can be extended later to an STL one based on the custom interface (a full one is quite an effort for a lot of syntactic sugar)

To be further refined.

PR #860, opened 2021-06-21
Iox #415 Make `ServiceRegistry` aware of complete `ServiceDescription`

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
This pull request implements a new `ServiceRegistry`, which stores the complete `ServiceDescription` and takes n:m communication into account. 

**Will be done in follow-up PR:**
* Replace `std::multimap` with `cxx::trie` (will be done in #859)
* Increase the number of services a user can search for

## References

- Closes partly #415


iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,164 @@
...
 namespace iox
 {
 namespace roudi
 {
-void ServiceRegistry::add(const CaproIdString_t& service, const CaproIdString_t& instance)
+cxx::expected<ServiceRegistry::ServiceRegistryError>
+ServiceRegistry::add(const capro::ServiceDescription& serviceDescription)
 {
-    cxx::set::add(m_serviceMap[service].instanceSet, instance);
+    // Forbid duplicate service descriptions entries
```
elfenpiff, line 29: Please use the `cxx::set` again or a `std::set` and replace it later with an optimized implementation (the `set` is one part of the `map` implementation).

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,164 @@
...
+    for (auto& element : m_serviceDescriptionVector)
+    {
+        if (element == serviceDescription)
+        {
+            return cxx::error<ServiceRegistryError>(ServiceRegistryError::SERVICE_DESCRIPTION_ALREADY_ADDED);
```
elfenpiff, line 34: This error just increases the error handling overhead and has no value for the user. If a user calls `add`, there are two possibilities:
1. it is not added, but after the call -> it is added
2. it is added, and after the call -> it is still added

Therefore, the user does not require any action here and if they would like to know if a certain service was registered they always can use `find`.

iceoryx_posh/source/runtime/posh_runtime_impl.cpp
```diff
@@ -383,26 +383,26 @@ PoshRuntimeImpl::findService(const capro::ServiceDescription& serviceDescription
...
+    ServiceContainer serviceContainer;
     uint32_t numberOfElements = requestResponse.getNumberOfElements();
-    uint32_t capacity = static_cast<uint32_t>(instanceContainer.capacity());
+    uint32_t capacity = static_cast<uint32_t>(serviceContainer.capacity());
 
-    // Limit the instances (max value is the capacity of instanceContainer)
-    uint32_t numberOfInstances = ((numberOfElements > capacity) ? capacity : numberOfElements);
-    for (uint32_t i = 0; i < numberOfInstances; ++i)
+    // Limit the services (max value is the capacity of serviceContainer)
+    uint32_t numberOfServices = ((numberOfElements > capacity) ? capacity : numberOfElements);
```
elfenpiff, line 391: Please use `cxx::min(capacity, numberOfElements);`

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,164 @@
...
+    }
+    m_serviceMap.insert({serviceDescription.getServiceIDString(), m_serviceDescriptionVector.size() - 1});
+    m_instanceMap.insert({serviceDescription.getInstanceIDString(), m_serviceDescriptionVector.size() - 1});
+    return cxx::success<>();
 }
 
-void ServiceRegistry::remove(const CaproIdString_t& service, const CaproIdString_t& instance)
+bool ServiceRegistry::remove(const capro::ServiceDescription& serviceDescription)
 {
-    cxx::set::remove(m_serviceMap[service].instanceSet, instance);
```
elfenpiff, line 30: Please use the suggested data structure: `std::map<std::map<CaproIdString_t, std::set<CaproIdString_t>>, CaproIdString_t> m_services` and refactor this method.

The implementation would be.
1. Remove event entry from `std::set<CaproIdString_t>`
2. If the set from 1. is empty remove the instance entry from `std::map<CaproIdString_t, std::set<CaproIdString_t>>`
3. If the map from 2. is empty remove the service entry from `m_services.

You can safely remove `m_serviceDescriptionVector`, `m_instanceMap` and `m_serviceMap` since the data structure from above is sufficient enough

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,164 @@
...
+
+    return removedElement;
 }
 
-void ServiceRegistry::find(InstanceSet_t& instances,
+void ServiceRegistry::find(ServiceDescriptionVector_t& searchResult,
                            const CaproIdString_t& service,
                            const CaproIdString_t& instance) const
 {
-    if (instance == iox::cxx::string<100>(capro::AnyInstanceString))
```
elfenpiff, line 37: Please refactor find as well with the suggested data structure: `std::map<std::map<CaproIdString_t, std::set<CaproIdString_t>>, CaproIdString_t> m_services`.

Then you can write the algorithm with much less code, somehow like this:
```cpp
cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> searchResult;

auto serviceStartIter = (service == capro::AnyServiceString) ? m_service.begin() : m_services.find(service);
auto serviceEndIter = (service == capro::AnyServiceString) ? m_service.end() : serviceStartIter + 1;

for(auto iter = serviceStartIter; iter != m_service.end() && iter != serviceEndIter; ++iter) {
  auto instanceStartIter = (service == capro::AnyInstanceString) ? iter.begin() : iter.find(service);
  auto instanceEndIter = (service == capro::AnyInstanceString) ? iter.end() : instanceStartIter + 1;
  for(auto iter2 = instanceStartIter; iter2 != iter.end() && iter2 != instanceEndIter; ++iter2) {
     for(auto event : iter2 )
       searchResult.emplace_back(iter.first, iter2.first, event);
  }
}
// maybe you have to rename iter to serviceIter, iter2 to instanceIter but this could would be much cleaner and faster
```

elfenpiff (commented):
First part reviewed - except the tests

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: A set is not part of a map implementation. Our current `cxx::set` abstracts away the duplicate removal but not in an efficient way. Still, I already discussed it with @mossmaurice and what we need for fast removal and insertion is a structure with 1) random access and 2) efficient insertion and removal. A set cannot help us with 1) in the long term, we need a vector with some addons, quite similar to the `Objectpool`. But the Objectpool can be improved to allow O(1) insertion. 

Long story short we can use the `cxx::set` abstraction for now but need some thin wrapper around a vector with properties 1) and 2) later.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -30,24 +33,44 @@ namespace roudi
...
+    /// @brief Removes given service description from registry
+    /// @param[in] serviceDescription, service to be removed
+    /// @return true, if service description was removed, false otherwise
+    bool remove(const capro::ServiceDescription& serviceDescription);
+
+    /// @brief Removes given service description from registry
+    /// @param[in] searchResult, reference to the vector which will be filled with the results
+    /// @param[in] service, string or wildcard to search for
+    /// @param[in] service, string or wildcard to search for
+    void find(ServiceDescriptionVector_t& searchResult,
```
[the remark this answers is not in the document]
MatthiasKillat: It may be easier more natural to use like this, but consider the case where we can store 1000 services or more.

There is a reason to have an API like this with our memory constraints, and this is that the return type would always need to support maximum number of elements in the returned container. We have no truly dynamic containers and cannot easily (or at all) define those. It is possible that the user may define the return type as a template parameter, but this is also bad to use for return values (cannot be inferred, needs to be specified explicitly).

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: Yes, it is also not an error if it already exists. Whether the user should be informed is debatable. I am fine with either way, just do not classify it as error.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: I need to look a at the current implementation but I am sure that any solution with maps and sets is inferior to prefix tree implementations I propose (and the latter is easier to implement than a map or set).

The reason is simple: is completing most operations in O(#length of string to insert), as opposed to anything with maps which at least need O(log(#elements in container)*#length of keys to compare). These structures are much harder to implement with the logarithmic guarantees as well, compared to linear in the key length. This is especially true for the multimaps needed here. 

I have not checked the correctness of the above code, but if it is correct the estimate still holds true and gets worse with any iteration done above.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -30,24 +33,44 @@ namespace roudi
...
+    enum class ServiceRegistryError
     {
-        InstanceSet_t instanceSet;
+        INVALID_STATE,
+        SERVICE_DESCRIPTION_ALREADY_ADDED,
+        SERVICE_REGISTRY_FULL,
     };
-    using serviceMap_t = std::map<CaproIdString_t, instance_t>;
+    /// @todo #415 should be connected with iox::MAX_NUMBER_OF_SERVICES
+    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 100U;
```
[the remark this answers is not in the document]
MatthiasKillat: Note that we need to consider number of services in the order of 1000 in our data structure design. So 100 is deceptively low here, but of course there will be a constant.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -30,24 +33,44 @@ namespace roudi
...
+
+    /// @brief Adds given service description to registry
+    /// @param[in] serviceDescription, service to be added
+    /// @return ServiceRegistryError, error wrapped in cxx::expected
+    cxx::expected<ServiceRegistryError> add(const capro::ServiceDescription& serviceDescription);
+
+    /// @brief Removes given service description from registry
+    /// @param[in] serviceDescription, service to be removed
+    /// @return true, if service description was removed, false otherwise
+    bool remove(const capro::ServiceDescription& serviceDescription);
```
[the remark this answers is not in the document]
MatthiasKillat: I think it is intended to indicate whether the service was there and removed or was not there and not removed. The logical state of the registry is the same in both cases, so we do not really need to communicate this. Personally I do those kind of things as well for more information if the user wants (the information was already computed and passing it is inexpensive). The user is technically free to ignore the return value (I know there are differing opinions on this one).

This kind of style is definitely used in several APIs, but I am fine either way.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -30,24 +33,44 @@ namespace roudi
 class ServiceRegistry
 {
   public:
-    static constexpr uint32_t MAX_INSTANCES_PER_SERVICE = 100u;
-    using CaproIdString_t = capro::IdString_t;
-    using InstanceSet_t = cxx::vector<CaproIdString_t, MAX_INSTANCES_PER_SERVICE>;
-    struct instance_t
+    enum class ServiceRegistryError
```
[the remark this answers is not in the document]
elBoberido: +1 for `ServiceRegistry::Error`

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -30,24 +33,44 @@ namespace roudi
...
-              const CaproIdString_t& instance = capro::AnyInstanceString) const;
-    const serviceMap_t& getServiceMap() const;
+    /// @brief Returns all service descriptions as copy
+    /// @return ServiceDescriptionVector_t, copy of complete service registry
+    const ServiceDescriptionVector_t getAllServices() const;
 
   private:
-    mutable serviceMap_t m_serviceMap;
+    /// @todo #859 replace std::map with prefix tree
+    ::std::multimap<CaproIdString_t, uint64_t> m_serviceMap;
```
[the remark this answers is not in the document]
MatthiasKillat: We cannot use a structure like this in a static memory world, it just does not work resource wise. The way it is designed in the PR it must be a multimap. 

 It will not be a multimap in the end anyway but a combination of two prefix trees and some vector to store the actual data. This will be theoretically optimal for all operations, unlike the above construct which cannot be (regardless of how it is used due to logarithmic lower bounds  of map operations for optimal implementations).
I do not know how this is supposed to work from the code alone, but I suspect there is a misunderstanding of what exactly the underlying algorithm needs to accomplish. For example we want to store full descriptor objects `ServiceDescription', not just strings which is missing in the proposal above. This is due to the fact that it can and will contain additional data describing the service in the future (e.g. for DDS).

I would also argue `fast access` and efficiency  cannot be achieved with the proposed structure. Furthermore the memory requirements if there are capacities involved for these structures are prohibitive.
Logically a solution with tries is also very concise and efficient, we just need handle the actual storage. 

I for one am not able to provide map and set implementations which are both space and time efficient at the same time given our constraints and would efficiently work here.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: As described in another comment this structure is inefficient for storage (considering we would need static capacities for all of them) and also and does not lend itself for efficient operations, especially wildcard search. Furthermore the way it is proposed it it does not allow us to locate a full `ServiceDescription` object which will contain more than 3 strings in the foreseeable future - we will need to embed custom information of DDS topics I think (like QoS), have not fully thought it through. But even if we do not need it for QoS, it is good to have to attach all information about a service here in one place, or at least point to it. 

I am also not sure how a map as key would really behave, since it has no natural `operator<` which is a requirement for key types (would maybe fall back to addresses of the keys here, but this would be wrong for non-dynamic objects). Besides any map where keys or values are more than lightweight objects is doomed efficiency wise, since insertion and removal require copies of internal data (which would consist of maps of sets here). This is just not going to work efficiently

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
mossmaurice: That makes sense, adapted. Lately I've seen it being added as a nested class, which I prefer instead of outside.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
mossmaurice: Well, I had a quick try. However, due to the additional 5 Introspection topics, some tests were failing. I'd prefer to do this in one go, once we have the user-facing `DiscoveryInfo` API (or what ever other name we choose). Is that ok?

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
mossmaurice: > I think it is intended to indicate whether the service was there and removed or was not there and not removed. 

Exactly, that's the idea. There shall be the possibility to check whether remove was successful or remove was unsuccessful, when the element was not contained in the registry.

> And when the services was not added it is still removed.

That's not entirely correct. It will not be removed if it wasn't added before. The `bool` is used to signal this. I'll keep the additional info to be able to pass this to the user via the new `ServiceDiscovery` API.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
mossmaurice: @elfenpiff Please have a look at commit 47e89f10e2f916c3c3a1aca84a996096c31fa8d0, where I estimated the complexity of an approach quite similar to yours. I used `std::multimap<std::pair<CaproIdString_t, CaproIdString_t>, `ServiceDescription`>` (Attempt A).

```
find(K1,K2): O(log n+ #result)
find(K1,*) or find(*,K2): O(#entriesInMap + n^3)
find(*,*): O(#entriesInMap + #entriesInMap + n^3)

where K1 is the service string and K2 the instance string
```

On the other hand with the current approach in this PR (Attempt B):

```
find(K1,K2): O(log n + log n + max(#PossibleServices + #possiblesInstances) + #intersection)
find(K1,*) or find(*,K2): O(log n + #result)
find(*,*): O(1)

where K1 is the service string and K2 the instance string
```

IMHO the attempt B is superior to the attempt A. Of course your approach with using a nested `set` might speed up things. However, we would still need to search with `equal_range` (O(log n)) in two maps.

> Why is it a multimap?

`multimap`'s are used as there a service or instance string can be used more than once, imagine e.g.:
|Service (1st `multimap`) | Instance (2nd `multimap`) | Event|
|----|----|-----|
|Radar|FrontLeft|Objects|
|Radar|FrontLeft|Odometry|
|Radar|FrontRight|Objects|
|Video|FrontLeft|Objects|

> For example we want to store full descriptor objects `ServiceDescription', not just strings which is missing in the proposal above. This is due to the fact that it can and will contain additional data describing the service in the future (e.g. for DDS).

That exactly, the reason. Thanks for pointing that out! For comparability with DDS it might be useful to have e.g. QoS settings attached to every `ServiceDescription.`

> I would also argue fast access and efficiency cannot be achieved with the proposed structure. 

I second that! IMHO a prefix tree is exactly what we need for efficient string look-ups. Let's have a look at the first code snippets of #859 and then do the final decision here.

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
```diff
@@ -83,10 +83,10 @@ class PoshRuntime
 
     /// @brief find all services that match the provided service description
     /// @param[in] serviceDescription service to search for
-    /// @return cxx::expected<InstanceContainer, FindServiceError>
-    /// InstanceContainer: on success, container that is filled with all matching instances
+    /// @return cxx::expected<ServiceContainer, FindServiceError>
+    /// ServiceContainer: on success, container that is filled with all matching service descriptions
     /// FindServiceError: if any, encountered during the operation
-    virtual cxx::expected<InstanceContainer, FindServiceError>
+    virtual cxx::expected<ServiceContainer, FindServiceError>
```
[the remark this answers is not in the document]
mossmaurice: > It is very easy to increase the instance container size to iox::MAX_NUMBER_OF_SERVICES and then this can never happen.

Indeed, will do.

> Otherwise we increase again the error handling in unnecessary cases and it would be inconsistent with the behavior of the other roudi communicating methods.

I'm tempted to disagree here. This class will be used quite close to the user-facing API, so we need to return the errors to the user in most case. Will have a thought about this.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
mossmaurice: As @MatthiasKillat in general I agree with you @elfenpiff a `auto result = registry.find("bla", "blubb", "fuu")` is much nicer. A typical ADAS or robotics system will for sure have something around 1000 services. We would not let all the services be findable, but let's assume a size of 200. That's a already a huge number and it will have some runtime effect as we would return lots of `cxx::string`'s via copy here. I'd propose to do a measurement once all building blocks are in place for the new discovery API. This can easily be changed as it is no user-facing API.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: I don't see why we should use our `cxx::set` wrapper here. To me it seems like we need a bit more advanced but still light-weight data structure than the `cxx::vector`. @elfenpiff Can you elaborate?

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: See my answer in the [other thread](https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r655665071)

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: @elfenpiff That's a neat and small implementation of `find()`! However, as stated in the other threads I'm not convinced that we will be able to build a fixed size `cxx::map`, that can be place in shared memory. What is your take on that?

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: To my knowledge `cxx::set` is a wrapper around a vector which takes care of duplicate elimination, albeit in an inefficient way (linear search). If we can use this instead of implementing duplicate elimination here, we should.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: To elaborate: Map keys need `operator<`, but if it does not exist it will use `std::less` in the map I think and use comparison on pointers to dynamically allocated data objects (allocated by the map). In other words: pointers to memory are the real key in this case, which is fine. Not fully sure though, our own map would have to behave in some reasonable way there as well and I do not know yet what this would be.

The storage size problem would be severe in any case for nested fixed-size containers.
This means we would need to implement a complex solution with a lot of drawbacks. We can have a simpler (still fairly complex) solution which at least avoids some of the drawbacks, is faster, generalizes in a better way to e.g. more keys, allows prefix based search and more.
The data structure itself is also somewhat easier to implement. But I fully intend to create the fixed-size map as well, just that we can design better algorithms use suitable abstractions instead for other problems.

iceoryx_posh/source/roudi/service_registry.cpp
elfenpiff: This should be feasable. I see a map implementation as a challenge but not the static memory part and in some previous company I already have seen such a map. The tree can be implemented in the same fashion like the linked list which we already have (from the memory point of view).

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
elfenpiff: > > Why is it a multimap?

> multimap's are used as there are service or instance string that can be used more than once, imagine e.g.:

This is why one should use a `std::map<CaproIdString_t, std::map<CaproIdString_t, std::set<CaproIdString_t>>` then you have quiet an efficient layout of containers which can be reused. The more I think about it the more I doubt that the `cxx::trie` prefix tree is the right data structure for this problem.

1. We most likely will not reuse it somewhere else in iceoryx
2. But there are multiple places where a `cxx::map` can be reused
3. `cxx::trie` implementation is complex, cost a lot of effort and does not gain much performance (especially for a task like service registration which is not on the hot path)
4. A very good alternative is just two maps and a set.

Here I somehow fear we are pursuing the wrong approach. Normally we say:
- Make it work, make it beautiful and then make it fast

Here it seems we start with make it fast for a use case outside the hot path and want to pay for a very specialized data structure.

@MatthiasKillat
> The way it is designed in the PR it must be a multimap. We want to support fast wildcard searches and the individual strings do not uniquely identify the ServiceDescription.

https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r655486388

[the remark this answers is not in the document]
elfenpiff: @mossmaurice For our use case runtime estimation with the big O notations are the wrong approach.

Let's take a look the definition: https://en.wikipedia.org/wiki/Big_O_notation and apply it to `find(K1,K2) = O(log n+ #results)`.

This means we have: ` | find(K1, K2) | < M * (log(n) + #result)` for a real number M > 0. In our use case `n` is the number of services and lets say we grow and support a massive amount of services like `10000` (at the moment it is 512). Therefore we have `log2(10000) = 13.2877`. For 100000 it would be `log2(100000) = 16.609`.

This means the number of operations for 10000 are `|find(K1, K2)| < M * (13.2877 + #result)`.

The big o notations provides you runtime estimation for huge numbers and we are having tiny numbers here and then the constant `M` becomes very important.

Since one approach can have a constant with a value of `M = 2` and another one with `M = 1000`. Now lets use it for the runtime estimation for your two approaches with n = 100000 (number of services). For simplicity, to demonstrate my argument, I get rid of all the `#result` etc stuff.

Approach A with `M=1000`: `|find(K1, K2)| < M * (log(n)) = 1000 * log2(100000) = 1000 * 16.609 = 16609`
Approach B with `M=2`: `|find(K1, K2)| < M * (log(n) * log(n)) = 2 * (log2(100000) + log2(100000)) = 2 * (16.609 + 16.609) = 2 * 33.218 = 66.436`.

As you can see your inferior algorithm B can be much faster even the big o notation states otherwise. The big o notation states only that for `n` marching to infinity approach A is faster.


This is why premature optimization is the root of all evil since we do not know the value of `M`. The value of `M` can be high or low and this number is influenced by the CPU cache, cache locality of the data etc. things which are extremely hard to measure in a complex algorithm.

To conclude this do not optimize prematurely, use a common data structure like the `std::map` or upcoming `cxx::map` to solve this in a readable manner and then write a benchmark for optimization purposes when we need it.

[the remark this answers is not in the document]
MatthiasKillat: I understand O Notation and the notions of hidden constants (ignoring them for the estimate is the whole point, they are implementation dependent anyway). It is also of no use to insert specific numbers here, logarithms are small, I get this (no point in computing logarithms here). Nevertheless for these algorithms the estimates hold for small values already, so O Notation is not wrong to use here to get an estimate. The constant is important, but we will not know it regardless of approach.

Also, much of the time is used for many string comparisons, each of them being fairly expensive (number of them is logarithmic for each find call in a map).

Further: it is not necessarily `find(K1,K2)` which is expensive, but find(K1, *) since it naively leads to some quadratic complexity and improper use of maps to perform the search (iteration and such). And quadratic complexity is felt early, already for hundreds of elements with string comparisons being in play.
 
I am very concerned about the nesting of structures like map which will blow up memory consumption.

This is therefore not premature optimization, it is using the right tool for the job. Furthermore, if the abstraction is defined properly it will be very readable, the complexity is hidden in the search structure itself. 

The map with logarithmic guarantees is **hard** to implement, even more so with our static memory constraints. And a naive map will have e.g. linear search complexity which will be a problem (not felt in something like our icedelivery example) but very much so for real applications with hundreds or thousands of services, so we need to do it properly from the start.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: This is true, it is feasible with the same approach as the list, but it would still be the wrong structure for the job due to this nesting of maps and sets. 

And it is a challenge I intend to tackle in the near future when we have a working approach. We would even be able to compare efficiency then, and my bet is against any map based approach (we can try with `std::map`, which is likely better optimized then what we will be able to come up without serious time investment).

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
MatthiasKillat: ad 1) This depends on what the future will bring, we cannot know. Still a useful abstraction and not a hugely complicated one.

ad 2) That is why we will also implement it, but we would need a set here to. While they are similar and will share some code, in a proper implementation (tree structure is almost the ame...) they need their own implementation and so on. Currently it is just used for introspection as far as I know, but yes it is a useful structure to have and we may be able to avoid some bad designs in the future with a map.

ad 3) The trie implementation is less complex then the map, especially with a minimalistic interface only. I have a dynamic memory prototype already (with an incomplete interface but basic operations). Hot path

ad 4) I dispute that for reasons stated, especially with static memory where all these structures need a well chosen capacity and it is constants gallore everywhere with some weird interactions like `MAXIMUM_NUMBER_OF_INSTANCES_PER_SERVICE` and the same for events and similar. We had these kind of problems in the past, only the involved containers where simpler (nested `vector`).

iceoryx_posh/test/integrationtests/test_roudi_findservice.cpp
```diff
@@ -358,7 +386,7 @@ TEST_F(RoudiFindService_test, InterfacePort)
         auto caproMessage = maybeCaProMessage.value();
         if ((caproMessage.m_serviceDescription.getServiceIDString() == IdString_t("service1"))
             && (caproMessage.m_serviceDescription.getInstanceIDString() == IdString_t("instance1"))
-            && ((caproMessage.m_serviceDescription.getEventIDString() == IdString_t(iox::capro::AnyEventString))))
+            && ((caproMessage.m_serviceDescription.getEventIDString() == IdString_t(iox::capro::InvalidString))))
```
elBoberido, line 389: Was this change intentional? I guess this changes with #854?

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,163 @@
...
+    // O(log n + log n + max(#PossibleServices + #possiblesInstances) + #intersection)
+    if (instance != capro::IdString_t(capro::AnyInstanceString)
+        && service != capro::IdString_t(capro::AnyServiceString))
     {
-        for (auto& instance : m_serviceMap[service].instanceSet)
+        cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> possibleServices;
+        cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> possibleInstances;
+
+        auto rangeServiceMap = m_serviceMap.equal_range(service);
+        for (auto entry = rangeServiceMap.first; entry != rangeServiceMap.second; ++entry)
```
elBoberido, line 119: cheers for inconsistencies in the STL for not being able to use range-based for loops here

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,163 @@
...
     else
     {
-        auto& instanceSet = m_serviceMap[service].instanceSet;
-        auto iter = std::find(instanceSet.begin(), instanceSet.end(), instance);
-        if (iter != instanceSet.end())
+        // Find (*, K2)
+        // O(log n + #result)
+        if (service == capro::IdString_t(capro::AnyServiceString)
+            && instance != capro::IdString_t(capro::AnyInstanceString))
+        {
```
elBoberido, line 147: Please make this an `else_if`. It will reduce indentation and gives you 4 more character before clang-format wraps the line ;)

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,163 @@
...
+        // Find (K1, *)
+        // O(log n + #result)
+        else if (instance == capro::IdString_t(capro::AnyInstanceString)
+                 && service != capro::IdString_t(capro::AnyServiceString))
+        {
+            auto range = m_serviceMap.equal_range(service);
+            for (auto entry = range.first; entry != range.second; ++entry)
+            {
+                searchResult.push_back(m_serviceDescriptionVector[entry->second]);
+            }
```
elBoberido, line 163: Since this is used quite often, can we have a lambda which could be used like
```cpp
storeIndices(containerToStore, range);
```

elBoberido (commented):
service registry tests not yet reviewed

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -42,6 +42,14 @@ class PublisherPortUser;
 class SubscriberPortRouDi;
 class SubscriberPortUser;
 } // namespace popo
+namespace capro
+{
+class ServiceDescription;
+}
+namespace posix
+{
+class UnixDomainSocket;
```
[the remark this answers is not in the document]
mossmaurice: That must have slipped in somehow. Removed.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -30,24 +33,44 @@ namespace roudi
...
+    /// @return true, if service description was removed, false otherwise
+    bool remove(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    /// @brief Removes given service description from registry
+    /// @param[in] searchResult, reference to the vector which will be filled with the results
+    /// @param[in] service, string or wildcard to search for
+    /// @param[in] service, string or wildcard to search for
+    void find(ServiceDescriptionVector_t& searchResult,
+              const CaproIdString_t& service = capro::AnyServiceString,
+              const CaproIdString_t& instinstanceance = capro::AnyInstanceString) const noexcept;
```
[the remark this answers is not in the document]
mossmaurice: You're right, this is one of the places where #854 touches this PR. I'll update this one once #854 is on `master`.

iceoryx_posh/test/integrationtests/test_roudi_findservice.cpp
mossmaurice: Yes, to both questions. Otherwise the test would fail.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -708,17 +697,17 @@ popo::ApplicationPortData* PortManager::acquireApplicationPortData(const Runtime
     }
 }
 
-void PortManager::addEntryToServiceRegistry(const capro::IdString_t& service,
-                                            const capro::IdString_t& instance) noexcept
+void PortManager::addEntryToServiceRegistry(const capro::ServiceDescription& service) noexcept
 {
-    m_serviceRegistry.add(service, instance);
+    m_serviceRegistry.add(service).or_else([](auto&) {
+        /// @todo #415 return something?
```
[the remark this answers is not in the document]
mossmaurice: Added

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,163 @@
...
+                    removedValue = it->second;
+                    it = m_serviceMap.erase(it);
+                    removedEntry = true;
+                    continue;
+                }
+                else if (removedEntry && it->second > removedValue)
+                {
+                    // update index due to removed element
+                    it->second--;
+                }
```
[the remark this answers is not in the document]
mossmaurice: The latter branch is only taken if a matching index was found aka `removedEntry == true`. However, you're right that things are a bit entangled here. Refactored to make it more understandable.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: Done

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: With only two occurrences, I prefer to keep things how they are. `ServiceRegistry::find` is very readable and fits well on one monitor currently.

elBoberido: well, I count 4 ;)

CHANGELOG.md
```diff
@@ -105,6 +105,20 @@ ServiceDescription("First", "Second", "DontCare") myServiceDescription2;
...
+```cpp
+// before
+poshRuntime.offerService(myServiceDescription);
+poshRuntime.stopOfferService(myServiceDescription);
+poshRuntime.findService(myServiceDescription);
+
+// after
+discoveryInfo.offerService(myServiceDescription);
+discoveryInfo.stopOfferService(myServiceDescription);
+discoveryInfo.findService("ServiceA", Wildcard_t);
```
elBoberido, line 119: To make this comparable, the `findService` from the `before` part needs to have a service description with `AnyInstance`. Could you please adjust it?

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -32,23 +34,44 @@ static const capro::IdString_t Wildcard{"*"};
...
-    static constexpr uint32_t MAX_INSTANCES_PER_SERVICE = 100u;
-    using InstanceSet_t = cxx::vector<capro::IdString_t, MAX_INSTANCES_PER_SERVICE>;
-    struct instance_t
+    enum class Error
     {
-        InstanceSet_t instanceSet;
+        INVALID_STATE,
+        SERVICE_DESCRIPTION_ALREADY_ADDED,
+        SERVICE_REGISTRY_FULL,
+        SERVICE_DESCRIPTION_INVALID,
```
elBoberido, line 42: Today I talked with @MatthiasKillat about his `PrefixTree` implementation and he told me that is should be possible to insert empty strings as key. The `std::multimap` also supports empty strings as keys. Therefore we could have a `ServiceDescription` with empty strings and our code wouldn't break. Since there is no technical reason for this limitation, I would suggest to allow such `ServiceDescription` since it liberates us from the burden to handle this scenario.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -16,45 +17,163 @@
...
+        {
+            return cxx::error<Error>(Error::SERVICE_DESCRIPTION_ALREADY_ADDED);
+        }
+    }
+
+    if (!m_serviceDescriptionVector.push_back(serviceDescription))
+    {
+        return cxx::error<Error>(Error::SERVICE_REGISTRY_FULL);
+    }
+    m_serviceMap.insert({serviceDescription.getServiceIDString(), m_serviceDescriptionVector.size() - 1});
```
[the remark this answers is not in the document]
elBoberido: Any opinion on this? Just resolving would also be fine.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
+    {
+        return cxx::error<Error>(Error::SERVICE_DESCRIPTION_INVALID);
+    }
+
+    // Forbid duplicate service descriptions entries
+    for (auto& element : m_serviceDescriptionVector)
+    {
+        if (element == serviceDescription)
+        {
+            return cxx::error<Error>(Error::SERVICE_DESCRIPTION_ALREADY_ADDED);
```
elBoberido, line 38: I'm not sure if this really should count as error. We have multi publisher support, therefore it is natural that the same service description might be added to the service registry

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
+    uint64_t index = 0U;
+    for (auto iterator = m_serviceDescriptionVector.begin(); iterator != m_serviceDescriptionVector.end();)
+    {
+        if (m_serviceDescriptionVector[index] == serviceDescription)
+        {
+            m_serviceDescriptionVector.erase(iterator);
+            removedElement = true;
+            // There can be not more than one element
+            break;
+        }
```
elBoberido, line 68: Let's assume we have multiple publisher and only one of them does a `STOP_OFFER` then the service gets removed from the service registry, even if there still are publishing publishers?

elBoberido (commented):
tests still not reviewed

iceoryx_posh/source/roudi/service_registry.cpp
elBoberido: any input on this? Resolving would be fine since the duplicated code is minimal, but you might revise your decision with four occurrences ;)

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
+cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceDescription& serviceDescription) noexcept
 {
-    cxx::set::add(m_serviceMap[service].instanceSet, instance);
+    if (!serviceDescription.isValid())
+    {
+        return cxx::error<Error>(Error::SERVICE_DESCRIPTION_INVALID);
+    }
+
+    // Forbid duplicate service descriptions entries
+    for (auto& element : m_serviceDescriptionVector)
```
MatthiasKillat, line 34: We can avoid this linear search later, could be worth adding a todo. We already have search structures in place that allow us to find quickly whether the service already exists, that is the major point of this data structure itself. If so, we return but not with an error (see below).

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: No, it is not an error. But it also should not be stored multiple times but I think you mean just that.

We hence should return success.

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
MatthiasKillat: I prefer it as you proposed, makes te code more expressive. The compiler will almost certainly optimize the extra variable away, so the generated code would be the same.

The structures may also not be maps later, hence the name will change. Will be done in a follow up once the internal search structure is ready.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: We may need multiset like tracking, but this should not be done in the prefixtree nor a map. We can have a count in a struct that we store in a container as the m_serviceDescriptionVector now (this is the wrong container for the job in the long run anyway, we should use the `ObjectPool`). The removing would decrease the associated count and only if it drops to 0 it is actually removed.

Storing actual copies multiple times would surely be the wrong approach though.
A vector is bad here, all the following elements will be moved upon removal. This will also completely lead to wrong results with the indexing, since it invalidates the indices stored in the maps/tries later.

This is a major problem in the current algorithm. Deleting anything from `m_serviceDescriptionVector` should not change the positions of other elements currently in the vector. It only should be logically removed (i.e. invalidated), not erased. The `ObjectPool` is made to deal with something like this. It works, but could be optimized and the code made a little nicer (in another PR).

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
+            removedElement = true;
+            // There can be not more than one element
+            break;
+        }
+        index++;
+        iterator++;
+    }
+
+    auto removeIndexFromMap = [](std::multimap<capro::IdString_t, uint64_t>& map, uint64_t index) {
+        for (auto it = map.begin(); it != map.end();)
```
MatthiasKillat, line 74: Those kind of iteration defeats the purpose of having efficient search  structures and is not necessary here.
If we want to remove a `Servicedescription` with `service` and `instance` we do the following (I prefer pseudo code to explain, it is easy to write this in C++).
1. find indices corresponding to service in `m_serviceMap, call those I1
2. find indices corresponding to instance in `m_instanceMap`, call those I2
3. compute the intersection I = I1 \intersection I2. This is one (sevice exists) or no element
4. If I is empty, the service does not exist and we have nothing to remove and can return.
5. Otherwise we remove the single element from the container and the maps.

Removing from the maps is rather tricky to do efficiently right now, but the prefix tree will support an operation that allows to remove a specific value to be removed from a specific key.

This is needed since e.g. for the `m_serviceMap` we already know the key, service. And after the intersection, we also know the value, the only value in the intsersection I. `std::map` does not efficiently support this though, so for the time being we could iterate over the map.

But for the container (currently the `m_serviceDescriptionVector`) the intersection approach must be used for efficiency. Otherwise the operations all degenerate into linear searches.

Note that this approach naturally generalizes to an arbitrary number of search attributes.

Also note that this is partly the algorithm for finding as well, at the internal level where we still operate with indices in our container (which the client will of course not see). So this can be suitably factored out. After all, removing is finding the indices to be removed and removing corresponding data. Finding for the client is just finding the indices and translating them to the actual data to be returned.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
+            possibleServices.push_back(entry->second);
+        }
+
+        auto rangeInstanceMap = m_instanceMap.equal_range(instance);
+        for (auto entry = rangeInstanceMap.first; entry != rangeInstanceMap.second; ++entry)
+        {
+            possibleInstances.push_back(entry->second);
+        }
+
+        ::std::set_intersection(possibleServices.begin(),
```
MatthiasKillat, line 121: Note that they need to be sorted before. They should be, with the ranges coming from an ordered map.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
+                                possibleServices.end(),
+                                possibleInstances.begin(),
+                                possibleInstances.end(),
+                                ::std::back_inserter(intersection));
+
```
MatthiasKillat, line 126: The code up to here can be reused in removal as well if it is suitably factored out. It can locate the index to be removed, if it exists. This is sketched in the comment regarding removal.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,149 @@
...
         {
-            instances.push_back(*iter);
+            searchResult.push_back(m_serviceDescriptionVector[entry->second]);
         }
     }
+    else
+    {
+        // Find (*, *)
+        // O(1)
+        searchResult = m_serviceDescriptionVector;
```
MatthiasKillat, line 156: In general the question arises whether we do always want to pass the result by copy, especially in the all wildcard case. Those will be many matches and they are not fast to copy if the result is of the order 1000.

MatthiasKillat (changes requested):
I believe removal should be implemented in a way that utilizes the search structures, not linear brute force. I also think that using the `m_serviceDescriptionVector` is not the correct approach and the algorithm may even fail if the following happens.

We have stored an index to some service description in one of the maps, say it is 37 for description SD1. Now we remove an unrelated service description stored at index 21. By doing so all elements in the vector will be moved forward since we erase element at position 21. Hence SD1 will no be at position 36. But the maps still assume that SD1 is at 37, causing the lookups and removal operations to be compromised. 

We can postpone the inefficient linear time removal, this is just optimization which I can do once I have the search structure ready. But the invalidation of indices by removal from the vector is a problem which will lead errors if I am not mistaken.

MatthiasKillat:
It appears my comment summarizing the review got lost somehow. Long story short, the inefficient removal by linear search can be changed later. 

But erasing elements from  `m_serviceDescriptionVector` will invalidate indices currently in the maps and cause subsequent lookups and removals to behave incorrectly.

Just consider what happens if we remove a ServiceDescription currently at index 21 and there are some ServiceDescriptions following this one, e.g. one a index 37. The latter will be moved to index 36. This is inefficient in the first place, but also the maps will be unaware of this (and they should). We cannot use a vector here like this. 

Specifically we can use a vector, but must not erase the objects, just invalidate them. In essence the `ObjectPool` does just that and some convenience here and there. The interface of the `ObjectPool` could be nicer (I wrote it hastily in the distant past...) and some optimizations are possible, but otherwise it would work here and some structure with similar properties is the way to go here.

iceoryx_posh/source/roudi/service_registry.cpp
elBoberido: Right, I didn't meant to store it multiple times. Maye we need an additional counter but that might make it a little brittle. How does the introspection do it? My guess would be we need the same mechanism.

iceoryx_posh/source/roudi/service_registry.cpp
elBoberido: Right, we need a container with a ref-counting like `ServiceDescription`. Could be an array of `cxx::optional<{ServiceDescripton, Counter}>` and a FIFO with the free indices. Similar to what you do in the PrefixTree. As intermediate step we could have an array of `cxx::optional<ServiceDesctiption>` and one with `int16_t` for the counter. Finding a free slot would be quite fast since the array is small and packed on adjacent cache lines ... we could also use the object pool, though.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
elBoberido: I like the approach with the lambda. That's quite modern and gives the user maximum freedom

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -600,17 +593,16 @@ runtime::IpcMessage PortManager::findService(const capro::IdString_t& service,
...
 
-    // add all found instances to instanceString
-    runtime::IpcMessage instanceMessage;
+    runtime::IpcMessage response;
 
-    ServiceRegistry::InstanceSet_t instances;
-    m_serviceRegistry.find(instances, service, instance);
-    for (auto& instance : instances)
+    ServiceRegistry::ServiceDescriptionVector_t searchResult;
+    m_serviceRegistry.find(searchResult, service, instance);
```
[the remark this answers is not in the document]
elBoberido: Or even a lambda like suggested by you in a previous comment. This gives the user the maximum flexibility and she doesn't even have to store the service description but work directly with the ones found like
```cpp
find("Foo", Wildcard, [] (auto& serviceDescription) { std::cout << serviceDescription << std::endl; });
```

CHANGELOG.md
mossmaurice: @MatthiasKillat Indeed, adapt the example accordingly. Our coding guidelines are still a bit of a mishmash. Do you maybe want to raise this topic in the next developer meetup?

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
mossmaurice: I agree, if we can simplify the logic we should do it. Refactored things to be able to store an invalid `ServiceDescription`. However, there are still `isValid()` checks on user-facing API level.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: Let's resolve this topic to #829 and #830 once `cxx::set` and `cxx::map` are available.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: AFAIK the introspection was not developed with n:m communication in mind and has the same issue. @FerdinandSpitzschnueffler had a look some time ago, maybe she can enlighten us.

Changed the return to `cxx::success`. My feeling is we'll need a counter in the further course of #415.

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: Ok, got your point.

Unfortunately with the `cxx::vector` that is not possible as `erase()` expects an iterator currently. So I have to iterate over the complete vector and can't use two `equal_range` searches with intersection which I used in `ServiceRegistry::find()`. Once we have an ready-made alternative like the ObjectPool I'm happy to refactor things, but for the very first version I'd like to stick with the `cxx::vector`. Let's optimise this when integrating #829 into the `ServiceRegistry`. I added a todo in #415.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: See [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/860/files#r682603230)

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: We'll see how things develop and measure once we have the first implementations ready. Then we can iterate over it again. My personal mantra: Make it work, make it fast, make it beautiful.

iceoryx_posh/source/roudi/port_manager.cpp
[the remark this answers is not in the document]
mossmaurice: As `ServiceRegistry::find()` is abstracted away from the user and will not be used directly. I would not implement this now. If this turns out useful we can implement it later on. For my eyes the order doesn't matter too much. No hard feelings on this particular issue, though.

mossmaurice:
> However, erasing elements from `m_serviceDescriptionVector` will invalidate indices currently in the maps and cause subsequent lookups and removals to behave incorrectly.

Please have a look at the test case `MultipleServiceDescriptionAddedInNonLinearOrderFindsCorrectServices`. With this test case the correct removal behaviour is verified. After the element is removed from the vector, the index is updated in the two maps. I agree that is not the most efficient way and I'm happy to refactor it once the `ObjectPool` is cleaned up and ready to use in a follow-up PR.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: Good catch @elBoberido! I added a plain and simple solution using `cxx::pair` holding the `ServiceDescription` and a `ReferenceCounter`. We can beautify it once we have decided on the final data structure.

CHANGELOG.md
```diff
@@ -108,6 +108,20 @@ ServiceDescription("First", "Second", "DontCare") myServiceDescription2;
 ServiceDescription("Foo", "Bar", "Baz") myServiceDescription3;
 ```
 
+The service-related methods have been move from `PoshRuntime` to a separate class (TBD):
```
elfenpiff, line 111: We should add a checkbox to our PR template: `Check if public API changes are added to CHANGELOG.md` for the reviewer.

Later this can be very helpful. Thanks for adding it directly.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -32,23 +35,44 @@ static const capro::IdString_t Wildcard{"*"};
...
+    /// @return true, if service description was removed, false otherwise
+    bool remove(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    /// @brief Removes given service description from registry
+    /// @param[in] searchResult, reference to the vector which will be filled with the results
+    /// @param[in] service, string or wildcard to search for
+    /// @param[in] service, string or wildcard to search for
+    void find(ServiceDescriptionVector_t& searchResult,
+              const capro::IdString_t& service = Wildcard,
+              const capro::IdString_t& instinstanceance = Wildcard) const noexcept;
```
elfenpiff, line 65: Is after this PR another PR pending? I am asking since we are providing here only service and instance and event is missing.
If the service registry is aware of the full service description we cannot use this feature for the search. This makes the PR for me incomplete - except when there is another PR pending, then forget my gibberish.

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
```diff
@@ -47,12 +47,12 @@ class NodeData;
 enum class FindServiceError
 {
     INVALID_STATE,
-    UNABLE_TO_WRITE_TO_ROUDI_CHANNEL,
-    INSTANCE_CONTAINER_OVERFLOW
+    UNABLE_TO_WRITE_TO_ROUDI_CHANNEL, /// @todo #415 remove as IPC channel won't be used
```
elfenpiff, line 50: Can we remove `FindServiceError` directly in here? I am asking since you already created the container with the size and with that the todo is already done?!

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -32,23 +35,44 @@ static const capro::IdString_t Wildcard{"*"};
...
-        InstanceSet_t instanceSet;
+        INVALID_STATE,
+        SERVICE_REGISTRY_FULL,
     };
-    using serviceMap_t = std::map<capro::IdString_t, instance_t>;
+    /// @todo #415 should be connected with iox::MAX_NUMBER_OF_SERVICES
+    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 100U;
+    using ReferenceCounter_t = uint64_t;
+    using ServiceDescriptionVector_t =
+        cxx::vector<cxx::pair<capro::ServiceDescription, ReferenceCounter_t>, MAX_SERVICE_DESCRIPTIONS>;
```
elfenpiff, line 47: The `cxx::pair` produces hard to read code. While I was reviewing I constantly had to look up what `.first` and `.second` meant. Instead of using a `cxx::pair` could you create a small struct like:
```cpp
struct ServiceDescriptionEntry {
  capro::ServiceDescription serviceDescription{};
  ReferenceCounter_t referenceCounter = 0U;
};
```

Then the code would be much more readable.

iceoryx_posh/source/runtime/posh_runtime_impl.cpp
```diff
@@ -410,26 +411,26 @@ PoshRuntimeImpl::findService(const cxx::variant<Any_t, capro::IdString_t> servic
...
     uint32_t numberOfElements = requestResponse.getNumberOfElements();
-    uint32_t capacity = static_cast<uint32_t>(instanceContainer.capacity());
+    uint32_t capacity = static_cast<uint32_t>(serviceContainer.capacity());
 
-    // Limit the instances (max value is the capacity of instanceContainer)
-    uint32_t numberOfInstances = ((numberOfElements > capacity) ? capacity : numberOfElements);
-    for (uint32_t i = 0; i < numberOfInstances; ++i)
+    // Limit the services (max value is the capacity of serviceContainer)
+    uint32_t numberOfServices = algorithm::min(capacity, numberOfElements);
+    for (uint32_t i = 0; i < numberOfServices; ++i)
```
elfenpiff, line 420: Please add `U`.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
-    cxx::set::remove(m_serviceMap[service].instanceSet, instance);
+    bool removedElement{false};
+    uint64_t index = 0U;
+    for (auto iterator = m_serviceDescriptionVector.begin(); iterator != m_serviceDescriptionVector.end();)
+    {
+        auto& element = m_serviceDescriptionVector[index].first;
+        if (element == serviceDescription)
+        {
+            auto& refCounter = m_serviceDescriptionVector[index].second;
+            refCounter--;
```
elfenpiff, line 60: Please write `--refCounter;`

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+                removedElement = true;
+            }
+            else
+            {
+                return true;
+            }
+            // There can't be more than one element
+            break;
+        }
+        index++;
```
elfenpiff, line 73: Please write `++index;`

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+        iterator++;
```
elfenpiff, line 74: Please write `++iterator;`

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+            else if (it->second > index)
+            {
+                // update index due to removed element
+                it->second--;
+            }
+            it++;
+        }
+    };
+
+    removeIndexFromMap(m_serviceMap, index);
```
elfenpiff, line 94: The lambdas should only be called when there was actually an element removed.
Could you please add `if ( removedElement ) { ...`

Otherwise we are iterating twice through some maps without doing anything.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+void ServiceRegistry::find(ServiceDescriptionVector_t& searchResult,
                            const capro::IdString_t& service,
-                           const capro::IdString_t& instance) const
+                           const capro::IdString_t& instance) const noexcept
 {
-    if (instance == Wildcard)
+    cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> intersection;
+
+    // Find (K1, K2)
+    // O(log n + log n + max(#PossibleServices + #possiblesInstances) + #intersection)
```
elfenpiff, line 107: You can write it much simpler, it is just: `O(log(n))` 

Here is why:
```
log n + log n = log ( n * n ) = log ( n ^ 2 )
log ( n ^ 2 ) = 2 * log n
```

See: https://en.wikipedia.org/wiki/List_of_logarithmic_identities

And the big-O-notation: https://en.wikipedia.org/wiki/Big_O_notation states
that `O(log(n))` that the `Find(K1, K2)` function has an upper limit which is `log(n)` with some constant.
`Find(K1, K2) = O(log(n)) => Find(K1, K2) <= M * log(n)` with `M > 0`. So the `M` constant contains the factor 2 and all the constants with `max(#PossibleServices + #possibleInstances) + #intersection)`.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+
+        for (auto& value : intersection)
         {
-            instances.push_back(instance);
+            searchResult.push_back(m_serviceDescriptionVector[value]);
         }
     }
-    else
+    // Find (*, K2)
+    // O(log n + #result)
```
elfenpiff, line 137: It is just `O(log n)`

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+    else if (service == Wildcard && instance != Wildcard)
+    {
+        auto range = m_instanceMap.equal_range(instance);
+        for (auto entry = range.first; entry != range.second; ++entry)
+        {
+            searchResult.push_back(m_serviceDescriptionVector[entry->second]);
+        }
+    }
+    // Find (K1, *)
+    // O(log n + #result)
```
elfenpiff, line 147: It is just `O(log n)`

iceoryx_posh/source/roudi/service_registry.cpp
elfenpiff, line 107: You are stating here the runtime without arguing how you come up with this result. I would remove all runtime comments and would put them into the design document.

MatthiasKillat: The estimates depend on implementation (or at least  high level algorithm). If not sufficient implementation details (or the algorithm) are present in the design argument it is also hard to argue there. For these kind of algorithms it is usually easy to see by inspecting loops and knowing bounds for containers like map and so on, so I do not miss a proof here.

I think the comment can stay, it is still useful as it reminds us that efficiency was considered.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: Nope, #result may depend on n superlogarithmically and then it would be wrong. Leave as is.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: Nope, #result may depend on n superlogarithmically and then it would be wrong. Leave as is.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+        for (auto entry = range.first; entry != range.second; ++entry)
         {
-            instances.push_back(*iter);
+            searchResult.push_back(m_serviceDescriptionVector[entry->second]);
         }
     }
+    else
+    {
+        // Find (*, *)
+        // O(1)
```
MatthiasKillat, line 159: No, it is `O(n)` if we copy the whole vector since `#result = n` here. We measure the effort as function of `n`, the number of elements which it certainly depends on.

This is why passing only a reference to the underlying container (if possible) would gain a lot of efficiency, since then it would be `O(1)`. I suppose the operation is only used rarely and we can think about this later.

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -120,11 +120,6 @@ ServiceDescription::ServiceDescription(const IdString_t& service,
 
 bool ServiceDescription::operator==(const ServiceDescription& rhs) const
 {
-    if (!isValid() || !rhs.isValid())
```
MatthiasKillat, line 123: What happens now if any of those is invalid? I may have missed it, maybe we do not even distinguish invalid from valid descriptions. Ideally there should be no invalid descriptions that can be created by the user.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: We hence can simplify this to `O( log n + #resultset)`, but not `O(log n)`.

CHANGELOG.md
mossmaurice: This is already in the current PR template. However, it was added after this PR was created. Added it manually.

CHANGELOG.md
```diff
@@ -108,6 +108,20 @@ ServiceDescription("First", "Second", "DontCare") myServiceDescription2;
...
+
+```cpp
+// before
+poshRuntime.offerService(myServiceDescription);
+poshRuntime.stopOfferService(myServiceDescription);
+poshRuntime.findService({"ServiceA", iox::capro::AnyInstanceString});
+
+// after
+discoveryInfo.offerService(myServiceDescription);
```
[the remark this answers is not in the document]
mossmaurice: I still did not come up with any better name nor did anyone else. Happy about other ideas :-)

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
mossmaurice: Completely agree! Done.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
mossmaurice: @MatthiasKillat is right, this API is modeled after ara::com from AUTOSAR. Agree, this can be added later but currently out of scope for #415 see [DDD](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/design/service-discovery.md)

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
mossmaurice: Will be done in a follow-up PR, already a todo in #415

iceoryx_posh/include/iceoryx_posh/runtime/posh_runtime.hpp
```diff
@@ -47,12 +47,12 @@ class NodeData;
...
+    INSTANCE_CONTAINER_OVERFLOW       /// @todo #415 set container to iox::MAX_NUMBER_OF_SERVICES and remove error
 };
 
-/// @brief Used to search for any string (wildcard)
-struct Any_t
+/// @brief Used to search for any string
+struct Wildcard_t
```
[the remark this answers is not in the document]
mossmaurice: As I mentioned it before last week, please put this topic onto the next iceoryx developer meetup. That would be the right place to discuss things like coding rules.

iceoryx_posh/source/capro/service_description.cpp
mossmaurice: This was changed due to the fact that the `ServiceRegistry` can now store invalid `ServiceDescriptions` and we will check for validity on a higher-level closer to the user API. The changes allows the `ServiceRegistry` to match and remove two invalid `ServiceDescription`s. The topic on how to prevent the user from creating invalid `ServiceDescription` was heavily discussed in #854 #852 and #853 so far we did not come up with a nice user-friendly solution. It's not super user-unfriendly if we stick with the current approach IMHO.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -268,21 +266,19 @@ void PortManager::handleInterfaces() noexcept
             }
         }
         // also forward services from service registry
-        auto serviceMap = m_serviceRegistry.getServiceMap();
+        /// @todo #415 do we still need this? yes but return a copy here to be stored in shared memory via new
```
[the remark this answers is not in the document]
mossmaurice: Sure, we can discuss that. Might be out of scope for iceoryx v2.0

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -600,17 +593,16 @@ runtime::IpcMessage PortManager::findService(const capro::IdString_t& service,
...
     }
 
-    // add all found instances to instanceString
-    runtime::IpcMessage instanceMessage;
+    runtime::IpcMessage response;
 
-    ServiceRegistry::InstanceSet_t instances;
-    m_serviceRegistry.find(instances, service, instance);
-    for (auto& instance : instances)
+    ServiceRegistry::ServiceDescriptionVector_t searchResult;
```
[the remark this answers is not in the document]
mossmaurice: Let's move the coding rule discussion to the developer meetup. Everything is mixed currently.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -600,17 +593,16 @@ runtime::IpcMessage PortManager::findService(const capro::IdString_t& service,
...
+    m_serviceRegistry.find(searchResult, service, instance);
+    for (auto& service : searchResult)
     {
-        instanceMessage << instance;
+        response << static_cast<cxx::Serialization>(service.first).toString();
```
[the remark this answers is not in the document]
mossmaurice: Good point, something I'll keep in mind.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,45 +17,153 @@
...
+cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceDescription& serviceDescription) noexcept
 {
-    cxx::set::add(m_serviceMap[service].instanceSet, instance);
+    // Forbid duplicate service descriptions entries
+    for (auto& element : m_serviceDescriptionVector)
+    {
+        if (element.first == serviceDescription)
+        {
+            // Due to n:m communication we don't store twice but increase the reference counter
+            element.second++;
```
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: Changed it for all three mentioned occurrences to be consistent.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: For now I'd vote for keeping the comments here, as the design document does so far not consider any implementation details.

MatthiasKillat (approved):

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
elBoberido: Why is an empty `ServiceDescription` invalid? If there are no technical reasons, we should not artificially introduce and invalid `ServiceDescription`. After all it was not necessary until now. Please remove the `isValid` method.

iceoryx_posh/source/roudi/service_registry.cpp
elBoberido: any input on this?

iceoryx_posh/source/roudi/port_manager.cpp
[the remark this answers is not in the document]
elBoberido: This is not about user facing API but about performance. With the lambda approach we do not only save a copy but do not need to iterate over a vector with data larger data than a cache line, which will further impact the performance.

elBoberido (commented):
Sorry, tests still not reviewed.

There are also quite a few issues which are deferred to a follow up PR. Please document these like the the other `todo #issueNumber` else it is quite hard to keep track of this

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
elfenpiff: @mossmaurice I disagree. The reason is, what can a user do with this kind of information: One removes a service and after the call it is not there - this is the contract. If you return false I would assume that you tried to remove the service and failed therefore it is still present.

In my opinion it does not make sense to return here a bool. Do you have an argument for that? And if so, why not return an enum so that it is more readable and one really knows what is going on!
Please also consider that we should try to reduce unnecessary code bloat and error handling. If we do return here something we are forced by our standards to handle it and we turn this line
```cpp
serviceRegistry.remove(myService);
```
into
```cpp
if ( serviceRegistry.remove(myService) ) {
   //...
} else {
  //...
}
```

And as final argument: YAGNI, you aint gonna need it. I do not see the use case and if the use case arises we can still add it.

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
elfenpiff: In your new fix PR I see that the iterator is always incremented but not part of the for loop header, could you please change this: https://github.com/eclipse-iceoryx/iceoryx/pull/917

iceoryx_posh/source/roudi/service_registry.cpp
elfenpiff: @elBoberido can you list your for occurrences here please so that @mossmaurice can see them as well.

Issue #889, opened 2021-07-31
Add CI job that runs clang-tidy


r7vme:
## Brief feature description

iceoryx already has a [.clang-tidy](https://github.com/eclipse-iceoryx/iceoryx/blob/master/.clang-tidy) config, but unfortunately codebase does not pass all checks.

## Detailed information

I propose:
1. Reduce check numbers to the point when codebase passes them
1. Add CI job that runs clang-tidy
1. Incrementally add more checks and fix violations

PR #888, opened 2021-07-31
Add clang format ci job

r7vme:
## Pre-Review Checklist for the PR Author

1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Changelog updated [in the unreleased section][changelog] including API breaking changes
1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CHANGELOG.md

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

- Closes #887

PR #890, opened 2021-07-31
iox-#889 Add clang-tidy workflows

r7vme:
Build iceoryx and then run clang-tidy on all cpp files exclude tests.

1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Closes **TBD**


.github/workflows/tidy.yml
```diff
@@ -0,0 +1,35 @@
+name: Check with clang-tidy
```
r7vme, line 1: This generally should not be separate workflow, but job that runs after one of the builds. Need to discuss the right place :thinking:

tools/check_clang_tidy.sh
```diff
@@ -0,0 +1,15 @@
...
+
+fail() {
+    printf "\033[1;31merror: %s: %s\033[0m\n" ${FUNCNAME[1]} "${1:-"Unknown error"}"
+    exit 1
+}
+
+hash run-clang-tidy-10.py || fail "run-clang-tidy-10.py not found"
+cd $(git rev-parse --show-toplevel)
+[ -f build/compile_commands.json ] || fail "build/compile_commands.json not found"
+# TODO: only process files change in PR
```
r7vme, line 12: Right now iceoryx does not pass checks and overall checks running quite long on my i7 with 8 HT cores it took 5+ minutes. So there tow things to consider:


1. How to make codebase complaint?
  a. reduce number of checks and add them gradually
  b.  try to fix violations in separate PR
1. Should we run clang-tidy for changed or all files?

r7vme:
@dkroenke this PR is to spin off a discussion, see questions that I have in comments.

dkroenke:
@r7vme Thanks for the contribution.
I think we may need to consolidate the CI Jobs a bit to reduce the load on ht GitHub Runners.

A possible way is to have "Pre-Flight" Checks that run over the codebase before the actual build and test jobs are starting. With that, we can detect early code style violations and can abort the other pendings jobs if necessary.

Considering https://github.com/eclipse-iceoryx/iceoryx/pull/888 which is a quick formatting check too, we could merge these two jobs into one workflow file  (maybe the job could be called "CI Pre-Checks").

To add the dependencies between the jobs we could use something like this: https://github.com/marketplace/actions/wait-on-check or we need to modify every job like this:

```yaml
name: Build & Test

on:
  workflow_run:
    workflows: ["CI Pre-Checks""]
    types:
      - completed # checks only if the depending job has run but not if it was successful

jobs:
  iceoryx-ubuntu:
    # the 2nd condition
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    (...)
```

.github/workflows/tidy.yml
dkroenke: See https://github.com/eclipse-iceoryx/iceoryx/pull/890#issuecomment-890816831

tools/check_clang_tidy.sh
dkroenke: One thing to consider is that we need to give the developer a chance for correct linting the code before commitring & pushing to the CI, otherwise the contributors would hammer the CI.

For that we added some weeks ago the [git-hooks ](https://github.com/eclipse-iceoryx/iceoryx/tree/master/tools/git-hooks) as automatic checks while coding.  In https://github.com/eclipse-iceoryx/iceoryx/blob/master/tools/git-hooks/pre-commit you can find a short script that already do the clang-format on the touched files. We should add something similar for clang-tidy.

In the beginning I would vote for to clang-tidy only the touched files. After some time we can then add a complete scan once the codebase is migrated. We need to consider that we have no guaranteed performance on a GitHub Runner and that they are usually limited because they are for free.

elBoberido:
@elfenpiff might be interesting for you since you are the creator of the the clang-tidy file

.clang-tidy
```diff
@@ -1,4 +1,5 @@
 Checks: '-*,readability-*,-readability-named-parameter,-readability-avoid-const-params-in-decls,-readability-else-after-return,performance-*,-readability-redundant-access-specifiers,hicpp-*,-hicpp-named-parameter,-hicpp-avoid-c-arrays,cppcoreguidelines-*,-hicpp-no-array-decay,-hicpp-signed-bitwise,-hicpp-vararg,-cppcoreguidelines-avoid-c-arrays,-cppcoreguidelines-pro-bounds-constant-array-index,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-pro-type-vararg,concurrency-*,clang-analyzer-*,cert-*,bugprone-*'
+WarningsAsErrors: '-*,readability-*,-readability-named-parameter,-readability-avoid-const-params-in-decls,-readability-else-after-return,performance-*,-readability-redundant-access-specifiers,hicpp-*,-hicpp-named-parameter,-hicpp-avoid-c-arrays,cppcoreguidelines-*,-hicpp-no-array-decay,-hicpp-signed-bitwise,-hicpp-vararg,-cppcoreguidelines-avoid-c-arrays,-cppcoreguidelines-pro-bounds-constant-array-index,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-pro-type-vararg,concurrency-*,clang-analyzer-*,cert-*,bugprone-*'
```
elfenpiff, line 2: I think we have to make some exceptions here since clang-tidy detects a lot of errors inside the gtest-macros (undeclared copy/move ctors). But if everything compiles and works it would be fine by me!

tools/check_clang_tidy.sh
elfenpiff: @r7vme My approach was to first add all checks and disable checks whenever they do not make sense for us.
With this in mind we could then reduce the warnings with every PR or disable another check.

I would run clang-tidy only for the changed files and we should create an issue to fix all clang-tidy warnings.

.clang-tidy
r7vme: The only solution for that was to NOT run clang-tidy for tests.

I also tried other solutions, like using `-isystem` for gtest headers (which in theory should prevent clang-tidy), but this did not help.

tools/git-hooks/pre-commit
```diff
@@ -43,7 +48,7 @@ fi
 cpp_files=$(git diff --cached --name-only --diff-filter=ACMRT | grep -E "\.(c|cpp|inl|h|hpp)$")
 if [[ -n ${cpp_files} ]]; then
 
-    CLANG_CMD="clang-format-10"
+    CLANG_CMD="clang-format-12"
```
elfenpiff, line 51: This is very ubuntu/debian specific. The clang package which is provided by https://clang.llvm.org comes with `clang-format` only. I would recommend the following alternative.

```sh
## when it comes first clang-format-12 has precedence if it exists
if [[ ! -z $(command -v clang-format-12 2>/dev/null) ]]; then 
    CLANG_CMD=$(command -v clang-format-12)
elif [[ ! -z $(command -v clang-format 2>/dev/null) ]]; then
    CLANG_CMD=$(command -v clang-format)
else
    echo clang-format is not installed
    exit
fi
```

tools/git-hooks/pre-commit
```diff
@@ -43,7 +48,7 @@ fi
...
     if ! command -v $CLANG_CMD &> /dev/null
     then
         CLANG_MAJOR_VERSION=$(clang-format --version | sed -n "s/clang-format version \([0-9]*\)\.[0-9]*\.[0-9]*/\1/p")
```
elfenpiff, line 54: Here you should use `CLANG_CMD` instead of `clang-format`.

Could you maybe rename `CLANG_CMD` into `CLANG_FORMAT_CMD` and `CLANG_MAJOR_VERSION` into `CLANG_FORMAT_MAJOR_VERSION`.

tools/git-hooks/pre-commit
```diff
@@ -43,7 +48,7 @@ fi
 cpp_files=$(git diff --cached --name-only --diff-filter=ACMRT | grep -E "\.(c|cpp|inl|h|hpp)$")
 if [[ -n ${cpp_files} ]]; then
 
-    CLANG_CMD="clang-format-10"
+    CLANG_CMD="clang-format-12"
     if ! command -v $CLANG_CMD &> /dev/null
```
elfenpiff, line 52: With the trick above we can remove this if statement.

tools/scripts/check_formatting.sh
```diff
@@ -27,7 +27,7 @@ fail() {
 }
 
 hash git || fail "git not found"
-hash clang-format-10 || fail "clang-format-10 not found"
+hash clang-format-12 || fail "clang-format-12 not found"
```
elfenpiff, line 30: This is again very ubuntu/debian specific. Here we could do the same trick like in the script above.

```sh
## when it comes first clang-format-12 has precedence if it exists
if [[ ! -z $(command -v clang-format-12 2>/dev/null) ]]; then 
    CLANG_FORMAT_CMD=$(command -v clang-format-12)
elif [[ ! -z $(command -v clang-format 2>/dev/null) ]]; then
    CLANG_FORMAT_CMD=$(command -v clang-format)
else
    echo clang-format is not installed
    exit
fi
```

To ensure the correct version we can do again a sed command like in the script above

```
CLANG_FORMAT_MAJOR_VERSION=$(${CLANG_FORMAT} --version | sed -n "s/clang-format version \([0-9]*\)\.[0-9]*\.[0-9]*/\1/p")

if [[ $CLANG_FORMAT_MAJOR_VERSION -lt 12 ]]; then
    echo clang-format major version must be 12 or greater
    exit
fi
```

tools/scripts/clang_tidy_check.sh
```diff
@@ -0,0 +1,71 @@
...
+    printf "\033[1;31merror: %s: %s\033[0m\n" ${FUNCNAME[1]} "${1:-"Unknown error"}"
+    exit 1
+}
+
+WORKSPACE=$(git rev-parse --show-toplevel)
+cd "${WORKSPACE}"
+
+if ! [[ -f build/compile_commands.json ]]; then # or we just exit here with an error
+    sudo apt-get update && sudo apt-get install -y libacl1-dev libncurses5-dev bison clang-tidy-12
+    export CXX=clang++-12
```
elfenpiff, line 41: Again ubuntu/specific. Could we please use again the trick drafted above to find the correct command and version. And select either `clang++-12` or `clang++` if it exists. Same for `clang-12`.

tools/scripts/clang_tidy_check.sh
```diff
@@ -0,0 +1,71 @@
...
+fail() {
+    printf "\033[1;31merror: %s: %s\033[0m\n" ${FUNCNAME[1]} "${1:-"Unknown error"}"
+    exit 1
+}
+
+WORKSPACE=$(git rev-parse --show-toplevel)
+cd "${WORKSPACE}"
+
+if ! [[ -f build/compile_commands.json ]]; then # or we just exit here with an error
+    sudo apt-get update && sudo apt-get install -y libacl1-dev libncurses5-dev bison clang-tidy-12
```
elfenpiff, line 40: I would remove this completely or we turn this script into a CI script which would not have the requirement to be distribution agnostic.

elfenpiff (changes requested):

elfenpiff (changes requested):
git hooks must be distribution agnostic implemented. This means `clang-format-12` or `clang-tidy-12` (which is not the standard binary name in most distributions and in the official source code of llvm) should be `clang-format` and `clang-tidy` or we introduce a short function which looks up if `-12` version is available uses it and otherwise it uses the default name in most distributions.

.github/workflows/lint_master.yml
```diff
@@ -1,4 +1,4 @@
-name: Axivion
```
mossmaurice, line 1: What will happen to the Axivion job?

.github/workflows/lint_pull_request.yml
```diff
@@ -0,0 +1,32 @@
...
+          echo $(clang --version)
+
+      - name: Install iceoryx dependencies and clang-tidy
+        # Softwares installed in ubuntu-20.04 instance
+        # https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu2004-README.md
+        run: sudo apt-get update && sudo apt-get install -y libacl1-dev libncurses5-dev
+
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: ZedThree/clang-tidy-review@7c3cca0396a179f2fb014dc55d8c6b97d640673c
```
mossmaurice, line 24: I'd appreciate a comment on why this custom version was used.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/.clang-tidy
```diff
@@ -0,0 +1,17 @@
...
+
+CheckOptions:
+  - { key: readability-identifier-naming.ClassCase,               value: lower_case }
+  - { key: readability-identifier-naming.EnumCase,                value: lower_case }
+  - { key: readability-identifier-naming.StructCase,              value: lower_case }
+  - { key: readability-identifier-naming.UnionCase,               value: lower_case }
+  - { key: readability-identifier-naming.MethodCase,              value: lower_case }
+  - { key: readability-identifier-naming.FunctionCase,            value: lower_case }
+  - { key: readability-identifier-naming.NamespaceCase,           value: lower_case }
+  - { key: readability-identifier-naming.PrivateMemberPrefix,     value: m_ }
```
mossmaurice, line 11: Is this also valid for `struct`'s? We have excluded `struct`s from this rule in the `CONTRIBUTING.md`.

mossmaurice (approved):
LGTM in general, left a few questions.

.github/workflows/lint_master.yml
dkroenke: It is here just a renaming with a more generic name

iceoryx_hoofs/include/iceoryx_hoofs/cxx/.clang-tidy
dkroenke: Yes that applies to structs too, afaik Clang-tidy is not able to differentiate between classes or structs when enforcing this rule. For easier handling I would propose to remove this exclusion from the Contributing.md

@elBoberido since this rule came from you, do you agree on that?

elBoberido: Sure, but the clang-tidy rule is for private member and should therefore also not affect public member not having a `m_` prefix

elBoberido (commented):
Looks good. Just one minor inconsistency regarding the `clang-format`/`clang-tidy` version

elBoberido (approved):

mossmaurice (approved):
LGTM, thanks for finishing this off!

PR #894, opened 2021-08-02
Iox #27 server port implementation preparatory

### LINKED ISSUE #27: Request/Response communication with iceoryx

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This is a preparatory PR for the server port implementation in order to keep that PR small.

The ChunkSender, ChunkDistributor and RpcHeader are extended with the functionality needed by the ServerPort. Additionally, the ServerPortOptions are added.

## References

- Partly closes #27


iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.hpp
```diff
@@ -97,19 +97,33 @@ class ChunkDistributor
...
+    /// @brief Deliver the provided shared chunk to the chunk queue with the provided ID. The chunk will NOT be added
+    /// to the chunk history
+    /// @param[in] uniqueQueueId is an unique ID which identifies the queue to which this chunk shall be delivered
+    /// @param[in] lastKnownQueueIndex is used for a fast lookup of the queue with uniqueQueueId; if the queue is not
+    /// found at the index, the queue is searched by iteration over all stored queues and the new index is stored in
+    /// this parameter
+    /// @param[in] chunk is the SharedChunk to be delivered
+    /// @return ChunkDistributorError if the queue was not found
+    cxx::expected<ChunkDistributorError>
+    deliverToQueue(const UniqueQueueId uniqueQueueId, uint32_t& lastKnownQueueIndex, mepoo::SharedChunk chunk) noexcept;
```
elBoberido, line 122: @elfenpiff I did not yet implement this since I want some feedback before I do it and also write the tests. I'm not quite sure about this since the optimization with `lastKnownQueueIndex` must be a reference which actually points to a location in shared chunk. You can see that in `ClientPort`, which has this knowledge, but further down the abstraction it is unknown that the chunk is a `ResponseHeader`. One alternative would be to request the queue index for a unique queue id separately and store that one in the `ResponseHeader`. The `deliverToQueue` method would still need to do sanity checks on the index and potentially iterate over all stored queues but that would be rare and therefore have virtually zero performance impact. What do you think?

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -48,6 +51,7 @@ using PublisherPortUserType = iox::popo::PublisherPortUser;
 using SubscriberPortRouDiType = iox::popo::SubscriberPortRouDi;
 using SubscriberPortUserType = iox::popo::SubscriberPortUser;
 using UniquePortId = popo::TypedUniqueId<popo::BasePortData>;
+using UniqueQueueId = popo::TypedUniqueId<popo::ChunkQueueDataId_t>;
```
elBoberido, line 54: @elfenpiff at first I thought this would be a good idea, but it seems that a roudi ID needs to be set before this can be used. I'm slightly leaning towards having a 64 bit unique entity ID, as suggested by you. Do you think this should be restricted to only queues or something that could also be used for other types, i.e. to never have the same IDs even if they are used for `Foo` or `Bar`?

elBoberido: @elfenpiff I think we once talked about this, but I forgot the conclusion we came to

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.hpp
elBoberido: @elfenpiff I think I go with the route to request the queue index before sending the response

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
elfenpiff: @elBoberido a unique entity id which is independent of the type so every entity in iceoryx, port, queue, semaphore or whatever we can think of gets this unique id is the right approach in my opinion.
A uint64 provides enough room for every one so this should not be a problem and one never would like to infer the type from a given unique id so why not use one single incrementing number.

iceoryx_posh/test/moduletests/test_popo_server_options.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+TEST(ServerOptions_test, DeserializingBogusDataFails)
+{
+    const auto bogusSerialization = iox::cxx::Serialization::create("hypnotoad", "brain slug", "rock star");
+    iox::popo::ServerOptions::deserialize(bogusSerialization)
+        .and_then([&](auto&) { FAIL() << "Deserialization is expected to fail!"; })
+        .or_else([&](auto&) { SUCCEED(); });
+}
+
+using QueueFullPolicyUT = std::underlying_type_t<iox::popo::QueueFullPolicy2>;
```
FerdinandSpitzschnueffler, line 60: `QueueFullPolicyUT` doesn't seem to be used.

iceoryx_posh/test/moduletests/test_popo_server_options.cpp
```diff
@@ -0,0 +1,92 @@
...
+iox::cxx::Serialization enumSerialization(ConsumerTooSlowPolicyUT clientTooSlowPolicy)
+{
+    constexpr uint64_t REQUEST_QUEUE_CAPACITY{42U};
+    const iox::NodeName_t NODE_NAME{"harr-harr"};
+    constexpr bool OFFER_ON_CREATE{true};
+
+    return iox::cxx::Serialization::create(REQUEST_QUEUE_CAPACITY, NODE_NAME, OFFER_ON_CREATE, clientTooSlowPolicy);
+}
+
+TEST(ServerOptions_test, DeserializingValidClientTooSlowPolicyIsSuccessful)
```
FerdinandSpitzschnueffler, line 71: Isn't this already tested in `SerializationRoundTripIsSuccessful`?

FerdinandSpitzschnueffler (commented):
I've only reviewed the design document and the server options so far.

iceoryx_posh/test/moduletests/test_popo_server_options.cpp
elBoberido: You are right, C&P leftover

iceoryx_posh/test/moduletests/test_popo_server_options.cpp
elBoberido: Kind of. But that one uses serialized data which are created by `ServerOptions::serialize` and this one data which is done by a function in the test. `SerializationRoundTripIsSuccessful` would work if `serialize`/`deserialize` do a memcpy. This one checks that deserialize does the right thing, which means that `serialize` also did the right thing. Maybe it would be better to do the round trip as last test, but it's easier to do it this way to catch errors during development.

elBoberido:
Btw, I did not update the changelog because this PR will not finish the feature. I just assume that the changelog is updated once the feature is complete or a PR introduces a breaking change. Please correct me if I'm wrong.

iceoryx_posh/source/popo/server_options.cpp
```diff
@@ -0,0 +1,54 @@
...
+    using ClientTooSlowPolicyUT = std::underlying_type_t<ConsumerTooSlowPolicy>;
+
+    ServerOptions serverOptions;
+    ClientTooSlowPolicyUT clientTooSlowPolicy;
+
+    auto deserializationSuccessful = serialized.extract(
+        serverOptions.requestQueueCapacity, serverOptions.nodeName, serverOptions.offerOnCreate, clientTooSlowPolicy);
+
+    if (!deserializationSuccessful
+        || clientTooSlowPolicy > static_cast<ClientTooSlowPolicyUT>(ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA))
```
MatthiasKillat, line 44: Ok, this works since DISCARD_OLDEST_DATA is currently the max in the enum. The problem is when we extend the enum (carelessly by just appending a new value) but do forget to change the code here we do not correctly deserialize. 

We need to rely on tests catching the error for the new enum value then (but we need to add the tests with the changes and may forget to do so...). There is unfortunately no way to get the max of an enum or enum class I know of.

Some trick for enums to make it more manageable is 
```
enum class MyEnum {
          MIN=0, A=MIN, B, C, MAX = C };
```
This makes it more clear when extending with D that it is probably required to adapt MAX = D but it is also only weak protection and might not be worth it (someone may add D but forget changing MAX).

Also cf. https://here-be-braces.com/extending-cpp-enums/ , I do not know of a simple alternative, it appears there is none.

MatthiasKillat (commented):
Looks ok from what I can see. I will continue the review of the tests in a second review.

doc/design/request_response_communication.md
```diff
@@ -80,14 +80,14 @@ It must be ensured that only one server with a given `ServiceDescription` can ru
 ![rpc header](diagrams/request_response/request_response_header.svg)
 
 Since request and response need to encode different meta-information, we also need different header for the messages.
-The common data is aggregated in `RpcBaseHeader` which contains a `UniquePortId` to the port owning the `ClientChunkQueueData_t` and a sequence ID.
-The `UniquePortId` is used to identify the queue which receives the response and used as identifier for `ChunkDistributor::deliverToPort`.
+The common data is aggregated in `RpcBaseHeader` which contains a `cxx::UniqueId` to the `ClientChunkQueueData_t` and a sequence ID.
+The `cxx::UniqueId` is used to identify the queue which receives the response and used as identifier for `ChunkDistributor::deliverToQueue`.
 This method will iterate over all stored queues and matches the `ChunkQueueData::m_uniqueIdOfOwner`.
-As optimization, the `ChunkDistributor::deliverToPort` returns the index of the queue to do a fast lookup and downgrades to a full lookup if the port IDs don't match.
+As optimization, the `lastKnownQueueIndex` will be used to do a fast lookup and downgrades to a full lookup if the queue IDs don't match.
```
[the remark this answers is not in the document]
elBoberido: They are stored in a vector and therefore it will be fast.

FerdinandSpitzschnueffler (commented):
Reviewed chunk queue data and chunk distributor files.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -201,12 +201,80 @@ inline void ChunkDistributor<ChunkDistributorDataType>::deliverToAllStoredQueues
...
+        retry = false;
+        if (!pushToQueue(queue.get(), chunk))
+        {
+            if (isBlockingQueue)
+            {
+                retry = true;
+            }
+            else
+            {
+                ChunkQueuePusher_t(queue.get()).lostAChunk();
```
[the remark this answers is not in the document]
elBoberido: The `ChunkQueuePusher_t` is only necessary to set the flag for a lost chunk. Does this answer your question or did you mean something different?

This was also modeled like the `deliverToAllQueues` from line 134 onwards.

iceoryx_posh/source/popo/server_options.cpp
elBoberido: Yes, it is not ideal. On the good side, if there is something added the deserialization of the new value would fail and we would immediately detect it in tests.

iceoryx_posh/test/moduletests/test_popo_chunk_distributor.cpp
```diff
@@ -412,68 +516,108 @@ TYPED_TEST(ChunkDistributor_test, addToHistoryWithoutDelivery)
...
+    constexpr uint32_t DATA_TO_SEND{987};
+    auto chunk = this->allocateChunk(DATA_TO_SEND);
+    ASSERT_FALSE(sut.deliverToQueue(queueData->m_uniqueId, UNKNOWN_QUEUE_INDEX, chunk).has_error());
 
     ChunkQueuePopper<typename TestFixture::ChunkQueueData_t> queue(queueData.get());
     auto maybeSharedChunk = queue.tryPop();
-
     ASSERT_THAT(maybeSharedChunk.has_value(), Eq(true));
-    EXPECT_THAT(this->getSharedChunkValue(*maybeSharedChunk), Eq(451u));
+    auto receivedData = *(static_cast<uint32_t*>(maybeSharedChunk.value().getUserPayload()));
```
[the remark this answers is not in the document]
elBoberido: Is `because I'm stupid` an acceptable answer for you?

iceoryx_posh/test/moduletests/test_popo_chunk_distributor.cpp
```diff
@@ -222,6 +238,110 @@ TYPED_TEST(ChunkDistributor_test, RemoveAllQueuesWhenContainingMultipleQueues)
...
+    auto sutData = this->getChunkDistributorData();
+    typename TestFixture::ChunkDistributor_t sut(sutData.get());
+
+    auto queueData = this->getChunkQueueData();
+    ASSERT_FALSE(sut.tryAddQueue(queueData.get()).has_error());
+    ASSERT_FALSE(sut.tryRemoveQueue(queueData.get()).has_error());
+
+    sut.getQueueIndex(queueData->m_uniqueId, EXPECTED_QUEUE_INDEX)
+        .and_then([&](const auto& index) { EXPECT_THAT(index, Eq(EXPECTED_QUEUE_INDEX)); })
+        .or_else([] { FAIL() << "Expected to get an index!"; });
```
FerdinandSpitzschnueffler, line 342: I would have expected that no queue is found and a `nullopt` is returned.

elBoberido: Funny, that's also what the CI expects

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.hpp
```diff
@@ -92,9 +93,21 @@ class ChunkSender : public ChunkDistributor<typename ChunkSenderDataType::ChunkD
...
     void send(mepoo::ChunkHeader* const chunkHeader) noexcept;
 
+    /// @brief Send an allocated chunk to a specific ChunkQueuePopper
+    /// @param[in] chunkHeader, pointer to the ChunkHeader to send; the ownership of the pointer is transferred to this
+    /// method
+    /// @param[in] uniqueQueueId is an unique ID which identifies the queue to which this chunk shall be delivered
+    /// @param[in] lastKnownQueueIndex is used for a fast lookup of the queue with uniqueQueueId
+    /// @return true when successful, false otherwise
+    /// @note This method does not add the chunk to the history
+    bool sendToQueue(mepoo::ChunkHeader* const chunkHeader,
```
elfenpiff, line 107: Is it valid to have the `nullptr` as argument for `chunkHeader`? If not I would change the `chunkHeader` argument into a reference to communicate this to the user that this is never a `nullptr`.

The same argument goes also for the other methods but you did not touch the fishy, but if you touch the fishy you can turn the pointer into a reference.

https://www.youtube.com/watch?v=8gdsueMGUfk

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
```diff
@@ -147,30 +147,51 @@ template <typename ChunkSenderDataType>
...
+}
+
+template <typename ChunkSenderDataType>
+inline bool ChunkSender<ChunkSenderDataType>::sendToQueue(mepoo::ChunkHeader* const chunkHeader,
+                                                          const cxx::UniqueId uniqueQueueId,
+                                                          const uint32_t lastKnownQueueIndex) noexcept
+{
+    mepoo::SharedChunk chunk(nullptr);
+    // BEGIN of critical section, chunk will be lost if the process terminates in this section
+    if (getChunkReadyForSend(chunkHeader, chunk))
```
elfenpiff, line 168: What happens when `chunkHeader == nullptr`?

iceoryx_posh/source/popo/server_options.cpp
elfenpiff: I prefer the approach of adding an additional enum called `END` and then one can add additional enum values without touching any piece of code.

elBoberido: This does not fully solve the problem since one can add new enum values after `END` and also leaks an implementation detail to the user since they would be able to configure `ConsumerTooSlowPolicy::END`. Usually it is not a good idea to have surrogates since they can also misused.

Since this is an enum class, one also has to edit the code when new values are added if there is a switch statement without a default value.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.hpp
```diff
@@ -96,19 +97,30 @@ class ChunkDistributor
...
-    /// @param[in] chunk queue to which this chunk shall be delivered
-    /// @param[in] shared chunk to be delivered
-    /// @return false if a queue overflow occured, otherwise true
-    bool deliverToQueue(cxx::not_null<ChunkQueueData_t* const> queue, mepoo::SharedChunk chunk) noexcept;
+    /// @brief Deliver the provided shared chunk to the chunk queue with the provided ID. The chunk will NOT be added
+    /// to the chunk history
+    /// @param[in] uniqueQueueId is an unique ID which identifies the queue to which this chunk shall be delivered
+    /// @param[in] lastKnownQueueIndex is used for a fast lookup of the queue with uniqueQueueId
+    /// @param[in] chunk is the SharedChunk to be delivered
+    /// @return ChunkDistributorError if the queue was not found
```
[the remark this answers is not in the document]
elBoberido: Of course. It can also happen that the call is blocking. This all depends on the `SubscriberTooSlowPolicy` and `QueueFullPolicy`

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.hpp
elBoberido: Well, it is the same behavior as with `ChunkSender::send`. One can argue if that is the desired behavior but since that can easily becomes a bottomless pit I would argue it should be discussed/implemented in a separate issue.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
elBoberido: Then `getChunkReadyForSend` will return false and the error handler will be called

iceoryx_posh/test/moduletests/test_popo_chunk_distributor.cpp
```diff
@@ -230,6 +246,115 @@ TYPED_TEST(ChunkDistributor_test, RemoveAllQueuesWhenContainingMultipleQueues)
...
+
+    auto sutData = this->getChunkDistributorData();
+    typename TestFixture::ChunkDistributor_t sut(sutData.get());
+
+    auto queueData = this->getChunkQueueData();
+    ASSERT_FALSE(sut.tryAddQueue(queueData.get()).has_error());
+    ASSERT_FALSE(sut.tryRemoveQueue(queueData.get()).has_error());
+
+    auto maybeIndex = sut.getQueueIndex(queueData->m_uniqueId, EXPECTED_QUEUE_INDEX);
+    ;
```
FerdinandSpitzschnueffler, line 354:
```suggestion
    
```

FerdinandSpitzschnueffler (approved):

elfenpiff (approved):

Issue #832, opened 2021-06-02
Refactor IpcChannel


elfenpiff:
## Brief feature description

The current IpcChannel abstraction is just a type alias with the big downside that all ipc channels (message queue, unix domain sockets, named pipes and the upcoming qnx message passing) have to support all the features of the message queue.

We require an IpcChannel abstraction which reduces the interface to a very small subset of common functionality (send, receive).

## Detailed information
- [ ] Create IpcChannel abstraction with the following interface.
```cpp
template<typename IpcType> // could be MessageQueue, UnixDomainSocket, NamedPipe etc.
class IpcChannel {
  public:
    IpcChannel(const IpcType::connectionInformation_t & );
    bool send(const IpcMessage_t& msg);
    bool timedSend(const IpcMessage_t & msg, const units::Duration & timeout);
    cxx::optional<IpcMessage_t> receive();
    cxx::optional<IpcMessage_t> timedReceive(const units::Duration & timeout);
};
```
  The IpcChannel object should be always in a valid and usable state, if the ctor can fail use the creation pattern.
- [ ]  replace current IpcChannel alias with new IpcChannel abstraction
- [ ] refactor UnixDomainSockets so that it is a true unix domain socket abstraction.
- [ ] refactor NamedPipe so that the MessageQueue compatibility code is removed
- [ ] remove std::string from everything.

elBoberido:
The interface is not sufficient since we potentially need to do some checks when RouDi or an application starts like if an IpcChannel is already available or deleted from the file system.

Removing `std::string` could be done once we switch to a binary serialization. I already have a PoC for this, just need some time to finish it.

PR #895, opened 2021-08-02
iox-#889 Fix clang-tidy violations for iceoryx_hoofs package

r7vme:
1. Fix all `clang-tidy-10` violations in iceoryx_hoofs package.
2. Add CI job `run-clang-tidy` that runs check (w/o failing as agreed with @dkroenke) for iceoryx_hoofs

To test locally:
1. Make sure to install clang-tidy-10
2. `./tools/ci/run-clang-tidy.sh`

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Partially addresses #889


.clang-tidy
```diff
@@ -11,7 +31,6 @@ CheckOptions:
   - { key: readability-identifier-naming.ProtectedMemberPrefix,   value: m_ }
   - { key: readability-identifier-naming.MemberCase,              value: camelBack }
   - { key: readability-identifier-naming.ConstexprVariableCase,   value: UPPER_CASE }
-  - { key: readability-identifier-naming.EnumCase,                value: UPPER_CASE }
```
elfenpiff, line 14: Why did you remove this?
We actually have this coding rule, did it produce somewhere problems?

r7vme: @elfenpiff this is duplicated `readability-identifier-naming.EnumCase`, above we have 

```
{ key: readability-identifier-naming.EnumCase,                value: CamelCase }
```

Most of the enum classes I saw use camel case

r7vme: @elfenpiff or you think we need to switch enum classes to UPPER_CASE?

elBoberido: @elfenpiff @r7vme I think we use mostly `UPPER_CASE` for `EnumConstantCase`. A notable exception to this is the `Error` enum which uses a `k` prefix for its values.

r7vme: ack, will remove rule above with camel case then

elBoberido: @r7vme the rule with camel case is valid, but for `EnumCase` not `EnumConstantCase`

r7vme: done

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
```diff
@@ -92,8 +92,8 @@ FileLock& FileLock::operator=(FileLock&& rhs) noexcept
 
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_name = std::move(rhs.m_name);
-        m_fd = std::move(rhs.m_fd);
+        m_name = std::move(rhs.m_name); // NOLINT TODO(r7vme)
```
r7vme, line 95: @elfenpiff @elBoberido  All parts marked with `TODO(r7vme)` are related to usage of `CreationPattern_t::operator=(std::move(rhs));` and std::move afterwards, which means usage of moved objects (which is violation).

What `CreationPattern_t::operator=(std::move(rhs));` does? I don't understand completely this part

iceoryx_hoofs/source/posix_wrapper/message_queue.cpp
```diff
@@ -117,10 +117,10 @@ MessageQueue& MessageQueue::operator=(MessageQueue&& other)
         }
         CreationPattern_t::operator=(std::move(other));
 
-        m_name = std::move(other.m_name);
-        m_attributes = std::move(other.m_attributes);
-        m_mqDescriptor = std::move(other.m_mqDescriptor);
-        m_channelSide = std::move(other.m_channelSide);
+        m_name = std::move(other.m_name);                 // NOLINT TODO(r7vme)
```
r7vme, line 120: same here

iceoryx_hoofs/source/posix_wrapper/named_pipe.cpp
```diff
@@ -136,8 +138,8 @@ NamedPipe& NamedPipe::operator=(NamedPipe&& rhs) noexcept
         IOX_DISCARD_RESULT(destroy());
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_sharedMemory = std::move(rhs.m_sharedMemory);
-        m_data = std::move(rhs.m_data);
+        m_sharedMemory = std::move(rhs.m_sharedMemory); // NOLINT TODO(r7vme)
```
r7vme, line 141: same here

iceoryx_hoofs/source/posix_wrapper/semaphore.cpp
```diff
@@ -41,14 +39,14 @@ Semaphore& Semaphore::operator=(Semaphore&& rhs) noexcept
 
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_name = std::move(rhs.m_name);
-        m_isCreated = std::move(rhs.m_isCreated);
-        m_isNamedSemaphore = std::move(rhs.m_isNamedSemaphore);
-        m_handle = std::move(rhs.m_handle);
-        m_isShared = std::move(rhs.m_isShared);
+        m_name = std::move(rhs.m_name);                         // NOLINT
```
r7vme, line 42: same here

iceoryx_hoofs/source/posix_wrapper/semaphore.cpp
```diff
@@ -190,7 +188,7 @@ Semaphore::Semaphore(CreateUnnamedSharedMemorySemaphore_t, const unsigned int va
 Semaphore::Semaphore(OpenNamedSemaphore_t, const char* name, const int oflag) noexcept
     : m_isCreated(false)
 {
-    if (m_name.unsafe_assign(name) == false)
+    if (!m_name.unsafe_assign(name))
```
r7vme, line 191: are we fine with `!` instead of `== false` as suggested by `readability-simplify-boolean-expr`?

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/memory_map.cpp
```diff
@@ -146,8 +151,8 @@ MemoryMap& MemoryMap::operator=(MemoryMap&& rhs) noexcept
 
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_baseAddress = std::move(rhs.m_baseAddress);
-        m_length = std::move(rhs.m_length);
+        m_baseAddress = std::move(rhs.m_baseAddress); // NOLINT TODO(r7vme)
```
r7vme, line 154: same here

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/shared_memory.cpp
```diff
@@ -115,9 +116,9 @@ SharedMemory& SharedMemory::operator=(SharedMemory&& rhs) noexcept
 
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_name = rhs.m_name;
-        m_hasOwnership = std::move(rhs.m_hasOwnership);
-        m_handle = std::move(rhs.m_handle);
+        m_name = rhs.m_name;                            // NOLINT TODO(r7vme)
```
r7vme, line 119: same here

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -125,11 +125,11 @@ UnixDomainSocket& UnixDomainSocket::operator=(UnixDomainSocket&& other) noexcept
 
         CreationPattern_t::operator=(std::move(other));
 
-        m_name = std::move(other.m_name);
-        m_channelSide = std::move(other.m_channelSide);
-        m_sockfd = std::move(other.m_sockfd);
-        m_sockAddr = std::move(other.m_sockAddr);
-        m_maxMessageSize = std::move(other.m_maxMessageSize);
+        m_name = std::move(other.m_name);                     // NOLINT TODO(r7vme)
```
r7vme, line 128: same here

r7vme:
@elBoberido @elfenpiff @dkroenke please take a look this PR fixes clang-tidy violations in iceoryx_hoofs

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
elfenpiff: @r7vme 
Hey this is absolutely correct and weird. The problem is the following:
- MemoryMap inherits from designpattern::creation
- creation has two members m_isInitialized and m_errorValue which have to be moved with every move operation
- creation has an explicit move ctor and move assignment operator which invalidates the source after a move operation - they should always be used

When we now implement a resource handling class which inherits from creation it has to call first the move ctor/assignment operator before moving the members to the new source otherwise the two members m_isInitialized and m_errorValue are either undefined or we set them manually and then the origin is not invalidated
We solve this by delegating the move operation first to the parent so that the members are moved correctly.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/message_queue.hpp
```diff
@@ -111,9 +111,11 @@ class MessageQueue : public DesignPattern::Creation<MessageQueue, IpcChannelErro
 
   private:
     IpcChannelName_t m_name;
-    struct mq_attr m_attributes;
+    struct mq_attr m_attributes
```
elfenpiff, line 114: Hehehe funky ... but actually this is the wrong formatting.
I think clang-format thinks you define a struct here and therefore does a line break but actually you are defining a variable `m_attributes` of type `mq_attr`.

With c++ you do not need to write `struct` anymore and can solve this formatting issue by writing:
`mq_attr m_attributes{};` without the leading `struct`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/message_queue.hpp
```diff
@@ -111,9 +111,11 @@ class MessageQueue : public DesignPattern::Creation<MessageQueue, IpcChannelErro
...
+    {
+    };
     mqd_t m_mqDescriptor = INVALID_DESCRIPTOR;
-    IpcChannelSide m_channelSide;
+    IpcChannelSide m_channelSide{};
```
elfenpiff, line 118: Does this makes sense? This is an enum class and I have no idea what the default constructor of an enum class variable initialize it to. Here I would suggest writing explicitly:
`IpcChannelSide m_channelSide = IpcChannelSide::CLIENT;`. 

It is later set in the ctor correctly so it does not matter but if someone does not initialize it we have at least a defined value of `CLIENT` in here.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/mutex.hpp
```diff
@@ -81,14 +81,15 @@ class mutex
     /// @brief  Tries to lock the mutex object. If it is not possible to lock
     ///         the mutex object try_lock will return an error. If the c
     ///         function fails it will return false, otherwise true.
-    bool try_lock();
+    bool tryLock();
```
elfenpiff, line 84: You did the right thing but here we explicitly followed the C++ STL code guidelines and for now I would stick to them.

The reason is when a user switches from C++ STL to our heapless, exceptionsless constructs that they have to adapt as less code as possible.

For this reason I also would stick to the lower case name `mutex`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -140,9 +140,11 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     static constexpr int32_t INVALID_FD = -1;
 
     UdsName_t m_name;
-    IpcChannelSide m_channelSide;
+    IpcChannelSide m_channelSide{};
```
elfenpiff, line 143: Please initialize it with an explicit enum class value like `IpcChannelSide m_channelSide = IpcChannelSide::CLIENT;`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -140,9 +140,11 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
...
     int32_t m_sockfd{INVALID_FD};
-    struct sockaddr_un m_sockAddr;
+    struct sockaddr_un m_sockAddr
```
elfenpiff, line 145: Please write `sockaddr_un m_sockAddr{};` to solve the formatting issue.

iceoryx_hoofs/include/iceoryx_hoofs/log/logger.hpp
```diff
@@ -40,44 +40,57 @@ class Logger
...
+    // NOLINTNEXTLINE(readability-identifier-naming) todo: rename to a proper name
     LogStream LogInfo() noexcept;
+    // NOLINTNEXTLINE(readability-identifier-naming) todo: rename to a proper name
     LogStream LogDebug() noexcept;
+    // NOLINTNEXTLINE(readability-identifier-naming) todo: rename to a proper name
     LogStream LogVerbose() noexcept;
 
   protected:
-    Logger(std::string ctxId, std::string ctxDescription, LogLevel appLogLevel);
+    Logger(const std::string& ctxId, const std::string& ctxDescription, LogLevel appLogLevel);
```
elfenpiff, line 85: `LogLevel appLogLevel` should be also const.

iceoryx_hoofs/include/iceoryx_hoofs/log/logging.hpp
```diff
@@ -27,7 +27,9 @@ namespace iox
 {
 namespace log
 {
-Logger& CreateLogger(std::string ctxId, std::string ctxDescription, LogLevel appDefLogLevel = LogLevel::kWarn) noexcept;
+Logger& createLogger(const std::string& ctxId,
+                     const std::string& ctxDescription,
+                     LogLevel appDefLogLevel = LogLevel::kWarn) noexcept;
```
elfenpiff, line 32: `LogLevel appLogLevel` should be also const.

iceoryx_hoofs/include/iceoryx_hoofs/log/logmanager.hpp
```diff
@@ -37,8 +37,11 @@ enum class LogLevelOutput : uint8_t
 class LogManager
 {
   public:
+    // NOLINTNEXTLINE(readability-identifier-naming) todo: rename to a proper name
     static LogManager& GetLogManager();
-    static Logger& CreateLogContext(std::string ctxId, std::string ctxDescription, LogLevel appDefLogLevel) noexcept;
+    static Logger&
+    // NOLINTNEXTLINE(readability-identifier-naming) todo: rename to a proper name
+    CreateLogContext(const std::string& ctxId, const std::string& ctxDescription, LogLevel appDefLogLevel) noexcept;
```
elfenpiff, line 44: Add const to `LogLevel appDefLogLevel`.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/posix_access_rights.hpp
```diff
@@ -42,8 +42,8 @@ class PosixGroup
 {
   public:
     using string_t = cxx::string<100>;
-    PosixGroup(gid_t f_id);
-    PosixGroup(string_t f_name);
+    explicit PosixGroup(gid_t f_id);
```
elfenpiff, line 45: Please add const to argument.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/posix_access_rights.hpp
```diff
@@ -68,8 +68,8 @@ class PosixUser
     using groupVector_t = cxx::vector<PosixGroup, MaxNumberOfGroups>;
     using string_t = cxx::string<100>;
 
-    PosixUser(uid_t f_id);
-    PosixUser(const string_t& f_name);
+    explicit PosixUser(uid_t f_id);
```
elfenpiff, line 71: Please add const to argument.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/signal_handler.hpp
```diff
@@ -69,7 +69,9 @@ class SignalGuard
 
   private:
     Signal m_signal;
-    struct sigaction m_previousAction;
+    struct sigaction m_previousAction
```
elfenpiff, line 72: Please write `sigaction m_previousAction{};` to get rid of the formatting issue.

iceoryx_hoofs/source/log/logger.cpp
```diff
@@ -123,21 +136,25 @@ void Logger::Print(const LogEntry entry) const
     auto sec = std::chrono::duration_cast<std::chrono::seconds>(entry.time);
     std::time_t time = sec.count();
 
-    auto timeInfo = std::localtime(&time);
+    // TODO: std::localtime is thread-unsafe, may be remove
```
elfenpiff, line 139: Can you elaborate this comment in more detail. How is `std::localtime` thread-unsafe? How would affect it the following code?

iceoryx_hoofs/source/posix_wrapper/message_queue.cpp
```diff
@@ -321,7 +321,9 @@ cxx::expected<IpcChannelError> MessageQueue::timedSend(const std::string& msg, c
 
 cxx::expected<bool, IpcChannelError> MessageQueue::isOutdated()
 {
-    struct stat sb;
+    struct stat sb
```
elfenpiff, line 324: Please write `stat sb{};` And if you are inspired by a better name let the inspiration directly flow into this piece of code :D

iceoryx_hoofs/source/posix_wrapper/signal_handler.cpp
```diff
@@ -57,7 +58,9 @@ void SignalGuard::restorePreviousAction() noexcept
 
 SignalGuard registerSignalHandler(const Signal signal, const SignalHandlerCallback_t callback) noexcept
 {
-    struct sigaction action;
+    struct sigaction action
```
elfenpiff, line 61: Please write `sigaction action{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/signal_handler.cpp
```diff
@@ -68,10 +71,14 @@ SignalGuard registerSignalHandler(const Signal signal, const SignalHandlerCallba
         return SignalGuard();
     }
 
+    // system struct, no way to avoid union
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
     action.sa_handler = callback;
     action.sa_flags = 0;
 
-    struct sigaction previousAction;
+    struct sigaction previousAction
```
elfenpiff, line 79: Please write `sigaction previousAction{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -218,17 +224,22 @@ Timer::OsTimer::OsTimer(const units::Duration timeToWait, const std::function<vo
     }
 
     // Create the struct in order to configure the timer in the OS
-    struct sigevent asyncCallNotification;
+    struct sigevent asyncCallNotification
```
elfenpiff, line 227: Please write `sigevent asyncCallNotification{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -280,7 +291,9 @@ void Timer::OsTimer::executeCallback() noexcept
 cxx::expected<TimerError> Timer::OsTimer::start(const RunMode runMode, const CatchUpPolicy catchUpPolicy) noexcept
 {
     // Convert units::Duration to itimerspec
-    struct itimerspec interval;
+    struct itimerspec interval
```
elfenpiff, line 294: Please write `itimerspec interval{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -327,7 +340,9 @@ cxx::expected<TimerError> Timer::OsTimer::stop() noexcept
         return cxx::success<void>();
     }
 
-    struct itimerspec interval;
+    struct itimerspec interval
```
elfenpiff, line 343: Please write `itimerspec interval{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -383,7 +398,9 @@ cxx::expected<TimerError> Timer::OsTimer::restart(const units::Duration timeToWa
 
 cxx::expected<units::Duration, TimerError> Timer::OsTimer::timeUntilExpiration() noexcept
 {
-    struct itimerspec currentInterval;
+    struct itimerspec currentInterval
```
elfenpiff, line 401: Please write `itimerspec currentInterval{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -423,7 +440,9 @@ TimerError Timer::OsTimer::getError() const noexcept
 
 cxx::expected<units::Duration, TimerError> Timer::now() noexcept
 {
-    struct timespec value;
+    struct timespec value
```
elfenpiff, line 443: Please write `timespec value{};` to solve formatting issue.

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -272,7 +272,9 @@ cxx::expected<std::string, IpcChannelError> UnixDomainSocket::receive() const no
 {
     // we also support timedReceive. The setsockopt call sets the timeout for all further recvfrom calls, so we must set
     // it to 0 to turn the timeout off
-    struct timeval tv;
+    struct timeval tv
```
elfenpiff, line 275: Please write `timeval tv{};` to solve formatting issue.

iceoryx_hoofs/source/units/duration.cpp
```diff
@@ -47,7 +47,9 @@ struct timespec Duration::timespec(const TimeSpecReference& reference) const noe
     }
     else
     {
-        struct timespec referenceTime;
+        struct timespec referenceTime
```
elfenpiff, line 50: Please write `timespec referenceTime{};` to solve formatting issue.

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -212,13 +213,17 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
         }
         else
         {
-            uint32_t pid;
-            uid_t userId;
-            int64_t transmissionTimestamp;
+            uint32_t pid{0};
```
elfenpiff, line 216: Please write `uint32_t pid{0U};`

Sadly there is only a clang-tidy check for `0u` -> `0U` but not when the `U` is missing at all. Therefor we have to write a custom clang-tidy rule.

r7vme:
@elBoberido @elfenpiff @dkroenke @mossmaurice 

Sorry for a huge PR. If you have time please take a look. And I welcome any suggestions.

There is also [a new CI job](https://github.com/eclipse-iceoryx/iceoryx/pull/895/checks?check_run_id=3264093438) that runs clang-tidy check for iceoryx_hoofs package.

iceoryx_posh/source/roudi/roudi.cpp
mossmaurice: @elfenpiff That's a good idea, happy to write another custom clang tidy rule :) I wrote one which adds `noexcept` to all methods. First version is available [here](https://github.com/mossmaurice/llvm-project/tree/add-custom-noexcept-all-the-things-rule).

@r7vme Could you give it a try and think about we can integrate this in our iceoryx workflow? Probably we should fork the llvm repo on the ApexAI organization.

cc @jpsamper2009

dkroenke: @mossmaurice This is addressed here: https://github.com/eclipse-iceoryx/iceoryx/issues/902

iceoryx_hoofs/source/posix_wrapper/semaphore.cpp
```diff
@@ -41,14 +39,16 @@ Semaphore& Semaphore::operator=(Semaphore&& rhs) noexcept
...
-        m_isCreated = std::move(rhs.m_isCreated);
-        m_isNamedSemaphore = std::move(rhs.m_isNamedSemaphore);
-        m_handle = std::move(rhs.m_handle);
-        m_isShared = std::move(rhs.m_isShared);
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_name = std::move(rhs.m_name); // NOLINT(hicpp-invalid-access-moved)
+        // the rest are trivially-copyable, so no std::move
+        m_isCreated = rhs.m_isCreated;
+        m_isNamedSemaphore = rhs.m_isNamedSemaphore;
+        m_handle = rhs.m_handle;
```
elfenpiff, line 47: Here you have to use move since the `m_handle` type is in mac os a unique_ptr like structure which can only be moved and not copied.
This is a trade off of platform independent code.

dkroenke: done

elfenpiff (dismissed):

dkroenke:
@elBoberido Would you take a look? I cannot approve because I did changes in this PR.

iceoryx_hoofs/source/posix_wrapper/named_pipe.cpp
```diff
@@ -136,8 +139,10 @@ NamedPipe& NamedPipe::operator=(NamedPipe&& rhs) noexcept
         IOX_DISCARD_RESULT(destroy());
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_sharedMemory = std::move(rhs.m_sharedMemory);
-        m_data = std::move(rhs.m_data);
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_sharedMemory = std::move(rhs.m_sharedMemory); // NOLINT(hicpp-invalid-access-moved)
+        // the rest are trivially-copyable, so no std::move
+        m_data = rhs.m_data;
```
elBoberido, line 145: Please revert. This makes the code less robust. In case the pointer is changed to a smart pointer, e.g. a relative pointer, this assumption is not true anymore and makes refactoring harder.

iceoryx_hoofs/source/posix_wrapper/named_pipe.cpp
```diff
@@ -153,6 +158,8 @@ template <typename Prefix>
 IpcChannelName_t NamedPipe::convertName(const Prefix& p, const IpcChannelName_t& name) noexcept
 {
     return IpcChannelName_t(cxx::TruncateToCapacity,
+                            // name is checked for emptyness, so it's ok to get a first member
+                            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
```
elBoberido, line 162: This comment is wrong, the name is not checked in this method and could be empty

iceoryx_hoofs/source/posix_wrapper/semaphore.cpp
```diff
@@ -41,14 +39,16 @@ Semaphore& Semaphore::operator=(Semaphore&& rhs) noexcept
...
-        m_isNamedSemaphore = std::move(rhs.m_isNamedSemaphore);
-        m_handle = std::move(rhs.m_handle);
-        m_isShared = std::move(rhs.m_isShared);
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_name = std::move(rhs.m_name); // NOLINT(hicpp-invalid-access-moved)
+        // the rest are trivially-copyable, so no std::move
+        m_isCreated = rhs.m_isCreated;
+        m_isNamedSemaphore = rhs.m_isNamedSemaphore;
+        m_handle = std::move(rhs.m_handle); // NOLINT(hicpp-move-const-arg) because m_handle cannot be copied in macos
+        m_isShared = rhs.m_isShared;
```
elBoberido, line 48: Please revert this

iceoryx_hoofs/source/posix_wrapper/shared_memory_object/memory_map.cpp
```diff
@@ -146,8 +150,10 @@ MemoryMap& MemoryMap::operator=(MemoryMap&& rhs) noexcept
 
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_baseAddress = std::move(rhs.m_baseAddress);
-        m_length = std::move(rhs.m_length);
+        // the rest are trivially-copyable, so no std::move
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_baseAddress = rhs.m_baseAddress; // NOLINT(hicpp-invalid-access-moved)
+        m_length = rhs.m_length;
```
elBoberido, line 156: please revert

iceoryx_hoofs/source/posix_wrapper/signal_handler.cpp
```diff
@@ -30,9 +30,10 @@ SignalGuard::SignalGuard(const Signal signal, const struct sigaction& previousAc
 
 SignalGuard::SignalGuard(SignalGuard&& rhs) noexcept
 {
-    m_signal = std::move(rhs.m_signal);
-    m_previousAction = std::move(rhs.m_previousAction);
-    m_doRestorePreviousAction = std::move(rhs.m_doRestorePreviousAction);
+    // NOTE: trivially copyable members do not require std::move
+    m_signal = rhs.m_signal;
+    m_previousAction = rhs.m_previousAction;
+    m_doRestorePreviousAction = rhs.m_doRestorePreviousAction;
```
elBoberido, line 36: please revert

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -54,6 +54,7 @@ UnixDomainSocket::UnixDomainSocket(UnixDomainSocket&& other) noexcept
     *this = std::move(other);
 }
 
+// NOLINTNEXTLINE(readability-function-size) todo: make a struct out of aruments
```
elBoberido, line 57: I don't know if we have this rule but lately we usually do not just use a `todo` but also add a corresponding issue number

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -125,11 +127,13 @@ UnixDomainSocket& UnixDomainSocket::operator=(UnixDomainSocket&& other) noexcept
...
-        m_sockfd = std::move(other.m_sockfd);
-        m_sockAddr = std::move(other.m_sockAddr);
-        m_maxMessageSize = std::move(other.m_maxMessageSize);
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_name = std::move(other.m_name); // NOLINT(hicpp-invalid-access-moved)
+        // the rest are trivially-copyable, so no std::move
+        m_channelSide = other.m_channelSide;
+        m_sockfd = other.m_sockfd;
+        m_sockAddr = other.m_sockAddr;
+        m_maxMessageSize = other.m_maxMessageSize;
```
elBoberido, line 136: please revert

iceoryx_posh/test/moduletests/test_roudi_process_manager.cpp
```diff
@@ -121,7 +122,7 @@ TEST_F(ProcessManager_test, HandleProcessShutdownPreparationRequestWorks)
 {
     m_sut->registerProcess(m_processname, m_pid, m_user, m_isMonitored, 1U, 1U, m_versionInfo);
 
-    auto user = iox::posix::PosixUser::getUserOfCurrentProcess().getName();
+    auto user = iox::posix::PosixUser{iox::posix::PosixUser::getUserOfCurrentProcess().getName()};
```
elBoberido, line 125: please use just `auto user = iox::posix::PosixUser::getUserOfCurrentProcess();`

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/posix_access_rights.hpp
```diff
@@ -68,8 +68,8 @@ class PosixUser
     using groupVector_t = cxx::vector<PosixGroup, MaxNumberOfGroups>;
     using string_t = cxx::string<100>;
 
-    PosixUser(uid_t f_id);
-    PosixUser(const string_t& f_name);
+    explicit PosixUser(const uid_t f_id);
+    explicit PosixUser(const string_t& f_name);
```
elBoberido, line 72: This might count as public API change and should be mentioned in the changelog. @dkroenke what do you think?

elBoberido (dismissed):

iceoryx_hoofs/source/posix_wrapper/signal_handler.cpp
elBoberido: Please don't disable formatting if there are other viable options. We are very hesitant to disable clang-format and currently it is only disabled in a few places. When we do it all over the place everybody will do it which will result in a mess.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -140,9 +140,9 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
     static constexpr int32_t INVALID_FD = -1;
 
     UdsName_t m_name;
-    IpcChannelSide m_channelSide;
+    IpcChannelSide m_channelSide = IpcChannelSide::CLIENT;
```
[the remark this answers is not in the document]
elfenpiff: I like this actually more. It is visually more pleasant to read and I use the braces one only when the compiler forces me.

I know the core guidelines are suggesting it :/ but why does the better approach has to be uglier and non-intuitively.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
```diff
@@ -98,7 +98,7 @@ class UnixDomainSocket : public DesignPattern::Creation<UnixDomainSocket, IpcCha
 
     /// @brief checks whether the unix domain socket is outdated
     /// @return true if the unix domain socket is outdated, false otherwise, IpcChannelError if error occured
-    cxx::expected<bool, IpcChannelError> isOutdated() noexcept;
+    static cxx::expected<bool, IpcChannelError> isOutdated() noexcept;
```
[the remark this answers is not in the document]
elfenpiff: @elBoberido I would disagree. `isOutdated` does not access the object at all and can be therefore static.
But you can also call it like it would be a member function. `mySocket.isOutdated()` works fine.

I think this is a consequence of our ugly IPC design which has to be refactored.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/named_pipe.hpp
```diff
@@ -67,7 +67,7 @@ class NamedPipe : public DesignPattern::Creation<NamedPipe, IpcChannelError>
 
     /// @brief for compatibility with IpcChannelError
     /// @return always false
-    cxx::expected<bool, IpcChannelError> isOutdated() noexcept;
+    static cxx::expected<bool, IpcChannelError> isOutdated() noexcept;
```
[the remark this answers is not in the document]
elfenpiff: @elBoberido here the same issue `isOutdated` is always false for a named pipe and the function never accesses the underlying object therefore it could remain static.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/access_control.hpp
```diff
@@ -113,9 +113,9 @@ class AccessController
 
     cxx::vector<PermissionEntry, MaxNumOfPermissions> m_permissions;
 
-    cxx::expected<smartAclPointer_t, AccessControllerError> createACL(const int32_t f_numEntries) const;
-    bool createACLEntry(const acl_t f_ACL, const PermissionEntry& f_entry) const;
-    bool addAclPermission(acl_permset_t f_permset, acl_perm_t f_perm) const;
+    static cxx::expected<smartAclPointer_t, AccessControllerError> createACL(const int32_t f_numEntries);
+    static bool createACLEntry(const acl_t f_ACL, const PermissionEntry& f_entry);
+    static bool addAclPermission(acl_permset_t f_permset, acl_perm_t f_perm);
```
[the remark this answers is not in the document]
elfenpiff: Yes I would prefer it this way. With a static member function you clearly state that the method is independent of the object and therefore it is guaranteed that the state is unchanged after a call. (No magic mutable which changes somehow)

It makes it also more clear that this is in the end a free helper function which is only required for the implementation of that class.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/semaphore.hpp
```diff
@@ -269,7 +269,7 @@ class Semaphore : public DesignPattern::Creation<Semaphore, SemaphoreError>
     /// undefined behavior.
     ///
     /// @return returns false when sem_init fails otherwise true
-    bool init(iox_sem_t* handle, const int pshared, const unsigned int value) noexcept;
+    static bool init(iox_sem_t* handle, const int pshared, const unsigned int value) noexcept;
```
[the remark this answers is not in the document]
elfenpiff: Me like it

iceoryx_hoofs/source/concurrent/loffli.cpp
```diff
@@ -40,6 +40,7 @@ void LoFFLi::init(cxx::not_null<Index_t*> freeIndicesMemory, const uint32_t capa
     {
         for (uint32_t i = 0; i < m_size + 1; i++)
         {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) upper-bound of index is checked above
```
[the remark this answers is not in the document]
elfenpiff: You are somehow right but it makes sense to provide an argument why the next line should be ignored by the linter. How would you handle it? @elBoberido

iceoryx_hoofs/source/posix_wrapper/access_control.cpp
```diff
@@ -34,7 +34,7 @@ bool AccessController::writePermissionsToFile(const int32_t f_fileDescriptor) co
         return false;
     }
 
-    auto maybeWorkingACL = createACL((m_permissions.size() + m_useACLMask) ? 1 : 0);
+    auto maybeWorkingACL = createACL((m_permissions.size() + static_cast<unsigned long>(m_useACLMask)) != 0U ? 1 : 0);
```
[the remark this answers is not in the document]
elfenpiff: @elBoberido 
> shouldn't this be m_permissions.size() + (m_useACLMask ? 1 : 0)?
You are right and I support your suggestion.

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
```diff
@@ -92,8 +92,9 @@ FileLock& FileLock::operator=(FileLock&& rhs) noexcept
 
         CreationPattern_t::operator=(std::move(rhs));
 
-        m_name = std::move(rhs.m_name);
-        m_fd = std::move(rhs.m_fd);
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_name = std::move(rhs.m_name); // NOLINT(hicpp-invalid-access-moved)
+        m_fd = std::move(rhs.m_fd);     // NOLINT(hicpp-invalid-access-moved, hicpp-move-const-arg)
```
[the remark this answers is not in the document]
elfenpiff: In the line above ` CreationPattern_t::operator=(std::move(rhs));` is called therefore the linter thinks that `rhs` is no longer valid which is actually only partially true since only the creation pattern members are moved.

iceoryx_hoofs/source/posix_wrapper/message_queue.cpp
```diff
@@ -117,10 +118,12 @@ MessageQueue& MessageQueue::operator=(MessageQueue&& other)
         }
         CreationPattern_t::operator=(std::move(other));
 
-        m_name = std::move(other.m_name);
-        m_attributes = std::move(other.m_attributes);
-        m_mqDescriptor = std::move(other.m_mqDescriptor);
-        m_channelSide = std::move(other.m_channelSide);
+        // for details see here iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
+        m_name = std::move(other.m_name); // NOLINT(hicpp-invalid-access-moved)
```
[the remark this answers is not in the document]
elfenpiff: See comment: https://github.com/eclipse-iceoryx/iceoryx/pull/895#discussion_r690428540

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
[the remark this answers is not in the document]
elBoberido: @elfenpiff I disagree here, even if is a bad API it doesn't make it better if there is a static `ìsOutdated` function without any parameter. The IPC channel currently requires such a method and if one starts using `Foo::isOutdated()` and later on changes the using declaration to a message queue, the code breaks. We need to fulfill the contract and the contract says it is a method.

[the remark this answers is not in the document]
elfenpiff: @elBoberido I think the contract states that `Foo.isOutdated()` is compilable.
But we can make it unstatic again :D

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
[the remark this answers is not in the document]
elBoberido: I guess this is something like having `const` for parameter passes by value ;)

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/named_pipe.hpp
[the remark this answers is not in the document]
elBoberido: @elfenpiff as already mentioned, this breaks the API since it can be used as `Foo::bar` but is only expected to be used as `foo.bar`

iceoryx_hoofs/source/concurrent/loffli.cpp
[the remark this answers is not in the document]
elBoberido: Somehow misunderstood this comment. Above refers to the for loop. Please be more explicit.

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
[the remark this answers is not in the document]
elBoberido: The file does not contain an explanation for the lint but only for the move of the base class.

Wouldn't the proper solution be to just move the base class?
```cpp
CreationPattern_t::operator=(std::move(*static_cast<CreationPattern_t*>(&rhs)));
```
Yes, it's ugly but without the pointer detour the copy elision optimization is prevented ... and we really need to get rid of the creation pattern ;)

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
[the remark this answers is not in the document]
elBoberido: @elfenpiff sure but that would then break the API and you know Hyrum's Law, even if it doesn't make sense, there will be developer using it

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/posix_access_rights.hpp
dkroenke: Added an entry in changelog for that

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
dkroenke: Added #832 as ticket number and for later refactoring

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/unix_domain_socket.hpp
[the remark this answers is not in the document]
dkroenke: @elBoberido @elfenpiff 
I removed the static here and in the other occurences since it is dangerous to use when the underlying IPC Channel changes.

iceoryx_hoofs/source/posix_wrapper/timespec.cpp
```diff
@@ -16,15 +16,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "iceoryx_hoofs/internal/posix_wrapper/timespec.hpp"
```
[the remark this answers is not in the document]
dkroenke: Removed

iceoryx_hoofs/source/log/logstream.cpp
```diff
@@ -99,7 +100,8 @@ LogStream& LogStream::operator<<(const LogRawBuffer& value) noexcept
     for (int8_t i = 0; i < value.size; ++i)
     {
         // the '+value' is there to not interpret the uint8_t as char and print the character instead of the hex value
-        ss << (i > 0 ? " " : "") << std::setw(2) << +value.data[i];
+        ss << (i > 0 ? " " : "") << std::setw(2)
+           << +value.data[i]; // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
```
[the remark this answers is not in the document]
dkroenke: Yes, i cleaned up the NOLINT statement a bit.

iceoryx_hoofs/test/moduletests/test_posix_signal_handler.cpp
```diff
@@ -66,7 +66,7 @@ class SignalHandler_test : public Test
 
     void registerSignal(const int signal, void (*const callback)(int))
     {
-        struct sigaction action;
+        struct sigaction action = {};
```
elfenpiff, line 69: Please initialize it like `struct sigaction action{};`

elfenpiff (changes requested):

iceoryx_hoofs/test/moduletests/test_posix_signal_handler.cpp
dkroenke: With that, clang-format will do it that way:
```cpp
        struct sigaction action
        {
        };
```

The other proposal to remove the `struct` keyword is not applicable everywhere.
Using ` = {};` might look a bit redundant but we avoid using `// clang-format`. I would prefer to use it that way consistently.

iceoryx_hoofs/source/posix_wrapper/unix_domain_socket.cpp
```diff
@@ -262,7 +269,7 @@ cxx::expected<std::string, IpcChannelError> UnixDomainSocket::receive() const no
 {
     // we also support timedReceive. The setsockopt call sets the timeout for all further recvfrom calls, so we must set
     // it to 0 to turn the timeout off
-    struct timeval tv;
+    struct timeval tv = {};
```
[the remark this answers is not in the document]
dkroenke: Unfortunately we cannot do that (yet):
```
C++ designated initializers only available with ‘-std=c++2a’
```

iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
```diff
@@ -142,6 +142,13 @@ class Creation
     ErrorType m_errorValue;
 };
 
+template <typename Class, typename Error>
+void creationMove(Creation<Class, Error>& lhs, Creation<Class, Error>&& rhs) noexcept
+{
+    using CreationPattern_t = Creation<Class, Error>;
+    lhs.operator=(std::forward<CreationPattern_t>(rhs));
+}
```
elBoberido, line 150: Please don't do it this way. With this you still have to do `creationMove(*this, std::move(rhs))` which caused the linter warnings or `creationMove(*this, std::move(*static_cast<Creation<Foo>*>(&rhs)))` which we want to prevent with this method.

Please rewrite this to
```cpp
void creationMove(Creation<Class, Error>& rhs)
{
    *this = std::move(rhs);
}
```

This might cause some warnings though. In this case, please use a macro.

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
```diff
@@ -90,13 +90,13 @@ FileLock& FileLock::operator=(FileLock&& rhs) noexcept
                       << "\" in the move constructor/move assingment operator" << std::endl;
         }
 
-        CreationPattern_t::operator=(std::move(rhs));
-
         m_name = std::move(rhs.m_name);
-        m_fd = std::move(rhs.m_fd);
+        m_fd = rhs.m_fd;
```
elBoberido, line 94: please revert and check that all the code changes doing similar stuff is also reverted

iceoryx_hoofs/source/posix_wrapper/semaphore.cpp
```diff
@@ -39,15 +37,21 @@ Semaphore& Semaphore::operator=(Semaphore&& rhs) noexcept
     {
         closeHandle();
 
-        CreationPattern_t::operator=(std::move(rhs));
-
         m_name = std::move(rhs.m_name);
+        // NOLINTNEXTLINE(hicpp-move-const-arg) might break silently if object becomes non-trivial
```
elBoberido, line 41: Are these linter attributes still necessary? If yes, we need to find a solution which makes them obsolete. It's kind of embarrassing to have these in our code base.

elBoberido (commented):
Just skimmed over some files and didn't do a thorough review. Is this really a PR with 12858 insertions and 365 removals? I think there is something fishy. Do we maybe have a line ending change?

It would also have been better to create multiple PRs for this since it has quite some changes where small details have a big impact ebeded in an ocean of trivial changes.

iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/creation.hpp
elBoberido: On a second thought, maybe we need to do this completely different https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-virtual

dkroenke:
@elfenpiff @elBoberido @r7vme I see that this does not make much sense here with this huge PR, too much open topics and clang-tidy exceptions in the code. I propose that this PR goes into draft state until the reviewers have agreed on migration strategy.

elBoberido:
Maybe some parts can be reverted and and only the non controversial stuff be merged first. The naming for example makes up most of the PR and I think we can all agree on this. Stuff like with the move of the creation pattern should be left out and be done in a follow up.

This PR should then also be squashed before merging else there are huge intermediate changes which mess up history. There are for sure not 12858 additions in this PR. Something is fishy with that.

dkroenke:
> Maybe some parts can be reverted and and only the non controversial stuff be merged first. The naming for example makes up most of the PR and I think we can all agree on this. Stuff like with the move of the creation pattern should be left out and be done in a follow up.
> 
> This PR should then also be squashed before merging else there are huge intermediate changes which mess up history. There are for sure not 12858 additions in this PR. Something is fishy with that.

@elBoberido 
The reason for the huge diff was a generated file from a unittest that accidently was committed. I modified the test and added the file to gitignore. 
The controversial parts in the move assignment operators are for now excluded from clang-tidy and needs to be addressed later, once an agreement is found.
@elfenpiff This is an agenda point for the next developer meetup.

Maybe that has more chances now.

elfenpiff (approved):

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/posix_access_rights.hpp
dkroenke: @FerdinandSpitzschnueffler You are right, I reworked the Changelog accordingly.

FerdinandSpitzschnueffler (approved):

elfenpiff (approved):

Issue #911, opened 2021-08-25
Support for dynamic size types


elfenpiff:
## Brief feature description

At the moment it is impossible to send types in a zero copy manner which have a size that is only known at runtime. The current approach is to use a `std::vector` or similar container and copy/serialize the content into an untyped message via the untyped publisher.
I propose to adjust our containers `cxx::vector`, `cxx::string` and maybe later `cxx::(forward)list` so that they support dynamic size instead of only compile time size.

## Detailed information

The cxx containers should be extended so that a custom allocator can be attached which allocates memory inside of the shared memory. 

 * The extension should be done in a manner so that the capacity template parameter still works, e.g. `cxx::vector<int, 10> bla;` should create a vector with a stack allocator of size 10.
 * The allocator should not effect function developers which are using such containers as input parameters. This would be bad: `void myFunction(cxx::vector<int, SpecificAllocator> & v);` we want `void myFunction(cxx::vector<int> &v)` without providing the allocator type.
 * Since the containers are zero copy containers the capacity has to be reserved first before filling them, otherwise intransparent copies occur due to resizing allocated memory which cannot be done in a contiguous way.
   ```cpp
   cxx::vector<int, SomeAllocator> myVec;
   myVec.reserve(10); // only after reserve the container can be filled with elements
   myVec.emplace_back(1234);
   ```

## Tasks

- [ ] Write design document
- [ ] Implement allocator which does not need to be configured, (see mempool config of roudi)

elfenpiff:
@Indra5196 @rex-schilasky this is maybe interesting for you. Especially the Usage example, would you be happy with such an API or do you see some problems?

mossmaurice:
Nice feature & ideas!

Few questions that popped up regarding 74f6d10:

* What happens if the user is using the `ShmAllocator` but has no shared memory mapped to virtual address space?
* I assume the `ShmAllocator` would be part of `iceoryx_posh` and not `iceoryx_hoofs`, correct?
* > The allocator should not effect function developers .. without providing the allocator type
  * Can you elaborate on how you want to solve it? 

You can also answer them later in the design PR or maybe it's also a good idea to put this onto the agenda of tomorrow's iceoryx developer meetup.

@ZhenshengLee might also be interesting for you.

elfenpiff:
[...]

PR #912, opened 2021-08-26
Iox #911 design document dynamic size types

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Design document task in #911

### DIFF
--- a/doc/design/drafts/dynamic_size_types.md
+++ b/doc/design/drafts/dynamic_size_types.md
@@ -26,13 +26,29 @@ Those obstacles can be overcome when iceoryx containers like `cxx::string`,
 memory in a part of the shared memory which is also in the process space of the
 subscriber available.
 
-To accomplish this we require:
+To accomplish dynamic size types we require:
 
-1. an allocator concept
+1. an allocator concept which supports states for allocators
 2. have to adjust the containers like `cxx::string`, `cxx::vector` and 
    `cxx::{forward}list`
 3. the publisher has to provide access to the shared memory allocator via the
    sample
+
+### Allocator Type Does Not Effect Container Type
+
+One important restriction is for the allocator concept is that the specification
+of the allocator does not change the type like the STL C++ allocator concept
+does. Function developers for instance do not want to restrict their functions
+to shared memory allocated types which would result in some implementation
+overhead. One example could be that someone analyses an image with a function
+like
+`void analyseImage(cxx::vector<int, SharedMemoryAllocator> & a)`. This function
+can now be only used with a vector which is using the `SharedMemoryAllocator`.
+One could define the allocator type as a template but this would increase the
+compile time and can make it nearly impossible to write proprietary code since
+templates require the implementation in the header. Furthermore, the function
+developer has to be aware of the implementation detail that there is something
+like a `SharedMemoryAllocator`.
 
 ## Memory Layout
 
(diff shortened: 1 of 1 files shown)


doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,433 @@
...
+chunk where all the contents are stored inside one chunk. Hereby is the
+Memory Manager the class which handles the distribution of shared memory chunks.
+To support different sizes the Memory Manager uses bucket allocators with
+preconfigured bucket sizes called Memory Pool.
+
+```
++--------------------- Memory Manager -------------------+
+|  MemPool with chunk size = 1000                        |
+|  +----------------+----------------+----------------+  |
+|  |  data          |                |                |  |
```
FerdinandSpitzschnueffler, line 90: I think this should be ``value``.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,433 @@
...
+its local process.
+
+## Design Allocator
+
+The allocator should be designed in a manner so that it can be used inside the
+shared memory in an IPC context. This means the design has to consider that
+vtables, virtual and inheritance are not allowed as well as function pointers
+or the use of `cxx::function` or `cxx::function_ref`.
+
+Therefore this allocator concept is less flexibel and more complex as it would
```
r7vme, line 144:
```suggestion
Therefore this allocator concept is less flexible and more complex as it would
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+### Usage
+
+```cpp
+cxx::vector<int, 20> a;
+a.emplace_back(123); // can be used without a preceding reserve call
+
+cxx::vector<int> c(heapAllocator);
+c.reserve(42); // reserve required
```
r7vme, line 377: what should happen if no reserve was called?

elfenpiff: In the case of the static vector it is not required and nothing will happen. For all other vectors `cxx::Expects` is called with the expectation that reserve was called before - if not `cxx::Expects` will call `std::terminate()` but this can be configured to something else later when we have a more fine grained error handling concept.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,433 @@
...
+of a vector. It makes the following operations possible
+```cpp
+void f(const vector<int, 20> &a) {}
+
+vector<int> v1;
+vector<int, 20> v2;
+
+// this works thanks to implicit conversions from vector<int>
+f(v1);
+v2 = v1;
```
FerdinandSpitzschnueffler, line 336: What happens when the size of `v1` is greater than 20?

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,433 @@
...
++--------------------------------------------------------+
+```
+
+### Layout With Dynamic Sized Types
+
+With dynamic sized types we move the data part outside of the container and
+add an allocator to the container which allows the container to acquire
+memory for their data part. When we assume that `MyData::image` and
+`MyData::text` require 100 bytes for their data part the memory layout could
+look like this.
```
FerdinandSpitzschnueffler, line 112: What happens when there is not enough memory for the data Part?

doc/design/drafts/dynamic_size_types.md
FerdinandSpitzschnueffler: Can reserve fail?

doc/design/drafts/dynamic_size_types.md
elfenpiff: I adjusted the allocator interface so that it provides the memory packed inside an expected with a custom allocator error type. This is error is forwarded via `reserve` when it fails.

doc/design/drafts/dynamic_size_types.md
elfenpiff: It fails, a detailed explanation was added to the document.

doc/design/drafts/dynamic_size_types.md
elfenpiff: The allocator returns an error via expected which is forwarded to the user when reserve fails.

elfenpiff:
@elBoberido @FerdinandSpitzschnueffler since it is a design document only I would be happy if you can accept the PR so that it can be merged. If too many questions are opened I would be happy to add the to the design document or to move the design document into the draft folder.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+|  |   |   |   |   |   |   |   |   |   |   |   |   |   | |
+|  +---+---+---+---+---+---+---+---+---+---+---+---+---+ |
++--------------------------------------------------------+
+```
+
+### Layout With Dynamic Sized Types
+
+With dynamic sized types we move the data part outside of the container and
+add an allocator to the container which allows the container to acquire
+memory for their data part. When we assume that `MyData::image` and
```
FerdinandSpitzschnueffler, line 110: Please describe this in more detail. The data part will now be stored in separate chunks and the management part will be inside one chunk, now including (a pointer to) the allocator, right?

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+        // suggested allocator
+
+  data; // contains the user data which is stored in the container
+}
+```
+
+### Static Layout
+
+When we require a sample with our publisher and write `MyData` we acquire a
+chunk where all the contents are stored inside one chunk. Hereby is the
```
FerdinandSpitzschnueffler, line 81:
```suggestion
When we require a sample with our publisher and write `MyData`, we acquire a
chunk where all the contents are stored. Hereby is the
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+```
+
+This does not require any adaptions on the subscriber side since the subscriber
+already mapped this memory region which is handled by the memory manager into 
+its local process.
+
+## Design Allocator
+
+The allocator should be designed in a manner so that it can be used inside the
+shared memory in an IPC context. This means the design has to consider that
```
FerdinandSpitzschnueffler, line 140:
```suggestion
The allocator should be designed to be used inside the
shared memory within an IPC context. This means the design has to consider that
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+vtables, virtual and inheritance are not allowed as well as function pointers
+or the use of `cxx::function` or `cxx::function_ref`.
+
+Therefore this allocator concept is less flexible and more complex as it would
+be when everything would run in the same process since we have to use relative
+pointer and have to allocate everything on the stack.
```
FerdinandSpitzschnueffler, line 146:
```suggestion
pointers and have to allocate everything on the stack.
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+               |   - expected<relative_ptr, AllocatorError> allocate(uint64_t) |
+               |   - void free(relative_ptr chunk)                             |
+               |   - template<AnotherAllocator>                                |
+               |     bool isMoveCompatible(const AnotherAllocator & rhs)       |
+               |                                                               |
+               |   # relative_ptr m_ptrToAllocator                             |
+               +---------------------------------------------------------------+
+```
+The `Allocator` described in this diagram is similar to a C++20 concept which can
+be verified at compiletime without inheritance. The verification can be realized
```
FerdinandSpitzschnueffler, line 207:
```suggestion
be verified at compile time without inheritance. The verification can be realized
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   Stack based containers do not require a `reserve()` call beforehand but
+   should provide this call as well for compatibility reasons.
+
+   * It is not possible to provide solely the capacity (`reserve` parameter) via
+     the constructor of the container since frameworks like ROS2 cannot forward
+     such arguments when for instance `loan` was called. But it is possible to
+     provide it as an alternative.
+
+6. The allocator type should not change the underlying container type, e.g. the
+   allocator should not provided as template argument, otherwise
```
FerdinandSpitzschnueffler, line 300:
```suggestion
   allocator should not be provided as template argument, otherwise
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   ```
+   After a copy- or move assignment operation the allocator which was provided
+   in the constructor will not change.
+
+   When the copy- or move-constructor is used the allocator of the origin shall
+   be used in the newly created container.
+
+   * Real copy- and move operations could have some unwanted side effects. Lets
+     assume one would like to copy/move a shared memory located vector into a local
+     cache. If the allocator would change the container has to allocate another
```
FerdinandSpitzschnueffler, line 327:
```suggestion
     cache. If the allocator would change, the container has to allocate another
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+       |  - vector& operator=(vector<T> &&)          |
+       |                                             |
+       |                                             |
+       |  // same API as current cxx::vector         |
+       |                                             |
+       |  # uint8_t m_data[Capacity]                 |
+       |  # RangeAllocator m_allocator               |
+       +---------------------------------------------+
+```
+All non stack based version of cxx containers will not have constructors similar
```
FerdinandSpitzschnueffler, line 380:
```suggestion
All non stack based versions of cxx containers will not have constructors similar
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+// this works since vector<int, 10> is also a vector<int> and can be converted
+v2 = v1;
+v1 = v2; // this can fail when v1 capacity is smaller then v2.size, previously 
+         // it was only possible when v1.capacity < v2.capacity
+f(v1);
+f(v2);
+```
+
+The only restriction are functions which require a stack based vector of a 
+specific size as non const reference. This is usally the case when some function
```
FerdinandSpitzschnueffler, line 420:
```suggestion
The only limitation is functions which require a stack based vector of a 
specific size as non const reference. This is usally the case when some function
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+allocator. The internal structure of those containers looks similar to the
+structure of their `cxx` pendants
+```cpp
+template<typename T>
+class vector {
+  //...
+  T * memoryStorageOfVectorElements;
+};
+```
+The pointer hereby is an absolut pointer pointing to a piece of memory in the
```
FerdinandSpitzschnueffler, line 497:
```suggestion
The pointer hereby is an absolute pointer pointing to a piece of memory in the
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+     struct MyData {
+       cxx::vector<int> value;
+     };
+
+     dataOnSharedMemory = subscriber->take();
+     MyData cache{.value = cxx:vector<int>(HeapAllocator)};
+     cache = dataOnSharedMemory; // would suddenly allocate shared memory
+     ```
+
+The draft we provide here is using the `cxx::vector` but the techniques
```
FerdinandSpitzschnueffler, line 343:
```suggestion
The provided draft uses the `cxx::vector` but the techniques
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+       |  // same API as current cxx::vector         |
+       |                                             |
+       |  # uint8_t m_data[Capacity]                 |
+       |  # RangeAllocator m_allocator               |
+       +---------------------------------------------+
+```
+All non stack based version of cxx containers will not have constructors similar
+to the STL. The only provided constructor has one argument which specifies the
+type of allocator one would like to use for that container. The default constructor
+shall construct a vector which has a capacity of 0.
```
FerdinandSpitzschnueffler, line 383: Can you please elaborate? Why can't we have typical c'tors with the allocator as last parameter with a default value, e.g. the heap allocator? This would be more user friendly.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+
+1. Send via untyped publisher and provide the size at runtime to the publisher.
+   But defining message types with multiple dynamic sized containers would be
+   a real challenge when they should be constructed inside of this sample.
+
+2. One can perform a worst case estimation at compile time and declare
+   a type with container members which fullfil those estimations.
+   This could lead to a lot of wasted memory and if the worst case estimations
+   have to be adjusted all the applications which are using this type may have
+   to be adjusted as well since the capacity is part of the type.
```
MatthiasKillat, line 17: These estimations are also often hard to come by for more complicated structures, e.g. specific trees with internal meta data nodes etc. The estimates then are very pessimistic and as stated here lead to a lot of wasted memory which cannot be avoided by this approach. Hence it is not really an option if memory is a concern.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,496 @@
...
+       |  - vector(vector<T> &&)                     |
+       |  - vector& operator=(const vector<T> &)     |
+       |  - vector& operator=(vector<T> &&)          |
+       |                                             |
+       |                                             |
+       |  // same API as current cxx::vector         |
+       |                                             |
+       |  # uint8_t m_data[Capacity]                 |
+       |  # RangeAllocator m_allocator               |
+       +---------------------------------------------+
```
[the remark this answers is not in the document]
elfenpiff: I added this idea to the design document under open questions to keep this idea documented.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   a real challenge when they should be constructed inside of this sample.
+
+2. One can perform a worst case estimation at compile time and declare
+   a type with container members which fullfil those estimations.
+   This could lead to a lot of wasted memory and if the worst case estimations
+   have to be adjusted all the applications which are using this type may have
+   to be adjusted as well since the capacity is part of the type.
+
+3. Another approach is to use the heap and copy/serialize the data into a sample
+   before sending it with iceoryx. This would imply multiple copies since the
```
MatthiasKillat, line 20: Leads to one serialization and one deserialization and is optimal *if data has to be serialized anyway*.

No copies needed if done correctly, but tota effort is effort of Omega(size of data to be sent), i.e. at least linear as a copy with more complex logic (leading to increased costs compared to copies).

doc/design/drafts/dynamic_size_types.md
elfenpiff: @FerdinandSpitzschnueffler I added a more detailed description. Can you please check it out.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,685 @@
...
+|  |   |   |   |   |   |   |   |   |   |   |   |   |   | |
+|  |   |   |   |   |   |   |   |   |   |   |   |   |   | |
+|  +---+---+---+---+---+---+---+---+---+---+---+---+---+ |
++--------------------------------------------------------+
+```
+
+### Layout With Dynamic Sized Types
+
+With dynamic sized types we split the container data into two parts, the
+the actual data and a management part which contains a pointer to the data part.
```
FerdinandSpitzschnueffler, line 109:
```suggestion
actual data and a management part which contains a pointer to the data part.
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,685 @@
...
+The data part lives then outside of the container and an allocator inside the
+management part of the container allows the container to acquire
+memory for their data part. When we assume that `MyData::image` and
```
FerdinandSpitzschnueffler, line 112:
```suggestion
memory for its data part. When we assume that `MyData::image` and
```

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   before sending it with iceoryx. This would imply multiple copies since the
+   publisher has to copy the data from the container into the sample and every
+   subscriber may has to deserialize/copy the data out of the iceoryx sample.
+
+Those obstacles can be overcome when iceoryx containers like `cxx::string`,
+`cxx::vector` or `cxx::{forward}list` support custom allocators which allocate
+memory in a part of the shared memory which is also in the process space of the
+subscriber available.
+
+To accomplish dynamic size types we require:
```
MatthiasKillat, line 29: To support ...

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+
+1. an allocator concept which supports states for allocators
```
MatthiasKillat, line 31: What is 1. supposed to mean?

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+to shared memory allocated types which would result in some implementation
+overhead. One example could be that someone analyses an image with a function
+like
+`void analyseImage(cxx::vector<int, SharedMemoryAllocator> & a)`. This function
+can now be only used with a vector which is using the `SharedMemoryAllocator`.
+One could define the allocator type as a template but this would increase the
+compile time and can make it nearly impossible to write proprietary code since
+templates require the implementation in the header. Furthermore, the function
+developer has to be aware of the implementation detail that there is something
+like a `SharedMemoryAllocator`.
```
MatthiasKillat, line 51: Basically the approach of https://en.cppreference.com/w/cpp/memory/polymorphic_allocator but we will have additional complications due to shared memory (across processes).

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+  mgmt; // contains internal variables to handle the logic and later the
+        // suggested allocator
+
+  data; // contains the user data which is stored in the container
+}
+```
+
+### Static Layout
+
+When we require a sample with our publisher and write `MyData`, we acquire a
```
MatthiasKillat, line 80: request or acquire a sample

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+|  |   |   |   |   |   |   |   |   |   |   |   |   |   | |
+|  |   |   |   |   |   |   |   |   |   |   |   |   |   | |
+|  +---+---+---+---+---+---+---+---+---+---+---+---+---+ |
++--------------------------------------------------------+
+```
+
+### Layout With Dynamic Sized Types
+
+With dynamic sized types we move the data part outside of the container and
+add an allocator to the container which allows the container to acquire
```
MatthiasKillat: important: I think we actually only should add a reference/pointer to the allocator in general.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+|  |   |   |   |   |   |   |   |   |   |   |   |   |   | |
+|  +---+---+---+---+---+---+---+---+---+---+---+---+---+ |
++--------------------------------------------------------+
+```
+
+This does not require any adaptions on the subscriber side since the subscriber
+already mapped this memory region which is handled by the memory manager into 
+its local process.
+
+## Design Allocator
```
MatthiasKillat, line 139: transpose

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+  |  AllocatorError [enum class] |
+  |                              |
+  |     # OUT_OF_MEMORY          |
+  |     # MEMORY_SIZE_TOO_LARGE  |
+  +------------------------------+
+
+  +---------------------------------------------------------------+
+  |  Allocator [Concept]                                          |
+  |                                                               |
+  |   - expected<relative_ptr, AllocatorError> allocate(uint64_t) |
```
MatthiasKillat, line 161: We should not return relaive_ptr or similar. There is no need, we can return raw pointers. Only if we want to store them for IPC use we need to consider fancy pointers. This is not the responsibility of allocate.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+  |   - void free(relative_ptr chunk)                             |
+  |   - template<AnotherAllocator>                                |
+  |     bool isMoveCompatible(const AnotherAllocator& rhs)        |
+  +---+--------+--------------------------------------------------+
+      |        |
+      |  +-----+---------------------------------------------------------+
+      |  | RangeAllocator                                                |
```
MatthiasKillat, line 168: Probably to early for specific allocators but sure, we want to support different specializations.

If we mention the different allocators here we should explain their purpose (HeapAllocator is also not an ideal name I think).

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+be verified at compile time without inheritance. The verification can be realized
+with a `IsAllocatorTypeTrait` in C++14.
+The shared memory and IPC restrictions are forcing us to implement a
+VariantAllocator which provides us type independent access to the underlying
+allocator like inheritance would provide. Hereby the implementation for
+`allocate`, `free` and `isMoveCompatible` is always the same and can be
+implemented similar to the `cxx::variant` move and copy operations. This would
+make the `VariantAllocator` easier extendable since the user just has to add
+any additional allocator as type in the variadic template list. An alternative
+way of implementing this `VariantAllocator` would be with a switch statement
```
MatthiasKillat, line 218: switch is a bad alternative and essentially variant abstracts something like this.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   myFunkyFunction(b);
+   myFunkyFunction(c);
+   ```
+
+4. All containers must be shared memory compatible, this means no vtables and
+   virtual. Furthermore the usage of function pointers, `cxx::function` and
+   `cxx::function_ref` is forbidden in all containers and their underlying 
+   constructs.
+
+5. Before using any cxx container the user as to call `reserve()` once otherwise
```
MatthiasKillat, line 289: has to

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   the container will have a default capacity of zero. It is not allowed to
+   call `reserve()` multiple times since this could violate the zero copy
+   guarantee.
+   Stack based containers do not require a `reserve()` call beforehand but
+   should provide this call as well for compatibility reasons.
+
+   * It is not possible to provide solely the capacity (`reserve` parameter) via
```
MatthiasKillat, line 296: I think for this it is also important what kind of template types we support for say `vector`.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+7. The allocator should not be part of copy and move-assignment operations.
+   The container should behave as closely as possible to an STL container with
+   custom allocator.
+   ```cpp
+   cxx::vector a(HeapAllocator);
+   cxx::vector b(SharedMemoryAllocator);
+   cxx::vector c(AnotherAllocator);
+   cxx::vector d(HeapAllocator);
+
+   a = b; // the content is copied from shared memory into the heap
```
MatthiasKillat, line 316: I think the behavior needs to be clarified more here. I think `a` will/should deallocate its current memory (or extend it), allocate more memory if needed and then copy the data from b to this memory.

However, in general this operation needs to know something about the source memory here (and it will, since it has access to the allocator of the source `b`).

TLDR: the actual semantics need more detail here I think.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   ```
+   After a copy- or move assignment operation the allocator which was provided
+   in the constructor will not change.
+
+   When the copy- or move-constructor is used the allocator of the origin shall
+   be used in the newly created container.
+
+   * Real copy- and move operations could have some unwanted side effects. Lets
+     assume one would like to copy/move a shared memory located vector into a local
+     cache. If the allocator would change, the container has to allocate another
```
MatthiasKillat, line 329: Can/should the allocator of an individual vector ever change once constructed? What would be the semantics?

doc/design/drafts/dynamic_size_types.md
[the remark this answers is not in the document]
MatthiasKillat: Not sure what `payload_vector` is here, I probably missed something.

But we may also request the allocator from the publisher in a way. Then we could provide this allocator to the vector as sketched in the document, but of course we would have to be careful with the lifetime.

Of course the vector would now have a pointer to the allocator, which I think the proposal here tries to avoid.
I think 2. is very error-prone and cumbersome to use.

doc/design/drafts/dynamic_size_types.md
```diff
@@ -0,0 +1,562 @@
...
+   /* error handling */
+   return;
+}
+sample->sharedMemoryVector.emplace_back(someValue);
+```
+
+
+Copy and move operations should be possible without the developer knowing
+what kind of underlying allocator is used. Additionally, it should be as efficient
+as possible. That means the move operation should be a real move operation
```
MatthiasKillat, line 458: We could also define that move only works if source and destination allocator are the same (and maybe other conditions hold), otherwise we could always fall back to a copy.

Another approach is simply that the destiantion takes over the memory and allocator of the source and discards its own. If the allocator is not owned by the vector (but just a fancy pointer to it), this should be well defined and cheap, at least for vector.

The source of the move will not need allocator nor data (it should not be accessed anymore, but we can make it move aware for added safety at the cost of some runtime checks whether the allocator is valid/set).

elBoberido:
@elfenpiff can we close this. It also starts to smell ;)

If you just want to have a quick link to this PR, you can open a page on `Discussion` and link it from there.

elBoberido:
@elfenpiff  I think there is still too much discussion ongoing and `doc/design/draft` is more for stuff which is settled and anyone could start with implementation.

Like mentioned. My proposal would be to open a page in `Discussion` and link this PR from there. Then it is still easy to find this PR.

PR #917, opened 2021-09-01
iox-#415 Rename variable and test case name in `ServiceRegistry_test` and fix bug in `ServiceDescription` c'tor

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Fixes left-over findings from #860
  * Rename `sut` in `ServiceRegistry_test`
  * Rename one test case in `ServiceRegistry_test`
  * Fix bug in `ServiceDescription::ServiceDescription(const cxx::Serialization& serial)`
  * Replace `bool` with `void` as return type of `ServiceRegistry::remove`

## References

- Closes partly #415

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,7 +128,7 @@ information about logging and error handling.
 * Aliases have a `_t` postfix : `using FooString_t = iox::cxx::string<100>;`
 * Objects created from a method returning a `cxx::optional<Foo>` shall be named `maybeFoo`
 * Objects created from a method returning a `cxx::expected<Foo, FooError>` shall
-contain the name `result` e.g. `fooResult`
+contain the name `result` e.g. `getChunkResult` for the method `getChunk()`
 
 ### Doxygen
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,6 +126,9 @@ information about logging and error handling.
     * Public members of structs and classes do not have the `m_` prefix
 * Namespaces in `lower_snake_case` : `my_namespace`
 * Aliases have a `_t` postfix : `using FooString_t = iox::cxx::string<100>;`
+* Objects created from a method returning a `cxx::optional<Foo>` shall be named `maybeFoo`
+* Objects created from a method returning a `cxx::expected<Foo, FooError>` shall
+contain the name `result` e.g. `fooResult`
 
 ### Doxygen
 
--- a/iceoryx_posh/source/capro/service_description.cpp
+++ b/iceoryx_posh/source/capro/service_description.cpp
@@ -73,26 +73,41 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
 {
     std::underlying_type<Scope>::type scope = 0;
     std::underlying_type<Interfaces>::type interfaceSource = 0;
-    serial.extract(m_serviceString,
-                   m_instanceString,
-                   m_eventString,
-                   m_classHash[0u],
-                   m_classHash[1u],
-                   m_classHash[2u],
-                   m_classHash[3u],
-                   scope,
-                   interfaceSource);
+    auto deserializationSuccessful = serial.extract(m_serviceString,
+                                                    m_instanceString,
+                                                    m_eventString,
+                                                    m_classHash[0u],
+                                                    m_classHash[1u],
+                                                    m_classHash[2u],
+                                                    m_classHash[3u],
+                                                    scope,
+                                                    interfaceSource);
+    if (!deserializationSuccessful)
+    {
+        m_serviceString = iox::capro::InvalidIdString;
+        m_instanceString = iox::capro::InvalidIdString;
+        m_eventString = iox::capro::InvalidIdString;
+        return;
+    }
+
     if (scope > static_cast<std::underlying_type<Scope>::type>(Scope::INVALID))
     {
-        m_scope = Scope::INVALID;
+        m_serviceString = iox::capro::InvalidIdString;
+        m_instanceString = iox::capro::InvalidIdString;
+        m_eventString = iox::capro::InvalidIdString;
+        return;
     }
     else
     {
         m_scope = static_cast<Scope>(scope);
     }
+
     if (interfaceSource > static_cast<std::underlying_type<Interfaces>::type>(Interfaces::INTERFACE_END))
     {
-        m_interfaceSource = Interfaces::INTERFACE_END;
+        m_serviceString = iox::capro::InvalidIdString;
+        m_instanceString = iox::capro::InvalidIdString;
+        m_eventString = iox::capro::InvalidIdString;
+        return;
     }
     else
     {
--- a/iceoryx_posh/source/capro/service_description.cpp
+++ b/iceoryx_posh/source/capro/service_description.cpp
@@ -82,7 +82,8 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
                                                     m_classHash[3u],
                                                     scope,
                                                     interfaceSource);
-    if (!deserializationSuccessful)
+    if (!deserializationSuccessful || scope >= static_cast<std::underlying_type<Scope>::type>(Scope::INVALID)
+        || interfaceSource >= static_cast<std::underlying_type<Interfaces>::type>(Interfaces::INTERFACE_END))
     {
         m_serviceString = iox::capro::InvalidIdString;
         m_instanceString = iox::capro::InvalidIdString;
@@ -90,29 +91,8 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
         return;
     }
 
-    if (scope > static_cast<std::underlying_type<Scope>::type>(Scope::INVALID))
-    {
-        m_serviceString = iox::capro::InvalidIdString;
-        m_instanceString = iox::capro::InvalidIdString;
-        m_eventString = iox::capro::InvalidIdString;
-        return;
-    }
-    else
-    {
-        m_scope = static_cast<Scope>(scope);
-    }
-
-    if (interfaceSource > static_cast<std::underlying_type<Interfaces>::type>(Interfaces::INTERFACE_END))
-    {
-        m_serviceString = iox::capro::InvalidIdString;
-        m_instanceString = iox::capro::InvalidIdString;
-        m_eventString = iox::capro::InvalidIdString;
-        return;
-    }
-    else
-    {
-        m_interfaceSource = static_cast<Interfaces>(interfaceSource);
-    }
+    m_scope = static_cast<Scope>(scope);
+    m_interfaceSource = static_cast<Interfaces>(interfaceSource);
 }
 
 ServiceDescription::ServiceDescription() noexcept
--- a/iceoryx_posh/source/capro/service_description.cpp
+++ b/iceoryx_posh/source/capro/service_description.cpp
@@ -155,11 +155,11 @@ ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept
 {
     ServiceDescription deserializedObject;
 
-    using ScopeType = std::underlying_type<Scope>::type;
-    using InterfaceType = std::underlying_type<Interfaces>::type;
+    using ScopeUnderlyingType = std::underlying_type<Scope>::type;
+    using InterfaceUnderlyingType = std::underlying_type<Interfaces>::type;
 
-    ScopeType scope{0};
-    InterfaceType interfaceSource{0};
+    ScopeUnderlyingType scope{0};
+    InterfaceUnderlyingType interfaceSource{0};
 
     auto deserializationSuccessful = serialized.extract(deserializedObject.m_serviceString,
                                                         deserializedObject.m_instanceString,
@@ -170,8 +170,8 @@ ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept
                                                         deserializedObject.m_classHash[3u],
                                                         scope,
                                                         interfaceSource);
-    if (!deserializationSuccessful || scope >= static_cast<ScopeType>(Scope::INVALID)
-        || interfaceSource >= static_cast<InterfaceType>(Interfaces::INTERFACE_END))
+    if (!deserializationSuccessful || scope >= static_cast<ScopeUnderlyingType>(Scope::INVALID)
+        || interfaceSource >= static_cast<InterfaceUnderlyingType>(Interfaces::INTERFACE_END))
     {
         return cxx::error<cxx::Serialization::Error>(cxx::Serialization::Error::DESERIALIZATION_FAILED);
     }
--- a/iceoryx_posh/source/capro/service_description.cpp
+++ b/iceoryx_posh/source/capro/service_description.cpp
@@ -155,8 +155,12 @@ ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept
 {
     ServiceDescription deserializedObject;
 
-    std::underlying_type<Scope>::type scope = 0;
-    std::underlying_type<Interfaces>::type interfaceSource = 0;
+    using ScopeType = std::underlying_type<Scope>::type;
+    using InterfaceType = std::underlying_type<Interfaces>::type;
+
+    ScopeType scope{0};
+    InterfaceType interfaceSource{0};
+
     auto deserializationSuccessful = serialized.extract(deserializedObject.m_serviceString,
                                                         deserializedObject.m_instanceString,
                                                         deserializedObject.m_eventString,
@@ -166,8 +170,8 @@ ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept
                                                         deserializedObject.m_classHash[3u],
                                                         scope,
                                                         interfaceSource);
-    if (!deserializationSuccessful || scope >= static_cast<std::underlying_type<Scope>::type>(Scope::INVALID)
-        || interfaceSource >= static_cast<std::underlying_type<Interfaces>::type>(Interfaces::INTERFACE_END))
+    if (!deserializationSuccessful || scope >= static_cast<ScopeType>(Scope::INVALID)
+        || interfaceSource >= static_cast<InterfaceType>(Interfaces::INTERFACE_END))
     {
         return cxx::error<cxx::Serialization::Error>(cxx::Serialization::Error::DESERIALIZATION_FAILED);
     }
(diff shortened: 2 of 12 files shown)


elfenpiff (changes requested):
I reviewed it and to address review findings in a follow up pull request is a horrible idea since now a reviewer has to click through all the review findings to make sure they were addressed. And as it turns out not all of them were addressed and even not tracked within the issue.

Here I list them, please fix every issue and open point from the former PR otherwise we create another PR to address again review findings from a former PR and it is getting harder and harder to see which issue was handled and which not:

* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r655431580
* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r655433529
* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r655455522
* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r655455522
* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r659822894
* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r659861246
* https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r659889975

Either address the issue in this PR or add them as todo in the issue with a link to the discussion.
Furthermore @elBoberido and @MatthiasKillat should be the ones who approve this PR since they where the initial approvers

elfenpiff:
@mossmaurice Shouldn't the PR title be something like `iox-#123 Address...` since you checked `Update the PR title 
Follow the same conventions as for commit messages` ?

mossmaurice:
> I reviewed it and to address review findings in a follow up pull request is a horrible idea since now a reviewer has to click through all the review findings to make sure they were addressed.

Thanks for your feedback! I added the three findings that were fixed in this PR under "Notes for Reviewer".
TBH I think the iterative approach used here is not horrible, but exactly what we need in order to keep the PRs as small as possible. This is what we've done in the past and IMHO we should stick with it for future development.

>  And as it turns out not all of them were addressed and even not tracked within the issue.

I have the feeling that we have a different understanding on what is "an important finding" vs. what is "just nice to have"-personal-preference. I addressed all the findings in #860 and resolved them if there was no more feedback/answer from the original reviewer. All the important findings e.g. small bugs are tracked in #415 in a todo list. Usually, I tend to track smaller things or "nice-to-have" stuff with an inline comment `/// @todo #123`. Especially for pull requests that are open for several months, I would kindly ask you to do a re-review after ~1 months and take part in the discussion about a certain issue if you don't agree.

> Here I list them, please fix every issue and open point from the former PR otherwise we create another PR to address again review findings from a former PR and it is getting harder and harder to see which issue was handled and which not:

For easier reference let me replace the links with some short description:

1. > Connect `MAX_NUMBER_OF_SERVICES` with `MAX_SERVICE_DESCRIPTIONS`

Tracked via an inline comment.

2. >  `bool remove(ServiceDescription&)` or `void remove(ServiceDescription&)`

@MatthiasKillat and myself were both ok with the `bool`. However, I see your point, will adapt it.

3. >  `registry.find("bla", "blubb", "fuu")` vs. `registry.find(result, "bla", "blubb", "fuu")`

As written in the thread: @MatthiasKillat and myself concluded something like this is always a trade-off. As this is an internal API that will never be used by the user directly it's not worth the hassle IMHO and can also be easily adapted if we experience it to be error-prone.

4. >   https://github.com/eclipse-iceoryx/iceoryx/pull/860#discussion_r659822894

Resolved, comment from when the predecessor PR was not merged to `master` yet.

5. >  `iterator` not part of for loop head

Done on purpose to show it goes hand in hand with the index. This will be refactored once we integrate #859 

6. > Counted two occurrences of `m_serviceMap.equal_range`

@MatthiasKillat told me he won't support `equal_range` with #859 Hence, this will be refactored soonish. 

> Furthermore @elBoberido and @MatthiasKillat should be the ones who approve this PR since they where the initial approvers

I have the feeling that we need a vote for a [BDFL](https://en.wikipedia.org/wiki/Benevolent_dictator_for_life)? :smirk: Kidding aside, especially during vacation time we can't always wait for everyone to return. In general, I try to address & merge my open pull requests ASAP but sometimes the workload does not allow this. Let's improve our review speed and finding-fix-time with the next PRs :blush: For the record #860 was open 72 days.

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -73,26 +73,41 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
...
+                                                    interfaceSource);
+    if (!deserializationSuccessful)
+    {
+        m_serviceString = iox::capro::InvalidIdString;
+        m_instanceString = iox::capro::InvalidIdString;
+        m_eventString = iox::capro::InvalidIdString;
+        return;
+    }
+
     if (scope > static_cast<std::underlying_type<Scope>::type>(Scope::INVALID))
```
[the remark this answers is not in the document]
mossmaurice: Fixed

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -73,26 +73,41 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
...
+        m_instanceString = iox::capro::InvalidIdString;
+        m_eventString = iox::capro::InvalidIdString;
+        return;
     }
     else
     {
         m_scope = static_cast<Scope>(scope);
     }
+
     if (interfaceSource > static_cast<std::underlying_type<Interfaces>::type>(Interfaces::INTERFACE_END))
```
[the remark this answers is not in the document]
mossmaurice: Fixed

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -73,26 +73,41 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
...
+    auto deserializationSuccessful = serial.extract(m_serviceString,
+                                                    m_instanceString,
+                                                    m_eventString,
+                                                    m_classHash[0u],
+                                                    m_classHash[1u],
+                                                    m_classHash[2u],
+                                                    m_classHash[3u],
+                                                    scope,
+                                                    interfaceSource);
+    if (!deserializationSuccessful)
```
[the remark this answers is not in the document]
mossmaurice: I agree, that makes it more compact. Also adapted it in `client_options.cpp` to be consistent.

elfenpiff (dismissed):

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -73,31 +73,26 @@ ServiceDescription::ServiceDescription(const cxx::Serialization& serial) noexcep
...
-                   m_instanceString,
-                   m_eventString,
-                   m_classHash[0u],
-                   m_classHash[1u],
-                   m_classHash[2u],
-                   m_classHash[3u],
-                   scope,
-                   interfaceSource);
-    if (scope > static_cast<std::underlying_type<Scope>::type>(Scope::INVALID))
+    auto deserializationSuccessful = serial.extract(m_serviceString,
```
elBoberido, line 76: I guess this has to do with my comment regarding the invalid `ServiceDescription` from the previous PR? If yes, could you please refactor this further to something like this
```cpp
cxx::expected<ServiceDescription, cxx::Serialization::Error>
ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept;
```
so that the caller gets informed when this fails?

elfenpiff: Oh I totally overlooked this because I did not look in which method this is performed :( ...

@elBoberido you are right and shouldn't it be a static factory method of `ServiceDescription`?

Just as inspiration, in every class in which the CTor can fail we have the static factory method `create`. I would not use the creation pattern here (it should be slowly removed) but I would name the method `create` to be consistent with all the other occurrences all over the code.

Then we can make a rule. CTor succeeds always, if the CTor can fail we implement a static factory method called `create`.

elBoberido: @elfenpiff ahem, you just have to look closely to spot the `static`. If you can't see it, just move your eyes closer to the screen. If you still can't see it, you might need some glasses ;)

Regarding the naming. As rule of thumb I would agree, but depending on the context it should be possible to deviate from this naming. There is for example `Duration::from_seconds` which deviates and makes the code better to understand. Similarly I would also suggest to call this `deserialize` since it has a special meaning. This is maybe more important if there is a `create` function because of the regular ctor could fail.

mossmaurice: Done

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -231,7 +231,16 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
         }
         else
         {
-            capro::ServiceDescription service(cxx::Serialization(message.getElementAtIndex(2)));
+            auto maybeService =
```
elBoberido, line 234: It's a long time since we hacked on iceoryx so correct me if I'm wrong but didn't we use `maybeFoo` for `optional<Foo> bar()` and `fooResult` for `expected<Bar> foo()`?

iceoryx_posh/source/runtime/posh_runtime_impl.cpp
```diff
@@ -419,8 +419,13 @@ PoshRuntimeImpl::findService(const cxx::variant<Wildcard_t, capro::IdString_t> s
     uint32_t numberOfServices = algorithm::min(capacity, numberOfElements);
     for (uint32_t i = 0U; i < numberOfServices; ++i)
     {
-        capro::ServiceDescription service(cxx::Serialization(requestResponse.getElementAtIndex(i)));
-        serviceContainer.push_back(service);
+        auto maybeService =
```
elBoberido, line 422: depending on the outcome from above, this might also become something like `deserializeResult`

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -231,7 +231,16 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
...
+            {
+                LogError() << "Deserialization failed when '" << message.getElementAtIndex(2).c_str()
+                           << "' was provided\n";
+                break;
+            }
+            auto service = maybeService.value();
+
             cxx::Serialization portConfigInfoSerialization(message.getElementAtIndex(7));
 
             if (!service.isValid())
```
elBoberido, line 246: Further down there is the de-serialization of the publisher and subscriber option but without error handling. Could you either also fix those o create and issue?

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -148,7 +148,11 @@ TEST_F(ServiceDescription_test, ServiceDescriptionSerializationCreatesServiceDes
                                                      static_cast<uint16_t>(testScope),
                                                      static_cast<uint16_t>(testInterfaceSource));
 
-    ServiceDescription serviceDescription1 = ServiceDescription(serialObj);
+
+    auto maybeService = ServiceDescription::deserialize(serialObj);
```
elBoberido, line 152: To bypass the naming situation, you could do something like this
```cpp
ServiceDescription::deserialize(serialObj).and_then([&](const auto & sut){
    EXPECT_THAT(...);
    ...
}).or_else([&](const auto error){
    FAIL() << "Deserialization should not fail but failed with: " << error;
});
```

What do you think?

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -180,16 +185,16 @@ TEST_F(ServiceDescription_test, ServiceDescriptionObjectInitialisationWithOutOfB
                                                      testHash[3],
                                                      invalidScope);
 
-    ServiceDescription serviceDescription1 = ServiceDescription(serialObj);
+    auto maybeService = ServiceDescription::deserialize(serialObj);
 
-    EXPECT_THAT(serviceDescription1.getScope(), Eq(Scope::INVALID));
+    EXPECT_TRUE(maybeService.has_error());
```
elBoberido, line 190: Please also check for the correct error

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -207,12 +212,21 @@ TEST_F(ServiceDescription_test,
                                                      static_cast<uint16_t>(testScope),
                                                      invalidInterfaceSource);
 
-    ServiceDescription serviceDescription1 = ServiceDescription(serialObj);
+    auto maybeService = ServiceDescription::deserialize(serialObj);
+
+    EXPECT_TRUE(maybeService.has_error());
```
elBoberido, line 217: same here

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -207,12 +212,21 @@ TEST_F(ServiceDescription_test,
...
+
+TEST_F(ServiceDescription_test, ServiceDescriptionObjectInitialisationWithEmptyStringLeadsToInvalidDeserialization)
+{
+    std::string emptyString;
+    iox::cxx::Serialization invalidSerialObj{emptyString};
+
+    auto maybeService = ServiceDescription::deserialize(invalidSerialObj);
 
-    EXPECT_THAT(serviceDescription1.getSourceInterface(), Eq(Interfaces::INTERFACE_END));
+    EXPECT_TRUE(maybeService.has_error());
```
elBoberido, line 227: and here :)

elBoberido (commented):
Just some minor things

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -181,6 +150,34 @@ ServiceDescription::operator cxx::Serialization() const
                                       interface);
 }
 
+cxx::expected<ServiceDescription, cxx::Serialization::Error>
+ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept
+{
+    ServiceDescription deserializedObject;
+
+    std::underlying_type<Scope>::type scope = 0;
+    std::underlying_type<Interfaces>::type interfaceSource = 0;
```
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -231,7 +231,16 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
...
-            capro::ServiceDescription service(cxx::Serialization(message.getElementAtIndex(2)));
+            auto maybeService =
+                capro::ServiceDescription::deserialize(cxx::Serialization(message.getElementAtIndex(2)));
+            if (maybeService.has_error())
+            {
+                LogError() << "Deserialization failed when '" << message.getElementAtIndex(2).c_str()
+                           << "' was provided\n";
+                break;
+            }
+            auto service = maybeService.value();
```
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_posh/source/roudi/roudi.cpp
mossmaurice: I don't quite know, if we agreed on this before, but I like the idea a lot! Adapted the code and added a hint in `CONTRIBUTING.md`.

iceoryx_posh/test/moduletests/test_capro_service.cpp
mossmaurice: I like it. It even lead to  more stringent const correctness code.

CONTRIBUTING.md
```diff
@@ -126,6 +126,9 @@ information about logging and error handling.
     * Public members of structs and classes do not have the `m_` prefix
 * Namespaces in `lower_snake_case` : `my_namespace`
 * Aliases have a `_t` postfix : `using FooString_t = iox::cxx::string<100>;`
+* Objects created from a method returning a `cxx::optional<Foo>` shall be named `maybeFoo`
+* Objects created from a method returning a `cxx::expected<Foo, FooError>` shall
+contain the name `result` e.g. `fooResult`
```
elBoberido, line 131: This is not quite what we do, at least not what I recall. It is more like `auto getChunkResult = blup.getChunk(...);`

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -181,6 +150,38 @@ ServiceDescription::operator cxx::Serialization() const
                                       interface);
 }
 
+cxx::expected<ServiceDescription, cxx::Serialization::Error>
+ServiceDescription::deserialize(const cxx::Serialization& serialized) noexcept
+{
+    ServiceDescription deserializedObject;
+
+    using ScopeType = std::underlying_type<Scope>::type;
+    using InterfaceType = std::underlying_type<Interfaces>::type;
```
elBoberido, line 159: This is not quite correct, since the scope type is actually `Scope`, so this must be either `ScopeUnderlyingType` or shorter `ScopeUT`. Since this is confined to the class, I think the abbreviation is not too confusing.

iceoryx_posh/source/roudi/roudi.cpp
elBoberido: @mossmaurice done with #989

CONTRIBUTING.md
mossmaurice: True. What do you think about?
```suggestion
* Objects created from a method returning a `cxx::expected<Foo, FooError>` shall
contain the name `result` e.g. `somehtingResult`
```

iceoryx_posh/source/capro/service_description.cpp
mossmaurice: True indeed, done.

CONTRIBUTING.md
mossmaurice: Added an explicit example

elBoberido: I would add a concrete example like
```suggestion
* Objects created from a method returning a `cxx::expected<Foo, FooError>` shall
contain the name `result` e.g. `allGloryToTheHypnotoadResult` for a call to `allGloryToTheHypnotoad()`
```

elBoberido (approved):

dkroenke (commented):
Since there is `ServiceDescription::deserialize` introduced and the serialization constructor removed, it would be good to have an entry in the Changelog with a description of the changes in the public API.

elBoberido (approved):

dkroenke (approved):

PR #928, opened 2021-09-21
iox-#926 Replace relocatable pointer implementation

### LINKED ISSUE #926: Relocatable Pointers - Version 2.0
## Re-implement relocatable pointers as a building block for linked structures for shared memory

Relocatable pointers can be used to implement linked structures like lists and trees in a natural way (closely resembling raw-pointer implementations). The performance overhead is minor.

The benefit is that those structures can be copied efficiently (memcpy) and can be transmitted via shared memory.

## Detailed information

The existing relocatable pointer implementation is badly designed (base class etc.) and needs to be redone. Test coverage is also insufficient. It is currently not used but there is a use case in e.g. the prefix tree and any linked structure for shared memory we want to develop in the future.

1. It shall be entirely template-based (no inheritance).
2. It shall resemble STL smart pointers. 
3. There must be void specialization that will have a slightly different interface. 

It will be used first used in the prefix tree to demonstrate how to create a relocatable prefix tree (so it can be transmitted more efficiently via shared memory).

MatthiasKillat:
Signed-off-by: Matthias Killat <matthias.killat@apex.ai>

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #926


iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -0,0 +1,136 @@
...
+{
+  public:
+    /// @brief Construct from raw pointer.
+    relocatable_ptr(T* ptr = nullptr);
+    // {
+    //     m_offset = to_offset(ptr);
+    // }
+
+    /// @brief Construct from other relocatable pointer.
+    relocatable_ptr(const relocatable_ptr& other);
```
elBoberido, line 55: This contradicts the class description since it states that the class is trivially copyable but this ctor make it non trivially copyable

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -0,0 +1,136 @@
...
+    operator const T*() const;
+
+  private:
+    using offset_t = uint64_t;
+
+    // This is safe since it is equivalent to point to the relocatable pointer
+    // second byte of the relocatable pointer itself which we define to be illegal.
+    // (it is no reasonable use-case)
+    // Note that 0 is equivalent to point to the relocatable pointer itself (i.e. this).
+    static constexpr offset_t NULL_POINTER_OFFSET = 1;
```
elBoberido, line 110: Why not just 0? A relocatable pointer which points to itself doesn't make sense and some CPU have an optimized instruction for comparing to 0.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    offset_t m_offset;
```
elBoberido, line 112: Could you please move this to the end of the class and maybe separate it by another `private` access specifier to also separate the member from the methods?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+
+namespace iox
+{
+namespace rp
+{
+template <typename T>
+relocatable_ptr<T>::relocatable_ptr(T* ptr)
+{
+    m_offset = to_offset(ptr);
+}
```
elBoberido, line 30: nitpicking, could you please use the constructor member initializer list

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+
+template <typename T>
+relocatable_ptr<T>::relocatable_ptr(const relocatable_ptr& other)
+{
+    m_offset = to_offset(other.get());
```
elBoberido, line 35: same here

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+}
+
+template <typename T>
+relocatable_ptr<T>::relocatable_ptr(relocatable_ptr&& other)
+{
+    m_offset = to_offset(other.get());
```
elBoberido, line 41: and here

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+{
+    return from_offset(m_offset);
+}
+
+template <typename T>
+template <typename S>
+S& relocatable_ptr<T>::operator*()
+{
+    // not actually evaluated in the error case (compiler fails earlier since S = void leads to void&)
+    static_assert(!std::is_same<S, void>::value, "relocatable_ptr<void> does not support operator*");
```
elBoberido, line 83: do we really need a second template parameter for this check?

elBoberido (commented):
First round. Tests follow.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -0,0 +1,136 @@
...
+///       There should be no need for this, since as pass-around type
+///       regular pointers do the job just fine and do not incur
+///       the slight runtime overhead of a relocatable_ptr.
+///       There should be no memory overhead on 64 bit systems.
+template <typename T>
+class relocatable_ptr
+{
+  public:
+    /// @brief Construct from raw pointer.
+    relocatable_ptr(T* ptr = nullptr);
```
elfenpiff, line 49: Could you please add `noexcept` to every method

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    offset_t to_offset(const void* ptr) const;
+
+    T* from_offset(offset_t offset) const;
+};
+
+/// @brief Compare relocatable_ptr with respect to logical equality.
+/// @return true if rhs and lhs point to the same location, false otherwise
+template <typename T>
+bool operator==(const relocatable_ptr<T>& lhs, const relocatable_ptr<T>& rhs);
```
elfenpiff, line 124: Why do you implement this as a free function instead of a member method?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+
+template <typename T>
+typename relocatable_ptr<T>::offset_t relocatable_ptr<T>::to_offset(const void* ptr) const
+{
+    if (ptr == nullptr)
+    {
+        return NULL_POINTER_OFFSET;
+    }
+    auto p = reinterpret_cast<offset_t>(ptr);
+    return p - self();
```
elfenpiff, line 134: You could also write a fancy one liner here:
```cpp
return ( ptr ) ? reinterpret_cast<offset_t>(ptr) - self() : nullptr; 
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+#define IOX_HOOFS_RELOCATABLE_POINTER_RELOCATABLE_PTR_INL
+
+#include "iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp"
+
+namespace iox
+{
+namespace rp
+{
+template <typename T>
+relocatable_ptr<T>::relocatable_ptr(T* ptr)
```
elfenpiff, line 27: Why not use C++ member initializer list like:
```cpp
relocatable_ptr<T>::relocatable_ptr(T* ptr)
  : m_offset(to_offset(ptr)) {}
```

elfenpiff, line 27: Please add `inline` to every template implementation otherwise we will get double definition errors.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+{
+    m_offset = to_offset(ptr);
+}
+
+template <typename T>
+relocatable_ptr<T>::relocatable_ptr(const relocatable_ptr& other)
+{
```
elfenpiff, line 34: Initialization via member initializer list?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+    m_offset = to_offset(other.get());
+}
+
+template <typename T>
+relocatable_ptr<T>::relocatable_ptr(relocatable_ptr&& other)
+{
```
elfenpiff, line 40: Initialization via member initializer list?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
```diff
@@ -0,0 +1,164 @@
...
+}
+
+template <typename T>
+T* relocatable_ptr<T>::from_offset(offset_t offset) const
+{
+    if (offset == NULL_POINTER_OFFSET)
+    {
+        return nullptr;
+    }
+    return reinterpret_cast<T*>(offset + self());
```
elfenpiff, line 144: Here you could also write a fancy one liner:
```cpp
return (offset == NULL_POINTER_OFFSET) ? nullptr : reinterpret_cast<T*>(offset + self());
```

iceoryx_hoofs/test/moduletests/test_relocatable_ptr.cpp
```diff
@@ -58,67 +99,290 @@ class RelocatablePointer_test : public Test
...
+TYPED_TEST(Relocatable_ptr_typed_test, copyAssignmentOfNullptrWorks)
+{
+    using T = typename TestFixture::DataType;
+    // we cannot construct an actual object if T = void
+    // and it is not necessary fro most tests, we just need some non-nullptr
+    T* p = reinterpret_cast<T*>(NON_NULL_PTR);
+    iox::rp::relocatable_ptr<T> rp1;
+    iox::rp::relocatable_ptr<T> rp2(p);
+    rp2 = rp1;
+    EXPECT_EQ(rp1.get(), nullptr);
```
elfenpiff, line 147: It would be nice to add the automatic bool conversion operator to check if the relocatable contains a pointer or not.
```cpp
explicit operator bool() noexcept { return get() != nullptr; }
```

Then we could write code like
```cpp
if ( myRelocatable) {
  myRelocatable->dostuff = 5;
}
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
MatthiasKillat: Actually only to allow pointing to itself. There should be no need for this, except for in weird data structures where `next = this` indicates some kind of end.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
MatthiasKillat: Yes, I usually like having an order of type definitions, constants, variables and functions (separate for public and private sections).

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
MatthiasKillat: It is good practice to do so with these operators I think and necessary if one wants to have asymmetric operators: compare an S with a T and a T with S. Not directly the case here, or more precisely happens due to conversion: we can compare raw pointers with relocatable pointers. First I provided specific overloads for that (which needed to be non-member for the above reasons) but this causes issues with type deduction due to the implicit conversion (which I want to keep).

This is the historical reason. Another reason is basically summarized here https://embeddedartistry.com/fieldatlas/how-non-member-functions-improve-encapsulation/

There is also something like this in the usual books, but I did not look it up.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
MatthiasKillat: Sure, but I am fairly certain the compiler will create the same code regardless and I prefer not to crowded lines myself when it is not necessary.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
MatthiasKillat: Yes, should be done. However, for primitive types there is no performance gain to my knowledge.

MatthiasKillat: Yes, I forgot that. I really dislike the boilerplate needed and some of it purely due to the use of those inl files (or more precisely the out of class definition).

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
MatthiasKillat: Yes. But I am more interested in finding a way to avoid the conditional jump. Not sure what code the ternary operator generates.

iceoryx_hoofs/test/moduletests/test_relocatable_ptr.cpp
MatthiasKillat: This may already work due to the availability of implicit conversion to raw pointer (I think, have to check). I know for sure it will cause ambiguity if we add the operator bool, and I would not drop the operator T* for it (it is essential for convenience/usability).

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
elBoberido: Okay, in case of this
```c++
struct Foo{
relocatable_ptr<Foo>{this};
};
```
If offset 0 would encode a nullptr this wouldn't be possible anymore. So the issue is not that the pointer is referencing itself but it can have an offset of 0 in the class which contains the pointer. Could you make this more explicit in the comment.

elBoberido: Okay, in case of this
```c++
struct Foo{
relocatable_ptr<Foo>{this};
};
```
If offset 0 would encode a nullptr this wouldn't be possible anymore. So the issue is not that the pointer is referencing itself but it can have an offset of 0 in the class which contains the pointer. Could you make this more explicit in the comment?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
elBoberido: I guess this boils down to personal preference. It just happens that most of our classes have the members at the end.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+#include <cstdint>
+#include <type_traits>
+
+namespace iox
+{
+namespace rp
+{
+/// @brief Smart pointer type that allows objects using it to be trivially copyable.
+///        This applies only if it points to memory owned by the object itself (i.e.
```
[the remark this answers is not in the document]
MatthiasKillat: Yes we have: if we just remap the memory of the pointer and pointee with *the same offset*. But we must differentiate here. The `relocatble_pointer` needs a copy ctor in general and cannot be trivially copyable in general. We can only use memcpy in certain cases (pointer and pointee remapped/copied with the same offset = relocated).  I will adapt the description to reflect this.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relocatable_ptr.inl
MatthiasKillat: Yes, otherwise the class template instantiation will fail at class scope for `T=void` regardless of whether we use `operator*`, which is not what we want. By introducing S compilation will only fail if we actually try to use it. There may other ways to accomplish that, but to my knowledge they are all similarly relying on postponing the compilation of the operator by using an additional template argument.

They are also not shorter. E.g. `enable_if`is overkill here since we do not switch between alternatives (we just fail in the void case).

elfenpiff (approved):

Issue #987, opened 2021-12-10
Remove the requirement for `INVALID_STATE` for the cxx::expected


elBoberido:
## Brief feature description

Currently the `cxx::expected` requires to indication an invalid state after a move. For enums this means that there is an enum value called `INVALID_STATE`. But this can also be misused for other error cases and there is also the possibility to move the content out of the `expected`, which cannot be detected. The behavior of a moved `cxx::expected` should be symmetric to a moved content of the `cxx::expected`.

## Detailed information

The requirement of the `INVALID_STATE` enum value create a dependency inversion since the enum needs to implement the requirement of the `cxx::expected`. Since we still have to write tests to ensure that no `cxx::expected` with a moved content is used, it is superfluous to have a specific moved state for the `cxx::expected`. To simplify the code paths a move `cxx::expected` should not have an own moved state but should rely on the moved state of the content value. This makes the behavior of a `cxx::expected` symmetric to one with a moved content value.

PR #1004, opened 2022-01-03
Iox #987 remove the requirement of invalid state from cxx expected

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR remove the requirement of `INVALID_STATE` from the `cxx::expected` and adjust the code to make everything compile.
The removal of the `INVALID_STATE` enum value is done in #1005.

## References

- Relates to #987

### DIFF
--- a/iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -60,39 +60,54 @@ struct TestClass
     int m_b;
 };
 
-} // namespace
-
-namespace iox
+struct NonTrivialTestClass
 {
-namespace cxx
-{
-template <>
-struct ErrorTypeAdapter<TestClass>
-{
-    static TestClass getInvalidState()
+    NonTrivialTestClass(int a, int b)
+        : m_a(a)
+        , m_b(b)
     {
-        return TestClass(-1, -1);
     }
+
+    NonTrivialTestClass(const NonTrivialTestClass& other)
+    {
+        *this = other;
+    }
+    NonTrivialTestClass(NonTrivialTestClass&& other)
+    {
+        *this = std::move(other);
+    }
+
+    NonTrivialTestClass& operator=(const NonTrivialTestClass& rhs)
+    {
+        if (this != &rhs)
+        {
+            m_a = rhs.m_a;
+            m_b = rhs.m_b;
+            m_moved = rhs.m_moved;
+        }
+        return *this;
+    }
+
+    NonTrivialTestClass& operator=(NonTrivialTestClass&& rhs)
+    {
+        if (this != &rhs)
+        {
+            m_a = rhs.m_a;
+            m_b = rhs.m_b;
+            m_moved = rhs.m_moved;
+            rhs.m_moved = true;
+        }
+        return *this;
+    }
+
+    int m_a{0};
+    int m_b{0};
+    bool m_moved{false};
 };
 
-template <>
-struct ErrorTypeAdapter<std::string>
-{
-    static std::string getInvalidState()
-    {
-        return std::string("IAmInvalid");
-    }
-};
-} // namespace cxx
-} // namespace iox
-
-namespace
-{
 class expected_test : public Test
 {
 };
-
-using TestClassAdapter = iox::cxx::ErrorTypeAdapter<TestClass>;
 
 enum class TestError : uint8_t
 {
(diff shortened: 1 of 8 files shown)


iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -187,58 +202,64 @@ TEST_F(expected_test, ConstCreateLValueAndGetErrorResultsInCorrectError)
...
-TEST_F(expected_test, CreateWithErrorAndMoveCtorLeadsToInvalidState)
-{
-    auto sut = expected<int, TestClass>::create_error(22, 33);
-    auto movedValue{std::move(sut)};
-    IOX_DISCARD_RESULT(movedValue);
-    ASSERT_TRUE(sut.has_error());
-    EXPECT_THAT(sut.get_error(), Eq(TestClassAdapter::getInvalidState()));
+    ASSERT_FALSE(sutSource.has_error());
+    EXPECT_TRUE(sutSource.value().m_moved);
+    ASSERT_TRUE(!sutDestination.has_error());
```
mossmaurice, line 214: Why not?
```suggestion
    ASSERT_FALSE(sutDestination.has_error());
```

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -187,58 +202,64 @@ TEST_F(expected_test, ConstCreateLValueAndGetErrorResultsInCorrectError)
...
 
-TEST_F(expected_test, CreateWithErrorAndMoveAssignmentLeadsToInvalidState)
-{
-    auto sut = expected<int, TestClass>::create_error(44, 55);
-    auto movedValue = std::move(sut);
-    IOX_DISCARD_RESULT(movedValue);
-    ASSERT_TRUE(sut.has_error());
-    EXPECT_THAT(sut.get_error(), Eq(TestClassAdapter::getInvalidState()));
+    ASSERT_TRUE(sutSource.has_error());
+    EXPECT_TRUE(sutSource.get_error().m_moved);
```
mossmaurice, line 228: I don't like how the `expected` usage is showcased here. We typically use an `enum class` as `ErrorType`. An `enum class` can't have a custom move c'tor, right? How can one find out if a moved `expected` is still valid to use?  What we want is defined behaviour aka destructive move, I don't yet see how we can achieve this with the proposed changes in this PR. Please add a test case which uses an `enum class` as `ErrorType`.

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -334,22 +355,50 @@ TEST_F(expected_test, ErrorTypeOnlyCreateErrorLeadsToError)
...
+
+TEST_F(expected_test, ErrorTypeOnlyCreateValueWithoutValueMoveAssignmentLeadsToNoError)
+{
+    auto sutSource = expected<NonTrivialTestClass>::create_value();
+    auto sutDestination = std::move(sutSource);
+    EXPECT_FALSE(sutSource.has_error());
+    EXPECT_FALSE(sutDestination.has_error());
+}
+
+TEST_F(expected_test, ErrorTypeOnlyMoveCtorLeadsMovedSource)
```
mossmaurice, line 374:
```suggestion
TEST_F(expected_test, ErrorTypeOnlyMoveCtorLeadsToMovedSource)
```

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -334,22 +355,50 @@ TEST_F(expected_test, ErrorTypeOnlyCreateErrorLeadsToError)
...
+TEST_F(expected_test, ErrorTypeOnlyCreateValueWithoutValueMoveCtorLeadsToNoError)
 {
-    auto sut = expected<TestError>::create_error(TestError::ERROR2);
-    auto movedValue{std::move(sut)};
-    IOX_DISCARD_RESULT(movedValue);
-    ASSERT_TRUE(sut.has_error());
-    EXPECT_THAT(sut.get_error(), Eq(TestError::INVALID_STATE));
+    auto sutSource = expected<NonTrivialTestClass>::create_value();
+    auto sutDestination{std::move(sutSource)};
+    EXPECT_FALSE(sutSource.has_error());
```
mossmaurice, line 362: Please add a check which verifies that `sutSource` was moved.

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -334,22 +355,50 @@ TEST_F(expected_test, ErrorTypeOnlyCreateErrorLeadsToError)
...
+    EXPECT_FALSE(sutDestination.has_error());
+}
+
+TEST_F(expected_test, ErrorTypeOnlyCreateValueWithoutValueMoveAssignmentLeadsToNoError)
+{
+    auto sutSource = expected<NonTrivialTestClass>::create_value();
+    auto sutDestination = std::move(sutSource);
+    EXPECT_FALSE(sutSource.has_error());
```
mossmaurice, line 370: Same here, please add a check with verifies that `sutSource` was moved.

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -334,22 +355,50 @@ TEST_F(expected_test, ErrorTypeOnlyCreateErrorLeadsToError)
...
+
+    ASSERT_TRUE(sutSource.has_error());
+    EXPECT_TRUE(sutSource.get_error().m_moved);
+    ASSERT_TRUE(sutDestination.has_error());
+    EXPECT_FALSE(sutDestination.get_error().m_moved);
+    EXPECT_EQ(sutDestination.get_error().m_a, A);
+    EXPECT_EQ(sutDestination.get_error().m_b, B);
 }
 
 TEST_F(expected_test, ErrorTypeOnlyMoveAssignmentLeadsToInvalidState)
```
mossmaurice, line 389:
```suggestion
TEST_F(expected_test, ErrorTypeOnlyMoveAssignmentLeadsToMovedSource)
```

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
```diff
@@ -334,22 +355,50 @@ TEST_F(expected_test, ErrorTypeOnlyCreateErrorLeadsToError)
...
+    constexpr int B{223};
+    auto sutSource = expected<NonTrivialTestClass>::create_error(A, B);
+    auto sutDestination = std::move(sutSource);
+
+    ASSERT_TRUE(sutSource.has_error());
+    EXPECT_TRUE(sutSource.get_error().m_moved);
+    ASSERT_TRUE(sutDestination.has_error());
+    EXPECT_FALSE(sutDestination.get_error().m_moved);
+    EXPECT_EQ(sutDestination.get_error().m_a, A);
+    EXPECT_EQ(sutDestination.get_error().m_b, B);
```
mossmaurice, line 401: Could you please add a unit test cases which uses an `expected<EnumClassType>::create_error(..)`?

mossmaurice (changes requested):

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
elBoberido: Destructive move is a language feature and not a library feature and C++ will (most probably) never have destructive moves. Destructive moves means that the source is not accessible anymore after the move but with C++ that is not the case. All we can do is place moved from objects in a `valid but unspecified state`. In this case, it contains a moved inner type. This is not only valid but also a specified state of the `expected` and therefore there is no undefined behavior. The moved from inner type must of course also obey the rule to be placed in a `valid but unspecified state`. In our case we would of course specify the state since we do not want undefined behavior.

This is a test to check that the `expected` indeed moves the inner type instead of e.g. just doing a copy. Having a move test where a `enum class` is used as `ErrorType` has no value since PODs are copied on move and therefore the moved from value will remain the same and we cannot check if a move occurred.

A moved expected behaves exactly the same as e.g.
```cpp
Foo foo;
fooExp.and_then([&] (auto& f) { foo = std::move(f); });
```
or
```cpp
auto foo = std::move(fooExp.value());
```
or
```cpp
Bar make_bar(Foo& f) {
    return Bar(std::move(f));
}

auto b = make_bar(fooExp.value());
```
or
```cpp
auto err = std::move(fooExp.get_error());
```

In the end the `expected` is more of a transport type than a container type. It's not stored in classes and used as return value. The usage is therefore local an easy to reason about. No need to artificially make it more complex when there are still many ways which lead to an `expected` with a moved inner type. In the end, I didn't even had to change anything in the code while removing this requirement, except the `INVALID_STATE` value, which shows that this is not handled in the code anyway because we do not access moved from objects and also check this in our tests. Furthermore, it behaves exactly the same as a moved `variant`.

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
elBoberido: Not possible, since `sut` has no value which can be checked.

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
elBoberido: Same here.

FerdinandSpitzschnueffler (dismissed):

FerdinandSpitzschnueffler (approved):

iceoryx_hoofs/test/moduletests/test_cxx_expected.cpp
mossmaurice: > Destructive move is a language feature and not a library feature and C++ will (most probably) never have destructive moves.

You're right. What I meant was a "invalidating move". We already do this throughout the code base and it's important e.g. when classes handle resources. 

> The moved from inner type must of course also obey the rule to be placed in a `valid but unspecified state`

My philosophy so far is, that with iceoryx we go one step further and for non-POD types go into a "invalid state and specified state".

> How can one find out if a moved `expected` is still valid to use?

What I realised with the `enum class`'s is, that it's not actually a dangerous thing because if a user accesses a moved-from `expected` which contains an error, he can only mistake the error and not by mistake take the happy path. 

>  I didn't even had to change anything in the code while removing this requirement

That's a good finding and very smelly. Thanks for taking the time to remove the `INVALID_STATE`!

mossmaurice (approved):

Issue #590, opened 2021-03-02
Improve modularity by creating architecure guidelines


mossmaurice:
## Brief feature description

iceoryx is developed in several CMake packages that aim to be independent. Current state is there are cross dependencies between each of them and no architecture guidelines have been created so far.

## Detailed information

Analysed iceoryx repository with [Lattix Architect](https://www.lattix.com/products-architecture-issues/). [Attached](https://github.com/eclipse-iceoryx/iceoryx/files/6071770/iceoryx_b929aa_Architecture.Metrics_Metrics.Architecture.xlsx) are the the dependecy structure matrix and the architecture metrixs of Lattix (commit b929aa8d5022471a318cdc8e6994f1cdd6aa3850).

### General

* [x] Analyse Lattix
  * [x] Dependecy structure matrix (DSM)
  * [x] Architectual metrics
  * [x] Conceptual architecture
* [x] Namespaces & CMake package/library refactoring
* [x] Add CMake target for quality level 5/playground features and move "maybe obsolote" constructs there
  * [x] Add script to compare used namespaces to a whitelist in binary (`nm -C myBinary -f posix` with C++filt for demangling) to make sure no experimental/playground code is baked into releases
* [ ] RouDi modularisation #539
* [x] Create architecture documentation
  * [x] PlantUML or plain ASCII art? 
  * [x] Package view
  * [x] Layer view
* [ ] Summarise guidelines and write a CI checker
  * [ ] Remove `iox::helper` namespace
* [ ] Create `iceoryx_posh_core` CMake target and address the `@todo #590` in the CMake files
* [ ] Remove unused files and move quality level 2 files to a separate CMake package

### Lattix Architect findings

* [x] ~~Move POSH specific errors from iceoryx_utils to iceoryx_posh~~ Moved to #1099 
* [x] `roudi_gtest.hpp` includes `iceoryx_posh/internal/roudi_environment/roudi_environment.hpp`
* [ ] Move `RouDiEnvironment` to test folder & namespace
* [ ] `iceoryx_hoofs` does not have a clear architecture and shall be better structured

mossmaurice:
FYI @dlo-86 
[Structure101](https://structure101.com) would be another option besides Lattix to analyse the architecture. However, I've had a detailed look at Lattix and found the DSM and its metrics quite helpful.

elBoberido:
@mossmaurice do you think this is related to iceoryx_hoofs?

mossmaurice:
[...]

PR #1007, opened 2022-01-05
iox-#861 Fix TypedUniqueId error handler usage and refactor to UniquePortId

### LINKED ISSUE #861: Fix error handling of TypedUniqueId and refactor it to UniquePortId
## Required information

The `TypedUniqueId` ctor, `setUniqueRouDiId` and `getUniqueRouDiId` are calling the error handler with an empty lambda as error callback and `ErrorLevel::FATAL` or `ErrorLevel::MODERATE`. In case of `ErrorLevel::FATAL` this is a clear mistake, since nothing happens due to the empty lambda.

Usually `nullptr` is used when no custom error callback is defined. This is implicitly casted to an empty `std::function`.

@elfenpiff @MatthiasKillat I would suggest to also use an `nullptr` for a quick fix and the decide if this needs a bigger refactoring. What do you think?

Additionally, the `TypedUniqueId` was also used once via an alias by `using UniquePortId = TypedUniqueId<BasePort>`. Since the `TypedUniqueId` requires a `RouDi` ID, it is quite specific and therefore it will be refactored to `UniquePortId`.

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

Currently, the `TypedUniqueId` set's an empty lambda as error callback. This is an error since no error handling will be done. At other places a `nullptr` is used for the error callback when a non fatal error level is desired but no custom error callback is needed. The `nullptr` is implicitly casted to an empty `std::function`. This PR adjusts the `TypedUniqueId` error handler usage to the remaining code base.

Additionally it turned out that a small refactoring of the  `TypedUniqueId` makes sense. Since it was only used as for the `UniquePortId` and also contains the unique RouDi ID it was renamed to `UniquePortId`.

## References

- Closes #861


iceoryx_posh/source/popo/building_blocks/typed_unique_id.cpp
```diff
@@ -29,8 +30,7 @@ void setUniqueRouDiId(const uint16_t id) noexcept
 {
     if (hasDefinedUniqueRouDiId)
     {
-        errorHandler(
-            Error::kPOPO__TYPED_UNIQUE_ID_ROUDI_HAS_ALREADY_DEFINED_UNIQUE_ID, [] {}, ErrorLevel::MODERATE);
+        errorHandler(Error::kPOPO__TYPED_UNIQUE_ID_ROUDI_HAS_ALREADY_DEFINED_UNIQUE_ID, nullptr, ErrorLevel::MODERATE);
```
elBoberido, line 33: This is quite error prone. Shall I create an issue to swap the order of the error callback and error level parameter?

This would then become
```cpp
errorHandler(Error::kPOPO__TYPED_UNIQUE_ID_ROUDI_HAS_ALREADY_DEFINED_UNIQUE_ID, ErrorLevel::MODERATE);
```
I guess it is more common to just use a non fatal error level than to use a custom error callback.

Another option would be the builder pattern
```cpp
errorHandler(Error::kPOPO__TYPED_UNIQUE_ID_ROUDI_HAS_ALREADY_DEFINED_UNIQUE_ID)
    .withErrorLevel(ErrorLevel::MODERATE)
    .customErrorCallback([]{
        std::cout << "You are doomed" << std::endl;
        std::terminate();
    });
```

What do you think?

iceoryx_posh/source/roudi/application/roudi_app.cpp
```diff
@@ -92,6 +92,11 @@ RouDiApp::RouDiApp(const config::CmdLineArgs_t& cmdLineArgs, const RouDiConfig_t
     {
         popo::internal::setUniqueRouDiId(cmdLineArgs.uniqueRouDiId.value());
     }
+    else
+    {
+        /// @todo iox-#861: we need a default RouDi ID. Which one to use
+        popo::internal::setUniqueRouDiId(1);
```
elBoberido, line 98: I somehow feel this should not be necessary. Can we set the `UniqueRouDiId` to a default value and don't call the error handler if it is not set to get rid of all the `popo::internal::setUniqueRouDiId` in the tests?

Currently it runs with `0` when not explicitly set and works fine due to the mistake with the error handler. I do not see a reason why we need to enforce the setting when we have a working default value.

@elfenpiff what do you think?

iceoryx_examples/singleprocess/single_process.cpp
```diff
@@ -115,6 +115,9 @@ int main()
     // set the log level to error to see the essence of the example
     iox::log::LogManager::GetLogManager().SetDefaultLogLevel(iox::log::LogLevel::kError);
 
+    /// @todo iox-#861: we need a default RouDi ID. Which one to use
```
elfenpiff, line 118: Thats easy we use the digit sum of the sum of the ASCII codes of the sentence `all glory to the hypnotoad`. But I can be also convinced to use the product of the ASCII codes of that sentence.

iceoryx_posh/source/popo/building_blocks/typed_unique_id.cpp
elfenpiff: I prefer the builder pattern. It is the most expressive approach.

iceoryx_posh/source/roudi/application/roudi_app.cpp
elfenpiff: @elBoberido I completely agree. But the default value should be calculated as described in the comment of the single process example.

iceoryx_examples/singleprocess/single_process.cpp
elBoberido: It must be more sophisticated. We also need to take the position of the moon and other celestial objects into account.

iceoryx_posh/source/roudi/application/roudi_app.cpp
elBoberido: @elfenpiff I would like to refactor the `TypedUniqueId` a bit. Instead of having a `setRouDiId` free function I would add a `TypedUniqueId::setDomainId`. This static function will call the error handler if there is already an instance of the `TypedUniqueId`. Therefore it must be called only once and before `TypedUniqueId` is used. For tests, the error handler can be overwritten. What do you think.

elBoberido: @elfenpiff forget my suggestion. Since `TypedUniqueId` is a template, one cannot use my proposal. What do you think of the code I just committed?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/typed_unique_id.hpp
```diff
@@ -37,8 +40,7 @@ namespace internal
 /// @param[in] id the unique id which you would like to set
 void setUniqueRouDiId(const uint16_t id) noexcept;
 
-/// @brief Sets the RouDi id to an undefined state
-void unsetUniqueRouDiId() noexcept;
```
elfenpiff, line 41: Please ad some doxygen comment.

elfenpiff (approved):

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
```diff
@@ -0,0 +1,89 @@
...
+    /// @brief constructor which creates an invalid id
+    UniquePortId(InvalidPortId_t) noexcept;
+
+    bool isValid() const noexcept;
+
+    /// @brief Has to be set on roudi startup so that a unique roudi id is set
+    ///         for all newly generated unique ids. If you call it when a unique
+    ///         id is already set an error is generated in the errorHandler.
+    /// @param[in] id the unique id which you would like to set
+    static void setUniqueRouDiId(const uint16_t id) noexcept;
```
mossmaurice, line 67:
```suggestion
    static void setUniqueBaseId(const uint16_t id) noexcept;
```

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    /// @brief Getter for the unique roudi id
+    /// @return value of the unique roudi id
+    static uint16_t getUniqueRouDiId() noexcept;
```
mossmaurice, line 71: For me it's a bit confusing that a class `UniquePortId` has a method `getUniqueRouDiId`. Could we generalise this to something like?

```suggestion
    static uint16_t getUniqueBaseId() noexcept;
```

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
```diff
@@ -0,0 +1,89 @@
...
+{
+namespace popo
+{
+/// @brief Struct to signal the constructor to create an invalid id
+struct InvalidPortId_t
+{
+};
+constexpr InvalidPortId_t InvalidPortId = InvalidPortId_t();
+
+/// @brief Unique ID depending on a type. If you would like to assign different
```
mossmaurice, line 37: Could you elaborate the usage of the RouDi ID a bit further here?

iceoryx_posh/source/popo/building_blocks/unique_port_id.cpp
```diff
@@ -0,0 +1,83 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp"
+#include "iceoryx_posh/iceoryx_posh_types.hpp"
+
+namespace iox
+{
+namespace popo
+{
+// in the future this should probably be moved to RouDi but currently it is only used for the ports
```
mossmaurice, line 25: Is this a specific class only to be used for RouDi and its ports or is it something generic? Please remove the comment if it will stay here.

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
```diff
@@ -26,160 +28,155 @@ using namespace ::testing;
 using namespace iox::popo;
 using namespace iox::cxx;
 
-TEST(TypedUniqueId_RouDiId, SettingTheRouDiIdWorks)
+TEST(UniquePortId_test, SettingTheRouDiIdWorks)
 {
     ::testing::Test::RecordProperty("TEST_ID", "473467bf-1a6f-4cd2-acd8-447a623a5301");
     uint16_t someId = 1243u;
```
mossmaurice, line 34: Do `U` fancy capital literals? :smirk:

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
elBoberido: Not sure about this. I'd rather label it with the content and the content is specified to be the unique RouDi ID.

Let's wait for a third (probably totally different) opinion :laughing:

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
elBoberido: I try ... please close if you are satisfied

iceoryx_posh/source/popo/building_blocks/unique_port_id.cpp
elBoberido: You are correct. It will stay here and also for a good reason. It is not generic but needed for the chunk header to set the origin id. This is in turn is then used for the ROS2 guid for publisher

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
elBoberido: Well, I find them horrible https://github.com/eclipse-iceoryx/iceoryx/discussions/617 ... but it seems almost nobody agrees and I have to live this nightmare day by day 😜

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
elfenpiff: I would leave it as it is.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
```diff
@@ -0,0 +1,92 @@
...
+                                         cxx::newtype::MoveAssignable>
+{
+  public:
+    using ThisType::ThisType;
+
+    /// @brief The constructor creates an id which is greater than the
+    ///         previous created id
+    UniquePortId() noexcept;
+
+    /// @brief Constructor which creates an invalid id
```
elfenpiff, line 59: This again smells somehow like a nullable type. Do we really need this or is it possible to pack the `UniquePortId` into an `optional` when it is not valid.
What is the use case for having an invalid `UniquePortId`?

We even could get rid of `isValid` and with that I am sure maybe even of some error handling or additional tests?!

iceoryx_posh/include/iceoryx_posh/mepoo/chunk_header.hpp
```diff
@@ -161,7 +161,7 @@ struct ChunkHeader
     uint8_t m_reserved{0};
     // currently just a placeholder
     uint16_t m_userHeaderId{NO_USER_HEADER};
-    UniquePortId m_originId{popo::InvalidId};
+    popo::UniquePortId m_originId{popo::InvalidPortId};
```
elfenpiff, line 164: Shouldn't a chunkheader have always a valid port id? I think it may makes more sense to get rid of the default value here, remove the default ctor from the unique port id (if there is one) and force to set it here correctly.

If the chunkheader can have an invalid port id we should turn this into a `optional<popo::UniquePortId>` instead.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
elBoberido: @mossmaurice if you do not have strong feelings regarding the naming, I would keep it as is

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
elBoberido: @elfenpiff I agree with you, but if you don't mind, I would create a separate issue for this. The PR is already quite big and this change would impact many other files. Are you okay with doing it in a follow up PR?

iceoryx_posh/include/iceoryx_posh/mepoo/chunk_header.hpp
elBoberido: It is also used in the `TypedMemPool`, which doesn't have a port ID. Therefore an `optional` would be the best solution

iceoryx_posh/include/iceoryx_posh/mepoo/chunk_header.hpp
```diff
@@ -142,7 +142,7 @@ struct ChunkHeader
     template <typename T>
     friend class popo::ChunkSender;
 
-    void setOriginId(UniquePortId originId) noexcept;
+    void setOriginId(popo::UniquePortId originId) noexcept;
```
elfenpiff, line 145: Please add const to the argument.

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
```diff
@@ -26,160 +28,155 @@ using namespace ::testing;
...
+    iox::popo::UniquePortId::setUniqueRouDiId(someId);
+    // now we know that setUniqueRouDiId was called and therefore the error handler must also be called
+    ASSERT_TRUE(detectedError.has_value());
+    ASSERT_TRUE(detectedErrorLevel.has_value());
+    EXPECT_THAT(detectedError.value(),
+                Eq(iox::Error::kPOPO__TYPED_UNIQUE_ID_ROUDI_HAS_ALREADY_DEFINED_CUSTOM_UNIQUE_ID));
+    EXPECT_THAT(detectedErrorLevel.value(), Eq(iox::ErrorLevel::SEVERE));
+
+    // reset unique RouDi ID
+    iox::popo::UniquePortId::setUniqueRouDiId(iox::roudi::DEFAULT_UNIQUE_ROUDI_ID);
```
elfenpiff, line 70: Could you turn those `TEST` into test fixtures `TEST_F` and add this line to `void TearDown()`. Then it does not get forgotten and as it seems it is part of the cleanup in every test.

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
```diff
@@ -26,160 +28,155 @@ using namespace ::testing;
...
     a = b;
     EXPECT_THAT(a, Eq(b));
 }
 
-TYPED_TEST(TypedUniqueId_test, MoveConstructorSetsSameID)
+TEST(UniquePortId_test, MoveConstructorSetsSameID)
 {
     ::testing::Test::RecordProperty("TEST_ID", "2abb1afa-094a-4d28-a3cc-328212ad2f7b");
-    typename TestFixture::UniqueIDType a;
+    UniquePortId a;
```
elfenpiff, line 105: Since we would like to not have undefined behavior and an undefined state would it make sense to test here that a moved `UniquePortId` has a defined state. This means that the uint64_t value of a moved object is INVALID_ID or something?

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
```diff
@@ -26,160 +28,155 @@ using namespace ::testing;
...
     EXPECT_TRUE(a.isValid());
 }
 
-TYPED_TEST(TypedUniqueId_test, InvalidIdIsInvalid)
+TEST(UniquePortId_test, InvalidIdIsInvalid)
 {
     ::testing::Test::RecordProperty("TEST_ID", "d0576b8d-65d2-4b53-88c0-05078f434a41");
-    typename TestFixture::UniqueIDType a(InvalidId);
```
elfenpiff, line 182: Maybe we should test that:
* Moving valid id to invalid id makes the destination valid
* Moving invalid id to valid id makes the destination invalid
* Moving id makes the source always invalid

For the move constructor and assignment operator.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/unique_port_id.hpp
elBoberido: After some thinking a came to the conclusion that we might not want to use an optional here. The `UniquePortId` is used in the `ChunkHeader` and we must be very careful when member of this class changes and always increase the version number. There could be recorded data which needs to be replayed and subtle bugs might occur when the version match but the binary layout and semantics change. With an complex type like an `optional` it might be possible to accidentally change the binary layout of the `ChunkHeader` without noticing it by just refactoring the `optional`.

In this special case, I would keep an invalid `UniquePortId`. To not break source compatibility I would also keep the `UniquePortId` instead of using the underlying type but I would also add some tests to check to notice when the underlying type changes or the `UniquePortId` gains new members.

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
elBoberido: I would argue that a moved `UniquePortId` does not need to be invalid, in the end it is just a `uint64_t` which is also not invalid after move. It is still in a defined state since the underlying value is not touched

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
elBoberido: Don't we already have this tests in the `NewType` and wouldn't this somewhat negate the benefits of the pattern. After all, it is a wrapper around a POD which does not have an undefined state after move.

iceoryx_posh/test/moduletests/test_popo_unique_port_id.cpp
elBoberido: Well, I count only two tests but it's late and I might overlook something ;)

I would keep it as is but use a GenericRAII instead, since it is only used where it is explicitly set before.

iceoryx_posh/test/moduletests/test_mepoo_chunk_header.cpp
```diff
@@ -23,6 +23,8 @@
 
 #include "test.hpp"
 
+#include <type_traits>
```
mossmaurice, line 26: Please change the year in the copyright header

mossmaurice (commented):
LGTM, found a few nitpicks.

iceoryx_posh/source/popo/building_blocks/typed_unique_id.cpp
elBoberido: I did not create an issue for this since @mossmaurice tackles this as part of #590 ... oh, I guess the name of that issue should be changed :)

mossmaurice (dismissed):

iceoryx_posh/source/mepoo/chunk_header.cpp
```diff
@@ -23,6 +23,8 @@ namespace iox
 {
 namespace mepoo
 {
+constexpr uint8_t ChunkHeader::CHUNK_HEADER_VERSION;
```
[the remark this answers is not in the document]
elBoberido: oopsi, I mixed the file up and changed the `chunk_header.hpp`

elfenpiff (approved):

mossmaurice (approved):

Issue #743, opened 2021-04-15
Improve documentation and website for v2.0


mossmaurice:
# Brief feature description

### Website/Markdown settings

* [x] Open Bugzilla issue to "enforce https"
* [x] Create favicon for website
* [x] Switch to `mike` v1.0 to avoid duplication for aliases
* [x] ~~Add `.markdownlint.yaml` file~~ Moved to #1295

### Common

* [x] Improve utils documentation (in the long run: find more suitable name for utils, e.g. safe building blocks), see also #468
* [x] Add overview graphic introducing the main players (RouDi, Runtime/user apps, shared Memory, data packages/topics) in overview.md (see suggestions [here](https://github.com/eclipse-iceoryx/iceoryx/pull/646#discussion_r604950011) and [here]
(https://github.com/eclipse-iceoryx/iceoryx/pull/646#discussion_r604960591))
  * [x] Include GIF on iceoryx.io
  * [x] Add flow diagram for WaitSet and Listner (draw.io?)
* [x] Add link to CONTRIBUTING.md?
* [x] ~~Add integration hints to other platforms~~ Moved to #1295 
* [x] Describe introspection client in more detail, maybe refer to ROS 2 use-case?
* [x] Validate all the iceoryx articles at https://iceoryx.io/ -> Related to https://github.com/eclipse-iceoryx/iceoryx/issues/743
* [x] Double check: are the iceoryx integration tests covering all the examples?
* [x] Is the changelog is complete?
* [x] ~~Add separator for the API change sections~~ Moved to #1295 

### New articles/examples

* [x] Add chunk header example
* [x] Add example sending complex data types
* [x] Add section on CI setup (e.g. Axivion)
* [x] Add pub/sub matching criteria table
* [x] Add architecture article under concepts
* [x] Replace all GitHub link with fixed version `https://github.com/eclipse-iceoryx/iceoryx/tree/v2.0.0/`
* [x] ~~How to write integration tests~~ Moved to #1295 
* [x] Record asciicast for
  * [x] icecrystal (outdated)
  * [x] icedelivery
  * [x] icediscovery
  * [x] icediscovery_in_c
  * [x] ~~icedocker~~
  * [x] ~~iceensemble (outdated)~~
  * [x] ~~iceperf~~
  * [x] maybe: request_response - the waitset/listener example is described, recorded the basic example
  * [x] request_response_in_c

### Obligatory fixes

* [x] Fix doxygen function links
* [x] Fix broken links
[...]

Issue #73, opened 2020-03-23
Write ReadMe for Iceoryx Utils


elfenpiff:
## Brief feature description

Readme introducing the constructs which are used in iceoryx

## Detailed information

If someone implements new features they have to know which building blocks are already available and can be used. Additionally, we should explain our best practices.

PR #1015, opened 2022-01-11
iox-#743 Add QoS policy article

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Create "Concept" section in navigation
* Add QoS policies article

## References

- Closes partly #743 
- Comes from [#73](https://github.com/ros2/rmw_iceoryx/pull/73/files)


elBoberido (approved):
Looks good. You can apply my suggestion or merge it like it is

elBoberido:
Oh, please link the issue in the sidebar. I know that the issue will be closed, but unfortunately the PR won't show up in the linked PR section of the issue and then it's a hassle to find all related PRs.

budrus:
@mossmaurice queue size and history settings could also be mentioned in the QoS policies. Such settings also belong to QoS in DDS and in the Cylone + iceoryx combination the DDS history QoS settings are also mapped to the iceoryx history settings

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,30 @@
+# QoS policies
+
+Unlike other middlewares like for example [DDS used in ROS 2](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html) iceoryx tries to keep things simple and offers only basic Quality of Service (QoS) settings. The QoS settings are called options and can be used to optimize and tailor the communication.
```
MatthiasKillat, line 3: I do not like the wording "Unlike <others> iceoryx tries ... ". This is is too much subjective opinion and we should avoid it for this reason. 

It is more like state what we do and what it can be compared to, whether it is good or bad, complex or simple is for others to judge in my opinion. Furthermore I would not write "tries to", either we do it or we don't. There is no try ...

In short: this should be more matter of fact (less opinion) and no try.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,30 @@
...
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings, which are also available in DDS, are:
```
MatthiasKillat, line 7: They are not directly available in DDS I would say, it is not immediately obvious how the concepts/options map to each other. I would leave DDS out here.

In DDS we have a `history depth` which roughly corresponds  to our `queueCapacity` on subscriber side. However, in DDS this is also available on the writer, which roughly corresponds to our `historyCapacity`. It also depends on the setting Durability setting, late joining readers in DDS only get values with `Non-Volatile` durability, e.g. `TRansient Local`.

So it is not straightforward to map the concepts, and for this reason I would leave it out or phrase it differently and mention that similar settings exist for DDS (durability and history depth QoS).

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,30 @@
...
+
+* `PublisherOptions::historyCapacity`: This will enable late-joining subscribers to read the n last samples
```
MatthiasKillat, line 9: I would use the present tense here. "This enables late-joining subscribers to request the last n samples" Or most recent n samples. n of course being the historyCapacity, but it does probably not require mentioning.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,30 @@
...
+* `SubscriberOptions::queueCapacity`: Size of the subscriber queue where samples are stored before they are passed to the user
+* `SubscriberOptions::historyRequest`: The number of samples a late-joining subscriber will request from a publisher
```
MatthiasKillat, line 11: It may be obvious but still: 
A subscriber can at most request historyCapacity samples, i.e. if the publisher buffers less samples the subscriber will be out of luck. I am not sure what happens if it tries to request more than the publisher supports. I think they still match but the subscriber gets less samples. This may be surprising for a user and worth noting for this reason.

mossmaurice: Added.

doc/website/concepts/qos-policies.md
mossmaurice: Ok, I reformatted the list to a table.

doc/website/concepts/qos-policies.md
mossmaurice: I personally would prefer this subjective comparison here as DDS has and will influence iceoryx heavily. Removed the try.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,36 @@
...
+Two criteria have to be fulfilled in order for a publisher and a subscriber to be connected.
+
+ 1. Same `capro::ServiceDescription`
+ 2. Matching `SubscriberTooSlowPolicy` and `QueueFullPolicy` in `PublisherOptions` and `SubscriberOptions`
+
+### Compatible policies
+
+| `SubscriberTooSlowPolicy` | `QueueFullPolicy`     | Result   |
+|---------------------------|-----------------------|----------|
+| `WAIT_FOR_SUBSCRIBER`     | `BLOCK_PUBLISHER`     | Match    |
```
MatthiasKillat, line 33: Maybe it would be useful to document the resulting behavior here. E.g. in the first row the publisher blocks and waits for the subscriber to finish processing - no data is dropped.

If this is explained elsewhere this can be ignored.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,36 @@
+# QoS policies
+
+Unlike other middlewares like for example [DDS used in ROS 2](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html)
```
elBoberido, line 3: This reads a bit weird. Depending on the emphasis one could read that iceoryx is like DDS. How about
```md
Unlike other middlewares, e.g. DDS used in ROS 2, iceoryx ...
```
or
```md
Compared to other middlewares like e.g. DDS used in ROS 2, iceoryx ...
```

Or maybe just remove the DDS part.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,36 @@
...
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings are similar to the respective DDS setting:
+
+| DDS                       | iceoryx                             | Explanation                                                                                                                    |
+|---------------------------|-------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
+| History depth of writer   | `PublisherOptions::historyCapacity` | This enables late-joining subscribers to request the n last samples                                                            |
+| History depth of reader   | `SubscriberOptions::queueCapacity`  | Size of the subscriber queue where samples are stored before they are passed to the user                                       |
+| -                         | `SubscriberOptions::historyRequest` | The number of samples a late-joining subscriber will request from a publisher, must be equal or smaller than `historyCapacity` |
```
elBoberido, line 15: Why the comparison to DDS and not e.g. AUTOSAR? This is totally unrelated to iceoryx and confuses more than it helps. Please keep only the iceoryx related settings

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,36 @@
...
+
+The history depth of DDS depend on the durability setting. Late-joining readers in DDS
+only get values with `Non-Volatile` durability, e.g. `Transient Local`.
```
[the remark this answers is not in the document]
MatthiasKillat: I agree that we should not explain (or even mention) the DDS concepts here. At most there existence (but even that is a stretch). At the level it can be explained here (not much detail) it does not really help and is not relevant to an iceoryx user I would argue.

If at one point we want a more detailed comparison we should devote a whole document to it. While probably not needed, the reader can just skip it if he is not interested.

doc/website/concepts/qos-policies.md
budrus: @elBoberido AUTOSAR has nothing that could be called similar. But I'm with you, as things are slightly different anyway, we can focus on iceoryx. The QoS mapping from iceoryx to DDS is a non-trivial topic that can be discussed when iceoryx is combined with a DDS stack like Cyclone DDS. Also the functionality we provide with iceoryx often ends up in a combination of history and durability QoS in DDS. This would make it even more complex to describe  the right mapping here

mossmaurice: Ok, I agree. Comparing can be even dangerous if it is easily misunderstood. Removed the DDS parts.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
...
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings are:
+
+| Option                              | Explanation                                                                                                                    |
+|-------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
+| `PublisherOptions::historyCapacity` | This enables late-joining subscribers to request the n last samples                                                            |
+| `SubscriberOptions::queueCapacity`  | Size of the subscriber queue where samples are stored before they are passed to the user                                       |
+| `SubscriberOptions::historyRequest` | The number of samples a late-joining subscriber will request from a publisher, must be equal or smaller than `historyCapacity` |
```
elBoberido, line 15: I'm not quite sure if it really must be smaller since the subscriber might not know the `historyCapacity`. I think if will silently accept if the request is larger and only delivers the available samples, similar to when the publisher is started and does not yet have enough samples.

I'm wondering what happens when the `historyRequest` is larger than the `queueCapacity`?

elBoberido (commented):
Looks good to me. Please check the explanation for `historyRequest` and also what happens when `historyRequest` is larger than `queueCapacity`.

doc/website/concepts/qos-policies.md
mossmaurice: Correct, it silently accepts it, but there is a warning in RouDi:

```
[Warning]: Chunk history request exceeds history capacity! Request is 20. Capacity is 10.
```

Re-phrased to "should be".

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
...
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings are:
+
+| Option                              | Explanation                                                                                                                                       |
+|-------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------|
+| `PublisherOptions::historyCapacity` | This enables late-joining subscribers to request the n last samples                                                                               |
+| `SubscriberOptions::queueCapacity`  | Size of the subscriber queue where samples are stored before they are passed to the user                                                          |
+| `SubscriberOptions::historyRequest` | The number of samples a late-joining subscriber will request from a publisher, should be equal or smaller than `historyCapacity` otherwise capped |
```
elBoberido, line 15: I had a brief look to check what happens when `historyRequest` is larger than `queueCapacity` and it seems there is either
- safe overflowing for `QueueFullPolicy::DISCARD_OLDEST_DATA`, i.e. latest gratest
- newer samples are discarded when  `QueueFullPolicy::BLOCK_PUBLISHER`, i.e. lost samples with a gap
Could you please check if I'm right?

For the first case I think it's the right behavior. For the second one I would be surprised. Since RouDi puts the history into the queue, we cannot block RouDi. In order to prevent a gap in the samples, I would suggest to either:
- catch this in the runtime when a subscriber is created
    - either saturate `historyRequest` value to `queueCapacity` and print a warning
    - or abort the creation of the subscriber
- catch it in ChunkDistributor::tryAddQueue and push only the newest samples to the queue to make it full
    - this should also be done as optimization for  `QueueFullPolicy::DISCARD_OLDEST_DATA` since we know that there will be an overflow

@budrus what do you think?

elBoberido (commented):
Looks good but I think I discovered a bug

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
...
+iceoryx keeps things simple and offers basic Quality of Service (QoS) settings. The QoS
+settings are called options and can be used to optimize and tailor the communication.
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings are:
+
+| Option                              | Explanation                                                                                                                                       |
+|-------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------|
+| `PublisherOptions::historyCapacity` | This enables late-joining subscribers to request the n last samples                                                                               |
```
elfenpiff, line 13: This is correct but the implementation is "kaputt".
See #780 

Here we have to also explain the contract in detail. Is the subscriber subscribing to a publisher or a service - I think service is here the right answer but our implementation assumes publisher. We have to also define what a service is.

To summarize the problem:
1. Single publisher - everything works
2. Multi publisher broken - you get the last `n` samples of every publisher so this means for `m` publishers you get currently in the worst case `m * n` samples in random order, not in the order they were published on the service
3. Multi publisher when publisher called stop offer or is removed - If the publisher who send the last n samples is removed or called stop offer you will never get the last `n` samples since they are vanished. You get arbitrary `n` samples at best or nothing.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
...
+| `SubscriberOptions::queueCapacity`  | Size of the subscriber queue where samples are stored before they are passed to the user                                                          |
+| `SubscriberOptions::historyRequest` | The number of samples a late-joining subscriber will request from a publisher, should be equal or smaller than `historyCapacity` otherwise capped |
+
+For more information about the options see the corresponding example [`iceoptions`](../getting-started/examples/iceoptions.md).
+
+## Publisher subscriber matching criteria
+
+Two criteria have to be fulfilled in order for a publisher and a subscriber to be connected.
+
+ 1. Same `capro::ServiceDescription`
```
elfenpiff, line 23: Question. Is it possible to have two publishers with the same service description and different policies? So that we have in the end two identical named services to which subscribers can subscribe?

If so then our service handling is broken as well since the QoS is then definitely a part of the service description.

elfenpiff (commented):
@mossmaurice Oooh I think you discovered a hornet nest in iceoryx, knocked on it and asked if someone is home and the hornets are now angry on all of us for disturbing them. ;)

doc/website/concepts/qos-policies.md
MatthiasKillat: If we say request 5 samples (and there are 5 available) and the queue can only hold three I would think we should store the latest 3 (i.e. sample 3 to 5). This is just happens automatically if we are requesting all and discard the first 2. 

Now this only holds in isolation, for multi-publish we may receive interleaved other samples of a different publisher and hence lose some of the three samples. This is not really tied to history but how multi-publish currently works (see above).

This is all consistent with the `QueueFullPolicy::DISCARD_OLDEST_DATA` I would think (as does @elBoberido as far as I understand). 

Now if we let the publisher block if the queue is full, I would expect the publisher to send the first 3 samples and then refuse sending more (to any subscriber). Whether this is desirable is another matter. This again does not consider the effects of multi-publish (which may cause other samples to be in the queue before blocking at 3 samples).

One could argue that such a setting of hsitroyRequest > queueCapacity should be illegal (which would make things easier), but the semantics above are reasonable. Currently those settings are valid. The details of the behavior and *intended semantics* are not entirely clear to me (needs further analysis).

doc/website/concepts/qos-policies.md
MatthiasKillat: This should be possible since those options/QoS do not represent the **structure of the data** but how it is made available (i.e. reliable, how much is cached, will the publisher block etc.). The Servicedescription should from my point of view only be concerned with the former. If we would like to have a DDS analogy it is the type information (or more precisely only a unique identifier for this information).

Note that I dislike the naming and design as it is. I would like to have a **ServiceIdentifier** which would be basically the current **ServiceDescription** which allows me to uniquely identify/locate the corresponding **Service/Type information**. 

The world would then consist of many available topics (services but this terminology is at odds with Autosar) with unique identifiers (say our 3 strings) and there may be many publishers for each of them with **different options**. Finally there are a number of subscribers with their own options which want to get data from a specific topic and will do so if topic and its own requirements *match* in some with what is offered in a way still to be precisely defined. This matching behavior hence depends on the **ServiceIdentifier AND the options** and is what this document seems to be about...

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
+# QoS policies
+
+Unlike other middlewares, e.g. [DDS used in ROS 2](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html),
+iceoryx keeps things simple and offers basic Quality of Service (QoS) settings. The QoS
+settings are called options and can be used to optimize and tailor the communication.
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings are:
```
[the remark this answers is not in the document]
mossmaurice: Nope, have a look [here](https://iceoryx.io/v1.0.1/getting-started/examples/iceoptions/). There is e.g. `subscribeOnCreate` or `nodeName`

elfenpiff (changes requested):
You checked the PR title check but the title does not follow our guidelines. Please add `iox-#???`.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
+# QoS policies
+
+Unlike other middlewares, e.g. [DDS used in ROS 2](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html),
+iceoryx keeps things simple and offers basic Quality of Service (QoS) settings. The QoS
```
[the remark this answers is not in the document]
mossmaurice: @elfenpiff @MatthiasKillat I would like to stick with the comparison to DDS. I think it's important to mention the influences of iceoryx. The overflow is described in [iceoptions](https://github.com/eclipse-iceoryx/iceoryx/tree/master/iceoryx_examples/iceoptions). Link is already there. This article is mainly about the matching criteria and not the features itself.

doc/website/concepts/qos-policies.md
mossmaurice: @elfenpiff Added to `iceoptions` readme as this does not belong to the QoS article.

doc/website/concepts/qos-policies.md
mossmaurice: @elfenpiff Yes, this is currently possible. The QoS options are independent from the `ServiceDescription`. We won't change this before `v2.0`, hence the description presents the current status quo.

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
...
+
+For more information about the options see the corresponding example [`iceoptions`](iceoptions.md).
+
+## Publisher subscriber matching criteria
+
+Two criteria have to be fulfilled in order for a publisher and a subscriber to be connected.
+
+ 1. Same `capro::ServiceDescription`
+ 2. Matching `SubscriberTooSlowPolicy` and `QueueFullPolicy` in `PublisherOptions` and `SubscriberOptions`
+
```
elBoberido, line 25: @mossmaurice there is a third criteria but I forgot the exact semantics. I think @MatthiasKillat can tell you about it

iceoryx_examples/iceoptions/README.md
```diff
@@ -81,6 +84,18 @@ If we want to enforce the contract that the publisher needs to support a `histor
...
+    In case of n:m communication, the history feature will **not** provide the overall last n samples based on delivery point in time!
+
+    The following two scenarios are examples of issues when using n:m together with the history feature:
+
+    1. Multiple publishers
+    The last n samples of every publisher are received. This means for m publishers in the worst
+    case m * n samples in random order, not in the order they were published on the topic.
+
+    2. Multiple publishers after the publisher called `stopOffer()` or is removed
+    The last n samples will never be received since they vanished. An arbitrary number of samples or nothing is received.
```
elBoberido, line 97: For me this important information does not belong into the example but in the `qos-policies` under a `Limitations` section. This could be linked from here but the information should be there.

doc/website/concepts/qos-policies.md
elBoberido: @mossmaurice could you please check whether my assumption from the second bullet point of the first list is correct? If yes, there is another limitation for the history when `QueueFullPolicy::BLOCK_PUBLISHER` is used

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,33 @@
+# QoS policies
+
+Unlike other middlewares, e.g. [DDS used in ROS 2](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html),
+iceoryx keeps things simple and offers basic Quality of Service (QoS) settings. The QoS
+settings are called options and can be used to optimize and tailor the communication.
+
+## `PublisherOptions` and `SubscriberOptions`
```
elBoberido, line 7: Do we need the same for `ClientOptions` and `ServerOptions`?

mossmaurice: Good point, updated.

doc/website/concepts/qos-policies.md
mossmaurice: Added.

iceoryx_examples/iceoptions/README.md
mossmaurice: I think it's more on the spot in the example as more people will look at the example than into the QoS article. I added the warning to the QoS and linked to `iceoptions`. Is that ok for you?

doc/website/concepts/qos-policies.md
mossmaurice: > newer samples are discarded when QueueFullPolicy::BLOCK_PUBLISHER, i.e. lost samples with a gap
Could you please check if I'm right?

@elBoberido This can't happen anymore as third matching criteria was introduced by @MatthiasKillat `SubscriberOptions::historyRequest <= PublisherOptions::historyCapacity`.

mossmaurice: > newer samples are discarded when QueueFullPolicy::BLOCK_PUBLISHER, i.e. lost samples with a gap
Could you please check if I'm right?

@elBoberido You're right. I adapted `iceoptions` and there is indeed a gap, when starting the subscriber late:

```
publisherOptions.historyCapacity = 10U;
subscriberOptions.queueCapacity = 1U;
-----
iox-cpp-subscriber-with-options got value: 4
iox-cpp-subscriber-with-options got value: 9
iox-cpp-subscriber-with-options got value: 10
iox-cpp-subscriber-with-options got value: 11
iox-cpp-subscriber-with-options got value: 12
iox-cpp-subscriber-with-options got value: 13
iox-cpp-subscriber-with-options got value: 14
iox-cpp-subscriber-with-options got value: 15
```

I documented that as a limitation as the code freeze for `v2.0`. Is that ok from your end?

elfenpiff (dismissed):

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,66 @@
...
+settings are called options and can be used to optimize and tailor the communication.
+
+## `ServerOptions` and `ClientOptions`
+
+The two most important settings are:
+
+| Option                                 | Explanation                                                                  |
+|----------------------------------------|------------------------------------------------------------------------------|
+| `ServerOptions::requestQueueCapacity`  | This enables servers to store n requests before they are passed to the users |
+| `ClientOptions::responseQueueCapacity` | This enables clients to store n respones before they are passed to the users |
```
elBoberido, line 14: I like `respones` ;)

Also, technically they are already at the user side. How about

```suggestion
| `ClientOptions::responseQueueCapacity` | This enables clients to store n responses at the users side |
```

Could you also change this for server and subscriber?

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Server and client matching criteria
+
+Two criteria have to be fulfilled in order for a server and a client to be connected.
+
+1. Same `capro::ServiceDescription`
+2. Matching `ConsumerTooSlowPolicy` and `QueueFullPolicy` in `ServerOptions`/`ClientOptions`
```
elBoberido, line 21: It feels weird to have this before `### Compatible policies` where the reader does not yet know the matching criteria. Could you move it below that section, like it is done with publish-subscribe?

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,66 @@
...
+
+    For more information about the options see the corresponding example [`iceoptions`](iceoptions.md).
+
+!!! info
+    If the `PublisherOptions::historyCapacity` is larger than `SubscriberOptions::queueCapacity` and blocking behaviour
+    is active, late-joining subscribers will not receive the latest and greatest sample, effectively loosing some.
+
+## Publisher and subscriber matching criteria
+
+If `requiresPublisherHistorySupport` is set, additionally to the matching criteria of server and client, there is there is a third one for publishers and subscribers:
```
elBoberido, line 62: `there is there is` :)

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,66 @@
...
+| `ClientOptions::responseQueueCapacity` | This enables clients to store n respones before they are passed to the users |
+
+## Server and client matching criteria
+
+Two criteria have to be fulfilled in order for a server and a client to be connected.
+
+1. Same `capro::ServiceDescription`
+2. Matching `ConsumerTooSlowPolicy` and `QueueFullPolicy` in `ServerOptions`/`ClientOptions`
+
+### Compatible policies
```
elBoberido, line 23: Just a food for thought. Only this section talks about compatibility, everywhere else matching is used. Would it make sense to call this `Policy matching`?

Maybe also adding a sentence like
```
The `ConsumerTooSlowPolicy` and `QueueFullPolicy` are used to control the behavior for full queues. Since some combinations would be contradicting, not all of them match to connect producer and consumer.
```

iceoryx_examples/iceoptions/README.md
```diff
@@ -81,6 +84,10 @@ If we want to enforce the contract that the publisher needs to support a `histor
 `requirePublisherHistorySupport` to `true`. In this case the subscriber will only connect if the publisher history support is at least as large as is request.
 By default this is set to `false` and best-effort behavior is used.
 
+!!! warning
+    In case of n:m communication, the history feature will **not** provide the overall last n samples based on delivery point in time!
+    For more information about this limitation see the [QoS article](https://iceoryx.io/v2.0.0/concepts/qos-policies/).
```
elBoberido, line 89: Please set this to `latest` and let a script do the work for you ... be lazy :laughing:

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,66 @@
...
+    If the `PublisherOptions::historyCapacity` is larger than `SubscriberOptions::queueCapacity` and blocking behaviour
+    is active, late-joining subscribers will not receive the latest and greatest sample, effectively loosing some.
+
+## Publisher and subscriber matching criteria
+
+If `requiresPublisherHistorySupport` is set, additionally to the matching criteria of server and client, there is there is a third one for publishers and subscribers:
+
+1. Same `capro::ServiceDescription`
+2. Matching `ConsumerTooSlowPolicy` and `QueueFullPolicy` in `PublisherOptions`/`SubscriberOptions`
+3. `SubscriberOptions::historyRequest` <= `PublisherOptions::historyCapacity`
```
elBoberido, line 66: This is only true when `requiresPublisherHistorySupport` is true.

iceoryx_examples/iceoptions/README.md
mossmaurice: I will save a script some execution time and leave it here. We need to be frugal with resources! :grin:

doc/website/concepts/qos-policies.md
mossmaurice: That is mentioned four lines above.

doc/website/concepts/qos-policies.md
mossmaurice: Fixed.

doc/website/concepts/qos-policies.md
mossmaurice: Added the sentece, but left the title as it is. "Compatible policies" is easier to understand for me.

elfenpiff (approved):

doc/website/concepts/qos-policies.md
elBoberido: Okay, I somehow overlooked it

iceoryx_examples/iceoptions/README.md
elBoberido: This makes the script more complex and will break for v3. Please keep it uniform on master

mossmaurice: IMHO the script will need this complexity anyway. Imagine we are on `master` at version `v2.1.16` and want to create both a `v2.1.17` on a release branch and `v3.0.0`. 

cc @dkroenke

elfenpiff: My idea would be, everything points to master and then we start the magic script with a release tag and the links are updated.

elfenpiff (dismissed):

iceoryx_examples/iceoptions/README.md
elBoberido: @elfenpiff exactly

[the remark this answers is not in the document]
mossmaurice: I have to strongly disagree here, sorry!

This is what we have done in the past to avoid relative links. There are several links to the iceoryx.io already in the examples. The reason is the following:

The `Readme.md`s of all examples are currently the only `.md` which are shared between GitHub and [iceoryx.io](https://iceoryx.io).

For iceoryx.io the links have to look like:
```
[QoS article](qos-policies.md) // mkdocs-autolinks-plugin will resolve the link
#or
[QoS article](../../doc/website/concepts/qos-policies.md)
#or
[QoS article](https://iceoryx.io/v2.0.0/concepts/qos-policies/)
```

For reading the document via GitHub's Web GUI:
```
[QoS article](https://iceoryx.io/v2.0.0/concepts/qos-policies/)
#or
[QoS article](../../doc/website/concepts/qos-policies.md) // User would land on a GitHub page with broken links
#or
[QoS article](https://iceoryx.io/v2.0.0/concepts/qos-policies/)
```

The decision for `v1.0` was to avoid relative links and use absolute links. IMHO we should stick to it.

cc @FerdinandSpitzschnueffler @dkroenke

mossmaurice: Ok, done.

elBoberido: @mossmaurice would it be possible to have a snapshot of master on the homepage? Not as default but selectable in the combo box, e.g.
```
latest <- this is the default; maybe instead of `lastest` the version number like it is now
master
v2.0.0
v1.0.1
v1.0.0 
```
What do you think? This is nothing for v2 but maybe afterwards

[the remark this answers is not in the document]
elBoberido: I would keep this as it was for 1.0 and then we need to create some tooling around this to fix it.

mossmaurice: @elBoberido Sure, we just need to setup a CI job for that!

mossmaurice: Created the issue: https://github.com/eclipse-iceoryx/iceoryx/issues/1201

elBoberido (approved):

iceoryx_examples/iceoptions/README.md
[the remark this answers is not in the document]
elfenpiff: @mossmaurice Then please use the link `[QoS article](../../doc/website/concepts/qos-policies.md)`

As you mentioned it works on the website and the markdown document and for 3.0 we create some tooling to clean this up.

doc/website/concepts/qos-policies.md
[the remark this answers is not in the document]
budrus: I reliable means no data is lost, Then only the blocking option is reliable. This is how we map now in Cyclone

iceoryx_examples/iceoptions/README.md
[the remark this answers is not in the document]
mossmaurice: @elBoberido The reason is in my comment above:

`[QoS article](../../doc/website/concepts/qos-policies.md) // User would land on a GitHub page with broken links`

Additionally, using relative links turned out to be a debacle with `v1.0.0`. Hence, the conclusion was to remove them.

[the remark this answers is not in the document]
elBoberido: Okay, then let's keep it as with v1 also for v2. It's an inconvenience to be redirected from github to the homepage but the links are working

doc/website/concepts/qos-policies.md
```diff
@@ -0,0 +1,69 @@
...
+
+## `PublisherOptions` and `SubscriberOptions`
+
+The three most important settings are:
+
+| Option                              | Explanation                                                                                                                                       |
+|-------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------|
+| `PublisherOptions::historyCapacity` | This enables late-joining subscribers to request the n last samples                                                                               |
+| `SubscriberOptions::queueCapacity`  | Size of the subscriber queue where samples are stored at the user side                                                                            |
+| `SubscriberOptions::historyRequest` | The number of samples a late-joining subscriber will request from a publisher, should be equal or smaller than `historyCapacity` otherwise capped |
```
budrus, line 43: @mossmaurice The behavior of today would maybe best be described with `The number of samples a late-joining subscriber will request from every offered publisher` Or? 
I would write here as it behaves today

mossmaurice: Added

budrus (approved):

elBoberido (approved):

PR #1027, opened 2022-01-19
Remove error handler calls from hoofs

### LINKED ISSUE #590: Improve modularity by creating architecure guidelines

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Templatize `ErrorHandler` to prepare usage in different modules
* Remove all `errorHandler()` calls from hoofs as described in the error handling document

## References

- Closes #590


iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -17,20 +17,23 @@
 #ifndef IOX_HOOFS_ERROR_HANDLING_ERROR_HANDLING_HPP
 #define IOX_HOOFS_ERROR_HANDLING_ERROR_HANDLING_HPP
 
+/// @todo #590 rename this file to error_handler.hpp
```
mossmaurice, line 20: Done in #1028

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -268,17 +271,83 @@ class ErrorHandler
...
 ///     });
 ///
 /// errorHandler(Error::kTEST__ASSERT_CALLED);
 /// ASSERT_TRUE(called);
 /// @endcode
-void errorHandler(const Error error,
-                  const std::function<void()>& errorCallBack = std::function<void()>(),
-                  const ErrorLevel level = ErrorLevel::FATAL) noexcept;
+/// @tparam[in] Error type of the enum which is used to report the error
+/// @todo use enable_if with is_enum
```
mossmaurice, line 283: Done in #1028

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -268,17 +271,83 @@ class ErrorHandler
...
+template <typename Error>
+inline void errorHandler(const Error error,
+                         const std::function<void()>& errorCallBack = std::function<void()>(),
+                         const ErrorLevel level = ErrorLevel::FATAL) noexcept
+{
+    ErrorHandler<Error>::handler(error, errorCallBack, level);
+}
+
+/// @todo #590 specialise for posh and c binding error handling
```
mossmaurice, line 292: Done in #1028

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -268,17 +271,83 @@ class ErrorHandler
...
+    else
+    {
+        std::stringstream ss;
+        ss << "iceoryx_posh error! " << error;
+
+        reactOnErrorLevel(level, ss.str().c_str());
+    }
+}
+
+/// @todo #590 move the below to .inl
```
mossmaurice, line 311: Done in #1028

elfenpiff (commented):
Can we please keep this PR on hold until we figured out how the overall error concept in iceoryx looks like.
To call terminate is most likely not an option for ASIL D and we may have to make this either configurable or that some user can define custom callbacks to notify maybe some kind of watchdog or even throw an exception.

In those cases an error handler may is the more viable solution.

MatthiasKillat:
From a quick view I think the approach is fine and in the scope well-defined. It is important that the error handling semantics will not be changed for now, only the syntax and organization of the code.

I.e. we check for the same conditions as before and react in the same way, just use different code. In other words there is no regression and also no new handling introduced.

This is a first step to allow individual error-handling for different modules. Changes of the overall concept should indeed be postponed and require more discussion. We will not have any overall concept change in iceoryx 2.0.

NB: this is just a first impression, I will have a closer look tomorrow.

mossmaurice:
@elfenpiff 
> Can we please keep this PR on hold until we figured out how the overall error concept in iceoryx looks like. To call terminate is most likely not an option for ASIL D and we may have to make this either configurable or that some user can define custom callbacks to notify maybe some kind of watchdog or even throw an exception.
> 
> In those cases an error handler may is the more viable solution.

Thanks for the remark, got your point. As we're not 100% sure how the final concept will look like, I would suggest to take an iterative approach here. My proposal is to get step by step to the full blown concept. Two goals are reached in #1027 and #1028:

- Cleaner architecture: hoofs is independent from both posh and C binding
- We are able to detect based on the enum type which module reported the error

As @MatthiasKillat pointed out, there is no semantic change. A next step towards the final design could be adding hooks to report and persistently save all errors from any module. This would enable reading them e.g. via [UDS](https://en.wikipedia.org/wiki/Unified_Diagnostic_Services) (both in `Expects`/`Ensures` and `errorHandler`).

elfenpiff (approved):

elBoberido:
@elfenpiff @MatthiasKillat @mossmaurice 
I also don't think this should be kept on hold. Like already said, it is a small iterative step and does not change the concept by much. I wouldn't call the `cxx::Expects` in the error handler though but the other way around. The `cxx::Expects` should call the errror handler. At one point we have to delegate the error code to a 3rd party and `cxx::Expects` cannot do this since the error value is already erased. So after some thinking about this, I would keep the calls to the error handler in hoofs but create separate enums for each component. The `cxx::Expects` should also have an entry in this enum, e.g. `EXPECTS_CONDITION_FAILED`.

I felt asleep on the keyboard and when I woke up I had a PoC of a refactored logger on my laptop. @elfenpiff told me that it also happens to him at times. Might be a virus and seems contagious.
I think the same concept could be used to make the error handler more ergonomic and provide a hook for a 3rd party framework. I'll create a design document draft for the logger refactoring and then we can discuss if this would also be an option for the error handler. Small teaser, I envision something like this
```cpp
iox_error(FATAL, HoofsError::POSIX_SEMAPHORE_FOO) << "what the heck: " << 42 \
    << "! This is not supposed to happen";
```

Btw, I'm wondering for what the custom error callback is good for? Do we need it? Can we remove it?
I had a quick grep over the codebase and we never provide a callback but I had some fun on #861 because it was accidentally used. We should kill it with fire but not in this PR.

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
```diff
@@ -379,57 +378,44 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
-
-    Timer sut(TIMEOUT, [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10)); });
-
-    ASSERT_FALSE(sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT).has_error());
-    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10));
-    ASSERT_FALSE(sut.restart(TIMEOUT, Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::TERMINATE).has_error());
-    std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10));
-
-    TIMING_TEST_EXPECT_TRUE(hasTerminated);
+    EXPECT_DEATH(
```
elBoberido, line 408: AFAIK the timer already started a thread and gtest will print warnings when a process with running threads is forked. We play Russian roulette with this and on a previous PR I had to disable one of these tests because it hang on macOS. Maybe it is not the worst idea to use the error handler in this case and get rid of most of the death tests.

Slightly offtopic but in this case we might also want to delete this class since it is not used in iceoryx and will also not be certified in this state.

elBoberido (commented):
I don't see a reason to create the template. When this is specialized we would essentially need to duplicate the code multiple times. It is much easier to only have a template function which erases the enum class and provides the error string to the error handler, e.g.
```cpp
template <typename E>
void errorHandlerWrapper(E error, callback, ErrorLevel) {
   errorHandler(static_cast<underlying_type>(error), to_string(error), callback, ErrorLevel);
}
```
Did I misunderstood something? If not, please revert those changes

iceoryx_hoofs/source/file_reader/file_reader.cpp
```diff
@@ -33,34 +35,31 @@ FileReader::FileReader(const std::string& f_fileName, const std::string& f_fileP
...
-            {
-                std::cerr << "\033[5;31m"
-                          << "Could not open file '" << m_file << "'. Exiting!"
-                          << "\033[0m" << std::endl;
-                std::terminate();
-                break;
-            }
-            }
-        });
+        switch (m_errorMode)
```
MatthiasKillat, line 38: Is semantically (almost) equivalent (the code `Error::kFILEREADER__FAILED_TO_OPEN_FILE` was removed, intentionally I presume to have no such codes in hoofs).

It looks a bit weird though in general, as is kind of invokes its own error handling inside the switch. If everything is generalized properly, the error handler should handle the decision whether to terminate etc. based on a severity level and a safety mode (`ErrorMode` is kind of like this).

The user only should provide the safety/handling mode, error severity and optional custom code/messages. It is important to have the code as concise as possible for this but also used in the same way in all places. 

NB: I am just trying to think through how it could ideally work in the future, nothing to change here for now.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -113,11 +112,7 @@ void Timer::OsTimer::callbackHelper(sigval data) noexcept
             // prohibits other threads from entering the flag protected region
             handle.m_callbackIsAboutToBeExecuted.test_and_set(std::memory_order_acq_rel);
 
-            if (handle.m_timer == nullptr)
-            {
-                errorHandler(Error::kPOSIX_TIMER__INCONSISTENT_STATE);
-                return;
-            }
+            cxx::Expects((handle.m_timer != nullptr) && "Timer in inconsistent state");
```
MatthiasKillat, line 115: In a future version of `Expects` the message should be a separate (optional?) argument and not appended with `&&` like in the good old days for `assert` in C.

Only a style issue and personal opinion though, nothing for this PR unless this is already possible.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -166,11 +161,8 @@ void Timer::OsTimer::callbackHelper(sigval data) noexcept
     }
     else
     {
-        if (handle.m_catchUpPolicy == CatchUpPolicy::TERMINATE)
-        {
-            errorHandler(Error::kPOSIX_TIMER__CALLBACK_RUNTIME_EXCEEDS_RETRIGGER_TIME);
-        }
-        return;
+        cxx::Ensures((handle.m_catchUpPolicy != CatchUpPolicy::TERMINATE)
```
MatthiasKillat, line 164: Personally I think the distinction between `Ensures`and `Expects` is arbitrary and not needed. Especially if I can and want to call either in the middle of a function. The names are also not meaningfully distinct.

We just need one of them and might as well call it `Assert` (though I am fine with `Expect` as well). Just a better flow when reading the code IMO (also closer to say GTest, where you also *expect* some condition to hold at specific points of the test).

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
```diff
@@ -379,57 +378,44 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
-        [&](const iox::Error, const std::function<void()>, const iox::ErrorLevel) { hasTerminated = true; });
-
     Timer sut(TIMEOUT, [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10)); });
-
     ASSERT_FALSE(sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT).has_error());
-
     std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10));
-    TIMING_TEST_EXPECT_FALSE(hasTerminated);
+    // EXPECT_NO_DEATH
+    EXPECT_TRUE(true);
```
MatthiasKillat, line 385: This makes no sense from my point of view (tautology). If we reach the code, the test passes anyway (and is listed as successful). Otherwise we have crashed and the line does nothing.

I know that it is repeated at most 5 times and needs to pass at least once, but is this really meaningful/needed?

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
```diff
@@ -379,57 +378,44 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
-        [&](const iox::Error, const std::function<void()>, const iox::ErrorLevel) { hasTerminated = true; });
-
     Timer sut(TIMEOUT, [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10)); });
-
     ASSERT_FALSE(sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::IMMEDIATE).has_error());
-
     std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10));
-    TIMING_TEST_EXPECT_FALSE(hasTerminated);
+    // EXPECT_NO_DEATH
+    EXPECT_TRUE(true);
```
MatthiasKillat, line 393: Required and meaningful? See above.

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
```diff
@@ -379,57 +378,44 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
 });
 
 TIMING_TEST_F(Timer_test, CatchUpPolicyTerminateTerminatesWhenCallbackIsLongerThenTriggerTime, Repeat(5), [&] {
```
MatthiasKillat, line 396: More like `...CallbackRunsLonger...` but functionally irrelevant.

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
MatthiasKillat: Yes, as far as I know fork has no information about any child threads and the resources (e.g. mutex) they use. I think it will copy the entire process space but only have the clone of the calling thread running in the fork Everything else may not run or is undefined, the threads may not have their own proper mutexes etc. (though their states are somewhat copied in memory, see below).

Likely this is called on linux.
https://man7.org/linux/man-pages/man2/fork.2.html

TLDR: I am not sure whether the threads in a forked process can operate properly and have the resources they need (independent from the original). From the above I think this is not the case and a tricky problem. As DEATH tests fork, we cannot use them if multiple threads run (or need to understand the limitations, I do not fully understand them).

iceoryx_posh/test/integrationtests/test_posh_mepoo.cpp
```diff
@@ -416,7 +416,7 @@ TEST_F(Mepoo_IntegrationTest, WrongSampleSize)
     constexpr uint32_t SAMPLE_SIZE = 2048U;
     constexpr uint32_t REPETITION = 1U;
     iox::cxx::optional<iox::Error> receivedError;
-    auto errorHandlerGuard = iox::ErrorHandler::setTemporaryErrorHandler(
+    auto errorHandlerGuard = iox::ErrorHandler<iox::Error>::setTemporaryErrorHandler(
```
MatthiasKillat, line 419: We may be able to get rid of the explicit template argument by defining a specific error handler alias (via using) in the ''posh' namespace. This could actually be good practice and leads to modular error handling without. We would then call the `iox::posh::ErrorHandler` explicitly or implicitly (the latter may be undesirable if the error handler is not configured correctly for all namespaces where it is called).

MatthiasKillat (approved):
Fine by me and a step in a direction to modular/custom error handling that I like. I mentioned some proposals and potential improvements for the future. Requires some more thought and iterations to really arrive at something generic and powerful but there is no immediate pressure to do so.

MatthiasKillat:
I retracted my approval because I am not sure whether we can get rid of the template argument in
```template <typename Error>
class ErrorHandler```
I would like to, if possible (as would @elBoberido ). I ust to not fully understand all implications and goals.

The template argument has the advantage the we actually enforce what kind or error this particular handler can handle at compile time (in contrast to providing a template method where the type is variable). This is only partially possible with template methods (would need `enable_if` which gets out of hand quickly)

This is closely related to the usage like ```auto errorHandlerGuard = iox::ErrorHandler<iox::Error>::setTemporaryErrorHandler ... ```. Here would like to somehow define it in a way that enforces to automatically call the correct error handler in the whole posh namespace etc. with the right error handler, maybe with a `using` directive in the module namespace. This alias will need to be called explicitly though

Sorry, this is not fully thought out on my side, I need to think this through over the weekend.
It is mainly a usage problem and I am not fully aware whether the goal is to actually enforce that some handler can only handle specific error types. Functionally it is all fine.

File: iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp [hunk 13]
```diff
@@ -209,20 +205,22 @@ enum class ErrorLevel : uint32_t
     MODERATE
 };
 
+template <typename Error>
 using HandlerFunction = std::function<void(const Error error, const std::function<void()>, const ErrorLevel)>;
```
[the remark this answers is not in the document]
elBoberido: Agree. If time is running out for v2.0 then definitely shortly after

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -268,17 +271,87 @@ class ErrorHandler
...
+                                                const std::function<void()>& errorCallBack,
+                                                const ErrorLevel level) noexcept
+{
+    if (errorCallBack)
+    {
+        errorCallBack();
+    }
+    else
+    {
+        std::stringstream ss;
```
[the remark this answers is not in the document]
elBoberido: Funnily the error callback was (almost) not used in our codebase. There was only one occasion where @mossmaurice removed it and we had a few places where it accidentally was used wrongly and bypassed the error handler. I guess removing this is the best we can do

iceoryx_hoofs/source/posix_wrapper/timer.cpp
elBoberido: I put `Requires` to the polls :laughing:

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
mossmaurice: @elBoberido Agree 100%, `DEATH_TESTS` should therefore always be avoided. Let's remove the `posix::Timer` together with resolving #754.

elfenpiff (changes requested):
You checked the PR title check but the title does not follow our guidelines. Please add `iox-#???`.

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
mossmaurice: See [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1027#discussion_r790965363)

mossmaurice:
@elBoberido Thanks for the review.

> The cxx::Expects should call the errror handler. At one point we have to delegate the error code to a 3rd party and cxx::Expects cannot do this since the error value is already erased.
I think you mixed something up. Probably you meant the `FileReader`? This is a bit misleading to see in the code, but this is not the error mode of the `ErrorHandler` but of the `FileReader` itself. Hence, I completely agree with `The cxx::Expects should call the errror handler.` :heavy_check_mark: 

> The cxx::Expects should also have an entry in this enum, e.g. EXPECTS_CONDITION_FAILED.

Will be implemented by #1035

> Btw, I'm wondering for what the custom error callback is good for? Do we need it? Can we remove it?

That's true indeed :+1: Not in this PR.

> I don't see a reason to create the template. 

Well, `setTemporaryErrorHandler` is the reason I templatized, but you're right it's possible to erase the type and just use the underlying integer value. I'll give it a try.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -276,9 +245,95 @@ class ErrorHandler
...
+const char* toString(const Error error) noexcept;
+
+template <typename Error>
+cxx::optional<iox::TypedHandlerFunction<Error>> typedHandler;
+
+
+/// @todo #590 move the implementations below to .inl
+template <typename Error>
+inline void errorHandler(const Error error,
+                         const std::function<void()>& errorCallBack IOX_MAYBE_UNUSED,
```
mossmaurice, line 295: The removal will be done either in #1028 or a follow-up.

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
```diff
@@ -58,8 +58,8 @@ TEST(c2cpp_enum_translation_test, SubscriberState)
 // the clang sanitizer detects this successfully and this leads to termination, and with this the test fails
 #if !defined(__clang__)
     iox::Error errorValue = iox::Error::kNO_ERROR;
-    auto errorHandlerGuard = iox::ErrorHandler::setTemporaryErrorHandler(
-        [&](const iox::Error e, const std::function<void()>, const iox::ErrorLevel) { errorValue = e; });
+    auto errorHandlerGuard = iox::ErrorHandlerMock::setTemporaryErrorHandler<iox::Error>(
```
dkroenke, line 61: You need to include `#include "iceoryx_hoofs/testing/error_handler_mock.hpp"` to make this work with clang

mossmaurice:
> I'll ask the heretical question, is this relevant for v2? The code freeze is not that far away and it binds resources on the reviewer side and on developer fixing merge conflicts due to breaking API changes. If we have too many of this side topics I fear we might run out of time at the end of February :/ This is not only related to your PR but in general to all PRs we currently have.

I understand what you mean. First prio are definitely the features #415 and #27. However, as this PR solves the last "illegal" dependency from `iceoryx_hoofs` to `iceoryx_posh`, I'd appreciate it a lot if we can merge it before `v2.0`. A clean architecture will enable us faster development in future. I have several other things in mind for `v2.0` that are on the [board](https://github.com/eclipse-iceoryx/iceoryx/projects/3).

elBoberido (commented):
Looks good. Just minor things. Could also be merged as is.

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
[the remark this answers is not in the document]
mossmaurice: I was thinking about this also back and forth. My conclusion was: The `ErrorHandlerMock` is not really a mock more of a `ErrorHandlerHook`. I kept the name as the term `Mock` will trigger a "Hey this is only for tests"-thought when developer read it. That's why I think we should both keep `LoggerMock` and `ErrorHandlerMock`. Other names could be misleading.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -209,45 +202,18 @@ enum class ErrorLevel : uint32_t
...
 ///             error(MODULE_NAME__MY_FUNKY_ERROR)
 ///         Attention: Create an error after the following convention:
 ///             MODULE_NAME__A_CLEAR_BUT_SHORT_ERROR_DESCRIPTION
 ///         And a long name is alright!
 ///
-///     2.) Call errorHandler(Error::kMODULE_NAME__MY_FUNKY_ERROR);
+///     2.) Specialize the following methods for your NewEnumErrorType:
+///         - const char* toString(const NewEnumErrorType error)
+///
+///     3.) Call errorHandler(Error::kMODULE_NAME__MY_FUNKY_ERROR);
```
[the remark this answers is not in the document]
mossmaurice: There wasn't a decision on that topic yet. If we will change it, I would for sure do it in a follow-up PR.

iceoryx_hoofs/source/error_handling/error_handling.cpp
```diff
@@ -17,43 +17,26 @@
...
-#include "iceoryx_hoofs/log/logging.hpp"
-#include "iceoryx_hoofs/log/logmanager.hpp"
-
-#include <iostream>
-#include <sstream>
-
 namespace iox
 {
-const char* ErrorHandler::ERROR_NAMES[] = {ICEORYX_ERRORS(CREATE_ICEORYX_ERROR_STRING)};
+const char* ERROR_NAMES[] = {ICEORYX_ERRORS(CREATE_ICEORYX_ERROR_STRING)};
```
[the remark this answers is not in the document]
mossmaurice: Have a look at #1028. The names will be unique for each module e.g. `C_BINDING_ERROR_NAMES[]`.  Hence no clashes are possible.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,80 @@
...
+    if (errorEnumType == expectedErrorEnumType)
+    {
+        // We undo the type erasure
+        auto typedError = static_cast<ErrorEnumType>(error);
+        typedHandler<iox::Error>.and_then(
+            [&](TypedHandlerFunction<Error> storedHandler) { storedHandler(typedError, level); });
+    }
+    else
+    {
+        GTEST_FAIL() << errorName;
```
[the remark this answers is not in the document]
mossmaurice: Sure, that makes a lot of sense. If you don't mind, I would do that in #1028. Is that ok for you?

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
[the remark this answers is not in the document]
elBoberido: Yeah, I'm a bit torn on this but like I said, I don't have an idea where I would say it is much better than the current naming.

iceoryx_hoofs/source/error_handling/error_handling.cpp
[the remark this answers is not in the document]
elBoberido: Okay, so it will be taken care of when the split actually happens

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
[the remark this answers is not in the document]
elBoberido: Would be fine, but this is a 5 minute task and there aren't any approvals which would be invalidated :smile:

iceoryx_binding_c/test/test.hpp
```diff
@@ -17,6 +17,8 @@
 #ifndef IOX_DDS_GATEWAY_TEST_TEST_HPP
 #define IOX_DDS_GATEWAY_TEST_TEST_HPP
 
+#include "iceoryx_hoofs/testing/mocks/error_handler_mock.hpp"
```
elfenpiff, line 20: Please adjust the copyright header.

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
```diff
@@ -58,8 +58,8 @@ TEST(c2cpp_enum_translation_test, SubscriberState)
 // the clang sanitizer detects this successfully and this leads to termination, and with this the test fails
 #if !defined(__clang__)
     iox::Error errorValue = iox::Error::kNO_ERROR;
-    auto errorHandlerGuard = iox::ErrorHandler::setTemporaryErrorHandler(
```
elfenpiff, line 61: Please adjust the copyright header.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
elfenpiff: This is hard. Furthermore #1028  states the following: ` Separate errors of posh and C binding` and as it seems currently the include guards, include headers, cmake files are not yet adjusted there. Please create a separate PR for this the other one is already huge.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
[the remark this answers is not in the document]
elfenpiff: @elBoberido @mossmaurice I think we introduced the `k` prefix so that it conforms to the autosar guidelines and then we found out that this is actually pretty nice since otherwise we would run into some trouble with our macro.

So whoever does this - be prepared for fun :D

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -276,9 +242,44 @@ class ErrorHandler
 /// errorHandler(Error::kTEST__ASSERT_CALLED);
 /// ASSERT_TRUE(called);
 /// @endcode
+/// @tparam[in] Error type of the enum which is used to report the error
+/// @todo #590 use enable_if with is_enum
```
elfenpiff, line 246: Lets assume some user would like to transmit additional information like a state or stack trace inside a struct. With enforcing enums here we have to write more code while restricting the user unnecessarily.

Iceoryx does not have a problem in supporting any kind of type here so maybe we should leave it this way.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
elfenpiff: Maybe we should create small separate issues for the error handler since I am already a little bit lost. And you have several comments where you state, will be done in this PR or will be done in another follow up  or in this issue and so on. And when I look at those PRs, issues etc. either the task is done incompletely or nothing is mentioned in the issue to this particular task.

For me as a reviewer it is hard to track all of this.

iceoryx_hoofs/source/error_handling/error_handling.cpp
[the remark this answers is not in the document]
elfenpiff: I think we have to do some restructuring. Issue #590  states nothing about the error handling, we have several hints that certain tasks are performed in several PRs or in this issue but this is nowhere really tracked or mentioned.

Why is the error handler part of #590 ? (I know why but this is hard to keep track when this is nowhere mentioned)

elfenpiff (changes requested):
Please create another issue where you only explain the work which is done here on the error handler. Issue #590 should have a todo list which links to all the separate issues. Then we have nice small understandable issues where we can keep track of stuff.

iceoryx_hoofs/testing/error_handler_mock.cpp
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
```
elfenpiff, line 1: Please move this file into the `mocks` subdirectory.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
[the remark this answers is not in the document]
elBoberido: @elfenpiff that's no problem. One has just to replace `#define CREATE_ICEORYX_ERROR_ENUM(name) k##name,` with `#define CREATE_ICEORYX_ERROR_ENUM(name) name,`

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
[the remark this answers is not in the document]
MatthiasKillat: If it has full functionality and is an error handler used in tests it should not be called a `Mock` of any kind. I think `TestingErrorHandler` is better for this reason.

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
```diff
@@ -95,8 +95,8 @@ TEST(c2cpp_enum_translation_test, SubscriberEvent)
...
-        [&](const iox::Error e, const std::function<void()>, const iox::ErrorLevel) { errorValue = e; });
+    auto errorHandlerGuard = iox::ErrorHandlerMock::setTemporaryErrorHandler<iox::Error>(
```
MatthiasKillat, line 98: Except that `std::function` is gone this does not look and feel better (naming with Mock, template arg) to use. We can fix this later of course.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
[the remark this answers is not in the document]
mossmaurice: Done :smile:

iceoryx_hoofs/source/error_handling/error_handling.cpp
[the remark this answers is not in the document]
mossmaurice: That's true. I'll close this PR here and re-open it as part of a new issue.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
mossmaurice: On the 2nd thought, I think what we need to do is separate the `ErrorHandler` from `error_handling.hpp`. `iceoryx_hoofs`'s `error_handling.hpp` will have only one entry `EXPECTS_ENSURES_FAILED`. Added to #1099

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
mossmaurice: I'll close this PR and will open an new one based on #1099

PR #1033, opened 2022-01-21
[do-not-review] iox-#27 Add ServerPort

### LINKED ISSUE #27: Request/Response communication with iceoryx

elBoberido:

1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Closes #27


iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_roudi.hpp
```diff
@@ -62,6 +65,8 @@ class ServerPortRouDi : public BasePort
     const MemberType_t* getMembers() const noexcept;
     MemberType_t* getMembers() noexcept;
 
+    void handleCaProProtocollViolation(iox::capro::CaproMessageType messageType) const noexcept;
```
elfenpiff, line 68: I know it states do not review therefore will not tell you that Protocol is only written with a single `l`
And that your argument is missing some const.

iceoryx_posh/source/popo/ports/server_port_data.cpp
```diff
@@ -23,13 +23,20 @@ namespace popo
...
-                               const NodeName_t& nodeName,
+                               const ServerOptions& serverOptions,
                                mepoo::MemoryManager* const memoryManager,
                                const mepoo::MemoryInfo& memoryInfo) noexcept
-    : BasePortData(serviceDescription, runtimeName, nodeName)
-    , m_chunkSenderData(memoryManager, SERVER_SUBSCRIBER_POLICY, 0, memoryInfo)
-    , m_chunkReceiverData(cxx::VariantQueueTypes::FiFo_MultiProducerSingleConsumer, SERVER_PUBLISHER_POLICY)
+    : BasePortData(serviceDescription, runtimeName, serverOptions.nodeName)
+    , m_chunkSenderData(
+          memoryManager, static_cast<SubscriberTooSlowPolicy>(serverOptions.clientTooSlowPolicy), 0, memoryInfo)
```
elfenpiff, line 31: Is it possible that `clientTooSlowPolicy` contains an integer which is not convertable to `SubscriberTooSlowPolicy`?
Just asking - not reviewing ;)

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,21 +39,94 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
+        const auto ret = m_chunkSender.tryAddQueue(static_cast<ClientChunkQueueData_t*>(caProMessage.m_chunkQueueData),
+                                                   caProMessage.m_historyCapacity);
+        if (!ret.has_error())
+        {
+            responseMessage.m_type = capro::CaproMessageType::ACK;
+            responseMessage.m_chunkQueueData = static_cast<void*>(&getMembers()->m_chunkReceiverData);
+            responseMessage.m_historyCapacity = 0;
+        }
+    }
+    break;
```
elfenpiff, line 104: I would move the break into the scope of the `case`. It feels otherwise a little weird.

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,21 +39,94 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
+        return cxx::make_optional<capro::CaproMessage>(responseMessage);
+    }
+
+    switch (caProMessage.m_type)
+    {
+    case capro::CaproMessageType::CONNECT:
+    {
+        cxx::Expects(caProMessage.m_chunkQueueData != nullptr && "Invalid response queue passed to server");
+        const auto ret = m_chunkSender.tryAddQueue(static_cast<ClientChunkQueueData_t*>(caProMessage.m_chunkQueueData),
+                                                   caProMessage.m_historyCapacity);
```
elfenpiff, line 96: why not use `and_then`?

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,21 +39,94 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
+            responseMessage.m_type = capro::CaproMessageType::ACK;
+            responseMessage.m_chunkQueueData = static_cast<void*>(&getMembers()->m_chunkReceiverData);
+            responseMessage.m_historyCapacity = 0;
+        }
+    }
+    break;
+    case capro::CaproMessageType::DISCONNECT:
+    {
+        const auto ret =
+            m_chunkSender.tryRemoveQueue(static_cast<ClientChunkQueueData_t*>(caProMessage.m_chunkQueueData));
```
elfenpiff, line 108: why not use `and_then`?

iceoryx_posh/source/popo/ports/server_port_data.cpp
elBoberido: This is just a stop gap solution. `SubscriberTooSlowPolicy` will be removed and replaced with `ConsumerTooSlowPolicy`, the type of `clientTooSlowPolicy`. Since that change affects quite many files, it is deferred to a later point in time. See also the `todo` in  `port_queue_policies.hpp`

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
elBoberido: c&p from `publisher_port_roudi.cpp` without thinking too much about the coding style

iceoryx_hoofs/test/moduletests/test_posix_signal_watcher.cpp
```diff
@@ -58,19 +58,22 @@ class SignalWatcher_test : public Test
```
elBoberido, line 58: Please ignore this. Part of #1063

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_receiver.hpp
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
elBoberido, line 1: Please ignore this. Part of #1063

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_receiver.inl
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
elBoberido, line 1: Please ignore this. Part of #1063

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.hpp
```diff
@@ -51,6 +51,23 @@ from<mepoo::MemoryManager::Error, popo::AllocationError>(const mepoo::MemoryMana
```
elBoberido, line 51: Please ignore this. Part of #1063

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender.inl
```diff
@@ -40,6 +40,37 @@ from<mepoo::MemoryManager::Error, popo::AllocationError>(const mepoo::MemoryMana
```
elBoberido, line 40: Please ignore this. Part of #1063

iceoryx_posh/test/moduletests/test_popo_chunk_receiver.cpp
```diff
@@ -236,4 +236,35 @@ TEST_F(ChunkReceiver_test, Cleanup)
     EXPECT_THAT(m_memoryManager.getMemPoolInfo(0).m_usedChunks, Eq(0U));
```
elBoberido, line 236: Please ignore this. Part of #1063

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
elBoberido, line 1: Please ignore this. Part of #1063

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/client_port_roudi.hpp
```diff
@@ -70,7 +70,7 @@ class ClientPortRouDi : public BasePort
     const MemberType_t* getMembers() const noexcept;
     MemberType_t* getMembers() noexcept;
 
-    void handleCaProProtocollViolation(iox::capro::CaproMessageType messageType) noexcept;
+    void handleCaProProtocolViolation(iox::capro::CaproMessageType messageType) noexcept;
```
elfenpiff, line 73: Please add a const to the argument and I think the method can be made const as well.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/client_port_roudi.hpp
```diff
@@ -70,7 +70,7 @@ class ClientPortRouDi : public BasePort
     const MemberType_t* getMembers() const noexcept;
```
elfenpiff, line 70: Please update the copyright year.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_data.hpp
```diff
@@ -22,7 +22,7 @@
 #include "iceoryx_posh/internal/mepoo/memory_manager.hpp"
```
elfenpiff, line 22: Please update the copyright year

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_roudi.hpp
```diff
@@ -62,6 +65,13 @@ class ServerPortRouDi : public BasePort
     const MemberType_t* getMembers() const noexcept;
     MemberType_t* getMembers() noexcept;
 
+    void handleCaProProtocolViolation(const capro::CaproMessageType messageType) const noexcept;
```
elfenpiff, line 68: I think this can be made static - I suggestion from clang-tidy.

iceoryx_posh/source/popo/ports/server_port_data.cpp
elfenpiff: I was wondering what the `0` means in this line and realized it is actually a `nullptr` and signals no memory manager set. Could you maybe introduce `constexpr mepoo::MemoryManager* INVALID_MEMORY_MANAGER = nullptr` in `ServerPortData` and use it here.

Otherwise other reviewers have no clue what the `0` means.

iceoryx_posh/source/popo/ports/server_port_data.cpp
```diff
@@ -23,13 +23,17 @@ namespace popo
...
+    , m_chunkReceiverData(cxx::VariantQueueTypes::SoFi_MultiProducerSingleConsumer,
```
elfenpiff, line 32: It seems that `cxx::VariantQueueTypes::SoFi_MultiProducerSingleConsumer` and `QueueFullPolicy::DISCARD_OLDEST_DATA` are values which can maybe change in a safety critical environment to no discard data and a fifo. Would it therefore make sense to introduce some public constexpr member variables which state this somehow?

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,19 +39,108 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
+}
+
 cxx::optional<capro::CaproMessage> ServerPortRouDi::tryGetCaProMessage() noexcept
 {
+    // get offer state request from user side
+    const auto offeringRequested = getMembers()->m_offeringRequested.load(std::memory_order_relaxed);
+
+    const auto isOffered = getMembers()->m_offered.load(std::memory_order_relaxed);
+
+    if (isOffered)
```
elfenpiff, line 54: A personal note, I would prefer a structure like
```cpp
if ( isOffered && !offeringRequested ) {
  capro::CaproMessage caproMessage(capro::CaproMessageType::STOP_OFFER, this->getCaProServiceDescription());
  return dispatchCaProMessageAndGetPossibleResponse(caproMessage);
} else if ( !isOffered && offeringRequested ) {
  capro::CaproMessage caproMessage(capro::CaproMessageType::OFFER, this->getCaProServiceDescription());
  return dispatchCaProMessageAndGetPossibleResponse(caproMessage);
}
```

Which is for me much more readable. Otherwise I have two additional if statements and my brain has to fiddle this whole thing together. I think with the approach above you cut the cyclomatic complexity in half.

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,19 +39,108 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
 
 cxx::optional<capro::CaproMessage>
-ServerPortRouDi::dispatchCaProMessageAndGetPossibleResponse(const capro::CaproMessage& /*caProMessage*/) noexcept
+ServerPortRouDi::dispatchCaProMessageAndGetPossibleResponse(const capro::CaproMessage& caProMessage) noexcept
 {
-    capro::CaproMessage responseMessage(
-        capro::CaproMessageType::NACK, this->getCaProServiceDescription(), capro::CaproMessageSubType::NOSUBTYPE);
+    const auto isOffered = getMembers()->m_offered.load(std::memory_order_relaxed);
+
+    if (isOffered)
```
elfenpiff, line 80: Personal taste: I would prefer something like:
```cpp
return (isOffered) ? handleCapProMessageForStateOffered(caProMessage)
                   : handleCapProMessageForStateNotOffered(caProMessage);
```

So two lines instead of 8.

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,19 +39,108 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
+    switch (caProMessage.m_type)
+    {
+    case capro::CaproMessageType::STOP_OFFER:
+        getMembers()->m_offered.store(false, std::memory_order_relaxed);
+        m_chunkSender.removeAllQueues();
+        return caProMessage;
+    case capro::CaproMessageType::OFFER:
+        return responseMessage;
+    case capro::CaproMessageType::CONNECT:
+        cxx::Expects(caProMessage.m_chunkQueueData != nullptr && "Invalid response queue passed to server");
```
elfenpiff, line 111: Wouldn't this statement kill roudi? So the only thing I have to do is to send an invalid connect request and everything goes down? Wouldn't it be better to send here a `NACK` and print an error into the log?

Would it make sense to also add some sanity check here and maybe verify if the `chunkQueueData` pointer is valid in the sense that it originates from a valid shared memory segment which is accessible? Otherwise some application can add a chunk queue data where the server port has no access to and then you kill the server with a segfault?

iceoryx_posh/source/popo/ports/server_port_user.cpp
```diff
@@ -39,15 +38,26 @@ ServerPortUser::MemberType_t* ServerPortUser::getMembers() noexcept
...
+    }
+
+    return cxx::success<const RequestHeader*>(static_cast<const RequestHeader*>(getChunkResult.value()->userHeader()));
 }
 
-void ServerPortUser::releaseRequest(const RequestHeader* const /*requestHeader*/) noexcept
+void ServerPortUser::releaseRequest(const RequestHeader* const requestHeader) noexcept
 {
-    /// @todo
+    cxx::Ensures(requestHeader != nullptr && "requestHeader must not be a nullptr");
```
elfenpiff, line 55: Could you please remove this. I would only print out an error log message since we can ensure here that we never reach an inconsistent state. Or maybe a call to the error handler with severity moderate?

iceoryx_posh/source/popo/ports/server_port_user.cpp
```diff
@@ -61,20 +71,72 @@ bool ServerPortUser::hasLostRequestsSinceLastCall() noexcept
...
 
 cxx::expected<ResponseHeader*, AllocationError>
-ServerPortUser::allocateResponse(const uint32_t /*userPayloadSize*/) noexcept
+ServerPortUser::allocateResponse(const RequestHeader* const requestHeader,
+                                 const uint32_t userPayloadSize,
+                                 const uint32_t userPayloadAlignment) noexcept
 {
-    /// @todo
-    return cxx::error<AllocationError>(AllocationError::RUNNING_OUT_OF_CHUNKS);
+    cxx::Ensures(requestHeader != nullptr && "requestHeader must not be a nullptr");
```
elfenpiff, line 78: What about returning `cxx::error` with an allocation error like invalid argument? Or event better we get the request header as reference and we can remove those lines and can guarantee that the request header is always not null?

iceoryx_posh/source/popo/ports/server_port_user.cpp
```diff
@@ -61,20 +71,72 @@ bool ServerPortUser::hasLostRequestsSinceLastCall() noexcept
...
+    if (requestHeader == nullptr)
```
elfenpiff, line 79: Those lines would be gone if request header would be a reference.

iceoryx_posh/source/popo/ports/server_port_user.cpp
```diff
@@ -61,20 +71,72 @@ bool ServerPortUser::hasLostRequestsSinceLastCall() noexcept
...
+    auto responseHeader =
+        new (allocateResult.value()->userHeader()) ResponseHeader(requestHeader->m_uniqueClientQueueId,
+                                                                  requestHeader->m_lastKnownClientQueueIndex,
+                                                                  requestHeader->getSequenceId());
+
+    return cxx::success<ResponseHeader*>(responseHeader);
 }
 
-void ServerPortUser::freeResponse(ResponseHeader* const /*responseHeader*/) noexcept
+void ServerPortUser::freeResponse(ResponseHeader* const responseHeader) noexcept
```
elfenpiff, line 102: Please use a reference here, this would remove the `nullptr` check, the `cxx::Ensures` and simplify the code and the error handling at all.

iceoryx_posh/source/popo/ports/server_port_user.cpp
```diff
@@ -61,20 +71,72 @@ bool ServerPortUser::hasLostRequestsSinceLastCall() noexcept
...
+    cxx::Ensures(responseHeader != nullptr && "responseHeader must not be a nullptr");
+    if (responseHeader == nullptr)
+    {
+        return;
+    }
+    m_chunkSender.release(responseHeader->getChunkHeader());
 }
 
-void ServerPortUser::sendResponse(ResponseHeader* const /*responseHeader*/) noexcept
+void ServerPortUser::sendResponse(ResponseHeader* const responseHeader) noexcept
```
elfenpiff, line 112: Again, please use a reference to get rid of the `nullptr` check and additional error handling.

elfenpiff (changes requested):
Tests will follow after lunch ...

iceoryx_posh/source/popo/ports/server_port_data.cpp
elBoberido: @elfenpiff you are wrong and by being wrong you showed why magic numbers should be avoided. This is the history request and I will add a constexpr for this

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_roudi.hpp
elBoberido: I extended the debug output to make it more similar to `ClientPortUser` which requires an access to members and therefore it must remain a method.

iceoryx_posh/source/popo/ports/server_port_data.cpp
elfenpiff: @elBoberido I just wanted to check if you are paying attention ;)

iceoryx_posh/source/popo/ports/server_port_data.cpp
elBoberido: Good question. We don't do this for pub-sub either. That does not mean we shouldn't do it but the question is if an additional constexpr will be of much use. I guess if we ever change this, it would be in the context of adding an additional field to `ServerOptions`.

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
elBoberido: Good point

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_user.hpp
```diff
@@ -51,9 +52,9 @@ class ServerPortUser : public BasePort
 
     /// @brief Tries to get the next request from the queue. If there is a new one, the ChunkHeader of the oldest
     /// request in the queue is returned (FiFo queue)
-    /// @return optional that has a new chunk header or no value if there are no new requests in the underlying queue,
+    /// @return cxx::expected that has a new RequestHeader if there are new requests in the underlying queue,
     /// ChunkReceiveResult on error
-    cxx::expected<cxx::optional<const RequestHeader*>, ChunkReceiveResult> getRequest() noexcept;
+    cxx::expected<const RequestHeader*, ChunkReceiveResult> getRequest() noexcept;
```
elfenpiff, line 57: In theory every user has to perform a `nullptr` check before accessing the success expected. Would it be possible to return here a `std::reference_wrapper<const RequestHeader>` to avoid exactly this? Or is it a valid use case that the request header is a nullptr?

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_user.hpp
```diff
@@ -69,10 +70,14 @@ class ServerPortUser : public BasePort
 
     /// @brief Allocate a response, the ownerhip of the SharedChunk remains in the ServerPortUser for being able to
     /// cleanup if the user process disappears
+    /// @param[in] requestHeader, the request header for the corresponding response
     /// @param[in] userPayloadSize, size of the user user-paylaod without additional headers
+    /// @param[in] userPayloadAlignment, alignment of the user user-paylaod without additional headers
     /// @return on success pointer to a ChunkHeader which can be used to access the chunk-header, user-header and
     /// user-payload fields, error if not
-    cxx::expected<ResponseHeader*, AllocationError> allocateResponse(const uint32_t userPayloadSize) noexcept;
+    cxx::expected<ResponseHeader*, AllocationError> allocateResponse(const RequestHeader* const requestHeader,
```
elfenpiff, line 78: Same here, the user has to perform a `nullptr` when success is returned. Could we add here a `std::reference_wrapper<ResponseHeader>` as well?

iceoryx_posh/source/popo/ports/server_port_data.cpp
elfenpiff: Could we add a type alias for the queue type and a constexpr for the behavior to `ServerOptions` for now and use them as default?

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
elBoberido: I disagree on this. It is not easily visible when the else branch is taken since it does not only depend on one "complex" condition on the `else if` but one has also to parse the `if` condition and suddenly there is a 2x2 matrix which decides if the branch is taken. Whit the current code, one can easily see that the `else` branch is only taken when not offered and this matches the logic from `dispatchCaProMessageAndGetPossibleResponse` ... let's wait for the opinion of the second reviewer. There is a good chance to get a third opinion.

iceoryx_posh/source/popo/ports/server_port_user.cpp
elBoberido: I will remove it from here, but the `ChunkReceiver` will call the error handler with error level `SEVERE` anyway.

iceoryx_posh/source/popo/ports/server_port_user.cpp
elBoberido: Good point. I didn't thought about extending `AllocationError` and was hesitant to introduce another error enum but extending the `AllocationError` enum would be the best option

iceoryx_posh/test/moduletests/test_popo_server_port_common.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+        constexpr uint64_t NUMBER_OF_REQUESTS{1U};
+        pushRequests(sut.requestQueuePusher, NUMBER_OF_REQUESTS);
+        auto requestResult = sut.portUser.getRequest();
+        ASSERT_FALSE(requestResult.has_error());
+        auto requestHeader = requestResult.value();
+
+        sut.portUser.allocateResponse(requestHeader, USER_PAYLOAD_SIZE, USER_PAYLOAD_ALIGNMENT)
+            .and_then([&](auto& responseHeader) { testFunction(requestHeader, responseHeader); })
+            .or_else([&](const auto& error) { FAIL() << "Expected ResponseHeader but got error: " << error; });
```
elfenpiff, line 170: Could you please use `GTEST_FAIL()` instead of `FAIL` in the whole test code base. According to `gtest.h` it was just defined since `FAIL` clashes sometimes with other libraries. Then it is also clear that you are using a gtest macro and not some custom defined one.

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+    ::testing::Test::RecordProperty("TEST_ID", "15a399ee-b162-4b42-8aab-da13571fb478");
+    auto& sut = serverPortWithOfferOnCreate;
+
+    sut.portUser.offer();
+
+    sut.portRouDi.tryGetCaProMessage()
+        .and_then(
+            [&](const auto& caproMessage) { FAIL() << "Expected no CaPro message but got: " << caproMessage.m_type; })
+        .or_else([&]() { SUCCEED(); });
```
elfenpiff, line 81: Could you please use `GTEST_SUCCEED` instead of `SUCCEED`.

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+    auto& sut = serverPortWithoutOfferOnCreate;
+
+    sut.portUser.offer();
+
+    sut.portRouDi.tryGetCaProMessage()
+        .and_then([&](const auto& caproMessage) { EXPECT_THAT(caproMessage.m_type, Eq(CaproMessageType::OFFER)); })
+        .or_else([&]() { FAIL() << "Expected CaPro message but got none"; });
+}
+
+TEST_F(ServerPort_test, TryGetCaProMessageOnOfferWhenOfferingHasNullopt)
```
elfenpiff, line 71: The name is a bit misleading. What about `TryGetCaProMessageOnOfferWhenPortAlreadyOffers`

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    // this is not part of the server port but holds the chunk from `sendResponse`
+    clientResponseQueue.clear();
+
+    EXPECT_THAT(getNumberOfUsedChunks(), Eq(0U));
+}
+
+// BEGIN tryGetCaProMessage tests
+
+TEST_F(ServerPort_test, TryGetCaProMessageOnOfferWhenNotOfferingHasCaProMessageTypeOffer)
```
elfenpiff, line 59: What about `TryGetCaProMessageOnOfferWhenPortIsNotOffering`?

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    sut.portUser.offer();
+
+    sut.portRouDi.tryGetCaProMessage()
+        .and_then(
+            [&](const auto& caproMessage) { FAIL() << "Expected no CaPro message but got: " << caproMessage.m_type; })
+        .or_else([&]() { SUCCEED(); });
+}
+
+TEST_F(ServerPort_test, TryGetCaProMessageOnStopOfferWhenOfferingHasCaProMessageTypeStopOffer)
```
elfenpiff, line 84: What about `TryGetCaProMessageOnStopOfferWhenPortIsOffering`?

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+    auto& sut = serverPortWithOfferOnCreate;
+
+    sut.portUser.stopOffer();
+
+    sut.portRouDi.tryGetCaProMessage()
+        .and_then([&](const auto& caproMessage) { EXPECT_THAT(caproMessage.m_type, Eq(CaproMessageType::STOP_OFFER)); })
+        .or_else([&]() { FAIL() << "Expected CaPro message but got none"; });
+}
+
+TEST_F(ServerPort_test, TryGetCaProMessageOnStopOfferWhenNotOfferingHasNullopt)
```
elfenpiff, line 96: What about `TryGetCaProMessageOnStopOfferWhenPortIsNotOffering`?

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+        .or_else([&]() { SUCCEED(); });
+}
+
+// END tryGetCaProMessage tests
+
+// BEGIN test CaPro transitions
+
+TEST_F(ServerPort_test, StateNotOfferedWithCaProMessageTypeConnectReactsWithNack)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "ceaef856-2a8d-46c0-9167-fe1ca6fad736");
```
elfenpiff, line 115: Could we pack this and the following test cases until `StateNotOfferedWithCaProMessageTypeStopOfferReactsWithNack` into a function with `CaproMessageType` as argument?

iceoryx_posh/source/popo/ports/server_port_user.cpp
elBoberido: Like I already mentioned in the PR description, I'm not that happy with the interface. It is modeled after the pub-sub interface where we also use pointer. I think in this case we also cannot easily use references since it make the user interface also error prone. Since this is used in combination with `getRequest`, which returns a pointer, it would be kind of awkward to call `releaseRequest(*req)` so we should also return a reference with `getRequest`. But one can easily write `auto req = requestResult.value()` and create a copy on the stack which results in an error when `releaseRequest` is called. One solution to this problem might be to use a `std::reference_wrapper`. This would fix the problem of the unintentional copy and would guarantee to not be `null`. I do not fully estimate the consequences of this change and if we do it, we should also do it for pub-sub. Therefore I would suggest to keep it as is for the next release and create an issue to replace the raw pointer with `std::reference_wrapper` where appropriate or create an own abstraction if `std::reference_wrapper` does not fulfill our needs.

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    sut.portRouDi.dispatchCaProMessageAndGetPossibleResponse(caproMessage)
+        .and_then([&](const auto& responseCaproMessage) {
+            EXPECT_THAT(responseCaproMessage.m_serviceDescription, Eq(sut.portData.m_serviceDescription));
+            EXPECT_THAT(responseCaproMessage.m_type, Eq(iox::capro::CaproMessageType::NACK));
+        })
+        .or_else([&]() { FAIL() << "Expected CaPro message but got none"; });
+}
+
+TEST_F(ServerPort_test, StateNotOfferedWithCaProMessageTypeStopOfferReactsWithNack)
```
elfenpiff, line 143: I think you miss a test for a capro protocol violation. Could you add this for state not offered and state offered.
Maybe you could send `CaproMessageType::SUB` or `CaproMessageType::PING` and check for the violation.

iceoryx_posh/source/popo/ports/server_port_user.cpp
elBoberido: I just remembered why I did the check here instead of letting the next layer call the error handler. The problem is that the pointer will be de-referenced before `release` is called and therefore the check must remain here. I can create a `cxx::error` here but still with error level `SEVERE`. What do you think?

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_user.hpp
elBoberido: No, it should never be a `nullptr`. Please se my comment [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1033#discussion_r800678244) and let's discuss it there.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+    ::testing::Test::RecordProperty("TEST_ID", "30e36e69-4f23-41ee-aac7-d34240f075ae");
+    auto& sut = serverPortWithOfferOnCreate;
+
+    constexpr uint64_t NUMBER_OF_REQUESTS{2U};
+    pushRequests(sut.requestQueuePusher, NUMBER_OF_REQUESTS);
+
+    EXPECT_TRUE(sut.portUser.hasNewRequests());
+}
+
+TEST_F(ServerPort_test, HasNewRequestsWithMultipleRequestsAndAllButOneRemovedIsTrue)
```
elfenpiff, line 191: Could you add a test for the limits. `HasNewRequestsIsTrueWithMaximumAmountOfRequests`.

iceoryx_posh/test/moduletests/test_popo_server_port_roudi.cpp
elBoberido: What do you think about an alternative like
```cpp
for(const auto caproMessageType: { CaproMessageType::CONNECT,
                                   CaproMessageType::DISCONNECT,
                                   CaproMessageType::STOP_OFFER})
{
    SCOPED_TRACE(caproMessageType);
    // here comes the test
}
```

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+// END hasNewRequests tests
+
+// BEGIN getRequest tests
+
+TEST_F(ServerPort_test, GetRequestWithoutOfferResultsIn_NO_CHUNK_AVAILABLE)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "6c555224-8b03-46c5-a988-059ac2656149");
+    auto& sut = serverPortWithoutOfferOnCreate;
+
+    sut.portUser.getRequest()
```
elfenpiff, line 225: Could we maybe adjust the API here. I would introduce a `ServerPortRequestResult` which contains the `ChunkReceiveResult` values as a subset and adds `SERVER_DOES_NOT_OFFER`. You could also add a translation function which translates the `ChunkReceiveResult` into a `ServerPortRequestResult` so that it can be handled easier.

In this case I would then recommend to return `ServerPortRequestResult::SERVER_DOES_NOT_OFFER` which states the actual reason.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+        .and_then([&](const auto&) { FAIL() << "Expected ChunkReceiveResult::NO_CHUNK_AVAILABLE but got request"; })
+        .or_else([&](const auto& error) { EXPECT_THAT(error, Eq(ChunkReceiveResult::NO_CHUNK_AVAILABLE)); });
+}
+
+TEST_F(ServerPort_test, GetRequestWithNoRequestsResultsIn_NO_CHUNK_AVAILABLE)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "2b78536b-6902-4e55-aef2-654da5afdb80");
+    auto& sut = serverPortWithOfferOnCreate;
+
+    sut.portUser.getRequest()
```
elfenpiff, line 235: Here I would add another value to `ServerPortRequestResult` `NO_REQUESTS_PENDING`.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+            EXPECT_THAT(errorLevel, Eq(iox::ErrorLevel::SEVERE));
+            detectedError.emplace(error);
+        });
+
+    sut.portUser.releaseRequest(static_cast<const RequestHeader*>(sharedChunk.getChunkHeader()->userHeader()));
+
+    EXPECT_TRUE(detectedError.has_value());
+}
+
+TEST_F(ServerPort_test, ReleaseRequestWithNullptrRequestHeaderCallsTheErrorHandler)
```
elfenpiff, line 396: When we use references we could remove this test.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+}
+
+TEST_F(ServerPort_test,
+       AllocateResponseWithInvalidPayloadParameterReturns_INVALID_PARAMETER_FOR_USER_PAYLOAD_OR_USER_HEADER)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "6015d3d6-73ae-46fe-994d-445f3f8f366d");
+    auto& sut = serverPortWithOfferOnCreate;
+
+    constexpr uint64_t USER_PAYLOAD_SIZE{23U};
+    constexpr uint64_t USER_PAYLOAD_ALIGNMENT{15U};
```
elfenpiff, line 524: Am I right that here the payload alignment is the only invalid parameter. Could you please rename it into `USER_INVALID_PAYLOAD_ALIGNMENT`.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+        sut.portUser.freeResponse(res);
+        // since the response is already freed, it should not be in the UsedChunkList anymore and the error handler
+        // should be called
+        sut.portUser.freeResponse(res);
+    });
+
+    EXPECT_TRUE(detectedError.has_value());
+}
+
+TEST_F(ServerPort_test, FreeResponseWithWithNullptrResponseHeaderCallsTheErrorHandler)
```
elfenpiff, line 590: Again we could remove this if we switch to references.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+
+    EXPECT_TRUE(detectedError.has_value());
+}
+
+TEST_F(ServerPort_test, SendResponseWithWithoutOfferReleasesTheChunkToTheMempool)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "dc4e31b1-18bf-4c42-9084-dd7abd52609b");
+    auto& sut = serverPortWithoutOfferOnCreate;
+
+    allocateResponseWithRequestHeaderAndThen(sut, [&](const auto, auto res) {
```
elfenpiff, line 635: Here I would prefer to inform the user that the send has not succeeded. Could we return a `bool` in `sendResponse` which returns `false` when the port is not in the offer state.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: I'm a bit torn on this. We also do not do it for `SubscriberPort` when not subscribed and just return a `NO_CHUNK_AVAILABLE`. I guess this is due to the fact that we can be unsubscribed even after calling `subscribe` if there is no publisher. With the `ServerPort` it is a bit different, since the port controls the state by itself and it is indeed binary. I was hesitant about creating a separate error enum since the API wouldn't mirror the pub-sub and client errors but would have an own one.
What do you think after knowing my considerations? I'm happy to change it if you think the more explicit error message outweighs the familiarity of the API.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: The payload size is also invalid since it must be a multiple of the alignment

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: This again boils down to the question if we want familiarity with the pub-sub API?

elfenpiff: No, I see a good API as a process and we converge with every feature and every PR to a more fitting API which covers better and better our use case.
Therefore, I would say we start with the pub-sub API here but develop it further and when this plays out well we create an issue to refine the pub sub API.

With this we improve the API step by step since we cannot think of everything right from the beginning.

iceoryx_posh/source/popo/ports/server_port_user.cpp
elBoberido: @elfenpiff I created #1071 and will close all your comments regarding references if you agree?

iceoryx_posh/test/moduletests/test_popo_server_port_common.hpp
elBoberido: Done for the files from this PR and created #1072 to replace it from the remaining code base since this PR is already quite big

iceoryx_posh/source/popo/ports/server_port_data.cpp
elBoberido: I think I need to add this option to the server. Currently there is a `serverTooSlowPolicy` option for the client which does not make sense if there isn't a `requestQueueFullPolicy` on the server side. Will add this to the options but this can only be tested with integration tests and the discovery loop which connects the clients and servers.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,732 @@
...
+{
+    ::testing::Test::RecordProperty("TEST_ID", "45823507-b83b-496b-965e-a48bd3c07b9e");
+    auto& sut = serverPortWithOfferOnCreate;
+
+    // the client is not yet connected to the `clientResponseQueue` which ID is used to send the responses to
+    allocateResponseWithRequestHeaderAndThen(sut, [&](const auto, auto res) {
+        constexpr uint64_t NUMBER_OF_REQUEST_CHUNKS{1U};
+        constexpr uint64_t NUMBER_OF_RESPONSE_CHUNKS{1U};
+        EXPECT_THAT(this->getNumberOfUsedChunks(), Eq(NUMBER_OF_REQUEST_CHUNKS + NUMBER_OF_RESPONSE_CHUNKS));
+        sut.portUser.sendResponse(res);
```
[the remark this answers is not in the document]
elBoberido: After some thinking, I suggest a different API. Since you mentioned you want this to be able to react on potentially lost Responses, e.g. because someone requested a resource, a failed transmission should return a response. The signature of `sendResponse` would become `cxx::expected<ResponseHeader*> sendResponse(ResponseHeader* res);` or `cxx::expected<(ResponseHeader*, ResponseTransmissionFailure)> sendResponse(ResponseHeader* res);`

What do you think?

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: I added a test. I'm not quite sure how much these functions need to be tested since they are delegating a call to the `ChunkReceiver` which is already extensively tested.

iceoryx_posh/source/popo/ports/server_port_data.cpp
```diff
@@ -21,15 +21,28 @@ namespace iox
 {
 namespace popo
 {
+/// @todo iox-#27 remove when the port queue policies are consolidated
```
elBoberido, line 24: Since this PR is already quite big, I didn't want to do this cleanup now since it will affect quite many files. If you agree, I will create a follow up PR while this PR is still in review

elBoberido (commented):
@elfenpiff please check my responses to your findings. I implemented everything but the API changes. I left some fun for tomorrow

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
```diff
@@ -38,19 +39,117 @@ ServerPortRouDi::MemberType_t* ServerPortRouDi::getMembers() noexcept
...
+        return responseMessage;
+    case capro::CaproMessageType::DISCONNECT:
+        m_chunkSender.tryRemoveQueue(static_cast<ClientChunkQueueData_t*>(caProMessage.m_chunkQueueData))
+            .and_then([&responseMessage]() { responseMessage.m_type = capro::CaproMessageType::ACK; });
+        return responseMessage;
+    default:
+        break;
+    }
+
+    handleCaProProtocolViolation(caProMessage.m_type);
```
elfenpiff, line 131: Could you please move this into the `default` case like in the method below. It is a bit more consistent and clear that when no value fits its automatically a protocol violation.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elfenpiff: You are most likely right but better safe than sorry.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elfenpiff: Oh I didn't know that this was required. Could you then please add to both variables `INVALID`.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elfenpiff: To summarize for `getRequest` I follow your argumentation and you can leave it has you mentioned it earlier but here I would really like to inform the user. What if a server allocates some resource and thinks, because of the lack of error, that this resource was send to the client.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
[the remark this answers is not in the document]
elfenpiff: I prefer `cxx::expected<(ResponseHeader*, ResponseTransmissionFailure)> sendResponse(ResponseHeader* res);` Then the error enum would clearly state to other code readers whats going on.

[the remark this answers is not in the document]
elBoberido: I will go this route but with a struct instead of a tuple

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_roudi.hpp
```diff
@@ -36,14 +37,16 @@ class ServerPortRouDi : public BasePort
...
-    explicit ServerPortRouDi(cxx::not_null<MemberType_t* const> serverPortDataPtr) noexcept;
+    explicit ServerPortRouDi(MemberType_t& serverPortData) noexcept;
 
     ServerPortRouDi(const ServerPortRouDi& other) = delete;
     ServerPortRouDi& operator=(const ServerPortRouDi&) = delete;
     ServerPortRouDi(ServerPortRouDi&& rhs) noexcept = default;
     ServerPortRouDi& operator=(ServerPortRouDi&& rhs) noexcept = default;
     ~ServerPortRouDi() = default;
 
+    ConsumerTooSlowPolicy getClientTooSlowPolicy() const noexcept;
```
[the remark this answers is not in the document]
elBoberido: One is the generic enum name and the other one gives more context to the enum since it is also used at different places. For me it's like having and `int64_t` as data type and calling the variable `myInt64`. It is technically correct but doesn't give context

iceoryx_posh/include/iceoryx_posh/popo/server_options.hpp
```diff
@@ -41,6 +41,10 @@ struct ServerOptions
     /// @brief The option whether the server should already be offered when creating it
     bool offerOnCreate{true};
 
+    /// @brief The option whether the client should block when the request queue is full
+    /// @note Corresponds with ClientOptions::serverTooSlowPolicy
+    QueueFullPolicy2 requestQueueFullPolicy{QueueFullPolicy2::DISCARD_OLDEST_DATA};
```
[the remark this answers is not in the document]
elBoberido: Nope, this is permanent. There are two `ChunkSender` and two `ChunkReceiver` and one might want or need to configure the behavior of both directions independently.

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
elBoberido: Then I would still need to have it here since there are cases where a branch does not return and leaves the switch statement. Alternatively I would need to ensure that all branches return which is also error prone when new branches are added. I'd rather move the `handleCaProProtocolViolation` out the the default case from the method below. What do you think?

elfenpiff (commented):
Could you please update also the `doc/design/request_response_communication.md` design document. Then the API changes are in sync with the design document and now we know what has been changed so it is easier to adjust then doing it afterwards where we have to check everything.

iceoryx_posh/source/popo/ports/server_port_roudi.cpp
elBoberido: @elfenpiff ping :)

iceoryx_posh/source/popo/ports/server_port_user.cpp
elBoberido: @elfenpiff ping :)

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: done

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: done

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
elBoberido: not yet done, but the comment below has the same request and the discussion happens there

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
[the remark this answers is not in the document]
elBoberido: It seems to be non-trivial to implement this. I added a todo and will implement it in a separate PR since some building blocks need adjustments before the overflow value can be returned

iceoryx_posh/source/popo/ports/server_port_data.cpp
elBoberido: Corresponding PR is #1075

elfenpiff (approved):

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/server_port_roudi.hpp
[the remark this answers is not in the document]
MatthiasKillat: In the long-run it might be better to return the options to the reader with one method instead of multiple individual getter methods. Read only access can be "ensured" by const ref or copy (more expensive).

But it is not important now.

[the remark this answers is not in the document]
elBoberido: @MatthiasKillat it might be necessary with the integration into RouDi and the discovery loop.

MatthiasKillat (commented):
Implementation looks fine but I did not review the test cases yet (so cannot approve). If this is needed and you need further reviewers let me know @elBoberido

iceoryx_posh/source/popo/ports/server_port_data.cpp
```diff
@@ -21,15 +21,28 @@ namespace iox
...
 ServerPortData::ServerPortData(const capro::ServiceDescription& serviceDescription,
                                const RuntimeName_t& runtimeName,
-                               const NodeName_t& nodeName,
+                               const ServerOptions& serverOptions,
                                mepoo::MemoryManager* const memoryManager,
                                const mepoo::MemoryInfo& memoryInfo) noexcept
-    : BasePortData(serviceDescription, runtimeName, nodeName)
-    , m_chunkSenderData(memoryManager, SERVER_SUBSCRIBER_POLICY, 0, memoryInfo)
-    , m_chunkReceiverData(cxx::VariantQueueTypes::FiFo_MultiProducerSingleConsumer, SERVER_PUBLISHER_POLICY)
+    : BasePortData(serviceDescription, runtimeName, serverOptions.nodeName)
```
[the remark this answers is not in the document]
elBoberido: Yes, it might be necessary when the new ports are integrated into the discovery loop. Will do it on demand. This is also related to #1048

elfenpiff (dismissed):

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,836 @@
...
+TEST_F(ServerPort_test, SendResponseWithWithoutOfferReleasesTheChunkToTheMempool)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "dc4e31b1-18bf-4c42-9084-dd7abd52609b");
+    auto& sut = serverPortWithoutOfferOnCreate;
+
+    allocateResponseWithRequestHeaderAndThen(sut, [&](const auto, auto res) {
+        constexpr uint64_t NUMBER_OF_REQUEST_CHUNKS{1U};
+        constexpr uint64_t NUMBER_OF_RESPONSE_CHUNKS{1U};
+        EXPECT_THAT(this->getNumberOfUsedChunks(), Eq(NUMBER_OF_REQUEST_CHUNKS + NUMBER_OF_RESPONSE_CHUNKS));
+        sut.portUser.sendResponse(res);
```
MatthiasKillat, line 684: Can this even happen without the server having offered? I would think the server responds automatically when offered (and resources like memory chunks are sufficient). But I may be wrong here and testing that we have no leak either way makes sense anyway.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,836 @@
...
+    IOX_DISCARD_RESULT(sut.portUser.getRequest());
+
+    EXPECT_FALSE(sut.portUser.hasNewRequests());
+}
+
+// END hasNewRequests tests
+
+// BEGIN getRequest tests
+
+TEST_F(ServerPort_test, GetRequestWithoutOfferResultsIn_NO_CHUNK_AVAILABLE)
```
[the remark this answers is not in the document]
elBoberido: Right, the enum was changed after the test was written. Fixed

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,836 @@
...
+    constexpr uint64_t REQUEST_DATA_1{13};
+    constexpr uint64_t REQUEST_DATA_2{73};
+
+    constexpr uint64_t NUMBER_OF_REQUESTS{1U};
+    pushRequests(sut.requestQueuePusher, NUMBER_OF_REQUESTS, REQUEST_DATA_1);
+    IOX_DISCARD_RESULT(sut.portUser.getRequest());
+    pushRequests(sut.requestQueuePusher, NUMBER_OF_REQUESTS, REQUEST_DATA_2);
+
+    sut.portUser.getRequest()
+        .and_then([&](const auto& req) { EXPECT_THAT(this->getRequestData(req), Eq(REQUEST_DATA_2)); })
```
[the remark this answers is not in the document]
elBoberido: Okay, this kind of confusing. The API returns a `ReqeustHeader*` and the check that we really received the requests in the correct order is done by the counter in the payload of the request. Therefore the `getRequestData`.

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,836 @@
...
+    auto& sut = serverPortWithOfferOnCreate;
+
+    pushRequests(sut.requestQueuePusher, QUEUE_CAPACITY + 1);
+    IOX_DISCARD_RESULT(sut.portUser.hasLostRequestsSinceLastCall());
+    pushRequests(sut.requestQueuePusher, 1);
+
+    EXPECT_TRUE(sut.portUser.hasLostRequestsSinceLastCall());
+}
+
+TEST_F(ServerPort_test, HasLostRequestsSinceLastCallWhenNoRequestAreLostAfterRemovingRequestFromQueueReturnsTrue)
```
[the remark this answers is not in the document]
elBoberido: True ... I mean false is correct. C&P issue

iceoryx_posh/test/moduletests/test_popo_server_port_user.cpp
```diff
@@ -0,0 +1,836 @@
...
+{
+    ::testing::Test::RecordProperty("TEST_ID", "248ddc67-5717-4b33-8b04-80d8246fedb3");
+    auto& sut = serverPortWithOfferOnCreate;
+
+    pushRequests(sut.requestQueuePusher, QUEUE_CAPACITY + 1);
+    IOX_DISCARD_RESULT(sut.portUser.hasLostRequestsSinceLastCall());
```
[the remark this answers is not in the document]
elBoberido: Yes, the test name was wrong and you are correct. We cause a lost request with line 503 and need to reset the indicator. Afterward a request is taken out of the queue and another one is pushed. This should not result in a lost request

PR #1043, opened 2022-01-25
Iox #1029 implement strict history capacity matching

### LINKED ISSUE #1029: Introduce strict matching mode for historyCapacity
## Brief feature description

Currently we allow matching of subscribers with HistoryRequest > HistoryCapacity of the publisher. Strictly speaking asking for a contract which can never be fulfilled (even if the publisher had that many samples before) and hence could be forbidden under certain conditions. This covers the special case of `HistoryRequest > HistoryCapacity = 0`.

## Detailed information

We will introduce a flag `requirePublisherHistorySupport` to define whether we want to strictly enforce `HistoryRequest <= HistoryCapacity`. If this flag is `true`, publishers and subscribers not satisfying the inequality will not be connected. 

By default it will be set to `false` which leads to the current behavior, i.e. does not break any existing systems.

## Use Case
Matching between Readers and Writers in Cyclone DDS where at least one of them has durability TRANSIENT LOCAL (out of TRANSIENT LOCAL and VOLATILE).

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #1029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - Extend cxx::optional constructor for in place construction so that copy/move for values inside the optional even could be deleted[\#967](https://github.com/eclipse-iceoryx/iceoryx/issues/967)
 - Add templated `from`/`into` free functions to formalize conversions from enums and other types [#992](https://github.com/eclipse-iceoryx/iceoryx/issues/992)
 - UniqueId class for unique IDs within a process [#1010](https://github.com/eclipse-iceoryx/iceoryx/issues/1010)
-- Add requirePublisherHistorySupport option at subscriber side (if set to true requires historyRequest <= historyCapacity to be eligible for connection)
+- Add requirePublisherHistorySupport option at subscriber side (if set to true requires historyRequest <= historyCapacity to be eligible for connection) [#1029](https://github.com/eclipse-iceoryx/iceoryx/issues/1029)
 
 **Bugfixes:**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Extend cxx::optional constructor for in place construction so that copy/move for values inside the optional even could be deleted[\#967](https://github.com/eclipse-iceoryx/iceoryx/issues/967)
 - Add templated `from`/`into` free functions to formalize conversions from enums and other types [#992](https://github.com/eclipse-iceoryx/iceoryx/issues/992)
 - UniqueId class for unique IDs within a process [#1010](https://github.com/eclipse-iceoryx/iceoryx/issues/1010)
+- Add requirePublisherHistorySupport option at subscriber side (if set to true requires historyRequest <= historyCapacity to be eligible for connection)
 
 **Bugfixes:**
 
--- a/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
@@ -45,6 +45,11 @@ typedef struct
     /// @brief describes whether a publisher blocks when subscriber queue is full
     ENUM iox_QueueFullPolicy queueFullPolicy;
 
+    /// @brief Indicates whether we require the publisher to have historyCapacity >= historyRequest.
+    ///        If true and the condition is not met (i.e. historyCapacity < historyRequest), the subscriber will
+    ///        not be connected to the publisher.
+    bool requirePublisherHistorySupport;
+
     /// @brief this value will be set exclusively by iox_sub_options_init and is not supposed to be modified otherwise
     uint64_t initCheck;
 } iox_sub_options_t;
--- a/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
@@ -1,5 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2020 - 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2020 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_examples/iceoptions/README.md
+++ b/iceoryx_examples/iceoptions/README.md
@@ -72,11 +72,20 @@ would encounter an overflow, the oldest sample is released to create space for t
 subscriberOptions.queueCapacity = 10U;
 ```
 
-`historyRequest` will enable a subscriber to receive the last n samples on subscription e.g. in case it was started later than the publisher. The publisher needs to have its `historyCapacity` enabled, too.
+`historyRequest` will enable a subscriber to receive the last n samples on subscription e.g. in case it was started later than the publisher.
+If the publisher does not have a sufficient `historyCapacity` (larger than `historyRequest`), it will still be connected but we will not be able to
+receive the requested amount of historical data (if it was available). Instead we will receive the largest amount of historical sample
+the publisher has available, i.e. best-effort.
+
+If we want to enforce the contract that the publisher needs to support a `historyCapacity` of at least `historyRequest`, we can do so by setting
+`requirePublisherHistorySupport` to `true`. In this case the subscriber will only connect if the publisher history support is at least as large as is request.
+By default this is set to `false` and best-effort behavior is used.
 
 <!--[geoffrey][iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp][history]-->
 ```cpp
 subscriberOptions.historyRequest = 5U;
+
+subscriberOptions.requiresPublisherHistorySupport = false;
 ```
 
 Topics are automatically subscribed on creation, if you want to disable that feature and control the subscription
--- a/iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp
+++ b/iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 - 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2020 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp
+++ b/iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp
@@ -41,8 +41,14 @@ int main()
     // publisher has send. The history request ensures that we at least get the last 5
     // samples sent by the publisher when we subscribe (if at least 5 were already sent
     // and the publisher has history enabled).
+
+    // we do not require the publisher to support the history we request,
+    // i.e. we will still connect even if its historyCapacity is smaller than what we request
+
     //! [history]
     subscriberOptions.historyRequest = 5U;
+
+    subscriberOptions.requiresPublisherHistorySupport = false;
     //! [history]
 
     // when the subscriber is created, no attempts are made to connect to any publishers that may exist
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/ports/pub_sub_port_types.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/ports/pub_sub_port_types.hpp
@@ -1,0 +1,39 @@
+// Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+#ifndef IOX_POSH_POPO_PORTS_PUB_SUB_PORT_TYPES_HPP
+#define IOX_POSH_POPO_PORTS_PUB_SUB_PORT_TYPES_HPP
+
+#include "iceoryx_posh/iceoryx_posh_types.hpp"
+#include "iceoryx_posh/internal/popo/building_blocks/chunk_receiver_data.hpp"
+#include "iceoryx_posh/internal/popo/building_blocks/locking_policy.hpp"
+
+#include <cstdint>
+
+namespace iox
+{
+namespace popo
+{
+/// @todo iox#1051 move definitions for publish subscribe communication here
+
+using SubscriberChunkQueueData_t = ChunkQueueData<DefaultChunkQueueConfig, ThreadSafePolicy>;
+
+using SubscriberChunkReceiverData_t =
+    ChunkReceiverData<MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY, SubscriberChunkQueueData_t>;
+
+} // namespace popo
+} // namespace iox
+
+#endif // IOX_POSH_POPO_PORTS_PUB_SUB_PORT_TYPES_HPP
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
@@ -50,6 +50,9 @@ struct PublisherPortData : public BasePortData
         ChunkSenderData<MAX_CHUNKS_ALLOCATED_PER_PUBLISHER_SIMULTANEOUSLY, ChunkDistributorData_t>;
 
     ChunkSenderData_t m_chunkSenderData;
+
+    PublisherOptions m_options;
+
     std::atomic_bool m_offeringRequested{false};
     std::atomic_bool m_offered{false};
 };
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2020 - 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2020 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_posh/include/iceoryx_posh/popo/subscriber_options.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/popo/subscriber_options.hpp
@@ -17,7 +17,7 @@
 #ifndef IOX_POSH_POPO_SUBSCRIBER_OPTIONS_HPP
 #define IOX_POSH_POPO_SUBSCRIBER_OPTIONS_HPP
 
-#include "iceoryx_posh/internal/popo/ports/subscriber_port_data.hpp"
+#include "iceoryx_posh/iceoryx_posh_types.hpp"
 #include "port_queue_policies.hpp"
 
 #include "iceoryx_hoofs/cxx/serialization.hpp"
@@ -33,7 +33,7 @@ struct SubscriberOptions
 {
     /// @brief The size of the receiver queue where chunks are stored before they are passed to the user
     /// @attention Depending on the underlying queue there can be a different overflow behavior
-    uint64_t queueCapacity{SubscriberPortData::ChunkQueueData_t::MAX_CAPACITY};
+    uint64_t queueCapacity{iox::MAX_SUBSCRIBER_QUEUE_CAPACITY};
 
     /// @brief The max number of chunks received after subscription if chunks are available
     uint64_t historyRequest{0U};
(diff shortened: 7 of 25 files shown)


iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -585,6 +584,163 @@ TEST_F(PortManager_test, DoDiscoveryPublisherCanWaitAndSubscriberDiscardOldestLe
     EXPECT_THAT(subscriber.getSubscriptionState(), Eq(iox::SubscribeState::SUBSCRIBED));
 }
 
+TEST_F(PortManager_test, SubscriberRequiringHistoryDoesNotConnectToPublisherWithoutHistorySupport)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "43f3ea1e-777a-4cc6-8478-13f981b7c941");
+
+    constexpr uint64_t historyRequest = 1;
```
MatthiasKillat, line 591: Note that I would like to introduce the coding convention that local constants (like here) should not be all caps. I find the flow to read them like this much more pleasant. As an added bonus it is easier to distinguish them from "configuration constants" with a much a larger scope.

Those constants here mainly serve to express intent and make the tests easier to understand (they are the essential part of the test configuration).

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
```diff
@@ -50,6 +50,9 @@ struct PublisherPortData : public BasePortData
         ChunkSenderData<MAX_CHUNKS_ALLOCATED_PER_PUBLISHER_SIMULTANEOUSLY, ChunkDistributorData_t>;
 
     ChunkSenderData_t m_chunkSenderData;
+
+    PublisherOptions m_options;
```
elBoberido, line 54: While I think it is a good idea to save this in the port in order to be able to do the matching, I'm a little bit hesitant that we run into issues since some of the options are additionally stored at a other locations. It might happen that they are mutated and then out of sync. Maybe a issue to check if this duplication can be removed would be good.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Extend cxx::optional constructor for in place construction so that copy/move for values inside the optional even could be deleted[\#967](https://github.com/eclipse-iceoryx/iceoryx/issues/967)
 - Add templated `from`/`into` free functions to formalize conversions from enums and other types [#992](https://github.com/eclipse-iceoryx/iceoryx/issues/992)
 - UniqueId class for unique IDs within a process [#1010](https://github.com/eclipse-iceoryx/iceoryx/issues/1010)
+- Add requirePublisherHistorySupport option at subscriber side (if set to true requires historyRequest <= historyCapacity to be eligible for connection)
```
elfenpiff, line 24: Please add a link to the issue like in the line above.

iceoryx_posh/include/iceoryx_posh/popo/subscriber_options.hpp
```diff
@@ -33,7 +33,7 @@ struct SubscriberOptions
 {
     /// @brief The size of the receiver queue where chunks are stored before they are passed to the user
     /// @attention Depending on the underlying queue there can be a different overflow behavior
-    uint64_t queueCapacity{SubscriberPortData::ChunkQueueData_t::MAX_CAPACITY};
+    uint64_t queueCapacity{iox::MAX_SUBSCRIBER_QUEUE_CAPACITY};
```
elBoberido, line 36: I'm not quite sure if this is the way to go. I know that some stuff needed to be refactored. Maybe further down the PR it's becomes clear.

iceoryx_examples/iceoptions/README.md
```diff
@@ -72,11 +72,20 @@ would encounter an overflow, the oldest sample is released to create space for t
...
 ```
 
-`historyRequest` will enable a subscriber to receive the last n samples on subscription e.g. in case it was started later than the publisher. The publisher needs to have its `historyCapacity` enabled, too.
+`historyRequest` will enable a subscriber to receive the last n samples on subscription e.g. in case it was started later than the publisher.
+If the publisher does not have a sufficient `historyCapacity` (larger than `historyRequest`), it will still be connected but we will not be able to
+receive the requested amount of historical data (if it was available). Instead we will receive the largest amount of historical sample
+the publisher has available, i.e. best-effort.
+
+If we want to enforce the contract that the publisher needs to support a `historyCapacity` of at least `historyRequest`, we can do so by setting
+`requirePublisherHistorySupport` to `true`. In this case the subscriber will only connect if the publisher history support is at least as large as is request.
```
elfenpiff, line 81: What if the publisher history support is large enough but does not contain the requested amount of samples?

iceoryx_binding_c/include/iceoryx_binding_c/subscriber.h
```diff
@@ -45,6 +45,11 @@ typedef struct
     /// @brief describes whether a publisher blocks when subscriber queue is full
```
elfenpiff, line 45: Please update the copyright header year in all files you have touched.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
elBoberido: Hmm, not sure about this. We already use a different coding convention all over the place and it would be another exception to remember.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -126,6 +157,111 @@ TEST_F(PublisherSubscriberCommunication_test, AllSubscriberInterfacesCanBeSubscr
...
+    constexpr uint64_t historyRequest = 1;
+    constexpr uint64_t historyCapacity = 0;
+    constexpr bool requireHistory = true;
+
+    auto publisher = createPublisher<int>(historyCapacity);
+    auto subscriber = createSubscriber<int>(historyRequest, requireHistory);
+
+    this->InterOpWait();
+
+    ASSERT_TRUE(publisher);
```
elfenpiff, line 173: Could we send also a simple integer from the publisher to the subscriber to verify that `subscriber` is subscribed and not something else.
It is a communication test but we do not perform any kind of actual  communication here.

I would add something for all the tests.
When they do not connect then sending an int should run into nothing otherwise the subscriber should receive something.

elBoberido (commented):
Did not yet review the tests, but looks good so far

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
elfenpiff: Fine by me but I would bet that we have to change it to all caps when we want to certify the code because then we have to blindly follow the rules. I have no idea if we can make them so fine grained.

iceoryx_posh/include/iceoryx_posh/popo/subscriber_options.hpp
elBoberido: I'm still not sure if this is a good idea or what consequences this will have when not `SubscriberPortData::ChunkQueueData_t::MAX_CAPACITY` is used. It might be better to do it like for the RPC feature where a `client_server_port_types.hpp` was created to prevent cyclic dependencies

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -585,6 +584,163 @@ TEST_F(PortManager_test, DoDiscoveryPublisherCanWaitAndSubscriberDiscardOldestLe
...
+    constexpr uint64_t historyCapacity = 0;
+    constexpr bool requireHistory = true;
+
+    PublisherOptions publisherOptions{
```
elfenpiff, line 595: Line 595 - 607 seems to be always more or less the same. Could we move this into a function and make the variation points arguments? This would reduce the amount of code and complexity of the following tests.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -126,6 +157,111 @@ TEST_F(PublisherSubscriberCommunication_test, AllSubscriberInterfacesCanBeSubscr
...
+    ::testing::Test::RecordProperty("TEST_ID", "31cbd36d-32f1-4bc7-9980-0cdf5f248035");
+
+    constexpr uint64_t historyRequest = 1;
+    constexpr uint64_t historyCapacity = 0;
+    constexpr bool requireHistory = true;
+
+    auto publisher = createPublisher<int>(historyCapacity);
+    auto subscriber = createSubscriber<int>(historyRequest, requireHistory);
+
+    this->InterOpWait();
```
elBoberido, line 171: This can be removed since they are connected on create if they match ... please also remove it from the other test cases

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
elBoberido: Well, in this test it must not be subscribed

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -126,6 +157,111 @@ TEST_F(PublisherSubscriberCommunication_test, AllSubscriberInterfacesCanBeSubscr
     }
 }
 
+TEST_F(PublisherSubscriberCommunication_test, SubscriberRequiringHistoryDoesNotConnectToPublisherWithoutHistorySupport)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "31cbd36d-32f1-4bc7-9980-0cdf5f248035");
+
+    constexpr uint64_t historyRequest = 1;
+    constexpr uint64_t historyCapacity = 0;
+    constexpr bool requireHistory = true;
```
elBoberido, line 166: How about calling this `requireHistorySupport`? It would make the intention more obvious.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -126,6 +157,111 @@ TEST_F(PublisherSubscriberCommunication_test, AllSubscriberInterfacesCanBeSubscr
...
+    auto publisher = createPublisher<int>(historyCapacity);
+    auto subscriber = createSubscriber<int>(historyRequest, requireHistory);
+
+    this->InterOpWait();
+
+    ASSERT_TRUE(publisher);
+    EXPECT_FALSE(publisher->hasSubscribers());
+}
+
+TEST_F(PublisherSubscriberCommunication_test, SubscriberRequiringNoHistoryDoesConnectToPublisherWithNoHistorySupport)
```
elBoberido, line 177: The name is ambiguous. Maybe rename to
```suggestion
TEST_F(PublisherSubscriberCommunication_test, SubscriberNotRequiringHistorySupportDoesConnectToPublisherWithNoHistorySupport)
```

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -126,6 +157,111 @@ TEST_F(PublisherSubscriberCommunication_test, AllSubscriberInterfacesCanBeSubscr
...
+    auto subscriber = createSubscriber<int>(historyRequest, requireHistory);
+
+    this->InterOpWait();
+
+    ASSERT_TRUE(publisher);
+    EXPECT_TRUE(publisher->hasSubscribers());
+}
+
+TEST_F(PublisherSubscriberCommunication_test,
+       SubscriberRequiringHistoryDoesConnectToPublisherWithSufficientHistorySupport)
```
elBoberido, line 195: Similar with this one. Since there is always a `historyRequest` using only `RequiringHistory` might be misunderstood.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -126,6 +157,111 @@ TEST_F(PublisherSubscriberCommunication_test, AllSubscriberInterfacesCanBeSubscr
...
+    constexpr bool requireHistory = true;
+
+    auto publisher = createPublisher<int>(historyCapacity);
+    auto subscriber = createSubscriber<int>(historyRequest, requireHistory);
+
+    this->InterOpWait();
+
+    ASSERT_TRUE(publisher);
+    EXPECT_FALSE(publisher->hasSubscribers());
+}
```
elBoberido, line 245: Déjà vu ... isn't this already tested on line 160?

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -71,6 +71,21 @@ class PublisherSubscriberCommunication_test : public RouDi_GTest
             options);
     }
 
+    // additional overload for convenient use in the test cases
+    template <typename T>
+    std::unique_ptr<iox::popo::Publisher<T>> createPublisher(const uint64_t historyCapacity)
```
elBoberido, line 76: Just wondering, wouldn't this also work
```cpp
template <typename T>
iox::popo::Publisher<T>&& createPublisher(...) {
    return std::move(iox::popo::Publisher<T>(...));
}
```

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
elBoberido: Alternatively, use a function to create the default options and then only mutate the variation points by setting the relevant members. With this you would not need to create constants for the parameter to avoid magic numbers since the name of the variable will be in the same line.

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -763,6 +919,8 @@ TEST_F(PortManager_test, AcquireNodeDataAfterDestroyingPreviouslyAcquiredOnesIsS
     acquireMaxNumberOfNodes(nodeName, runtimeName);
 }
 
+// MAKI tests
```
elBoberido, line 922: Please remove this if you think all tests are done

iceoryx_examples/iceoptions/README.md
MatthiasKillat: They will connect. And why shouldn't they, it just means this data was never available. If at any time enough samples where available the publisher will have the latest samples available from then on.

It will just refuse connection if it is *never* possible to fulfill the contract, no matter how many samples it has published. This is well-defined I think.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
MatthiasKillat: We can do so, but actually I consider this wasteful since the tests run in isolation. We should separate tests which check matching and actual data transfer to reduce redundancy of tests. We currently do not have proper connection matching/discovery checks for many cases I think.

I would prefer the rationale that we

1. check whether we establish connections in all cases we need to
2. test whether we can transmit data whenever a connection is established but only for one connected case (otherwise this will scale badly if we add more cases where we expect connections)

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
MatthiasKillat: Yes.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
MatthiasKillat: Sure. General rule should be: the larger the scope, the more expressive a name needs to be. Interfaces need expressive names as well, local scope not so much if it means the following lines become more readable.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
MatthiasKillat: Can do, but I tried to keep the already long names as short as reasonably possible.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
elBoberido: @MatthiasKillat yes, I mean a new issue, just to not forget about this potential issue. If not before v2 we should definitely look afterwards on this and consolidate the options to a single source of truth.

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
MatthiasKillat: Yes, basically. Case is redundant (3 or 1 should not matter and if it does we could argue for other numbers as well).

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
MatthiasKillat: Not sure whether this would work but the current factory functions follow the style of existing ones and we should be careful not to mix to much (in tests), as it is unnecessarily confusing.

iceoryx_posh/include/iceoryx_posh/popo/subscriber_options.hpp
elBoberido: Because when someone sets a different value to `SubscriberPortData::ChunkQueueData_t::MAX_CAPACITY` we might break the port

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
MatthiasKillat: A leftover ...

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
MatthiasKillat: I think we can define coding rules like this for certification, whether we want to is another issue. There will be different tool settings needed for tests and library code for sure, since e.g. we use e.g. dynamic memory in tests for simplicity etc. From this point of view we could allow it for tests without a problem I think.

I kind of dislike to make the code uglier than it needs to be for this purpose only, and overusing ALL_CAPS_SNAKE_CASE is annoying to read.

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/publisher_port_data.hpp
MatthiasKillat: cf. https://github.com/eclipse-iceoryx/iceoryx/issues/1048

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/pub_sub_port_types.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <cstdint>
+
+namespace iox
+{
+namespace popo
+{
+/// @todo move definitions for publish subscribe communication here
+
+using ChunkQueueData_t = ChunkQueueData<DefaultChunkQueueConfig, ThreadSafePolicy>;
```
elBoberido, line 31: Please rename this to `SubscriberChunkQueueData_t` to prevent ambiguities

iceoryx_posh/include/iceoryx_posh/internal/popo/ports/pub_sub_port_types.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include "iceoryx_posh/internal/popo/building_blocks/chunk_receiver_data.hpp"
+#include "iceoryx_posh/internal/popo/building_blocks/locking_policy.hpp"
+
+#include <cstdint>
+
+namespace iox
+{
+namespace popo
+{
+/// @todo move definitions for publish subscribe communication here
```
elBoberido, line 29: Please create an issue for this and link it here

MatthiasKillat: Done, https://github.com/eclipse-iceoryx/iceoryx/issues/1051

iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp
```diff
@@ -41,8 +41,14 @@ int main()
...
     // and the publisher has history enabled).
+
+    // we do not require the publisher to support the history we request,
+    // i.e. we will still connect even if its historyCapacity is smaller than what we request
+
     //! [history]
     subscriberOptions.historyRequest = 5U;
+
+    subscriberOptions.requiresPublisherHistorySupport = false;
     //! [history]
```
elBoberido, line 52: Could you either run geoffrey or update the example readme manually? It might also ne a good idea to mention the new option in more detail in the readme, similar to the comment you added here

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -585,6 +609,120 @@ TEST_F(PortManager_test, DoDiscoveryPublisherCanWaitAndSubscriberDiscardOldestLe
...
+
+    auto publisher = createPublisher(publisherOptions);
+    auto subscriber = createSubscriber(subscriberOptions);
+
+    ASSERT_TRUE(publisher);
+    ASSERT_TRUE(subscriber);
+    EXPECT_TRUE(publisher.hasSubscribers());
+}
+
+TEST_F(PortManager_test, SubscriberRequiringHistorySupportDoesNotConnectToPublisherWithInSufficientHistorySupport)
```
elBoberido, line 669: nitpicking: `Insufficient` instead of `InSufficient`

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -585,6 +609,120 @@ TEST_F(PortManager_test, DoDiscoveryPublisherCanWaitAndSubscriberDiscardOldestLe
...
+TEST_F(PortManager_test, SubscriberRequiringHistorySupportDoesNotConnectToPublisherWithNoHistorySupport)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "8f17d9cb-8f46-4742-99c3-b609a4d2319a");
+
+    auto publisherOptions = createTestPubOptions();
+    auto subscriberOptions = createTestSubOptions();
+
+    publisherOptions.historyCapacity = 0;
+    subscriberOptions.historyRequest = 3;
+    subscriberOptions.requiresPublisherHistorySupport = true;
```
elBoberido, line 697: This is similar to the test at line 613 and following. Can this be removed?

elBoberido (commented):
Just minor things/nitpicks

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
MatthiasKillat: Yes, removed. It is the same equivalency class if nothing is broken. It might be broken for any arbitrary number though, in principle at least (but this cannot be feasibly checked).

iceoryx_examples/iceoptions/iox_subscriber_with_options.cpp
MatthiasKillat: Was already done in a previous commit.

elBoberido: yeah, just overlooked it

elBoberido (approved):

elfenpiff (approved):

Issue #1036, opened 2022-01-22
Remove creation design pattern class with in place implementation


elfenpiff:
## Brief feature description

Inheriting from the creation design pattern class has certain drawbacks which can be very error prone.
1. In move and copy operations one has to delegate the call to the base class first
2. `m_isInitialized`, `m_errorValue` are sometimes overlooked or not set correctly
3. the creation class has to be a friend class of the resource which is using it

A better implementation alternative would be to create a static method in such classes called `create` and create the system resource directly in there and provide the handle to that successfully created resource to the constructor of the class. The constructor does not create the system resource it just acquires a existing one and handles its lifetime.

Another approach is to introduce a builder pattern when the class constructors requires more then zero arguments to improve the readability.

## Classes where Creation pattern must be replaced with static create method
- [x] SharedMemoryObject
- [x] MemoryMap
- [x] SharedMemory
- [x] UnixDomainSocket (high prio)
- [x] MessageQueue (medium prio)
- [x] NamedPipe (medium prio)
- [x] FileLock
- [x] Semaphore (will be solved in #751)
- [x] Mutex (high prio)
  - [x] Implement builder pattern
  - [ ] Integrate new API into iceoryx and remove old constructor, all `@todo iox-#1036` should be gone
- [x] move `creation.hpp` to `iceoryx_dust` (high prio)

PR #1047, opened 2022-01-26
Iox #415 remove InvalidIdString and isValid() from ServiceDescription

### LINKED ISSUE #415: service registry as a built-in topic

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR removes `InvalidIdString` and `isValid()` from `ServiceDescription` and replaces the Wildcard string with a nullopt. With this, the `ServiceDescription` can contain any string, empty strings by default.

## References

- Partially closes #415

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@
 - Remove creation pattern from `MemoryMap` and replace it with `MemoryMapBuilder` [\#1036](https://github.com/eclipse-iceoryx/iceoryx/issues/1036)
 - Fix error handling of `TypedUniqueId` and refactor it to `UniquePortId` [\#861](https://github.com/eclipse-iceoryx/iceoryx/issues/861)
 - Updating Codecov API and enforce CMake version 3.16 for building iceoryx [\#774](https://github.com/eclipse-iceoryx/iceoryx/issues/774) and [\#1031](https://github.com/eclipse-iceoryx/iceoryx/issues/1031)
-- Remove `InvalidIdString` and `isValid()` from `ServiceDescription`, replace Wildcard string with struct [\#415](https://github.com/eclipse-iceoryx/iceoryx/issues/415)
+- Remove `InvalidIdString` and `isValid()` from `ServiceDescription`, replace Wildcard string with `iox::cxx::nullopt` [\#415](https://github.com/eclipse-iceoryx/iceoryx/issues/415)
 
 **API Breaking Changes:**
 
@@ -226,8 +226,8 @@ The `iox::cxx::expected` has dropped the requirement for `INVALID_STATE`. With t
 of `ErrorTypeAdapter` for custom types must therefore also be removed in the user code.
 
 The `InvalidIdString` was removed from `ServiceDescription` and the Wildcard string was replaced
-with a struct. With this, every string is allowed within the `ServiceDescription`. The default
-`ServiceDescription` consists of empty strings.
+with a `iox::cxx::nullopt`. With this, every string is allowed within the `ServiceDescription`.
+The default `ServiceDescription` consists of empty strings.
 
 ## [v1.0.1](https://github.com/eclipse-iceoryx/iceoryx/tree/v1.0.0) (2021-06-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@
 - Remove creation pattern from `MemoryMap` and replace it with `MemoryMapBuilder` [\#1036](https://github.com/eclipse-iceoryx/iceoryx/issues/1036)
 - Fix error handling of `TypedUniqueId` and refactor it to `UniquePortId` [\#861](https://github.com/eclipse-iceoryx/iceoryx/issues/861)
 - Updating Codecov API and enforce CMake version 3.16 for building iceoryx [\#774](https://github.com/eclipse-iceoryx/iceoryx/issues/774) and [\#1031](https://github.com/eclipse-iceoryx/iceoryx/issues/1031)
+- Remove `InvalidIdString` and `isValid()` from `ServiceDescription`, replace Wildcard string with struct [\#415](https://github.com/eclipse-iceoryx/iceoryx/issues/415)
 
 **API Breaking Changes:**
 
@@ -223,6 +224,10 @@ serviceDiscovery.findService("ServiceA", Wildcard);
 The `iox::cxx::expected` has dropped the requirement for `INVALID_STATE`. With this, the
 `ErrorTypeAdapter` which was necessary for non enum types was also removed. The specialization
 of `ErrorTypeAdapter` for custom types must therefore also be removed in the user code.
+
+The `InvalidIdString` was removed from `ServiceDescription` and the Wildcard string was replaced
+with a struct. With this, every string is allowed within the `ServiceDescription`. The default
+`ServiceDescription` consists of empty strings.
 
 ## [v1.0.1](https://github.com/eclipse-iceoryx/iceoryx/tree/v1.0.0) (2021-06-15)
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
@@ -382,6 +382,7 @@ constexpr T into(const F value) noexcept;
 struct Wildcard_t
 {
 };
+constexpr Wildcard_t Wildcard{};
 
 } // namespace cxx
 } // namespace iox
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
@@ -378,12 +378,6 @@ constexpr T into(const F value) noexcept;
   private:                                                                                                             \
     type m_##name = defaultValue;
 
-/// @brief Used to search for any string
-struct Wildcard_t
-{
-};
-constexpr Wildcard_t Wildcard{};
-
 } // namespace cxx
 } // namespace iox
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
@@ -378,6 +378,10 @@ constexpr T into(const F value) noexcept;
   private:                                                                                                             \
     type m_##name = defaultValue;
 
+/// @brief Used to search for any string
+struct Wildcard_t
+{
+};
 
 } // namespace cxx
 } // namespace iox
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -103,8 +103,7 @@ class ServiceDescription
                        ClassHash m_classHash = {0u, 0u, 0u, 0u},
                        Interfaces interfaceSource = Interfaces::INTERNAL) noexcept;
 
-    /// @brief compare operator. If wildcards AnyServiceString, AnyInstanceString or AnyEventString are used, the
-    /// corresponding member comparisons are skipped.
+    /// @brief compare operator.
     bool operator==(const ServiceDescription& rhs) const noexcept;
 
     /// @brief negation of compare operator.
@@ -139,7 +138,7 @@ class ServiceDescription
     bool isValid() const noexcept;
 
     ///@{
-    /// Getters for the integer and string IDs
+    /// Getters for the string IDs
     IdString_t getServiceIDString() const noexcept;
     IdString_t getInstanceIDString() const noexcept;
     IdString_t getEventIDString() const noexcept;
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2019, 2021 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -29,7 +29,6 @@ namespace iox
 {
 namespace capro
 {
-
 /// @brief Used to search for any string
 constexpr iox::cxx::nullopt_t Wildcard;
 
--- a/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
@@ -30,7 +30,6 @@ namespace iox
 {
 namespace capro
 {
-static const IdString_t InvalidIdString{""};
 static constexpr int32_t MAX_NUMBER_OF_CHARS = 64;
 static constexpr size_t CLASS_HASH_ELEMENT_COUNT{4U};
 
@@ -100,7 +99,7 @@ class ServiceDescription
     ServiceDescription(const IdString_t& service,
                        const IdString_t& instance,
                        const IdString_t& event,
-                       ClassHash m_classHash = {0u, 0u, 0u, 0u},
+                       ClassHash m_classHash = {0U, 0U, 0U, 0U},
                        Interfaces interfaceSource = Interfaces::INTERNAL) noexcept;
 
     /// @brief compare operator.
@@ -132,10 +131,6 @@ class ServiceDescription
     /// @brief Returns the scope of a ServiceDescription
     Scope getScope() const noexcept;
 
-    /// @brief Returns true for valid ServiceDescription
-    /// false for ServiceDescription that contain InvalidStrings.
-    /// @return bool, true if ServiceDescription is valid, false otherwise
-    bool isValid() const noexcept;
 
     ///@{
     /// Getters for the string IDs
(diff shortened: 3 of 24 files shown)


iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -96,7 +96,8 @@ class PortManager
     void deletePortsOfProcess(const RuntimeName_t& runtimeName) noexcept;
 
     const std::atomic<uint64_t>* serviceRegistryChangeCounter() noexcept;
-    runtime::IpcMessage findService(const capro::IdString_t& service, const capro::IdString_t& instance) noexcept;
+    runtime::IpcMessage findService(const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& service,
```
elfenpiff, line 99: An alternative could be to use `cxx::optional<capro::IdString_t>` instead of the `cxx::variant<cxx::Wildcard_t, capro::IdString_t>` construct.

The API would be a bit more understandable, as well as the code and we could get rid of the wildcard struct. But then downside may be that it is not clear for the user that a nullopt is then like a wildcard - on the other hand, what else could it be in the findService context!?

iceoryx_posh/include/iceoryx_posh/internal/roudi/roudi.hpp
```diff
@@ -17,6 +17,7 @@
 #ifndef IOX_POSH_ROUDI_ROUDI_MULTI_PROCESS_HPP
 #define IOX_POSH_ROUDI_ROUDI_MULTI_PROCESS_HPP
 
+#include "iceoryx_hoofs/cxx/convert.hpp"
```
elfenpiff, line 20: I think this is required only in `roudi.cpp` therefore I would move it into the cpp file. This reduces a bit the compile time and makes the header dependencies more clear.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
```diff
@@ -335,6 +335,12 @@ constexpr T from(const F value) noexcept;
 template <typename T, typename F>
 constexpr T into(const F value) noexcept;
 
+/// @brief Used to search for any string
```
elfenpiff, line 338: I would move this into `capro/service_description.hpp` since it is somehow part of that logic. The helplets have nothing to do with strings and string search.

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -17,6 +17,7 @@
 #ifndef IOX_POSH_RUNTIME_SERVICE_DISCOVERY_HPP
 #define IOX_POSH_RUNTIME_SERVICE_DISCOVERY_HPP
 
+#include "iceoryx_hoofs/cxx/helplets.hpp"
```
elfenpiff, line 20: If you move the `Wildcard_t` into the service description header you could remove this include here.

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -29,11 +29,11 @@ ServiceDescription::ClassHash::ClassHash() noexcept
 
 ServiceDescription::ClassHash::ClassHash(const std::initializer_list<uint32_t>& values) noexcept
 {
-    uint64_t index = 0u;
+    uint64_t index = 0U;
     for (auto& v : values)
     {
         data[index++] = v;
-        if (index == 4u)
+        if (index == 4U)
```
elfenpiff, line 36: You touched the fishy! :fish: :tropical_fish: 

Why not use `CLASS_HASH_ELEMENT_COUNT` instead of this magic number `4U`.

iceoryx_posh/source/popo/ports/interface_port_data.cpp
```diff
@@ -22,10 +22,7 @@ namespace iox
 namespace popo
 {
 InterfacePortData::InterfacePortData(const RuntimeName_t& runtimeName, const capro::Interfaces interface) noexcept
-    : BasePortData(capro::ServiceDescription(
-                       capro::InvalidIdString, capro::InvalidIdString, capro::InvalidIdString, {0, 0, 0, 0}, interface),
-                   runtimeName,
-                   "")
+    : BasePortData(capro::ServiceDescription("", "", "", {0, 0, 0, 0}, interface), runtimeName, "")
```
elfenpiff, line 25: A ServiceDescriptionBuilder would be perfect here. Then we could write code like:
```cpp
    : BasePortData(capro::ServiceDescriptionBuilder().interface(interface).create()), runtimeName, "")
```
which looks much cleaner.

This is nothing which has to be done here - but if you want I wont stop you ;)

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
```diff
@@ -133,13 +131,9 @@ class ServiceDescription
...
 
-    /// @brief Returns true for valid ServiceDescription
-    /// false for ServiceDescription that contain InvalidStrings.
-    /// @return bool, true if ServiceDescription is valid, false otherwise
-    bool isValid() const noexcept;
 
     ///@{
-    /// Getters for the integer and string IDs
+    /// Getters for the string IDs
     IdString_t getServiceIDString() const noexcept;
```
MatthiasKillat, line 137: We still return by value here. This is bad for performance everywhere it is used. I would prefer to have it return a `const IdString_t&`. Furthermore I am still not sure whether in a strict camel case model it would not be `FooIdString`, but I do not care much.

 This might break the usage somewhere, so in the future an alternative would be to have functions like `const IdString_t& serviceIdString()` or just `const IdString_t& service()`.

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
MatthiasKillat: Yes, the variant is bad and optional would be better. It would be even better to have overloads with `WildCard_t` in addition but then we would have 4 different combinations. Those can be used if compile time dispatch is possible (and the runtime dispatch can also variant can also call them internally. This gets annoying to extend and maintain though.

Ideally we can remove variant here though, but it is not a priority for internal calls (i.e. not exposed to the user).

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -58,13 +57,13 @@ class ServiceRegistry
...
 
-    /// @brief Removes given service description from registry
+    /// @brief Searches for given service description in registry
     /// @param[in] searchResult, reference to the vector which will be filled with the results
     /// @param[in] service, string or wildcard to search for
     /// @param[in] instance, string or wildcard to search for
     void find(ServiceDescriptionVector_t& searchResult,
-              const capro::IdString_t& service = Wildcard,
-              const capro::IdString_t& instance = Wildcard) const noexcept;
+              const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& service,
```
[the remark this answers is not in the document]
MatthiasKillat: Not sure this is possible with `variant` without wrapping it. In any case I prefer overloads in theory, i.e. have a call with *service*, one with *instance*, one with both and one without arguments. This improves dispatching as well and internally the cases are implemented in a switch statement otherwise anyway.

The reasoning is to prefer early dispatching mechanisms.

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -41,8 +42,8 @@ class ServiceDiscovery
     /// ServiceContainer: on success, container that is filled with all matching instances
     /// FindServiceError: if any, encountered during the operation
     cxx::expected<ServiceContainer, FindServiceError>
-    findService(const cxx::variant<Wildcard_t, capro::IdString_t> service,
-                const cxx::variant<Wildcard_t, capro::IdString_t> instance) noexcept;
+    findService(const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& service,
```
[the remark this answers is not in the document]
MatthiasKillat: I think the Wildcard belongs to the same namespace as the `ServiceDescription` but we may change some of those later, so I am not sure yet and until we know it is not important.

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -358,15 +346,35 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
...
         else
         {
-            capro::IdString_t service{cxx::TruncateToCapacity, message.getElementAtIndex(2)};
-            capro::IdString_t instance{cxx::TruncateToCapacity, message.getElementAtIndex(3)};
+            cxx::variant<cxx::Wildcard_t, capro::IdString_t> service;
+            cxx::variant<cxx::Wildcard_t, capro::IdString_t> instance;
+            bool isServiceWildcard = false;
+            bool isInstanceWildcard = false;
+            cxx::convert::fromString(message.getElementAtIndex(2).c_str(), isServiceWildcard);
+            if (!isServiceWildcard)
```
MatthiasKillat, line 361: I see no need to negate here, drop the `!` and just switch the statements in the `if` and `else` branches. Maybe  the compiler optimizes this away, maybe not, but for readability prefer less complexity if it is not needed.

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -358,15 +346,35 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
...
+            {
+                service.emplace<capro::IdString_t>(cxx::TruncateToCapacity, message.getElementAtIndex(3));
+            }
+            else
+            {
+                service.emplace<cxx::Wildcard_t>();
+            }
+            cxx::convert::fromString(message.getElementAtIndex(4).c_str(), isInstanceWildcard);
+            if (!isInstanceWildcard)
```
MatthiasKillat, line 370: same as above

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -100,25 +100,25 @@ void ServiceRegistry::remove(const capro::ServiceDescription& serviceDescription
...
-                           const capro::IdString_t& instance) const noexcept
+                           const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& service,
+                           const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& instance) const noexcept
 {
     cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> intersection;
 
     // Find (K1, K2)
     // O(log n + log n + max(#PossibleServices + #possiblesInstances) + #intersection)
-    if (instance != Wildcard && service != Wildcard)
+    if (instance.index() != 0 && service.index() != 0)
```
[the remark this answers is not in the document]
MatthiasKillat: Would be more readable, yes. 

The real question is still whether we want the `variant` in the interface, which is the only reason for the index in the first place...

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -100,25 +100,25 @@ void ServiceRegistry::remove(const capro::ServiceDescription& serviceDescription
...
     {
         cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> possibleServices;
         cxx::vector<uint64_t, MAX_SERVICE_DESCRIPTIONS> possibleInstances;
 
-        auto rangeServiceMap = m_serviceMap.equal_range(service);
+        auto rangeServiceMap = m_serviceMap.equal_range(*service.get_at_index<1>());
```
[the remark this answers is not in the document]
MatthiasKillat: I would not replace the index, but instead make the result of the call more explicit to convey intention.

I am not sure whether we can get a reference to the data contained in the `variant` (without copying), but if we can I would introduce a helper variable like `auto& serviceName = *service.get_at_index<1>();` and use this to make the code more expressive without any overhead (if it can be a reference the compiler will optimize it away).

If it is currently copied anyway it makes no difference performance-wise but will improve readability.

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -22,16 +22,18 @@ namespace iox
...
+ServiceDiscovery::findService(const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& service,
+                              const cxx::variant<cxx::Wildcard_t, capro::IdString_t>& instance) noexcept
 {
     /// @todo #415 remove the string mapping, once the find call is done via shared memory
     capro::IdString_t serviceString;
     capro::IdString_t instanceString;
+    bool isServiceWildcard = false;
+    bool isInstanceWildcard = false;
 
     if (service.index() == 0U)
```
MatthiasKillat, line 34: We could maybe also introduce a helper function 
```
bool isWildCard(const cxx::variant< ... >& value) {
return service.index() == 0U;
}
```

This will be inlined and make the code more expressive.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -546,22 +544,14 @@ void PortManager::destroySubscriberPort(SubscriberPortType::MemberType_t* const
...
 {
-    // send find to all interfaces
-    capro::CaproMessage caproMessage(capro::CaproMessageType::FIND, {service, instance, roudi::Wildcard});
-
-    for (auto interfacePortData : m_portPool->getInterfacePortDataList())
-    {
-        iox::popo::InterfacePort interfacePort(interfacePortData);
-        interfacePort.dispatchCaProMessage(caproMessage);
-    }
-
```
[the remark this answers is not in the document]
MatthiasKillat: I also do not understand this, but I suspect it has something to do with the FIND message being changed and interface ports not being used properly (in the sense that they will not answer the request anyway).

But that is just a guess and requires clarification. @FerdinandSpitzschnueffler

MatthiasKillat (commented):
Looks OK, minor clarifications needed and I consider the `variant` interface problematic in the long-term but this PR is not the place to change that.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @mossmaurice I decided against because it was not used anywhere, not even tested, and as @MatthiasKillat wrote, it's not that easy with a variant.

iceoryx_posh/source/roudi/port_manager.cpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: After changing the `findService` parameters, this code had to be adapted since the `CaproMessage` c'tor requires a `ServiceDescription`. I discussed several ideas with @elfenpiff, but then we came to the conclusion that the `CaproMessage` itself could need some refactoring and that we can send the `find` here but there will be no answer anyway as @MatthiasKillat stated.

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -358,15 +346,35 @@ void RouDi::processMessage(const runtime::IpcMessage& message,
     }
     case runtime::IpcMessageType::FIND_SERVICE:
     {
-        if (message.getNumberOfElements() != 4)
+        if (message.getNumberOfElements() != 6)
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: right

iceoryx_posh/source/popo/ports/interface_port_data.cpp
elfenpiff: Actually no since in #1036 we remove the creation pattern from the posix wrappers.

But maybe you can implement this here or in a follow up PR. With the helper macro it is also not so much code. Here maybe 10 lines or some more.
Take a look at the MemoryMapBuilder: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object/memory_map.hpp

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I agree with @MatthiasKillat and moved the Wildcard (now a nullopt) to the `ServiceDescription`, so the namespace is now `capro`.

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Obsolete, because I replaced variant with optional.

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Obsolete, because I replaced variant with optional.

iceoryx_posh/source/roudi/roudi.cpp
FerdinandSpitzschnueffler: No else branches anymore, would leave it as it is.

iceoryx_posh/source/popo/ports/interface_port_data.cpp
FerdinandSpitzschnueffler: Although it's a good idea, I won't do this in this PR. I added a todo to #415 .

CHANGELOG.md
```diff
@@ -219,6 +220,10 @@ The `iox::cxx::expected` has dropped the requirement for `INVALID_STATE`. With t
 `ErrorTypeAdapter` which was necessary for non enum types was also removed. The specialization
 of `ErrorTypeAdapter` for custom types must therefore also be removed in the user code.
 
+The `InvalidIdString` was removed from `ServiceDescription` and the Wildcard string was replaced
```
elfenpiff, line 223: Can you please add a short code snippet with how it was used before and how it is used now.

MatthiasKillat (approved):

CHANGELOG.md
MatthiasKillat: I think we will have the Wildcard usage in the upcoming findService example. Nothing else change for the user,except for the inability to create an invalid ServiceDescription now (which made no sense anyway). Considering this, I am not sure what needs to be shown.

MatthiasKillat (approved):

elfenpiff (approved):

Issue #982, opened 2021-12-02
Design and implement new `StatusPort`


mossmaurice:
## Brief feature description

After [this discussion](https://github.com/eclipse-iceoryx/iceoryx/discussions/818) the conclusion was a `StatusPort` or `BroadcastPorts` is beneficial for #415 and for the introspection topics, in order to prevent any user to `take()` the data and not `release()` which could lead to a out-of-memory error.

## Detailed information

* [ ] Design document for `StatusPort`
* [ ] Implement `StatusPorts`
* [ ] Replace `InterfacePorts` with `StatusPorts`

elfenpiff:
@mossmaurice maybe I missed something but the purpose of the interface port is to forward capro messages for instance to gateway ports.
And the purpose of the StatePort is to deliver some kind of state. How should we be able to replace the InterfacePort with a StatusPort?

And shouldn't it be called StatePort instead of StatusPort?

mossmaurice:
@elfenpiff 
> maybe I missed something but the purpose of the interface port is to forward capro messages for instance to gateway ports.

That's correct. More broadly speaking `InterfacePort`s are used to relay discovery information to gateways. After the `StatusPort`s are available we should consolidate them with both the `InterfacePort`s and the introspection 2.0 by sending the `ServiceRegistry` via a `StatusPort`. Fore more information also see the [design document](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/design/service-discovery.md#decision). Very likely it will be fleshed out from this issue.

> And the purpose of the StatePort is to deliver some kind of state. How should we be able to replace the InterfacePort with a StatusPort?
> And shouldn't it be called StatePort instead of StatusPort?

I think that's a misunderstanding. I'm talking about [these](https://github.com/eclipse-iceoryx/iceoryx/discussions/818) `StatusPort` or `BroadcastPort`s.

mossmaurice:
Not needed anymore. Closing.

PR #1064, opened 2022-02-03
Iox #982 Add `StatusPort` design document

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Add design `StatusPort` design document
 
## References

- Closes partly #982
- Related to #415

### DIFF
--- a/doc/design/diagrams/status-port/status-port-overview.puml
+++ b/doc/design/diagrams/status-port/status-port-overview.puml
@@ -1,0 +1,51 @@
+@startuml
+
+title StatusPort overview
+
+skinparam classAttributeIconSize 0
+
+enum ActiveChunk {
+  FIRST
+  SECOND
+}
+
+class Transaction {
+  +activeChunk: ActiveChunk
+  +abaCounter: uint32_t
+}
+
+class StatusPortData {
+  +latestTransaction: std::atomic<Transaction>
+  +serviceDescription: capro::ServiceDescription
+  +chunks: SharedChunk[2]
+}
+
+class StatusPortReader<T> {
+  +StatusPortReader(statusPortDataPtr: cxx::not_null<StatusPortData* const)
+  +take(callable: cxx::function_ref<void(const T&)>): void
+  ' Open question if copyTake() is wanted
+  'copyTake(callable: cxx::function_ref<void(const T&)>): void)
+  -getMembers() const: const StatusPortData*
+  -getMembers(): StatusPortData*
+  -enableEvent(triggerHandle: iox::popo::TriggerHandle&&, event: const MyTriggerClassEvents): void
+  -disableEvent(event: const MyTriggerClassEvents): void
+  ' Some more methods need to be implemented in the StatusPortReader to be attachable to a Listener
+  ' See https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/waitset/ice_waitset_trigger.cpp
+  ' Open question if copyTake() is wanted
+  '-m_copyOfUserData: T
+  -m_statusPortDataPtr: StatusPortData*
+}
+
+
+class StatusPortWriter<T> {
+  +StatusPortWriter(statusPortDataPtr: cxx::not_null<StatusPortData* const)
+  +store(callable: cxx::function_ref<void(T&)>): void
+  -getMembers() const: const StatusPortData*
+  -getMembers(): StatusPortData*
+  -m_statusPortDataPtr: StatusPortData*
+}
+
+StatusPortWriter ..> StatusPortData
+StatusPortReader ..> StatusPortData
+
+@enduml
--- a/doc/design/status_port.md
+++ b/doc/design/status_port.md
@@ -1,0 +1,453 @@
+# StatusPort
+
+## Summary and requirements
+
+The `StatusPort` is an alternative to the publish subscriber communication in `iceoryx_posh`.
+
+The target of the `StatusPort` are use-case with the following properties:
+
+* Data is transferred once or is rarely updated
+* There are many readers which are interested in the data
+* Data can be persistent
+    * The lifetime of the transferred data is bound to the lifetime of `StatusPortData`
+* Readers don't access the data directly but via a lambda
+    * Preventing torn reads, since the `StatusPortReader` detects if the data
+    changed during `take()` operation (Frankenstein check) and re-executes the lambda
+* Has to be attachable to `Listener`
+* `StatusPort` could potentially be used to communicate between different ASIL domains
+
+Potential applications are:
+
+* Introspection topics
+* Service discovery
+
+## Design
+
+### Terminology
+
+* Transaction: Atomic state of the world, which is changed by a write operation
+* Current transaction: Transaction in local scope, read in the beginning of each operation
+* Latest transaction: Transaction in the shared memory managment segment
+* Chunk: Untyped piece of memory located in the shared memory payload segment
+* Read position: The chunk, which was used the last time to write data
+* Write position: The opposite chunk, not currently being used by the `StatusPortReader`s
+
+### Contract & properties
+
+* 1:n communication aka broadcast
+* Two memory chunks from shared memory payload segment are used
+* Data exchanged needs to be trivially-copyable
+* Storing a chunk cannot fail
+* Reading a chunk cannot fail
+    * If no data was sent yet, `callable` is not called
+    * If data was updated in-between read, re-call `callable`
+* Not part of the user API in the first version, only used internally
+    * Might become user API for freedom-from-interference separation
+* Reader only needs write access to the data structure and no read access
+* Writing and Reading will be tried indefinitely till possible, hence starvation
+is possible
+* `StatusPortData` is created in the shared memory segment if either a `StatusPortWriter`
+or `StatusPortReader` is created
+    * Two chunks in the shared memory payload segment are bound to the lifetime
+    of the `StatusPortData` (either via `StatusPortData` c'tor or
+    `StatusPort{Writer,Reader}` c'tor)
+
+### Discarded ideas & design alternatives
+
+#### Atomic pointer
+
+The pointer to the currently active chunk could also be stored in an `std::atomic`.
+
+```cpp
+std::atomic<T*> activeChunk{nullptr}
+```
+
+However, it would need the full 64-bit and which is not needed when managing
+just two chunks. Hence an `abaCounter` would need to be stored in a separate
+`std::atomic` variable.
+
+#### Omit `ServiceDescription` and discovery
+
+Methods like `offer()` or `stopOffer()` are not needed to keep things simple.
+However, omitting the `ServiceDescription` completely is not possible as it is
+needed to associate the right `StatusPortData` in shared memory with the correct
+writer or reader object on the stack. The `StatusPort` shall follow the
+service-based design.
+
+#### `atomic<T>::exchange` in `StatusPortReader::take()`
+
+A [CAS](https://en.wikipedia.org/wiki/Compare-and-swap) operation is not possible
+because CAS needs write access and a `StatusPortReader` shall not be allowed to
+write data into shared memory.
+
+#### Differentiate between Writer/Reader between RouDi/User?
+
+To be consistent with the previously available publisher and subscribers one
+could argue to follow the same naming. However, the `StatusPort` might be used to
+communicate between ASIL domains in future and hence become user API. As a result
+the naming shall stay Writer/Reader.
+
+#### Can part of the previous port infrastructure be re-used?
+
+Potential candidates:
+
+1. `BasePort`
+   1. The previously available publisher and subscribers follow a much more complex
+   communication pattern. Hence, most of the data in `BasePort` is irrelevant and
+   `StatusPortData` shall stay separate.
+1. `popo::Sample`
+   1. Due to the fact that data is accessed solely via `callable`'s there is no
+   benefit in using the non-owning `popo::Sample` view.
+
+### Solution
+
+#### Class structure
+
+Depicted below is the class diagram.
+
+![StatusPort class diagram](diagrams/status-port/status-port-overview.svg)
+
+The `StatusPortWriter` is a template where `T` is the transferred data type.
+It has a constructor which takes a pointer to the `StatusPortData`
+object in the shared memory management segment. The pointer is acquired in the
+same manner as previous publishers and subscribers via RouDi's unix domain
+socket. The only difference is that one `StatusPortData` object is shared between
+`StatusPortReader` and `StatusPortWriter`. `store()` takes a `function_ref` to
+be able to manipulate the type in-place in shared memory.
+`StatusPortReader` is a template as well and has the same constructor. The first
+and single argument of`take()` is a `function_ref`, which is executed as long as
+the `take()` was unsuccessful. Furthermore, the `StatusPortReader` has the methods
+necessary to be attachable to a `popo::Listener` (e.g. `{enable,disable}Event`).
+
+`StatusPortData` stores the latest atomic `Transaction` as well as two pointers
+to `SharedChunk`s in the shared memory payload segment. The lifetime of the two
+chunks is bound to the lifetime of the `StatusPortData` object. To be able to to
+associate the correct readers and writers a `ServiceDescription` is used.
+
+#### Frankenstein object corner cases
+
+How can we ensure in `StatusPortWriter::store()` that there is no `StatusPortReader`
+reading the chunk to which we want to write to? This could be for example a slow
+`StatusPortReader`. Is a reference counter needed? No, this is not the case. Instead
+the `StatusPortReader` checks if his view of on the world (aka the two chunks)
+has changed in-between.
+
+To illustrate this, let's assume that that there is one `StatusPortWriter` and one
+`StatusPortReader` in this scenario. Further assume that the type transferred via
+the `StatusPort` is
+
+```cpp
+T = cxx::string<36>
+```
+
+The following samples are transmitted by the `StatusPortWriter`
+
+1. "Time And Relative Dimension In Space"
+2. "Twin Ion Engines Fighter"
+3. "USS Enterprise"
+
+After the second transaction the `StatusPortReader` see's the latest transaction
+below and starts reading the `chunk[1]` aka `SECOND`.
+
+```text
+Latest transaction in shared memory managment segment
++----------------------------------+
+|       ActiveChunk::SECOND        |
+|       abaCounter: 1              |
++----------------------------------+
+```
+
+Now the the `StatusPortWriter` starts writing concurrently the third sample and
+overtakes the slower `StatusPortReader`.
+Due the latest transaction depicted below, `writePosition` is calculated to
+`chunks[1]` aka `SECOND`. Below you can find the location `^` at which the writer
+is currently writing and the reader is reading. As a result the strings are mixed
+up at this very moment in time, a so called Frankenstein object. While in this
+case, the reader might still read uncorrupted data if he finished before the
+writer, one can easily imagine that this can lead to very subtile, nasty bugs.
+To avoid this problem the `StatusPortReader` compares the initial transaction
+with the latest transaction and  re-calls the callable as long as they are not
+the same.
+
+```text
+Shared memory payload segment
++----------------------------------------------------+
+|        SharedChunk chunks[2]                       |
+|        +----------------------------------------+  |
+| FIRST  | "Twin Ion Engines Fighter"             |  |
+|        +----------------------------------------+  |
+| SECOND | "USS Enterprlative Dimension In Space" |  |
+|        +----------------------------------------+  |
+|                      ^                 ^           |
+|                      Writer            Slow reader |
++----------------------------------------------------+
+
+Latest transaction in shared memory managment segment
++----------------------------------+
+|       ActiveChunk::FIRST         |
+|       abaCounter: 2              |
++----------------------------------+
+```
+
+> ***
+> Special care has to be taken, when writing the `callable`!
+> ***
+>
+> It must be ensured that the `callable` does not crash or ends up in an endless
+> loop, when working on the data
+
+The memory order ``std::memory_order_release` and `std::memory_order_acquire`
+ensure that the operations below the `load()` "happened-before" and the memory of
+`chunks[]` is synchronized.
+
+```text
+// Write data via callable
+callable(*(m_statusPortDataPtr->chunks[currentWritePosition]));
+        \         /
+         \       /
+          \     /
+           \   /
+            \ /
+    latestTransaction.store(newTransaction,   ----------+
+       std::memory_order_release)                       |
+                                                        +--------->   latestTransaction.load(std::memory_order_acquire)
+                                                                                / \
+                                                                               /   \
+                                                                              /     \
+                                                                             /       \
+                                                                            /         \
+                                                        // Read sychronized data via callable
+                                                        callable(*(m_statusPortDataPtr->chunks[currentReadPosition].data));
+```
+
+### Code example
+
+```cpp
+// Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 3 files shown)


doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+} // namespace iox
+```
+
+## Open issues
+
+* Naming
+    * `StatusPort{Writer,Reader}` or just `StatusReader` and `StatusWriter`?
+* Does `copyTake()` make sense as an additional contract with the user?
+    * It would be possible that the user provides an arbitrary lambda, that does
+    not need to be written with care
```
MatthiasKillat, line 422: I am also not sure that the read/take algorithm does not have a problem where we may read corrupt data without detecting it. I will have to check some material and think about it and will get back to this later.

MatthiasKillat (changes requested):

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
+# StatusPort
+
+## Summary and requirements
+
+The `StatusPort` is an alternative to the publish subscriber communication in `iceoryx_posh`.
+
+The target of the `StatusPort` are use-case with the following properties:
+
+* Data is transferred once or is rarely updated
```
elfenpiff, line 9: Requirements to the data would be:
* data must be self-contained (pointers, indices to structures outside are forbidden - caused by torn read)
* data must be trivially copy-able
* all the other SHM constraints of the data must be satisfied

The reader must be able to handle frankenstein objects or mermaids ;)

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* There are many readers which are interested in the data
+* Data can be persistent
+    * The lifetime of the transferred data is bound to the lifetime of `StatusPortData`
+* Readers don't access the data directly but via a lambda
+    * Preventing torn reads, since the `StatusPortReader` detects if the data
+    changed during `take()` operation (Frankenstein check) and re-executes the lambda
+* Has to be attachable to `Listener`
```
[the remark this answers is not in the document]
elfenpiff: When it is attachable to the `Listener` it is also attachable to the `Waitset` therefore I would add this requirement here since it comes at zero cost.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* `StatusPort` could potentially be used to communicate between different ASIL domains
```
elfenpiff, line 17: But with very heavy restriction to the data and with no history. Furthermore you stated that you never or only rarely send data which is most likely not the case for communication between different ASIL domains.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Introspection topics
+* Service discovery
+
+## Design
+
+### Terminology
+
+* Transaction: Atomic state of the world, which is changed by a write operation
+* Current transaction: Transaction in local scope, read in the beginning of each operation
+* Latest transaction: Transaction in the shared memory managment segment
```
elfenpiff, line 30: Please define this differently with only the StatusPort scope in mind. There is nowhere shared memory mentioned and the StatusPort does not require shared memory at all, also does not know what a management segment is.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Chunk: Untyped piece of memory located in the shared memory payload segment
```
elfenpiff, line 31: How about pointer to untyped memory? I would define everything only inside the scope of the knowledge of a status port. And this state port has a pointer to some piece of memory located somewhere but this is not important for the status port.

This is important when we describe how the status port is integrated.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Read position: The chunk, which was used the last time to write data
+* Write position: The opposite chunk, not currently being used by the `StatusPortReader`s
+
+### Contract & properties
+
+* 1:n communication aka broadcast
+* Two memory chunks from shared memory payload segment are used
+* Data exchanged needs to be trivially-copyable
+* Storing a chunk cannot fail
```
elfenpiff, line 40: Maybe add here why this is. Since the underlying memory is trivially copyable and transferred via `memcpy`

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Reading a chunk cannot fail
```
elfenpiff, line 41: But it can fail ... you can read corrupt data and you have to handle it.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+    * If no data was sent yet, `callable` is not called
+    * If data was updated in-between read, re-call `callable`
+* Not part of the user API in the first version, only used internally
+    * Might become user API for freedom-from-interference separation
+* Reader only needs write access to the data structure and no read access
```
elfenpiff, line 46: Hehe ;) so the reader is allowed to change the data but not to read it

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+
+```cpp
+std::atomic<T*> activeChunk{nullptr}
+```
+
+However, it would need the full 64-bit and which is not needed when managing
+just two chunks. Hence an `abaCounter` would need to be stored in a separate
+`std::atomic` variable.
+
+#### Omit `ServiceDescription` and discovery
```
elfenpiff, line 69: The ServiceDescription has nothing to do with the status port.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Reader only needs write access to the data structure and no read access
+* Writing and Reading will be tried indefinitely till possible, hence starvation
+is possible
+* `StatusPortData` is created in the shared memory segment if either a `StatusPortWriter`
+or `StatusPortReader` is created
+    * Two chunks in the shared memory payload segment are bound to the lifetime
+    of the `StatusPortData` (either via `StatusPortData` c'tor or
+    `StatusPort{Writer,Reader}` c'tor)
+
+### Discarded ideas & design alternatives
```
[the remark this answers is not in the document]
elfenpiff: I agree, otherwise this design documented is a little bit bloated.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+#### `atomic<T>::exchange` in `StatusPortReader::take()`
+
+A [CAS](https://en.wikipedia.org/wiki/Compare-and-swap) operation is not possible
+because CAS needs write access and a `StatusPortReader` shall not be allowed to
+write data into shared memory.
+
+#### Differentiate between Writer/Reader between RouDi/User?
+
+To be consistent with the previously available publisher and subscribers one
+could argue to follow the same naming. However, the `StatusPort` might be used to
```
elfenpiff, line 86: I doubt that the `StatusPort` can be used to communicate between ASIL domains since it has some very heavy restrictions on the data.
* self contained
* trivially copyable
* changes never or only rarely
* no history possible
* can be corrupted when reading (when something writes new data)

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+#### Class structure
+
+Depicted below is the class diagram.
+
+![StatusPort class diagram](diagrams/status-port/status-port-overview.svg)
+
+The `StatusPortWriter` is a template where `T` is the transferred data type.
+It has a constructor which takes a pointer to the `StatusPortData`
+object in the shared memory management segment. The pointer is acquired in the
+same manner as previous publishers and subscribers via RouDi's unix domain
```
elfenpiff, line 113: Please change unix domain socket to IPC channel since we will use message passing in QNX, unix domain sockets for linux and mac and named pipes for windows.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+socket. The only difference is that one `StatusPortData` object is shared between
+`StatusPortReader` and `StatusPortWriter`. `store()` takes a `function_ref` to
+be able to manipulate the type in-place in shared memory.
+`StatusPortReader` is a template as well and has the same constructor. The first
+and single argument of`take()` is a `function_ref`, which is executed as long as
+the `take()` was unsuccessful. Furthermore, the `StatusPortReader` has the methods
+necessary to be attachable to a `popo::Listener` (e.g. `{enable,disable}Event`).
```
elfenpiff, line 120: You forgot `invalidateTrigger`

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+
+`StatusPortData` stores the latest atomic `Transaction` as well as two pointers
+to `SharedChunk`s in the shared memory payload segment. The lifetime of the two
+chunks is bound to the lifetime of the `StatusPortData` object. To be able to to
```
elfenpiff, line 124: To be able to to  ... ;) one less to please

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+overtakes the slower `StatusPortReader`.
+Due the latest transaction depicted below, `writePosition` is calculated to
+`chunks[1]` aka `SECOND`. Below you can find the location `^` at which the writer
+is currently writing and the reader is reading. As a result the strings are mixed
+up at this very moment in time, a so called Frankenstein object. While in this
+case, the reader might still read uncorrupted data if he finished before the
+writer, one can easily imagine that this can lead to very subtile, nasty bugs.
+To avoid this problem the `StatusPortReader` compares the initial transaction
+with the latest transaction and  re-calls the callable as long as they are not
+the same.
```
elfenpiff, line 170: This is why only self contained data is allowed. Lets assume instead of a string a pointer to some data is stored in here.
Then it is possible that the reader either dereferences a no longer valid pointer or a pointer which is completely corrupted. Furthermore containers like vector and string are also strictly forbidden. Assume the following example.
```cpp
myStateport.take([](const cxx::vector & v){
  for(auto iter = v.begin(); iter != v.end(); ++iter) {
    // while iterating the writer clears the vector. when we are already at position 1 we have `iter == v.end() + 1` and this means we will iterate until we reach an unaccessible memory location and segfault.
  }
});
```

This means every data structure with any kind of logic may cause a catastrophic failure. Therefore only plain old data, c arrays or structs of them should be allowed. And even here we have some serious problems. Look at this one here
```cpp
template<uint64_t N>
using charArray = char[N];

myStateport.take([](const charArray<100> & helloWorld){
  std::cout << helloWorld << std::endl;
  // lets assume the charArray contains "hello world\0" and is updated from the writer to "bla\0 fuu asldalsdkalsdkalsdasd"
  // this piece of code would segfault since there is no final "\0" anymore and the writer may assume that a reader only prints "bla" since it is terminated with "\0". I know it is a little bit constructed but nevertheless one have to be aware of the extreme restrictions we have here.
});
```

elfenpiff: In short the usage of any class you find in `cxx` is forbidden as type.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+|       ActiveChunk::FIRST         |
+|       abaCounter: 2              |
++----------------------------------+
+```
+
+> ***
+> Special care has to be taken, when writing the `callable`!
+> ***
+>
+> It must be ensured that the `callable` does not crash or ends up in an endless
```
elfenpiff, line 196: How does one verify this in a callable. One suggestion would be the following:
1. Create an integration tests which every callable has to pass.
2. This integration test has two threads, one low prio thread in which the callable is called, one high prio thread which randomly changes the data with random content. This test is not allowed to segfault.

doc/design/diagrams/status-port/status-port-overview.puml
```diff
@@ -0,0 +1,51 @@
+@startuml
```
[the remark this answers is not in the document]
mossmaurice: I agree. In this case the we are already at a later stage. There is already code of a prototype available.

doc/design/status_port.md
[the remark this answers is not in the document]
mossmaurice: Correct, because the reader is read-only. Added the reason.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+Potential applications are:
+
+* Introspection topics
+* Service discovery
+
+## Design
+
+### Terminology
+
+* Transaction: Atomic state of the world, which is changed by a write operation
```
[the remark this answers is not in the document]
mossmaurice: You are right. In the code comments I used "world view" for `latestTransaction`. A `Transaction` is a change to the world view.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Current transaction: Transaction in local scope, read in the beginning of each operation
+* Latest transaction: Transaction in the shared memory managment segment
+* Chunk: Untyped piece of memory located in the shared memory payload segment
+* Read position: The chunk, which was used the last time to write data
+* Write position: The opposite chunk, not currently being used by the `StatusPortReader`s
+
+### Contract & properties
+
+* 1:n communication aka broadcast
+* Two memory chunks from shared memory payload segment are used
```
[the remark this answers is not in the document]
mossmaurice: You are right. Removed it.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Data exchanged needs to be trivially-copyable
```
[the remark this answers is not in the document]
mossmaurice: You are right. I added the lock-free requirement.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+* Storing a chunk cannot fail
+* Reading a chunk cannot fail
+    * If no data was sent yet, `callable` is not called
+    * If data was updated in-between read, re-call `callable`
+* Not part of the user API in the first version, only used internally
```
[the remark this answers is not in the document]
mossmaurice: Moved to a new section Remarks

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+    * Might become user API for freedom-from-interference separation
+* Reader only needs write access to the data structure and no read access
+* Writing and Reading will be tried indefinitely till possible, hence starvation
+is possible
+* `StatusPortData` is created in the shared memory segment if either a `StatusPortWriter`
```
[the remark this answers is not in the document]
mossmaurice: You are correct. I've moved it to the "Class structure" section.

doc/design/status_port.md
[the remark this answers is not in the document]
mossmaurice: Ok, got your point and agree. I put these alternatives here to prevent questions "Why didn't you do x?". If you don't mind I'd keep this section as it is and make sure with the next design document to keep the alternative section short and precise.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+
+`StatusPortData` stores the latest atomic `Transaction` as well as two pointers
+to `SharedChunk`s in the shared memory payload segment. The lifetime of the two
+chunks is bound to the lifetime of the `StatusPortData` object. To be able to to
+associate the correct readers and writers a `ServiceDescription` is used.
+
+#### Frankenstein object corner cases
+
+How can we ensure in `StatusPortWriter::store()` that there is no `StatusPortReader`
+reading the chunk to which we want to write to? This could be for example a slow
```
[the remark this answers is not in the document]
mossmaurice: TBH I prefer the question-style and hope it is ok for you :upside_down_face:

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+and single argument of`take()` is a `function_ref`, which is executed as long as
+the `take()` was unsuccessful. Furthermore, the `StatusPortReader` has the methods
+necessary to be attachable to a `popo::Listener` (e.g. `{enable,disable}Event`).
+
+`StatusPortData` stores the latest atomic `Transaction` as well as two pointers
+to `SharedChunk`s in the shared memory payload segment. The lifetime of the two
+chunks is bound to the lifetime of the `StatusPortData` object. To be able to to
+associate the correct readers and writers a `ServiceDescription` is used.
+
+#### Frankenstein object corner cases
```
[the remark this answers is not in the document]
mossmaurice: Agree, done.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+
+Latest transaction in shared memory managment segment
++----------------------------------+
+|       ActiveChunk::FIRST         |
+|       abaCounter: 2              |
++----------------------------------+
+```
+
+> ***
+> Special care has to be taken, when writing the `callable`!
```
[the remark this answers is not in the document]
mossmaurice: No, the meaning is verbatim here: "Someone who is writing the code which is passed to the method as `callable`"

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+> ***
+>
+> It must be ensured that the `callable` does not crash or ends up in an endless
+> loop, when working on the data
+
+The memory order ``std::memory_order_release` and `std::memory_order_acquire`
+ensure that the operations below the `load()` "happened-before" and the memory of
```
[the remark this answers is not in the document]
mossmaurice: You are right, the "happen-before" relationship is not the case for read and write. I re-wrote the paragraph.

doc/design/status_port.md
```diff
@@ -0,0 +1,453 @@
...
+                                                                                / \
+                                                                               /   \
+                                                                              /     \
+                                                                             /       \
+                                                                            /         \
+                                                        // Read sychronized data via callable
+                                                        callable(*(m_statusPortDataPtr->chunks[currentReadPosition].data));
+```
+
+### Code example
```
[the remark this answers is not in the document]
mossmaurice: See your point. As the `StatusPort` is already in the prototype phase, I omitted any additional pseudo code. If you want I can add the `.hpp` alongside the design document and refer to it. What do you think?

doc/design/diagrams/status-port/status-port-overview.puml
```diff
@@ -0,0 +1,51 @@
...
+}
+
+class StatusPortData {
+  +latestTransaction: std::atomic<Transaction>
+  +serviceDescription: capro::ServiceDescription
+  +chunks: SharedChunk[2]
+}
+
+class StatusPortReader<T> {
+  +StatusPortReader(statusPortDataPtr: cxx::not_null<StatusPortData* const)
```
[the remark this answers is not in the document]
mossmaurice: Sure, we discussed that before. I used a pointer to be consistent with the current code.

doc/design/diagrams/status-port/status-port-overview.puml
```diff
@@ -0,0 +1,51 @@
...
+  ' Some more methods need to be implemented in the StatusPortReader to be attachable to a Listener
+  ' See https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_examples/waitset/ice_waitset_trigger.cpp
+  ' Open question if copyTake() is wanted
+  '-m_copyOfUserData: T
+  -m_statusPortDataPtr: StatusPortData*
+}
+
+
+class StatusPortWriter<T> {
+  +StatusPortWriter(statusPortDataPtr: cxx::not_null<StatusPortData* const)
```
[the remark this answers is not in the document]
mossmaurice: See this [comment](https://github.com/eclipse-iceoryx/iceoryx/pull/1064#discussion_r799720340)

doc/design/diagrams/status-port/status-port-overview.puml
```diff
@@ -0,0 +1,51 @@
...
+  +store(callable: cxx::function_ref<void(T&)>): void
+  -getMembers() const: const StatusPortData*
```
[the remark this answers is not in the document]
mossmaurice: See this [comment](https://github.com/eclipse-iceoryx/iceoryx/pull/1064#discussion_r799720340)

doc/design/status_port.md
mossmaurice: True, this needs to be seen.

doc/design/diagrams/status-port/status-port-overview.puml
```diff
@@ -0,0 +1,51 @@
...
+skinparam classAttributeIconSize 0
+
+enum ActiveChunk {
+  FIRST
+  SECOND
+}
+
+class Transaction {
+  +activeChunk: ActiveChunk
+  +abaCounter: uint32_t
```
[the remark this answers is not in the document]
mossmaurice: Cool idea! On second thought, I'm not sure if this is actually needed. The overflow of an `uint32_t` is well defined. I would prefer a more readable way to follow the slogan "Write code for humans and not machines". For me the behaviour is more obvious if we use two variables for the two separate responsibilities. What do you think?

doc/design/status_port.md
mossmaurice: Added.

doc/design/status_port.md
mossmaurice: Removed

doc/design/status_port.md
mossmaurice: That's better, adapted it.

doc/design/status_port.md
mossmaurice: Not sure if I got you here. `memcpy` is not used, instead the de-referenced pointer is passed to the `callable`. I added the "why".

doc/design/status_port.md
mossmaurice: I know what you mean, but for me the contract is still "Reading a chunk shall not fail" as the `StatusPort` ensures that the `callable` is called until the read was successful. Which basically means once we return from `read` everything was good.

doc/design/status_port.md
mossmaurice: Upsiiii, fixed :sweat_smile:

doc/design/status_port.md
mossmaurice: Yes, it does. The `ServiceDescription` is part of `StatusPortData`.

doc/design/status_port.md
mossmaurice: You're definitely right when it comes to the restrictions. However, my current view is that it might be the better approach than other alternatives to separate ASIL partitions. Needs to be seen.

doc/design/status_port.md
mossmaurice: Fixed.

doc/design/status_port.md
mossmaurice: You overlooked `e.g.` ;) This is not meant to be a complete list.

doc/design/status_port.md
mossmaurice: You are completely right. `read()` should only be used to very simple data types. For more complex data types I'd suggest to add a 2nd method `copy()`, which does a `memcpy` into a local object and checks if the data is un-corrupted before passing it to the `callable`.

doc/design/status_port.md
mossmaurice: Brilliant idea! Probably such an integration test would need to run for several days to be sure to have statistically meaningful data to argue for minor probability of problems, right?

doc/design/status_port.md
MatthiasKillat: Without copying (memcpy and verify) before performing any operation the data type is basically constrained to be `trivially copyable` and all its operations (if the exist) must be `lock-free thread-safe`, which is fairly restrictive.

Still, it should work with structs with POD at least.

doc/design/diagrams/status-port/status-port-overview.puml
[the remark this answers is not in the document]
mossmaurice: No worries here :relaxed: I completely agree and just didn't implement it as I didn't want to distract from the actual problem. Will refactor it.

doc/design/diagrams/status-port/status-port-overview.puml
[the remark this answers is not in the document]
mossmaurice: I don't see why the overflow is a problem. You can think of the `abaCounter` as a [wheel of fortune](https://huepfefrosch.de/wp-content/uploads/2016/08/Gluecksrad-4.jpg) with consecutive, monotonically nondecreasing integers. The algorithm will call the inequality operator when overflowing and notice that the state of the world has changed `std::numeric_limits<uint32_t>::max != 0` (status_port.md:331).

doc/design/status_port.md
mossmaurice: @elfenpiff Ok, got your point and agree. I added the hint to the design doc, but leave the `copy` code snippet in for now. To easily make the question vividly when implementing it, is that ok for you?

mossmaurice:
Not needed anymore. Closing.

Issue #1052, opened 2022-01-28
ServiceDescription with single string, remove service/instance/event distinction


elfenpiff:
## Brief feature description

The ServiceDescription should be modified so that it only contains one string instead of 3 and iceoryx should not make the distinction between service/instance/event since it has no meaning in our world.

The ServiceDescription would then only consist of one single `cxx::string` which would also have the advantage that we could reduce the string length to 1 in an embedded low memory environment and use the underlying char for service enumeration which would be quiet memory efficient.
For use cases like ara com or someip a binding could introduce logic to reestablish the service/instance/event logic by introducing for instance a separator and name the underlying iceoryx services like `/ServiceName/InstanceName/EventName`.

In `findService` we could then also remove the wildcard option and introduce an option like `extactMatch` or `partialMatch`. This would allow a user for instance to look for `toad` and they would find `AllHailHypnotoad`, `AllGloryToTheHypnotoad` with `partialMatch`. For more fine grained searches the ara com / someip binding could introduce extra logic to filter the results further.

Furthermore it would give other frameworks the advantage to encode additional features cleanly into the service description like for instance a type, a class hash - by just defining some service syntax on their own. 

At the moment we have cyclone dds which uses only partially those 3 identifiers in the internal iceoryx integration which makes it more clear that this is not the way to go. Or look at ros2 when it uses dds and how the services are mapped there.

elfenpiff:
@budrus @FerdinandSpitzschnueffler @MatthiasKillat @elBoberido @mossmaurice What do you think?

If you have additions to the feature descriptions please add them directly in the description.

budrus:
Maybe a good idea. We have today different mappings for rmw_iceoryx, wirhin Cyclone DDS and some kind of ara::com best practice. The constraint would be that this is all possible again and then we have to adapt these

FerdinandSpitzschnueffler:
Good idea. I think especially for the partial search, the prefix tree @MatthiasKillat is working on will be very efficient. 
The question is: when do we want to introduce this change? If we want it for the next release, parts of the refactoring of `ServiceDiscovery` and `ServiceRegistry` that is currently ongoing will be obsolete soon.

elBoberido:
We could also encode the data type into this, e.g. `"/my/service/data_type/version"` -> `"/my/service/int64_t/1"`. Not quite sure if it should be in this string or if there should be a second string for the data type description e.g.
```cpp
struct ServiceDescription {
  string name;
[...]

PR #1069, opened 2022-02-07
Iox #415 refactor service registry

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Reviewer notes

Only change the internal implementation of `ServiceRegistry` to use a vector, but NOT the interface (except for extension with one method `removeAll`).

## References

- Closes nothing

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
@@ -131,7 +131,8 @@ constexpr uint32_t MAX_PROCESS_NUMBER = 300U;
 /// This limitation is coming due to the fixed capacity of the cxx::vector (This doesn't limit the offered number of
 /// instances)
 /// @todo #415 increase number back to 50 once service registry is available via shared memory
-constexpr uint32_t MAX_NUMBER_OF_SERVICES = 100U;
+/// @todo #1074 define and set the limits concerning the service discovery
+constexpr uint32_t MAX_NUMBER_OF_SERVICES = 10U;
 
 // Nodes
 constexpr uint32_t MAX_NODE_NUMBER = 1000U;
--- a/iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
@@ -131,7 +131,7 @@ constexpr uint32_t MAX_PROCESS_NUMBER = 300U;
 /// This limitation is coming due to the fixed capacity of the cxx::vector (This doesn't limit the offered number of
 /// instances)
 /// @todo #415 increase number back to 50 once service registry is available via shared memory
-constexpr uint32_t MAX_NUMBER_OF_SERVICES = 10U;
+constexpr uint32_t MAX_NUMBER_OF_SERVICES = 100U;
 
 // Nodes
 constexpr uint32_t MAX_NODE_NUMBER = 1000U;
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -22,8 +22,6 @@
 #include "iceoryx_hoofs/cxx/vector.hpp"
 #include "iceoryx_posh/capro/service_description.hpp"
 
-#include "iceoryx_hoofs/internal/objectpool/objectpool.hpp"
-
 #include <cstdint>
 #include <map>
 #include <utility>
@@ -53,7 +51,7 @@ class ServiceRegistry
     };
 
     /// @todo #415 should be connected with iox::MAX_NUMBER_OF_SERVICES
-    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 100U;
+    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 2000U;
     static constexpr uint32_t NO_INDEX = MAX_SERVICE_DESCRIPTIONS;
 
 
@@ -61,17 +59,21 @@ class ServiceRegistry
 
     using Entry_t = cxx::optional<ServiceDescriptionEntry>;
     using ServiceDescriptionContainer_t = cxx::vector<Entry_t, MAX_SERVICE_DESCRIPTIONS>;
-    // using ServiceDescriptionPool_t = cxx::ObjectPool<cxx::optional<ServiceDescriptionEntry>,
-    // MAX_SERVICE_DESCRIPTIONS>;
 
     /// @brief Adds given service description to registry
     /// @param[in] serviceDescription, service to be added
     /// @return ServiceRegistryError, error wrapped in cxx::expected
     cxx::expected<Error> add(const capro::ServiceDescription& serviceDescription) noexcept;
 
-    /// @brief Removes given service description from registry if service is found
+    /// @brief Removes given service description from registry if service is found,
+    ///        in case of multiple occurences only one is removed
     /// @param[in] serviceDescription, service to be removed
     void remove(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    /// @brief Removes given service description from registry if service is found,
+    ///        all occurences are removed
+    /// @param[in] serviceDescription, service to be removed
+    void removeAll(const capro::ServiceDescription& serviceDescription) noexcept;
 
     /// @brief Searches for given service description in registry
     /// @param[in] searchResult, reference to the vector which will be filled with the results
@@ -88,12 +90,6 @@ class ServiceRegistry
     const ServiceDescriptionVector_t getServices() const noexcept;
 
   private:
-    /// @todo #859 replace std::multimap with prefix tree
-    ::std::multimap<capro::IdString_t, uint64_t> m_serviceMap;
-    ::std::multimap<capro::IdString_t, uint64_t> m_instanceMap;
-    ::std::multimap<capro::IdString_t, uint64_t> m_eventMap;
-    ServiceDescriptionVector_t m_serviceDescriptionVector;
-
     ServiceDescriptionContainer_t m_serviceDescriptions;
     // ServiceDescriptionPool_t m_serviceDescriptions;
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -25,7 +25,6 @@
 
 
 #include <cstdint>
-#include <map>
 #include <utility>
 
 namespace iox
@@ -91,9 +90,6 @@ class ServiceRegistry
 
     static constexpr uint32_t NO_INDEX = MAX_SERVICE_DESCRIPTIONS;
 
-    // tag type for internal overloads
-    using Wildcard_t = iox::capro::Wildcard_t;
-
     ServiceDescriptionContainer_t m_serviceDescriptions;
 
     // store the last known free Index (if any is known)
@@ -102,45 +98,7 @@ class ServiceRegistry
     uint32_t m_freeIndex{NO_INDEX};
 
   private:
-    // functions for the different search cases
-    // tag types are not needed in all cases to distinguish overloads but kept for consistency
-
     uint32_t findIndex(const capro::ServiceDescription& serviceDescription) const noexcept;
-
-    void find(const capro::IdString_t& service,
-              Wildcard_t,
-              Wildcard_t,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
-
-    void find(Wildcard_t,
-              const capro::IdString_t& instance,
-              Wildcard_t,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
-
-    void find(Wildcard_t,
-              Wildcard_t,
-              const capro::IdString_t& event,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
-
-    void find(const capro::IdString_t& service,
-              const capro::IdString_t& instance,
-              Wildcard_t,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
-
-    void find(const capro::IdString_t& service,
-              Wildcard_t,
-              const capro::IdString_t& event,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
-
-    void find(Wildcard_t,
-              const capro::IdString_t& instance,
-              const capro::IdString_t& event,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
-
-    void find(const capro::IdString_t& service,
-              const capro::IdString_t& instance,
-              const capro::IdString_t& event,
-              ServiceDescriptionVector_t& searchResult) const noexcept;
 
     void getAll(ServiceDescriptionVector_t& searchResult) const noexcept;
 };
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -39,21 +39,21 @@ class ServiceRegistry
     };
 
     using ReferenceCounter_t = uint64_t;
+
     struct ServiceDescriptionEntry
     {
-        ServiceDescriptionEntry(const capro::ServiceDescription& desc, ReferenceCounter_t count)
-            : serviceDescription(desc)
-            , referenceCounter(count)
+        ServiceDescriptionEntry(const capro::ServiceDescription& sd, ReferenceCounter_t count)
+            : serviceDescription(sd)
+            , count(count)
         {
         }
+
         capro::ServiceDescription serviceDescription{};
-        ReferenceCounter_t referenceCounter = 0U;
+        ReferenceCounter_t count{0U};
     };
 
     /// @todo #415 should be connected with iox::MAX_NUMBER_OF_SERVICES
-    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 2000U;
-    static constexpr uint32_t NO_INDEX = MAX_SERVICE_DESCRIPTIONS;
-
+    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 5000U;
 
     using ServiceDescriptionVector_t = cxx::vector<ServiceDescriptionEntry, MAX_SERVICE_DESCRIPTIONS>;
 
@@ -66,7 +66,7 @@ class ServiceRegistry
     cxx::expected<Error> add(const capro::ServiceDescription& serviceDescription) noexcept;
 
     /// @brief Removes given service description from registry if service is found,
-    ///        in case of multiple occurences only one is removed
+    ///        in case of multiple occurrences only one occurrence is removed
     /// @param[in] serviceDescription, service to be removed
     void remove(const capro::ServiceDescription& serviceDescription) noexcept;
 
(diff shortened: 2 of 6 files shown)


iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -38,25 +39,42 @@ class ServiceRegistry
...
-    /// @brief Removes given service description from registry if service is found
+    /// @brief Removes given service description from registry if service is found,
+    ///        in case of multiple occurrences only one occurrence is removed
     /// @param[in] serviceDescription, service to be removed
     void remove(const capro::ServiceDescription& serviceDescription) noexcept;
 
+    /// @brief Removes given service description from registry if service is found,
+    ///        all occurences are removed
+    /// @param[in] serviceDescription, service to be removed
+    void removeAll(const capro::ServiceDescription& serviceDescription) noexcept;
```
MatthiasKillat, line 76: We may also want a method to remove all services (i.e. clear it). But this is part of the next step: improving the interface.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -38,25 +39,42 @@ class ServiceRegistry
...
+        }
+
         capro::ServiceDescription serviceDescription{};
-        ReferenceCounter_t referenceCounter = 0U;
+        ReferenceCounter_t count{0U};
     };
 
     /// @todo #415 should be connected with iox::MAX_NUMBER_OF_SERVICES
-    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 100U;
+    static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 5000U;
```
MatthiasKillat, line 56: Deciding how to configure the values is also part of an interface change PR.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,87 +17,83 @@
 
 #include "iceoryx_posh/internal/roudi/service_registry.hpp"
```
MatthiasKillat, line 18: Remark for reviewers:
The diff is almost useless, all implementations where changed, i.e. nothing except the interfaces remain.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -445,4 +465,109 @@ TEST_F(ServiceRegistry_test, AddingVariousServiceDescriptionAndGetServicesDoesNo
...
+    {
+        auto id = randomString();
+        ServiceDescription sd(fixedId, fixedId, id);
+
+        auto result = sut.add(sd);
+        if (result.has_error())
+        {
+            break;
+        }
+        lastAdded = sd;
```
MatthiasKillat, line 543: Needs to be stored for removal afterwards since we do not know when the registry is full if we add random values (duplicates occupy no extra space).

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -38,25 +39,40 @@ class ServiceRegistry
     };
 
     using ReferenceCounter_t = uint64_t;
+
     struct ServiceDescriptionEntry
     {
+        ServiceDescriptionEntry(const capro::ServiceDescription& sd, ReferenceCounter_t count)
```
elfenpiff, line 45: Please move all implementations into cpp file.

FerdinandSpitzschnueffler (dismissed):

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,87 +17,83 @@
...
+        auto& entry = m_serviceDescriptions[index];
+        entry->count++;
+        return cxx::success<>();
+    }
+
+    // entry does not exist, find a free slot if it exists
+
+    // fast path to a free slot (which was occupied by previously removed entry),
+    // prefer to fill entries close to the front
+    if (m_freeIndex != NO_INDEX)
```
elfenpiff, line 39: Actually one could improve this even further.
1. Introduce the fifo which can be found in `internal/concurrent/fifo.hpp` (its single push/single pop with an atomic) as `concurrent::FiFo<uint32_t, MAX_SERVICE_DESCRIPTIONS> m_freeIndices`
2. move all indices which are freed in there with just `m_freeIndices.push(indexToFree)`.
3. At this position we take a potential free index with `m_freeIndices.pop` and if it is not a nullopt use this one.

The solution above would not even increase the lines of code or would make the code more complex but would optimize it even further.

My personal opinion either keep it simple and make it work first before optimizing it but when one goes that route than take the low hanging fruits with you.

elfenpiff: Or we could use the index queue ... wouldn't than this service registry be thread safe in a lock free manner?! I have not really thought this through

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
elfenpiff, line 45: It looks like you are always using `1` as count for the `ServiceDescriptionEntry` wouldn't it make more sense to set it always to one and remove the second argument?
If we need it later we always can add it again - but YAGNI.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -38,25 +39,40 @@ class ServiceRegistry
...
+        }
+
         capro::ServiceDescription serviceDescription{};
-        ReferenceCounter_t referenceCounter = 0U;
+        ReferenceCounter_t count{0U};
     };
 
     /// @todo #415 should be connected with iox::MAX_NUMBER_OF_SERVICES
+    ///            increase limit to at least >= 1000
     static constexpr uint32_t MAX_SERVICE_DESCRIPTIONS = 100U;
```
elfenpiff, line 57: Why not set it here directly to `iox::MAX_NUMBER_OF_SERVICES` otherwise I may fear that you brake master.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -72,11 +88,61 @@ class ServiceRegistry
...
-    ::std::multimap<capro::IdString_t, uint64_t> m_instanceMap;
-    ::std::multimap<capro::IdString_t, uint64_t> m_eventMap;
-    ServiceDescriptionVector_t m_serviceDescriptionVector;
+    using Entry_t = cxx::optional<ServiceDescriptionEntry>;
+    using ServiceDescriptionContainer_t = cxx::vector<Entry_t, MAX_SERVICE_DESCRIPTIONS>;
+
+    static constexpr uint32_t NO_INDEX = MAX_SERVICE_DESCRIPTIONS;
+
+    // tag type for internal overloads
+    struct Wildcard
```
elfenpiff, line 97: Why not use the definition from: `iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp:33`?

This is compile time and has no overhead.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,87 +17,83 @@
...
-            }
-            else if (it->second > index)
-            {
-                // update index due to removed element
-                it->second--;
-            }
-            it++;
+            entry.reset();
+            // reuse the slot in the next insertion
+            m_freeIndex = index;
```
elfenpiff, line 83: When you would use the fifo approach you could write here `m_freeIndices.push(index);`.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,87 +17,83 @@
...
         }
-    };
+    }
+}
 
-    if (removedElement)
+void ServiceRegistry::removeAll(const capro::ServiceDescription& serviceDescription) noexcept
```
elfenpiff, line 88: I do not completely understand the use case for this method but `removeAll` is somehow misleading. What do you think about `purge`. Or that we rename `remove` into `deregister` and this into `remove`? If this is part of a follow up PR I would leave it as it is as long it is mentioned in some issue.

elfenpiff (changes requested):

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -72,11 +88,61 @@ class ServiceRegistry
...
+    // store the last known free Index (if any is known)
+    // we should not use a queue (or stack) here since they are not optimal
+    // for the filling pattern of a vector (prefer entries close to the front)
+    uint32_t m_freeIndex{NO_INDEX};
+
+  private:
+    // functions for the different search cases
+    // tag types are not needed in all cases to distinguish overloads but kept for consistency
+
+    uint32_t find(const capro::ServiceDescription& serviceDescription) const noexcept;
```
elfenpiff, line 112: There are 10 overloads of `find` and it seems you can create three groups of them when you consider the behavior. This is hard. Could we please change this.

The user api `find` can be named as it is right now. 

This `find` is actually something like `getIndexOfServiceDescription`.

And the remaining finds, why not make one method out of it and name it `fillSearchResultVector`.
```cpp
void fillSearchResultVector(const cxx::optional<capro::IdString_t> & service, const cxx::optional<capro::IdString_t> & instance, const cxx::optional<capro::IdString_t> & event, ServiceDescriptionVector_t& searchResult);
```

In the end I think the code does not become much more complex since one could write it like:
```cpp
for(auto & entry : m_serviceDescriptions) {
  bool isEqual = static_cast<bool>(entry);
  isEqual &= (service) ? (entry->serviceDescription.getServiceIDString() == *service) : true;
  isEqual &= (instance) ? (entry->serviceDescription.getInstanceIDString() == *instance) : true;
  isEqual &= (event) ? (entry->serviceDescription.getEventIDString() == *event) : true;
  if ( isEqual ) {
    searchResult.emplace_back(*entry);
  }
}
```

This would be much less code and complexity as with this approach.

elfenpiff (changes requested):
Please refactor the `find` method variations as suggested.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -106,84 +102,178 @@ void ServiceRegistry::find(ServiceDescriptionVector_t& searchResult,
...
+        return find(WILDCARD, WILDCARD, *event, searchResult);
+    }
+
+    return findAll(searchResult);
+}
+
+const ServiceRegistry::ServiceDescriptionVector_t ServiceRegistry::getServices() const noexcept
+{
+    ServiceDescriptionVector_t allEntries;
+    findAll(allEntries);
```
mossmaurice, line 145: ~~Why is `findAll()` needed? You could also invoke the copy c'tor of `cxx::vector`.~~

Forgot that you are storing `cxx::optional`'s... maybe `getAll()` is better, because we're not really searching anything here.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -445,4 +465,109 @@ TEST_F(ServiceRegistry_test, AddingVariousServiceDescriptionAndGetServicesDoesNo
...
+    // deliberately contains no `0` (need to exclude some char)
+    static const char chars[] = "123456789"
+                                "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                                "abcdefghijklmnopqrstuvwxyz";
+
+    constexpr auto N = string_t::capacity();
+    size = std::min(N, size);
+
+    char a[N + 1];
+    for (uint64_t i = 0; i < size; ++i)
```
mossmaurice, line 490: Nitpick ;)

```suggestion
    for (uint64_t i = 0U; i < size; ++i)
```

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -445,4 +465,109 @@ TEST_F(ServiceRegistry_test, AddingVariousServiceDescriptionAndGetServicesDoesNo
...
+    a[size] = '\0';
+
+    string_t s(a);
+    return s;
+}
+
+TEST_F(ServiceRegistry_test, CanAddMaximumNumberOfDifferentServiceDescriptions)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "76aef6cb-7886-4d64-9188-09bd1be2d335");
+    uint32_t numEntriesAdded = 0;
```
mossmaurice, line 503: Nitpick ;)

```suggestion
    uint32_t numEntriesAdded = 0U;
```

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: Not really important, but I do not care either way. The ctor was just needed for an emplace.

MatthiasKillat:
@elfenpiff Note that it builds on MacOS and Windows with small size (like 100 entries, did not test the limit), but fails with 5000. I suspect it is the stack size but did not look into it yet. Not sure what is set in these environments or whether there is another problem.

We need to increase the limits and ensure it builds on these as well.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: Only this implementation should be affected and will be moved.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: @elfenpiff I considered this, actually my `TypedAllocator` works like this. But here I opted against it, because I want that earlier slots in the vector are filled first.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: Will see whether it works.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: Can do, we already have this dependency.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: Sure, but I will instead remove the optimization.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: I was thinking about the naming myself.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: For the the first part which returns an index I agree but for the second part I do not.

I will not check `isEqual &= (service)` every time I look at an entry and it is in no way clear the compiler can and will optimize it (I tend to think it will not). That is why I check it **once** and then invoke the specific case handler (i.e. overload of find). This is exactly what overloading is for. Furthermore it is event conceivable to have these overloads public for cases where we already know that a specific wildcard is needed (at compile time). This would require testing them in a public interface which I do not want at this point though.

Furthermore, I saw in more sophisticated approaches (with a search tree) that it might be required to have very different case handling. Having different overloads for different cases makes this way easier.

TLDR: The code you present may look more compact but at the cost of being slower by doing unnecessary work and making it harder to change individual cases. Rejected for this reason.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,87 +17,83 @@
...
+    }
+
+    // search from start
+    for (auto& entry : m_serviceDescriptions)
+    {
+        if (!entry)
         {
-            // Due to n:m communication we don't store twice but increase the reference counter
-            element.referenceCounter++;
+            entry.emplace(serviceDescription, 1U);
```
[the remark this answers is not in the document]
MatthiasKillat: Can do.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -17,87 +17,83 @@
...
+{
+    auto index = find(serviceDescription);
+    if (index != NO_INDEX)
     {
-        removeIndexFromMap(m_serviceMap, index);
-        removeIndexFromMap(m_instanceMap, index);
-        removeIndexFromMap(m_eventMap, index);
+        auto& entry = m_serviceDescriptions[index];
+        entry.reset();
+        // reuse the slot in the next insertion
```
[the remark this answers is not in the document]
MatthiasKillat: Not really a good optimization. A queue or stack would be more general but not quite helpful to fill the first slot.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: Initially it was another `find` overload but I wanted to be more expressive. It corresponds to all wildcard search.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -106,84 +102,178 @@ void ServiceRegistry::find(ServiceDescriptionVector_t& searchResult,
                            const cxx::optional<capro::IdString_t>& instance,
                            const cxx::optional<capro::IdString_t>& event) const noexcept
 {
-    ServiceDescriptionVector_t serviceInstanceResult;
-    // Find (K1, K2)
-    // O(log n + log n + max(#PossibleServices + #possiblesInstances) + #intersection)
-    if (instance && service)
+    constexpr Wildcard WILDCARD;
```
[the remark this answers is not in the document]
MatthiasKillat: No it is not, only the type is there but no value of this type. I defined a local constant in the cpp though.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
MatthiasKillat: Yes but this really has to stop because it is clutter and serves no purpose (here). There is a time and place where unsigned and signed literals make a difference, but this is not one.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
elfenpiff: I disagree, lets ask @elBoberido as referee.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -131,7 +131,7 @@ constexpr uint32_t MAX_PROCESS_NUMBER = 300U;
 /// This limitation is coming due to the fixed capacity of the cxx::vector (This doesn't limit the offered number of
 /// instances)
 /// @todo #415 increase number back to 50 once service registry is available via shared memory
-constexpr uint32_t MAX_NUMBER_OF_SERVICES = 10U;
+constexpr uint32_t MAX_NUMBER_OF_SERVICES = 100U;
```
mossmaurice, line 134: I think you can remove the `@todo` above?

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
elBoberido: @MatthiasKillat @elfenpiff As far as I know, this is more or less a stop gap solution right? When we notice that the search is too slow we would use the `PrefixTree` anyway? Is this assumption correct?
Furthermore, even if there is not yet a consensus on #1052, my bet would be that we will go with that route.

With all of this in mind, this looks like premature optimization and might even be removed in the not too distant future. While I understand the pursuit for performance, there are far more (yes, plural) elephants in the room which could give us a far bigger performance boost than this micro optimizations. Cases like false sharing, data distributed over two cache lines, context switches, etc. So all the stuff where the CPU needs to wait because of data synchronization or data not available. Stuff like that is more than an order of magnitude more relevant for performance issues than the stuff we constantly argue about. Please take a glass of wine and carefully read this blog post https://pkolaczk.github.io/server-slower-than-a-laptop/. This is the stuff we need to worry about.

So I'm leaning towards @elfenpiff since it tries to optimize something which does not even cause problems but this optimization adds costs in the form of maintenance and might make refactoring even harder.

MatthiasKillat: @elBoberido I am not sure it really will be a stop gap solution. And yes, we should strive to use atomics correctly as well etc. The #1052 would make the tree more attractive as well. Or a variation of it. But the tree is necessarily more complicated.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: There are a lot of problems related to this, hence a new issue #1074.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
MatthiasKillat: Quite the contrary, more todos await in #1074

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: @elfenpiff Changed under the condition that we (strongly) consider using the prefix tree for this and similar problems in the future. I dislike to use suboptimal (runtime etc.) algorithms just because they are simple. This is often deceiving and minor inefficiencies add up.

FerdinandSpitzschnueffler (approved):

elfenpiff (dismissed):

mossmaurice (approved):

FerdinandSpitzschnueffler (approved):

Issue #938, opened 2021-10-08
application can't create publisher repeatedly with previous one already destroyed


qclzdh:
## Required information

**Operating system:**
Ubuntu 20.04 LTS

**Compiler version:**
GCC 9.3.0

**Observed result or behaviour:**
test application reported some error log:
         **_[ Error ]: Request publisher received no valid publisher port from RouDi
         [Warning]: Service '10:Radar-test14:FrontLeft-test6:Object1:01:01:01:01:01:0' already in use by another process.
         [Warning]: ICEORYX error! POSH__RUNTIME_PUBLISHER_PORT_NOT_UNIQUE_**
Roudi log:
**_2021-10-08 15:27:33.443 [ Debug ]: Registered new application iox-cpp-publisher-helloworld
2021-10-08 15:27:33.444 [ Debug ]: Created new ApplicationPort for application iox-cpp-publisher-helloworld
2021-10-08 15:27:33.445 [ Debug ]: Created new PublisherPort for application iox-cpp-publisher-helloworld
2021-10-08 15:27:33.446 [ Debug ]: Created new PublisherPort for application iox-cpp-publisher-helloworld
2021-10-08 15:27:33.446 [Warning]: Process 'iox-cpp-publisher-helloworld' violates the communication policy by requesting a PublisherPort which is already used by 'iox-cpp-publisher-helloworld' with service '10:Radar-test14:FrontLeft-test6:Object1:01:01:01:01:01:0'.
2021-10-08 15:27:33.446 [Warning]: ICEORYX error! POSH__PORT_MANAGER_PUBLISHERPORT_NOT_UNIQUE
2021-10-08 15:27:33.446 [ Error ]: Could not create PublisherPort for application iox-cpp-publisher-helloworld
2021-10-08 15:27:33.541 [ Debug ]: Destroyed publisher port_**

**Expected result or behaviour:**
I expect the same publisher should be created successfully while previous one already destroyed.

**Conditions where it occurred / Performed steps:**
1. re-build whole iceoryx with "ONE_TO_MANY_ONLY=ON"
2. add following code in iceoryx_examples/icehello/iox_publisher_helloworld.cpp
    ` iox::capro::ServiceDescription serviceDescription{"Radar-test", "FrontLeft-test", "Object"};
      std::shared_ptr<iox::popo::Publisher<RadarObject>> p_publisher =
                  std::make_shared<iox::popo::Publisher<RadarObject>>(serviceDescription);
[...]

PR #1084, opened 2022-02-09
iox-#27 Add client and server port to PortPool and PortManager  [stacked PR #2]

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR adds the client and server to the `PortPool` and `PortManager`. With this changes, the discovery loop is also updated to handle the new ports and it is taken care of resource cleanup.

## References

- Relates to #27


iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
```diff
@@ -40,7 +40,7 @@ typedef struct
     bool offerOnCreate;
 
     /// @brief describes whether a publisher blocks when subscriber queue is full
```
elfenpiff, line 42: Please adjust the comment.

iceoryx_binding_c/include/iceoryx_binding_c/internal/c2cpp_enum_translation.hpp
```diff
@@ -24,7 +24,8 @@
 
 namespace c2cpp
 {
-iox::popo::SubscriberTooSlowPolicy subscriberTooSlowPolicy(const ENUM iox_SubscriberTooSlowPolicy policy) noexcept;
+iox::popo::ConsumerTooSlowPolicy consumerTooSlowPolicy(const ENUM iox_ConsumerTooSlowPolicy policy) noexcept;
```
elfenpiff, line 27: I am sorry again, really - but please adjust the copyright header to keep the wheel of madness running.

If you add a something like:
```
// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
// Praise (H) 2020 - Doom by the honorable Hypnotoad. All glory to the Hypnotoad.
//
// Licensed under the Apache License, Version 2.0 (the "License");
// you may not use this file except in compliance with the License.
// You may obtain a copy of the License at
```

I will approve it.

iceoryx_binding_c/include/iceoryx_binding_c/runtime.h
```diff
@@ -34,7 +34,7 @@ void iox_runtime_init(const char* const name);
 uint64_t iox_runtime_get_instance_name(char* const name, const uint64_t nameLength);
 
 /// @brief initiates the shutdown of the runtime to unblock all potentially blocking publisher
```
elfenpiff, line 36: Please adjust the comment

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -215,6 +219,158 @@ void PortManager::doDiscoveryForSubscriberPort(SubscriberPortType& subscriberPor
...
+        m_portIntrospection.reportMessage(caproMessage);
+        this->sendToAllMatchingServerPorts(caproMessage, clientPortRoudi);
+    });
+
+    /// @todo iox-#27 report to port introspection
+
+    // delete client port from list after DISCONNECT was processed
+    m_portPool->removeClientPort(clientPortData);
+
+    LogDebug() << "Destroyed client port";
```
elfenpiff, line 246: I would think it maybe helpful to add the process name and service description here.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -215,6 +219,158 @@ void PortManager::doDiscoveryForSubscriberPort(SubscriberPortType& subscriberPor
...
+void PortManager::doDiscoveryForClientPort(popo::ClientPortRouDi& clientPort) noexcept
+{
+    clientPort.tryGetCaProMessage().and_then([this, &clientPort](auto caproMessage) {
+        if ((capro::CaproMessageType::CONNECT == caproMessage.m_type)
+            || (capro::CaproMessageType::DISCONNECT == caproMessage.m_type))
+        {
+            /// @todo iox-#27 report to port introspection
+            if (!this->sendToAllMatchingServerPorts(caproMessage, clientPort))
+            {
+                LogDebug() << "capro::CONNECT/DISCONNECT, no matching server!!";
```
elfenpiff, line 275: Could you add the process name of the port owner and the service description? This goes also for all the other log messages.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -215,6 +219,158 @@ void PortManager::doDiscoveryForSubscriberPort(SubscriberPortType& subscriberPor
...
+        {
+            /// @todo iox-#27 add to service registry?
+        }
+        else if (capro::CaproMessageType::STOP_OFFER == caproMessage.m_type)
+        {
+            /// @todo iox-#27 remove from service registry
+        }
+        else
+        {
+            // protocol error
```
elfenpiff, line 365: Maybe we should also add a `LogError` here with process name and service description.

iceoryx_posh/source/roudi/port_pool.cpp
```diff
@@ -160,6 +160,52 @@ PortPool::addSubscriberPort(const capro::ServiceDescription& serviceDescription,
...
+cxx::expected<popo::ClientPortRouDi::MemberType_t*, PortPoolError>
+PortPool::addClientPort(const capro::ServiceDescription& serviceDescription,
+                        mepoo::MemoryManager* const memoryManager,
+                        const RuntimeName_t& runtimeName,
+                        const popo::ClientOptions& clientOptions,
+                        const mepoo::MemoryInfo& memoryInfo) noexcept
+{
+    if (!m_portPoolData->m_clientPortMembers.hasFreeSpace())
+    {
+        errorHandler(Error::kPORT_POOL__CLIENTLIST_OVERFLOW, nullptr, ErrorLevel::MODERATE);
```
elfenpiff, line 182: Could you please add a `LogError` here with service description and runtime name.

iceoryx_posh/source/roudi/port_pool.cpp
```diff
@@ -160,6 +160,52 @@ PortPool::addSubscriberPort(const capro::ServiceDescription& serviceDescription,
...
+cxx::expected<popo::ServerPortRouDi::MemberType_t*, PortPoolError>
+PortPool::addServerPort(const capro::ServiceDescription& serviceDescription,
+                        mepoo::MemoryManager* const memoryManager,
+                        const RuntimeName_t& runtimeName,
+                        const popo::ServerOptions& serverOptions,
+                        const mepoo::MemoryInfo& memoryInfo) noexcept
+{
+    if (!m_portPoolData->m_serverPortMembers.hasFreeSpace())
+    {
+        errorHandler(Error::kPORT_POOL__SERVERLIST_OVERFLOW, nullptr, ErrorLevel::MODERATE);
```
elfenpiff, line 200: Could you please add a `LogError` here with service description and runtime name.

elfenpiff (commented):
In my opinion we should add to every `Log*` message as much information as possible since we do not know what may becomes important during debugging.
Furthermore, I would suggest to always call `LogError` before calling the errorHandler to provide some context to the user. What do you think?

The rest looks fine - only minor things.

iceoryx_binding_c/include/iceoryx_binding_c/publisher.h
elBoberido: Why, the comment is stil true. It's only the data type that changed it's name, note the member

iceoryx_posh/source/roudi/port_pool.cpp
elBoberido: Good point. I C&P this code and will also add the log to the source location. I have to look at the error handling document if it must be `LogWarn` or `LogError`

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: Unfortunately the process name is not available at this point, but I added the service description

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: same as above, the process name is not available ... or did you mean the runtime name?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -279,23 +451,24 @@ void PortManager::handleInterfaces() noexcept
...
+                for (auto& interfacePortData : interfacePortsForInitialForwarding)
+                {
+                    auto interfacePort = popo::InterfacePort(interfacePortData);
+                    // do not offer on same interface
+                    if (serverPort.getCaProServiceDescription().getSourceInterface()
+                        != interfacePort.getCaProServiceDescription().getSourceInterface())
+                    {
+                        interfacePort.dispatchCaProMessage(caproMessage);
+                    }
+                }
```
elBoberido, line 471: @mossmaurice to continue our discussion from the previous PR. I think I can remove the service discovery items since all the publisher and server are already forwarded to the new interface ports

iceoryx_posh/test/moduletests/test_roudi_portpool.cpp
```diff
@@ -391,6 +452,226 @@ TEST_F(PortPool_test, RemoveSubscriberPortIsSuccessful)
     EXPECT_EQ(subscriberPortDataList.size(), 0U);
 }
 
+// END SubscriberPort tests
+
+// BEGIN ClientPort tests
+
+TEST_F(PortPool_test, AddClientPortIsSuccessful)
```
elBoberido, line 459: I'm not that happy with the naming but all tests in this file follow the same schema and this makes it easier to check if all ports have the same tests

iceoryx_posh/source/popo/ports/client_port_data.cpp
```diff
@@ -35,7 +37,8 @@ ClientPortData::ClientPortData(const capro::ServiceDescription& serviceDescripti
     : BasePortData(serviceDescription, runtimeName, clientOptions.nodeName)
     , m_chunkSenderData(memoryManager, clientOptions.serverTooSlowPolicy, HISTORY_CAPACITY_ZERO, memoryInfo)
     , m_chunkReceiverData(getResponseQueueType(clientOptions.responseQueueFullPolicy),
-                          clientOptions.responseQueueFullPolicy)
+                          clientOptions.responseQueueFullPolicy,
+                          memoryInfo)
```
budrus, line 41: memoryInfo is still used for nothing, or?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -225,14 +233,175 @@ void PortManager::doDiscoveryForSubscriberPort(SubscriberPortType& subscriberPor
...
 
+void PortManager::destroyClientPort(popo::ClientPortData* const clientPortData) noexcept
+{
+    cxx::Ensures(clientPortData != nullptr && "clientPortData must not be a nullptr");
+
+    // create temporary client ports to orderly shut this client down
+    popo::ClientPortRouDi clientPortRoudi(*clientPortData);
+    popo::ClientPortUser clientPortUser(*clientPortData);
+
+    clientPortRoudi.releaseAllChunks();
```
budrus, line 254: Hmm, it would make more sense to first disconnect and then release all chunks to ensure that there won't arrive new chunks in between. The order is the same for the subscriber. The question is now if there is a reason for this order....

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -225,14 +233,175 @@ void PortManager::doDiscoveryForSubscriberPort(SubscriberPortType& subscriberPor
...
+
+void PortManager::destroyServerPort(popo::ServerPortData* const serverPortData) noexcept
+{
+    cxx::Ensures(serverPortData != nullptr && "serverPortData must not be a nullptr");
+
+    // create temporary server ports to orderly shut this server down
+    popo::ServerPortRouDi serverPortRoudi{*serverPortData};
+    popo::ServerPortUser serverPortUser{*serverPortData};
+
+    serverPortRoudi.releaseAllChunks();
```
budrus, line 339: same same

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -225,14 +233,175 @@ void PortManager::doDiscoveryForSubscriberPort(SubscriberPortType& subscriberPor
...
+        else
+        {
+            LogWarn() << "CaPro protocol error for server from runtime '" << serverPort.getRuntimeName()
+                      << "' and with service description '" << serverPort.getCaProServiceDescription()
+                      << "'! Cannot handle CaProMessageType '" << caproMessage.m_type << "'";
+            errorHandler(
+                Error::kPORT_MANAGER__HANDLE_SERVER_PORTS_INVALID_CAPRO_MESSAGE, nullptr, iox::ErrorLevel::MODERATE);
+        }
+
+        this->sendToAllMatchingClientPorts(caproMessage, serverPort);
```
budrus, line 401: They also would get the messages that are a protocol error, but their state machine can handle

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -279,23 +450,24 @@ void PortManager::handleInterfaces() noexcept
                 }
             }
         }
-        // also forward services from service registry
-        /// @todo #415 do we still need this? yes but return a copy here to be stored in shared memory via new
-        /// StatusPort's
-        /// @todo iox-#27 I guess this was necessary since a service could be offered via ServiceDiscovery;
-        /// this was removed and I somehow have the feeling this breaks the interface ports with the changes from this
-        /// PR if the CaproServiceType is something different than NON
```
budrus, line 287: I think it should be fine to remove this. We could have services without publishers that were offered via the runtime. This was to support ara::com services with methods only that were covered by another communication channel. I think with the servers now it is fine to forward the server offers and there are no more services.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -436,6 +609,75 @@ void PortManager::sendToAllMatchingSubscriberPorts(const capro::CaproMessage& me
...
+                                               popo::ServerPortRouDi& serverSource) noexcept
+{
+    for (auto clientPortData : m_portPool->getClientPortDataList())
+    {
+        popo::ClientPortRouDi clientPort(*clientPortData);
+        if (clientPort.getCaProServiceDescription() == serverSource.getCaProServiceDescription()
+            && !(serverSource.getClientTooSlowPolicy() == popo::ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA
+                 && clientPort.getResponseQueueFullPolicy() == popo::QueueFullPolicy::BLOCK_PRODUCER))
+        {
+            // send OFFER to client
```
budrus, line 622: could also be STOP_OFFER here?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -436,6 +609,75 @@ void PortManager::sendToAllMatchingSubscriberPorts(const capro::CaproMessage& me
...
+{
+    bool serverFound = false;
+    for (auto serverPortData : m_portPool->getServerPortDataList())
+    {
+        popo::ServerPortRouDi serverPort(*serverPortData);
+        if (clientSource.getCaProServiceDescription() == serverPort.getCaProServiceDescription()
+            && !(serverPort.getClientTooSlowPolicy() == popo::ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA
+                 && clientSource.getResponseQueueFullPolicy() == popo::QueueFullPolicy::BLOCK_PRODUCER))
+        {
+            // send CONNECT to server
```
budrus, line 661: could also be a DISCONNECT?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -646,6 +918,58 @@ PortManager::acquireSubscriberPortData(const capro::ServiceDescription& service,
...
+}
+
+cxx::expected<popo::ServerPortData*, PortPoolError>
+PortManager::acquireServerPortData(const capro::ServiceDescription& service,
+                                   const popo::ServerOptions& serverOptions,
+                                   const RuntimeName_t& runtimeName,
+                                   mepoo::MemoryManager* const payloadDataSegmentMemoryManager,
+                                   const PortConfigInfo& portConfigInfo) noexcept
+{
+    // it is not allowed to have two servers with the same ServiceDescription;
```
budrus, line 947: Just an idea, can we fix #938 if we do check for the server to be deleted here. Sure do the same for the publisher where this bug was reported. Otherwise we would have the same problem with servers. And there it is more likely as two same servers is always not allowed. Something like 

```
    for (auto serverPortData : m_portPool->getServerPortDataList())
    {
        popo::ServerPortRouDi serverPort(*serverPortData);

        if (serverPort.toBeDestroyed())
        {
            destroyServerPort(serverPortData);
        }
    }
```

iceoryx_posh/source/popo/ports/client_port_data.cpp
```diff
@@ -27,6 +27,8 @@ cxx::VariantQueueTypes getResponseQueueType(const QueueFullPolicy policy) noexce
                                                           : cxx::VariantQueueTypes::FiFo_MultiProducerSingleConsumer;
 }
 
+constexpr uint64_t ClientPortData::HISTORY_CAPACITY_ZERO;
```
budrus, line 30: I think I did not see this used somewhere. Do you have more info?

iceoryx_posh/include/iceoryx_posh/capro/service_description.hpp
```diff
@@ -182,6 +183,8 @@ bool serviceMatch(const ServiceDescription& first, const ServiceDescription& sec
 /// @return the reference to `stream` which was provided as input parameter
 std::ostream& operator<<(std::ostream& stream, const ServiceDescription& service) noexcept;
 
+log::LogStream& operator<<(log::LogStream& stream, const ServiceDescription& service) noexcept;
```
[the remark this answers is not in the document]
elBoberido: No, I will update them. I noticed that I missed the doxygen documentation on almost all PRs

iceoryx_posh/source/capro/service_description.cpp
```diff
@@ -235,5 +235,13 @@ std::ostream& operator<<(std::ostream& stream, const ServiceDescription& service
     return stream;
 }
 
+log::LogStream& operator<<(log::LogStream& stream, const ServiceDescription& service) noexcept
+{
+    /// @todo #415 Add classHash, scope and interface
+    stream << "Service: " << service.getServiceIDString() << ", Instance: " << service.getInstanceIDString()
```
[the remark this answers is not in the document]
elBoberido: I've also the feeling we should get rid of scope and refactor classHash into something more generic

iceoryx_posh/source/popo/ports/client_port_data.cpp
elBoberido: correct

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: Probably just an oversight and I copied the behavior

iceoryx_posh/source/popo/ports/client_port_data.cpp
elBoberido: It is just to avoid a magic number and is defined in the header. It is used a few lines further down but since I also use it now outside of the class in the tests it needs a memory location.

iceoryx_posh/source/capro/service_description.cpp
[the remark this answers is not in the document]
budrus: I think I also saw a scope like thing in ROS now

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: In this case not, since the `ServerPort::tryGetCaProMessage` will either return a `nullopt`, `OFFER` or `STOP_OFFER`

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: Moving the release of the chunks after the disconnect is not sufficient. It must be moved to the location after the disconnect was announced to the server

budrus: @elBoberido it was introduced here https://github.com/eclipse-iceoryx/iceoryx/commit/0ae3aa814318e1e0aa060446f14f64259ddf3409 for the subscriber. I thought about it and find no reason why we shouldn't do it the other way round. First ensuring that no one will push to the queue anymore and then clearing the queue seems to be the right order for me. This affects subscriber, client and server

elBoberido: I changed it for all ports. I guess it is not a problem because this part of the code should be guarded by a mutex and therefore the order shouldn't matter but once we remove the locks, it should be as you described it.

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: @budrus for the publisher this could be done in `doesViolateCommunicationPolicy``similar to what I did with the server port. If you agree, I'll create a PR for that issue

iceoryx_posh/test/moduletests/test_roudi_portmanager.cpp
```diff
@@ -15,195 +15,10 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-#include "iceoryx_hoofs/cxx/convert.hpp"
```
elBoberido, line 18: This code was moved to `test_roudi_portmanager_fixture.cpp`

iceoryx_posh/test/moduletests/test_roudi_portmanager_fixture.hpp
```diff
@@ -0,0 +1,227 @@
+// Copyright (c) 2019 - 2021 by Robert Bosch GmbH. All rights reserved.
```
elBoberido, line 1: Almost all of this was copied from `test_roudi_portmanager.cpp`

iceoryx_posh/source/roudi/port_manager.cpp
budrus: @elBoberido This I do not fully get. A publisher can still deliver data to a connected subscriber after the subscriber made a `releaseAllChunks()` and before we unsubscribe (with a lock). The Roudi code here prevents with a lock that new subscribers/publishers can come in parallel and change the connections

iceoryx_posh/source/roudi/port_manager.cpp
budrus: @elBoberido Which means that we'll never be in the else branch. But if `ServerPort::tryGetCaProMessage` would return something else, it is like I described, or? But fine for me as this is all checked on all sides

budrus (approved):

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: @budrus okay, maybe I confused where the lock is. Anyway, the `releaseAllChunks` now happens after dispatching the capro message

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: @budrus yes. I copied it from the publisher port without thinking too much. The else branch should not be taken but it would nevertheless be a good idea to just return in the else branch.

elBoberido: Fixed

iceoryx_posh/test/moduletests/test_roudi_portmanager_client_server.cpp
```diff
@@ -0,0 +1,672 @@
...
+    auto serverPortUser = createServer(serverOptions);
+    auto clientPortUser1 = createClient(clientOptions);
+    auto clientPortUser2 = createClient(clientOptions);
+
+    EXPECT_THAT(clientPortUser1.getConnectionState(), Eq(ConnectionState::CONNECTED));
+    EXPECT_THAT(clientPortUser2.getConnectionState(), Eq(ConnectionState::CONNECTED));
+}
+
+TEST_F(PortManager_test,
+       DoDiscoveryWithMultipleClientsNotConnectedAndOfferingServerResultsSomeClientsConnectedWhenSomeClientsCallConnect)
```
elfenpiff, line 374: Ouh ... I think we need snake case for test names ... the title is very expressive (which is good!) but hard to read due to camel case

iceoryx_posh/test/moduletests/test_roudi_portmanager_client_server.cpp
```diff
@@ -0,0 +1,672 @@
...
+
+TEST_F(PortManager_test, ClientWithDiscardOldestDataAndServerWithDiscardOldestDataAreConnected)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "56871f9d-d7c1-4c3c-b86c-9a1e1dc9fd74");
+    auto clientOptions = createTestClientOptions();
+    clientOptions.responseQueueFullPolicy = QueueFullPolicy::DISCARD_OLDEST_DATA;
+    auto serverOptions = createTestServerOptions();
+    serverOptions.clientTooSlowPolicy = ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA;
+
+    auto clientPortUser1 = createClient(clientOptions);
```
elfenpiff, line 507: Why are you using always 2 client ports in those tests. It is fine by me I was just wondering since I thought one client port should suffice.

iceoryx_posh/test/moduletests/test_roudi_portmanager_client_server.cpp
```diff
@@ -0,0 +1,672 @@
...
+    auto requestHeader = allocateRequestResult.value();
+    requestHeader->setSequenceId(SEQUENCE_ID);
+    clientPortUser.sendRequest(requestHeader);
+
+    auto getRequestResult = serverPortUser.getRequest();
+    ASSERT_FALSE(getRequestResult.has_error());
+    auto receivedRequestHeader = getRequestResult.value();
+    EXPECT_THAT(receivedRequestHeader->getSequenceId(), Eq(SEQUENCE_ID));
+
+    auto allocateResponseResult =
```
elfenpiff, line 656: Do we require tests for the following case:
* serverPortUser calls allocateResponse multiple times for the same receivedRequestHeader
* `sendRequest`, `sendResponse` is called multiple times for the same header
* client sends two requests A, B and server sends responses in opposite order B, A.
* client sends so many requests (A,B,C,D) that the clients response queue (of size 2) overflows as expected so that only (C, D) responses are present.
* server is blocked when client response queue is full and the policy states blocking
* server is not blocked when client response queue is full and the policy states not blocking.

Or are those tests already implemented somewhere else and I overlooked them?

iceoryx_posh/test/moduletests/test_capro_service.cpp
```diff
@@ -428,6 +429,25 @@ TEST_F(ServiceDescription_test, LessThanOperatorReturnsFalseIfEventStringOfFirst
...
+TEST_F(ServiceDescription_test, LogStreamConvertsServiceDescriptionToString)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "42bc3f21-d9f4-4cc3-a37e-6508e1f981c1");
+    Logger_Mock loggerMock;
+
+    const IdString_t SERVICE_ID{"S"};
+    const IdString_t INSTANCE_ID{"I"};
+    const IdString_t EVENT_ID{"E"};
+    const std::string SERVICE_DESCRIPTION_AS_STRING{"Service: S, Instance: I, Event: E"};
+    auto sut = ServiceDescription{SERVICE_ID, INSTANCE_ID, EVENT_ID};
```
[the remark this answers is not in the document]
elBoberido: Why should it matter?

iceoryx_posh/test/moduletests/test_roudi_portmanager_client_server.cpp
```diff
@@ -0,0 +1,672 @@
...
+    ::testing::Test::RecordProperty("TEST_ID", "14070d7b-d8e1-4df5-84fc-119e5e126cde");
+    auto clientOptions = createTestClientOptions();
+    clientOptions.connectOnCreate = true;
+
+    auto clientPortUser = createClient(clientOptions);
+
+    EXPECT_THAT(clientPortUser.getConnectionState(), Eq(ConnectionState::WAIT_FOR_OFFER));
+}
+
+TEST_F(PortManager_test, DoDiscoveryWithClientConnectOnCreateAndNoServerResultsInClientNotConnected)
```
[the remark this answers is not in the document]
elBoberido: Yes, `WAIT_FOR_OFFER` is not `CONNECTED` ;)
Shall I use `Ne(CONNECTED)` to remove the ambiguity?

iceoryx_posh/test/moduletests/test_roudi_portmanager_client_server.cpp
elBoberido: Yeah, we need to write this somewhere down. The question is whether we need to change all the old names

elfenpiff (approved):

budrus (approved):

Issue #1077, opened 2022-02-09
integrationtest to verify limits of iceoryx


elfenpiff:
## Brief feature description

In `iceoryx_posh_types.hpp` defines maximum supported amounts of resources like ports, history size, mempools, shared memory segments etc. free to use by the user.

We require tests to check if those limits are compatible with each other and maybe connect them with if necessary. Additionally, we require integrationtests which verify that iceoryx runs stable even when all resources are in use.

## Detailed information

In a first step it would make sense to split up `iceoryx_posh_types.hpp` into a `iceoryx_config.hpp` file and into a compile time `iceoryx_config_verificator.hpp`. In the config file only the global `constexpr` are defined and we put all the `static_asserts` into the verification file. All the remaining class and function definitions have to be moved to more fitting places. Maybe a `iceoryx_common.hpp` file would make sense.
Finally, when having tightly coupled `MAX_` values then only one of those values should be present in `iceoryx_config.hpp`. The coupling of the other value should happen in `iceoryx_config_internal.hpp` so that the user cannot create incompatible configurations.

### List of coupling problems (not complete):
* Can we move certain `MAX_` values to `platform_settings.hpp` since most system resources have platform specific limits
* Maximum amount of publishers and maximum amount of supported entries in service registry must be equal
* When pursuing approaches for pub/sub find service: max number of apps and max number of subscribers per publisher must be equal
* Do we really require a maximum history size of 2048 when there is no mempool with 2048 chunks of a specific size? A user would get an out of chunks error before the history is completely filled
* Does it make sense to couple maximum chunks hold in parallel to the history size?
* Do we require global `MAX_*` specifications for message queues in there. This is an implementation detail which is platform dependent. Could this be integrated into the posix wrapper?

### Internally used resources

A user does not know what kind of resources we are using internally. So it may happen that they are surprised that they configured 6 publishers but they can only create 1 (since the remaining 5 are used by the introspection). This can be solved by having a `MAX_PUBLISHERS` variable in the public `iceoryx_config.hpp` and a `MAX_PUBLISHER_IN_SYSTEM=MAX_PUBLISHER + PUBLISHERS_USED_BY_INTROSPECTION` in the `iceoryx_config_internal.hpp`.

But still it may happen that a user then breaks its own contract and uses more publishers than configured and maybe internal services are then crashing. In a further step it may makes sense to decouple internal resources completely so that they are always available for roudi.

## To Do

[...]

Issue #1065, opened 2022-02-04
Refactor ServiceDiscovery tests


FerdinandSpitzschnueffler:
## Brief feature description

Many test and variable names in `test_service_discovery.cpp` could be renamed to make the tests more understandable. Some of the tests could also be simplified and it should be checked if tests are missing (see also https://github.com/eclipse-iceoryx/iceoryx/pull/1088#discussion_r805521216) and https://github.com/eclipse-iceoryx/iceoryx/pull/1088#discussion_r806695908.

PR #1088, opened 2022-02-10
iox-#415 Replace IPCchanned-based `findService()` with pub/sub-based one

### LINKED ISSUE #415: service registry as a built-in topic

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. x ] Relevant issues are linked
1. [ ] Assign PR to reviewer

## Notes for Reviewer
* Remove all IPC-chanel-based `findService()` and `changeRegistryCounter` methods
* Send a copy of `serviceRegistry` via a new pub/sub pair
* Make `ServiceDiscovery` attachable to `WaitSet`

### Todo
* Add notification-based unit test cases for `ServiceDiscovery`

## References

- Closes #415 (partly)


iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -38,19 +40,23 @@ class ServiceDiscovery
...
-    findService(const cxx::optional<capro::IdString_t>& service,
-                const cxx::optional<capro::IdString_t>& instance,
-                const cxx::optional<capro::IdString_t>& event) noexcept;
-
-    /// @brief requests the serviceRegistryChangeCounter from the shared memory
-    /// @return pointer to the serviceRegistryChangeCounter
-    /// @todo #415 If this method is still used after refactoring, consider to return a reference so that a nullptr
-    /// check is not necessary
-    virtual const std::atomic<uint64_t>* getServiceRegistryChangeCounter() noexcept;
+    ServiceContainer findService(const cxx::optional<capro::IdString_t>& service,
```
mossmaurice, line 45: @FerdinandSpitzschnueffler I slighty had to adapt the API, removing the `expected` as a return value, as this method can't fail anymore.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -473,31 +458,38 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsMaxServices)
...
-        std::string instance = "i" + iox::cxx::convert::toString(i);
-        iox::capro::ServiceDescription SERVICE_DESCRIPTION(
-            SERVICE, IdString_t(iox::cxx::TruncateToCapacity, instance), "foo");
-        publishers.emplace_back(SERVICE_DESCRIPTION);
-    }
+    IOX_DISCARD_RESULT(ServiceDiscoveryPointer);
+}
 
-    auto serviceContainer = sut.findService(SERVICE, iox::capro::Wildcard, iox::capro::Wildcard);
+TEST_F(ServiceDiscovery_test, ServiceDiscoveryIsAttachableToWaitSet)
```
mossmaurice, line 471: @MatthiasKillat This is how the attaching to the `WaitSet` looks like. Under the hood the `popo::Subscriber` is attached. I'll continue writing the notification-based unit test cases, or did you work on this already?

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -140,11 +135,16 @@ class PortManager
...
 
+    void publishCurrentServiceRegistry() const noexcept;
+
   private:
     RouDiMemoryInterface* m_roudiMemoryInterface{nullptr};
     PortPool* m_portPool{nullptr};
     ServiceRegistry m_serviceRegistry;
     PortIntrospectionType m_portIntrospection;
+
+    /// @todo #415 can we avoid the nullptr here?
```
elfenpiff, line 146: Yes with either a reference or a `cxx::optional<PublisherPortRouDiType::MemberType_t>`

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -62,8 +62,22 @@ PortManager::PortManager(RouDiMemoryInterface* roudiMemoryInterface) noexcept
     }
     auto introspectionMemoryManager = maybeIntrospectionMemoryManager.value();
 
+    /// @todo #415 Move the hard-coded service description to iceoryx_posh_types.hpp?
```
elfenpiff, line 65: Sorry, but I cant resist: https://www.youtube.com/watch?v=ZXsQAXx_ao0

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -140,11 +135,16 @@ class PortManager
...
+    PublisherPortRouDiType::MemberType_t* m_serviceRegistryPublisherPortData{nullptr};
```
MatthiasKillat, line 147: In the long run I would prefer an additional wrapper object (to be named) which contains the Port, the `ServiceRegistry` and other required data. For example currently we lack a mutex for find and add/remove (was broken before with the map as well). This may or may not be required later (depends on final design), mainly whether we call find concurrently.

I prefer to group related information if possible. Currently there is no need to have the registry in shared memory though, but this is also an interesting option for the future.

TLDR: We should maybe consider extensibility and group the data.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +651,43 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
     }
 }
 
+void PortManager::publishCurrentServiceRegistry() const noexcept
+{
+    // Send the new serviceRegistry here and ring the bell to inform all ServiceDiscovery instances about the change
+    if (m_serviceRegistryPublisherPortData)
+    {
+        PublisherPortUserType publisher(m_serviceRegistryPublisherPortData);
```
MatthiasKillat: Can the `Portmanager` have this `PublisherPortUserType publisher` as a member? I know the object creation is lightweight but I am just wondering why we need to create it each time we send.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +651,43 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
...
+                                                           alignof(ServiceRegistry),
+                                                           CHUNK_NO_USER_HEADER_SIZE,
+                                                           CHUNK_NO_USER_HEADER_ALIGNMENT);
+
+        if (!maybeChunkHeader.has_error())
+        {
+            auto sample = static_cast<ServiceRegistry*>(maybeChunkHeader.value()->userPayload());
+
+            // Copy the complete registry
+            *sample = m_serviceRegistry;
```
MatthiasKillat, line 675: We must ensure it is not concurrently updated. This happens in the discovery thread only. As long as we only publish in this thread as we are fine.

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -21,97 +21,47 @@ namespace iox
...
 {
-    /// @todo #415 remove the string mapping, once the find call is done via shared memory
-    capro::IdString_t serviceString;
-    capro::IdString_t instanceString;
-    capro::IdString_t eventString;
-    bool isServiceWildcard = !service;
-    bool isInstanceWildcard = !instance;
-    bool isEventWildcard = !event;
+    ServiceContainer searchResult;
+    roudi::ServiceRegistry::ServiceDescriptionVector_t tempSearchResult;
```
MatthiasKillat, line 29: Could be moved to the point after which we receive the registry.

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -21,97 +21,47 @@ namespace iox
...
-    IpcMessage requestResponse;
-
-    if (!PoshRuntime::getInstance().sendRequestToRouDi(sendBuffer, requestResponse))
+    m_serviceRegistry.find(tempSearchResult, service, instance, event);
+    for (auto& service : tempSearchResult)
     {
-        LogError() << "Could not send FIND_SERVICE request to RouDi\n";
-        errorHandler(Error::kIPC_INTERFACE__REG_UNABLE_TO_WRITE_TO_ROUDI_CHANNEL, nullptr, ErrorLevel::MODERATE);
-        return cxx::error<FindServiceError>(FindServiceError::UNABLE_TO_WRITE_TO_ROUDI_CHANNEL);
+        searchResult.push_back(service.serviceDescription);
```
MatthiasKillat, line 39: We need to consider changing the registry o have the option of not getting the counter as well, then we do not have to transform the results. Can be combined with the lambda interface.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
MatthiasKillat: @mossmaurice I did not write the unit test cases for this.

Personally I am not sure it is nice to force the user to use the waitset like this, especially as the events are not ideally named (as they are `iox::popo::SubscriberEvent::DATA_RECEIVED` and expose an implementation detail). What we can and do wait for is a change of the registry.

On the other hand we can (I think) attach multiple waitsets, i.e. wait in different threads for change.

I am also wondering whether it is possible to miss changes to the registry, i.e. it is cumbersome to use together with find). This is what I am experimenting with and were I think a dedicated API that is easier to use would be better. In theory we can have both, attaching where the user needs to create her own abstractions for wait or we can already provide easier to use abstractions for wait.

Another reason to have the ability to attach(via the waitset interface) is that it works with the listener as well.

MatthiasKillat (commented):
@mossmurice Preliminary review only, the approach should work but I am not sure we will not need he counter anyway. It may also not scale well, but we discussed this already. However, I will refactor the registry again slightly (to also incorporate a counter).

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
mossmaurice: I agree. I will think about it makes sense to include it in this PR already.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: So far, I think we're clean here. Will double check.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
mossmaurice: Good point with the event name, fixed that. As we discussed today, not event are missed. Let's use your dedicated API as an example.

iceoryx_posh/source/runtime/service_discovery.cpp
mossmaurice: I don't the user needs to be aware of how many `popo::Publisher`s are available.

iceoryx_posh/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ add_library(iceoryx_posh # @todo iox-#590 Rename this to iceoryx_posh_core
     source/runtime/node_data.cpp
     source/runtime/node_property.cpp
     source/runtime/shared_memory_user.cpp
+    source/roudi/service_registry.cpp              # @todo iox-#415 Move the service registry into runtime namespace?
```
mossmaurice, line 147: I would tackle this todo with either a follow-up in #415, maybe even after `v2.0`

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: This method here is indirectly called via `deletePortsOfProcess()`, which is called by `removeProcessAndDeleteRespectiveSharedMemoryObjects()`, which is called if a process is unresponsive, and `monitorProcesses()`, which is called cyclically every 100ms. As this is the usually process/port management I'm unsure if I oversaw something or if I discovered a bug? :thinking:

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: I think it would be possible in theory. However, we are using the pattern like above everywhere in the code when working on shared memory ports data. The object changing the data in shared memory lives on the stack and is created on demand. Hence, I would like to stick with the current approach.

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
mossmaurice: Wrapped the pointer in an `cxx::optional`, but decide against grouping things in `struct` for now. We can refactor that in future.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: As the `iceoryx_posh_typeshpp` is already clutter. I prefer if the common info lives in `service_registry.hpp`

doc/design/service-discovery.md
```diff
@@ -172,9 +172,9 @@ Create a new publisher in RouDi which sends a `ServiceRegistryTopic`. This publi
 change in the service registry and to transmit the service discovery registry. The complete old service registry
 (saved locally) would be compared to the new service registry in a new class, extending the public user API.
 
-![overview diagram](diagrams/service-discovery/overview-alternative-b.svg)
+![overview diagram](diagrams/service-discovery/overview-alternative-d.svg)
```
elfenpiff, line 175: I am a little bit confused. The design document states 5 alternatives and argues in detail that alternative E is used.

And it states that alternative D can replace the interface port - which would be possible but an extreme performance hit (this is just a side note, please excuse my jibber jabber).

Furthermore alternative D states that a publisher is `used both to signal a change in the service registry and to transmit the service discovery registry`. 
Is this what you are currently implementing? If so then some parts seem to be missing. Could you please update the design document in a way that it is helpful for the review process and the reviewer knows what was/shall be implemented and why.

iceoryx_posh/CMakeLists.txt
elfenpiff: Please create a separate issue for this and link it in the todo and not use the uber issue 415. We have reached there a point that we cannot retrace there anything anymore since we have to read up on a huge amount of comments and when we would like to find one (browser find does not work) we have to click load additional comments about 100 times.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -69,7 +69,8 @@ constexpr uint32_t MAX_SUBSCRIBER_QUEUE_CAPACITY = MAX_CHUNKS_HELD_PER_SUBSCRIBE
 // 1x publisherPort mempool introspection
 // 1x publisherPort process introspection
 // 3x publisherPort port introspection
-constexpr uint32_t PUBLISHERS_RESERVED_FOR_INTROSPECTION = 5;
+// 1x publisherPort service registry
+constexpr uint32_t PUBLISHERS_RESERVED_FOR_INTROSPECTION_AND_SERVICE_REGISTRY = 6;
```
elfenpiff, line 73: Please create here 3 variables:
```cpp
constexpr uint32_t PUBLISHERS_RESERVED_FOR_INTROSPECTION = 5;
constexpr uint32_t PUBLISHERS_RESERVED_FOR_SERVICE_REGISTRY = 1;
constexpr uint32_t NUMBER_OF_INTERNAL_PUBLISHERS = PUBLISHERS_RESERVED_FOR_INTROSPECTION  + PUBLISHERS_RESERVED_FOR_SERVICE_REGISTRY;
```

iceoryx_posh/include/iceoryx_posh/internal/roudi/port_manager.hpp
```diff
@@ -140,11 +135,15 @@ class PortManager
     cxx::optional<RuntimeName_t>
     doesViolateCommunicationPolicy(const capro::ServiceDescription& service IOX_MAYBE_UNUSED) const noexcept;
 
+    void publishCurrentServiceRegistry() const noexcept;
```
elfenpiff, line 138: Please rename this to `publishServiceRegistry` a distinction between current registry and some other state would only make sense if we would distinguish different states we want to publish.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -31,6 +31,7 @@ namespace iox
 {
 namespace roudi
 {
+static capro::ServiceDescription serviceRegistryService{"ServiceRegistry", "RouDi_ID", "ServiceRegistry"};
```
elfenpiff, line 34: Please never use static globals otherwise we run again into the C++ static initialization order fiasco. Please use the `iceoryx_posh_types.hpp` and store there 3 constexpr `capro::IdStrings` with the service name.

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -18,12 +18,18 @@
...
+#include "iceoryx_posh/popo/subscriber.hpp"
 #include "iceoryx_posh/runtime/posh_runtime.hpp"
 
 namespace iox
 {
 namespace runtime
 {
+enum class ServiceDiscoveryEvent : popo::EventEnumIdentifier
+{
+    SERVICE_REGISTRY_HAS_CHANGED
```
elfenpiff, line 31: Please rename it into `SERVICE_REGISTRY_UPDATED` ... `HAS_CHANGED` describes a state and the other one sounds like an event. See the subscriber state `HAS_DATA` and the event `DATA_RECEIVED`.

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -38,19 +44,23 @@ class ServiceDiscovery
...
+
+  private:
+    void enableEvent(popo::TriggerHandle&& triggerHandle, const ServiceDiscoveryEvent event) noexcept;
+    void disableEvent(const ServiceDiscoveryEvent event) noexcept;
+    void invalidateTrigger(const uint64_t uniqueTriggerId);
+    iox::popo::WaitSetIsConditionSatisfiedCallback
+    getCallbackForIsStateConditionSatisfied(const popo::SubscriberState state);
+    roudi::ServiceRegistry m_serviceRegistry;
+    popo::Subscriber<roudi::ServiceRegistry> m_serviceRegistrySubscriber{
+        roudi::serviceRegistryService, {1U, 1U, iox::NodeName_t("Service Registry"), true}};
```
elfenpiff, line 63: Please create constexpr variables for all arguments. 
Why do we need a node name here? Aren't nodes supposed to be unique but with this approach we would have a `Service Registry` node in every process?!

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -62,8 +62,25 @@ PortManager::PortManager(RouDiMemoryInterface* roudiMemoryInterface) noexcept
     }
     auto introspectionMemoryManager = maybeIntrospectionMemoryManager.value();
 
+    popo::PublisherOptions registryPortOptions;
+    registryPortOptions.historyCapacity = 1U;
```
elfenpiff, line 66: Please defined `static constexpr ` member variables and use them here.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -62,8 +62,25 @@ PortManager::PortManager(RouDiMemoryInterface* roudiMemoryInterface) noexcept
...
+
+    m_serviceRegistryPublisherPortData =
+        acquirePublisherPortData(serviceRegistryService,
+                                 registryPortOptions,
+                                 IPC_CHANNEL_ROUDI_NAME,
+                                 introspectionMemoryManager,
+                                 PortConfigInfo())
+            .or_else([](auto&) {
+                LogError() << "Could not create PublisherPort for service registry!";
+                errorHandler(Error::kPORT_MANAGER__NO_PUBLISHER_PORT_FOR_SERVICE_REGISTRY);
```
elfenpiff, line 78: Please add the error level `Fatal` here and add a `LogFatal`.

When roudi is unable to create a publisher during startup then no application is able to do it during runtime and the system is broken either way.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +659,43 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
...
+void PortManager::publishCurrentServiceRegistry() const noexcept
+{
+    // Send the new serviceRegistry here and ring the bell to inform all ServiceDiscovery instances about the change
+    if (m_serviceRegistryPublisherPortData.has_value())
+    {
+        PublisherPortUserType publisher(m_serviceRegistryPublisherPortData.value());
+        auto maybeChunkHeader = publisher.tryAllocateChunk(sizeof(ServiceRegistry),
+                                                           alignof(ServiceRegistry),
+                                                           CHUNK_NO_USER_HEADER_SIZE,
+                                                           CHUNK_NO_USER_HEADER_ALIGNMENT);
```
elfenpiff, line 671: Why not use `and_then` then we do not need a `maybeChunkHeader`?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +659,43 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
     }
 }
 
+void PortManager::publishCurrentServiceRegistry() const noexcept
+{
+    // Send the new serviceRegistry here and ring the bell to inform all ServiceDiscovery instances about the change
```
elfenpiff, line 664: Please remove the comment, the method is simple enough and in the end you rephrased the method name with some lyrical finesse ;)

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +659,43 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
...
+        auto maybeChunkHeader = publisher.tryAllocateChunk(sizeof(ServiceRegistry),
+                                                           alignof(ServiceRegistry),
+                                                           CHUNK_NO_USER_HEADER_SIZE,
+                                                           CHUNK_NO_USER_HEADER_ALIGNMENT);
+
+        if (!maybeChunkHeader.has_error())
+        {
+            auto sample = static_cast<ServiceRegistry*>(maybeChunkHeader.value()->userPayload());
+
+            // Copy the complete registry
```
elfenpiff, line 677: Yes, the next line exactly states that ... please remove the comment.

Reminder: A comment should explain the why and not the what. The code already does this.

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -21,97 +21,53 @@ namespace iox
...
+                                               const cxx::optional<capro::IdString_t>& event) noexcept
 {
-    /// @todo #415 remove the string mapping, once the find call is done via shared memory
-    capro::IdString_t serviceString;
-    capro::IdString_t instanceString;
-    capro::IdString_t eventString;
-    bool isServiceWildcard = !service;
-    bool isInstanceWildcard = !instance;
-    bool isEventWildcard = !event;
+    // Copy the new service registry if has changed
```
elfenpiff, line 28: Please remove comment.

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -21,97 +21,53 @@ namespace iox
...
+    if (event == ServiceDiscoveryEvent::SERVICE_REGISTRY_HAS_CHANGED)
     {
-        auto deserializationResult =
-            capro::ServiceDescription::deserialize(cxx::Serialization(requestResponse.getElementAtIndex(i)));
-        if (deserializationResult.has_error())
-        {
-            return cxx::error<FindServiceError>(FindServiceError::DESERIALIZATION_FAILED);
-        }
-        serviceContainer.push_back(deserializationResult.value());
+        m_serviceRegistrySubscriber.enableEvent(std::move(triggerHandle), popo::SubscriberEvent::DATA_RECEIVED);
```
elfenpiff, line 49: I would add a switch case in here. In the default case we call the error handler with moderate since somehow we have an internal hickup since we attached the wrong event.
Can happen when one calls `discovery.enableEvent(handle, static_cast<ServiceDiscoveryEvent>(12893))` and this may be caused by memory corruption.

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -21,97 +21,53 @@ namespace iox
...
     }
+}
 
-    if (numberOfElements > capacity)
+void ServiceDiscovery::disableEvent(const ServiceDiscoveryEvent event) noexcept
+{
+    if (event == ServiceDiscoveryEvent::SERVICE_REGISTRY_HAS_CHANGED)
```
elfenpiff, line 55: Same here, please add switch case where the default case calls the error handler.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -473,31 +485,126 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsMaxServices)
...
-            SERVICE, IdString_t(iox::cxx::TruncateToCapacity, instance), "foo");
-        publishers.emplace_back(SERVICE_DESCRIPTION);
+        (*notification)();
     }
 
-    auto serviceContainer = sut.findService(SERVICE, iox::capro::Wildcard, iox::capro::Wildcard);
+    EXPECT_TRUE(callbackWasCalled);
+}
+
+TEST_F(ServiceDiscovery_test, ServiceDiscoveryNotifiedbyWaitSetFindsSingleService)
```
elfenpiff, line 532: 1. Please add tests for the maximum supported amount of services and service results. A tests where you have maximum amount of `radar`, `frontleft`, `event1` ... `eventMAX_SUPPORTED` services and the user searches with one wildcard and a user searches with only wildcards to get all of them.

2. Furthermore, we require wild card tests where some services are found and some are not.

3. Test where we have one/multiple/all services are coming up and do a stop offer and cannot be found anymore

4. Test where one/multiple/all services are performing an offer and we find more and more (with wildcards and without)

elfenpiff (changes requested):
Please add integration tests where you verify that the officially supported limits work.

elfenpiff:
Please try to squash/pack one liner commits together before opening a PR. But not in this PR anymore (otherwise we screw up the already reviewed file list for the reviewers).

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -17,58 +17,70 @@
...
-    auto serviceContainer = sut.findService(iox::capro::Wildcard, iox::capro::Wildcard, iox::capro::Wildcard);
-    ASSERT_FALSE(serviceContainer.has_error());
+    auto searchResult = sut.findService(iox::capro::Wildcard, iox::capro::Wildcard, iox::capro::Wildcard);
 
-    auto searchResult = serviceContainer.value();
 
     for (auto& service : searchResult)
     {
-        EXPECT_THAT(service.getServiceIDString().c_str(), StrEq("Introspection"));
+        EXPECT_THAT(service.getInstanceIDString().c_str(), StrEq("RouDi_ID"));
```
MatthiasKillat, line 83: The sudden capitalization in `RouDi` is not reader-friendly. Not sure we should get rid of this by using a lower case `d`.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -231,9 +243,9 @@ TEST_F(ServiceDiscovery_test, FindServiceWithInstanceAndEventWildcardReturnsAllM
     serviceContainerExp.push_back(SERVICE_DESCRIPTION3);
 
     auto serviceContainer = sut.findService(SERVICE, iox::capro::Wildcard, iox::capro::Wildcard);
-    ASSERT_FALSE(serviceContainer.has_error());
-    ASSERT_THAT(serviceContainer.value().size(), Eq(3U));
-    EXPECT_TRUE(serviceContainer.value() == serviceContainerExp);
+
+    ASSERT_THAT(serviceContainer.size(), Eq(3U));
+    EXPECT_TRUE(serviceContainer == serviceContainerExp);
```
MatthiasKillat, line 248: Those kind of tests are always shaky, since they rely on an implementation detail by assuming an order of returned services which could change. OK for now, but we should strive to write tests that do not use such kind of assumptions.

This could be fixed by e.g. sorting both containers before comparison.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
MatthiasKillat: I agree, we need more tests for these scenarios but not in this PR. We should only check here whether we can attach and are correctly notified in case of a registry change.

The missing tests concern `findService` itself which this PR does not change semantically (only the implementation) and some of the existing cases need refactoring as well as a check for completeness (e.g. test partial wildcard searches etc.). There is an issue for this https://github.com/eclipse-iceoryx/iceoryx/issues/1065 and I suggest using this issue to improve the `findService` tests.

There is no point in checking attach and `findService` in combination if both are tested separately and we just call `findService` in an (anonymous) function on notification.

MatthiasKillat (changes requested):
Looks overall ok but requires some minor changes. Extensive test refactoring should be done in #1065.

iceoryx_posh/source/roudi/port_manager.cpp
elfenpiff: Please don't do this. Yes you are right the file is cluttered but at least we have a central place for all the iceoryx settings. If you place them there we do not know what we are able to configure and what not, the only chance we then have is to look at all classes all over the code and look for settings.

Please move this to the `iceoryx_posh_types.hpp`. We already have also an issue where we would clean this file partially: #1077

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
elfenpiff: @MatthiasKillat I agree the only problem I have is that we may release this feature but we actually did not really test it. I am fine that this is done in #1065 but then this issue has to be fully implemented until 28th of February (code freeze).

FYI @mossmaurice @elBoberido @FerdinandSpitzschnueffler @budrus

elfenpiff: I looked into the issue #1065 and spoke with @FerdinandSpitzschnueffler and as it seems the points I mentioned above are not considered in this issue!

The tests should be added here since the feature was implemented exactly in this pull request and tests should be strongly coupled with features.

Otherwise the two points from the reviewer checklist of every PR do not make sense:
* Unit tests have been written for new behavior
* Tests follow the [best practice for testing

It is also harder and more time consuming for a reviewer to check such things in a follow up PR.

doc/design/service-discovery.md
mossmaurice: This is correct. This PR implements `Alternative D` as an intermediate solution until #1096 is finished and the `StatusPort` is available. Adapted the design doc.


@FerdinandSpitzschnueffler Please update it on your PR with the changes you did.

iceoryx_posh/CMakeLists.txt
mossmaurice: I would not open a completly new issue for such a small refactoring. But overall I agree with you that #415 is no too large. As an intermediate step I added the todo in the first comment in #415, so you can always search it with `Ctrl-F`.

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
mossmaurice: Refactored the strings.

The node name can be handy to see that the different internal publishers in RouDi belong to different modules or in case of an app visually show the users that this is an special subscriber. No, uniqueness comes in place here, it's just a grouping based on names.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: I agree with @MatthiasKillat. IMHO it would be too much overhead to define these local values always `static constexpr` instead we should really try to make all the building blocks in iceoryx `constexpr`'able e.g. `cxx::string` or the `popo::PublisherOptions` would be good candidates to start with.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: True, no `return` is needed here.

iceoryx_posh/CMakeLists.txt
elfenpiff: @mossmaurice Can you please add the todo to the issue description. Then we can cross it out and it clearly belongs to the issue. Otherwise it is just a comment.

doc/design/service-discovery.md
elfenpiff: @mossmaurice If this PR is implementing alternative D you have to adjust alternative D in the design document since it implicitly talks about to publishers. One to signal a change and the other to transmit the service discovery registry.

iceoryx_posh/source/roudi/port_manager.cpp
elfenpiff: @mossmaurice Lets state it differently. We have to follow coding guidelines which explicitly state: no magic numbers.

We have three options now.
1. Let the magic number stick here and add a exception as comment for the static code analysis (which makes the code less readable and clutters it with useless comments).
2. Create a `static constexpr ` member variable and use it here. Then we do not have to write an exception comment for the static code analysis and the code is more readable.
3. Do nothing and create technical debt which someone has to fix - so we create more work for the future. 

The choice is yours...

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
mossmaurice: @elfenpiff I agree with you. However, it seems some of the tests are already implemented:

1. `FindServiceReturnsMaxServices` (line 469)
2. `FindServiceWithServiceWildcardReturnsCorrectServices` (line 251)
   1. Just three not one/many/max 
3. `FindServiceDoesNotReturnNotOfferedServices` (line 403)
4. Not implemented yet

I agree with @MatthiasKillat that unit test cases should not be duplicate for polling and push-based approach. So I would go ahead an implement 4). Does that make sense to you?

MatthiasKillat: If maximum and wildcard tests are not part of https://github.com/eclipse-iceoryx/iceoryx/issues/1065 then they need to be added or a new issues created (I favor adding them).

iceoryx_posh/source/roudi/port_manager.cpp
MatthiasKillat: Yeah it is ok but slightly suboptimal.

iceoryx_posh/source/roudi/port_manager.cpp
MatthiasKillat: We are safe then and should avoid mutexes for this (not needed, performance). However, these subtle assumptions are easy to get lost and maybe a comment is in order for anyone changing the code.

iceoryx_posh/source/runtime/service_discovery.cpp
MatthiasKillat: I just have an idea how to considerably optimize the registry and discovery with counters and more. But this is for the future.

doc/design/service-discovery.md
```diff
@@ -262,53 +262,45 @@ myListener.attachEvent(caproSubscriber, DATA_RECEIVED, createNotificationCallbac
 #### Event-based notification: Alternative D
 
 ```cpp
-class DiscoveryInfo
+class ServiceDiscovery
```
MatthiasKillat, line 265: Because of those kind of changes I would argue strongly against code (or at least much code, apart from algorithmic snippets and struct definitions) in those design documents. Can stay like this but should be considered in the future.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -69,8 +69,10 @@ constexpr uint32_t MAX_SUBSCRIBER_QUEUE_CAPACITY = MAX_CHUNKS_HELD_PER_SUBSCRIBE
 // 1x publisherPort mempool introspection
 // 1x publisherPort process introspection
 // 3x publisherPort port introspection
-// 1x publisherPort service registry
-constexpr uint32_t PUBLISHERS_RESERVED_FOR_INTROSPECTION_AND_SERVICE_REGISTRY = 6;
```
MatthiasKillat: Not for now, but in the future it may be useful to break this down into headers for the configuration of individual components (which are included by the main config). Otherwise we end up with one monolithic config file. This would be just for logical grouping.

Resolve me.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -134,6 +136,10 @@ constexpr uint32_t MAX_PROCESS_NUMBER = 300U;
 /// @todo #415 increase number back to 50 once service registry is available via shared memory
 /// @todo #1074 define and set the limits concerning the service discovery
 constexpr uint32_t MAX_NUMBER_OF_SERVICES = 10U;
+constexpr const char SERVICE_REGISTRY_SERVICE_NAME[] = "ServiceRegistry";
```
MatthiasKillat, line 139: I would say the service is ServiceDiscovery and the event ServiceRegistry. Not that it matters much but I find it more logical.

doc/design/service-discovery.md
elfenpiff, line 265: Please remove the implementation from the design document. This is getting extremely fast out of date and absolutely not helpful. A link to the implementation (file path) would be nice.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,26 +660,23 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
...
-        }
+        publisher
+            .tryAllocateChunk(sizeof(ServiceRegistry),
+                              alignof(ServiceRegistry),
+                              CHUNK_NO_USER_HEADER_SIZE,
+                              CHUNK_NO_USER_HEADER_ALIGNMENT)
+            .and_then([&](auto& chunk) {
+                auto sample = static_cast<ServiceRegistry*>(chunk->userPayload());
+
+                *sample = m_serviceRegistry;
```
MatthiasKillat, line 676: Here a comment would be in order that this copy is not concurrent to any modification of `m_serviceRegistry` (as it happens in the same thread where those modifications are happening) and hence safe.

The problem is this does not hold anymore if anyone modifies it elsewhere in the future. Not much we can do about this to enforce it though.

doc/design/service-discovery.md
```diff
@@ -340,7 +332,7 @@ StatusPortReader {
 };
 
 // User facing API
-class DiscoveryInfo
+class ServiceDiscovery
```
elfenpiff, line 335: Why adding an implementation for an alternative which we are currently not pursuing?

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -133,6 +136,10 @@ constexpr uint32_t MAX_PROCESS_NUMBER = 300U;
...
```
elfenpiff, line 139: Me like it :D

MatthiasKillat (commented):
We should strongly consider to clean up the design document and remove any alternative we have not implemented and do not plan to implement. Furthermore it should contain as little code as possible. We can do this in a follow up from my point of view.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +660,40 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
     }
 }
 
+void PortManager::publishServiceRegistry() const noexcept
+{
+    if (m_serviceRegistryPublisherPortData.has_value())
```
elfenpiff, line 665: I think we should print at least a warning or even a call to the error handler when this is not set.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -659,19 +660,40 @@ popo::InterfacePortData* PortManager::acquireInterfacePortData(capro::Interfaces
...
+    {
+        PublisherPortUserType publisher(m_serviceRegistryPublisherPortData.value());
+        publisher
+            .tryAllocateChunk(sizeof(ServiceRegistry),
+                              alignof(ServiceRegistry),
+                              CHUNK_NO_USER_HEADER_SIZE,
+                              CHUNK_NO_USER_HEADER_ALIGNMENT)
+            .and_then([&](auto& chunk) {
```
elfenpiff, line 673: Here we should also print at least a warning if we cannot allocate another chunk.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
elfenpiff: @mossmaurice I agree with implementing only 4) Thanks for looking up the test cases!

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: @elfenpiff I don't really see an issue here, also in terms of maintainability. Let's ping @elBoberido to get his opinion

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: Added

doc/design/service-discovery.md
mossmaurice: I disagree, design documents always go along with code and tests. In my experience things can never be developed alone.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -473,31 +485,124 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsMaxServices)
...
+
+    listener
+        .attachEvent(
+            sut, ServiceDiscoveryEvent::SERVICE_REGISTRY_CHANGED, iox::popo::createNotificationCallback(testCallback))
+        .or_else([](auto) { GTEST_FAIL() << "Could not attach to listener"; });
+
+    const iox::capro::ServiceDescription SERVICE_DESCRIPTION("Moep", "Fluepp", "Shoezzel");
+    iox::popo::UntypedPublisher publisher(SERVICE_DESCRIPTION);
+
+    this->InterOpWait();
```
elfenpiff, line 583: Please do not use `InterOpWait` anymore since it fails to often in our CI. Use the following approach:
1. include `watch_dog.hpp` and use it as a member in the test like:
```cpp
void SetUp() {
  m_watchdog.watchAndActOnFailure([]{ EXPECT_TRUE(false); std::terminate(); });
}

const iox::units::Duration m_fatalTimeout = 2_s;
Watchdog m_watchdog{m_fatalTimeout};
```
2. Introduce a `std::atomic_bool wasCallbackCalled{false};` in your test in the first line.
3. Set this atomic in the notification callback to true.
4. Introduce a semi busy loop to wait for it
```cpp
while(!wasCallbackCalled.load()) {
  std::this_thread::yield();
}
```
5. At the end you can write something like `EXPECT_TRUE(wasCallbackCalled.load());` The watchdog watches the system so that there is no deadlock occurring and this approach is valid. This will make run the test much faster and reliable and if this does not work the watchdog will terminate the unit test to prevent it from blocking the CI.


Please follow this approach for every instance where you used `InterOpWait`

iceoryx_posh/source/roudi/port_manager.cpp
MatthiasKillat: Can this even happen by design?... I mean it should be created with the `Portmanager` and then live until its destruction, making the whole check only a performance hit...

doc/design/service-discovery.md
mossmaurice: Sorry, but I disagree here. This is not meant to be up to date but provide a higher-level pseudo code, you can think of it as a blueprint or prototype. I think it's good practise the develop a design doc together with some code and tests and later when actually doing the implementation updating the design doc with the latest knowledge gain.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
MatthiasKillat: For old tests we may postpone this to #1065 

For new ones I agree.

doc/design/service-discovery.md
elfenpiff: But this is not a higher level pseude code. It is a cut and paste from the actual implementation containing even the todos and remarks.

doc/design/service-discovery.md
mossmaurice: For readers to see the differences to the other alternatives right away and not confuse readers with naming differences which are irrelevant. Think of our futures selves which want to understand what happened in 2022 :wink:

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
mossmaurice: Good catch! Adapted for new test, existing tests to be done as part of #1065

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -47,13 +48,17 @@ class ServiceDiscovery_test : public RouDi_GTest
...
-    static bool callbackWasCalled;
+    static std::atomic_bool callbackWasCalled;
     static ServiceContainer serviceContainer;
 
     void SetUp()
     {
         callbackWasCalled = false;
         serviceContainer.clear();
+        m_watchdog.watchAndActOnFailure([] {
+            EXPECT_TRUE(false);
```
MatthiasKillat, line 59: Could use `FAIL();`. Do we really need to call terminate? Only needed if we cannot recover and run further tests. 

We may have a thread in deadlock at this point (we cannot know by waiting a finite time), so we probably need to. There always can be an arbitrary error in implementation causing a deadlock (but this is the case in any test and we do not use a watchdog every time ...). If we are sure that we can resolve the deadlock by e.g. destroying the sut we would not have to terminate.

Resolve me.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -66,10 +71,14 @@ class ServiceDiscovery_test : public RouDi_GTest
     {
         serviceContainer = serviceDiscoveryPointer->findService(
             service->getServiceIDString(), service->getInstanceIDString(), service->getEventIDString());
+        callbackWasCalled = true;
     }
+
+    const iox::units::Duration m_fatalTimeout = 2_s;
```
MatthiasKillat, line 77: `constexpr` is probably not possible.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -580,9 +589,12 @@ TEST_F(ServiceDiscovery_test, ServiceDiscoveryIsNotifiedByListenerAboutSingleSer
...
     iox::popo::UntypedPublisher publisher(SERVICE_DESCRIPTION);
 
-    this->InterOpWait();
+    while (!callbackWasCalled.load())
+    {
+        std::this_thread::yield();
+    }
 
-    EXPECT_TRUE(callbackWasCalled);
+    EXPECT_TRUE(callbackWasCalled.load());
```
MatthiasKillat, line 597: We would not have made it out of the loop otherwise I think but ok. (So in theory we could call an unconditional `SUCCEED();`.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -473,31 +494,130 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsMaxServices)
...
+    waitSet
+        .attachEvent(sut,
+                     ServiceDiscoveryEvent::SERVICE_REGISTRY_CHANGED,
+                     0U,
+                     iox::popo::createNotificationCallback(searchForService, serviceDescriptionToSearchFor))
+        .or_else([](auto) { GTEST_FAIL() << "Could not attach to wait set"; });
+
+    iox::popo::UntypedPublisher publisher(serviceDescriptionToSearchFor);
+
+    auto notificationVector = waitSet.wait();
```
MatthiasKillat, line 556: In theory this can deadlock here and block the tests forever. Strictly speaking we would require a watchdog here a well. In particular the other thread from the watchdog can unblock the wait with `waitset.markForDestruction();`(strange name but still). In this case the test would fail but subsequent tests could continue.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
mossmaurice: See your point. However, my eye always searches for an `EXPECT_*` at the end of a test case, so I would like to leave it here in this verbose way, if you don't mind.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
mossmaurice: True, I added a timeout parameter to the call of `timedWait`.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: Moved to `iceoryx_posh_types.hpp` :heavy_check_mark:

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -618,45 +596,141 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsMaxServices)
...
+        .attachEvent(sut,
+                     ServiceDiscoveryEvent::SERVICE_REGISTRY_CHANGED,
+                     0U,
+                     iox::popo::createNotificationCallback(testCallback))
+        .or_else([](auto) { GTEST_FAIL() << "Could not attach to wait set"; });
+
+    const iox::capro::ServiceDescription SERVICE_DESCRIPTION("Moep", "Fluepp", "Shoezzel");
+    iox::popo::UntypedPublisher publisher(SERVICE_DESCRIPTION);
+
+    auto notificationVector = waitSet.timedWait(1_s);
```
MatthiasKillat, line 644: 1s might not be enough in some cases if e.g. the CI is busy ... but ok for now.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
MatthiasKillat: It`s fine, just saying as I am hardwired to avoid computation if the result is already determined.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -17,32 +17,41 @@
...
+
 class ServiceDiscovery_test : public RouDi_GTest
 {
   public:
     void SetUp() override
     {
         searchResultOfFindServiceWithFindHandler.clear();
+        callbackWasCalled = false;
+        serviceContainer.clear();
+        m_watchdog.watchAndActOnFailure([] { GTEST_FAIL(); });
```
MatthiasKillat, line 54: Now this holds for every test. OK, but might be overkill.

MatthiasKillat (approved):

iceoryx_posh/source/roudi/port_manager.cpp
elfenpiff: As I stated, the choice is yours.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
elfenpiff, line 54: This is wrong. The watchdog should terminate the process when it runs into a deadlock to avoid blocking the CI for 6 hours.
What you are doing here is just fail the test and block the CI for 6 hours.

Please add a `std::terminate`.

elfenpiff (changes requested):

elfenpiff (approved):

PR #1089, opened 2022-02-10
Iox #27 add typed client and server api [stacked PR #5]

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR adds the typed API. The functionality is complete.

Test will be added in PR #1140 

## References

- Relates to #27

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
@@ -22,6 +22,7 @@
 #include "iceoryx_posh/internal/popo/request_deleter.hpp"
 #include "iceoryx_posh/internal/popo/response_deleter.hpp"
 #include "iceoryx_posh/internal/popo/rpc_interface.hpp"
+#include "iceoryx_posh/internal/popo/typed_port_api_trait.hpp"
 #include "iceoryx_posh/popo/client_options.hpp"
 #include "iceoryx_posh/popo/request.hpp"
 #include "iceoryx_posh/popo/response.hpp"
@@ -35,15 +36,8 @@ namespace popo
 template <typename Req, typename Res, typename BaseClientT = BaseClient<>>
 class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
 {
-    static_assert(!std::is_void<Req>::value, "The type `Req` must not be void. Use the UntypedClient for void types.");
-    static_assert(!std::is_void<Res>::value, "The type `Res` must not be void. Use the UntypedClient for void types.");
-
-    static_assert(!std::is_const<Req>::value, "The type `Req` must not be const.");
-    static_assert(!std::is_const<Res>::value, "The type `Res` must not be const.");
-    static_assert(!std::is_reference<Req>::value, "The type `Req` must not be a reference.");
-    static_assert(!std::is_reference<Res>::value, "The type `Res` must not be a reference.");
-    static_assert(!std::is_pointer<Req>::value, "The type `Req` must not be a pointer.");
-    static_assert(!std::is_pointer<Res>::value, "The type `Res` must not be a pointer.");
+    using RequestTypeAssert = typename TypedPortApiTrait<Req>::Assert;
+    using ResponseTypeAssert = typename TypedPortApiTrait<Res>::Assert;
 
   public:
     /// @brief Constructor for a client
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
@@ -38,11 +38,22 @@ class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
   public:
     explicit ClientImpl(const capro::ServiceDescription& service, const ClientOptions& clientOptions = {}) noexcept;
 
+    /// @brief Get a Request from loaned shared memory and construct the data with the given arguments.
+    /// @param[in] args Arguments used to construct the data.
+    /// @return An instance of the Request that resides in shared memory or an error if unable to allocate memory to
+    /// loan.
+    /// @details The loaned Request is automatically released when it goes out of scope.
     template <typename... Args>
     cxx::expected<Request<Req>, AllocationError> loan(Args&&... args) noexcept;
 
+    /// @brief Sends the given Request and then releases its loan.
+    /// @param request to send.
     void send(Request<Req>&& request) noexcept override;
 
+    /// @brief Take the Response from the top of the receive queue.
+    /// @return Either a Response or a ChunkReceiveResult.
+    /// @details The Response takes care of the cleanup. Don't store the raw pointer to the content of the Response, but
+    /// always the whole Response.
     cxx::expected<Response<const Res>, ChunkReceiveResult> take() noexcept;
 
   private:
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
@@ -37,6 +37,10 @@ class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
 {
   public:
     explicit ClientImpl(const capro::ServiceDescription& service, const ClientOptions& clientOptions = {}) noexcept;
+    ClientImpl(const ClientImpl&) = delete;
+    ClientImpl(ClientImpl&&) = delete;
+    ClientImpl& operator=(const ClientImpl&) = delete;
+    ClientImpl& operator=(ClientImpl&&) = delete;
 
     /// @brief Get a Request from loaned shared memory and construct the data with the given arguments.
     /// @param[in] args Arguments used to construct the data.
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
@@ -46,6 +46,9 @@ class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
     static_assert(!std::is_pointer<Res>::value, "The type `Res` must not be a pointer.");
 
   public:
+    /// @brief Constructor for a client
+    /// @param[in] service is the ServiceDescription for the new client
+    /// @param[in] clientOptions like the queue capacity and queue full policy by a client
     explicit ClientImpl(const capro::ServiceDescription& service, const ClientOptions& clientOptions = {}) noexcept;
     ClientImpl(const ClientImpl&) = delete;
     ClientImpl(ClientImpl&&) = delete;
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
@@ -1,0 +1,63 @@
+// Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IOX_POSH_POPO_CLIENT_IMPL_HPP
+#define IOX_POSH_POPO_CLIENT_IMPL_HPP
+
+#include "iceoryx_posh/capro/service_description.hpp"
+#include "iceoryx_posh/internal/popo/base_client.hpp"
+#include "iceoryx_posh/internal/popo/request_deleter.hpp"
+#include "iceoryx_posh/internal/popo/response_deleter.hpp"
+#include "iceoryx_posh/internal/popo/rpc_interface.hpp"
+#include "iceoryx_posh/popo/client_options.hpp"
+#include "iceoryx_posh/popo/request.hpp"
+#include "iceoryx_posh/popo/response.hpp"
+#include "iceoryx_posh/popo/trigger_handle.hpp"
+#include "iceoryx_posh/runtime/posh_runtime.hpp"
+
+namespace iox
+{
+namespace popo
+{
+template <typename Req, typename Res, typename BaseClientT = BaseClient<>>
+class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
+{
+  public:
+    explicit ClientImpl(const capro::ServiceDescription& service, const ClientOptions& clientOptions = {}) noexcept;
+
+    template <typename... Args>
+    cxx::expected<Request<Req>, AllocationError> loan(Args&&... args) noexcept;
+
+    void send(Request<Req>&& request) noexcept override;
+
+    cxx::expected<Response<const Res>, ChunkReceiveResult> take() noexcept;
+
+  private:
+    using BaseClientT::port;
+
+    cxx::expected<Request<Req>, AllocationError> loanUninitialized() noexcept;
+
+    using RequestSampleDeleter = RequestDeleter<typename BaseClientT::PortType>;
+    RequestSampleDeleter m_requestDeleter{port()};
+    using ResponseSampleDeleter = ResponseDeleter<typename BaseClientT::PortType>;
+    ResponseSampleDeleter m_responseDeleter{port()};
+};
+} // namespace popo
+} // namespace iox
+
+#include "iceoryx_posh/internal/popo/client_impl.inl"
+
+#endif // IOX_POSH_POPO_CLIENT_IMPL_HPP
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
@@ -35,6 +35,16 @@ namespace popo
 template <typename Req, typename Res, typename BaseClientT = BaseClient<>>
 class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
 {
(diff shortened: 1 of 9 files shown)


iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace iox
+{
+namespace popo
+{
+template <typename Req, typename Res, typename BaseClientT = BaseClient<>>
+class ClientImpl : public BaseClientT, public RpcInterface<Request<Req>>
+{
+  public:
+    explicit ClientImpl(const capro::ServiceDescription& service, const ClientOptions& clientOptions = {}) noexcept;
```
elfenpiff, line 39: I know it is an internal class but could you please add some doxygen.

iceoryx_posh/include/iceoryx_posh/internal/popo/server_impl.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace iox
+{
+namespace popo
+{
+template <typename Req, typename Res, typename BaseServerT = BaseServer<>>
+class ServerImpl : public BaseServerT, public RpcInterface<Response<Res>>
+{
+  public:
+    explicit ServerImpl(const capro::ServiceDescription& service, const ServerOptions& serverOptions = {}) noexcept;
```
elfenpiff, line 39: Please add some doxygen

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
```diff
@@ -0,0 +1,83 @@
...
+
+template <typename Req, typename Res, typename BaseClientT>
+cxx::expected<Request<Req>, AllocationError> ClientImpl<Req, Res, BaseClientT>::loanUninitialized() noexcept
+{
+    auto result = port().allocateRequest(sizeof(Req), alignof(Req));
+    if (result.has_error())
+    {
+        return cxx::error<AllocationError>(result.get_error());
+    }
+    else
```
elfenpiff, line 41: It is not a hard must but some compiler may throw a warning here since you have no return at the end of the function even when this covers all cases.
Would you mind to do just an early return here and remove the `else` and just put the code after the early return

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
```diff
@@ -0,0 +1,83 @@
...
+cxx::expected<Request<Req>, AllocationError> ClientImpl<Req, Res, BaseClientT>::loan(Args&&... args) noexcept
+{
+    return std::move(
+        loanUninitialized().and_then([&](auto& request) { new (request.get()) Req(std::forward<Args>(args)...); }));
+}
+
+template <typename Req, typename Res, typename BaseClientT>
+void ClientImpl<Req, Res, BaseClientT>::send(Request<Req>&& request) noexcept
+{
+    auto payload = request.release(); // release the Request ownership of the chunk before publishing
```
elfenpiff, line 61: Here it is maybe interesting why are you doing this before and not after `sendRequest` otherwise I find this comment a little bit obvious.

iceoryx_posh/include/iceoryx_posh/internal/popo/server_impl.inl
```diff
@@ -0,0 +1,86 @@
...
+cxx::expected<Response<Res>, AllocationError>
+ServerImpl<Req, Res, BaseServerT>::loanUninitialized(const Request<const Req>& request) noexcept
+{
+    const auto* requestHeader = &request.getRequestHeader();
+    auto result = port().allocateResponse(requestHeader, sizeof(Req), alignof(Req));
+    if (result.has_error())
+    {
+        return cxx::error<AllocationError>(result.get_error());
+    }
+    else
```
elfenpiff, line 57: Could you please perform just an early return and move the code inside the else branch outside and remove the `else`

iceoryx_posh/include/iceoryx_posh/internal/popo/server_impl.inl
```diff
@@ -0,0 +1,86 @@
...
+                                                                                      Args&&... args) noexcept
+{
+    return std::move(loanUninitialized(request).and_then(
+        [&](auto& response) { new (response.get()) Res(std::forward<Args>(args)...); }));
+}
+
+template <typename Req, typename Res, typename BaseServerT>
+void ServerImpl<Req, Res, BaseServerT>::send(Response<Res>&& response) noexcept
+{
+    auto payload = response.release(); // release the Request ownership of the chunk before publishing
```
elfenpiff, line 78: Again, why is more interesting than what.

iceoryx_posh/include/iceoryx_posh/popo/client.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace iox
+{
+namespace popo
+{
+template <typename Req, typename Res>
+class Client : public ClientImpl<Req, Res>
+{
+  public:
+    using ClientImpl<Req, Res>::ClientImpl;
```
elfenpiff, line 34: Why using a class when you could do it like:
```cpp
template<typename Req, typename Res>
using Client = ClientImpl<Req, Res>;
```

iceoryx_posh/include/iceoryx_posh/popo/server.hpp
```diff
@@ -0,0 +1,35 @@
...
+#define IOX_POSH_POPO_SERVER_HPP
+
+#include "iceoryx_posh/internal/popo/server_impl.hpp"
+
+namespace iox
+{
+namespace popo
+{
+template <typename Req, typename Res>
+class Server : public ServerImpl<Req, Res>
```
elfenpiff, line 27: Why using a class when you could use a template alias?

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
elBoberido: Oh, should have stated in the PR description that beside the tests, also doxygen is missing and will be added

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
elBoberido: Because `sendRequest` takes the ownership and if I don't release if from the `Request` if would also hold it. This might be a case where `cxx::unique_ptr` would be more appropriate to communicate the ownership transfer to `sendRequest`. Currently all the ports use raw pointer. Shall I open an issue to clean this up for all ports?

iceoryx_posh/include/iceoryx_posh/popo/client.hpp
elBoberido: Good question. Initially there was more code. The Publisher and Subscriber do it this way ... we also had some issues with doxygen not showing the documentation of the alias but I guess that can be solved

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
elBoberido: I already updated the comment to emphasis more the why instead of the what

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
```diff
@@ -0,0 +1,81 @@
...
+    auto payload = mepoo::ChunkHeader::fromUserHeader(requestHeader)->userPayload();
+    auto request = cxx::unique_ptr<Req>(reinterpret_cast<Req*>(payload), m_requestDeleter);
+    return cxx::success<Request<Req>>(Request<Req>{std::move(request), *this});
+}
+
+template <typename Req, typename Res, typename BaseClientT>
+template <typename... Args>
+cxx::expected<Request<Req>, AllocationError> ClientImpl<Req, Res, BaseClientT>::loan(Args&&... args) noexcept
+{
+    return std::move(
```
MatthiasKillat, line 51: Not sure why we need to move into the expected (return type). Ideally all should work directly with return value optimization. As far as I know these kind of move conversions are antipattern as they interfere with RVO or at least should not be required. There is also no real move happening from what  understand (expected owns its memory).

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
```diff
@@ -0,0 +1,81 @@
...
+{
+    auto result = port().getResponse();
+    if (result.has_error())
+    {
+        return cxx::error<ChunkReceiveResult>(result.get_error());
+    }
+    auto responseHeader = result.value();
+    auto payload = mepoo::ChunkHeader::fromUserHeader(responseHeader)->userPayload();
+    auto response = cxx::unique_ptr<const Res>(static_cast<const Res*>(payload), m_responseDeleter);
+    return cxx::success<Response<const Res>>(Response<const Res>{std::move(response)});
```
MatthiasKillat, line 75: Why is there a need to construct a `Response` from an `unique_ptr`? In theory it could be constructed from `payload` and some `responseDeleter` directly, could it not? Internally it would probably use a `unique_ptr` to not reinvent the wheel though (as it uses RAII with custom deleter).

Just thinking about the design here ...

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    template <typename... Args>
+    cxx::expected<Request<Req>, AllocationError> loan(Args&&... args) noexcept;
+
+    void send(Request<Req>&& request) noexcept override;
```
MatthiasKillat, line 44: We may also need the capability to release a loan without sending a request at some point. At least we had the case for a publisher in publish-response where we added this later.

May not be needed if everything works and since sending cannot fail we just assume that once the client has called loaned, it has committed itself to eventually send (otherwise it should not have called loan).

`release` or similar can be added later if needed.

MatthiasKillat (commented):
Looks good, only some design questions from my side.

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
budrus, line 44: @elBoberido I think we should check whether there is a server. If not we should return something like NOT_CONNECTED_TO_SERVER. Sure there are a lot of different states with clients and servers that can come and go, but if there is no connection to a server (which we can check with something like `port.getConnectionState()`) we should tell the user that it does not make sense to wait for responses. 
If we do not block the server queue and drop a request because of overflow, a SEVER_REQUEST_QUEUE_OVERFLOW would also be nice (not sure how easy this is but deep in the stack we have this information)

elBoberido: This capability comes from the `SmartChunk` base class and is available for `Request` and `Response`.

I'm not fully happy with the implementation though. It mimics the `unique_ptr` but I'd rather have something like a static `SmartChunk::release(SmartChunk&& sc)`. This has the benefit that e.g. clang will output a warning if the sample is used after a move.

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
elBoberido: Good point. It was modeled as per the Sample, which also takes a `unique_ptr`. I also have to check which implications this has but most probably could change it like you suggested.

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
elBoberido: @budrus the elf had the same suggestion. Additionally he suggested to even return the chunk in case of an overflow. This was not easily doable since it requires quite big changes in the lower layers. We wanted to change this once the functionality is merged to master but it seems that we are running out of time

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
elBoberido: You are right. This was again a c&p from the `Publisher`. I guess it should be changed there too

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.hpp
budrus: @elBoberido Returning the chunk is maybe too much. Detecting and returning an error enum would be fine for me

elBoberido: @budrus okay. Since this change would affect multiple open PRs, would you be fine to do this in a follow up and change it in the port, untyped C++ API, typed C++ API and C binding at once?

budrus: @elBoberido makes sense. So I'll resolve this thread for now

budrus (approved):

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
elBoberido: If you don't mind, I would keep this as is since it is an internal API and I'm not quite sure how much code needs to be changed since this part is shared with the `SmartChunk` which is the base of the `Sample`

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
MatthiasKillat: Ok, then I think we will leave it as is. But I o not understand why `SmartChunk` is movable but not copyable. I would imagine it could have multiple (read) owners in a proper design.

elBoberido: @MatthiasKillat I'm not sure if it was done in the initial implementation of the `Sample` or later on. When a `Publisher` creates a `Sample` it might make sense to not copy it but for a `Subscriber` it most probably should be possible. For this, we would need to replace the `cxx::unique_ptr` with a `cxx::shared_ptr`, which will be interesting to be implemented without the heap.

iceoryx_posh/include/iceoryx_posh/internal/popo/server_impl.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "iceoryx_posh/runtime/posh_runtime.hpp"
+
+namespace iox
+{
+namespace popo
+{
+template <typename Req, typename Res, typename BaseServerT = BaseServer<>>
+class ServerImpl : public BaseServerT, public RpcInterface<Response<Res>>
+{
+    static_assert(!std::is_void<Req>::value, "The type `Req` must not be void. Use the UntypedServer for void types.");
```
elfenpiff, line 38: You have here and in `ClientImpl` some code duplication. What about introducing a `HeaderConcept` like:
```cpp
template <typename T>
struct HeaderConcept {
    static_assert(!std::is_void<T>::value, "Must not be void");
    static_assert(!std::is_const<T>::value, "Must not be const");
    static_assert(!std::is_reference<T>::value, "Must not be a reference");
    static_assert(!std::is_pointer<T>::value, "Must not be a pointer");
    using Evaluate = void; // required and has to be used with typename HeaderConcept<int>::Evaluate otherwise the compiler ignores the static_assert's
};
```

And use it like:
```cpp
template <typename Req, typename Res, typename BaseServerT = int,
          typename RequiresForRequest = typename HeaderConcept<Req>::Evaluate,
          typename RequiresForResponse = typename HeaderConcept<Res>::Evaluate>
class ServerImpl {};
```

Then one would see that a header concept is required (looks a bit like C++20), if something changes we have only one piece of code we have to adjust and not 4 different pieces.

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
MatthiasKillat: It is possible to implement, even without dynamic memory (if I did not miss something, only done mentally), but not pretty or efficient without a factory which can indirectly store the refcounts.

iceoryx_posh/include/iceoryx_posh/internal/popo/server_impl.hpp
MatthiasKillat: @elfenpiff Looks interesting. Can be extended with checks for specific members etc. in the concept should this be needed (with SFINAE). I just do not like the need to introduce artificial template args for this (as with other meta-programming e.g. `enable_if`) which are not really supposed to be used apart from the default values.

I think it is possible/better to transfer the aggregated condition value to some internal bool, i.e. basically create a type trait for headers. Then we can `static_assert` on this as before, like `static_assert(HeaderConcept<Res>::value, "concept not satisfied"). I leave out the details, but its basically a `constexpr bool` on your conditions above.

Then we do not have artificial template args.

elBoberido: @elfenpiff this might be tricky since the asserts have subtle variations

elBoberido: @elfenpiff after a closer look, forget what I wrote

elBoberido: @elfenpiff I agree with @MatthiasKillat to not add additional template args so I adjusted your proposal a bit.

iceoryx_posh/include/iceoryx_posh/popo/publisher.hpp
```diff
@@ -45,16 +46,8 @@ class PublisherInterface
...
-    static_assert(!std::is_void<H>::value, "The user-header `H` must not be void.");
-
-    static_assert(!std::is_const<T>::value, "The type `T` must not be const.");
-    static_assert(!std::is_reference<T>::value, "The type `T` must not be a reference.");
-    static_assert(!std::is_pointer<T>::value, "The type `T` must not be a pointer.");
-
-    static_assert(!std::is_const<H>::value, "The user-header `H` must not be const.");
-    static_assert(!std::is_reference<H>::value, "The user-header `H` must not be a reference.");
-    static_assert(!std::is_pointer<H>::value, "The user-header must `H` not be a pointer.");
+    using DataTypeAssert = typename TypedPortApiTrait<T>::Assert;
```
MatthiasKillat, line 49: This works too to avoid template args (and we keep the fine grained errors of the static asserts).

MatthiasKillat (approved):

elfenpiff (approved):

PR #1094, opened 2022-02-11
Iox #415 additional find service method

### LINKED ISSUE #415: service registry as a built-in topic

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR introduces a functional `findService` to `ServiceDiscovery` and `ServiceRegistry` API. The implementation is not final and will be adapted once #1088 is merged.
Tests are already implemented, but the Changelog will be updated when the function works as intended. 

## References

- Partially closes #415

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -18,6 +18,7 @@
 #define IOX_POSH_ROUDI_SERVICE_REGISTRY_HPP
 
 #include "iceoryx_hoofs/cxx/expected.hpp"
+#include "iceoryx_hoofs/cxx/function_ref.hpp"
 #include "iceoryx_hoofs/cxx/optional.hpp"
 #include "iceoryx_hoofs/cxx/vector.hpp"
 #include "iceoryx_posh/capro/service_description.hpp"
@@ -80,6 +81,20 @@ class ServiceRegistry
               const cxx::optional<capro::IdString_t>& instance,
               const cxx::optional<capro::IdString_t>& event) const noexcept;
 
+    /// @brief Searches for given service description in registry and applies a function to each
+    ///        matching entry, the function may have side effects but cannot change the entry
+    /// @param[in] searchResult, reference to the vector which will be filled with the results
+    /// @param[in] service, string or wildcard (= iox::cxx::nullopt) to search for
+    /// @param[in] instance, string or wildcard (= iox::cxx::nullopt) to search for
+    /// @param[in] event, string or wildcard (= iox::cxx::nullopt) to search for
+    /// @param[in] function, the function to apply to the matching entries
+    void find(const cxx::optional<capro::IdString_t>& service,
+              const cxx::optional<capro::IdString_t>& instance,
+              const cxx::optional<capro::IdString_t>& event,
+              cxx::function_ref<void(const ServiceDescriptionEntry&)> function) const noexcept;
+
+    /// @todo #415 this may not be needed later or we can move applyToAll to the public interface,
+    ///       (we want to avoid large containers on the stack)
     /// @brief Returns all service descriptions as copy
     /// @return ServiceDescriptionVector_t, copy of complete service registry
     const ServiceDescriptionVector_t getServices() const noexcept;
@@ -101,6 +116,8 @@ class ServiceRegistry
     uint32_t findIndex(const capro::ServiceDescription& serviceDescription) const noexcept;
 
     void getAll(ServiceDescriptionVector_t& searchResult) const noexcept;
+
+    void applyToAll(cxx::function_ref<void(const ServiceDescriptionEntry&)> function) const noexcept;
 };
 } // namespace roudi
 } // namespace iox
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -111,7 +111,7 @@ class ServiceRegistry
 
     void getAll(ServiceDescriptionVector_t& searchResult) const noexcept;
 
-    void applyToAll(cxx::function_ref<void(const ServiceDescriptionEntry&)> function) const noexcept;
+    void applyToAll(cxx::function_ref<void(const ServiceDescriptionEntry&)> callable) const noexcept;
 };
 } // namespace roudi
 } // namespace iox
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -81,17 +81,11 @@ class ServiceRegistry
               const cxx::optional<capro::IdString_t>& instance,
               const cxx::optional<capro::IdString_t>& event) const noexcept;
 
-    /// @brief Searches for given service description in registry and applies a function to each
-    ///        matching entry, the function may have side effects but cannot change the entry
-    /// @param[in] searchResult, reference to the vector which will be filled with the results
-    /// @param[in] service, string or wildcard (= iox::cxx::nullopt) to search for
-    /// @param[in] instance, string or wildcard (= iox::cxx::nullopt) to search for
-    /// @param[in] event, string or wildcard (= iox::cxx::nullopt) to search for
-    /// @param[in] function, the function to apply to the matching entries
+    /// @copydoc ServiceDiscovery::findService
     void find(const cxx::optional<capro::IdString_t>& service,
               const cxx::optional<capro::IdString_t>& instance,
               const cxx::optional<capro::IdString_t>& event,
-              cxx::function_ref<void(const ServiceDescriptionEntry&)> function) const noexcept;
+              cxx::function_ref<void(const ServiceDescriptionEntry&)> callable) const noexcept;
 
     /// @todo #415 this may not be needed later or we can move applyToAll to the public interface,
     ///       (we want to avoid large containers on the stack)
--- a/iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
@@ -46,6 +46,16 @@ class ServiceDiscovery
                 const cxx::optional<capro::IdString_t>& instance,
                 const cxx::optional<capro::IdString_t>& event) noexcept;
 
+    /// @brief find all services that match the provided service description and call callback
+    /// @param[in] service service string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] instance instance string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] event event string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] callback to call with the ServiceContainer filled with all matching services
+    void findService(const cxx::optional<capro::IdString_t>& service,
+                     const cxx::optional<capro::IdString_t>& instance,
+                     const cxx::optional<capro::IdString_t>& event,
+                     const cxx::function_ref<void(const ServiceContainer&)>& callback) noexcept;
+
     /// @brief requests the serviceRegistryChangeCounter from the shared memory
     /// @return pointer to the serviceRegistryChangeCounter
     /// @todo #415 If this method is still used after refactoring, consider to return a reference so that a nullptr
--- a/iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
@@ -34,7 +34,7 @@ class ServiceDiscovery
     ServiceDiscovery& operator=(ServiceDiscovery&&) = delete;
     ~ServiceDiscovery() noexcept = default;
 
-    /// @brief find all services that match the provided service description
+    /// @brief Searches all services that match the provided service description
     /// @param[in] service service string to search for, a nullopt corresponds to a wildcard
     /// @param[in] instance instance string to search for, a nullopt corresponds to a wildcard
     /// @param[in] event event string to search for, a nullopt corresponds to a wildcard
@@ -46,17 +46,17 @@ class ServiceDiscovery
                 const cxx::optional<capro::IdString_t>& instance,
                 const cxx::optional<capro::IdString_t>& event) noexcept;
 
-    /// @brief find all services that match the provided service description and call callback
+    /// @brief Searches all services that match the provided service description and applies a function to each of them
     /// @param[in] service service string to search for, a nullopt corresponds to a wildcard
     /// @param[in] instance instance string to search for, a nullopt corresponds to a wildcard
     /// @param[in] event event string to search for, a nullopt corresponds to a wildcard
-    /// @param[in] callback to call with the ServiceContainer filled with all matching services
+    /// @param[in] function to apply to all matching services
     void findService(const cxx::optional<capro::IdString_t>& service,
                      const cxx::optional<capro::IdString_t>& instance,
                      const cxx::optional<capro::IdString_t>& event,
-                     const cxx::function_ref<void(const ServiceContainer&)>& callback) noexcept;
+                     const cxx::function_ref<void(const ServiceContainer&)>& function) noexcept;
 
-    /// @brief requests the serviceRegistryChangeCounter from the shared memory
+    /// @brief Requests the serviceRegistryChangeCounter from the shared memory
     /// @return pointer to the serviceRegistryChangeCounter
     /// @todo #415 If this method is still used after refactoring, consider to return a reference so that a nullptr
     /// check is not necessary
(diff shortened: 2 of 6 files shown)


iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -80,6 +81,20 @@ class ServiceRegistry
               const cxx::optional<capro::IdString_t>& instance,
               const cxx::optional<capro::IdString_t>& event) const noexcept;
 
+    /// @brief Searches for given service description in registry and applies a function to each
+    ///        matching entry, the function may have side effects but cannot change the entry
+    /// @param[in] searchResult, reference to the vector which will be filled with the results
+    /// @param[in] service, string or wildcard (= iox::cxx::nullopt) to search for
+    /// @param[in] instance, string or wildcard (= iox::cxx::nullopt) to search for
+    /// @param[in] event, string or wildcard (= iox::cxx::nullopt) to search for
+    /// @param[in] function, the function to apply to the matching entries
```
elfenpiff, line 90: Since the doxygen docu is identical to the `service_discovery.hpp` why not use `\\\ @copydoc ServiceDiscovery::findService`?

elfenpiff (dismissed):

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -80,6 +81,14 @@ class ServiceRegistry
               const cxx::optional<capro::IdString_t>& instance,
               const cxx::optional<capro::IdString_t>& event) const noexcept;
 
+    /// @copydoc ServiceDiscovery::findService
+    void find(const cxx::optional<capro::IdString_t>& service,
+              const cxx::optional<capro::IdString_t>& instance,
+              const cxx::optional<capro::IdString_t>& event,
+              cxx::function_ref<void(const ServiceDescriptionEntry&)> function) const noexcept;
```
mossmaurice, line 88: Nitpick: `callable` is more general than `function`, think e.g. of functors
```suggestion
              cxx::function_ref<void(const ServiceDescriptionEntry&)> callable) const noexcept;
```

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -46,7 +46,17 @@ class ServiceDiscovery
...
-    /// @brief requests the serviceRegistryChangeCounter from the shared memory
+    /// @brief Searches all services that match the provided service description and applies a function to each of them
+    /// @param[in] service service string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] instance instance string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] event event string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] function to apply to all matching services
+    void findService(const cxx::optional<capro::IdString_t>& service,
+                     const cxx::optional<capro::IdString_t>& instance,
+                     const cxx::optional<capro::IdString_t>& event,
+                     const cxx::function_ref<void(const ServiceContainer&)>& function) noexcept;
```
mossmaurice, line 57:
```suggestion
                     const cxx::function_ref<void(const ServiceContainer&)>& callable) noexcept;
```

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -119,7 +128,7 @@ void ServiceRegistry::find(ServiceDescriptionVector_t& searchResult,
 
             if (match)
             {
-                searchResult.emplace_back(*entry);
+                function(*entry);
```
mossmaurice, line 131: Same here

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -40,10 +40,37 @@ using iox::runtime::ServiceContainer;
...
+
+    void TearDown() override
+    {
+    }
+
     iox::runtime::PoshRuntime* runtime{&iox::runtime::PoshRuntime::initRuntime("Runtime")};
     ServiceDiscovery sut;
+
+    static ServiceContainer searchResultOfFindServiceWithLambda;
+    static void lambda(const ServiceContainer& s)
```
mossmaurice, line 56: Nitpick: This is not a lambda but a static function.
```suggestion
    static void staticFunction(const ServiceContainer& s)
```

mossmaurice (changes requested):
LGTM! One nitpick: Could you please link the commit 542f6efe to iox-#415. It's missing the `#`, thanks!

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: In a mathematical sense function is general enough. And functors in C++ are not even functors (but functions with state). Either is fine here, cf. https://en.wikipedia.org/wiki/Map_(higher-order_function) for concepts.

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
```diff
@@ -46,7 +46,17 @@ class ServiceDiscovery
                 const cxx::optional<capro::IdString_t>& instance,
                 const cxx::optional<capro::IdString_t>& event) noexcept;
 
-    /// @brief requests the serviceRegistryChangeCounter from the shared memory
+    /// @brief Searches all services that match the provided service description and applies a function to each of them
+    /// @param[in] service service string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] instance instance string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] event event string to search for, a nullopt corresponds to a wildcard
+    /// @param[in] function to apply to all matching services
```
MatthiasKillat, line 53:
```suggestion
    /// @param[in] callable function to apply to all matching services
```

iceoryx_posh/include/iceoryx_posh/runtime/service_discovery.hpp
MatthiasKillat: Either is fine IMO and I prefer function as it is called functional programming for a reason.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: No change needed, in functional programming (and math) everything that maps arguments to a result is a function ...

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
MatthiasKillat: Right, I would usually avoid calling these constructs like this in the code anyway and prefer handler/function/descriptive name (as it is a theoretical concept from https://en.wikipedia.org/wiki/Lambda_calculus).

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -40,10 +40,37 @@ using iox::runtime::ServiceContainer;
...
+    {
+        searchResultOfFindServiceWithLambda = s;
```
MatthiasKillat, line 58:
```suggestion
        searchResultOfFindServiceWithFindHandler = s;
```

But this is quite long ...

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -40,10 +40,37 @@ using iox::runtime::ServiceContainer;
...
+    };
 };
 
+ServiceContainer ServiceDiscovery_test::searchResultOfFindServiceWithLambda;
+
+void compareServiceContainers(const ServiceContainer& lhs, const ServiceContainer& rhs)
```
MatthiasKillat, line 64: I assume the underlying container lacks the operator (did not check). If so, ok, not the right PR to add it.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -85,6 +115,12 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsOfferedService)
     ASSERT_FALSE(serviceContainer.has_error());
     ASSERT_THAT(serviceContainer.value().size(), Eq(1U));
     EXPECT_THAT(*serviceContainer.value().begin(), Eq(SERVICE_DESCRIPTION));
+
+    sut.findService(SERVICE_DESCRIPTION.getServiceIDString(),
+                    SERVICE_DESCRIPTION.getInstanceIDString(),
+                    SERVICE_DESCRIPTION.getEventIDString(),
+                    lambda);
```
MatthiasKillat, line 122: Maybe a better name would be 

```suggestion
                    findHandler);
```

or

```suggestion
                    matchHandler);
```
as it is applied to all matching entries.
Applies to all tests.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -408,50 +527,76 @@ TEST_F(ServiceDiscovery_test, FindServiceDoesNotReturnNotOfferedServices)
...
+    sut.findService(SERVICE_DESCRIPTION3.getServiceIDString(), INSTANCE, EVENT, lambda);
+    EXPECT_TRUE(searchResultOfFindServiceWithLambda.empty());
 }
 
 TEST_F(ServiceDiscovery_test, NonExistingServicesAreNotFound)
 {
     ::testing::Test::RecordProperty("TEST_ID", "86b87264-4df4-4d20-9357-06391ca1d57f");
-    const iox::capro::ServiceDescription SERVICE_DESCRIPTION("service", "instance", "event");
+    const IdString_t SERVICE = "service";
+    const IdString_t INCORRECT_SERVICE = "ignatz";
```
MatthiasKillat, line 551:
```suggestion
    const IdString_t NONEXISTENT_SERVICE = "ignatz";
```

MatthiasKillat (commented):
Minor renaming suggested but otherwise it is ok.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
MatthiasKillat, line 56:
```suggestion
    static void findHandler(const ServiceContainer& s)
```

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
mossmaurice: I would prefer the name `callable` as this term is used [throughout the C++ standard](https://eel.is/c++draft/func.def#def:type,callable) and we should be as close to the standard as possible. No hard feelings, though.

MatthiasKillat (approved):

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -119,7 +128,7 @@ void ServiceRegistry::find(ServiceDescriptionVector_t& searchResult,
 
             if (match)
             {
-                searchResult.emplace_back(*entry);
+                callable(*entry);
```
elfenpiff, line 131: Oh I was so strictly against making the `cxx::function_ref` nullable since such simple mistakes like this can happen. This is so dangerous!

But you have to check the `callable` first before calling it since when it does not contain a callable this produces either undefined behavior or a call to the error handler - or a terminate.

So please add:
```
if ( !callable ) {
  return;
}
```

at the top of the method.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -145,13 +155,13 @@ uint32_t ServiceRegistry::findIndex(const capro::ServiceDescription& serviceDesc
     return NO_INDEX;
 }
 
-void ServiceRegistry::getAll(ServiceDescriptionVector_t& searchResult) const noexcept
+void ServiceRegistry::applyToAll(cxx::function_ref<void(const ServiceDescriptionEntry&)> function) const noexcept
```
elfenpiff, line 158: For consistency reasons, could you please rename `function` into `callable`.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -145,13 +155,13 @@ uint32_t ServiceRegistry::findIndex(const capro::ServiceDescription& serviceDesc
...
 {
     for (auto& entry : m_serviceDescriptions)
     {
         if (entry)
         {
-            searchResult.emplace_back(*entry);
+            function(*entry);
```
elfenpiff, line 164: Here we have the same problem. What happens when `function` is not set?

Please add here as well a check right in the beginning of the function.
```
if ( !callable ) {
  return;
}
```

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -89,6 +89,19 @@ ServiceDiscovery::findService(const cxx::optional<capro::IdString_t>& service,
...
+void ServiceDiscovery::findService(const cxx::optional<capro::IdString_t>& service,
+                                   const cxx::optional<capro::IdString_t>& instance,
+                                   const cxx::optional<capro::IdString_t>& event,
+                                   const cxx::function_ref<void(const ServiceContainer&)>& callable) noexcept
+{
+    /// @todo #415 change implementation once PR #1088 is merged
+    auto searchResult = findService(service, instance, event);
+    if (!searchResult.has_error())
+    {
+        callable(searchResult.value());
```
elfenpiff, line 101: What if `callable` is not set?

Please add this to the beginning of the function:
```
if ( !callable ) {
  return;
}
```

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -570,4 +570,42 @@ TEST_F(ServiceRegistry_test, SearchInFullRegistryWorks)
     ASSERT_EQ(searchResult.size(), 1);
 }
 
+using Entry = iox::roudi::ServiceRegistry::ServiceDescriptionEntry;
+
+TEST_F(ServiceRegistry_test, FunctionIsAppliedToAllEntriesInSearchResult)
```
elfenpiff, line 575: Please add here tests for the test case that the callable is not set. This can be done via the default ctor like this:
```cpp
cxx::function_ref<void(const ServiceContainer&)> searchFunction;
```

elfenpiff (changes requested):
Please add a callable check before using it.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -118,6 +118,11 @@ void ServiceRegistry::find(const cxx::optional<capro::IdString_t>& service,
                            const cxx::optional<capro::IdString_t>& event,
                            cxx::function_ref<void(const ServiceDescriptionEntry&)> callable) const noexcept
 {
+    if (!callable)
```
MatthiasKillat, line 121: The more I think about it: it is bad for `function_ref` to be `nullable`, since 
1.  it semantically mimics a reference which is always referencing something (can be dangling, but this is the case here as well)
2. it leads to design flaws like these checks which add nothing but costs and boilerplate

Since its main purpose is to be passed around this is counter productive.

The situation is different for `std::function`, which is supposed to be stored and there it can be useful to default initialize it to something, and that something is naturally ... nothing.

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -650,4 +650,13 @@ TEST_F(ServiceDiscovery_test, FindServiceReturnsContainerOverflowErrorWhenMoreTh
     EXPECT_TRUE(searchResultOfFindServiceWithFindHandler.empty());
 }
 
+TEST_F(ServiceDiscovery_test, FindServiceWithEmptyCallableDoesNotDie)
```
MatthiasKillat, line 653: I would surely hope it does not die but the real semantics we implicitly defined is that it does nothing, even if the search would match something. Feel free to test this, but it is another indicator of a flawed design of `function_ref` or the interface.

MatthiasKillat (approved):
`function_ref` should not be nullable but this is nothing to be resolved here ...

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: @FerdinandSpitzschnueffler Please also add a test case for that, where you call `find()` with supplying a `nullptr` as last argument.

iceoryx_posh/source/roudi/service_registry.cpp
mossmaurice: I see your point, but I'm not sure if we want to go in this strict and limiting direction for `iceoryx_hoofs`, see [my comment](https://github.com/eclipse-iceoryx/iceoryx/issues/1104#issuecomment-1040083614).

elfenpiff (approved):

iceoryx_posh/source/roudi/service_registry.cpp
FerdinandSpitzschnueffler: @mossmaurice I already added a test with an empty function_ref. Calling `find()` with a nullptr would just check that an empty function_ref is created from a nullptr and that should be checked in the function_ref tests.

mossmaurice (approved):

Issue #1066, opened 2022-02-04
Make `asStringLiteral` a template to specialize for enum to string conversion


elBoberido:
## Brief feature description

In order to have a nice output with `std::ostream` and `iox::log::LogStream` the stream operator has to be implemented for both of this types. With `asStringLiteral` as a template, this could be specialized for a specific enum and a blanket implementation for `iox::log::LogStream` could be provided. Unfortunately this cannot be done for `std::ostream`.

## Detailed information

This could be the template in hoofs
```cpp
template <typename T>
constexpr const char* asStringLiteral(const T value)
{
    static_assert(always_false_v(T), "'asStringLiteral' is not specialized for the specified type");
}
```

The logger could then do a blanket implementation like
```cpp
template <typename T>
LogStream& operator<<(LogStream& stream, T value) noexcept
{
    stream << asStringLiteral(value);
    return stream;
}
```

To ease the implementation for `std::ostream` @elfenpiff suggested the following pattern
```cpp
// in header file
template<typename T>
T& operator<<(T& stream, ChunkReceiveResult value) noexcept;

// in cpp file
T& operator<<(T& stream, ChunkReceiveResult value) noexcept
{
    stream << asStringLiteral(value);
    return stream;
}

template log::Stream& operator<<(log::LogStream & stream, ChunkReceiveResult value) noexcept;
template std::ostream& operator<<(std::ostream & stream, ChunkReceiveResult value) noexcept;
```

Maybe these two approaches could be combined to
```cpp
template<typename S, typename E>
S& operator<<(S& stream, E value) noexcept
{
    stream << asStringLiteral(value);
    return stream;
}
```
With some constraints on `E`

elBoberido:
@dkroenke since you already worked on clang-tidy rules and tried to better decouple out modules in hoofs, what do you think would be the best place for a `asStringLiteral` template? `cxx/helplets.hpp`?

Issue #1032, opened 2022-01-21
ErrorHandling concept with user defined actions


elfenpiff:
## Brief feature description

In a safety certified environment the user may would like to define what reaction certain errors cause. Like:
* call to `std::terminate`
* throw an exception
* inform some watch dog and go into a spinlock

Furthermore, to avoid death tests in unit tests customizing the behavior so that exceptions can be thrown can be useful. This would require to get rid of `noexcept`. The idea would be that iceoryx never throws an exceptions but allows the user to overload/define the error handling so that it is capable of throwing custom exception which is at the moment impossible since `noexcept` is in place.

### Details

Iceoryx as a framework should enable the user to pursue any error concept they like with a fitting interface. This error concept should then be used uniformly in all iceoryx components so that we have one single call like `errorHandler(` instead of a wild mixture of `std::terminate`, `cxx::Expects`, `cxx::Ensures`, `assert` etc.

This call may require the origin, cause, severity of the error so that user defined callbacks can be defined to react differently based on origin, severity and cause.

@elBoberido fell asleep on his keyboard and maybe the products of his dreams can provide us with further insights, [see here](https://github.com/eclipse-iceoryx/iceoryx/pull/1027#issuecomment-1018391791)

## Definition of done

The error handling concept is documented in: [doc/design/error-handling.md](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/design/error-handling.md). For this the current error handling concept may have to be discarded/rewritten.

Follow-up issues can then implement the error concept.

### Further insights

We have the ability to allow the user to throw exceptions in callbacks, `cxx::function` or `cxx::function_ref` but noexcept will cause always a `std::terminate`.
This is especially a problem in tests where we avoid fatal errors with `Expects` or the error handler. This is untestable since the errorhandler calls the custom handler and then continues - we require an exception here to make the code testable. 

Think of testing this piece of code:
```cpp
void doStuff(someClass * ptr) {
  iox::cxx::Expects(ptr != nullptr);

  ptr->doStuff();
}
```

A test like
```cpp
auto handle =
    iox::ErrorHandler::setTemporaryErrorHandler([&](auto, auto, auto) { wasErrorHandlerCalled = true; });
doStuff(nullptr);
EXPECT_TRUE(wasErrorHandlerCalled);
```
will always end up with a segfault since the code in `doStuff` continues after the error handler call. But if we would allow exceptions we could throw an exception in the error handler and safely return to the test.
[...]

PR #1100, opened 2022-02-12
iox-#1099 Separate test part from `ErrorHandler` and call `errorHandler()` in `iceoryx_hoofs` only once

### LINKED ISSUE #1099: Separate module specific errors from `iceoryx_hoofs`
## Brief feature description

#590 uncovered an illegal dependency from `iceoryx_hoofs` towards modules using it. While `iceoryx_hoofs` aims to be a library that [can be used independent](https://discourse.ros.org/uploads/short-url/cKUVQ4V8aUu7nB2D4IvldyozD64.pdf) from `iceoryx_posh` and the other modules, as of today it still contains all the error codes from all modules.

## Detailed information

* [x] Separate test specific code from the `ErrorHandler` and templatize `setTemporaryErrorHandler(..)`
* [x] Call `errorHandler(..)` only once in `iceoryx_hoofs` in `Expects` / `Ensures` 
* [x] Remove the 2nd argument of `errorHandler(..)` being `const std::function<void()>& errorCallBack IOX_MAYBE_UNUSED`
* [x] Create separate `error_handling.hpp` for each module using the `ErrorHandler`
   * [x] Rename this `error_handling.cpp` in `hoofs` to `error_handler.hpp`
* [x] Use `asStringLiteral` instead of `toString` from #1066 
* [x] Add explanation in error handling concept about what needs to be implemented for new module
* [ ] Add basic unit tests for `ErrorHandlerMock`

## References

* Comes from #590

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
*  Separate test specific code from the `ErrorHandler` and templatize `setTemporaryErrorHandler(..)`
* Call `errorHandler(..)` only once in `iceoryx_hoofs` in `Expects` / `Ensures`

## References

- Closes #1099 (partly)
- Successor to #1027


CONTRIBUTING.md
```diff
@@ -110,8 +110,8 @@ codebase follows these rules, things are work in progress.
     our code may contain additions which are not compatible with the STL (e.g. `iox::cxx::vector::emplace_back()`
     does return a bool)
 7) **Always use `iox::log::Logger`**, instead of `printf()`
-8) **Always use `iox::ErrorHandler()`**, when an error occurs that cannot or shall not be propagated via an
-    `iox::cxx::expected`, the `iox::ErrorHandler()` shall be used; exceptions are not allowed
+8) **Always use `iox::ErrorHandler()` or `cxx::Expects`/`cxx::Ensures`**, when an error occurs that cannot or shall not be propagated via an
```
elfenpiff, line 113: In my opinion this will be very soon outdated when the low level error handler becomes also available for iceoryx hoofs. Then it would make sense to never use `cxx::Expects` / `cxx::Ensures` (even when they use the error handler behind the scenes).
I would state here right from the beginning: always use the `iox::ErrorHandler` since we can emit an error code here which can be important for later bug reports or debugging.

iceoryx_binding_c/test/test.hpp
```diff
@@ -17,6 +17,8 @@
 #ifndef IOX_DDS_GATEWAY_TEST_TEST_HPP
 #define IOX_DDS_GATEWAY_TEST_TEST_HPP
 
+#include "iceoryx_hoofs/testing/mocks/error_handler_mock.hpp"
```
elfenpiff, line 20: Why do we actually require an error handler mock? We can always set a temporary error handler and check if it was called with the correct arguments?

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -17,20 +17,26 @@
 #ifndef IOX_HOOFS_ERROR_HANDLING_ERROR_HANDLING_HPP
 #define IOX_HOOFS_ERROR_HANDLING_ERROR_HANDLING_HPP
 
+/// @todo #1099 rename this file to error_handler.hpp
```
elfenpiff, line 20: This todo is not listed in the issue, please add it since it may comes with some bigger code refactoring.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -213,45 +206,18 @@ enum class ErrorLevel : uint32_t
...
-    static iox::HandlerFunction handler;
-    /// Needed, if you want to exchange the handler. Remember the old one and call it if it is not your error. The error
-    /// mock needs to be the last one exchanging the handler in tests.
-    static std::mutex handler_mutex;
-};
-
 /// @brief Howto use the error handler correctly
-///     1.) If the error you would like to handle is not listed in ICEORYX_ERRORS(error)\...
-///             macro just add them like:
+///     1.) Use the macro ICEORYX_ERRORS(error)\ to create the enum for your component and
```
elfenpiff, line 210: I think you can get rid of the slash `\` after `ICEORYX_ERRORS`

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -213,45 +206,18 @@ enum class ErrorLevel : uint32_t
...
+///             add new errors like:
 ///             error(MODULE_NAME__MY_FUNKY_ERROR)
 ///         Attention: Create an error after the following convention:
 ///             MODULE_NAME__A_CLEAR_BUT_SHORT_ERROR_DESCRIPTION
 ///         And a long name is alright!
 ///
-///     2.) Call errorHandler(Error::kMODULE_NAME__MY_FUNKY_ERROR);
+///     2.) Specialize the following methods for your NewEnumErrorType:
```
elfenpiff, line 217: This is new. Shouldn't we add this also to our error handling concept document?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant_queue.inl
```diff
@@ -17,8 +17,6 @@
 #ifndef IOX_HOOFS_CXX_VARIANT_QUEUE_INL
 #define IOX_HOOFS_CXX_VARIANT_QUEUE_INL
 
-#include "iceoryx_hoofs/error_handling/error_handling.hpp"
-
```
elfenpiff, line 21: I am sorry, please adjust the copyright year :/

iceoryx_hoofs/include/iceoryx_hoofs/internal/file_reader/file_reader.hpp
```diff
@@ -17,8 +17,6 @@
 #ifndef IOX_HOOFS_FILE_READER_FILE_READER_HPP
 #define IOX_HOOFS_FILE_READER_FILE_READER_HPP
 
-#include "iceoryx_hoofs/error_handling/error_handling.hpp"
```
elfenpiff, line 20: Copyright year

iceoryx_hoofs/source/error_handling/error_handling.cpp
```diff
@@ -17,43 +17,26 @@
 
 #include "iceoryx_hoofs/error_handling/error_handling.hpp"
 
-#include "iceoryx_hoofs/log/logger.hpp"
-#include "iceoryx_hoofs/log/logging.hpp"
-#include "iceoryx_hoofs/log/logmanager.hpp"
-
-#include <iostream>
-#include <sstream>
-
```
elfenpiff, line 26: Copyright year

iceoryx_hoofs/source/error_handling/error_handling.cpp
```diff
@@ -17,43 +17,26 @@
...
+const char* toString(const Error error) noexcept
 {
-    stream << ErrorHandler::toString(value);
-    return stream;
+    return ERROR_NAMES[static_cast<typename std::underlying_type<Error>::type>(error)];
 }
 
-void ErrorHandler::defaultHandler(const Error error,
-                                  const std::function<void()>& errorCallBack,
+void ErrorHandler::defaultHandler(const uint32_t error IOX_MAYBE_UNUSED,
```
elfenpiff, line 32: Actually the C++ best practice is here just to remove the variable name and write it like: `const uint32_t, const char * errorName`.
Then it is clear we ignore the argument and we do not have to use `IOX_MAYBE_UNUSED` to suppress compiler warnings.

iceoryx_hoofs/source/file_reader/file_reader.cpp
```diff
@@ -33,34 +35,31 @@ FileReader::FileReader(const std::string& f_fileName, const std::string& f_fileP
...
+                      << "Could not open file '" << m_file << "'."
+                      << "\033[0m" << std::endl;
+            return;
+        }
+        case ErrorMode::Terminate:
+        {
+            std::cerr << "\033[5;31m"
+                      << "Could not open file '" << m_file << "'. Exiting!"
+                      << "\033[0m" << std::endl;
+            cxx::Ensures(false);
```
elfenpiff, line 57: Since we will most likely use the error handler in the complete code base, why not use it here with a newly created enum to spare us some future work.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
```diff
@@ -113,11 +112,7 @@ void Timer::OsTimer::callbackHelper(sigval data) noexcept
             // prohibits other threads from entering the flag protected region
             handle.m_callbackIsAboutToBeExecuted.test_and_set(std::memory_order_acq_rel);
 
-            if (handle.m_timer == nullptr)
-            {
-                errorHandler(Error::kPOSIX_TIMER__INCONSISTENT_STATE);
-                return;
-            }
+            cxx::Expects((handle.m_timer != nullptr) && "Timer in inconsistent state");
```
elfenpiff, line 115: Please use the error handler here and in all remaining files where you added an `cxx::Expects` / `cxx::Ensures` or where you removed the `errorHandler`.

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
```diff
@@ -379,57 +378,44 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
...
-        [&](const iox::Error, const std::function<void()>, const iox::ErrorLevel) { hasTerminated = true; });
-
     Timer sut(TIMEOUT, [] { std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10)); });
-
     ASSERT_FALSE(sut.start(Timer::RunMode::PERIODIC, Timer::CatchUpPolicy::SKIP_TO_NEXT_BEAT).has_error());
-
     std::this_thread::sleep_for(std::chrono::milliseconds(TIMEOUT.toMilliseconds() * 10));
-    TIMING_TEST_EXPECT_FALSE(hasTerminated);
+    // EXPECT_NO_DEATH
+    EXPECT_TRUE(true);
```
elfenpiff, line 385: Please revert the changes (in whole file). `cxx::Expects` / `cxx::Ensures` are calling the error handler and terminate the program. When this test fails the whole test is terminated and the user is wondering what happened.

If possible, a unit test should never die! We always want a list of all the failing unit tests with the condition/code line which failed - this test setup makes this impossible since we do not know if the test failed or some segfault occurred.

Please do this by dropping the commit otherwise we clutter or history with changes which actually never happened.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
```
elfenpiff, line 1: Why do we need it and cannot use the `setTemporaryErrorHandler`?

Please drop this change from the history as well.

elfenpiff (changes requested):
Please squash or reorder and squash some commits and revert some changes by dropping some commits like:
1. Update copyright header, this really clutters our git history.
2. The last two one liner commits.

Please create one commit here instead of 4.

elBoberido:
@mossmaurice can't promise to look at this anytime soon since I'm quite busy with the request-response feature

CONTRIBUTING.md
mossmaurice: I don't think it makes sense to use `errorHandler()` in `iceoryx_hoofs` building blocks. IMHO it's more idiomatic and closer to a library philosophy to use `Expects` / `Ensures` and allow the users to call the `errorHandler()` inside `Expects` / `Ensures` (or `throw` or `std::terminate`).

However, we should not discuss this here, that's something for #1032

The goal for `v2.0` is to decouple `iceoryx_posh` and `iceoryx_binding_c` from `iceoryx_hoofs` as an intermediate step. Hence, the description above is still valid.

cc @MatthiasKillat

iceoryx_binding_c/test/test.hpp
mossmaurice: The `ErrorHandlerMock` contains call code which is only needed in test code. Now, `ErrorHandler` on the other hand, contains the code needed during runtime of an iceoryx system and is not cluttered with test-only code anymore. The templatization of `setTemporaryErrorHandler` is necessary to provide a user friendly way to use a different error enums for each module (posh, C binding).

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
mossmaurice: Good point, I'll add this on the follow-up PR to #1028, where the split into three different enums is actually happening.

iceoryx_hoofs/source/file_reader/file_reader.cpp
mossmaurice: I disagree with using the `errorHandler()` directly in `iceoryx_hoofs`. See [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/1100#discussion_r807013547). That's something for #1032. For `v2.0` I would like to implement our current error handling strategy as described in `doc/design/error-handling.md`.

iceoryx_hoofs/source/posix_wrapper/timer.cpp
mossmaurice: Rejected, quote from `doc/design/error-handling.md`:

`The error handler cannot be used in hoofs.`

See [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/1100#discussion_r807013547) for the background an my opinion. For `v2.0` I would like to implement our current error handling strategy.

I'm happy to discuss changes to the error handling strategy in #1032 for `v3.0`.

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
mossmaurice: You're right, `DEATH_TESTS` should be avoided. However, this is an intermediate step as the `posix::Timer` needs to be removed before `v2.0` e.g. by finishing off #754 or re-implementing it from scratch.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: Now, the `ErrorHandler` is free from any test code, see more in my [previous comment](https://github.com/eclipse-iceoryx/iceoryx/pull/1100#discussion_r807093601).

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,81 @@
...
+namespace iox
+{
+template <typename Error>
+using TypedHandlerFunction = std::function<void(const Error, const ErrorLevel)>;
+
+class ErrorHandlerMock : protected ErrorHandler
+{
+  public:
+    template <typename Error>
+    static cxx::GenericRAII setTemporaryErrorHandler(const TypedHandlerFunction<Error>& newHandler) noexcept;
```
elBoberido, line 32: Instead of having to call a function, why not make this the ctor?

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -17,20 +17,26 @@
...
+#include "iceoryx_hoofs/log/logger.hpp"
+#include "iceoryx_hoofs/log/logging.hpp"
+#include "iceoryx_hoofs/log/logmanager.hpp"
 
 #include <cassert>
 #include <functional>
 #include <iostream>
 #include <mutex>
+#include <sstream>
+
```
elBoberido, line 33: Are these new includes really necessary?

iceoryx_posh/test/moduletests/test_popo_sample.cpp
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
elBoberido, line 1: Please remove this file. This is a wrong merge conflict resolution

elBoberido (commented):
Looks good in general. I'm just not sure about the new death tests, since the make problems on macOS

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
mossmaurice: Removed, the uncessary ones.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: That would be possible. However, using `setTemporaryErrorHandler()` in tests is more meaningful for me. I use a mock to do set a certain behaviour.

iceoryx_hoofs/source/error_handling/error_handling.cpp
```diff
@@ -17,43 +17,24 @@
...
 {
-    stream << ErrorHandler::toString(value);
-    return stream;
+    return ERROR_NAMES[static_cast<typename std::underlying_type<Error>::type>(error)];
 }
 
-void ErrorHandler::defaultHandler(const Error error,
-                                  const std::function<void()>& errorCallBack,
-                                  const ErrorLevel level) noexcept
+void ErrorHandler::defaultHandler(const uint32_t, const char* errorName, const ErrorLevel level) noexcept
```
MatthiasKillat, line 32: Ok, the default handler does not care about the first parameter (I think it is basically the error code). Is this because it is already translated to `errorName`. The API may be a bit convoluted, but if so it already was before.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,82 @@
...
+};
+
+template <typename Error>
+cxx::optional<iox::TypedHandlerFunction<Error>> typedHandler;
+
+template <typename ErrorEnumType>
+inline void errorHandlerForTest(const uint32_t error, const char* errorName, const ErrorLevel level) noexcept
+{
+    uint32_t errorEnumType = error >> 16;
+    uint32_t expectedErrorEnumType =
```
MatthiasKillat, line 48: This is more like `noError` (and certainly not a type but a concrete value).

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,82 @@
...
+    static std::mutex handler_mutex;
+};
+
+template <typename Error>
+cxx::optional<iox::TypedHandlerFunction<Error>> typedHandler;
+
+template <typename ErrorEnumType>
+inline void errorHandlerForTest(const uint32_t error, const char* errorName, const ErrorLevel level) noexcept
+{
+    uint32_t errorEnumType = error >> 16;
```
MatthiasKillat, line 47: more like `errorCode`, certainly no type (I know it means probably type of error, but I think it is a misleading name). I would drop `Enum` in any case.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+template <typename Error>
+inline cxx::GenericRAII
+ErrorHandlerMock::setTemporaryErrorHandler(const TypedHandlerFunction<Error>& newHandler) noexcept
+{
+    return cxx::GenericRAII(
+        [&newHandler] {
+            std::lock_guard<std::mutex> lock(handler_mutex);
```
MatthiasKillat, line 71: snake case looks out of place but ok.

iceoryx_hoofs/testing/mocks/error_handler_mock.cpp
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "iceoryx_hoofs/testing/mocks/error_handler_mock.hpp"
+
+namespace iox
+{
+std::mutex ErrorHandlerMock::handler_mutex;
```
MatthiasKillat, line 21: Does this snake case break with our (local) conventions or is there more of a purpose?

I prefer snake case for certain library abstractions but not variables like this one, unless all variables in the module follow this convention.

But more importantly, it is a global variable in the iox namespace but is actually something specific to error handling. We should try to avoid polluting the iox namespace with this. If we cannot avoid the global mutex (can we confine it to some class as a static?) maybe it should live in something like `iox::error_handling::detail`.

The convention being that the any `detail` namespace is certainly not to be accessed by the user (even if it can be accessed to write a global variable like this one).

iceoryx_posh/test/moduletests/test_popo_sample.cpp
```diff
@@ -90,4 +90,4 @@ TEST_F(Sample_test, PublishingMovedSampleCallsErrorHandler)
     ASSERT_THAT(detectedError.value(), Eq(iox::Error::kPOSH__PUBLISHING_EMPTY_SAMPLE));
 }
 
-} // namespace
+} // namespace
```
MatthiasKillat, line 93: missing newline

MatthiasKillat (commented):
Ok in general (but we need to rethink some of the error handling in another issue). I also think ErrorHandlerMock is the wrong name, it is the `TestErrorHandler` or something like this.

What happened to `typed_unique_id.hpp`?

mossmaurice:
> Ok in general for the purpose of separating the error handling but we need to rethink some of the error handling in another issue. I also think ErrorHandlerMock is the wrong name, it is the `TestErrorHandler` or something like this.

Agree, that will happen as part of #1032. 

> What happened to `typed_unique_id.hpp`?

`typed_unique_id.hpp` was an empty left over file, which was forgotten in a previous PR.

iceoryx_binding_c/test/moduletests/test_c2cpp_enum_translation.cpp
```diff
@@ -58,8 +58,8 @@ TEST(c2cpp_enum_translation_test, SubscriberState)
 // the clang sanitizer detects this successfully and this leads to termination, and with this the test fails
 #if !defined(__clang__)
     iox::Error errorValue = iox::Error::kNO_ERROR;
-    auto errorHandlerGuard = iox::ErrorHandler::setTemporaryErrorHandler(
-        [&](const iox::Error e, const std::function<void()>, const iox::ErrorLevel) { errorValue = e; });
+    auto errorHandlerGuard = iox::ErrorHandlerMock::setTemporaryErrorHandler<iox::Error>(
```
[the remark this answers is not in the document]
mossmaurice: Ok, noted. Let's take out the naming discussion from this PR.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -280,9 +246,36 @@ class ErrorHandler
 /// errorHandler(Error::kTEST__ASSERT_CALLED);
 /// ASSERT_TRUE(called);
 /// @endcode
+/// @tparam[in] Error type which is used to report the error (typically an enum)
+template <typename Error>
 void errorHandler(const Error error,
                   const std::function<void()>& errorCallBack = std::function<void()>(),
                   const ErrorLevel level = ErrorLevel::FATAL) noexcept;
+
+using HandlerFunction = std::function<void(const uint32_t, const char*, const ErrorLevel)>;
```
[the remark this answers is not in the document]
mossmaurice: As this will be moved to test code with #1170, we're ok here to use `std::function`.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -280,9 +246,36 @@ class ErrorHandler
...
+
+/// @brief This handler is needed for unit testing, special debugging cases and
+///         other corner cases where we'd like to explicitly suppress the
+///         error handling.
+class ErrorHandler
```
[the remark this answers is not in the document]
mossmaurice: Let's take the naming discussion to a follow-up PR.

iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_expect.cpp
```diff
@@ -36,8 +37,8 @@ void ExpectDoesNotCallTerminateWhenObjectIsValid(const ExpectCall& callExpect)
     bool wasErrorHandlerCalled = false;
     SutType sut = FactoryType::createValidObject();
     {
-        auto handle =
-            iox::ErrorHandler::setTemporaryErrorHandler([&](auto, auto, auto) { wasErrorHandlerCalled = true; });
+        auto handle = iox::ErrorHandlerMock::setTemporaryErrorHandler<iox::Error>(
```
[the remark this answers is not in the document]
mossmaurice: I would like to make it explicit another class is used for testing. This is the reason why `ErrorHandler` would be ambiguous.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,81 @@
...
+using TypedHandlerFunction = std::function<void(const Error, const ErrorLevel)>;
+
+class ErrorHandlerMock : protected ErrorHandler
+{
+  public:
+    template <typename Error>
+    static cxx::GenericRAII setTemporaryErrorHandler(const TypedHandlerFunction<Error>& newHandler) noexcept;
+
+  protected:
+    /// Needed, if you want to exchange the handler. Remember the old one and call it if it is not your error. The error
```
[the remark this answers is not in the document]
mossmaurice: Re-phrased, the old comment.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,81 @@
...
+    /// mock needs to be the last one exchanging the handler in tests.
+    static std::mutex handler_mutex;
```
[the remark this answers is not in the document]
mossmaurice: Fixed

iceoryx_hoofs/source/error_handling/error_handling.cpp
mossmaurice: True, this can probably optimised. Same for:

```
void errorHandler(const Error error,
                  const std::function<void()>& errorCallBack = std::function<void()>(),
                  const ErrorLevel level = ErrorLevel::FATAL) noexcept;
```

The 2nd and 3rd parameter will be swapped. Not on this PR.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: Well, the concret values will map to type (`CBindingError`, `PoshError`, `HoofsError`), see #1170

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: I would appreciate to leave the naming as it is, if you're ok with it. Reason is that this makes it easier to understand and/or remember what is happening here and how the type of error enum is detected.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,82 @@
...
+    uint32_t errorEnumType = error >> 16;
+    uint32_t expectedErrorEnumType =
+        static_cast<typename std::underlying_type<ErrorEnumType>::type>(ErrorEnumType::kNO_ERROR) >> 16;
+
+    if (errorEnumType == expectedErrorEnumType)
+    {
+        // We undo the type erasure
+        auto typedError = static_cast<ErrorEnumType>(error);
+        typedHandler<iox::Error>.and_then(
+            [&](TypedHandlerFunction<Error> storedHandler) { storedHandler(typedError, level); });
```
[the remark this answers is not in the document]
mossmaurice: Yes, the handler is stored inside `typedHandler`. If the `cxx::optional` contains a value, this handler is invoked. It is not possible to encounter a dangling reference, as the reseting is done via a `cxx::GenericRAII`. Misuse is therefore very low.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: Fixed.

iceoryx_hoofs/testing/mocks/error_handler_mock.cpp
mossmaurice: Fixed the snake case naming. We can think about wrapping parts of the `error_handling` in objects in an extra namespace. For me this not an important topic, as the class is already separate in a testing package. Hence, not a big deal for me.

iceoryx_posh/test/moduletests/test_popo_sample.cpp
mossmaurice: fixed

iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_expect.cpp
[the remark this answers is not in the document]
MatthiasKillat: Yes, so `TestErrorhandler` maybe. But I am fine with changing this in another refactoring of the error handling.

MatthiasKillat (approved):

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
```diff
@@ -242,45 +230,18 @@ enum class ErrorLevel : uint32_t
...
+///     1.) Use the macro ICEORYX_ERRORS(error) to create the enum for your component and
+///             add new errors like:
 ///             error(MODULE_NAME__MY_FUNKY_ERROR)
 ///         Attention: Create an error after the following convention:
 ///             MODULE_NAME__A_CLEAR_BUT_SHORT_ERROR_DESCRIPTION
 ///         And a long name is alright!
 ///
-///     2.) Call errorHandler(Error::kMODULE_NAME__MY_FUNKY_ERROR);
+///     2.) Specialize the following methods for your NewEnumErrorType:
+///         - const char* toString(const NewEnumErrorType error)
```
elBoberido, line 242: How about using #1066 for this?

iceoryx_hoofs/test/moduletests/test_posix_timer.cpp
```diff
@@ -379,57 +378,44 @@ TEST_F(Timer_test, GetOverrunsFailsWithNoCallback)
 }
 
 TIMING_TEST_F(Timer_test, CatchUpPolicySkipToNextBeatContinuesWhenCallbackIsLongerThenTriggerTime, Repeat(5), [&] {
```
elBoberido, line 380: I think it's time to just delete the posix timer

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
elBoberido: For me it is the oposite. It's like with the `std::lock`. One creates the object and let it's dtor do everything.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
elBoberido: Shouldn't this just be `0` with a comment why it currently 0 and what will change?

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
elBoberido: How about `errorDomain`/`errorEnumDomain`?

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
```diff
@@ -0,0 +1,80 @@
...
+        [&newHandler] {
+            std::lock_guard<std::mutex> lock(m_handlerMutex);
+            typedHandler<iox::Error>.emplace(newHandler);
+            handler = errorHandlerForTest<Error>;
+        },
+        [] {
+            std::lock_guard<std::mutex> lock(m_handlerMutex);
+            typedHandler<iox::Error>.reset();
+            handler = defaultHandler;
+        });
```
elBoberido, line 77: I know that this is not new but just the current code moved to a new file, but this is broken and we need to change it in the future. It is not only the change of the handler function but also the access which must be guarded by a mutex. Maybe we can somehow circumvent this by allowing only once to change the handler but that is a bigger refactoring.

elBoberido (commented):
Can you please remove the empty `typed_unique_id.hpp`

mossmaurice:
> Can you please remove the empty `typed_unique_id.hpp`

@elBoberido It's already remove, maybe just harder to catch in GitHub web GUI due to being empty.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: If you don't mind I would like to leave it that whay, let's call it a forward implementation :upside_down_face:

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/mocks/error_handler_mock.hpp
mossmaurice: When I heard domain, in this context I think of a tld like `bar.com`. Happy to discuss the naming, but for now I would appreciate if we could merge it as it is.

iceoryx_hoofs/include/iceoryx_hoofs/error_handling/error_handling.hpp
mossmaurice: Added to #1099

Issue #1074, opened 2022-02-08
Define limits for service discovery in configuration


MatthiasKillat:
## Brief feature description

We need to define constants for service discovery in a configuration, e.g.
- number of supported services
- number of services per query

The limits must be set appropriately. Due to the static memory model there are severe constraints on some platforms (due to stack size etc.)

## Detailed information

Note: list of required constants is incomplete and requires consideration

Some existing constants have ambiguous names and must be reconsidered.

PR #1115, opened 2022-02-16
Iox #27 add server support to service registry

MatthiasKillat:

1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Closes Nothing

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -127,6 +127,9 @@ class ServiceRegistry
     void getAll(ServiceDescriptionVector_t& searchResult) const noexcept;
 
     void applyToAll(cxx::function_ref<void(const ServiceDescriptionEntry&)> callable) const noexcept;
+
+    cxx::expected<Error> add(const capro::ServiceDescription& serviceDescription,
+                             ReferenceCounter_t ServiceDescriptionEntry::*count);
 };
 } // namespace roudi
 } // namespace iox
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -46,8 +46,13 @@ class ServiceRegistry
     {
         ServiceDescriptionEntry(const capro::ServiceDescription& serviceDescription);
 
-        capro::ServiceDescription serviceDescription{};
-        ReferenceCounter_t count{1U};
+        capro::ServiceDescription serviceDescription;
+
+        // it can be discussed whether the type of service (server or publisher)
+        // should be part of the ServiceDescription but this would lead to more
+        // exchanges (e.g. comparison semantics)
+        ReferenceCounter_t publisherCount{0U};
+        ReferenceCounter_t serverCount{0U};
     };
 
     /// @todo #415 #1074 set limits properly and define location for the limits,
@@ -65,6 +70,16 @@ class ServiceRegistry
     ///        in case of multiple occurrences only one occurrence is removed
     /// @param[in] serviceDescription, service to be removed
     void removePublisher(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    /// @brief Adds a given server service description to registry
+    /// @param[in] serviceDescription, service to be added
+    /// @return ServiceRegistryError, error wrapped in cxx::expected
+    cxx::expected<Error> addServer(const capro::ServiceDescription& serviceDescription) noexcept;
+
+    /// @brief Removes a given server service description from registry if service is found,
+    ///        in case of multiple occurrences only one occurrence is removed
+    /// @param[in] serviceDescription, service to be removed
+    void removeServer(const capro::ServiceDescription& serviceDescription) noexcept;
 
     /// @brief Removes given service description from registry if service is found,
     ///        all occurences are removed
--- a/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
@@ -56,15 +56,15 @@ class ServiceRegistry
 
     using ServiceDescriptionVector_t = cxx::vector<ServiceDescriptionEntry, MAX_SERVICE_DESCRIPTIONS>;
 
-    /// @brief Adds given service description to registry
+    /// @brief Adds a given publisher service description to registry
     /// @param[in] serviceDescription, service to be added
     /// @return ServiceRegistryError, error wrapped in cxx::expected
-    cxx::expected<Error> add(const capro::ServiceDescription& serviceDescription) noexcept;
+    cxx::expected<Error> addPublisher(const capro::ServiceDescription& serviceDescription) noexcept;
 
-    /// @brief Removes given service description from registry if service is found,
+    /// @brief Removes a given publisher service description from registry if service is found,
     ///        in case of multiple occurrences only one occurrence is removed
     /// @param[in] serviceDescription, service to be removed
-    void remove(const capro::ServiceDescription& serviceDescription) noexcept;
+    void removePublisher(const capro::ServiceDescription& serviceDescription) noexcept;
 
     /// @brief Removes given service description from registry if service is found,
     ///        all occurences are removed
--- a/iceoryx_posh/source/roudi/port_manager.cpp
+++ b/iceoryx_posh/source/roudi/port_manager.cpp
@@ -687,7 +687,7 @@ void PortManager::publishServiceRegistry() const noexcept
 
 void PortManager::addEntryToServiceRegistry(const capro::ServiceDescription& service) noexcept
 {
-    m_serviceRegistry.add(service).or_else([&](auto&) {
+    m_serviceRegistry.addPublisher(service).or_else([&](auto&) {
         LogWarn() << "Could not add service " << service.getServiceIDString() << " to service registry!";
         errorHandler(Error::kPOSH__PORT_MANAGER_COULD_NOT_ADD_SERVICE_TO_REGISTRY, nullptr, ErrorLevel::MODERATE);
     });
@@ -696,7 +696,7 @@ void PortManager::addEntryToServiceRegistry(const capro::ServiceDescription& ser
 
 void PortManager::removeEntryFromServiceRegistry(const capro::ServiceDescription& service) noexcept
 {
-    m_serviceRegistry.remove(service);
+    m_serviceRegistry.removePublisher(service);
     publishServiceRegistry();
 }
 
--- a/iceoryx_posh/source/roudi/service_registry.cpp
+++ b/iceoryx_posh/source/roudi/service_registry.cpp
@@ -163,7 +163,6 @@ void ServiceRegistry::removeServer(const capro::ServiceDescription& serviceDescr
         {
             if (entry->serverCount >= 1U)
             {
-                entry->serverCount--;
                 if (--entry->serverCount >= 1U)
                 {
                     return;
--- a/iceoryx_posh/source/roudi/service_registry.cpp
+++ b/iceoryx_posh/source/roudi/service_registry.cpp
@@ -26,17 +26,17 @@ ServiceRegistry::ServiceDescriptionEntry::ServiceDescriptionEntry(const capro::S
 {
 }
 
-cxx::expected<ServiceRegistry::Error>
-ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescription) noexcept
+cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceDescription& serviceDescription,
+                                                           ReferenceCounter_t ServiceDescriptionEntry::*count)
 {
     auto index = findIndex(serviceDescription);
     if (index != NO_INDEX)
     {
-        // multiple publishers with the same service descripion are possible
+        // multiple entries with the same service descripion are possible
         // and we just increase the count in this case (multi-set semantics)
         // entry exists, increment counter
         auto& entry = m_serviceDescriptions[index];
-        entry->publisherCount++;
+        ((*entry).*count)++;
         return cxx::success<>();
     }
 
@@ -48,7 +48,7 @@ ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescriptio
     {
         auto& entry = m_serviceDescriptions[m_freeIndex];
         entry.emplace(serviceDescription);
-        entry->publisherCount = 1U;
+        (*entry).*count = 1U;
         m_freeIndex = NO_INDEX;
         return cxx::success<>();
     }
@@ -59,7 +59,7 @@ ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescriptio
         if (!entry)
         {
             entry.emplace(serviceDescription);
-            entry->publisherCount = 1U;
+            (*entry).*count = 1U;
             return cxx::success<>();
         }
     }
@@ -69,7 +69,7 @@ ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescriptio
     {
         auto& entry = m_serviceDescriptions.back();
         entry.emplace(serviceDescription);
-        entry->publisherCount = 1U;
+        (*entry).*count = 1U;
         return cxx::success<>();
     }
 
@@ -77,53 +77,15 @@ ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescriptio
 }
 
 cxx::expected<ServiceRegistry::Error>
+ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescription) noexcept
+{
+    return add(serviceDescription, &ServiceDescriptionEntry::publisherCount);
+}
+
+cxx::expected<ServiceRegistry::Error>
 ServiceRegistry::addServer(const capro::ServiceDescription& serviceDescription) noexcept
 {
-    auto index = findIndex(serviceDescription);
-    if (index != NO_INDEX)
-    {
-        // multiple publishers with the same service descripion are possible
-        // and we just increase the count in this case (multi-set semantics)
-        // entry exists, increment counter
-        auto& entry = m_serviceDescriptions[index];
-        entry->serverCount++;
-        return cxx::success<>();
-    }
-
-    // entry does not exist, find a free slot if it exists
-
-    // fast path to a free slot (which was occupied by previously removed entry),
-    // prefer to fill entries close to the front
-    if (m_freeIndex != NO_INDEX)
-    {
-        auto& entry = m_serviceDescriptions[m_freeIndex];
-        entry.emplace(serviceDescription);
-        entry->serverCount = 1U;
-        m_freeIndex = NO_INDEX;
-        return cxx::success<>();
-    }
-
-    // search from start
-    for (auto& entry : m_serviceDescriptions)
-    {
-        if (!entry)
-        {
-            entry.emplace(serviceDescription);
-            entry->serverCount = 1U;
-            return cxx::success<>();
-        }
-    }
-
-    // append new entry at the end (the size only grows up to capacity)
-    if (m_serviceDescriptions.emplace_back())
-    {
-        auto& entry = m_serviceDescriptions.back();
-        entry.emplace(serviceDescription);
-        entry->serverCount = 1U;
-        return cxx::success<>();
-    }
-
-    return cxx::error<Error>(Error::SERVICE_REGISTRY_FULL);
+    return add(serviceDescription, &ServiceDescriptionEntry::serverCount);
 }
 
 void ServiceRegistry::removePublisher(const capro::ServiceDescription& serviceDescription) noexcept
--- a/iceoryx_posh/source/roudi/service_registry.cpp
+++ b/iceoryx_posh/source/roudi/service_registry.cpp
@@ -26,7 +26,8 @@ ServiceRegistry::ServiceDescriptionEntry::ServiceDescriptionEntry(const capro::S
 {
 }
 
-cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceDescription& serviceDescription) noexcept
+cxx::expected<ServiceRegistry::Error>
+ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescription) noexcept
 {
     auto index = findIndex(serviceDescription);
     if (index != NO_INDEX)
@@ -72,7 +73,7 @@ cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceD
     return cxx::error<Error>(Error::SERVICE_REGISTRY_FULL);
 }
 
-void ServiceRegistry::remove(const capro::ServiceDescription& serviceDescription) noexcept
+void ServiceRegistry::removePublisher(const capro::ServiceDescription& serviceDescription) noexcept
 {
     auto index = findIndex(serviceDescription);
     if (index != NO_INDEX)
--- a/iceoryx_posh/source/roudi/service_registry.cpp
+++ b/iceoryx_posh/source/roudi/service_registry.cpp
@@ -36,7 +36,7 @@ ServiceRegistry::addPublisher(const capro::ServiceDescription& serviceDescriptio
         // and we just increase the count in this case (multi-set semantics)
         // entry exists, increment counter
         auto& entry = m_serviceDescriptions[index];
-        entry->count++;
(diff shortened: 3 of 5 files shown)


MatthiasKillat:
@elBoberido Since you are the one who needs this functionality feel free to review. If you do not have the time no problem, but maybe check the API change and whether it works for you.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -26,16 +26,17 @@ ServiceRegistry::ServiceDescriptionEntry::ServiceDescriptionEntry(const capro::S
 {
 }
 
-cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceDescription& serviceDescription) noexcept
+cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceDescription& serviceDescription,
+                                                           ReferenceCounter_t ServiceDescriptionEntry::*count)
```
elfenpiff, line 30: Could you please a reference instead of a pointer.

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
```diff
@@ -46,8 +46,13 @@ class ServiceRegistry
     {
         ServiceDescriptionEntry(const capro::ServiceDescription& serviceDescription);
 
-        capro::ServiceDescription serviceDescription{};
-        ReferenceCounter_t count{1U};
+        capro::ServiceDescription serviceDescription;
+
+        // it can be discussed whether the type of service (server or publisher)
```
mossmaurice, line 51: Should this comment not be in some issue instead of in the code?

iceoryx_posh/source/runtime/service_discovery.cpp
```diff
@@ -35,7 +35,10 @@ ServiceContainer ServiceDiscovery::findService(const cxx::optional<capro::IdStri
     ServiceContainer searchResult;
     for (auto& service : tempSearchResult)
     {
-        searchResult.push_back(service.serviceDescription);
+        if (service.publisherCount > 0)
```
mossmaurice, line 38: Why is the `serverCount` not checked here? How can I find a `ServerPort`?

MatthiasKillat: You cannot for now, I talked without @elBoberido  and we need to figure out the best way to make this available in the discovery. For now it is just to not break existing functionality.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: It is a pointer to member.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -29,7 +29,96 @@ namespace
...
+using SearchResult_t = iox::roudi::ServiceRegistry::ServiceDescriptionVector_t;
 
+struct PublisherTest
+{
+    auto add(const ServiceDescription& sd)
+    {
+        return registry.addPublisher(sd);
+    }
+
+    auto otherAdd(const ServiceDescription& sd)
```
mossmaurice, line 42: What about `complementaryAdd()`?

iceoryx_posh/include/iceoryx_posh/internal/roudi/service_registry.hpp
MatthiasKillat: I wanted to ad the rationale behind not encoding the information into the ServiceDescription (discussed it with @elBoberido ) and deemed it relevant in the code.

There is also some uncertainty about details how to handle the publisher/server distinction and there are differences that are relevant for the registry (e.g. there can only be one server of a specific description but it can be the same description as some publisher etc.).

Can be removed though, i am indifferent here.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -73,216 +168,195 @@ TEST_F(ServiceRegistry_test, AddMaximumNumberOfServiceDescriptionsWorks)
...
     for (auto& service : services)
     {
-        auto result = sut.add(service);
+        auto result = this->sut.add(service);
         ASSERT_FALSE(result.has_error());
     }
 }
 
-TEST_F(ServiceRegistry_test, AddMoreThanMaximumNumberOfServiceDescriptionsFails)
+TYPED_TEST(ServiceRegistry_test, AddMoreThanMaximumNumberOfServiceDescriptionsFails)
```
FerdinandSpitzschnueffler, line 176: I guess the correct test name is `AddServiceDescriptionsWhichWasAlreadyAddedAndReturnsOneResult`.

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -66,28 +69,73 @@ cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceD
...
+    return add(serviceDescription, &ServiceDescriptionEntry::publisherCount);
+}
+
+cxx::expected<ServiceRegistry::Error>
+ServiceRegistry::addServer(const capro::ServiceDescription& serviceDescription) noexcept
+{
+    return add(serviceDescription, &ServiceDescriptionEntry::serverCount);
+}
+
+void ServiceRegistry::removePublisher(const capro::ServiceDescription& serviceDescription) noexcept
```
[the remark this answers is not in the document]
MatthiasKillat: This is possible and the exact semantics for the discovery are not yet known. That is why for the discovery we essentially project the result onto publishers only, to be consistent with the previous functionality. I discussed it with @elBoberido.

I will discuss with @elBoberido  how to handle servers in the `ServiceDiscovery` later, but this PR is concerned with the `ServiceRegistry` only. There will be a separate PR for this once the semantics are clear.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
MatthiasKillat: Would that gain much? It is not more expressive in my book. It is also an internal artifact to avoid code duplication, no public API.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -73,216 +168,195 @@ TEST_F(ServiceRegistry_test, AddMaximumNumberOfServiceDescriptionsWorks)
...
-    for (auto& service : services)
-    {
-        auto result = sut.add(service);
-        ASSERT_FALSE(result.has_error());
-    }
-
-    auto result = sut.add(iox::capro::ServiceDescription("Foo", "Bar", "Baz"));
-    ASSERT_TRUE(result.has_error());
-    EXPECT_THAT(result.get_error(), Eq(ServiceRegistry::Error::SERVICE_REGISTRY_FULL));
-}
```
FerdinandSpitzschnueffler, line 101: Why is this test removed?

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
elBoberido: @mossmaurice this is mostly about adding internal API so that I can continue on the Request-Response feature. Currently it is not decided how to communicate this to the user. It is only tracked in the `ServiceRegistry` and `findService` would return only `ServiceDescriptions` for entries from `Publisher`. So we remain at status quo.

@MatthiasKillat had the idea to make this information part of `ServiceDescription`. Initially I was not too fond of this idea and MaKi also feared that it would be too invasive to change the `ServiceDescription` so close to the code freeze since it might result in many changes. After I took counsel with my pillow I had an idea how we could have it in the `ServiceDescription` and do not need to change much of our code. The additional benefit would be that the current `findService` API could remain as is and also be extended in a source compatible way to find only `Publisher` or `Server`.

Let's assume we add an additional `messagingPattern` method to the `ServiceDescription`
```cpp
enum class MessagingPattern {
ANY,
PUB_SUB,
REQ_RES
};
```
With the current `findService` API the `ServiceDiscovery` would assume that the user wants `MessagingPattern::ANY` and the user needs to check the returned `ServiceDescription` with the `MessagingPattern ServiceDescription::messagingPattern()` method. The `findService` API could be extended with a fourth parameter with a default value like `MessagingPattern messagingPattern = MessagingPattern::ANY`. This means that if there is a Publisher with `{"Foo", "Bar", "Baz"}` and a Server with `{"Foo", "Bar", "Baz"}` the user would get two results for `ANY` and one result for either `PUB_SUB` or `REQ_RES`.

To be backward compatible and enable the user to reuse a `ServiceDescription` for `Publisher` and `Server`, the `ServiceDescription` will by default be created with `MessagingPattern::ANY`
```cpp
auto sd = ServiceDescription{"Foo", "Bar", "Baz"};
auto publisher = Publisher{sd, {}};
auto server = Server{sd, {}};
```
This make it possible to create method/field combinations for ara::com where a method can be used to update a field.

I'm not sure if the user should be able to set the `MessagingPattern` of the `ServiceDescription` by herself and would therefore initially only allow it for the `ServiceDiscovery` which would do it when the result is returned.

Even if we do not want to proceed with the suggestion of MaKi and me, this PR does not prevent us from going a different route and adding a second counter is currently the least invasive way to add this to the `ServiceRegistry`. Another option would be to have a second vector but that wouldn't make things easier and just increase the memory usage.

The user side has not to be decided now and we can have a more detailed discussion in person, together with MaKi and Spitzi.

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
```diff
@@ -133,9 +133,8 @@ constexpr uint32_t MAX_PROCESS_NUMBER = 300U;
 /// Maximum number of services, which can be found.
 /// This limitation is coming due to the fixed capacity of the cxx::vector (This doesn't limit the offered number of
 /// instances)
-/// @todo #415 increase number back to 50 once service registry is available via shared memory
 /// @todo #1074 define and set the limits concerning the service discovery
-constexpr uint32_t MAX_NUMBER_OF_SERVICES = 10U;
+constexpr uint32_t MAX_NUMBER_OF_SERVICES = 100U;
```
elBoberido, line 137: Most probably we want something like `MAX_PUBLISHER + MAX_SERVER`. Maybe even `MAX_PUBLISHER * 0.8 + MAX_SERVER` or something like this if we assume there will be multi-publisher which share the same `ServiceDiscovery`. But that can be decided later on.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -619,6 +622,41 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
...
 }
 
+cxx::expected<PublisherPortRouDiType::MemberType_t*, PortPoolError>
+PortManager::acquirePublisherPortDataWithoutDiscovery(const capro::ServiceDescription& service,
+                                                      const popo::PublisherOptions& publisherOptions,
+                                                      const RuntimeName_t& runtimeName,
+                                                      mepoo::MemoryManager* const payloadDataSegmentMemoryManager,
+                                                      const PortConfigInfo& portConfigInfo) noexcept
+{
+    if (doesViolateCommunicationPolicy<iox::build::CommunicationPolicy>(service).and_then(
```
elBoberido, line 632: Seeing this, I think when a port is requested, RouDi should check whether the `ServiceDescription` matches one of the internal services like introspection or service discovery and deny the port creation if it is an internal service. Currently a user could create a publisher with the `ServiceDescription` of the service discovery and confuse all users of it.

Could you please create an github issues for this?

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
MatthiasKillat: @elBoberido I agree. But this should not affect the current pull request as it is not at odds with your proposal. I also want to store it as memory efficient as (reasonably, i.e. no compression) possible to keep the registry small (unless we gain something like search speed).

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -26,16 +26,17 @@ ServiceRegistry::ServiceDescriptionEntry::ServiceDescriptionEntry(const capro::S
...
     auto index = findIndex(serviceDescription);
     if (index != NO_INDEX)
     {
-        // multiple publishers with the same service descripion are possible
+        // multiple entries with the same service descripion are possible
         // and we just increase the count in this case (multi-set semantics)
         // entry exists, increment counter
         auto& entry = m_serviceDescriptions[index];
-        entry->count++;
+        ((*entry).*count)++;
```
elBoberido, line 39: Oh boy, seeing this in action I'm not sure anymore it was a good idea to give you the hint with the member pointer 😅

Nevertheless, let's keep it

iceoryx_posh/source/roudi/service_registry.cpp
```diff
@@ -66,28 +69,73 @@ cxx::expected<ServiceRegistry::Error> ServiceRegistry::add(const capro::ServiceD
...
     {
         auto& entry = m_serviceDescriptions[index];
 
-        if (entry->count > 1)
+        if (entry)
         {
-            entry->count--;
+            if (entry->publisherCount >= 1U)
+            {
+                if (--entry->publisherCount >= 1U)
```
elBoberido, line 102: Let's see if we can keep this after our certification tools run through the codebase :)

What do you think of
```cpp
if(--entry->publisherCount == 0U && entry->serverCount == 0)
{
    entry.reset();
    // reuse the slot in the next insertion
    m_freeIndex = index;
}
```

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -566,55 +666,82 @@ TEST_F(ServiceRegistry_test, SearchInFullRegistryWorks)
...
+    ::testing::Test::RecordProperty("TEST_ID", "3e8a9647-69cc-4cad-afb1-9188927aff04");
+    iox::capro::ServiceDescription service1("a", "a", "a");
+    iox::capro::ServiceDescription service2("a", "b", "b");
+    iox::capro::ServiceDescription service3("a", "a", "c");
+    iox::capro::ServiceDescription service4("a", "a", "d");
+
+    ASSERT_FALSE(this->sut.add(service1).has_error());
+    ASSERT_FALSE(this->sut.otherAdd(service2).has_error());
+    ASSERT_FALSE(this->sut.add(service3).has_error());
+    ASSERT_FALSE(this->sut.otherAdd(service4).has_error());
```
elBoberido, line 731: Could you please make number of calls to `add` and `otherAdd` distinct?

iceoryx_posh/source/roudi/port_manager.cpp
elfenpiff: Hehehe but then we close the door for iceoryx spoofing attacks ;)

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -619,6 +622,41 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
     return maybePublisherPortData;
 }
 
+cxx::expected<PublisherPortRouDiType::MemberType_t*, PortPoolError>
+PortManager::acquirePublisherPortDataWithoutDiscovery(const capro::ServiceDescription& service,
```
elfenpiff, line 626: The function `acquirePublisherPortData` is identical to this one, except the one line with the discovery. Could you please remove the whole content of that function and replace it with:
```cpp
return acquirePublisherPortDataWithoutDiscovery(service,
                                                     publisherOptions,
                                                     runtimeName,
                                                     payloadDataSegmentMemoryManager,
                                                     portConfigInfo)
                                                     .and_then([](auto publisherPortData){ doDiscoveryForPublisherPort(PublisherPortRouDiType(publisherPortData)); });
```

elfenpiff (changes requested):

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: There is no easier syntax for it. You may be able to get rid of one pair of braces but then it looks even more arcane.

elfenpiff: Yeah, why cant we use a normal pointer instead?

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: It is not equivalent as in theory you can call it with `entry->publisherCount==0`. In this case decrementing would lead to underflow and would mess everything up.

MatthiasKillat: I am indifferent but may change it (do not think in machine code it makes any difference). Regarding certification I do not know, code looks reasonable to me (otherwise I would not have it written like this...).

iceoryx_posh/source/roudi/service_registry.cpp
elBoberido: @elfenpiff the problem is this
```cpp
void doSomethingWithFooOnBar(Foo& foo, pointerToSomeMemberOfBarWhichNeedsToBeFound)
{
    for(auto& bar: allBars) {
        if(bar == foo.someCondition) {
            bar.pointerToSomeMemberOfBarWhichNeedsToBeFound = 42;
        }
    }
}
```

iceoryx_posh/include/iceoryx_posh/iceoryx_posh_types.hpp
MatthiasKillat: Yes we have issue https://github.com/eclipse-iceoryx/iceoryx/issues/1074 for this.

iceoryx_posh/source/roudi/port_manager.cpp
MatthiasKillat: Yes, essentially we blacklist some service descriptions which are internally used.

iceoryx_posh/source/roudi/service_registry.cpp
MatthiasKillat: I can also have all conditions in one `if` (short-circuiting) but I am not sure that is nice to read and understand. :-)

iceoryx_posh/source/roudi/port_manager.cpp
MatthiasKillat: Yes, but needs slight adaptation due to `doDiscoveryForPublisherPort` requiring an lvalue.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -91,198 +186,199 @@ TEST_F(ServiceRegistry_test, AddMoreThanMaximumNumberOfServiceDescriptionsFails)
...
     ASSERT_TRUE(result.has_error());
     EXPECT_THAT(result.get_error(), Eq(ServiceRegistry::Error::SERVICE_REGISTRY_FULL));
 }
 
-TEST_F(ServiceRegistry_test, AddServiceDescriptionsWhichWasAlreadyAddedAndReturnsOneResult)
+TYPED_TEST(ServiceRegistry_test, AddServiceDescriptionsWhichWasAlreadyAddedAndReturnsOneResult)
 {
-    ::testing::Test::RecordProperty("TEST_ID", "d8f61eb2-d082-4c26-9970-461427c3d200");
-    auto result1 = sut.add(ServiceDescription("Li", "La", "Launebaer"));
+    ::testing::Test::RecordProperty("TEST_ID", "a911f654-8314-4ea3-b9b2-1afa121a2b21");
```
FerdinandSpitzschnueffler, line 200: You copied the test id from the test above.

elBoberido (dismissed):
Looks good to me. Does @mossmaurice still have some concerns?

FerdinandSpitzschnueffler (changes requested):
Please change the mentioned test id.

elBoberido:
@MatthiasKillat I should not approve PRs without checking the reviewer checklist first. Pleas add the issue number to your last two commits

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
elBoberido: We need some kind of automation for this. It is impossible to the reviewer to catch such stuff. A new test could be based on any test from the file and one would need to manually check if the id is unique. What makes things even worse, one could even copy a test from a different file

MatthiasKillat: Agreed.

FerdinandSpitzschnueffler (approved):

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -587,6 +590,21 @@ PortManager::acquirePublisherPortData(const capro::ServiceDescription& service,
                                       const RuntimeName_t& runtimeName,
                                       mepoo::MemoryManager* const payloadDataSegmentMemoryManager,
                                       const PortConfigInfo& portConfigInfo) noexcept
+{
+    return acquirePublisherPortDataWithoutDiscovery(
+               service, publisherOptions, runtimeName, payloadDataSegmentMemoryManager, portConfigInfo)
+        .and_then([&](auto publisherPortData) {
+            PublisherPortRouDiType port(publisherPortData);
+            doDiscoveryForPublisherPort(port);
```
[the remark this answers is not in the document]
MatthiasKillat: Oversight, merged already, will fix in next PR.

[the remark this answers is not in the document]
MatthiasKillat: Fixed in https://github.com/eclipse-iceoryx/iceoryx/pull/1122

iceoryx_posh/source/roudi/service_registry.cpp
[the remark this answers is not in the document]
elfenpiff: @mossmaurice I think we need to define the terms first. Look at ROS2 since we are having here the same terminology:
* Topic: Publish/Subscribe communication: https://docs.ros.org/en/foxy/Tutorials/Topics/Understanding-ROS2-Topics.html
* Service: Request/Response communication: https://docs.ros.org/en/foxy/Tutorials/Services/Understanding-ROS2-Services.html

With this in mind the phone-book analogy does not fit since these are two different things.

Our service registry mixes both Topics/Services and we have to distinguish between them, either we do this by adding a `MessagingPattern` or we have a `TopicRegistry` and a `ServiceRegistry`. The naming is also not perfect in our case.


Just for fun, what do you think happens when we do not distinguish those things, then we will receive bug reports like:
Oh I found service XYZ but the subscriber/client cannot connect to it. To solve this impediment we then have to communicate to the user what type of service is behind XYZ?
For me the best approach would be by introducing `MessagingPattern`.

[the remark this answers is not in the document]
mossmaurice: @elfenpiff Good point! I had a look at the `rmw` API last week. There are several `findService`-like functions available:

* `rmw_get_topic_names_and_types()` => `ANY`
* `rmw_get_service_names_and_types_by_node()` => `REQ_RES`
* `rmw_get_publisher_names_and_types_by_node()` => `PUB_SUB`

There are also two methods to get more info like QoS:

* `rmw_get_publishers_info_by_topic()`
* `rmw_get_subscriptions_info_by_topic()`

My latest view on how things should work in `v3.0` is:

We differentiate between a user `ServiceDescription` and an internal representation of `ServiceDescription` enriched with more meta information. The information flow would be the following:

1. User creates a `ServiceDescription service`, which only describes the data
2. User creates a `popo::Publisher(service, publisherOptions)`
3. RouDi combines the `ServiceDescription`, QoS and the info what kind of producer was used (`popo::Publisher` or `popo::ServerPort`) into an internal representation and stores it in the `ServiceRegistry`
4. User does a `ServiceDiscovery::findService` and the internal presentation is converted back to `ServiceDescription`
5. To get the QoS info we could a similar method like in ROS 2 e.g. `ServiceDiscovery::getOptionsForProducer(ServiceDescription& service)`

What do you think?

I've implemented the possibilty to search for different producer kinds on #1134

cc @budrus @MatthiasKillat @elBoberido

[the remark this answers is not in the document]
elBoberido: @mossmaurice @elfenpiff could we please move this discussion to an issue or a github discussion? I just got notified about this and had to search for the right thread to find about it and I only did this because I knew that it must be in a closed thread. I'm not sure all find this, even when mentioned. At lease I would not have looked at the closed threads if I didn't knew it would be there.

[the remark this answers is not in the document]
mossmaurice: Created the [discussion](https://github.com/eclipse-iceoryx/iceoryx/discussions/1149).

Issue #1141, opened 2022-02-21
Refactoring of ServiceDescription


FerdinandSpitzschnueffler:
## Brief feature description

The `ServiceDescription` has a few members which seem unused. The scope might be removed, the classhash could for instance be used for storing type information. Furthermore the constructor takes a lot of parameters, so we could think of using a `ServiceDescriptionBuilder`. 

## Detailed information

- [ ] Clarify if we need all members
    - [ ] Can we remove the scope?
    - [ ] Do we need the interface source?
    - [ ] Will we use the classhash?
    - [ ] Do we want to have a single string instead of three string identifiers, see #1052 
- [ ] Use a `ServiceDescriptionBuilder` like proposed [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1047#discussion_r792652416)

PR #1137, opened 2022-02-21
iox-#27 add client and server port to runtime and roudi tests [stacked PR #3.1]

elBoberido:
## Pre-Review Checklist for the PR Author

1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Changelog updated [in the unreleased section][changelog] including API breaking changes
1. [x] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/release-notes/iceoryx-unreleased.md

## Notes for Reviewer

This PR contains the tests for #1087

Similar to publisher and subscriber, the code in the `ProcessManager` is indirectly tested via the `PoshRuntime` tests.
The `ProcessManager` is currently more or less untestable. I already created #1148 to refactor the `ProcessManager` and make it testable.

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
    - [x] Each unit test case has a unique UUID
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

- Relates to #27

PR #620, opened 2021-03-18
Iox #611 design draft shared memory object creation

mossmaurice:
## Pre-Review Checklist for the PR Author

1. [ ] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [ ] Tests follow the [best practice for testing][testing]
1. [ ] Branch follows the naming format (`iox-#123-this-is-a-branch`)
1. [ ] Commits messages are according to this [guideline][commit-guidelines]
    - [ ] Commit messages have the issue ID (`iox-#123 commit text`)
    - [ ] Commit messages are signed (`git commit -s`)
    - [ ] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [ ] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [ ] Relevant issues are linked
1. [ ] Add sensible notes for the reviewer
1. [ ] All checks have passed (except `task-list-completed`)
1. [ ] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/best-practice-for-testing.md

## Notes for Reviewer
* First version of a design draft as a follow-up of #404 
* Not for v1.0 but instead probably v2.0

## Checklist for the PR Reviewer

- [ ] Commits are properly organized and messages are according to the guideline
- [ ] Code according to our coding style and naming conventions
- [ ] Unit tests have been written for new behavior
- [ ] Public API changes are documented via doxygen
- [ ] Copyright owner are updated in the changed files
- [ ] PR title describes the changes

## Post-review Checklist for the PR Author

1. [ ] All open points are addressed and tracked via issues

## References

- Closes partly #611

PR #1154, opened 2022-02-22
iox-#722 Introduce defined behaviour for subscribers without read access

### LINKED ISSUE #722: Defined behaviour when creating subscriber without read access
## Brief feature description

If a user app does not have read access to any shared memory segment and a subscriber object is created, which `takes()` in case of a publisher is sending data, the current behaviour is a segfault.

## Detailed information

* Expected behaviour is that no data will ever be received from subscribers if the user has no read access to the shared memory segment to which the publisher is writing
  * This needs a runtime check in the discovery loop
  * Hence, add `posix::User to BasePort`
* Move `SegmentMapping` class to its own inl (segment_manager.hpp)

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

> The solution described in #722 recommends a runtime check inside `isCompatiblePubSub()`
> I decided against as we then have to attach the group info to every publisher and subscriber to compare them.
> IMHO this too complex for `v2.0` and a check on creation like it is already implemented for the `popo::Publisher` is sufficient

* Fixes a segfault if `popo::Subscriber::take()` was called without read access to the underlying shared memory segment

## References

- Closes #722 (partly)


iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -622,13 +622,15 @@ void PortManager::sendToAllMatchingSubscriberPorts(const capro::CaproMessage& me
...
-    auto responseMatch = !(server.getClientTooSlowPolicy() == popo::ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA
-                           && client.getResponseQueueFullPolicy() == popo::QueueFullPolicy::BLOCK_PRODUCER);
+    const auto requestMatch = !(client.getServerTooSlowPolicy() == popo::ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA
+                                && server.getRequestQueueFullPolicy() == popo::QueueFullPolicy::BLOCK_PRODUCER);
 
-    return requestMatch && responseMatch;
+    const auto responseMatch = !(server.getClientTooSlowPolicy() == popo::ConsumerTooSlowPolicy::DISCARD_OLDEST_DATA
+                                 && client.getResponseQueueFullPolicy() == popo::QueueFullPolicy::BLOCK_PRODUCER);
+
+    return serviceMatch && requestMatch && responseMatch;
```
elBoberido, line 633: Now I remember why I didn't do the service description check in this method. It does unnecessary checks for all ports even if the service descriptions do not match, which happens most of the time. Could you please do an early return when `serviceMatch` is false? Can you do the same for pub-sub?

mossmaurice: Done

iceoryx_posh/source/roudi/process_manager.cpp
```diff
@@ -416,6 +416,19 @@ void ProcessManager::addSubscriberForProcess(const RuntimeName_t& name,
 {
     findProcess(name)
         .and_then([&](auto& process) {
+            auto segmentInfo = m_segmentManager->getSegmentInformationWithReadAccessForUser(process->getUser());
```
elfenpiff, line 419: I may be wrong but I think this is the wrong approach. Here is why.
1. Let's assume publisher A is on shared memory segment A to which the subscriber B has no read access
2. But subscriber B has read access to shared memory segment B

When subscriber B would like to receive data it has to read from segment A.

With this algorithm you are just looking for any readable segment and not if segment A of the matching publisher is readable. Now the following happens.
1. Subscriber B is created since segment B is readable
2. Publisher A sends a sample to Subscriber B
3. Subscriber B is accessing the memory to which the sample is pointing and SEGFAULT.

elfenpiff (changes requested):
I think your approach is wrong and cannot be implemented in this way.

Lets assume a subscriber is created before the publisher. Then the subscriber does not know which segment the publisher is using, this is defined by the app/group and not via the service description.
When the subscriber has read access to some arbitrary segment your approach creates a subscriber but when now the publisher is storing its data on a segment to which the subscriber has no read access it will still segfault.

iceoryx_posh/source/roudi/process_manager.cpp
mossmaurice: @elfenpiff You're not wrong here. It is exactly as you described. IMHO the long-term solution is to have the groups stored in either the `ServiceDescription` or `popo::Publisher` and `popo::Subscriber` to be able to see if they match in `isCompatible{PubSub, ClientServer}`. However, this would be a major refactoring and is therefore out of scope for `v2.0`. This is something we should tackle together with the `ServiceDescription` refactoring in #1141.

For `v2.0` instead of silently failing with a segfault, which can lead to nasty debugging sessions for several days, I would like to add the proposed patches to fail early if the process has no access to any shared memory segment. This makes the world a little bit better and follows our iterative improvement process which we aim to establish for iceoryx.

@elBoberido @MatthiasKillat @elBoberido Would you prefer a segfault or a clear and concise error message for this case?

elfenpiff: @mossmaurice But your patch does not solve this - only the rare corner case that an application has neither read nor write access to any segment. But in this case we should ask ourselfs why can this application register itself at RouDi at all?

So for this corner case - app has no read/write access to anything - I would propose that RouDi denies the registration of the app with a NACK response to the initial register request.

But with your solution we would still end up in a segfault when the app has read access to something so it does not solves the problem.

I would not merge this into `v2.0` and move this whole issue to `v3.0` especially since time (review time & coding time) is precious this week.

Final note: `Unit tests have been written for new behavior` something I have to check as a reviewer and I cannot cross out since the (integration) tests for this are missing.

mossmaurice: >But with your solution we would still end up in a segfault when the app has read access to something so it does not solves the problem.

@elfenpiff Well, it's exactly this 'corner case' that we describe in the `ice_access_control` example. Imagine someone 
would copy and paste it and then run into the segfault. I completely agree that a `NACK` on registration makes more sense in the long run. I've sketched a hierarchical management of the entities in shared memory on #620. My proposal is to implement RAII e.g. if a node is destroyed also publisher, clients, etc. inside are destroyed.

> Final note: Unit tests have been written for new behavior something I have to check as a reviewer and I cannot cross out since the (integration) tests for this are missing.

Just yesterday a PR got merged with your approval without proper integration tests :clown_face: The [explanation](https://github.com/eclipse-iceoryx/iceoryx/pull/1137#discussion_r812065503) of @elBoberido makes sense to me.

iceoryx_examples/ice_access_control/README.md
```diff
@@ -153,7 +153,7 @@ other::--
 The cheeky app is started with the user _notallowed_. This user is not in any group that would allow him either read
 or write access to one of the shared memory segments. Hence, RouDi will print a warning in this case.
 
-Despite having no read access, subscribers can still be created. <!-- @todo In this case no data will ever arrive.-->
+When creating and requesting a subscriber RouDi will answer with an error, as there is no read access. Hence, an error will be printed and the cheeky app will stop.
```
MatthiasKillat, line 156:
```suggestion
When creating and requesting a subscriber RouDi will answer with an error, as there is no read access. Hence, an error will be printed and the app will call terminate.
```

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.inl
```diff
@@ -112,7 +112,7 @@ SegmentManager<SegmentType>::getSegmentInformationWithWriteAccessForUser(const p
 {
     auto groupContainer = user.getGroups();
 
-    SegmentUserInformation segmentInfo{cxx::nullopt_t(), 0u};
+    SegmentUserInformation segmentInfo{cxx::nullopt_t(), 0U};
```
MatthiasKillat, line 115: Can use `cxx::nullopt`

iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.inl
```diff
@@ -131,6 +131,31 @@ SegmentManager<SegmentType>::getSegmentInformationWithWriteAccessForUser(const p
...
+    // with the groups we can search for the readable segment of this user
+    for (const auto& groupID : groupContainer)
+    {
+        for (auto& segment : m_segmentContainer)
+        {
+            if (segment.getReaderGroup() == groupID)
+            {
+                segmentInfo.m_memoryManager = segment.getMemoryManager();
+                segmentInfo.m_segmentID = segment.getSegmentId();
+                return segmentInfo;
```
MatthiasKillat, line 151: So we return the "first" one if there is any and essentially none otherwise, OK.

So after this call we clearly know whether there is none.

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -528,7 +531,7 @@ bool PortManager::isCompatiblePubSub(const PublisherPortRouDiType& publisher,
     const bool historyRequestIsCompatible =
         !subOpts.requiresPublisherHistorySupport || subOpts.historyRequest <= pubOpts.historyCapacity;
 
-    return servicesMatch && blockingPoliciesAreCompatible && historyRequestIsCompatible;
+    return blockingPoliciesAreCompatible && historyRequestIsCompatible;
```
MatthiasKillat, line 534: We could also return early if the blocking policies are not compatible to be consistent. Would be more efficient. Whether we want the more expressive bool variables is a matter of taste, they will be optimized out in all likelihood.

iceoryx_posh/source/roudi/process_manager.cpp
MatthiasKillat: @mossmaurice I think the proper way to avoid this issue with the SEGFAULT the information about the segment which the publisher is writing to needs to be propagated to the subscriber when they initially establish connection. If this segment is in the readable list of the app, then they get connected. Otherwise the publisher will not add the corresponding susbcriber to its list of queues. This can be part of some publisher settings , essentially a private(i.e. not directly user configurable) part of the options which are considered during the matching process. We can also consider later the option to configure a segment affinity for publishers.

Later on another publisher which writes to a segment we can read from may become available, so we should not deny creation of the subscriber (and even a warning is questionable, it will just get no data in the meantime and not be connected).

Currently we fix only the SEGFAULT if we cannot read from any segment. This is better than before but still does not solve the main issue. Whether this is valuable enough for us right now I cannot say, but it is at least not worse than before.

MatthiasKillat (commented):
I think it is not the proper solution to the problem but also no regression.

iceoryx_posh/source/roudi/port_manager.cpp
mossmaurice: Already done on `master` in the meantime.

Issue #1142, opened 2022-02-21
ServiceDiscovery C binding


FerdinandSpitzschnueffler:
## Brief feature description

We need to implement the C binding for the `ServiceDiscovery`.

## Detailed information

- [x] implement `service_discovery.h` + tests
- [x] add attach/detach event functions to `listener.h` + tests
- [x] add attach/detach event and state functions to `wait_set.h` + tests
- [x] add `*_does_originate_from_*`, `*_get_*_origin` functions to `notification_info.h` + tests
- [x] add necessary enums to `enums.h`
- [x] if necessary, add enum translation to `c2cpp_enum_translation.cpp` and `cpp2c_enum_translation.cpp` + tests
- [x] implement `icediscovery_in_c` example

## References

* Comes from #415

PR #1165, opened 2022-02-24
Iox #1142 icediscovery in c

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
This PR introduces an example for service discovery in C.

## References

- Closes #1165

### DIFF
--- a/iceoryx_binding_c/include/iceoryx_binding_c/config.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/config.h
@@ -97,7 +97,7 @@ uint32_t iox_cfg_max_runtime_name_length();
 /// @brief the maximum size of a node name string + \0 terminator
 #define IOX_CONFIG_NODE_NAME_SIZE 101
 
-/// @brief the maximum size of a service description string identifier
-#define IOX_CONFIG_SERVICE_STRING_SIZE 100
+/// @brief the maximum size of a service description string identifier + \0 terminator
+#define IOX_CONFIG_SERVICE_STRING_SIZE 101
 
 #endif // IOX_BINDING_C_CONFIG_H
--- a/iceoryx_binding_c/include/iceoryx_binding_c/config.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/config.h
@@ -97,7 +97,7 @@ uint32_t iox_cfg_max_runtime_name_length();
 /// @brief the maximum size of a node name string + \0 terminator
 #define IOX_CONFIG_NODE_NAME_SIZE 101
 
-/// @brief the maximum size of a service description string identifier + \0 terminator
-#define IOX_CONFIG_SERVICE_STRING_SIZE 101
+/// @brief the maximum size of a service description string identifier
+#define IOX_CONFIG_SERVICE_STRING_SIZE 100
 
 #endif // IOX_BINDING_C_CONFIG_H
--- a/iceoryx_binding_c/include/iceoryx_binding_c/config.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/config.h
@@ -97,4 +97,7 @@ uint32_t iox_cfg_max_runtime_name_length();
 /// @brief the maximum size of a node name string + \0 terminator
 #define IOX_CONFIG_NODE_NAME_SIZE 101
 
+/// @brief the maximum size of a service description string identifier + \0 terminator
+#define IOX_CONFIG_SERVICE_STRING_SIZE 101
+
 #endif // IOX_BINDING_C_CONFIG_H
--- a/iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
@@ -23,9 +23,9 @@
 
 typedef struct
 {
-    char serviceString[IOX_CONFIG_SERVICE_STRING_SIZE];
-    char instanceString[IOX_CONFIG_SERVICE_STRING_SIZE];
-    char eventString[IOX_CONFIG_SERVICE_STRING_SIZE];
+    char serviceString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U]; // +1U for \0 terminator
+    char instanceString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U];
+    char eventString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U];
 } iox_service_description_t;
 
 #endif
--- a/iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,9 +23,9 @@
 
 typedef struct
 {
-    char serviceString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U]; // +1U for \0 terminator
-    char instanceString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U];
-    char eventString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U];
+    char serviceString[IOX_CONFIG_SERVICE_STRING_SIZE];
+    char instanceString[IOX_CONFIG_SERVICE_STRING_SIZE];
+    char eventString[IOX_CONFIG_SERVICE_STRING_SIZE];
 } iox_service_description_t;
 
 #endif
--- a/iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
@@ -17,13 +17,15 @@
 #ifndef IOX_BINDING_C_SERVICE_DESCRIPTION_H
 #define IOX_BINDING_C_SERVICE_DESCRIPTION_H
 
+#include "iceoryx_binding_c/config.h"
+
 #include <stdint.h>
 
 typedef struct
 {
-    char serviceString[100U];
-    char instanceString[100U];
-    char eventString[100U];
+    char serviceString[IOX_CONFIG_SERVICE_STRING_SIZE];
+    char instanceString[IOX_CONFIG_SERVICE_STRING_SIZE];
+    char eventString[IOX_CONFIG_SERVICE_STRING_SIZE];
 } iox_service_description_t;
 
 #endif
--- a/iceoryx_binding_c/test/moduletests/test_service_description.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_service_description.cpp
@@ -34,11 +34,11 @@ TEST(iox_service_description_test, StringSizesAreCorrect)
 {
     ::testing::Test::RecordProperty("TEST_ID", "f32a6d19-9ff7-4913-a1d8-39d46267b114");
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().serviceString)),
-                Eq(iox::capro::IdString_t().capacity()));
+                Eq(iox::capro::IdString_t().capacity() + 1U));
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().instanceString)),
-                Eq(iox::capro::IdString_t().capacity()));
+                Eq(iox::capro::IdString_t().capacity() + 1U));
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().eventString)),
-                Eq(iox::capro::IdString_t().capacity()));
+                Eq(iox::capro::IdString_t().capacity() + 1U));
 }
 
 } // namespace
--- a/iceoryx_binding_c/test/moduletests/test_service_description.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_service_description.cpp
@@ -32,13 +32,14 @@ using namespace iox::capro;
 
 TEST(iox_service_description_test, StringSizesAreCorrect)
 {
+    constexpr uint64_t ZERO_TERMINATOR_SIZE = 1U;
     ::testing::Test::RecordProperty("TEST_ID", "f32a6d19-9ff7-4913-a1d8-39d46267b114");
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().serviceString)),
-                Eq(iox::capro::IdString_t().capacity() + 1U));
+                Eq(iox::capro::IdString_t().capacity() + ZERO_TERMINATOR_SIZE));
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().instanceString)),
-                Eq(iox::capro::IdString_t().capacity() + 1U));
+                Eq(iox::capro::IdString_t().capacity() + ZERO_TERMINATOR_SIZE));
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().eventString)),
-                Eq(iox::capro::IdString_t().capacity() + 1U));
+                Eq(iox::capro::IdString_t().capacity() + ZERO_TERMINATOR_SIZE));
 }
 
 } // namespace
--- a/iceoryx_binding_c/test/moduletests/test_service_description.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_service_description.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Apex.AI Inc. All rights reserved.
+// Copyright (c) 2021 - 2022 Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_examples/icediscovery_in_c/README.md
+++ b/iceoryx_examples/icediscovery_in_c/README.md
@@ -28,7 +28,7 @@ To be able to search for services, we need to include:
 #include "iceoryx_binding_c/service_discovery.h"
 ```
 
-We create some stack storage for the service discovery and initialize it.
+We create some storage for the service discovery and initialize it.
 
 <!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][create service discovery handle]-->
 ```c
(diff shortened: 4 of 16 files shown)


iceoryx_examples/icediscovery_in_c/iox_c_find_service.c
```diff
@@ -0,0 +1,120 @@
...
+
+void printSearchResult(const iox_service_description_t service)
+{
+    printf(
+        "- Service: %s, Instance: %s, Event: %s\n", service.serviceString, service.instanceString, service.eventString);
+}
+
+void searchFrontCameras(const iox_service_description_t service, void* count)
+{
+    if (strcmp(service.instanceString, "FrontLeft") == 0 || strcmp(service.instanceString, "FrontRight") == 0)
```
MatthiasKillat, line 48: There is this advice to use https://www.cplusplus.com/reference/cstring/strncmp/ to avoid errors due to missing zero termination but I am not sure we want to do this here (code gets uglier).  Up to you.

The literal is obviously zero-terminated, the strings in `service` basically are returned from our API so ...

iceoryx_examples/icediscovery_in_c/sleep_for.h
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
```
MatthiasKillat, line 1: wrong copyright?

iceoryx_examples/icediscovery_in_c/sleep_for.h
```diff
@@ -0,0 +1,37 @@
...
+
+void sleep_for(uint32_t milliseconds)
+{
+    Sleep((uint64_t)milliseconds);
+}
+
+#else
+#include <unistd.h>
+
+void sleep_for(uint32_t milliseconds)
```
MatthiasKillat, line 31: Is the choice of `uint32_t` here deliberate to be able to safely multiply with 1000 and get a non-overflowing value? I think we can have an overflow in the multiplication and then things get weird (of course, this requires fairly large values > 2^32 /10³) 

According to https://man7.org/linux/man-pages/man3/usleep.3.html `useconds_t` appears to be an `unsigned int` (usually). But portability is shaky here anyway, we cannot do much except to work with microseconds in the call.

But it is only example code and not called with large values. If it were library code we would need to make it safer.

TLDR: For a lib we would need to make it safer by using `uint32_t` microseconds (and the windows call would need to perform a division by 1000). Not needed in example here.

iceoryx_examples/icediscovery_in_c/sleep_for.h
FerdinandSpitzschnueffler: Nope, file copied from `icedelivery_in_c` example.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+which points to `storage`, the memory location where the service discovery is
+stored.
+
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][create service discovery handle]-->
+```c
+iox_service_discovery_storage_t storage;
+iox_service_discovery_t serviceDiscovery = iox_service_discovery_init(&storage);
+```
+
+It is included via:
```
MatthiasKillat, line 34: What is it? The service discovery? The handle? ... 

I would somehow start with this (to use service discovery, we need to include ...) before mpving on to service discovery creation/initialization.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][print function to be applied to search results]-->
+```c
+void printSearchResult(const iox_service_description_t service)
+{
+    printf(
+        "- Service: %s, Instance: %s, Event: %s\n", service.serviceString, service.instanceString, service.eventString);
+}
+```
+
+We can not only search for exactly matching services, but also for wildcards
```
elfenpiff, line 63: , in combination with wildcards (`NULL`) we can search also for all instances and events with the service `Radar`.

elfenpiff, line 63: We can not only search for exact matching services,

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+using `NULL`:
+
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][search for all Radar services]-->
+```c
+iox_service_discovery_find_service_apply_callable(
+    serviceDiscovery, "Radar", NULL, NULL, printSearchResult, MessagingPattern_PUB_SUB);
+```
```
elfenpiff, line 70: The wildcard can be used for all strings which describe a service, e.g. service, instance and event.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+                                                         NULL,
+                                                         NULL,
+                                                         searchResult,
+                                                         SEARCH_RESULT_CAPACITY,
+                                                         &missedServices,
+                                                         MessagingPattern_PUB_SUB);
+```
+
+`searchResult` is a `iox_service_description_t` array of size
+`SEARCH_RESULT_CAPACITY`. The found services are written into this array, but
```
elfenpiff, line 90: The matching services are written

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][include service discovery]-->
+```c
+#include "iceoryx_binding_c/service_discovery.h"
+```
+
+We can now call three different find service functions. Let's start with
+
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][find service and apply callable]-->
+```c
+iox_service_discovery_find_service_apply_callable(
```
MatthiasKillat, line 45: The terminology of callable is more of a C++ one, in C only functions and function pointers are callable and we can only pass the latter (there are no C++ functors). Still, I would keep the name as it is correct but maybe unusual to a C user.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+`Camera` services but additionally count the front camera services:
+
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][search for all front camera services]-->
+```c
+iox_service_discovery_find_service_apply_callable_with_context_data(
+    serviceDiscovery, "Camera", NULL, NULL, searchFrontDevices, &numberFrontCameras, MessagingPattern_PUB_SUB);
+```
+
+This function is quite similar to the first find service function, but we
+pass an additional argument:
```
elfenpiff, line 107: pass an additional argument `numberFrontCameras`.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][store number of front cameras]-->
+```c
+uint32_t numberFrontCameras = 0U;
+```
+
+We use this variable to store the number of front cameras and provide it as
+second parameter to the function which is applied to all found services:
```
elfenpiff, line 115: and provide a pointer to it as second argument `count` to the function `searchFrontDevices` which is applied

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+                                                         NULL,
+                                                         searchResult,
+                                                         SEARCH_RESULT_CAPACITY,
+                                                         &missedServices,
+                                                         MessagingPattern_PUB_SUB);
+```
+
+`searchResult` is a `iox_service_description_t` array of size
+`SEARCH_RESULT_CAPACITY`. The found services are written into this array, but
+at maximum `SEARCH_RESULT_CAPACITY`. If the number of found services exceeds
```
MatthiasKillat, line 91:
```suggestion
up to a maximum of `SEARCH_RESULT_CAPACITY`. If the number of found services exceeds
```

iceoryx_examples/icediscovery_in_c/README.md
MatthiasKillat, line 90:
```suggestion
`SEARCH_RESULT_CAPACITY`. The found services are written into this array
```

iceoryx_examples/icediscovery_in_c/iox_c_find_service.c
```diff
@@ -0,0 +1,134 @@
...
+            printSearchResult(searchResult[i]);
+        }
+
+        //! [search for all front camera services]
+        iox_service_discovery_find_service_apply_callable_with_context_data(
+            serviceDiscovery, "Camera", NULL, NULL, searchFrontDevices, &numberFrontCameras, MessagingPattern_PUB_SUB);
+        //! [search for all front camera services]
+        printf("\nFound %u front cameras\n", numberFrontCameras);
+
+        sleep_for(1000);
```
elfenpiff, line 128: Could you please remove the magic number here:
```c
const uint32_t WAIT_TIME_IN_MS = 1000;
sleep_for(WAIT_TIME_IN_MS);
```

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,131 @@
...
+the array's capacity, the number of not stored services is written to
```
MatthiasKillat, line 92:
```suggestion
the array's capacity, the number of services that could not be stored is written to
```

iceoryx_examples/icediscovery_in_c/README.md
MatthiasKillat, line 115: Maybe it is more important that it is an output parameter to return the count. Do we want to add this information or is this clear from usage? To me it is but I may be biased.

iceoryx_examples/icediscovery_in_c/iox_c_find_service.c
```diff
@@ -0,0 +1,134 @@
...
+{
+    printf(
+        "- Service: %s, Instance: %s, Event: %s\n", service.serviceString, service.instanceString, service.eventString);
+}
+//! [print function to be applied to search results]
+
+//! [search function for all front devices]
+void searchFrontDevices(const iox_service_description_t service, void* count)
+{
+    if (strncmp(service.instanceString, "FrontLeft", IOX_CONFIG_SERVICE_STRING_SIZE - 1U) == 0
```
MatthiasKillat, line 53: This `-1U` here on user side is unfortunate and error-prone. Do we want to define the string size *without*  the zero terminator included and compensate with a +1 in the tests instead where needed?

Or even have two constants, i.e. `IOX_CONFIG_SERVICE_STRING_SIZE_WITH_ZERO_TERMINATON` in addition.

From https://www.cplusplus.com/reference/cstring/strncmp/ I am not sure we really need to subtract 1 here though for correctness. It will not exceed the buffer either way. In this case we would only drop the `-1U`.

MatthiasKillat (commented):
Looks good, some final minor changes should be considered.

iceoryx_examples/icediscovery_in_c/sleep_for.h
MatthiasKillat: @FerdinandSpitzschnueffler I think this is of no concern as we use it in the example and can be resolved.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,134 @@
...
+To be able to search for services, we need to include:
+
+<!--[geoffrey][iceoryx_examples/icediscovery_in_c/iox_c_find_service.c][include service discovery]-->
+```c
+#include "iceoryx_binding_c/service_discovery.h"
+```
+
+To get a handle to the service discovery we call `iox_service_discovery_init`.
+The service discovery requires memory of the amount of
+`sizeof(iox_service_discovery_storage_t)` to be initialized.
```
MatthiasKillat, line 33: Ok, but this is hidden from the user and an implementation detail. From my point of view the important information is that the user must provide an instance of `iox_service_discovery_storage_t` to an initialization function. The size of objects of this class is not relevant to the user and taking care of by the struct itself.

Technically the user can pass any buffer of sufficient size (requires a cast), but it is not intended to be used that way...

I would hence not mention the detail wit `sizeof`, but not a must change ...

MatthiasKillat (dismissed):

iceoryx_examples/icediscovery_in_c/README.md
MatthiasKillat, line 33: Technically the only information relevant to the user is that she has to pass a struct of type `iox_service_discovery_storage_t` which will be of the correct size (by construction/API). 

Hence one could leave out the information with `sizeof`. It may give people ideas to use it with raw buffers of that size (which will work if correctly aligned but is not encouraged).

Not a must change but should probably be omitted.

iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
```diff
@@ -17,13 +17,15 @@
...
+#include "iceoryx_binding_c/config.h"
+
 #include <stdint.h>
 
 typedef struct
 {
-    char serviceString[100U];
-    char instanceString[100U];
-    char eventString[100U];
+    char serviceString[IOX_CONFIG_SERVICE_STRING_SIZE + 1U]; // +1U for \0 terminator
```
elfenpiff, line 26: Please do not do this. `IOX_CONFIG_SERVICE_STRING_SIZE` should contain the actual capacity of the string including the null terminator.
Otherwise we have a weird behavior when for instance we have `sizeof` calculations etc.

iceoryx_binding_c/include/iceoryx_binding_c/config.h
```diff
@@ -97,4 +97,7 @@ uint32_t iox_cfg_max_runtime_name_length();
 /// @brief the maximum size of a node name string + \0 terminator
 #define IOX_CONFIG_NODE_NAME_SIZE 101
 
+/// @brief the maximum size of a service description string identifier
+#define IOX_CONFIG_SERVICE_STRING_SIZE 100
```
elfenpiff, line 101: Please change it to 101

iceoryx_binding_c/test/moduletests/test_service_description.cpp
```diff
@@ -32,13 +32,14 @@ using namespace iox::capro;
 
 TEST(iox_service_description_test, StringSizesAreCorrect)
 {
+    constexpr uint64_t ZERO_TERMINATOR_SIZE = 1U;
     ::testing::Test::RecordProperty("TEST_ID", "f32a6d19-9ff7-4913-a1d8-39d46267b114");
     EXPECT_THAT(sizeof(decltype(std::declval<iox_service_description_t>().serviceString)),
-                Eq(iox::capro::IdString_t().capacity()));
+                Eq(iox::capro::IdString_t().capacity() + ZERO_TERMINATOR_SIZE));
```
elfenpiff, line 38: The null terminator is always included in string literals, see
```c
const char blubb[] = "EvilHuhn";
printf("%lu \n", sizeof(blubb) / sizeof(char) ); // = 9, since the zero terminator is always part of the array.
```

iceoryx_examples/icediscovery_in_c/iox_c_find_service.c
```diff
@@ -0,0 +1,135 @@
...
+{
+    printf(
+        "- Service: %s, Instance: %s, Event: %s\n", service.serviceString, service.instanceString, service.eventString);
+}
+//! [print function to be applied to search results]
+
+//! [search function for all front devices]
+void searchFrontDevices(const iox_service_description_t service, void* count)
+{
+    if (strncmp("FrontLeft", service.instanceString, IOX_CONFIG_SERVICE_STRING_SIZE) == 0
```
elfenpiff, line 53: If clang-tidy spits out a warning here, please ignore it, see this little thing for instance:
```c
 #include <stdio.h>
 #include <string.h>
 #define ID_NAME 6
 int main() {
     const char service[ID_NAME] = "12453";
     return strncmp("12", service, 15); // clang-tidy warning
     return strncmp("12", service, 15 - 0); // NO clang-tidy warning
     return strncmp("12", service, 15 - 1230); // NO clang-tidy warning
 }

```

It may makes sense to add `-bugprone-not-null-terminated-result` as exclusion to our root `.clang-tidy` file

iceoryx_binding_c/include/iceoryx_binding_c/service_description.h
MatthiasKillat: I would argue the string size is n and the buffer size is n +1.

iceoryx_binding_c/test/moduletests/test_service_description.cpp
MatthiasKillat: For literals sure but `IdString_t::capacity` returns a capacity which excludes the null terminator.

iceoryx_binding_c/include/iceoryx_binding_c/config.h
MatthiasKillat: This is not very intuitive as it is the buffer size.

iceoryx_examples/icediscovery_in_c/README.md
```diff
@@ -0,0 +1,132 @@
...
+                                                         &missedServices,
+                                                         MessagingPattern_PUB_SUB);
+```
+
+`searchResult` is a `iox_service_description_t` array of size
+`SEARCH_RESULT_CAPACITY`. The matching services are written into this array, up
+to a maximum of `SEARCH_RESULT_CAPACITY`. If the number of found services
+exceeds the array's capacity, the number of services that could not be stored is
+written to `missedServices`. The number of stored services is returned. Since
+the `cameraPublishers` periodically offer/stop their services, you should see
```
MatthiasKillat, line 94:
```suggestion
the `cameraPublishers` periodically offer/stop their services, we should see
```

MatthiasKillat (approved):

iceoryx_examples/icediscovery_in_c/README.md
FerdinandSpitzschnueffler: I find this clear from usage. Additionally we state `We use this variable to store the number...`

iceoryx_examples/icediscovery_in_c/iox_c_find_service.c
FerdinandSpitzschnueffler: I added the exclusion

elfenpiff (approved):

dkroenke (approved):

PR #1253, opened 2022-03-08
iox-#1252 Check QNX and C++ STL system headers against whitelist

### LINKED ISSUE #1252: Check QNX and STL system headers against whitelist
## Brief feature description

Only certain system headers can be used in production code. Developers should be aware of adding new headers.

## Detailed information

* [x] Add a check in CI to compare against a whitelist of system headers
* [ ] ~~Run the script on QNX only on the installed headers (currently blocked by missing QNX job)~~
   * E.g. `build/generated/iceoryx/include/`
* [ ] ~~Check that C++ version of headers are used instead of the C ones (`cstdio` vs. `stdio.h`)~~
* [ ] ~~[Check](https://github.com/eclipse-iceoryx/iceoryx/pull/1253#discussion_r849545417) if `process.h` can be removed for QNX build~~

## References

* Follow-up to #765

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Remove `///`,`/* .. */`-comments, etc. before grepping for `std::`
* Add CI check when new QNX or C++ STL system headers are added
* Remove some unused system headers

## References

- Closes #1252


FerdinandSpitzschnueffler:
@mossmaurice Can you please rename `whitelist`, e.g. to `allowlist`?

tools/scripts/stl-header-whitelist.txt
```diff
@@ -0,0 +1,62 @@
...
+# iceoryx_posh / iceoryx_hoofs system headers
+algorithm
+assert.h
+atomic
+bitset
+cassert
+chrono
+climits
+cmath
+cpptoml.h
```
elfenpiff, line 34: This is not a STL header.

tools/scripts/stl-header-whitelist.txt
```diff
@@ -0,0 +1,62 @@
...
+initializer_list
+iomanip
+iostream
+limits
+map
+memory
+mutex
+new
+numeric
+process.h
```
elfenpiff, line 53: This is not a STL header.

tools/scripts/stl-header-whitelist.txt
```diff
@@ -0,0 +1,62 @@
...
+sys/socket.h
+sys/stat.h
+sys/time.h
+sys/types.h
+sys/un.h
+sys/wait.h
+unistd.h
+# iceoryx_posh / iceoryx_hoofs system headers
+algorithm
+assert.h
```
elfenpiff, line 27: This is not a STL header.

tools/scripts/stl-header-whitelist.txt
```diff
@@ -0,0 +1,62 @@
+# QNX platform system headers
+arpa/inet.h
```
elfenpiff, line 2: Those following `.h` files are either posix lib c header or OS headers which are present in most unix operating systems.

elfenpiff (changes requested):

tools/scripts/stl-header-whitelist.txt
mossmaurice: I adapted the title.

tools/scripts/stl-header-whitelist.txt
mossmaurice: I adapted the title.

budrus (dismissed):

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,73 @@
...
         GREP_PATH="${GREP_PATH} ${WORKSPACE}/${COMPONENT}/$DIR"
     done
 done
 
+echo "# QNX platform/ libc headers" >> $TEMP_FILE
+
+echo "# usage of headers for QNX"
+gcc -w -fpreprocessed -dD -E $(find $QNX_PLATFORM_DIR -type f -iname *.cpp -o -iname *.hpp) \
+ | grep -e "#include <" \
+ | grep -oP "\<\K[^<>]+(?=>)" \
```
elBoberido, line 44: Please save this regex in a variable and also add a comment what it does

tools/scripts/header-allowlist.txt
```diff
@@ -0,0 +1,62 @@
...
+cstdlib
+cstring
+ctime
+fstream
+functional
+initializer_list
+iomanip
+iostream
+limits
+map
```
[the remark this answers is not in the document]
mossmaurice: This is the current list of header that are used in `iceoryx_posh` and `iceoryx_hoofs`. Of course you are right, that this header will not be used in a production version using iceoryx. We'll narraw the headers down when having an alternative for map.

tools/scripts/header-allowlist.txt
```diff
@@ -0,0 +1,62 @@
...
+memory
+mutex
+new
+numeric
+process.h
+regex
+sstream
```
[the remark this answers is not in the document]
mossmaurice: See [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1253#discussion_r840797403), it's ok for now, but will be removed later.

tools/scripts/header-allowlist.txt
```diff
@@ -0,0 +1,62 @@
...
+stdint.h
+stdio.h
+stdlib.h
+string
```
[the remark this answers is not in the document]
mossmaurice: See [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1253#discussion_r840797403), it's ok for now, but will be removed later.

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,73 @@
...
 for COMPONENT in ${COMPONENTS[@]}; do
     for DIR in ${SOURCE_DIR[@]}; do
         GREP_PATH="${GREP_PATH} ${WORKSPACE}/${COMPONENT}/$DIR"
     done
 done
 
+echo "# QNX platform/ libc headers" >> $TEMP_FILE
+
+echo "# usage of headers for QNX"
+gcc -w -fpreprocessed -dD -E $(find $QNX_PLATFORM_DIR -type f -iname *.cpp -o -iname *.hpp) \
```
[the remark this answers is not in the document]
mossmaurice: I agree with you that this would be definitely a nicer solution! However, as `gcc`s preprocessing stage is invoked here and it e.g. doesn't handle `.inl` and `.in`, that's the reason why `.inl`s are currently passed unprocessed with comments, I would argue to go with common sense here and not consider extensions like `.cxx`, `.cc`, etc. as they are not used in iceoryx yet. Could you go along with that?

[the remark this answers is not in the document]
elBoberido: Could you then also please check the headers from `build/generated` or even better the headers which are installed? This would be the best but then the script would need to run on QNX

tools/scripts/header-allowlist.txt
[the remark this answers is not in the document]
elBoberido: Then the name is misleading. Why is it called `allowlist` when it actually is an `uselist`. Where is the documentation how to handle new header? The more I think about this, the more I come to the conclusion that it is a bad idea to handle it this way.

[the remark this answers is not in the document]
elfenpiff: I fear a little bit the same. But as a compromise would could document that only new files are only allowed the includes listed here and nothing more and then we can step by step refactor the old code.

But what about unit tests, examples, introspection client, ?dds gateway? etc. They do not have such restrictions, how would we handle them?

What about hidden include dependencies. Let's say `iostream` is allowed and the STL implementation requires to include `map` which is not allowed?

elfenpiff:
@elBoberido @mossmaurice What do you think about an approach where we create a CI target with a libc/libc++ implementation where we just delete all forbidden headers and compile iceoryx on that machine. Or we delete everything except the headers we allow.
This could be done by a simple `rm -rf ...` in a shell script.

The compilation will fail when one uses an illegal header and this is definitely future proof from a code perspective.

elBoberido:
> @elBoberido @mossmaurice What do you think about an approach where we create a CI target with a libc/libc++ implementation where we just delete all forbidden headers and compile iceoryx on that machine. Or we delete everything except the headers we allow. This could be done by a simple `rm -rf ...` in a shell script.
> 
> The compilation will fail when one uses an illegal header and this is definitely future proof from a code perspective.
> 
> Furthermore it would solve the issue of checking implicit dependencies when an allowed header includes and illegal one.

@elfenpiff that's not going to work. Consider we allow header `foo` which in turn includes header `bar` then we do not care if `foo` includes `bar` since we can ask the vendor of `foo` to give us a header which matches our production guidelines, which in turn apply to all transient dependencies.

tools/scripts/header-allowlist.txt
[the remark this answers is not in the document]
mossmaurice: @elBoberido That's a good point. I changed the name to `used-headers.txt` and documented the approach in `CONTRIBUTING.md`

@elfenpiff The header restriction only applies to `iceoryx_posh` and `iceoryx_hoofs` excluding their tests. Transitive dependencies are not in the scope of this script.

CONTRIBUTING.md
```diff
@@ -153,6 +154,15 @@ For overrides of virtual methods the `copydoc` tag can be used:
...
 A good example for code formatting and doxygen structure can be found in [swe_docu_guidelines.md (WIP)](./doc/aspice_swe3_4/swe_docu_guidelines.md)
 
+### External dependencies
+
+External dependencies such as the [STL](https://en.wikipedia.org/wiki/Standard_Template_Library) or
+other libaries shall be kept to a minium for `iceoryx_posh` and `iceoryx_hoofs`. If you think a new dependency is
+necessary, do the following:
+
+1. Engage with the maintainers beforehand by opening an issue to discuss the necessity
+1. If accepted, add the new header to `tools/scripts/used-headers.txt` for the CI to pass
```
elfenpiff, line 164: Please write 2. so that is also readable in my favorite editor vim

CONTRIBUTING.md
```diff
@@ -153,6 +154,15 @@ For overrides of virtual methods the `copydoc` tag can be used:
 
 A good example for code formatting and doxygen structure can be found in [swe_docu_guidelines.md (WIP)](./doc/aspice_swe3_4/swe_docu_guidelines.md)
 
+### External dependencies
+
+External dependencies such as the [STL](https://en.wikipedia.org/wiki/Standard_Template_Library) or
+other libaries shall be kept to a minium for `iceoryx_posh` and `iceoryx_hoofs`. If you think a new dependency is
+necessary, do the following:
+
+1. Engage with the maintainers beforehand by opening an issue to discuss the necessity
```
elfenpiff, line 163: This sounds a little bit weird. How about: Talk with the maintainers ... or Contact the maintainers beforehand...

iceoryx_posh/include/iceoryx_posh/internal/runtime/ipc_interface_base.hpp
```diff
@@ -23,20 +23,20 @@
...
+#include "iceoryx_hoofs/platform/time.hpp"
 #include "iceoryx_hoofs/platform/types.hpp"
 #include "iceoryx_hoofs/platform/unistd.hpp"
 #include "iceoryx_hoofs/posix_wrapper/named_pipe.hpp"
 #include "iceoryx_posh/iceoryx_posh_types.hpp"
 #include "iceoryx_posh/internal/runtime/ipc_message.hpp"
 
 #include <cstdint>
-#include <errno.h>
 #include <stdio.h>
```
elfenpiff, line 37: Please write here `#include <cstdio>` instead of `stdio.h`

iceoryx_posh/include/iceoryx_posh/internal/runtime/ipc_interface_base.hpp
```diff
@@ -23,20 +23,20 @@
...
 #include <stdlib.h>
```
elfenpiff, line 38: Please write here `#include <cstdlib>` instead of `stdlib.h`.

elfenpiff: Is it actually possible with your script to check that no libc headers are included in c++ files but instead the c++ version of the c header?
Like with `cstdio` instead of `stdio.h` since it is best practices.

iceoryx_posh/source/roudi/application/roudi_app.cpp
```diff
@@ -34,8 +35,7 @@
 #include "iceoryx_posh/internal/roudi/roudi.hpp"
 #include "iceoryx_posh/roudi/cmd_line_args.hpp"
 
-#include "stdio.h"
-#include <signal.h>
+#include <stdio.h>
```
elfenpiff, line 38: Please write `cstdio`

tools/scripts/used-headers.txt
```diff
@@ -0,0 +1,62 @@
+# QNX platform/ libc headers
+arpa/inet.h
```
elfenpiff, line 2: All QNX platform /libc headers should never be used in our C++ code base directly - except in platforms.
Our c++ code base should only include our platform files. If you could do this check we could avoid possible problems with other platforms..
Could you adjust the tests in that way?

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+TEMP_FILE=$(mktemp)
+GET_HEADER_NAME="\<\K[^<>]+(?=>)" # Matches the content between angle brackets
 
 for COMPONENT in ${COMPONENTS[@]}; do
     for DIR in ${SOURCE_DIR[@]}; do
         GREP_PATH="${GREP_PATH} ${WORKSPACE}/${COMPONENT}/$DIR"
     done
 done
 
+echo "# QNX platform/ libc headers" >> $TEMP_FILE
```
elBoberido, line 40: nitpick: in case there are other changes and you need to create additional commits, please add a space between `platform` and `/` like with `iceoryx_posh / iceoryx_hoofs`

tools/scripts/list_stl_dependencies.sh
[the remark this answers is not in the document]
elBoberido: Event though we currently don't have `.inl` files in the platform folder, please make it future proof nad handle it exactly like below with the header from iceoryx_hoofs and iceoryx_posh. These things tend to break when one is short on time.

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+        echo "One or more header is not on the list, please remove the header usage or extend the list!"
+        exit 1
+    fi
+    echo "No header divergence found!"
+fi
 
 echo
-echo files with stl dependency
+echo "# usage of std components"
+{ gcc -w -fpreprocessed -dD -E $(find $GREP_PATH -type f -iname *.cpp -o -iname *.hpp); cat $(find $GREP_PATH -type f -iname *.inl); } \
```
elBoberido, line 77: Shouldn't this and the following also be done for the files from `QNX_PLATFORM_DIR`?

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+QNX_PLATFORM_DIR=$WORKSPACE/iceoryx_hoofs/platform/qnx/
+USELIST=$WORKSPACE/tools/scripts/used-headers.txt
+TEMP_FILE=$(mktemp)
+GET_HEADER_NAME="\<\K[^<>]+(?=>)" # Matches the content between angle brackets
 
 for COMPONENT in ${COMPONENTS[@]}; do
     for DIR in ${SOURCE_DIR[@]}; do
         GREP_PATH="${GREP_PATH} ${WORKSPACE}/${COMPONENT}/$DIR"
     done
 done
```
elBoberido, line 38: `GREP_PATH` does not really tell anything about the variable. Can you find a better name, like with the `QNX_PLATFORM_DIR`?
Same with `TEMP_FILE`. Too generic. This makes maintaining harder.

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+ | uniq
+
+echo
+echo "# files with stl dependency"
 grep -RIne "std::" $GREP_PATH | sed -n  "s/\([^:]*\)\:[0-9]*\:.*std::[a-zA-Z_]*.*/\1/p" | xargs -I{} basename {} | sort | uniq
 
 echo
-echo using namespace with std component
+echo "# using namespace with std component"
 grep -RIne ".*using[ ]*namespace[ ]*std" $GREP_PATH | sed -n "s/\(.*\)/\ \ \1/p"
```
elBoberido, line 89: Please use `\s*` instead of `[ ]*`. Clang-format should replace all tabs with space and the compiler should issue a warning when something like no-break space is used but `\s` reduces the risk.

Additionally, the regex does not find `using namespace ::std;`.

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
 
 echo
-echo usage of std components
+echo "# usage of std components"
 grep -RIne "std::" $GREP_PATH | sed -n  "s/.*\(std::[a-zA-Z_]*\).*/\ \ \1/p" | sort | uniq
```
elBoberido, line 93: I have no idea what `sed` does and what to expect as output. For maintainability, please add a comment what one should expect as output.

Btw, this also prints `std::` without any component. Does this mean that the `sed` command is removing too much?

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+ | tee -a $TEMP_FILE \
+ | cat
+
+
+echo "# iceoryx_posh / iceoryx_hoofs headers" >> $TEMP_FILE
+# GCC can't preprocess .inl so we grep them in plain text
 echo
-echo usage of std components by file
-grep -RIne "std::" $GREP_PATH | sed -n  "s/\([^:]*\:[0-9]*\)\:.*\(std::[a-zA-Z_]*\).*/\ \ \1  \2/p" | sort | uniq
+echo "# usage of <..> header includes"
```
elBoberido, line 55: Could you please unify the output string and use either `"# iceoryx_posh / iceoryx_hoofs headers"` or `"# usage of <..> header includes"`?

I just ran the script and was confused by this. Same with the QNX header.

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+ | grep -oP $GET_HEADER_NAME \
+ | sort \
+ | uniq \
+ | tee -a $TEMP_FILE \
+ | cat
+
+if [[ "$SCOPE" == "check" ]]; then
+    echo
+    echo "Comparing the used system headers against the list.."
+    diff $TEMP_FILE $USELIST
```
elBoberido, line 67: Could you please store the output of the diff in a variable and print it after the message that something is fishy?

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+    if [ $? -eq 1 ]; then
+        echo "One or more header is not on the list, please remove the header usage or extend the list!"
```
elBoberido, line 69: This is confusing. How about ``Mismatch between expected and found header. Please check the diff below and remove/add the header in `tools/used-headers.txt`!``

In case you are bored, you could check for `>` and `<` in the output of the diff and explicitly list what needs to be removed or added.

elBoberido (commented):
I should stop reviewing this. The more I look at this the more I remember some C++ quirks and notice that the script misses some edge cases :grimacing: 

Could you also please check the generated header from `build/generated/iceoryx/include/`?

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
 grep -RIne "std::" $GREP_PATH | sed -n  "s/\([^:]*\)\:[0-9]*\:.*std::[a-zA-Z_]*.*/\1/p" | xargs -I{} basename {} | sort | uniq
 
 echo
-echo using namespace with std component
+echo "# using namespace with std component"
 grep -RIne ".*using[ ]*namespace[ ]*std" $GREP_PATH | sed -n "s/\(.*\)/\ \ \1/p"
 
 echo
-echo usage of std components
+echo "# usage of std components"
```
elBoberido, line 92: It seems this is printed twice. See also line 76

CONTRIBUTING.md
mossmaurice: There is no markdown plugin for vim? :astonished:

iceoryx_posh/include/iceoryx_posh/internal/runtime/ipc_interface_base.hpp
mossmaurice: Good point, I fixed it manually. And added the check for the C++ version as an item to the issue.

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,74 @@
...
+
+
+echo "# iceoryx_posh / iceoryx_hoofs headers" >> $TEMP_FILE
+# GCC can't preprocess .inl so we grep them in plain text
 echo
-echo usage of std components by file
-grep -RIne "std::" $GREP_PATH | sed -n  "s/\([^:]*\:[0-9]*\)\:.*\(std::[a-zA-Z_]*\).*/\ \ \1  \2/p" | sort | uniq
+echo "# usage of <..> header includes"
+{ gcc -w -fpreprocessed -dD -E $(find $GREP_PATH -type f -iname *.cpp -o -iname *.hpp); cat $(find $GREP_PATH -type f -iname *.inl); } \
+ | grep -e "#include <" \
```
elfenpiff, line 57: This is not sufficient, here look at the wonderful diversity of c++.

```cpp
%:include <iostream> // this is legal, see digraphs
#include "iostream" // legal when you add the include path with `-I`
#include <iostream> // classic
%:include "iostream" // another possibility

#include          <iostream> // also legal, spaces for the win ... also in all the above allowed
#include <   iostream  > // this is illegal, no spaces in < > are allowed
#include <iostream> #include <vector> // this is legal c++ code but the second include is not processed
```

So this is the standard weirdness ... now lets get weirder and have more fun
```cpp
#define HYPNOTOAD <iostream>
#include HYPNOTOAD // hehehe this is legal

#define HASSEL_THE_HOFF <iostream> \
#include <vector>

#include HASSEL_THE_HOFF // woohooo it compiles and we have iostream and vector
```

elfenpiff: I am aware that we do not have to handle every weird corner case but extra spaces or " instead of < should be handled.

tools/scripts/list_stl_dependencies.sh
mossmaurice: I meant to check just `iceoryx_hoofs` and `iceoryx_posh` here. We might include a large header, but just use a tiny little class from it. As of today, this is just for informational purposes.

tools/scripts/list_stl_dependencies.sh
mossmaurice: True indeed, replaced. As this is just for informational purposes currently I think we can live without `using namespace ::std;`

tools/scripts/list_stl_dependencies.sh
mossmaurice: AFAIK this just extracts all `std::` with all comments. As it was written by @elfenpiff I'm passing the question along. As of today this is just for informational purposes, if we want to run it on the CI we need to preprocess the file and removing the comments.

tools/scripts/list_stl_dependencies.sh
mossmaurice: If there is a difference in the files `diff` is outputting this information for us already.

tools/scripts/list_stl_dependencies.sh
mossmaurice: Removed the latter and kept the one without parsing the comments.

tools/scripts/list_stl_dependencies.sh
mossmaurice: Oh no why did you meantion ISO 646? :sweat_smile: You're completely correct with your analysis.! Instead of tackling every possible syntax in our little script, I suggest the following:

1. Run the alternative token check in `clang-tidy` and prevent any ISO 646 character from being used
2. `clang-format` ensures proper formatting
3. Preprocessor shenanigans like the ones you mentioned in the last paragraph should be found via a manual review

CONTRIBUTING.md
elfenpiff: Actually there is but I consider this bad style. In my opinion it is a feature of markdown that all the files are also readable in plain ASCII.

But as you can see during review this is not automatically converted so why not just enumerate it correctly.

mossmaurice: Well, three foods for thought:

1. It's readable in plain text
1. I'm very lazy
1. Maintainabilty is easier when moving lines around in a larger list

Would you install a markdown plugin for me or replace the numbers subsequently to right ones in your mind? :innocent:

tools/scripts/list_stl_dependencies.sh
elBoberido: @mossmaurice it is not about not having the info but at first I was wondering where to find the missing files and it took me some time to notice that the list is above the sentence that something is missing.

mossmaurice: Is the current message sufficient from your point of view?

`Mismatch of expected and found headers. Please check the diff above and remove/add the header in 'tools/used-headers.txt'!`

elBoberido (commented):
Almost there :)

Just in case you are also wondering where the `std::` comes from. It's from
```
using parentListPointer = typename std::
    conditional<IsConstIterator, const forward_list<T, Capacity>*, forward_list<T, Capacity>*>::type;
```

tools/scripts/used-headers.txt
```diff
@@ -0,0 +1,60 @@
...
+initializer_list
+iomanip
+iostream
+limits
+map
+memory
+mutex
+new
+numeric
+process.h
```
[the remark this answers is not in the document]
mossmaurice: As we don't have a QNX CI job yet to verify that, I added it as an item in #1252 

fyi @dkroenke

tools/scripts/list_stl_dependencies.sh
```diff
@@ -22,30 +22,76 @@
...
 
 echo
-echo usage of std components
-grep -RIne "std::" $GREP_PATH | sed -n  "s/.*\(std::[a-zA-Z_]*\).*/\ \ \1/p" | sort | uniq
+echo "# files with stl dependency"
+grep -RIne "std::" $GREP_PATH_HOOFS_POSH | sed -n  "s/\([^:]*\)\:[0-9]*\:.*std::[a-zA-Z_]*.*/\1/p" | xargs -I{} basename {} | sort | uniq
 
+echo
+echo "# using namespace with std component"
+grep -RIne ".*using\s*namespace\s*std" $GREP_PATH_HOOFS_POSH | sed -n "s/\(.*\)/\ \ \1/p"
```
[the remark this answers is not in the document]
mossmaurice: Fixed

elfenpiff (approved):

elBoberido (approved):

elfenpiff (approved):

elBoberido (approved):

Issue #996, opened 2021-12-16
Add `.expect` to expected and optional


elfenpiff:
## Brief feature description

Implement a generic functional interface and let the classes `optional` and `expected` inherit from it. This reduces code duplication in the implementation and test cases.

### Generic Functional Interface

Introduce a generic functional interface from which nullable classes can inherit. The advantage would be that methods like
* `and_then`
* `or_else`
* `value_or`
* `expect`
* ...

Need to be implemented and tested only once and every nullable user type can be enriched with this interface with minimal code chances. This would reduce a lot of code duplication in the implementation of `optional`, `expected` etc and other classes could be enriched and extended quickly with this interface.

Furthermore we would gain the advantage to quickly introduce other functional methods to those classes without touching them directly.

### .expect method

Add a method `.expect` to the functional interface which requires the underlying class to have a certain result otherwise it will terminate.
Coming from a PR suggestion: https://github.com/eclipse-iceoryx/iceoryx/pull/994#discussion_r770530769

### Example

Bad:
```cpp
m_semaphore{std::move(Semaphore::create(CreateUnnamedSingleProcessSemaphore, 0U)
                                .or_else([](auto) {
                                    std::cerr << "Unable to create semaphore for signal watcher" << std::endl;
                                    constexpr bool UNABLE_TO_CREATE_SEMAPHORE_FOR_SIGNAL_WATCHER = false;
                                    cxx::Ensures(UNABLE_TO_CREATE_SEMAPHORE_FOR_SIGNAL_WATCHER);
                                })
```
Cool & nifty:
```cpp
Semaphore::create(...).expect("Unable to create semaphore for signal watcher"); // with expects(...) doing exactly what you did
```

elBoberido:
@elfenpiff we could have the same for the `cxx::optional`. What do you think?

elfenpiff:
@elBoberido I like it!

elBoberido:
@elfenpiff the Rust `Option` and `Result` call this just `expect`, without the `s`. Maybe we should do the same?

PR #1324, opened 2022-04-08
iox-#996 add functional interface to expected

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
In the current `and_then`/`or_else` implementation is a bug which allows to provide callables which take a non-const reference even when the container is const. The new functional interface fixes this but here we encounter another problem. 

`std::function` and `cxx::function_ref` deduce always a const reference with `auto &` as lambda argument. This leads to compile error in all calls which use `and_then` and modify the given parameter reference. Most prominent example is the ice delivery publisher.
For more details, see: https://stackoverflow.com/questions/71797023/type-deduction-for-stdfunction-argument-types-with-auto-adds-const

Furthermore, the `expected` supported `and_then`/`or_else` calls with empty/invalid `cxx::function_refs`. This feature was added to the functional interface with tests.

## References

- See #996


iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_or_else.cpp
```diff
@@ -147,5 +148,78 @@ TYPED_TEST(FunctionalInterface_test, OrElseIsNotCalledWhenValid_ConstRValueCase)
...
+
+TYPED_TEST(FunctionalInterface_test, OrElseDoesNotCrashWithNullFunction_RValueCase)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "38ee906c-1314-440a-9328-976362555db0");
+    IOX_TEST_FUNCTIONAL_INTERFACE(OrElseDoesNotCrashWithNullFunction, std::move(sut));
+}
+
+TYPED_TEST(FunctionalInterface_test, OrElseDoesNotCrashWithNullFunction_ConstRValueCase)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "38ee906c-1314-440a-9328-976362555db0");
```
FerdinandSpitzschnueffler, line 221: We need the CI to check the test ids... This one is copied from above.

iceoryx_examples/icedelivery/iox_publisher.cpp
```diff
@@ -54,7 +54,7 @@ int main()
         //  * Sample can be held until ready to publish.
         //  * Data is default constructed during loan
         publisher.loan()
-            .and_then([&](auto& sample) {
+            .and_then([&](auto& sample) -> void {
```
FerdinandSpitzschnueffler, line 57: Probably you have to rerun Geoffrey.

FerdinandSpitzschnueffler, line 57: If we agree on the trailing return type (instead of a template approach or ???), can you please add a FAQ entry about why this is necessary?

iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_or_else.cpp
mossmaurice, line 221: Duplicate test id.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/expect_no_death.hpp
```diff
@@ -0,0 +1,48 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IOX_HOOFS_TESTUTILS_EXPECT_NO_DEATH_HPP
+#define IOX_HOOFS_TESTUTILS_EXPECT_NO_DEATH_HPP
+
+/// @brief gtest offers EXPECT_DEATH as test but no alternative to test the opposite, that
+///        the program does not terminate.
+///        EXPECT_NO_DEATH offers to verify that something does not terminate.
+/// @param[in] callable The callable which should not lead to termination
+#define EXPECT_NO_DEATH(callable)                                                                                      \
```
mossmaurice, line 24: I like it! :smile:

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/expect_no_death.hpp
```diff
@@ -0,0 +1,48 @@
...
+    EXPECT_EXIT(                                                                                                       \
+        {                                                                                                              \
+            callable();                                                                                                \
+            fprintf(stderr, "callable did not terminate");                                                             \
```
mossmaurice, line 28: You could use `std::cerr` here, would be more idiomatic C++.

iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_types.cpp
```diff
@@ -97,4 +97,66 @@ OptionalFactory::Type OptionalFactory::createInvalidObject() noexcept
...
+/// BEGIN ExpectedValueErrorFactory
+//////////////////////////////////
+ExpectedValueErrorFactory::value_t ExpectedValueErrorFactory::usedTestValue = 11;
+ExpectedValueErrorFactory::value_t ExpectedValueErrorFactory::anotherTestValue = 12;
+ExpectedValueErrorFactory::error_t ExpectedValueErrorFactory::usedErrorValue = 13;
+ExpectedValueErrorFactory::error_t ExpectedValueErrorFactory::anotherErrorValue = 14;
+
+void ExpectedValueErrorFactory::configureNextTestCase() noexcept
+{
+    // we increment all the values with an arbitrary value in every test case
```
mossmaurice, line 111: The values are not arbitrary. Did you initially meant to use random values here?

mossmaurice (changes requested):
Thanks a bunch for taking the time to create the question on Stackoverflow! I think using the generic `FunctionalInterface` struct is definitely a big improvent in order to avoid code and test duplication :tada: 

As this changes the behaviour on the high-level user API, we have to be very careful with this PR. Users might write the following code:

```
publisher.loan()
   .and_then([&](auto& sample) {
       sample = fooValue;
   }
```

Makes GCC output the  `error: assignment of member ‘SampleType’ in read-only object`, because the most cv-qualified overload is chosen, which is unintuitive for adding `[&](auto& sample) -> void`. If we chose this path, we should highlite it in our docs. My feeling is even if we document it well, bug reports might be dropping in about this issue.

I see two possible alternative to avoid this pitfall for the user:

1. Try the approach of templatizing the non-template methods, make them participate in SFINAE to prevent the need for adding the trailing return type syntax `(auto& foo) -> void` to the lambda, see [here](https://stackoverflow.com/questions/71797023/type-deduction-for-stdfunction-argument-types-with-auto-adds-const#comment126926381_71798130).
1. Narrow down the number of overloads excluding the const variant of `and_then` by providing several `FunctionalInterface` impls. The const-less `FunctionalInterface` one would be used just for the user API.  The compiler would then have not other choice, because the most cv-qualified overload would be the const-less one.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/expect_no_death.hpp
elfenpiff: I explicitly did not choose `std::cerr`. The reason is that `cerr` can be rerouted to another stream, or surpressed with statements like `internal::CaptureStderr()` something which would then lead to a `EXPECT_NO_DEATH` failure since the output was captured.

With `fprintf` this is not out of the box possible and therefore it should be more sustainable.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/expected.hpp
```diff
@@ -446,20 +392,6 @@ class IOX_NO_DISCARD expected<ValueType, ErrorType>
...
-    ///         expected which contains only an error
-    /// @code
-    ///     cxx::expected<int, int> someErrorProneFunction(){}
-    ///
-    ///     cxx::expected<int> isItSuccessful() {
-    ///         return someErrorProneFunction();
-    ///     }
-    /// @endcode
-    template <typename T>
-    operator expected<T>() noexcept;
```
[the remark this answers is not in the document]
elfenpiff: Since it returns always a copy. And if we return a copy a non const overload produces just code bloat since the logic is identical to the const version.

You can always get a copy independent of the constness of the object.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/expect_no_death.hpp
mossmaurice: Okidoki, please docment this in a little, cute comment above ;)

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/functional_interface.inl
```diff
@@ -111,7 +113,7 @@ inline Derived& AndThenWithValue<Derived, ValueType>::and_then(const and_then_ca
 {
```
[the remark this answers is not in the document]
elfenpiff: This was a discussion we had in the initial functional interface PR.

When we would use the `cxx::string` or `char`-array approach we would have to make this function a template so that it captures the size. You can have a wide variety of messages. We could also decide for a very long string but then we would waste stack memory in most cases.
Furthermore with the `cxx::string` template approach we would have to provide the size since the argument is most likely a string literal and not a `cxx::string`.

The user would have to write:
```cpp
myOptional.expect<11>("hello world"); // somehow ugly
myOptional.expect(someString); // most of the time it will not be used this way but here we could get rid of the template argument
```
-> we concluded that the `cxx::string` are not really fitting our use case.

The char array approach would look like:
```
struct Expect {
  template<uint64_t N>
  void expect( const char[N] msg);
};
```
With this we would cover all the string literal use cases and could write:
```cpp
myOptional.expect("hello world");
```
But now this causes a compile failure:
```cpp
myOptional.expect(someString); 
myOptional.expect(someString.c_str()); // also fails since we require the size as template argument and not a char ptr.
myOptional.expect<11>(stringWhichContainsHelloWorld.c_str()); // this may somehow work but this is evil
```
And think of all code which would be generated by the compiler, for all the length variations of `msg`.
-> so we concluded `char`-array is also not the way to go.

With the major use case in mind, that the user would like to do something like:
```cpp
SharedMemory shm = SharedMemoryBuilder::create().expect("unable to create shared memory");
```
And sometimes:
```cpp
SharedMemory shm = SharedMemoryBuilder::create().expect(myErrorMessage.c_str());
```
we decided then to follow the `const char* const` approach since it covers our major use case perfectly without having an unbearable API or generate a lot of code blaot.

iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_expect.cpp
```diff
@@ -72,15 +72,12 @@ TYPED_TEST(FunctionalInterface_test, ExpectDoesNotCallTerminateWhenObjectIsValid
...
 {
-    bool wasErrorHandlerCalled = true;
     SutType sut = FactoryType::createInvalidObject();
     {
-        auto handle = iox::ErrorHandlerMock::setTemporaryErrorHandler<iox::HoofsError>(
-            [&](auto, auto) { wasErrorHandlerCalled = true; });
-        callExpect(sut);
+        auto handle =
+            iox::ErrorHandlerMock::setTemporaryErrorHandler<iox::HoofsError>([&](auto, auto) { std::terminate(); });
+        EXPECT_DEATH(callExpect(sut), ".*");
```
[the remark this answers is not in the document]
elfenpiff: Sadly no. The reason is undefined behavior in a functional interface construct. The contract is either return a reference to the underlying value or die. And this has to be testable.

Assume you have an `optional` or an `expected` which does not contain a value.
Now we test `expect` for those objects. When they do not contain an error the expect has to terminate otherwise we have to dereference a nullptr or some garbage memory since they do not contain any value. So the only option is to terminate and check this.

Another alternative would of course be an exception in the custom user callback. This would make such code much more testable and would solve a lot of issues. But we have to remove `noexcept` before that in all methods and functions.

elfenpiff:
@mossmaurice @elBoberido @FerdinandSpitzschnueffler 

I solved the lambda auto type deduction now with templated Functors in `and_then`/`or_else`. This causes gcc to throw a `maybe-uninitialized` warning when the `m_data` member in the optional is not initialized with `{0}`.
This is a false positive but as the name states it is a `maybe`. In combination with `-Werror` this becomes deadly so I am disabling this warning for gcc in the cmake file.

iceoryx_dds/source/gateway/main.cpp
```diff
@@ -32,7 +32,7 @@ class ShutdownManager
   public:
     static void scheduleShutdown(int num)
     {
-        char reason;
+        char reason = '\0';
```
elfenpiff, line 35: Fun fact. gcc states here `maybe uninitialized` but detects this error only when `and_then` / `or_else` has a template functor argument.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -117,28 +117,32 @@ struct AndThenWithValue
     ///        happens.
     /// @param[in] callable will be called when valid
     /// @return reference to *this
-    Derived& and_then(const and_then_callback_t& callable) & noexcept;
+    template <typename Functor>
```
mossmaurice, line 120: Please add an `enable_if<cxx::is_invocable<Functor>` narrow down the set of candidates during overload resolution.

mossmaurice: A comment mentioning why this was templatized would be much appreciated :blush:

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -117,28 +117,32 @@ struct AndThenWithValue
     ///        happens.
     /// @param[in] callable will be called when valid
     /// @return reference to *this
-    Derived& and_then(const and_then_callback_t& callable) & noexcept;
```
mossmaurice, line 120: The alias in line 115 `and_then_callback_t` can be removed. Same for the ones in the struct `OrElseWithValue`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/function_ref.inl
```diff
@@ -29,7 +29,7 @@ inline function_ref<ReturnType(ArgTypes...)>::function_ref() noexcept
 template <class ReturnType, class... ArgTypes>
 template <typename CallableType, typename>
 inline function_ref<ReturnType(ArgTypes...)>::function_ref(CallableType&& callable) noexcept
-    : m_pointerToCallable(reinterpret_cast<void*>(std::addressof(callable)))
+    : m_pointerToCallable(const_cast<void*>(reinterpret_cast<const void*>(std::addressof(callable))))
```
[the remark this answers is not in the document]
mossmaurice: Good catch, thanks for fixing this!

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/functional_interface.inl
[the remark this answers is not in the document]
mossmaurice: Thanks for the detailed description!

There is a question which is still unanswered:

> What happens if the user passes a char array which is not null-terminated?

Let's assume the following

```
char myUnterminatedEvilString[3] = {'f','o','o'};

Foo::create().expect(myUnterminatedEvilString);
// std::cout reads the memory behind myUnterminatedEvilString till a '\0' is found
```

This is a major security problem, because we read the consecutive memory aka out-of-bounds access or am I mistaken here? That exactly the kind of cases where `std::span` should be used in the signature to pass the size alongside the pointer.

Instead of a generic `std::span` we could also make use of `std::string_view` in the signature of `expects(..)`. I know these two are not available in C++14 but I would be happy to backport them to C++14 :blush:

iceoryx_hoofs/test/moduletests/test_cxx_functional_interface_types.cpp
mossmaurice: What about `s/arbitrary value/static arbitrary value` then? This would make it clearer for me.

mossmaurice (commented):
Please adapt the reviewer hints in the PR description. It still mentions the old `[](auto& foo)-> void` changes. Also, please add the issue number under `## References`.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
elfenpiff: `enable_if` is ugly and has the effect that the user get even uglier template error messages - we should avoid this.

I added a `static_assert<cxx::is_invocable...` which has the same effect but in readable.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
elfenpiff: Oh I still need them. The old expected/optional had the requirement to work with unset `cxx::function_ref`, therefore I still have to check this in `and_then`/`or_else`.
And for that I require to convert the callable into the `function_ref` type - if it was already a `function_ref` which was unset I can use the bool operator to check if it is valid and only call it then.

The problem is the functor could be also a lambda which does not have a bool operator and then we cannot perform the check without a compile time error.
To have the same behavior we sadly have to go this way.

But again, another good reason why `cxx::function` & `cxx::function_ref` should not be nullable. It makes the code more error prone and in this case even slower.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/functional_interface.inl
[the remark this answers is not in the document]
elBoberido: I would opt for option 2 in combination with option 1 in order to prevent code bloat.
```
public:
template<uint64_t N>
... expect(const char message[N]) // supports use case 1
{
    return expect(message);
}

template<uint64_t N>
... expect(const cxx::string<N> & message) // supports use case 2
{
    return expect(message.c_str());
}
private:
... expect(const char* message, const uint64_t strSize);
```
There is still monopolization, but it should be optimized out and therefore not increase the binary size.

iceoryx_hoofs/cmake/IceoryxPlatform.cmake
```diff
@@ -44,6 +44,20 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Cla
...
+##    not executed at run time.
+## Here are very hard to debug false positives possible when the compiler is unable
+## to prove that a certain variable is initialized. This warning does not mean
+## that the variable is used uninitialized. It states that the variable is `maybe`
+## used uninitialized.
+## In combination with -Werror this leads to compile error and the only fix is
+## to assign already initialized variables again a value which can be very costly.
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    set(ICEORYX_WARNINGS PRIVATE ${ICEORYX_WARNINGS} -Wno-maybe-uninitialized)
+endif()
```
elBoberido, line 59: I downloaded the gcc 11.2 sources (latest release) and built them on ubuntu 20.04 but I don't get any warning when I remove these lines

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/functional_interface.inl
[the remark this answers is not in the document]
elfenpiff: So I choose secret option number 4.
```
template <typename StringContainer>
ValueType& expect(const StringContainer& msg) & noexcept {
  static_assert((std::is_array<StringContainer>::value
               && std::is_same<typename std::remove_all_extents<StringContainer>::type, char>::value)
                  || is_iox_string<StringContainer>::value,
              "Only char arrays and iox::cxx::strings are allowed as message type.");
//...
}
```

I allow every type and restrict it to either a char array or a `cxx::string` in the static assert so the developer gets a nice compiler failure.

I refrained from the overload once with char and once with string since expect already has 4 overloads (lvalue, rvalue and const variations). So adding two additional expect methods once with char array and once with string would produce in the end 12 additional methods.

iceoryx_hoofs/cmake/IceoryxPlatform.cmake
elBoberido: ... and after disabling the sanitizer I also get the error but only for the introspection client ... something is fishy with the client. I'll investigate that

elBoberido: Okay, I have a minimal example
https://github.com/ApexAI/iceoryx/blob/weird-maybe-uninitialized-warning/tools/introspection/source/introspection_app.cpp#L44:L68
While disabling the warning for iceoryx, this doesn't fix the problem since someone else could build with this warning enabled and encounter the same issue.

We either have to
1. understand why this happens
2. suppress the warning where it occurs
3. disable the warning in the introspection client itself and let all our user do the same if they encounter the problem

I'm for 1 and/or 2.

elBoberido: This would be a "fix" ... at least there is no warning anymore, even though it was most probably a false positive
https://github.com/ApexAI/iceoryx/commit/174f9bee476ba53e3a5241bfc05090d7e0c81acc

elfenpiff: @elBoberido I am against 1. The reason is that we actually do not suppress the cause of that error but symptoms.

My investigation showed that the source of the error is an uninitialized `m_data` member in the `optional` but the error occurred in some lines of the `function_ref` which was contained in the optional.

So we would suppress the error in the `function_ref` which is not the cause of the error and when you think longer about this it is not unlikely that we have other constructs which are contained in some kind of `optional` which also encounter `maybe-uninitialized` warnings.
So we ignore the warning then all over the place but not in this one single origin which causes it.

We also have a third option:
3. Initialize `m_data` in the `optional` with zero.

But I am against this one. Just think about the overhead when the object inside the optional is huge then we have a useless initialization of memory just to fix a broken compiler warning.

I would vote for 2.

elfenpiff: @elBoberido 
> This would be a "fix" ... at least there is no warning anymore, even though it was most probably a false positive
ApexAI@174f9be

I tried it out and also adapted in for the `or_else` path since we potentially have the same issue there. It fixed the issue so really thank you for your effort! This helped a lot and is definitely the cleanest and best solution.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -53,7 +60,13 @@ struct Expect
     /// @brief Expects that the object is valid, otherwise the method prints the
     ///        provided message and induces a fatal error
     /// @param[in] msg Message which will be printed when the object is invalid
-    void expect(const char* const msg) const noexcept;
+    // void expect(const char* const msg) const noexcept;
```
FerdinandSpitzschnueffler, line 63: Can this be removed?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -112,33 +129,42 @@ struct AndThenWithValue
     using and_then_callback_t = cxx::function_ref<void(ValueType&)>;
     using const_and_then_callback_t = cxx::function_ref<void(const ValueType&)>;
 
+    /// @note and_then has a template argument since otherwise we encounter issue
+    ///       with the type deduction and constness of auto arguments. A detailed
+    ///       discussion can be found here:
+    ///       https://stackoverflow.com/questions/71797023/type-deduction-for-stdfunction-argument-types-with-auto-adds-consthttps://stackoverflow.com/questions/71797023/type-deduction-for-stdfunction-argument-types-with-auto-adds-const
```
FerdinandSpitzschnueffler, line 135: It seems that you pasted the link twice.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -177,33 +203,42 @@ struct OrElseWithValue
     using or_else_callback_t = cxx::function_ref<void(ErrorType&)>;
     using const_or_else_callback_t = cxx::function_ref<void(const ErrorType&)>;
 
+    /// @note and_then has a template argument since otherwise we encounter issue
+    ///       with the type deduction and constness of auto arguments. A detailed
+    ///       discussion can be found here:
+    ///       https://stackoverflow.com/questions/71797023/type-deduction-for-stdfunction-argument-types-with-auto-adds-consthttps://stackoverflow.com/questions/71797023/type-deduction-for-stdfunction-argument-types-with-auto-adds-const
```
FerdinandSpitzschnueffler, line 209: same here

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -177,33 +203,42 @@ struct OrElseWithValue
     using or_else_callback_t = cxx::function_ref<void(ErrorType&)>;
     using const_or_else_callback_t = cxx::function_ref<void(const ErrorType&)>;
 
+    /// @note and_then has a template argument since otherwise we encounter issue
```
FerdinandSpitzschnueffler, line 206:
```suggestion
    /// @note or_else has a template argument since otherwise we encounter issue
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
```diff
@@ -126,6 +127,34 @@ struct is_function_pointer<ReturnType (*)(ArgTypes...)> : std::true_type
 {
 };
 
+template <uint64_t Capacity>
+class string;
+
+/// @brief struct to check whether an argument is a char array
+template <typename T>
+struct is_char_array
```
FerdinandSpitzschnueffler, line 135: Now that you moved it, I recognize that I never wrote tests for this... :/

mossmaurice (commented):
Could you please address the following two points?

* Please squash your commits together with a rebase, as this PR had several different attempts of solving the problem
* Rework the PR description, it still mentions adding `[](auto& foo)-> void{}`

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/functional_interface.inl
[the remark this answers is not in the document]
elfenpiff: Sadly there is no short way.

It is possible to implement something which is somehow similar to c++20 concepts with c++14 but this requires a lot of more code and the PR is already big enough.

Since the user cannot compile this and gets a sane error message right from the beginning and even my IDE (neovim) realizes that and prints me an error I can say we with good conscience that we are save here!

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
elfenpiff: @mossmaurice I think the convertible check is superfluous since I already check if this thing is invocable as required - the exact same check you perform in the constructor of the function ref.

Furthermore, it is a requirement of the `function_ref` that it is constructible from any callable with the defined signature.

And finally, when the `function_ref` & `function` is no longer nullable we also can get rid of those type aliases and the check inside `and_then` and `or_else`. Then the code become easier and more performant.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/functional_interface.hpp
```diff
@@ -53,7 +60,13 @@ struct Expect
...
     ///        provided message and induces a fatal error
     /// @param[in] msg Message which will be printed when the object is invalid
-    void expect(const char* const msg) const noexcept;
+    // void expect(const char* const msg) const noexcept;
+
+    /// @brief Expects that the object is valid, otherwise the method prints the
+    ///        provided message and induces a fatal error
+    /// @param[in] msg Message which will be printed when the object is invalid
+    template <typename StringContainer>
+    void expect(const StringContainer& msg) const noexcept;
```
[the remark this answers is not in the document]
elfenpiff: `String` and `string` look very similar in text and I wanted to make sure that someone does not mix it up and thinks we are using a `string` class here but sees right away that this is the template parameter.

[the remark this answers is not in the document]
mossmaurice: What about just `T` or `StringType`?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
```diff
@@ -126,6 +127,34 @@ struct is_function_pointer<ReturnType (*)(ArgTypes...)> : std::true_type
...
+struct is_cxx_string
+{
+    static constexpr bool value = false;
+};
+
+template <uint64_t N>
+struct is_cxx_string<::iox::cxx::string<N>>
+{
+    static constexpr bool value = true;
+};
```
FerdinandSpitzschnueffler, line 157:
```suggestion
};

```

iceoryx_hoofs/test/moduletests/test_cxx_type_traits.cpp
```diff
@@ -170,4 +171,41 @@ TEST(TypeTraitsTest, IsFunctionPointerResolvesToFalse)
     EXPECT_FALSE(result);
 }
 
+TEST(TypeTraitsTest, NonCharArraysAreIdentifiedCorrectly)
```
FerdinandSpitzschnueffler, line 174: Thanks for adding the tests! Can you please update the copyright year?

FerdinandSpitzschnueffler (dismissed):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/functional_interface.inl
```diff
@@ -37,8 +43,12 @@ inline void Expect<Derived>::expect(const char* const msg) const noexcept
 }
 
 template <typename Derived, typename ValueType>
-inline ValueType& ExpectWithValue<Derived, ValueType>::expect(const char* const msg) & noexcept
+template <typename StringType>
+inline ValueType& ExpectWithValue<Derived, ValueType>::expect(const StringType& msg) & noexcept
 {
+    static_assert(is_char_array<StringType>::value || is_cxx_string<StringType>::value,
```
mossmaurice, line 49: Why didn't you add the `static_assert` to the other methods? E.g. `ExpectWithValue<Derived, ValueType>::expect(const StringType& msg)`

elfenpiff: Since the other methods are forwarding the arguments to this method to reduce the code duplication and implement this only once.

So it is always verified and adding this to all methods would just increase the code duplication.

mossmaurice (approved):

FerdinandSpitzschnueffler (approved):

PR #1331, opened 2022-04-11
iox-#208 add insert method

### LINKED ISSUE #208: Additional functionality for `cxx::string`

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
Unlike the `std::string` `insert` method, a boolean is returned to indicate whether the insert was successful. Furthermore, insert will fail when `count` is greater than the size of the passed string.

## References

- Partially closes #208

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
@@ -450,6 +450,19 @@ class string
     typename std::enable_if<is_char_array<T>::value || is_cxx_string<T>::value, bool>::type
     unsafe_append(const T& t) noexcept;
 
+    /// @brief inserts a cxx:string or char array in the range [str[0], str[count]) at position pos. The insertion fails
+    /// if the string capacity would be exceeded or pos is greater than the string size or count is greater than the
+    /// string to be inserted.
+    ///
+    /// @param [in] pos position at which the string shall be inserted
+    /// @param [in] str the cxx::string or char array to be inserted
+    /// @param [in] count number of characters to be inserted
+    ///
+    /// @return true if the insertion was successful, otherwise false
+    template <typename T>
+    typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
+    insert(const uint64_t pos, const T& str, const uint64_t count) noexcept;
+
     /// @brief creates a substring containing the characters from pos until count; if pos+count is greater than the size
     /// of the original string the returned substring only contains the characters from pos until size();
     /// iox::cxx::nullopt is returned if pos is greater than the size of the original string;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
@@ -460,7 +460,7 @@ class string
     ///
     /// @return true if the insertion was successful, otherwise false
     template <typename T>
-    typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
+    constexpr typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
     insert(const uint64_t pos, const T& str, const uint64_t count) noexcept;
 
     /// @brief creates a substring containing the characters from pos until count; if pos+count is greater than the size
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
@@ -460,7 +460,7 @@ class string
     ///
     /// @return true if the insertion was successful, otherwise false
     template <typename T>
-    constexpr typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
+    constexpr typename std::enable_if<is_char_array<T>::value || is_cxx_string<T>::value, bool>::type
     insert(const uint64_t pos, const T& str, const uint64_t count) noexcept;
 
     /// @brief creates a substring containing the characters from pos until count; if pos+count is greater than the size
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
@@ -511,10 +511,6 @@ string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count)
     {
         return false;
     }
-    if (pos > m_rawstringSize)
-    {
-        return false;
-    }
     const auto new_size = m_rawstringSize + count;
     if (new_size > Capacity)
     {
@@ -522,9 +518,30 @@ string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count)
     }
 
     auto remainder = substr(pos);
+    if (!remainder)
+    {
+        return false;
+    }
 
+    // ignored for gcc since the code above ensures that there is no out of bounds access:
+    // Capacity >= m_rawstringSize + count and m_rawstringSize >= pos,
+    // therefore Capacity >= pos + count
+    //
+    // if it should be enabled again, the following code can be uncommented to suppress the warning
+    //
+    // if (Capacity < pos + count)
+    // {
+    //   return false;
+    // }
+#if (defined(__GNUC__) && __GNUC__ >= 7 && !defined(__clang__))
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#endif
     std::memcpy(&(m_rawstring[pos]), internal::GetData<T>::call(str), count);
     std::memcpy(&(m_rawstring[pos + count]), remainder->c_str(), remainder->size());
+#if (defined(__GNUC__) && __GNUC__ >= 7 && !defined(__clang__))
+#pragma GCC diagnostic pop
+#endif
 
     m_rawstring[new_size] = '\0';
     m_rawstringSize = new_size;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
@@ -516,6 +516,10 @@ string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count)
     {
         return false;
     }
+    if (Capacity < pos + count)
+    {
+        return false;
+    }
 
     auto remainder = substr(pos);
     if (!remainder)
@@ -523,25 +527,8 @@ string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count)
         return false;
     }
 
-    // ignored for gcc since the code above ensures that there is no out of bounds access:
-    // Capacity >= m_rawstringSize + count and m_rawstringSize >= pos,
-    // therefore Capacity >= pos + count
-    //
-    // if it should be enabled again, the following code can be uncommented to suppress the warning
-    //
-    // if (Capacity < pos + count)
-    // {
-    //   return false;
-    // }
-#if (defined(__GNUC__) && __GNUC__ >= 7 && !defined(__clang__))
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Warray-bounds"
-#endif
     std::memcpy(&(m_rawstring[pos]), internal::GetData<T>::call(str), count);
     std::memcpy(&(m_rawstring[pos + count]), remainder->c_str(), remainder->size());
-#if (defined(__GNUC__) && __GNUC__ >= 7 && !defined(__clang__))
-#pragma GCC diagnostic pop
-#endif
 
     m_rawstring[new_size] = '\0';
     m_rawstringSize = new_size;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
@@ -503,6 +503,36 @@ string<Capacity>::append(TruncateToCapacity_t, const T& t) noexcept
 }
 
 template <uint64_t Capacity>
+template <typename T>
+inline typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
+string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count) noexcept
+{
+    if (count > internal::GetSize<T>::call(str))
+    {
+        return false;
+    }
+    if (pos > m_rawstringSize)
+    {
+        return false;
+    }
+    const auto new_size = m_rawstringSize + count;
+    if (new_size > Capacity)
+    {
+        return false;
+    }
+
+    auto remainder = substr(pos);
+
+    std::memcpy(&(m_rawstring[pos]), internal::GetData<T>::call(str), count);
+    std::memcpy(&(m_rawstring[pos + count]), remainder->c_str(), remainder->size());
+
+    m_rawstring[new_size] = '\0';
+    m_rawstringSize = new_size;
+
+    return true;
+}
+
+template <uint64_t Capacity>
 inline optional<string<Capacity>> string<Capacity>::substr(const uint64_t pos, const uint64_t count) const noexcept
 {
     if (pos > m_rawstringSize)
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
@@ -504,7 +504,7 @@ string<Capacity>::append(TruncateToCapacity_t, const T& t) noexcept
 
 template <uint64_t Capacity>
 template <typename T>
-inline constexpr typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
+inline constexpr typename std::enable_if<is_char_array<T>::value || is_cxx_string<T>::value, bool>::type
 string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count) noexcept
 {
     if (count > internal::GetSize<T>::call(str))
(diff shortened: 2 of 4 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -503,6 +503,36 @@ inline
...
+{
+    if (count > internal::GetSize<T>::call(str))
+    {
+        return false;
+    }
+    if (pos > m_rawstringSize)
+    {
+        return false;
+    }
+    const auto new_size = m_rawstringSize + count;
```
elfenpiff, line 519: I think you can get rid of the two previous early returns with this return. This should cover everything.

FerdinandSpitzschnueffler: Unfortunately no. `count` could still be greater than string to be inserted. This is allowed in `std::string` but I decided against because it's undefined what will be inserted. `pos` could be greater than `m_rawStringSize` too. Since this is tested in `substr`, I moved the return after creating the `remainder` substring.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,190 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+    ASSERT_TRUE(this->testSubject.insert(0, "Ferdinand", 0));
+    EXPECT_THAT(this->testSubject.size(), Eq(0));
+    EXPECT_THAT(this->testSubject.c_str(), StrEq(""));
+}
+
+TEST(String10, InsertStringLiteralAtPositionGreaterStringSizeFails)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "a3a30ea5-f228-48f4-9497-08502e4f2c9a");
+    string<10> sut("Muesli");
+    ASSERT_FALSE(sut.insert(7, "s", 1));
```
mossmaurice, line 2592: Could you use a `pos` way larger than `7` here? My inner chicken always fears an of-by-one-error is these cases. Maybe this is yet another test case.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,190 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+    EXPECT_THAT(sut.c_str(), StrEq("Muesli"));
+}
+
+TEST(String10, InsertCxxStringAtTheBeginningOfTheStringSucceeds)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "df62df30-0838-4282-91f7-b39b1c861861");
+    string<10> sut("toad");
+    string<10> string_to_insert("Hypno");
+    ASSERT_TRUE(sut.insert(0, string_to_insert, 5));
+    EXPECT_THAT(sut.size(), Eq(9));
```
mossmaurice, line 2603: Why not?

```suggestion
    EXPECT_THAT(sut.size(), Eq(sut.size() + string_to_insert.size()));
```

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,190 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+    ASSERT_TRUE(this->testSubject.insert(0, string_to_insert, 0));
+    EXPECT_THAT(this->testSubject.size(), Eq(1));
+    EXPECT_THAT(this->testSubject.c_str(), StrEq("M"));
+}
+
+TYPED_TEST(stringTyped_test, InsertCxxStringWithCountGreaterThanSizeOfStringLiteralFails)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "9bd29af2-12c8-4dc1-8ee8-82610f1bb9a6");
+    string<1> string_to_insert("M");
+    ASSERT_FALSE(this->testSubject.insert(0, string_to_insert, 2));
```
mossmaurice, line 2650:
```suggestion
    ASSERT_FALSE(this->testSubject.insert(0, string_to_insert, string_to_insert().size() + 1));
```

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,190 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+
+TEST(String10, InsertCxxStringAtPositionGreaterStringSizeFails)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "a2517ea0-2842-4f56-a89a-b7fd01c6c6ba");
+    string<10> sut("Muesli");
+    string<1> string_to_insert("s");
+    ASSERT_FALSE(sut.insert(7, string_to_insert, 1));
+    EXPECT_THAT(sut.size(), Eq(6));
+    EXPECT_THAT(sut.c_str(), StrEq("Muesli"));
+}
```
mossmaurice, line 2692: You added test cases for string literals and `cxx::string`. It would make sense to add the same test cases also for `std::string`, right?

mossmaurice (commented):
Thanks for improving the `cxx::string`! LGTM in general. Left one change request and a few nitpicks.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,190 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+
+TEST(String10, InsertStringLiteralAtPositionGreaterStringSizeFails)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "a3a30ea5-f228-48f4-9497-08502e4f2c9a");
+    string<10> sut("Muesli");
+    ASSERT_FALSE(sut.insert(7, "s", 1));
+    EXPECT_THAT(sut.size(), Eq(6));
+    EXPECT_THAT(sut.c_str(), StrEq("Muesli"));
+}
+
```
mossmaurice, line 2596: What about a further test case `InsertStringLiteralWithCountPlusSizeLargerThanCapacity`?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -450,6 +450,19 @@ class string
...
+    /// string to be inserted.
+    ///
+    /// @param [in] pos position at which the string shall be inserted
+    /// @param [in] str the cxx::string or char array to be inserted
+    /// @param [in] count number of characters to be inserted
+    ///
+    /// @return true if the insertion was successful, otherwise false
+    template <typename T>
+    constexpr typename std::enable_if<internal::IsCharArray<T>::value || internal::IsCxxString<T>::value, bool>::type
+    insert(const uint64_t pos, const T& str, const uint64_t count) noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @mossmaurice The std::string method you proposed behaves differently. It inserts count copies of a passed character at the position pos. My implementation follows method 3 of https://en.cppreference.com/w/cpp/string/basic_string/insert which inserts characters in the range [str[0], str[count]) at position pos.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
FerdinandSpitzschnueffler: Since I want to exclude an off-by-one error in the implementation, I chose a position of `sut.size() + 1` on purpose. I will make this clearer in the code.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
FerdinandSpitzschnueffler: Because `sut.size()` has already changed. But I will add an `expectedString` as you proposed above.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
FerdinandSpitzschnueffler: The current implementation doesn't allow (via enable_if) to insert std::strings and I tend to keep it like this. @mossmaurice What do you think?

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
FerdinandSpitzschnueffler: This is tested in `InsertTooLargeStringLiteralFails`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -502,6 +502,40 @@ string<Capacity>::append(TruncateToCapacity_t, const T& t) noexcept
...
+    auto remainder = substr(pos);
+    if (!remainder)
+    {
+        return false;
+    }
+
+    std::memcpy(&(m_rawstring[pos]), internal::GetData<T>::call(str), count);
+    std::memcpy(&(m_rawstring[pos + count]), remainder->c_str(), remainder->size());
+
+    m_rawstring[new_size] = '\0';
```
MatthiasKillat, line 533: Not sure how the static checks (that fail in the pipeline with a specific gcc) work but we could move setting the zero-terminator before the memcpy operations. The static checks should not really know about where the zero-terminator will be (maybe that it directly follows the remaineder copy, but definitely not the exact position as it is runtime info).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -502,6 +502,40 @@ string<Capacity>::append(TruncateToCapacity_t, const T& t) noexcept
...
+    }
+
+    auto remainder = substr(pos);
+    if (!remainder)
+    {
+        return false;
+    }
+
+    std::memcpy(&(m_rawstring[pos]), internal::GetData<T>::call(str), count);
+    std::memcpy(&(m_rawstring[pos + count]), remainder->c_str(), remainder->size());
```
MatthiasKillat, line 531: Given the numbers in the warnings m = 18446744073709551615 = max 64 bit integer, B = 9223372036854775807 = m/2  
and reading

https://archive.fosdem.org/2017/schedule/event/valgrind_fortification/attachments/slides/1868/export/events/attachments/valgrind_fortification/slides/1868/valgrind_fortify.html#slide6

https://gcc.gnu.org/onlinedocs/gcc/Object-Size-Checking.html

I have the following theory.

1) gcc tries to determine the  size of the destination buffer, but cannot really bound it (insufficient static information due to pointer arithmetics etc). 
2) It then assumes the bound B by default (see above, whether this is smart or not I cannot say, but in general it is a large number).
3) It also only knows it has to copy count bytes to the buffer and assumes count = m in the worst case.
4) since M > B it issues a warning.

To test this, we could (temporarily) create 
`uint32_t c = (uint32_t) count;` and `uint32_t r = (uint32_t) remainder->size()` and only copy c and r bytes, respectively.

For our tests the casts are ok, but generally they are not. However, it is questionable whether the size (internally `m_rawSize` and return value of `size()` really should be 64 bit integers. I would assume we will not really have strings with >2^32 characters, especially in a static memory setup.

TLDR: the implemented logic is correct from what I can see, but certain compilers fail to detect this due to the way they determine worst case sizes and the type they use to store these sizes.

@FerdinandSpitzschnueffler

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -502,6 +502,40 @@ string<Capacity>::append(TruncateToCapacity_t, const T& t) noexcept
...
+    if (new_size > Capacity)
+    {
+        return false;
+    }
+    if (Capacity < pos + count)
+    {
+        return false;
+    }
+
+    auto remainder = substr(pos);
```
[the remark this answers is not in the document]
MatthiasKillat: Actually it would be https://en.cppreference.com/w/cpp/algorithm/copy_backward ...

MatthiasKillat:
As for string.inl:550:26: error: array subscript 18446744073709551615 is above array bounds of ‘char [11]’ [-Werror=array-bounds]
  550 |     subString.m_rawstring[length] = '\0';

it could likely be fixed by obscuring the destination of the write using a volatile pointer like so

```
volatile char* p =  &subString.m_rawstring[length];
*p = `\0`;
```
Previosly the compiler assumed it could write to any 64 bit integer value length value here, but actually the size of the arry is known at compile time so it will complain. I consider this a bug due to very overeager reaction, if the compiler does not know enough it should stay silent (but some versions apparently do not).

With volatile the compiler cannot infer anymore what the size of the buffer is, hence it will not complain anymore. Without `volatile` it still might, as there is no aliasing of p (still worth a try if we dislike volatile).

However, this may reduce optimization potential to some extent (only minor here I think). Something has to give here though.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,203 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+    const string<1> expectedString("M");
+    ASSERT_TRUE(this->testSubject.insert(0, "M", 1));
+    EXPECT_THAT(this->testSubject.size(), Eq(expectedString.size()));
+    EXPECT_THAT(this->testSubject, Eq(expectedString));
+}
+
+TYPED_TEST(stringTyped_test, InsertEmptyStringLiteralDoesNotChangeTheString)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "37d723b7-5b35-4f1d-ac33-23dd483121a1");
+    this->testSubject = "M";
```
MatthiasKillat, line 2552: Sometimes we use a member as sut and sometimes we create a local sut. This is somewhat inconsistent but ok for now. We should strive to be consistent in future tests. I propose using a local sut if the setup is simple and a member if it is complex.

Nothing to do here, carry on.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,203 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+    ASSERT_FALSE(this->testSubject.insert(0, "M", 2));
+    EXPECT_THAT(this->testSubject.size(), Eq(0));
+    EXPECT_THAT(this->testSubject.c_str(), StrEq(""));
+}
+
+TEST(String10, InsertTooLargeStringLiteralFails)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "0f11f387-f437-4a39-8207-259df4807c70");
+    const string<10> expectedString("Ferdinand");
+    string<10> sut(expectedString);
```
MatthiasKillat, line 2570: Ok, resulting string would exceed capacity.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -2504,4 +2504,203 @@ TYPED_TEST(stringTyped_test, AccessMaxPositionOfNotEmptyStringViaConstSubscriptO
...
+    string<10> sut("Ferdinand");
+    ASSERT_TRUE(sut.insert(sut.size(), "Spitzschnueffler", 1));
+    EXPECT_THAT(sut.size(), Eq(expectedString.size()));
+    EXPECT_THAT(sut, Eq(expectedString));
+}
+
+TYPED_TEST(stringTyped_test, InsertStringLiteralWithCount0DoesntChangeTheString)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "90e80302-1c14-48bb-b9e9-0f8fca017fea");
+    ASSERT_TRUE(this->testSubject.insert(0, "Ferdinand", 0));
```
MatthiasKillat, line 2589: Maybe try insert in the middle, e.g. at 2 or something.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
MatthiasKillat: I think we cannot use `c_str()` of `std::string` explicitly if we need to insert a string (it does not return a char array). Maybe for the better, so count cannot exceed the size of the inserted string.

MatthiasKillat (approved):
Looks good. Minor suggestions, no changes required.

elfenpiff (approved):

MatthiasKillat (approved):
Correct, memmove will handle overlap and copy backwards if required (and more efficient than a handcrafted loop). I remembered that at some point but I would imagine https://en.cppreference.com/w/cpp/algorithm/copy_backward will use it for primitive data...

PR #1356, opened 2022-05-11
Iox #208 add comparison between string and char

### LINKED ISSUE #208: Additional functionality for `cxx::string`

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
The compare method is now a template so that a cxx::string can be compared with another cxx::string (like before), a std::string or char array. Additionally, the comparison operators for std::string and char array were implemented. The comparison operators for char ptr have been removed. They once were added because of the implicit conversion from char ptr to std::string. This is now excluded via 'enable_if'. Furthermore, comparison operators for simple chars are added.

## References

- Partially closes #208


elfenpiff (commented):
Everything reviewed - except the test. They will follow soon.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -214,61 +214,222 @@ inline bool string<Capacity>::unsafe_assign(const std::string& str) noexcept
...
+{
+    auto result = memcmp(c_str(), &other, 1U);
+    if (result == 0)
     {
-        return 1;
+        if (m_rawstringSize > 1U)
+        {
+            return 1;
+        }
+        else
```
MatthiasKillat, line 355: The `< 1U` case would be the empty string case and for this we would always correctly compare -1 if `other != \0`. Hence it appears we can omit the case, but there is the following problem:

If `this` is the empty string (size 0) and `other` is  `\0` we would get result 0 and also return 0 which is wrong. I may miss something here but I am fairly certain.

Lexicographically the empty string is smaller than any character, in particular `\0` (which is not equal to the empty string). I think the logic is incorrect in this case and we need to properly handle the empty string. The easiest would be an initial size check (there may be more efficient ways in terms of control flow).

MatthiasKillat (commented):
Reviewed all but the tests. I will review the tests separately.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -214,61 +214,222 @@ inline bool string<Capacity>::unsafe_assign(const std::string& str) noexcept
...
+template <typename T>
+inline
+    typename std::enable_if<(is_cxx_string<T>::value || is_char_array<T>::value || std::is_same<T, std::string>::value),
+                            int64_t>::type
+    string<Capacity>::compare(const T& other) const noexcept
 {
-    uint64_t otherSize = other.size();
-    if (m_rawstringSize < otherSize)
+    uint64_t otherSize = internal::GetSize<T>::call(other);
+    auto result = memcmp(c_str(), internal::GetData<T>::call(other), std::min(m_rawstringSize, otherSize));
```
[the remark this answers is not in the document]
MatthiasKillat: One could argue it does not matter here if we do not have early `\0` characters in our string. I am not sure we allow those. If we do, `memcmp` is required as `strncmp` stops at the first `\0` character (or if the n characters where compared).

So they behave differently but it is only relevant if we can have a string with `\0` characters, otherwise they should behave the same and theoretically `memcmp` is better than as it will not check for `\0`.

TLDR: if my assumption regarding `\0` chars is correct, `memcpy` should be preferred for performance.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -301,93 +301,128 @@ class string
...
     ///
     /// @param [in] other is the string to compare with self
     ///
     /// @return an integer < 0 if the first character that does not match has a lower value in self than in other, 0 if
     /// the contents of both strings are equal, an integer > 0 if the first character that does not match has a greater
     /// value in self than in other
-    template <uint64_t N>
-    int64_t compare(const string<N>& other) const noexcept;
+    template <typename T>
+    typename std::enable_if<(is_cxx_string<T>::value || is_char_array<T>::value || std::is_same<T, std::string>::value),
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, that is what I meant by generating a specific type trait to bundle this.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -1089,153 +1090,634 @@ TYPED_TEST(stringTyped_test, CompareOperatorGreaterEqResultFalseWithDifferentCap
...
+}
+
+TYPED_TEST(stringTyped_test, CompareOperatorsWithDifferentCharWithDifferentSize)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "c9f2e472-3fbf-4e08-a0ee-6bb725e7c0d7");
+    using MyString = typename TestFixture::stringType;
+    constexpr auto STRINGCAP = MyString::capacity();
+
+    std::string temp1(STRINGCAP + 5U, 'M');
+    string<STRINGCAP + 5U> sut1;
```
elfenpiff, line 1641: Could we rename the sut into better understandable names. Since we compare them always with `testChar` I think the name `sutSmallerTestChar`, `sutEqualTestChar` and `sutGreaterTestChar` would help in understanding the single conditions you verify below.

And maybe we can create 3 `EXPECT_THAT` groups for smaller, equal and greater

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -1089,153 +1090,634 @@ TYPED_TEST(stringTyped_test, CompareOperatorGreaterEqResultFalseWithDifferentCap
...
+TYPED_TEST(stringTyped_test, CompareOperatorsWithDifferentChar)
 {
-    ::testing::Test::RecordProperty("TEST_ID", "99a40704-9d0b-49f9-bd5c-4c87b9f2ba00");
+    ::testing::Test::RecordProperty("TEST_ID", "c818b150-b926-4f3c-8405-6327303f12f6");
     using MyString = typename TestFixture::stringType;
     constexpr auto STRINGCAP = MyString::capacity();
-    string<STRINGCAP> testFixedString("M");
-    std::string testSTDString = "L";
-    EXPECT_THAT(testFixedString != testSTDString, Eq(true));
+    string<STRINGCAP> sut1("M");
```
elfenpiff, line 1594: could you rename this into `sutGreaterTestChar` and `sutSmallerTestChar` so that the `EXPECT_` condition can be easier reviewed.

elfenpiff: Please do this also for the tests above where it is applicable.

iceoryx_hoofs/test/moduletests/test_cxx_string.cpp
```diff
@@ -16,6 +16,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "iceoryx_hoofs/cxx/string.hpp"
+#include "iceoryx_hoofs/platform/platform_correction.hpp"
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Maybe not. This was just a try to satisfy the msvc, but I found another solution. I will remove it and we will see if the CI complains.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -214,61 +214,222 @@ inline bool string<Capacity>::unsafe_assign(const std::string& str) noexcept
...
+inline typename std::enable_if<(is_char_array<T>::value || std::is_same<T, std::string>::value), bool>::type
+operator>=(const T& lhs, const string<Capacity>& rhs) noexcept
+{
+    return (rhs.compare(lhs) <= 0);
+}
+
+template <uint64_t Capacity>
+inline int64_t string<Capacity>::compare(const char& other) const noexcept
+{
+    auto result = memcmp(c_str(), &other, 1U);
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Yes, I deliberately didn't add the check for an empty string but I forgot to implement a test for this case and will add it.

First I thought that `memcmp` is overkill, too. But the code is more readable without writing the comparison of one char and in case of equality add the size check.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
FerdinandSpitzschnueffler: Actually the first character of an empty string is `\0`. It is default constructed this way and it is also set in the clear method. 
For me, it makes sense to return 0 in this case. Furthermore, if you compare an empty std::string with `"\0"` (note that a char comparison is not allowed) it also returns 0.

MatthiasKillat: That may be the case for `std::string` and we can leave it like this. But in the theory of formal languages (https://en.wikipedia.org/wiki/Empty_string#:~:text=In%20formal%20treatments%2C%20the%20empty,%7C%CE%B5%7C%20%3D%200.) this makes no sense and teh STL is at odds with this.

We can leave it as is for STL consistency.

MatthiasKillat (approved):
Ok, I think the important test cases are covered. I think we have not tested all the different cases for a string to be lexicographically smaller than another though like 1) same length, some common prefix, different rest, 2) different length, common prefix, different rest etc. This gets combinatorially quite exhausting though.

I think this is also a case where we can (in theory) benefit from a generative testing approach (like I try to use for the ServiceDiscovery). The string is even better suited since the reference algorithm is basically in the `std::string`, so we do not even have to implement it ourselves. It is still complex to generate all the required test cases semi-automatically.

Ideally we would write tests for structures with STL  counterpart in a way like: 1) setup the test case, 2) verify that it behaves like the STL

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
FerdinandSpitzschnueffler: @MatthiasKillat @elfenpiff I reconsidered this case and decided to differ here from the std::string behavior. If an empty cxx::string is compared with '\0', compare will return -1.

elfenpiff (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -321,7 +321,11 @@ inline int64_t string<Capacity>::compare(char other) const noexcept
     auto result = memcmp(c_str(), &other, 1U);
     if (result == 0)
     {
-        if (m_rawstringSize > 1U)
+        if (empty())
+        {
+            return -1;
+        }
+        else if (m_rawstringSize > 1U)
```
MatthiasKillat, line 328: I would recommend avoiding this kind of `else` branching if there is an early return (which should be used as often as possible for easy to follow code flow and optimization).

No need for changes though and the compiler will optimize the jumps/returns itself in this case.

MatthiasKillat (approved):

Issue #1345, opened 2022-04-29
Refactor Logger


elBoberido:
## Brief feature description

Log messages cannot be forwarded to another logging framework. To make this possible, the logger must be refactored.

## Detailed information

When iceoryx is used with another framework it is not possible to forward log messages to the native logger of the framework.

The logger shall be refactored to be able to to switch the backend at compile time or at runtime.

## Tasks

- [ ] Design document
- [x] Implement/refactor logger
- [x] Create a custom logger to print the log only for failed tests

elfenpiff:
@elBoberido What do you think about dividing this issue in multiple sub issues? This issue is kept to keep track of the overall feature progress and every task gets its own issue which is linked in the task todo list here.

Then we can manage discussions at the right place and prevent comment cluttering in the main issue.

elBoberido:
@elfenpiff I'm a bit torn on this. For me the design document, implementation and tests belong together in one issue. Integration might be a different issue.

The question is also how this is handled when someone does not use subtasks? Is this person then not allowed to have multiple PRs for the same issue? Do we enforce to have subtasks? What happens when someone notices that it might be better to split the implementation in multiple PRs after the implementation started? What happens when the implementation started with a small refactoring and the it turned out to make sense to have the refactoring as separate PR?

There is in general also not much discussions going on in the issue itself but on the PRs.

From my point of view the only problem this would solve is that issues are closed when a PR is merged. Everything else should be handled by the developer working on the feature and as little as possible should be enforced.

Regarding this feature, I'd like to have the design document, implementation and tests on the same issue. The remaining tasks shouldn't could then be moved to a separate issue but currently I don't feel that it would improve my workflow or save me time and it also does not have a negative impact on the reviewer.

PR #1358, opened 2022-05-12
iox-#1345 Design document for logging

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

A proof of concept which implements the design of the PR is available at https://github.com/elBoberido/iceoryx/tree/logger-ng/iceoryx_hoofs/include/iceoryx_hoofs/log and https://github.com/elBoberido/iceoryx/blob/logger-ng/iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/logger.hpp.

The mechanism to replace the logger could also be applied to the error handler later on.

## References

- Relates to #1345


doc/design/diagrams/logging/logging_classes.puml
```diff
@@ -0,0 +1,85 @@
...
+
+enum LogLevel {
+    OFF = 0,
+    FATAL,
+    ERROR,
+    WARN,
+    INFO,
+    DEBUG,
+    TRACE,
+}
```
elBoberido, line 13: This is the same ordering like in the current implementation. Shall we follow swap the order to
```cpp
enum LogLevel {
    TRACE = 0,
    DEBUG,
    INFO,
    WARN,
    ERROR,
    FATAL,
    OFF,
}
```

doc/design/logging.md
```diff
@@ -0,0 +1,356 @@
...
+   passed tests but output them for failed tests. This keeps the noise low but
+   still provides useful information when needed. Additionally there shall be an
+   environment variable to circumvent the suppression when desired.
+
+### Solution
+
+#### Class diagram
+
+![logging class diagram](../website/images/logging_classes.svg)
+
```
elBoberido, line 85: It just occurred to me that I need to add an additional sequence diagram for a log message with a `LogStream` object. This will explain how the logger is obtained and replacing the logger at runtime is synced to other threads.

doc/design/logging.md
```diff
@@ -0,0 +1,356 @@
...
+In the example above `expensiveFunctionCall` is only executed when `cond` is `true`.
+If `cond` is a compile time constant set to `false`, the whole statement is compiled
+to a no-op and optimized away.
+
+This is the log macro with lazy evaluation
+
+```cpp
+#define IOX_LOG_INTERNAL(file, line, function, level)          \
+    if ((level) <= iox::log::Logger::minimalLogLevel()         \
+        && (iox::log::Logger::ignoreLogLevel()                 \
```
elBoberido, line 106: The `ignoreLogLevel` function would be deprecated if we do not allow to change to log level after `initLogger`, which I would prefer

doc/design/logging.md
```diff
@@ -0,0 +1,356 @@
...
+Although the macro contains an incomplete if-statement, the `LogStream` object at
+the end makes it safe to use since the compiler will complain if something else
+than a streaming operator or semicolon is used.
+
+#### Behavior before calling initLogger
+
+In order to have log messages before `initLogger` is called, the default logger
+is used with `LogLevel::INFO`. It is up to the implementation of the default
+logger what to do with these messages. For iceoryx the default logger is the
+`ConsoleLoger` (this can be changed via the platform abstraction) which will
```
elfenpiff, line 140: ConsoleLogger

elfenpiff (commented):
Looks good ...

doc/design/logging.md
```diff
@@ -0,0 +1,356 @@
...
+| LogLevel | Usage |
+|----------|-------|
+| FATAL    | For fatal an non-recoverable errors which essentially leads to a termination of the program. |
+| ERROR    | For severe but recoverable errors. |
+| WARN     | For cases when something unintentional happens but not considered as severe error. |
+| INFO     | For important status information a user should be aware of. |
+| DEBUG    | Anything that is helpful for debugging. |
+| TRACE    | Any comment could be replaced by logging call with log level trace. |
+
+The log levels `FATAL`, `ERROR` and `WARN` should not be used independent but in
```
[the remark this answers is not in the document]
elBoberido: I envision something like `IOX_ERROR(MODERATE, FooError::BAR) << "You are screwed";`

doc/design/diagrams/logging/logging_classes.puml
```diff
@@ -0,0 +1,85 @@
...
+  + LogStream& operator<<(const Callable&& c)
+}
+
+}
+
+ConsoleLogger --|> Logger
+TestingLogger --|> ConsoleLogger
+Logger <.. LogStream : <<friend>>
+LogStream .> LogHex : <<friend>>
+LogStream .> LogOct : <<friend>>
```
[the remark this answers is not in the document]
elfenpiff: Mmmmhhh this sound intriguing. Could you add this to the design document.

doc/design/logging.md
```diff
@@ -0,0 +1,360 @@
...
+![logger runtime replacement](../website/images/logger_runtime_replacement.svg)
+
+#### Replacing the default logger for a platform
+
+Each platform must specify the following aliases in the `platform_settings.hpp`
+header.
+
+```cpp
+using LogLevel = pbb::LogLevel;
+using pbb::asStringLiteral;
```
FerdinandSpitzschnueffler, line 173: `pbb` means platform building blocks?

FerdinandSpitzschnueffler (commented):
LGTM

doc/design/logging.md
elBoberido: @FerdinandSpitzschnueffler yes, I try to keep my keyboard mileage low. Happy for other naming suggestions.

FerdinandSpitzschnueffler: @elBoberido No complaints. Just wanted to know if my guess is correct.

doc/design/diagrams/logging/logging_classes.puml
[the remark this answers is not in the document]
elBoberido: @elfenpiff I'll try it in the poc and add it to the design document if it works as expected

elfenpiff (dismissed):

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+        iox::log::setActiveLogger(&myLogger);
+        iox::log::initLogger(iox::log::logLevelFromEnvOr(iox::log::LogLevel::INFO));
+    }
+
+  private:
+    void setupNewLogMessage(const char*, const int, const char*, iox::log::LogLevel logLevel) override
+    {
+        switch(logLevel) {
+            case iox::log::LogLevel::FATAL:
+                logString("💀: ");
```
MatthiasKillat, line 312: The example leaves out what to do with the general arguments (of type `const char*` etc.) of the call as it just ignores them. Maybe the purpose can be made more clear by using at least some of them.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+            case iox::log::LogLevel::TRACE:
+                logString("🐾: ");
+                break;
+            default:
+                logString("🐔: ");
+        }
+    }
+
+    void flush() override {
+        puts(m_buffer);
```
MatthiasKillat, line 335: OK, `m_buffer` is inherited through the parent class I presume. The user really needs to know what she can and cannot do with this buffer.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+    IOX_LOG(TRACE) << "Row row row your boat!";
+
+    return 0;
+}
+```
+
+## Open issues
+
+- do we need to change the log level after `initLogger`
+- do we want a `IOX_LOG_IF(cond, level)` macro
```
MatthiasKillat, line 358: I would like  `IOX_LOG_IF`. This is practical if we want to log anomalies (depending on a rare condition) in a lightweight way. Then we really should consider the proposed in  way to suppress condition evaluation as in https://www.foonathan.net/2016/09/assertions/ (if the level does not require evaluating it).

This kind of performance gain is important to me as `cond` can be very expensive to evaluate.

MatthiasKillat (commented):
Looks good overall and I like it. I recommend looking at https://www.foonathan.net/2016/09/assertions/ and related links for inspiration. It is useful for error handling as well and I suggest we incorporate some of the ideas.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+| TRACE    | Any comment could be replaced by logging call with log level trace. |
+
+The log levels `FATAL`, `ERROR` and `WARN` should not be used independent but in
+combination with the error handler.
+
+## Design
+
+### Considerations
+
+1. Similar to other iceoryx constructs, heap usage, exceptions and recursive
```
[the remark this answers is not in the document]
elBoberido: Well, it depends. Our MemPools are also dynamic memory but handled by our self.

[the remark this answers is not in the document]
elBoberido: No, it does not need recursion. It's just that many safety guidelines discourage recursion and therefore I added this constraint to the logger.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+   `std::stringstream` cannot be used.
+
+2. Since the logger shall be used by the `iceoryx_hoofs` itself, the base
+   implementation can also not make use of functionality in e.g. `iox::cxx` or
+   `iox::posix`
+
+3. To have a minimal runtime overhead, string formatting shall only happen when
+   the log level is above the output threshold. This lazy evaluation shall also
+   apply to other potentially expensive operations like the result of a function
+   call.
```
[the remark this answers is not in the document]
elBoberido: @MatthiasKillat yes, something like `IOX_LOG(TRACE) << "#### " << expensiveCall();` should not be expensive when the log level is set to `WARN`

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+special cases when file, line and function are supplied from other places than
+the log macro invocation, like the `Expects` and `Ensures` macros.
+
+Although the macro contains an incomplete if-statement, the `LogStream` object at
+the end makes it safe to use since the compiler will complain if something else
+than a streaming operator or semicolon is used.
+
+#### Behavior before calling initLogger
+
+In order to have log messages before `initLogger` is called, the default logger
```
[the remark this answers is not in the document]
elBoberido: Yes, we are not forced in order to have pre-main logging output

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+
+#### Replacing the logger at runtime
+
+The default console logger can be replaced at runtime to a custom logger which
+can be used to forward the log messages to a logger of another framework.
+
+This is done by deriving from the base `Logger` class and implementing the pure
+virtual `setupNewLogMessage` and `flush` methods. Finally, the derived logger
+must be activated by calling the static `Logger::activeLogger` function and
+passing the new logger as argument to the function. The logger must have a static
```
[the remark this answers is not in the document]
elBoberido: Yes, and it should be static.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+```cpp
+using LogLevel = pbb::LogLevel;
+using pbb::asStringLiteral;
+
+using Logger = pbb::Logger;
+using DefaultLogger = pbb::ConsoleLogger;
+using TestingLoggerBase = pbb::ConsoleLogger;
+```
+
+With `LogLevel` enum being the log level enum with the values defined in the
```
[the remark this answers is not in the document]
elBoberido: No, it is not intended to have this feature. I'm not sure if it is really valuable or used by many user but it would complicate the code.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+
+This is the sequence diagram of the setup of the testing logger:
+
+![logger testing sequence](../website/images/logger_testing_sequence.svg)
+
+#### Environment variables
+
+The behavior of the logger can be altered via environment variables and the
+`initLogger` function. Calling this function without arguments, it will check
+whether the environment variable `IOX_LOG_LEVEL` is set or use `LogLevel::INFO`
```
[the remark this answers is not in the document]
elBoberido: Okay, have to clarify this. The log level defined by `IOX_LOG_LEVEL` will be used

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+recommended to provide a public init function. If some initialization is needed,
+the `initLoggerHook` virtual function can be overloaded. This is called after
+the base logger is initialized.
+
+![logger compile time replacement](../website/images/logger_compile_time_replacement.svg)
+
+#### Dedicated logger for testing
+
+In order to have quiet tests, the logging output shall be suppressed for passed
+tests and printed to the console for failed tests.
```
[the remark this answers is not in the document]
elBoberido: The testing logger will be derived from the default console logger and use the heap instead of just flushing to the console.

Here is the proof of concept if your are interested in the details https://github.com/elBoberido/iceoryx/blob/logger-ng/iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/logger.hpp

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+  public:
+    static void init()
+    {
+        static MyLogger myLogger;
+        iox::log::setActiveLogger(&myLogger);
+        iox::log::initLogger(iox::log::logLevelFromEnvOr(iox::log::LogLevel::INFO));
+    }
+
+  private:
+    void setupNewLogMessage(const char*, const int, const char*, iox::log::LogLevel logLevel) override
```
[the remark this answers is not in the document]
elBoberido: It sets up the new log message with the timestamp, line number, etc. The actual output is not written to the buffer.

I'm happy for better naming suggestions.

doc/design/logging.md
elBoberido: It ignores it since it is file, and function name but I should make it more clear

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+
+![logging with logstream](../website/images/logging_with_logstream.svg)
+
+#### Macro with lazy evaluation
+
+The buildup of the log message is only done when the log level is above the output
+threshold. This is accomplished by a macro with an incomplete if-statement.
+
+```cpp
+#define LAZY() if (cond)
```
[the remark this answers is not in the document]
elBoberido: Yes, it can be omitted. But if you don't mind, I would keep it since it is closer to the actual implementation further down.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+
+LAZY() expensiveFunctionCall();
+```
+
+In the example above `expensiveFunctionCall` is only executed when `cond` is `true`.
```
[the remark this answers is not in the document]
elBoberido: Exactly. `cond` will be `compileTimeCond && runTimeCond`.

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: I clarified this a bit.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+Although the macro contains an incomplete if-statement, the `LogStream` object at
+the end makes it safe to use since the compiler will complain if something else
+than a streaming operator or semicolon is used.
+
+#### Behavior before calling initLogger
+
+In order to have log messages before `initLogger` is called, the default logger
+is used with `LogLevel::INFO`. It is up to the implementation of the default
+logger what to do with these messages. For iceoryx the default logger is the
+`ConsoleLogger` (this can be changed via the platform abstraction) which will
```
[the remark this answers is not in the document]
elBoberido: Yes, and it is also required to have important log messages from pre-main logger calls.

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: Added some clarification.

doc/design/logging.md
elBoberido: You are right. This exposes too much of the implementation. I'll change this to

```cpp
puts(Logger::buffer());
Logger::assumePushed();
```

doc/design/logging.md
elBoberido: This is also already taken care of with `if ((level) <= iox::log::Logger::minimalLogLevel() &&` which will be evaluated at compile time to `true` or `false`.

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+This is the log macro with lazy evaluation
+
+```cpp
+#define IOX_LOG_INTERNAL(file, line, function, level)          \
+    if ((level) <= iox::log::Logger::minimalLogLevel()         \
+        && (iox::log::Logger::ignoreLogLevel()                 \
+            || (level) <= iox::log::Logger::getLogLevel()))    \
+    iox::log::LogStream(file, line, function, level).self()
+
+#define IOX_LOG(level) IOX_LOG_INTERNAL(__FILE__, __LINE__, __FUNCTION__, iox::log::LogLevel::level)
```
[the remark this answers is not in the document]
elBoberido: I'll leave this to decide for later. The main question is if this struct should be part of the logger of live somewhere else since it might also be helpful for `Expects`/`Ensures`

elBoberido:
Okay, final review round. In case your concerns did not receive attention, please bring them up again. Since there wasn't much discussion about the general design, I will start with the implementation and take the risk to bigger changes due to this PR not yet being merged.

elBoberido:
> @elBoberido I think we need in the error handling and the logger a best practice how to deal with those things inside the signal handler. I would add a best practice section for this inside the log design document. What do you think about something like this:

I thinks it makes more sense to have a document for signal handling and then link from the error handling and logging design document to this document. This could then also put emphasis on how one can break free of the `WaitSet` and `Listener` if we keep our current implementation.

cc @MatthiasKillat

doc/design/logging.md
elBoberido: Adjusted as proposed

doc/design/diagrams/logging/logging_classes.puml
elBoberido: Another consideration. What do you think to rename `OFF` to `NONE`/`QUIET`/`SILENT` or something similar. `OFF` kind of does not fit with the other levels.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+
+- do we need to change the log level after `Logger::init`
+- do we want a `IOX_LOG_IF(cond, level)` macro
+- shall the TestingLogger register signals to catch SIGTERM, etc. and print the
+  log messages when the signal is raised? It might be necessary to wait for the
+  error handling refactoring before this can be done
+- instead of having the `Logger::init()` static function with hidden default
+  parameter this could be replaced by `Logger::init(LogLevel::WARN)`,
+  `Logger::initFromEnvOr(LogLevel::WARN)` and a builder like
+  `Logger::customize().logLevelFromEnvOr(LogLevel::WARN).init()`
```
elBoberido, line 393: I think I will try this when the class is implemented and if it feels nice I will also update the design document. What do you think?

doc/design/diagrams/logging/logging_classes.puml
elfenpiff: * `SILENT` & `QUIET` is like I will log to a file but you won't see it
* `NONE` feels right here

One question, is it possible for the user to log a message to the `NONE` LogLevel so something like `Log(LogLevel::NONE) << "hello";` ? Or do we prevent this by providing only `LogTrace()` - `LogFatal()`

doc/design/logging.md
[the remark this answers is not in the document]
elfenpiff: The more I think about logging and error handling I am think we have to define it a little bit clearer and consider @mossmaurice ideas with splitting error handling and error reporting.

* In my naivety I thought `LogFatal() << "a fatal error";` was the error reporting mechanism to some kind of error code log mechanism in an ECU but it seems that it completely legal to call `LogFatal()` just for fun and only the user on the console sees a funny message.

* As mentioned also in the error handling design document, sometimes we want to communicate a recoverable error via `expected` but also inform some kind of error reporting mechanism. So one would write code like:
```cxx
LogError() << "an error has happened".
IOX_REPORT_ERROR(MyErrorCode);
return cxx::error();
```
Would it maybe an idea that `LogWarn`, `LogError` and `LogFatal` always report the error and always requires an error code. I can't think of any use case where this wouldn't make sense? So to summarize, shouldn't `LogWarn` - `LogFatal` be our error reporting mechanism?

And the error handling would then always consist of first reporting the error and then handling it. Here I really like your vision: `IOX_ERROR(MODERATE, FooError::BAR) << "You are screwed";`

doc/design/logging.md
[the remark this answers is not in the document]
elfenpiff: But if it would be helpful for `Expects`/`Ensures` you would be unable to use it since the requirement is that the logger has no dependency to `hoofs`.

I think this struct would be helpful in `hoofs` but could only be defined in the logger since this is our lowest building block.

So what do you think, does it make sense to introduce such a struct here and use it?

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+#define IOX_LOG_INTERNAL(file, line, function, level)          \
+    if ((level) <= iox::log::Logger::minimalLogLevel()         \
+        && (iox::log::Logger::ignoreLogLevel()                 \
+            || (level) <= iox::log::Logger::getLogLevel()))    \
+    iox::log::LogStream(file, line, function, level).self()
+
+#define IOX_LOG(level) IOX_LOG_INTERNAL(__FILE__, __LINE__, __FUNCTION__, iox::log::LogLevel::level)
+```
+
+With `minimalLogLevel` and `ignoreLogLevel` being static constexpr functions
```
elfenpiff, line 121: You have to also take into account if the filtering is happening in the external logger as described in requirement 8. Then you are also not allowed to ignore it.
Maybe you need to introduce something like `isFilteredExternally`.

Or is this the purpose of `ignoreLogLevel`? If so I would rename it so since the purpose feels different.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+and the ignore active log level check to always forward the log message to the
+logger, independent of the active log level.
+
+The `self()` invocation is used to create an lvalue reference to the `LogStream`
+object. This eases the implementation of logging support for custom types since
+`IOX_LOG(INFO) << myType;` would require to implement an overload with a rvalue
+`LogStream` reference but `IOX_LOG(INFO) << "#### " << myType;` requires a
+lvalue reference.
+
+The `IOX_LOG` macro is intended for general use and the `IOX_LOG_INTERNAL` for
```
elfenpiff, line 135: Here it would absolutely make sense to add a struct `SourceCodeOrigin` in the logger which bundles line, file, function together.
This struct then could be used in `Expects` and `Ensures` and would bring some uniformity.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+
+## Design
+
+### Considerations
+
+1. Similar to other iceoryx constructs, heap usage, exceptions and recursive
+   calls are forbidden. This also means that `std::string` and
+   `std::stringstream` cannot be used.
+
+2. Since the logger shall be used by the `iceoryx_hoofs` itself, the base
```
elfenpiff, line 42: I would rephrase it to the logger shall not have any dependency. Since in my opinion the logger is also not allowed to use the future error handler since the error handler should use the logger for error reporting (my opinion at the moment but it may change :smile:  )

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### Behavior before calling Logger::init
+
+In order to have log messages before `Logger::init` is called, the default logger
+is used with `LogLevel::INFO`. It is up to the implementation of the default
+logger what to do with these messages. For iceoryx the default logger is the
+`ConsoleLogger` (this can be changed via the platform abstraction) which will
+print the log messages to the console.
+
+Although it is possible to use the logger without calling `Logger::init`, this is
```
mossmaurice, line 151: :cat2: 
```suggestion
Although, it is possible to use the logger without calling `Logger::init`, this is
```

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+
+This is done by deriving from the base `Logger` class and implementing the pure
+virtual `setupNewLogMessage` and `flush` methods. Finally, the derived logger
+must be activated by calling the static `Logger::activeLogger` function and
+passing the new logger as argument to the function. The logger must have a static
+lifetime and should therefore be placed in the data segment.
+
+The call to `iox::log::Logger::init` will finalize the option to replace the logger
+at runtime. Further efforts to replace the logger will call the error handler
+with a `MODERATE` error level and a log message to the old and new logger
```
mossmaurice, line 168:
```suggestion
with a `MODERATE` error level and a log message to the old and new logger.
```

mossmaurice, line 168: Great idea! :1st_place_medal:

doc/design/diagrams/logging/logger_runtime_replacement.puml
```diff
@@ -0,0 +1,49 @@
...
+return
+return
+
+MyLogger -> Logger ++ : init
+Logger -> Logger ++ : get
+Logger -> Logger ++ : activeLogger()
+return
+return
+Logger -> loggerInstance ++ : initLoggerInternal
+loggerInstance -> myLoggerInstance ++ : initiLogger
```
mossmaurice, line 42:
```suggestion
loggerInstance -> myLoggerInstance ++ : initLogger
```

doc/design/diagrams/logging/logger_compile_time_replacement.puml
```diff
@@ -0,0 +1,28 @@
...
+  the default logger
+  is configured via
+  the alias in
+  platform_settings.hpp
+end note
+return
+return
+
+Logger -> loggerInstance ++ : initLoggerInternal
+loggerInstance -> loggerInstance ++ : initiLogger
```
mossmaurice, line 22:
```suggestion
loggerInstance -> loggerInstance ++ : initLogger
```

doc/design/diagrams/logging/logging_classes.puml
elBoberido: Whoopsi. In my proof of concept it is not caught but it should not be possible.

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: Well, that was my point. It is not user facing and we could make use of it at several places but the question is where to put it so that we can use it everywhere. Therefore I would not do it in the initial implementation but wait for the error handler refactoring and then decide what to do.

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: Since this is similar to https://github.com/eclipse-iceoryx/iceoryx/pull/1358#discussion_r905896907 I will response there

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+| LogLevel | Usage |
+|----------|-------|
+| FATAL    | For fatal and non-recoverable errors which essentially lead to termination of the program. |
+| ERROR    | For severe but recoverable errors. |
+| WARN     | For cases when something unintentional happens which is not considered a severe error. |
+| INFO     | For important status information a user should be aware of. |
+| DEBUG    | Anything that is helpful for debugging. |
+| TRACE    | Any comment could be replaced by logging call with log level trace. |
+
+The log levels `FATAL`, `ERROR` and `WARN` should not be used independently but
```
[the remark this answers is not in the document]
elBoberido: @elfenpiff @mossmaurice 

Please don't mix the logger with an error handler/reporter (I know an elf who heavily complained about this idea when it was proposed a few years ago ;) ). Since this is a library, it is totally fine to do this in my application

```cpp
int main() {
    IOX_LOG(ERROR) << "yadda yadda";
};
```

Please let the logger do what a logger is supposed to do, log messages.

I think we all agree that a call like `IOX_LOG(FATAL) << "yadda yadda";` or `IOX_LOG(ERROR) << "yadda yadda";` should terminate an application. At least I don't know of any logger which has such a behavior. This means we would need to do something like this

```cpp
IOX_LOG(FATAL, MY_ERROR_CODE) << "yadda yadda"; // does reports the error and returns
IOX_FOO(FATAL, MY_ERROR_CODE); // or at least IOX_FOO(FATAL); since in debug builds `ERROR` would also not return and e.g. lead to termination
```

or something like this

```cpp
IOX_LOG(WARN, MY_ERROR_CODE_1) << "yadda yadda";  // does report the error and returns
IOX_FOO(FATAL, MY_ERROR_CODE_2) << "yadda yadda";  // does report the error but does not return, e.g. terminate
```

and compare it so the proposal of @MatthiasKillat 

```cpp
IOX_RAISE(WARN, MY_ERROR_CODE_1) << "yadda yadda";  // does report the error and returns
IOX_RAISE(FATAL, MY_ERROR_CODE_2) << "yadda yadda";  // does report the error but does not return, e.g. terminate
```

It's unambiguous since it always reports the error and that is the important part. The user does not have to think which macro to use when he concludes that this is something which needs to be stored in the black box. This handler also ensures that the program does not continue with the execution if the error level is `FATAL`.

One could of course argue if `IOX_RAISE` should be split into two macros and I would also argue that this is a bad idea and a source for errors, e.g.

```
IOX_REPORT(MY_ERROR_CODE_1) << "yadda yadda";
IOX_FOO(FATAL); // or just nothing for WARN
```

- okay, it it is nothing for WARN, how do I know that the user just forgot to add the line?
- why is there a `IOX_FOO(FATAL)` but no `IOX_FOO(WARN)`?
- if there is a `IOX_FOO(WARN)`, `IOX_FOO(ERROR)` and `IOX_FOO(FATAL)` which must be used in combination with `IOX_REPORT(...)` why not combine it?
- can you list me all error codes with an error level of FATAL? With `IOX_RAISE(FATAL, ...)` it is just a `grep -irn "IOX_RAISE(FATAL"`.

The error handler should be the lowest building block and the error handler macro can use the logger to do something like
```
IOX_RAISE(FATAL, MY_ERROR) << "yadda yadda".
```

The reason for this is that the logger is more complex than the error handler. The logger must do something with the message (console, file) but the error handler/reporter must in the most naive case only terminate. If the error handler/reporter needs to send the error code via e.g. network it must still handle cases where an non-recoverable error occurs, and therefore it must be lower level than the logger.

@mossmaurice since I always assume that the user of my software is a psychopath who knows where I'm living, I would keep the error code. Without the error code one needs a mapping of file, line, fuction to a specific error and this for each release of iceoryx. It's also not fun to test for these errors.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+
+3. To have a minimal runtime overhead, string formatting shall only happen when
+   the log level is above the output threshold. This lazy evaluation shall also
+   apply to other potentially expensive operations like the execution of a
+   function call.
+
+4. Additionally, the minimal log level shall be configurable by a compile time
+   switch and everything below this log level should be compiled to no-ops. With
+   this, developer can make excessive use of the most verbose log level and
+   disable it by default. When debugging, this can be turned on and help to find
```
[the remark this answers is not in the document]
elBoberido: @elfenpiff this is two fold. With the compile time switch, the tests will ensure that the behavior is the same, therefore I don't see a big issue with this. The log level can also be set at runtime and yes this would influence the runtime behavior but I think it would be a bad idea to always pay the price for a log message even if it will be discarded. It the user wants the same behavior he just needs to use the same compiler time switch and don't allow to change the log level (e.g. via environment variables), which is possible with the current proposal.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+   the root of the problem.
+
+5. There shall also be the possibility to alter the behavior of the logger via
+   environment variables. This can be functionality like setting the log level or
+   deciding whether the timestamps or file names shall be part of a log message.
+
+6. Since iceoryx is often used as transport layer in a higher level framework, it
+   shall be possible to replace the default logger at runtime and delegate the
+   log messages to the logger of such a framework. To prevent misuse, this can
+   only be done up to a specific time. Afterwards an error message shall be
```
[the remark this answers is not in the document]
elBoberido: @elfenpiff @mossmaurice done

@elfenpiff we will definitely have to replace the logger at runtime. Let's assume someone want's to integrate iceoryx into an higher level framework and also want's to make use of the functionality of that framework, e.g. DDS. Since iceoryx will be lower level it would result in a cyclic dependency if this would only be possible at compile time, therefore it needs to be done at runtime.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+The logger can be customized at compile time and at runtime. The former is done
+by the `Impl` template parameter and the latter by deriving from the logger.
+
+#### Logging with LogStream
+
+![logging with logstream](../website/images/logging_with_logstream.svg)
+
+#### Macro with lazy evaluation
+
+The buildup of the log message is only done when the log level is above the output
```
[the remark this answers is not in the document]
elBoberido: How about `The buildup of the log message is only done when the condition to log the message is fulfilled.`? This is less restrictive and also covers an `IOX_LOG_IF` in the future.

doc/design/logging.md
elBoberido: Already answered in https://github.com/eclipse-iceoryx/iceoryx/pull/1358#discussion_r906194754

It is an implementation detail and can easily be done once we know where we can put such a struct. It is not necessary to have it in the initial implementation.

doc/design/logging.md
elBoberido: As already answered in https://github.com/eclipse-iceoryx/iceoryx/pull/1358#discussion_r906234409

A minimal logger is more complex than a minimal error handler/reporter and therefore should be higher level than the error handler.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: Assuming that I understood everything correctly.

1.) I agree that we should not split logging and reporting in fatal error cases. I.e. if there is a fatal error, only use `IOX_RAISE(FATAL)` or some variation which also logs, optionally executes some code and terminates.
2.) For non-fatal errors I would argue it is the same, i.e. no split of logging and error reporting. The difference is that now the program continues (likely/necessarily on a different path).
3.) `IOX_LOG(<NON-FATAL>)` would be used if there is no error but it is useful to log some information, like that some (non-error) event happened, how much time some call took etc.

 This means we only explicitly log good or at least neutral events, while the bad ones require reporting an error and logging happens automatically.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: Sure, we do not want to use it though even if we manage it ourselves because we might run out at runtime. That is why I prefer a more general statement here. Allocating before logging begins is of course fine (and not dynamic).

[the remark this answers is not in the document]
MatthiasKillat: Fine either way, we do not use recursion for the error handler and can state it as a constraint. I just consider it a minor issue and would like to not require this for some algorithm to be considered safe (because it is more complicated than that).

doc/design/logging.md
```diff
@@ -0,0 +1,361 @@
...
+![logging class diagram](../website/images/logging_classes.svg)
+
+#### Logging with LogStream
+
+![logging with logstream](../website/images/logging_with_logstream.svg)
+
+#### Macro with lazy evaluation
+
+The buildup of the log message is only done when the log level is above the output
+threshold. This is accomplished by a macro with an incomplete if-statement.
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, I used and combined this with template trickery in the new error handling. The important part is that it is optimized away if the branching can be determined at compile time.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: I do not mind. I was just curious why we would define it like that.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: Expects/Ensures can be made obsolete but yes, we can bundle this later. I do not know yet where this should live. It is not critical, just saves us some boilerplate here and there while implementing the logging.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: I agree. This makes management and runtime replacement of loggers easier. Then we can assume it exists for the whole lifetime of the program and do not have to worry about it. There are minor issues after main terminates (order of destruction), but those are manageable:

It is generally problematic if other objects with static lifetime use logging after main returns. The usage must ensure that the logger is destroyed last. However, we cannot enforce it is created first among all statics. It can be solved by constructing it into static memory (placement new) and never calling the dtor. Of course, the dtor does not have to be logically mandatory for this, e.g. have some connection to the outside world (e.g. close a data base connection).

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: Ok. Just nice to have and probably easy to add. It would sure add some complexity but I think not much. We can consider it when we need it though.

It was just a question/suggestion.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: Yes, in tests we could use built-in dynamic memory.

doc/design/logging.md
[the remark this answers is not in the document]
MatthiasKillat: I was just wondering why we emphasize `new` and `setup` here. I would have tended to `createLogMessage` but do not feel strongly about it.

doc/design/logging.md
MatthiasKillat: Yes, if there is a compile time condition the branch will be optimized away. 

It is not a critical functionality, but nnice to have.  It should be used to log rare or unusual non-error events. For errors we  would use a conditional and `IOX_RAISE` or `IOX_RAISE_IF` in my opinion.

doc/design/logging.md
MatthiasKillat: I assume this refers to the \_\_FILE\_\_ \_\_LINE\_\_ etc.? Sure, not critical at all and easy to refactor.

doc/design/logging.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+## Open issues
+
+- do we need to change the log level after `Logger::init`
+- do we want a `IOX_LOG_IF(cond, level)` macro
+- shall the TestingLogger register signals to catch SIGTERM, etc. and print the
+  log messages when the signal is raised? It might be necessary to wait for the
+  error handling refactoring before this can be done
+- instead of having the `Logger::init()` static function with hidden default
```
MatthiasKillat, line 390: Does it make sense to enforce initialization of the logger by design without ever having to call `init`?

This can be done, but it cannot be done automatically as the very first static initialization I think. To do enforce this, a static dummy object defined in the executable first which calls `init` in its constructor may be sufficient (not 100% sure, need to research this).

The benefit is that then it can be used in any static, stack or heap object allocated afterwards.

MatthiasKillat (commented):
Looks good. The API and design is fine I think, implementation details can be changed later if needed.

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: @MatthiasKillat I'll use the implementation from the refactored error handler once it is available

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: I think `createLogMessage` works.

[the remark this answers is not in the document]
elBoberido: @MatthiasKillat I went with `createLogMessageHeader`

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
+# Logging
+
+## Summary and problem description
+
+Logging is a crucial part of a framework since it informs the developer and
+user of the framework about anomalies and helps with debugging by providing
+contextual information.
+
+Additionally, when integrated into a separate framework, the log messages should
```
elfenpiff, line 9: Not a strong must but here I would introduce a little requirements subsection and list the following 3 sentences as the 3 requirements we have to fulfill.

Then it is crystal clear what we aim for here.

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
...
+### Log levels and how to use them
+
+| LogLevel | Usage |
+|----------|-------|
+| FATAL    | For fatal and non-recoverable errors which essentially lead to termination of the program. |
+| ERROR    | For severe but recoverable errors. |
+| WARN     | For cases when something unintentional happens which is not considered a severe error. |
+| INFO     | For important status information a user should be aware of. |
+| DEBUG    | Anything that is helpful for debugging. |
+| TRACE    | Any comment could be replaced by logging call with log level trace. |
```
elfenpiff, line 29: I think this does not fit. Here a short example, `storable_function.inl:113`. The copy assignment could then log:
```
trace:: this operations is needed for destroy, then changed to source (rhs) operations
```

this would be only weird and confusing.

But when I think about trace, the word itself, then I would like to trace steps etc. So I would put as recommendation or guideline here that lifetime changes (creation/destruction) of vital objects could be logged with trace as well as important achievements in the lifetime of an application. When it is a little nebulous here an example.

* inform the user about any shared memory object that was created and mapped into process space
* when roudi is started up correctly
* whenever a waitset or listener is created or the underlying condition variables
* new applications which register or unregister at roudi

What do you think?

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
...
+
+The log levels `FATAL`, `ERROR` and `WARN` should not be used independently but
```
elfenpiff, line 31: Just because we already had a lengthy discussion about this sentence. Could you add a sentence which also states that the error handler should also be used only in combination with an appropriate log message.

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
...
+
+8. Additionally, there shall be a compile time option to forward all log messages
+   independent of the log level threshold for cases where the logger framework
+   does the filtering itself.
+
+9. For tests, the default logger shall be replaced to suppress the log messages in
+   passed tests but output them for failed tests. This keeps the noise low but
+   still provides useful information when needed. Additionally, there shall be an
+   environment variable to circumvent the suppression when desired.
+
```
elfenpiff, line 80: Since we already encountered the mutex in the logger. Should we add a point were we state that the logger shall be usable/exchangeable without the usage of a mutex.
Here I even would go so far to restrict it to an `atomic_flag` which would perfectly fit the use case of overriding it once.

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
...
+#### Behavior before calling Logger::init
+
+In order to have log messages before `Logger::init` is called, the default logger
+is used with `LogLevel::INFO`. It is up to the implementation of the default
+logger what to do with these messages. For iceoryx the default logger is the
+`ConsoleLogger` (this can be changed via the platform abstraction) which will
+print the log messages to the console.
+
+Although, it is possible to use the logger without calling `Logger::init`, this is
```
elfenpiff, line 120: What about the use case of a global static class which is using the logger. When the program terminates it is possible that after main is finished the logger goes out of scope before the other global static class does and then we have a segfault or undefined behavior.

Would it makes sense to add this as a restriction for now and implement a nifty counter approach for the logger later to solve this issue.

doc/design/diagrams/logging/logger_testing_classes.puml
```diff
@@ -0,0 +1,27 @@
+@startuml
+
+class TestingLoggerBase
+note bottom: type alias to 'Logger<BaseLogger>' with
+  specific 'BaseLogger' implementation
```
dkroenke, line 5: The generated svg diagram shows a syntax error and needs to be renewed:

```suggestion
note bottom of TestingLoggerBase
  type alias to 'Logger<BaseLogger>' with
  specific 'BaseLogger' implementation
end note
```

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
...
+2. Since the logger shall be used by the `iceoryx_hoofs` itself, the base
+   implementation can also not make use of functionality in e.g. `iox::cxx` or
+   `iox::posix`
+
+3. To have a minimal runtime overhead, string formatting shall only happen when
+   the log level is above the output threshold. This lazy evaluation shall also
+   apply to other potentially expensive operations like the execution of a
+   function call.
+
+4. Additionally, the minimal log level shall be configurable by a compile time
```
dkroenke, line 51:
```suggestion
4. Additionally, the minimal log level shall be configurable by a compile-time
```

doc/design/logging.md
```diff
@@ -0,0 +1,385 @@
...
+```
+
+Lazy evaluation with stream based logging seems to be challenging. An alternative
+might be a variadic macro like `IOX_LOG(INFO, "Hello World");`
+
+### Miscellaneous
+
+- do we need to change the log level after `Logger::init`
+- do we want a `IOX_LOG_IF(cond, level)` macro
+- shall the TestingLogger register signals to catch SIGTERM, etc. and print the
```
dkroenke, line 376:
```suggestion
- shall the `TestingLogger` register signals to catch SIGTERM, etc. and print the
```

dkroenke (commented):
Just a few remarks after initial review.

doc/design/logging.md
elBoberido: I'm not sure if this belongs to this document. Usually I would look at the design document for the error handler for this but I can add a `and vice versa` if that makes you happy :)

doc/design/logging.md
elBoberido: log4j uses the following definition `Designates finer-grained informational events than the DEBUG`. So how about `Finer-grained information than DEBUG` to keep it short?

doc/design/logging.md
elBoberido: Good point. I'm not quite sure where the best location for this would be. Maybe a generic `The logger shall not block, neither on logging nor on runtime exchange` under `Considerations`?

doc/design/logging.md
elBoberido: The nifty counter will be part of the `PolymorphicHandler` and we will get it for free once we use it. But that only helps you once you know that you have a problem. Have to think where it makes sense to add such an information.

doc/design/logging.md
```diff
@@ -0,0 +1,402 @@
...
+can be used to forward the log messages to a logger of another framework.
+
+This is done by deriving from the base `Logger` class and implementing the pure
+virtual `createLogMessageHeader` and `flush` methods. Finally, the derived logger
+must be activated by calling the static `Logger::setActiveLogger` function and
+passing the new logger as argument to the function. The logger must have a static
+lifetime and should therefore be placed in the data segment.
+
+The logger can only be set once and subsequent calls to `Logger::setActiveLogger`
+will not replace the logger anymore and log error messages to the existing logger
```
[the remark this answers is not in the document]
elBoberido: @ithier the `and log error messages ...` is correct since the user needs to be informed that something bad happened. Nevertheless, I rephrased the sentence.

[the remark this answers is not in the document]
ithier: @elBoberido OK. With the previous wording I had to read a few times to understand (or maybe I still didn't understand).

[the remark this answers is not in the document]
elBoberido: @ithier to my excuse I'm not a native speaker :sweat_smile: 

What is your understanding of the current wording?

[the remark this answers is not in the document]
ithier: @elBoberido The logger won't be replaced AND ADDITIONALLY error messages will be written to the existing logger.

What's not entirely clear is if you mean "a log message indicating an error" will be written if trying to set another logger, or "messages logged by the application with `IOX_ERROR` will keep using the initially set logger.

I would reword it a bit still because it doesn't sound so natural, it is probably nitpicking and not so important so I think it's not worth spending the effort haha.

doc/design/logging.md
elfenpiff: We can do this. In the last days I tried to figure out what and where should be printed in which log level. I came up with the following definitions:

* **Trace** - Lifetime changes of vital constructs or track important intermediate steps
  * example: creation/remove of shared memory segments; log that roudi is ready to rumble :smile: 
* **Debug** - Like info but less relevant for the user.
  * example: use default config file, set certain log level
* **Info** - All log outputs which could be relevant for the daily user

I think the remaining log levels are clear.

But to keep things short, I agree with your definition!

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: @ithier the old logger remains but an error message will be logged with the old logger and the logger which was supposed to replace the old logger. This is to ensure that all parties get informed about the attempt to replace the logger. How would you phrase it?

doc/design/logging.md
elBoberido: @elfenpiff I would say `DEBUG` has no relevance to the user but only the developer.

How about `Anything which could be relevant for the daily user` for `INFO`?

I also wouldn't narrow it down too much. The transitions are fluid and depending on the context a lifetime change could be a `DEBUG` in a case where it is critical and a `TRACE` in another case.

elfenpiff: > How about Anything which could be relevant for the daily user for INFO?

I do not want to delay this PR too much but I find this somehow interesting to nail down when to debug and when to trace since I have no idea when to do what.

But in the end one could say: Debug is the Info log level for the fellow developer.

doc/design/logging.md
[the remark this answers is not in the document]
ithier: @elBoberido 

The logger can only be set once and subsequent calls to `Logger::setActiveLogger`
will have no effect. Instead, an error message will be logged to both loggers to notify both parties of the unsuccessful attempt to switch the active logger and the application will continue to use the original logger for subsequent logs.

doc/design/logging.md
elBoberido: @elfenpiff I like the suggestion for debug :)

We would have
| LogLevel       | Definition |
|------------|------------|
| INFO | Anything which could be relevant for the daily user     |
| DEBUG  | Anything that is helpful for debugging, i.e. the info log level for the fellow developer     |
| TRACE  | Anything that might be helpful for debugging but would be too verbose in most cases |

doc/design/logging.md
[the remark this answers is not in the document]
elBoberido: @ithier that's so much better

elfenpiff (approved):

PR #1398, opened 2022-06-14
iox-#1032 Define error handling requirements

MatthiasKillat:
Signed-off-by: Matthias Killat <matthias.killat@apex.ai>

1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Closes **TBD**

### DIFF
--- a/doc/design/error_handling_requirements.md
+++ b/doc/design/error_handling_requirements.md
@@ -1,0 +1,228 @@
+# Error Handling
+
+## Faults, Errors and Failures
+
+We use the following error terminology.
+
+1. A **fault** is a design flaw, a programming error or can e.g. be caused by unforseen use
+   (oversight of a use case is basically also a design flaw).
+2. An **error** is the result of a fault and manifests as an incorrect or inconsistent (intermediate) result.
+   An error may not be externally observable by the user of the system.
+3. A **failure** occurs if the system observably acts in contradiction to its specification.
+
+This reasonably assumes that errors are always caused by faults. If these errors are not detected
+they can cause failures. By detecting and handling errors failures are avoided.
+
+The goal of this document is to define error handling requirements.
+
+## Error Detection vs. Error Handling
+
+To handle an error it obviously needs to be detected first. To do so, the code must contain conditionals that check
+for inconsistent states, unavailability of resources, return codes of system calls etc.
+This means that the error has to be anticipated by the system designer,
+i.e. we cannot handle unforseen errors since they are not detected by definition.
+
+Error handling is the execution of handling code once the error was detected.
+This ranges from simple logging to sophisticated recovery or shutdown code and depends on the error in question.
+We distinguish between recoverable and non-recoverable error handling.
+
+### Recoverable Error Handling
+
+A recoverable error handling strategy allows the system to continue operation after the handling code
+was executed successfully. Exceptions are an example of such a mechanism but also any handler code
+that does not result in termination. If the handling code itself fails (e.g. insufficient resources)
+no recovery is possible and the system terminates.
+
+### Non-recoverable Error Handling
+
+A non-recoverable error handling strategy will always call terminate after the handling code was executed.
+Handler code may be used to initiate a graceful shutdown. An example are (strict) asserts.
+
+## Requirements
+
+In the following we list the requirements of the Error Handling (EH) strategy. The term platform
+refers to the usage environment of the error handler. There can be various platforms,
+e.g. for regular operations, testing, different OS etc.
+
+### Functional
+
+1. The EH shall support error codes.
+    - configurable on a platform basis
+    - errors from different platforms shall not be usable (checked at compile time)
+    - uniqueness is not enforced (can be done by the layer above if desired)
+    - error codes shall be optional (if none is used, it is mapped to some default error code)
+1. The EH shall support multiple error levels.
+    - the individual error levels are defined by the platform
+    - there has to be at least a FATAL error level (most severe error level)
+    - the error levels shall be ordered from least severe to most severe
+    - if the code requires an ERROR level that is not supported by the platform it shall not compile
+1. The EH shall support a non-recoverable error handling strategy in case of FATAL errors.
+    - non-recoverable errors shall call terminate after optional handling and logging code has been executed
+1. The EH shall support a recoverable error handling strategies for non-FATAL errors.
+    - recoverable error handling shall support execution of sufficiently general user actons (callbacks)
+1. The detection mechanism of the EH shall allow logging the error.
+    - user defined-logging shall be supported
+    - there shall be a default implementaton for logging
+1. The EH shall allow to trace the error to its source location (file, function, line).
+1. The EH shall be disableable for non-FATAL errors.
+    - disabling shall be possible up to some level at compile time; this level shall be non-FATAL
+    - can be used if there is sufficient trust in the system to disable e.g. warning checks
+    - by default all levels defined by the platform are enabled
+    - disabling shall cause the affected error logic in the code to cause no runtime overhead
+
+### Technical Constraints
+
+1. The EH shall be uniquely defined for each platform.
+    - only one error handler can be active for a platform at any time
+    - the EH shall not be changed while the system runs
+    - the EH must be configurable at runtime in the init phase
+      (for tests, but it might be sufficient to change it at compile time if only required for tests)
+    - for tests it must be possible to override the FATAL error reaction causing termination
+    - If there is no error handler set the system shall not run (check at the very beginning or enforc)
+2. The EH shall not use any exceptions.
+    - this excludes user callbacks (impossible to enforce)
+3. The EH shall not use dynamic memory.
+    - this excludes user callbacks (impossible to enforce)
+4. The EH shall be non-blocking.
+    - recoverable errors shall eventually pass back control to the invoking function
+    - cannot be enforced with user callbacks (the user is responsible to use non-blocking calls only)
+5. The EH shall be highly-visible in the code to be able to distinguish it from ordinary function calls.
+    - allows simple search for error handling locations in the code
+    - if implemented via macros it can easily be disabled
+6. The EH may depend on logging to some extent.
+    - only if it is supposed to use the logging capabilities in some platform implementation
+    - the logger cannot use the (complete) EH due to this
+
+## API Proposal
+
+The implementation can use a combination of templates and macros to achieve the goals regarding usability,
+configurability and perfomance.
+
+### Recoverable Error Handling API
+
+Assume we already have checked for some error condition and determined that we have to raise an error, i.e.
+it is not sufficient to propagate the error to some upper layer. This is a design decision, especially for
+non-FATAL errors.
+Then we can provide the ability to raise the error like so:
+
+`RAISE_ERROR(error_code, error_level)`
+
+We can define default values for the final arguments
+By default `error_level` is `FATAL` and hence could be optionally omitted.
+
+Optionally we can consider passing some function (functor) that will be invoked in addition,
+but this is basically syntactic sugar at this point as we already have determined that an error occured
+and could call this functor directly. The ability to pass this function makes the implementation more difficult
+if sufficent generality is desired (optional arguments etc.).
+
+The installed error handler can now decide how to proceed with the raised error and delegate to logging etc,
+shutdown code etc. Providing this call is useful to allow specific shutdown code which otherwise would require a lookup
+in some error code table (which can still be done if required).
+
+The only requirment is that for FATAL errors the `RAISE_ERROR` invocation shall never return and lead to termination.
+
+It is also possible to provide no `error_code` at all, which leads t a generic FATAL error:
+
+`RAISE_ERROR()`.
+
+A further conisderation is allowing custom messages (in principle not necessary if we allow custom handler functions)
+which are delegated to the handling code. The handling code can then decide to log them or otherwise incorporate
+it into the reporting code.
+
+`RAISE_ERROR(error_code, FATAL, "Mutex corrupted")`.
+
+The problem of various optional arguments is to define a proper argument order and, if macros are used,
+the requirement of variadic macros. It is generally not possible to leave out arguments in the middle
+without defining an entirely different macro or function.
+The macro wrappers should be very thin and directy delegate to a template function that enables
+the compiler to optimize the code.
+
+### Non-recoverable Error Handling API
+
+Non recoverable errors could be raised with the general API as well as
+
+`RAISE_ERROR(error_code, FATAL, optional_message)`
+
+This assumes we have checked for some error `condition` first. We can incorporate the check in another macro
+and arrive at:
+
+`ASSERT(!condition, error_code, optional_message)`
+
+This is mainly syntactic sugar to increase readability and state intent in a more concise way.
+We can use both notations. Notice that this always operates at FATAL level and must be always active
+since there is no notion how we can continue normally from there (otherwise it would not be a FATAL error).
+
+Again optional arguments lead to a more involved implementation. Providing an optional handler is even more useful here
+(as the check is happens in the macro) and can look like this:
+
+`ASSERT(!condition, error_code, handler, optional_handler_args)`
+
+If the handler is not invocable with provided arguments (if any) then the compilation will fail.
+Unless we define another macro for the default error code, `error_code` must be provided.
+It is not advisable to manage this in the handler for consistency and to allow search for the error codes.
+
+### Generalization of ASSERT to Recoverable Errors
+
+If we want to support recoverable errors with a similar syntax to `ASSERT` we can define
+
+`EXPECT(condition, error_code, error_level, handler, optional_handler_args)`
+
+This is similar to Goofle Test terminology and basically states that we expect some `condition` to be true
+but do not necessarily terminate if it is not (depending on the `error_level`).
+If it is not true we raise an error of the required level.
+
+Hence
+`EXPECT(condition, error_code, FATAL, handler, optional_handler_args)`
+
+is equivalent to
+
+`ASSERT(condition, error_code, handler, optional_handler_args)`.
+
+If we do not need to call a handler and just provide some error message we can define
+
+`EXPECT(condition, error_code, error_level, error_message)`
+
+or even omit it entirely
+
+`EXPECT(condition, error_code, error_level)`
+
+We have the possibility to specify at compile time the minimum error level we intend to handle to e.g. exclude warnings.
+This can be done without causing overhead for the affected checks. Assuming there are the levels
+`WARNING`, `RECOVERABLE_ERROR`, `FATAL` we can suppress all checks for `WARNING` in a way that not even
+`condition` is evaluated in
+
+`EXPECT(condition, error_code, WARNING)`,
+
+i.e. achieve a zero-overhead abstraction. This is useful if `WARNING`s are indeed no problem that compromise the
+ability to continue but instead unusal system states we encounter.
+
+### Replacement of Expects and Ensures
+
+`ASSERT` should replace `Expects` and `Ensures`. In particular there is no use to distinguish between them unless it is
+possible to check the conditions **before** the function is actually invoked or immediately **after** it returns.
+It is not possible with the capabilities of the current C++ language to specify true preconditions and postconditions
+conveniently (e.g. without wrapping all functions manually or with some automation),
+hence both are basically custom asserts.
+
+This is consistent with the C++ error handling proposal <https://doc.bccnsoft.com/docs/cppreference2018/en/cpp/language/attributes/contract.html>
+which is encourages to use `assert` inside a function body.
+The proposed `ASSERT` is more general except in one regard: it cannot be disabled by design. Disabling it could easily
+be made possible, but this would lead to problems if thefailure conditions where ever encountered to be true
+since it would be impossible to safely proceed.
+
+## Error Propagation
+
+Idiomatic propagation of errors with `expected`, `optional` or return codes is not in scope here ad will be covered elsewhere.
+
+## Open Points
+
+1. Design considerations of functionality such as general handler code provided by the user
+1. Optional arguments if the proposed macro API path is chosen
+1. Replacement strategy and definition of the active error handler will be similar to the new logger (platforms specific)
+1. Prototype implementation and detailed design
+1. Error propataion with `expected` is another issue but does not concern the error handler invocation
+1. Replacement of noexcept to allow exceptions in client code (e.g. types, callbacks) is not in scope here
+1. Macro naming should be close to Google Test but ideally not the same to ease searches.
+   Currently it is just a rough proposal.
+
+TBD


MatthiasKillat:
This is a draft version and there are probably some aspects missing. I implemented some similar prototype a while ago in a private repo but it had some different requirements so it cannot be ported directly. If there is an overall agreement about the requirements I will start a prototype implementation.

In the meantime I will prepare a mock macro implementation to demonstrate that it generally works and illustrate the usage in more detail.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+Hence
+`EXPECT(condition, error_code, FATAL, handler, optional_handler_args)`
+
+is equivalent to
+
+`ASSERT(condition, error_code, handler, optional_handler_args)`.
+
+If we do not need to call a handler and just provide some error message we can define
+
+`EXPECT(condition, error_code, error_level, error_message)`
```
MatthiasKillat, line 183: It is probably best to design it in a way that allows passing an optional error message with `operator<<` instead.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - this excludes user callbacks (impossible to enforce)
+4. The EH shall be non-blocking.
+    - recoverable errors shall eventually pass back control to the invoking function
+    - cannot be enforced with user callbacks (the user is responsible to use non-blocking calls only)
+5. The EH shall be highly-visible in the code to be able to distinguish it from ordinary function calls.
+    - allows simple search for error handling locations in the code
+    - if implemented via macros it can easily be disabled
+6. The EH may depend on logging to some extent.
+    - only if it is supposed to use the logging capabilities in some platform implementation
+    - the logger cannot use the (complete) EH due to this
```
MatthiasKillat, line 94: Thread-safety is missing. But apart from the potential logger use this is no problem as the calls need no synchronization in the general case. Synchronization may arise in user-defined error-handling if a error data-base is to be accessed under mutex or similar. But this would happen in the configurable use code (some handler) or some default implementation and hence be the responsibility of this extension code.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+To handle an error it obviously needs to be detected first. To do so, the code must contain conditionals that check
+for inconsistent states, unavailability of resources, return codes of system calls etc.
+This means that the error has to be anticipated by the system designer,
+i.e. we cannot handle unforseen errors since they are not detected by definition.
+
+Error handling is the execution of handling code once the error was detected.
+This ranges from simple logging to sophisticated recovery or shutdown code and depends on the error in question.
+We distinguish between recoverable and non-recoverable error handling.
+
+### Recoverable Error Handling
```
elfenpiff, line 29: As a thought to the recoverable vs. non-recoverable error handling:
Should we maybe state that when a error is non recoverable in iceoryx we always call the error handler and the error handler has only one error level fatal? This would simplify a lot and would provide us with a clean rule.
All recoverable errors always use `expected` for error handling (maybe we should rename it to result soon).

A non recoverable error in iceoryx would be something for me like:
* accessing non existing memory like, `optional.value()` without having a value or `vector[i], we can decide for undefined behavior or segfault or call the error handler`. In the end non recoverable means in the context of iceoryx that the next line will either lead to undefined behavior, a segfault or some other crash.

This can maybe can be mitigated so that this becomes a recoverable error, for instance by using exceptions in the error handler but on our level this is non recoverable.

A recoverable error would be something where, caused by this failure, at no time undefined behavior or a crash is immanent from our point of view. Something like creating shared memory but the user does not have the permission.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+## Requirements
+
+In the following we list the requirements of the Error Handling (EH) strategy. The term platform
+refers to the usage environment of the error handler. There can be various platforms,
+e.g. for regular operations, testing, different OS etc.
+
+### Functional
+
+1. The EH shall support error codes.
+    - configurable on a platform basis
```
elfenpiff, line 50: I maybe misunderstand you here but the error codes should be configured by a component like hoofs. If the platform defines the error codes then it is possible that required error codes are not defined. What then?
Could you maybe define error code first so that I get a better understanding.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - errors from different platforms shall not be usable (checked at compile time)
+    - uniqueness is not enforced (can be done by the layer above if desired)
+    - error codes shall be optional (if none is used, it is mapped to some default error code)
+1. The EH shall support multiple error levels.
+    - the individual error levels are defined by the platform
```
elfenpiff, line 55: The error levels must be defined by the error handler and non configurable. Otherwise we only can use the error level `FATAL` in our code base since we never know if some platform may not have this specific error level.

Otherwise I think we can get rid completely of error levels when we state the error handler is only called on non recoverable errors.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - there has to be at least a FATAL error level (most severe error level)
+    - the error levels shall be ordered from least severe to most severe
+    - if the code requires an ERROR level that is not supported by the platform it shall not compile
+1. The EH shall support a non-recoverable error handling strategy in case of FATAL errors.
+    - non-recoverable errors shall call terminate after optional handling and logging code has been executed
+1. The EH shall support a recoverable error handling strategies for non-FATAL errors.
```
elfenpiff, line 61: This can become very compex, therefore I propose to only support the `cxx::expected` approach for non recoverable errors.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - errors from different platforms shall not be usable (checked at compile time)
+    - uniqueness is not enforced (can be done by the layer above if desired)
+    - error codes shall be optional (if none is used, it is mapped to some default error code)
+1. The EH shall support multiple error levels.
+    - the individual error levels are defined by the platform
+    - there has to be at least a FATAL error level (most severe error level)
+    - the error levels shall be ordered from least severe to most severe
+    - if the code requires an ERROR level that is not supported by the platform it shall not compile
+1. The EH shall support a non-recoverable error handling strategy in case of FATAL errors.
+    - non-recoverable errors shall call terminate after optional handling and logging code has been executed
```
elfenpiff, line 60: Would it make sense to maybe make this configurable as well? So that we have the 3 options:
* terminate
* infinite busy loop
* dead lock

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - can be used if there is sufficient trust in the system to disable e.g. warning checks
+    - by default all levels defined by the platform are enabled
+    - disabling shall cause the affected error logic in the code to cause no runtime overhead
+
+### Technical Constraints
+
+1. The EH shall be uniquely defined for each platform.
+    - only one error handler can be active for a platform at any time
+    - the EH shall not be changed while the system runs
+    - the EH must be configurable at runtime in the init phase
```
elfenpiff, line 78: Here I would explicitly not handle it like the logger. I would state that the EH must be only configurable at compile time or link time. Otherwise what happens when you reconfigure the error handler at runtime? You could forbid that but than you have to do error handling for the error handler.
You could allow it but then you open the door to a lot of misuse cases. What happens when multiple instance or change the error handler or an error occurrs concurrently while the error handler is changed.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+      (for tests, but it might be sufficient to change it at compile time if only required for tests)
+    - for tests it must be possible to override the FATAL error reaction causing termination
```
elfenpiff, line 80: No we don't need that. In test we could throw an exception in our `optional handling` code snippet which the error handler has to provide also for fatal errors.
But we need to get rid of `noexcept` for that to work.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - If there is no error handler set the system shall not run (check at the very beginning or enforc)
```
elfenpiff, line 81: We should be able to make this impossible. We could provide a default error handler which prints some message and always terminates.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+2. The EH shall not use any exceptions.
```
elfenpiff, line 82: We have the requirement that the EH should support exceptions in user callbacks.

doc/design/error_handling_requirements.md
MatthiasKillat: No, I think this would be a mistake and too restrictive. It also allows the possibility to easily upgrade into a "WARNING" is handled like a FATAL error mode without changing the code in all places.

Or downgrade to "WARNING" means logging only.

The problem with `expected` is that without propagation it is useless. The other problem is that personally I find it cumbersome to use and so it should only be used it where propagation is required.

`expected` and the error detection mechanism do not serve the same purpose.

doc/design/error_handling_requirements.md
MatthiasKillat: It is fine with me to configure it at compile time only. Makes stuff even easier.

doc/design/error_handling_requirements.md
MatthiasKillat: We should not always throw something in FATAL mode, because this would have impacts on the binary size (possible impacts on execution speed (mainly in the bad case where we not care too much about it, but this is not guaranteed).

Therefore we should leave the option open to override the FATAL behavior. It is not complicated in code nor does it cause additional costs.

MatthiasKillat:
I move forward with some little prototype next where I showcase that it works nicely. Should not require too much code for the core functionality.

doc/design/error_handling_requirements.md
MatthiasKillat: Yes, right. I mixed this up with error levels where I am still pondering. Error codes are defined by the modules (whatever a module is exactly, but it could be hoofs or some class).

It is only important that we cannot mix up error codes between modules I think.

If we encounter any error code that is unknown we could 
1.) not compile
2.) ignore the code but do everything else we can (it will probably be mainly for logging but could also be for lookup based reaction)
3.) ignore the error
4.) something else ...

3.) is certainly not the way to go
2.) is hard to say whether this will not break specific runtime cases
1.) is safe as it forces the designer to provide the error code at runtime

So we shall go with 1.) I think (we do that already with enum classes). Ideally we enforce this with language capabilities, such as enum classes.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+### Recoverable Error Handling API
+
+Assume we already have checked for some error condition and determined that we have to raise an error, i.e.
+it is not sufficient to propagate the error to some upper layer. This is a design decision, especially for
+non-FATAL errors.
+Then we can provide the ability to raise the error like so:
+
+`RAISE_ERROR(error_code, error_level)`
+
+We can define default values for the final arguments
```
FerdinandSpitzschnueffler, line 110:
```suggestion
We can define default values for the final arguments.
```

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+and could call this functor directly. The ability to pass this function makes the implementation more difficult
+if sufficent generality is desired (optional arguments etc.).
+
+The installed error handler can now decide how to proceed with the raised error and delegate to logging etc,
+shutdown code etc. Providing this call is useful to allow specific shutdown code which otherwise would require a lookup
+in some error code table (which can still be done if required).
+
+The only requirment is that for FATAL errors the `RAISE_ERROR` invocation shall never return and lead to termination.
+
+It is also possible to provide no `error_code` at all, which leads t a generic FATAL error:
```
FerdinandSpitzschnueffler, line 124:
```suggestion
It is also possible to provide no `error_code` at all, which leads to a generic FATAL error:
```

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+and arrive at:
+
+`ASSERT(!condition, error_code, optional_message)`
+
+This is mainly syntactic sugar to increase readability and state intent in a more concise way.
+We can use both notations. Notice that this always operates at FATAL level and must be always active
+since there is no notion how we can continue normally from there (otherwise it would not be a FATAL error).
+
+Again optional arguments lead to a more involved implementation. Providing an optional handler is even more useful here
+(as the check is happens in the macro) and can look like this:
```
FerdinandSpitzschnueffler, line 156:
```suggestion
(as the check happens in the macro) and can look like this:
```

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+### Replacement of Expects and Ensures
+
+`ASSERT` should replace `Expects` and `Ensures`. In particular there is no use to distinguish between them unless it is
+possible to check the conditions **before** the function is actually invoked or immediately **after** it returns.
+It is not possible with the capabilities of the current C++ language to specify true preconditions and postconditions
+conveniently (e.g. without wrapping all functions manually or with some automation),
+hence both are basically custom asserts.
+
+This is consistent with the C++ error handling proposal <https://doc.bccnsoft.com/docs/cppreference2018/en/cpp/language/attributes/contract.html>
+which is encourages to use `assert` inside a function body.
```
FerdinandSpitzschnueffler, line 208:
```suggestion
which encourages to use `assert` inside a function body.
```

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+The proposed `ASSERT` is more general except in one regard: it cannot be disabled by design. Disabling it could easily
+be made possible, but this would lead to problems if thefailure conditions where ever encountered to be true
```
FerdinandSpitzschnueffler, line 210:
```suggestion
be made possible, but this would lead to problems if the failure conditions where ever encountered to be true
```

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+
+1. Design considerations of functionality such as general handler code provided by the user
+1. Optional arguments if the proposed macro API path is chosen
+1. Replacement strategy and definition of the active error handler will be similar to the new logger (platforms specific)
+1. Prototype implementation and detailed design
+1. Error propataion with `expected` is another issue but does not concern the error handler invocation
+1. Replacement of noexcept to allow exceptions in client code (e.g. types, callbacks) is not in scope here
+1. Macro naming should be close to Google Test but ideally not the same to ease searches.
+   Currently it is just a rough proposal.
+
```
elBoberido, line 227: At the end some food for thoughts.

I like the idea to call this `RAISE_something`. What do you think about aligning the error levels with the log levels?

This would mean that the macro would look like this `RAISE(FATAL, error code)`, `RAISE(ERROR, error code)` and `RAISE(WARNING, error code)`. Maybe with an `IOX_` prefix. With this it would be obvious which log level is used for the custom message. `ERROR` would be our current `SEVERE` and `WARNING` our current `MODERATE`.

Additionally there would be `RAISE_IF(condition, error level, error code)` and `ASSERT` would be syntactic sugar for this macro.

doc/design/error_handling_requirements.md
elfenpiff: We could have a custom error handler only for the tests which throws the error code as exception.

elBoberido: So you never run the tests with the same compiler flags as the production code?

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+Unless we define another macro for the default error code, `error_code` must be provided.
+It is not advisable to manage this in the handler for consistency and to allow search for the error codes.
+
+### Generalization of ASSERT to Recoverable Errors
+
+If we want to support recoverable errors with a similar syntax to `ASSERT` we can define
+
+`EXPECT(condition, error_code, error_level, handler, optional_handler_args)`
+
+This is similar to Goofle Test terminology and basically states that we expect some `condition` to be true
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, it could be a future company name https://www.youtube.com/watch?v=0kAEthfslsE ...

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+      (for tests, but it might be sufficient to change it at compile time if only required for tests)
+    - for tests it must be possible to override the FATAL error reaction causing termination
+    - If there is no error handler set the system shall not run (check at the very beginning or enforc)
+2. The EH shall not use any exceptions.
+    - this excludes user callbacks (impossible to enforce)
+3. The EH shall not use dynamic memory.
+    - this excludes user callbacks (impossible to enforce)
+4. The EH shall be non-blocking.
+    - recoverable errors shall eventually pass back control to the invoking function
+    - cannot be enforced with user callbacks (the user is responsible to use non-blocking calls only)
```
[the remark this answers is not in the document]
MatthiasKillat: I will rephrase this while creating prototype. But I think the intention is clear and we agree on it.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+
+Assume we already have checked for some error condition and determined that we have to raise an error, i.e.
+it is not sufficient to propagate the error to some upper layer. This is a design decision, especially for
+non-FATAL errors.
+Then we can provide the ability to raise the error like so:
+
+`RAISE_ERROR(error_code, error_level)`
+
+We can define default values for the final arguments
+By default `error_level` is `FATAL` and hence could be optionally omitted.
```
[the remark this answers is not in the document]
MatthiasKillat: Fair enough, it is again easier since it avoids variadic macros, various of overloads or default parameters.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+
+Optionally we can consider passing some function (functor) that will be invoked in addition,
+but this is basically syntactic sugar at this point as we already have determined that an error occured
+and could call this functor directly. The ability to pass this function makes the implementation more difficult
+if sufficent generality is desired (optional arguments etc.).
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, it will not be part of the macro/function call anyway. I can show it in the prototype and we may decide to scrap it.

doc/design/error_handling_requirements.md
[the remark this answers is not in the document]
MatthiasKillat: This is just syntactic sugar. But true, fewer ways of achieving something makes it simpler to use. We will use the general version then.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+### Non-recoverable Error Handling API
+
+Non recoverable errors could be raised with the general API as well as
+
+`RAISE_ERROR(error_code, FATAL, optional_message)`
+
+This assumes we have checked for some error `condition` first. We can incorporate the check in another macro
+and arrive at:
+
+`ASSERT(!condition, error_code, optional_message)`
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, I am implementing it this way but the names can be still changed later.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+`condition` is evaluated in
+
+`EXPECT(condition, error_code, WARNING)`,
+
+i.e. achieve a zero-overhead abstraction. This is useful if `WARNING`s are indeed no problem that compromise the
+ability to continue but instead unusal system states we encounter.
+
+### Replacement of Expects and Ensures
+
+`ASSERT` should replace `Expects` and `Ensures`. In particular there is no use to distinguish between them unless it is
```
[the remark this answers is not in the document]
MatthiasKillat: We cannot have proper pre- and post-conditions that are enforced anyway, unfortunately. Let alone in the sense of e.g. Hoare Logic.

doc/design/error_handling_requirements.md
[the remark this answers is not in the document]
MatthiasKillat: We can do this. The capability will be there to define fine-grained macros with this kind of fnctionality. We shouldprobably not go overboard with slightly different ones though. DEBUG_ASSERT is fine though.

doc/design/error_handling_requirements.md
mossmaurice: I agree with @elfenpiff that for recoverable errors we should use `cxx::expected`. This is the most idiomatic way. However, those still need to be reported and stored in e.g. in a persitency module. What do you think about my proposal few threads above to call the `errorReporter/errorHandler` from `cxx::error`? I don't like our triplet of

1. log entry
2. `errorHandler` call
3. return of `cxx::error`

That's quite noisy and inconvenient.

doc/design/error_handling_requirements.md
```diff
@@ -0,0 +1,228 @@
...
+    - error codes shall be optional (if none is used, it is mapped to some default error code)
+1. The EH shall support multiple error levels.
+    - the individual error levels are defined by the platform
+    - there has to be at least a FATAL error level (most severe error level)
+    - the error levels shall be ordered from least severe to most severe
+    - if the code requires an ERROR level that is not supported by the platform it shall not compile
+1. The EH shall support a non-recoverable error handling strategy in case of FATAL errors.
+    - non-recoverable errors shall call terminate after optional handling and logging code has been executed
+1. The EH shall support a recoverable error handling strategies for non-FATAL errors.
+    - recoverable error handling shall support execution of sufficiently general user actons (callbacks)
```
mossmaurice, line 62:
```suggestion
    - recoverable error handling shall support execution of sufficiently general user actions (callbacks)
```

doc/design/error_handling_requirements.md
mossmaurice, line 81:
```suggestion
    - If there is no error handler set the system shall not run (check at the very beginning or enforce)
```

doc/design/error_handling_requirements.md
elBoberido: @mossmaurice the user must be able to override bode, the error handling and the error reporting.

AFAIK @MatthiasKillat is moving toward like

```cpp
if (someCall.has_error())
{
    IOX_RAISE(error level, error code) << "Be careful an recoverable error happened!";
    return cxx::error<BlubbError>(BlubbError::SOME_ERROR);
}
```

So one line less. I would also decouple the `cxx::error` from the error reporting/handling. Since hoofs is a library, we should not force our users to a specific pattern and the expected is totally fine to use without the error reporter/handler. Additionally, not every time we use the `expected` it is warranted to also report the error. Sometimes it might be better to bubble up the error until enough context is available. For example, we use the `expected` for `Subscriber::takeChunk` and when there is no chunk available we do not want to report this since this is not an error when we poll but we communicate it with the `cxx::error`.

mossmaurice: @elBoberido 
> the user must be able to override bode, the error handling and the error reporting.

For sure, that's what I meant. Shame on me if you understood it differently :sweat_smile: 

> Since hoofs is a library, we should not force our users to a specific pattern and the expected is totally fine to use without the error reporter/handler. 

Good point, that should definitely not happen. The "how" can be figured out for sure somehow (inheritance, special mode, etc.). My main fear is that reporting a recoverable error will be forgotten both when writing the code and when reviewing it. IMHO it is error-prone, just look at how chaotic it is in the current code. I would prefer to have concise rules like A) recoverable don't need to call `IOX_RAISE` B) non-recoverable error means call `IOX_RAISE` (see examples depicted below). Also, why do I have to show my intention of reporting an error to the user two times? For me the idiomatic way is the return type `cxx::error`.

> Additionally, not every time we use the expected it is warranted to also report the error.

I agreed with the `takeChunk` case, although that is already some sort of a misuse of `std::expected`. However, an error is still an error. I would feel much safer preemptively reporting all of them after reading [this book](https://www.amazon.de/-/en/Chris-Hobbs/dp/0367338858/ref=sr_1_1?crid=XVDHEP7F0YE1&keywords=chris+hobbs&qid=1656075258&sprefix=chris+hobbs%2Caps%2C97&sr=8-1).


I couldn't yet answer the question to external contributors if someone would ask: When should I use `IOX_RAISE` in the code? What is the purpose of it?

Summed up my two suggestions are:

1. Clear and concise rules for recoverable and non-recoverable errors (like @elfenpiff suggested)
2. Stronger differentiate between error reporting and error handling to the user (Who handles the error? Me, the API caller, the system above or god?)

## Recoverable errors

```

std::expected<FooBar, FooBarError> doSomethingHighLevel()
{

   if(doSomethingLowLevel().has_error())
   {
      // IMHO the next line is redundant and can be error-prone
      IOX_RAISE(error level, error code) << "Be careful an recoverable error happened!"; // better IOX_REPORT
      // Handle or pass up with automatic conversion by implementing the mapping as discussed before
      return cxx::error<FooBarError>(doSomethingLowLevel().get_error());
   }

}

std::expected<FuuBaz, FuuBazError> doSomethingLowLevel()
{
   if (someThing.has_error())
   {
       // IMHO the next line is redundant and can be error-prone
       IOX_RAISE(error level, error code) << "Be careful an recoverable error happened!"; // better IOX_REPORT
       return cxx::error<BlubbError>(BlubbError::SOME_ERROR);
   }
}

```

My wish would be to see the error callstack or propoagation of the errors through the different layers in the fault recorder.

If we stick with adding the `IOX_RAISE` manually, I still would like to rename it to something like `IOX_REPORT`. We are just reporting and error here and are not handling it. The error being handled by the user via `cxx::error` :exclamation:  Of course `IOX_RAISE` can be called underneath non-fatally.

## Non-recoverable errors

```
T& vector::operator[](uint64_t index)
{
    // This calls IOX_RAISE(error level, error code) underneath
    IOX_ASSERT(index < m_size) << "Out of bounds access";
    return reinterpret_cast<const T*>(m_data)[index];
}
```

@MatthiasKillat Could you add such specific examples for both use cases? Thanks!

mossmaurice: @elBoberido Sure, on macro `IOX_RAISE` is fine. I also agree with the error handler being the lowest building block. However, my point of the above post was mainly to not burden developers with the necessity to add `IOX_RAISE` for recoverable errors. IMHO we should report them silently underneath. For non-recoverable errors I'm fully for adding an explicit `IOX_RAISE`!

MatthiasKillat:
Will be moved to another PR with the actual implementation.

PR #1400, opened 2022-06-15
iox-#1365 implement wrapper class for pthreads

### LINKED ISSUE #1365: POSIX wrapper for pthreads
## Brief feature description

We need a POSIX wrapper for `pthreads` to get rid of exceptions and undefined behavior. 
For posix platforms the wrapper shall support:

* Set stack address
     * https://man7.org/linux/man-pages/man3/pthread_attr_setstack.3.html
     * consider suggestions in #1400 
* Setting scheduler priority
    * https://man7.org/linux/man-pages/man3/pthread_setschedprio.3.html
* Set CPU affinity
    * https://man7.org/linux/man-pages/man3/pthread_setaffinity_np.3.html
* Set stack size
    * See, https://man7.org/linux/man-pages/man3/pthread_attr_setstack.3.html
    * https://man7.org/linux/man-pages/man3/pthread_attr_setstacksize.3.html
* The destructor shall call join by default, see `std::jthread` https://en.cppreference.com/w/cpp/thread/jthread
* Explicit thread termination:
    * https://man7.org/linux/man-pages/man3/pthread_exit.3.html

Optionally, it would be awesome when it supports

* https://man7.org/linux/man-pages/man3/pthread_equal.3.html
* https://man7.org/linux/man-pages/man3/pthread_self.3.html
* Set the scheduler
    * https://man7.org/linux/man-pages/man3/pthread_attr_setschedparam.3.html
 * Set contention scope:
     * https://man7.org/linux/man-pages/man3/pthread_attr_setscope.3.html
 * Set cleanup handlers
     * https://man7.org/linux/man-pages/man3/pthread_cleanup_push.3.html

## Tasks

* [x] implement Thread class with minimal functionality needed for current code base
* [x] make it work on Windows and Mac
* [ ] replace all `std::thread` occurrences
* [ ] implement must have features
* [ ] add awesome optional features

## Design considerations

* pass an optional to create method
    * current state
    * https://github.com/eclipse-iceoryx/iceoryx/pull/1400#discussion_r903375242
* or separate handle and storage
    * proposed API: https://github.com/eclipse-iceoryx/iceoryx/pull/1400#discussion_r899220308
    * optional: potential to forget reset on an early return, https://github.com/eclipse-iceoryx/iceoryx/pull/1400#discussion_r898375219
    * https://github.com/eclipse-iceoryx/iceoryx/pull/1400#discussion_r906079855
* callable with generic signature?
    * can be achieved with a lambda capturing the needed arguments in the current implementation
    * see also https://github.com/eclipse-iceoryx/iceoryx/pull/1400#discussion_r899256494
* movable threads?
    *  https://github.com/eclipse-iceoryx/iceoryx/pull/1400#discussion_r898890414
    * easier if thread handle and storage are separated

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## Notes for Reviewer
This draft PR introduces a wrapper class for the thread. Please note that the changes in `RouDi` and `PeriodicTask` are made just to provide a feeling on how the thread will be used and will be removed once we agreed on the design. I chose these two classes because they use the former free functions `*etThreadName` which are now methods of the Thread class.

The next steps are:
- merge thread class, with tests disabled for Windows and Mac
- make it work on Windows and Mac, enable tests
- replace all occurrences of `std::thread`
- extend thread class with functionality listed in #1365  

## References

- Partially closes #1365


iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
...
+        return cxx::error<ThreadError>(ThreadError::EMPTY_CALLABLE);
+    }
+
+    uninitializedThread.emplace();
+    uninitializedThread->m_callable = callable;
+
+    // create thread
+    auto createResult =
+        posixCall(pthread_create)(
+            &uninitializedThread->m_threadHandle, nullptr, Thread::startRoutine, &uninitializedThread->m_callable)
```
elfenpiff, line 25: Could you create a `const pthread_attr_t * threadAttributes = nullptr` and use it here instead of a `nullptr`.
In a follow up PR you can use this variable and set all the thread attributes but for now it would make the core more readble.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
...
+    if (!callable)
+    {
+        std::cerr << "The thread cannot be created with an empty callable." << std::endl;
+        return cxx::error<ThreadError>(ThreadError::EMPTY_CALLABLE);
+    }
+
+    uninitializedThread.emplace();
+    uninitializedThread->m_callable = callable;
+
+    // create thread
```
elfenpiff, line 22: I would remove that comment.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
...
+    auto createResult =
+        posixCall(pthread_create)(
+            &uninitializedThread->m_threadHandle, nullptr, Thread::startRoutine, &uninitializedThread->m_callable)
+            .successReturnValue(0)
+            .evaluate();
+    if (createResult.has_error())
+    {
+        uninitializedThread->m_isJoinable = false;
```
elfenpiff, line 30: Could you write this line like `uninitializedThread->m_isJoinable = !createResult.has_error();` and move it above the condition.
Also I would rename the variable to something like `m_isThreadConstructed`. The reason is that we do not distinguish joinable and detached threads with this variable but successfully and failed threads.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -17,8 +17,13 @@
 #define IOX_HOOFS_POSIX_WRAPPER_PTHREAD_HPP
 
 #include "iceoryx_hoofs/cxx/expected.hpp"
+#include "iceoryx_hoofs/cxx/function.hpp"
```
elfenpiff, line 20: Please update the copyright header.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
+#ifndef IOX_HOOFS_POSIX_WRAPPER_THREAD_INL
```
elfenpiff, line 1: Please add a copyright header.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -119,15 +119,15 @@ storable_function<S, signature<ReturnType, Args...>>::~storable_function() noexc
 }
 
 template <typename S, typename ReturnType, typename... Args>
-ReturnType storable_function<S, signature<ReturnType, Args...>>::operator()(Args... args)
```
elfenpiff, line 122: Copyright header.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -112,12 +112,12 @@ class storable_function<StorageType, signature<ReturnType, Args...>>
     ///          Not specifying move or using a default implementation is fine.
     ///          This is also the case for std::function (for the gcc implementation at least).
     ///
-    ReturnType operator()(Args... args);
+    ReturnType operator()(Args... args) const;
```
elfenpiff, line 115: Copyright header.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
+            {
+            case EDEADLK:
+                LogError() << "A deadlock was detected when attempting to join the thread.";
+                break;
+            default:
+                LogError() << "This should never happen. An unknown error occurred.";
+                break;
+            }
+        }
+        m_isJoinable = false;
```
elfenpiff, line 43: I think this is not needed here, after this line the object cease to exist

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
+    }
 }
 
-ThreadName_t getThreadName(pthread_t thread) noexcept
+void Thread::setThreadName(const ThreadName_t& name) noexcept
+{
+    posixCall(iox_pthread_setname_np)(m_threadHandle, name.c_str())
+        .successReturnValue(0)
+        .evaluate()
+        .or_else([](auto& r) {
```
elfenpiff, line 52: You can replace the whole `or_else` with `.expect("This should never happen! Failed to set thread name.");` You do not need to print `getHumanReadableErrnum` since the posix call is already doing this for you on failure.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
 
     return ThreadName_t(cxx::TruncateToCapacity, tempName);
 }
 
+ThreadError Thread::errnoToEnum(const int errnoValue) noexcept
+{
+    switch (errnoValue)
+    {
+    case EAGAIN:
+        LogError() << "insufficient resources to create another thread";
```
elfenpiff, line 86: I think it would be nice to add the thread name to every message if a thread name was set up. So something like `getThreadName` and when the string is not empty add something like:

```cxx
auto details = getThreadName();
if ( !details.empty() ) details = ", thread = \"" + details + "\"";
LogError() << "insufficient resources to create another thread" << details;
```

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
+    default:
+        LogError() << "an unexpected error occurred in thread - this should never happen! errno: "
+                   << strerror(errnoValue);
+        return ThreadError::UNDEFINED;
+    }
+}
+
+void* Thread::startRoutine(void* callable)
+{
+    // necessary because the callback signature for pthread_create is void*(void*)
```
elfenpiff, line 106: I would remove the comment. I think it is also out of date since you cast the `callable` argument to `void()` and not `void*(void*)`

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -35,66 +32,73 @@ class Thread_test : public Test
...
+
+    optional<Thread> sut;
 };
 
-#if !defined(__APPLE__)
+TEST_F(Thread_test, CreateThread)
+{
+    constexpr uint64_t MY_FAVORITE_UINT = 13;
+    function<void()> callable = [&] { EXPECT_THAT(test_function(MY_FAVORITE_UINT), Eq(MY_FAVORITE_UINT)); };
+    ASSERT_FALSE(ThreadBuilder().create(sut, callable).has_error());
```
elfenpiff, line 57: How do you make sure that the `callable` is called at all.
I would capture a `bool callbackWasCalled = false;` set it to true inside the callback and check this callback from the outside.
Something like:
```cxx
bool wasCallbackCalled = false;
function<void()> callable = [&] { EXPECT_THAT(test_function(MY_FAVORITE_UINT), Eq(MY_FAVORITE_UINT)); wasCallbackCalled = true; };
ASSERT_FALSE(ThreadBuilder().create(sut, callable).has_error());
sut.reset(); // join the thread so that we are sure that the callback is finished
EXPECT_TRUE(wasCallbackCalled);
```

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -35,66 +32,73 @@ class Thread_test : public Test
...
-    std::thread* m_thread;
+        return val;
+    }
+
+    optional<Thread> sut;
 };
 
-#if !defined(__APPLE__)
+TEST_F(Thread_test, CreateThread)
+{
```
elfenpiff, line 54: Test ids are missing in all tests.

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -17,9 +17,6 @@
 #include "iceoryx_hoofs/posix_wrapper/thread.hpp"
 #include "test.hpp"
 
-#include <atomic>
```
elfenpiff, line 20: Copyright year.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
 {
-    posixCall(iox_pthread_setname_np)(thread, name.c_str()).successReturnValue(0).evaluate().or_else([](auto& r) {
-        // String length limit is ensured through cxx::string
-        // ERANGE (string too long) intentionally not handled to avoid untestable and dead code
-        std::cerr << "This should never happen! " << r.getHumanReadableErrnum() << std::endl;
-        cxx::Ensures(false && "internal logic error");
-    });
+    if (m_isJoinable)
+    {
+        auto joinResult = posixCall(pthread_join)(m_threadHandle, nullptr).successReturnValue(0).evaluate();
```
elBoberido, line 30: Can you create a `#define NULLPTR_NULLPTR_NULLPTR_NULLPTR_NULLPTR nullptr;` and use it instead of `nullptr` in order to force clang-format to break the lines at the dot?

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
+        if (joinResult.has_error())
+        {
+            switch (joinResult.get_error().errnum)
+            {
+            case EDEADLK:
+                LogError() << "A deadlock was detected when attempting to join the thread.";
```
elBoberido, line 36: I guess the thread would keep running ... or maybe keep deadlocking, right?

The philosophical question would be if we should call the error handler? Worst case one could get out of memory if this happens repeatedly.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
elBoberido: We have to be careful with this. If this is called from `getThreadName` at some point, we might have infinite recursion.

It also does not make sense in this case since the only place where this is currently used is in the builder which would give us the wrong thread name.

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
elBoberido: Why have the check in the thread at all? Just checking that it was called should be sufficient, shouldn't it?

MatthiasKillat (commented):
Ok for first replacement of selected threads. In general we should support

1) `(void*)(void*)` for a posix thread wrapper (and maybe `void(void*)`, `void(void)` for convenience, but they are easily derived from the former)
2) an `std::thread` substitute should support general signature `R(Args...)`, but this is harder without dynamic memory and some proper type erasure.
3) support of detachable threads? Also non-trivial without proper ownership concept of the thread state, which includes the callable, stack and some meta-infromation.
4) construction which can fail and life time management of thread state. This is also related to creating a somewhat user friendly (semi)fool-proof API.
5) allow the thread handles to be lightweight and hence movable (while the thread is running, as it is just a pointer copy).

Those are all non-trivial problems, made harder due to lack of dynamic memory (and less so exceptions).
2)

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,10 +33,58 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+    INSUFFICIENT_PERMISSIONS,
+    INSUFFICIENT_RESOURCES,
+    INVALID_ATTRIBUTES,
+    UNDEFINED
+};
+
+class Thread
+{
+  public:
+    using callable_t = cxx::function<void()>;
```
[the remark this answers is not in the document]
elfenpiff: Since we use here an alias we can add this any time in the future with adjusting a single line. For the moment I would pursue the simplest approach and see how far we get with this.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,10 +33,58 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+
+    Thread(const Thread&) = delete;
+    Thread& operator=(const Thread&) = delete;
+    Thread(Thread&& other) = delete;
```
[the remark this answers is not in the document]
elfenpiff: This goes also for `semaphores` and `mutex` and here the solution would be to acquire some memory outside of the class and store the object in there.
One could change this but then we have to solve a race condition at the start which makes this object much more complex and the benefit is litte. For us it would be actually non existing since we do not have this use case in iceoryx.

Here the problem:
```
posix::Thread t(someFunction);
posix::Thread t2 = std::move(t); // what if the thread is started but the function argument is not yet copied or called inside of `startRoutine`.
```
We could of course make the ctor blocking until the callable is copied and started inside of `startRoutine`. But how? We could add a semaphore (which can fail an then we have to add real complex error handling). A busy loop which burns the CPU?

Or we leave it as it is for now and see if we really need this feature.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,10 +33,58 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+
+    pthread_t m_threadHandle;
+    callable_t m_callable;
+    bool m_isJoinable{true};
+};
+
+class ThreadBuilder
+{
+  public:
+    cxx::expected<ThreadError> create(cxx::optional<Thread>& uninitializedThread,
```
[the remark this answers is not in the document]
elfenpiff: Could you explain heavy in what sense? Creating a thread costs a lot of performance and the optional which is provided by reference or the enum inside an expected is neglectable in this context.

elfenpiff:
@MatthiasKillat 

> 2.) We cast it to a callable_t* (this should only be valid if it actually is of the type already or derives from it, but why have the void* argument then?).

This is enforced by the posix standard. See: https://man7.org/linux/man-pages/man3/pthread_create.3.html

> I need to read on to see the full picture but I am not convinced this is safe in the general case nor does is it general enough to provide C type erased arguments 

Could you please elaborate this in detail.

> (let alone generic ones of symbolic signature R(Args...)).

This is covered since one can always use a lambda and capture the arguments which are required. Not with the full flexibility we have with dynamic memory but with the flexibility we need for our use case.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,10 +33,58 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+
+  private:
+    Thread() noexcept = default;
+
+    static ThreadError errnoToEnum(const int errnoValue) noexcept;
+
+    static void* startRoutine(void* callable);
+
+    pthread_t m_threadHandle;
+    callable_t m_callable;
```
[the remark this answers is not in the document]
elfenpiff: We should never support detached threads with a posix thread wrapper. The reason is that a posix wrapper ties the resource to its lifetime and we want to explicitly use RAII.

When we support detached threads then the lifetime of the thread is decoupled from the lifetime of the posix wrapper and this can introduce serious issues and race conditions as we have seen in the ara binding.
Then one can easily overlook that a thread is still running and one has to introduce some other mechanism to manage the lifetime.

If this is really required some day I would propose a thread builder for detached threads which simply returns an expected which states if this worked or not.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.hpp
```diff
@@ -123,7 +123,7 @@ class PeriodicTask
     units::Duration m_interval{units::Duration::fromMilliseconds(0U)};
     /// @todo use a refactored posix::Timer object once available
     posix::Semaphore m_stop{posix::Semaphore::create(posix::CreateUnnamedSingleProcessSemaphore, 0U).value()};
-    std::thread m_taskExecutor;
+    cxx::optional<posix::Thread> m_taskExecutor;
```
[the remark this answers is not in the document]
MatthiasKillat: Yes the alias is ok. The problem is that potentially invalid objects (PIO) are viral. If we use an such a PIO in another object, it becomes a PIO itself etc. I am not sure how to deal with this in general and we should strive to avoid them.

But low level APIs may fail and we cannot ignore this.

Ideally we have something during construction which may fail to construct the thread (a `std::thread` would throw but we cannot). But we can terminate or call some error handler.

So we could have a thread handle that cannot be invalid and we assign it only if the construction succeeded. Somehow like this (pseudo code) in the ctor:

```
using maybe_thread_handle_t = cxx::optional<posix::Thread>; // technically no handle yet ...
auto maybeThreadHandle = createThread(...);
if(!maybeThreadHandle) {
   terminate();
}
m_thread = maybeThreadHandle.value();
```

If we do not terminate, the invariant holds that m_thread is a valid handle (unless we reassign and all this but we could avoid this by wrapping sme of the code in a function that either creates a valid handle from a maybe invalid handle or terminates, leaving no other option to create it (only move it later).

Can also be done in the initializer list with e.g. immediately invoked lambdas (not so nice syntactically but this is the price).

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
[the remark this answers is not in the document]
MatthiasKillat: True, the thread creation is costly and I misunderstood the purpose of the optional here: it is actually used to provide some storage to construct the thread in, even if the optional is invalid. So this part of my comment is wrong, it makes sense to provide an invalid optional (as it still has the aligned memory we need).

Basically it is the thread storage and we could have an alias like

```using thread_storage_t = cxx::optional<Thread>;```

For me this helps to create a mental model about what is happening.

File: iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp (hunk 7 above)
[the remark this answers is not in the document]
Comment (elBoberido) [reply iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp:115]: @MatthiasKillat maybe we need to add a test for the scenario you described. If that does not break, then I see no reason to revert the change

File: iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.hpp (hunk 21 above)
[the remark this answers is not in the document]
Comment (elBoberido) [reply iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.hpp:126]: Exactly. That's why my personal ideal API would be
```cpp
thread_foo_t tft;
thread t { createThread(tft, ...).expect("Aaaaaaaaaaargh!") };
```
or
```cpp
thread_foo_t tft;
thread t { createThread(new thread_foo_t, ...).expect("Aaaaaaaaaaargh!") };
```
And at some point in the future with a better in process memory management just a
```cpp
thread t { createThread(...).expect("Aaaaaaaaaaargh!") };
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
...
+#define IOX_HOOFS_POSIX_WRAPPER_THREAD_INL
+
+#include "iceoryx_hoofs/posix_wrapper/thread.hpp"
+
+namespace iox
+{
+namespace posix
+{
+inline cxx::expected<ThreadError> ThreadBuilder::create(cxx::optional<Thread>& uninitializedThread,
+                                                        const cxx::function<void()>& callable) noexcept
```
MatthiasKillat, line 11: If this is intended to be called by the user we need to consider t:
`cxx::function` just as `std::function` is more of a storage type instead of a lightweight "pass-around" type, even if we have a reference here.
For example `std::thread` passes its callable via template parameter (supporting a variadic signature) https://en.cppreference.com/w/cpp/thread/thread/thread, but we do not need this in the same generality. 

We just need to pass something that is invocable with `(void)(void)` signature. This can be done with the following template method```
template< class Function>
inline cxx::expected<ThreadError> ThreadBuilder::create(cxx::optional<Thread>& uninitializedThread, Function &&f);
```

The benefit is that there is no need to construct a `cxx::function` when passing (this will implicitly happen unless our argument is already a `cxx::function`, which generally will not be the case.This also happens if we pass by reference, it will just construct a temporary and pass its reference.

Now this is not completely save, we would need to use `is_invocable_r` with the desired signature. This would make the later check whether the callable is null obsolete as well, since now everything is checked at compile time (another benefit).

File: iceoryx_hoofs/source/posix_wrapper/thread.cpp [hunk 23]
```diff
@@ -16,38 +16,97 @@
...
+        LogError() << "an unexpected error occurred in thread - this should never happen! errno: "
+                   << strerror(errnoValue);
+        return ThreadError::UNDEFINED;
+    }
+}
+
+void* Thread::startRoutine(void* callable)
+{
+    // necessary because the callback signature for pthread_create is void*(void*)
+    callable_t f = *static_cast<callable_t*>(callable);
```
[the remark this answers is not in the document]
MatthiasKillat: I looked closer at how it is invoked in `ThreadBuilder::create` with `&uninitializedThread->m_callable` and this is ok. I was under the impression we want to somehow support `void*` arguments (as a native pthread) and this caused me to make some mistake but we do not as of know.

To summarize:
1) We type erase with `void*` as `pthread_create` requires this.
2) We always pass `&uninitializedThread->m_callable`, which is stored as part of the object and always a `cxx::function<void(void)>`.
3) We wrap this locally to `f`, also a `cxx::function<void(void)>` (here I am not entirely sure whether it is a copy or wrapping the callable, but I think it is the latter).
4) We execute `f`.

The local copy is unfortunate though (or double wrap, see 3)). If it is a double wrap it is even more loss of performance (indirections again). 

Can you explain why it is needed? We should not move the thread while it is created and I assume `pthread_create` is blocking until everything is copied to it. 

It would be a problem if the thread is moved later than we will have a problem if we do not create this copy since the running thread would access the now invalid callable we moved from. But it cannot be moved as of now, so this cannot be it...

Considering this, I think we would still have a race here if we move after the thread was started, even with the copy to f.

1) pthread invoked Thread::startRoutine
2) copy to callable to `f` is initated but not yet completed
3) thread wrapper is moved
4) source of move is destroyed
5) copy to f resumes from an the destroyed callable (undefined)

TLDR: 
It is not wrong but we may be able to optimize it if we store the thread state elsewhere. I also think we do not need the local copy if we cannot move the thread but I may have overlooked something critical (it is fairly complex). I will have to rethink it some more and require more detail.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.inl
```diff
@@ -17,6 +17,8 @@
 #ifndef IOX_HOOFS_CONCURRENT_PERIODIC_TASK_INL
 #define IOX_HOOFS_CONCURRENT_PERIODIC_TASK_INL
 
+#include "iceoryx_hoofs/internal/concurrent/periodic_task.hpp"
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: The changes is this file are made just to provide a feeling on how the thread would be used and will be removed once we agreed on the design. Therefore, I won't change the copyright date.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
+#ifndef IOX_HOOFS_POSIX_WRAPPER_THREAD_INL
+#define IOX_HOOFS_POSIX_WRAPPER_THREAD_INL
+
+#include "iceoryx_hoofs/posix_wrapper/thread.hpp"
+
+namespace iox
+{
+namespace posix
+{
+inline cxx::expected<ThreadError> ThreadBuilder::create(cxx::optional<Thread>& uninitializedThread,
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Actually `create` was once a template function to have a more generic callable signature which didn't work out and I simply forgot to move the function to the cpp file.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,10 +33,58 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+    using callable_t = cxx::function<void()>;
+
+    Thread(const Thread&) = delete;
+    Thread& operator=(const Thread&) = delete;
+    Thread(Thread&& other) = delete;
+    Thread& operator=(Thread&& rhs) = delete;
+
+    ~Thread() noexcept;
+
+    // @todo set name during creation and remove method
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: The name is optional. By default a thread inherits the program name if it is not set with `pthread_setname_np`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: `test_cxx_function.cpp` provides a test class `Functor` which state is changed when `operator()` is called and all the corresponding tests pass.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -16,38 +16,97 @@
...
+void Thread::setThreadName(const ThreadName_t& name) noexcept
+{
+    posixCall(iox_pthread_setname_np)(m_threadHandle, name.c_str())
+        .successReturnValue(0)
+        .evaluate()
+        .or_else([](auto& r) {
+            // String length limit is ensured through cxx::string
+            // ERANGE (string too long) intentionally not handled to avoid untestable and dead code
+            LogError() << "This should never happen! " << r.getHumanReadableErrnum();
+            cxx::Ensures(false && "internal logic error");
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Yes, maybe it is too harsh but the error handling here has to be reworked anyway since the possible `open` errors are not considered. Therefore, I'd like to postpone the decision on how to react when setting the thread name fails until this is implemented.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @MatthiasKillat Your summary is correct. The local copy is not needed anymore and I removed it.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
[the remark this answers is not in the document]
MatthiasKillat: @elboberido The test exists and does not break because the memory that is modified by the functor (and owned by function) is type erased, so the compiler cannot know/prove that it is modified (or not modified).

I think the usage problem arises if we have a `const cxx::function` and want to call `operator()`. To resolve this, we can make `operator()` `const` but this is mainly a workaround. I see no easy way to get proper `const overloading` where the methods are actually different.

The reason for this is basically that at construction time it needs to be checked whether it is a functor with state (meaning its `operator()` cannot be const. For these the wrapping `cxx::function` would need to allow only non-const `operator()`.

TLDR: this may need complex case distinctions in `cxx::function` to really get right (if this is even possible). If for now it is sufficient and necessary to make it `const` we can do so.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
...
+                                                        const cxx::function<void()>& callable) noexcept
+{
+    if (!callable)
+    {
+        std::cerr << "The thread cannot be created with an empty callable." << std::endl;
+        return cxx::error<ThreadError>(ThreadError::EMPTY_CALLABLE);
+    }
+
+    uninitializedThread.emplace();
+    uninitializedThread->m_callable = callable;
```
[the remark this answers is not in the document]
MatthiasKillat: Addendum: If `cxx::function` double wraps a `cxx::function``when constructed with one it is non-intended behavior of xx::function` (a failure of overload resolution). It should call the copy ctor and basically copy the state of the source and all its type erased information.

Afterwards the destination is an independent copy of the original function (with the actual callable being type-erased of course). So long as we call the copy ctor and do not move (or copy) while the thread is running it should be alright.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
elfenpiff: I would like to provide as much information as possible to the user when something goes wrong. 

We could add a comment stating that `getThreadName` never calls this (inside the implementation). When we implement MC/DC tests for this class this case would be covered and so it never should occur in theory.

We could also store the thread name in a separate member variable and since it is quite short it should not blow up the object.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.inl
```diff
@@ -53,24 +55,25 @@ inline void PeriodicTask<T>::start(const units::Duration interval) noexcept
 {
     stop();
     m_interval = interval;
-    m_taskExecutor = std::thread(&PeriodicTask::run, this);
-    posix::setThreadName(m_taskExecutor.native_handle(), m_taskName);
+    cxx::function<void()> callable(*this, &PeriodicTask::run);
+    posix::ThreadBuilder().create(m_taskExecutor, callable).expect("PeriodicTask::start : Failed to create thread");
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Do you mean that the callable can have a generic signature? This is not that easy without using dynamic memory (or at least I don't see it at the moment). But you can use a lambda and invoke a callable with a generic signature.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.hpp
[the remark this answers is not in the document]
elfenpiff: @elBoberido @MatthiasKillat the more I think about it the more I am against separating storage from handle in this fashion. The problem is in my opinion ownership and lifetime.

## The solution I prefer, handle and storage combined in one Thread object

First, I assume that the final posix thread wrapper has an interface to set external stack memory so that you can use it like this:
```cxx
cxx::optional<posix::Thread> myThread;
ThreadBuilder().setStackMemory(somePtr, memorySize)
               .create(myThread, someCallback)
               .expect("Aaarrgh!");
```

Or even better that `posix::Thread` comes with an optional template parameter where the stack size is part of the object  so that you can use it like this:
```cxx
cxx::optional<posix::Thread<StackSizeInMb<10>> myThread;
ThreadBuilder().create(myThread, someCallback)
               .expect("Aaarrgh!");
```

When one would like to store the posix thread somewhere else this is easy peasy, just use
```
using threadHandle_t = cxx::optional<posix::Thread<StackSizeInMb<10>>
threadHandle_t * myThread = new threadHandle_t();
// or
threadHandle_t * memory = new (allocateSomeMemory(sizeof(threadHandle_t))) threadHandle_t();
```

The big advantage is, the thread owns the stack memory and ensures its lifetime.

## Solution with separate storage and handle

But when we would pursue an approach which separates storage from handle similar to this:
```cxx
threadStorage_t threadStorage;
threadHandle_t threadHandle = ThreadBuilder().create(...).expect("Aarrgh!");
```
We have the following problems where we do not have an answer to and which increases the code complexity when solving them:
1. Lifetime, what happens when `threadStorage` goes out of scope before `threadHandle`
   * undefined behavior, this can be ugly
   * `threadStorage` knows that it is in use
     * calls thread_join - which may surprises the users and can lead easily to deadlocks and would also require that the `threadHandle` knows the `threadStorage` so that it can release the memory when they go out of scope in the right order.
     * calls terminate - we had this in the first waitset and this just makes the code unnecessarily unstable in my opinion
2. Ownership, those two objects are tightly coupled together and I would strongly refrain from using `threadStorage` for different threads but it would be legal. I would be fine with restarting the same thread over and over again if this would be a use case but completely different threads would be a pain. In the end I would always couple those together like
```cxx
struct ThreadStuff {
  threadStorage_t storage;
  threadHandle_t handle;
};
```
3. Ownership the second, what happens when a second thread is created in the `threadStorage`? Terminate, thread_join or thread_exit of the existing thread, undefined behavior? Since `threadHandle` is independent of `threadStorage` we cannot check if `threadHandle` is set since a user maybe initialized a different handle. A solution would be then to check if the `threadStorage` is already in use. So it would not be as easily usable as suggested, one would have to write stuff like:
```cxx
threadStorage_t threadStorage;
threadHandle_t threadHandle;
//...
if ( threadStorage.isNotInUse() ) {
  threadHandle = ThreadBuilder().create(...).expect("Aaargh!");
} else {
  errorHandler(THREAD_STORAGE_NOT_AVAILABLE);
}
```

## Summary

In my opinion separating thread storage from handle has lifetime and ownership issues which can be solved but make the thread implementation much more complex and the usage more error prone and also unnecessary complex.

I also have a hard time seeing the advantage of separating storage from handle except that the handle would be movable which could be also solved by introducing a semaphore into the thread which signals when the `startRoutine` has copied the callable.
But we have no where in iceoryx the use case that the thread must be movable. And a user could solve this easily by storing it in the heap, and we could solve this easily by adding an allocator to our posh runtime (or somewhere else) where we already have plans for.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.inl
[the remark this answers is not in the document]
elBoberido: I just couldn't resist :laughing:

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.inl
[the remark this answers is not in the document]
elBoberido: yes, I meant the callable with the generic signature

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/thread.inl
```diff
@@ -0,0 +1,40 @@
...
+inline cxx::expected<ThreadError> ThreadBuilder::create(cxx::optional<Thread>& uninitializedThread,
+                                                        const cxx::function<void()>& callable) noexcept
+{
+    if (!callable)
+    {
+        std::cerr << "The thread cannot be created with an empty callable." << std::endl;
+        return cxx::error<ThreadError>(ThreadError::EMPTY_CALLABLE);
+    }
+
+    uninitializedThread.emplace();
```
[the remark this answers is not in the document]
elBoberido: This shows again some shortcomings with the API to pass an `optional<thread>` to the builder. It is a source of a potential nullable type in an optional which has a value. This could happen on an early return when another feature is added and the developer forgets to reset the `optional`.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,10 +33,58 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+    pthread_t m_threadHandle;
+    callable_t m_callable;
+    bool m_isJoinable{true};
+};
+
+class ThreadBuilder
+{
+  public:
+    cxx::expected<ThreadError> create(cxx::optional<Thread>& uninitializedThread,
+                                      const cxx::function<void()>& callable) noexcept;
```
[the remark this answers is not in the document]
elBoberido: @FerdinandSpitzschnueffler and this is fine for now but in the mid to long term we might have to rethink this. Once it is possible to set the stack address, there are additional options how this could be handled.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/periodic_task.hpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @MatthiasKillat @elBoberido @elfenpiff Would you agree to merge the current changes (once the CI is green) as a first replacement for the std::thread occurrences? After making it work on Windows and Mac and replacing std::thread, I would continue with implementing a way to set the stack address and reconsider all your comments and suggestions. I think the current implementation doesn't hamper further design approaches and should be easily expandable/adaptable.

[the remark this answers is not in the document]
elBoberido: @FerdinandSpitzschnueffler agreed

@elfenpiff these are valid concerns but the two approaches are not that different.

Considering that something like `cxx::optional<posix::Thread<StackSizeInMb<10>> myThread;` won't work since most probably there is not enough stack on your current thread to put the storage of the new thread onto that stack, all of your concerns regarding the separate storage from the handle can be solved the same way you propose to solve the stack problem when it is part of the handle.

```cpp
iox::cxx::unique_ptr<threadStorage_t> ts {new threadStorage_t};
auto myThread = ThreadBuilder(std::move(ts))
           .create(someCallback)
           .expect("Aaarrgh!");
```

The nice thing about this is that we do not need to check if the `optional` contains a value, since there is no `optional`. Of course we can argue that we abort when `create` fails but what about this

```cpp
cxx::optional<posix::Thread> myThread;
ThreadBuilder().setStackMemory(somePtr, memorySize)
               .create(myThread, someCallback)
               .or_else("Aaarrgh!");
...
myThread.reset();
...
myThread.getName(); // kaboom
```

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @elfenpiff @elBoberido @MatthiasKillat I summarized and linked the main discussions in #1365 so that nothing is forgotten.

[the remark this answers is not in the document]
elBoberido: @FerdinandSpitzschnueffler each type of API has it drawbacks and it all boils down to which drawbacks make us less uncomfortable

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @elBoberido Agreed. That's why I added the arguments for both to the issue.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,8 +33,64 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+    Thread& operator=(Thread&&) = delete;
+
+    ~Thread() noexcept;
+
+    // @todo set name during creation and remove method
+    void setName(const ThreadName_t& name) noexcept;
+
+    /// @brief Returns the name of the thread
+    /// @return An iox::cxx::string containing the name of the thread
+    ThreadName_t getName() noexcept;
```
elfenpiff, line 67: This should be a const method. And when the name can only be set at creation time we can also return here a const reference, so
```cxx
const ThreadName_t & getName() const noexcept;
```

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/thread.hpp
```diff
@@ -28,8 +33,64 @@ constexpr uint64_t MAX_THREAD_NAME_LENGTH = 15U;
...
+{
+    EMPTY_CALLABLE,
+    INSUFFICIENT_MEMORY,
+    INSUFFICIENT_PERMISSIONS,
+    INSUFFICIENT_RESOURCES,
+    INVALID_ATTRIBUTES,
+    UNDEFINED
+};
+
+class Thread
```
elfenpiff, line 50: Could you add a doxygen documentation for the class and add a code snippet explaining how to use it.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -49,5 +50,111 @@ ThreadName_t getThreadName(pthread_t thread) noexcept
     return ThreadName_t(cxx::TruncateToCapacity, tempName);
 }
 
+cxx::expected<ThreadError> ThreadBuilder::create(cxx::optional<Thread>& uninitializedThread,
+                                                 const Thread::callable_t& callable) noexcept
+{
+    if (!callable)
```
elfenpiff, line 56: Not part of this PR but we really should get rid of nullable callables.

iceoryx_hoofs/source/posix_wrapper/thread.cpp
```diff
@@ -49,5 +50,111 @@ ThreadName_t getThreadName(pthread_t thread) noexcept
...
+    {
+        std::cerr << "The thread cannot be created with an empty callable." << std::endl;
```
elfenpiff, line 58: Could you please use `LogError` instead of `std::cerr` in the whole file. You gain access by including `iceoryx_hoofs/internal/log/hoofs_logging.hpp`

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -35,64 +34,75 @@ class Thread_test : public Test
...
-        }
-    };
-
-    std::atomic_bool m_run{true};
-    std::thread* m_thread;
+    optional<Thread> sut;
 };
 
-#if !defined(__APPLE__)
+#if !defined(_WIN32) && !defined(__APPLE__)
```
elfenpiff, line 50: I think only the set and get name does not work on apple but the thread itself should be working. Could you run those tests also on apple?

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -35,64 +34,75 @@ class Thread_test : public Test
...
+TEST_F(Thread_test, CreateThreadWithNonEmptyCallableSucceeds)
```
elfenpiff, line 51: Could you also add a tests where it is verified that the dtor calls join. It could look like.
1. Start a thread A which waits until thread B signals it that it goes out of scope
2. Start thread B whose callable blocks (you can yous `iceoryx_hoofs/testing/Barrier.hpp` for that)
3. Now let B go out of scope, this should be blocking so you start the time measurement directly before B goes out of scope.
4. Now let thread A wait  a given amount of time T and then signal the blocking callback that it can continue (the Barrier). Then thread B should be able to join.
5. Now measure the time, the time must be longer then T.

What we tested here is that the dtor of B blocks because of the join call. So the test could be called dtor of thread blocks until callback is finished.

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
FerdinandSpitzschnueffler: I didn't implement create and join yet. This will be done in the next PR.

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -35,66 +38,120 @@ class Thread_test : public Test
...
+
+    ASSERT_FALSE(ThreadBuilder()
+                     .create(sut,
+                             [&] {
+                                 start = std::chrono::steady_clock::now();
+                                 std::this_thread::sleep_for(std::chrono::nanoseconds(TEST_WAIT_TIME.toNanoseconds()));
+                                 threadSync.notify();
+                             })
+                     .has_error());
+
```
elfenpiff, line 102: I think you forgot one line here, `blockingThread.reset()` since we would like to test if the dtor of the blocking thread blocks until the callback was finished.

But I may have an idea how to handle this all with only one thread:
```cxx
optional<Thread> blockingThread;
ASSERT_FALSE(ThreadBuilder()
                     .create(blockingThread,
                             [&] {
                                 threadSync.wait();
                                 std::this_thread::sleep_for(std::chrono::nanoseconds(TEST_WAIT_TIME.toNanoseconds()));
                             })
                     .has_error());

// since we want to ensure only roughly that the dtor blocks at least TEST_WAIT_TIME
// we can start the measurement already a little ahead right before we inform the thread and trigger the dtor
start = std::chrono::steady_clock::now(); 
threadSync.wait();
blockingThread.reset(); // destroy the object, ensure that this is blocking
end = std::chrono::steady_clock::now();

EXPECT_THAT(std::chrono::nanoseconds(end - start).count(), Gt(TEST_WAIT_TIME.toNanoseconds()));
```

FerdinandSpitzschnueffler: @elfenpiff You're right, I forgot the reset call but I replaced the test with your suggestion. Thanks!

iceoryx_hoofs/source/posix_wrapper/thread.cpp
elfenpiff: @mossmaurice thanks for the little research. 

I think we gain a lot of flexibility when we using our logger. `std::cout` prints only to console (of course you can redirect it but then you maybe also redirect the `cout` messages of top level applications which they do not want).

elfenpiff (dismissed):

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
```diff
@@ -35,66 +38,113 @@ class Thread_test : public Test
...
+//{
+// return a;
+//}
+// TEST_F(Thread_test, Blub)
+//{
+// iox::cxx::function<void()> callable = [&]() { someFunction(3); };
+// iox::cxx::function<void()> callable2;
+// optional<Thread> myThread;
+// ThreadBuilder().create(myThread, callable2).expect("Couldn't create a thread.");
+//}
```
elBoberido, line 148: Do you want to keep this?

elBoberido (approved):
Looks good to me. If the other reviewer are nor unhappy with this it could be merged.

iceoryx_hoofs/test/moduletests/test_posix_thread.cpp
FerdinandSpitzschnueffler: Whoops... I removed these lines since I had to rebase anyway.

elBoberido (approved):

iceoryx_hoofs/source/posix_wrapper/thread.cpp
MatthiasKillat: That is debatable and I disagree for the purpose of storage (what `std::function` and `cxx::function` is about). The issue here is that it is used for passing callable, which is not its primary purpose. This could be avoided with a template function as for `std::thread` which will make it impossible to pass anything that is not callable.

This requires distinguishing some cases in generic programming but is the way to go IMO.

PR #1418, opened 2022-06-24
iox-#1036 builder pattern for file lock

### LINKED ISSUE #1036: Remove creation design pattern class with in place implementation

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
The FileLock had some issues. For instance it never verified that the path prefix was a valid path (syntax) and it hardcoded the permissions. I added those as builder parameters and verified them.
To verify the path I had to add a variation of `isValidFilePath` called `isValidPath` which just does not check if the last character is a separator. The tests are also more or less the same with some minor addition (adding a last slash).

## References

- Closes #1036


doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -139,3 +141,17 @@
...
+6. `FileLock` uses the builder pattern. Path and permissions can now be set.
+    ```cpp
+    // before
+    auto fileLock = iox::posix::FileLock::create("lockFileName")
+                        .expect("Oh no I couldn't create the lock file");
+
+    // after
+    auto fileLock = iox::posix::FileLockBuilder().name("lockFileName")
+                                                 .path("/Now/I/Can/Add/A/Path")
+                                                 .permissions(iox::cxx::perms::owner_all)
```
FerdinandSpitzschnueffler, line 154:
```suggestion
                                                 .permission(iox::cxx::perms::owner_all)
```

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
```diff
@@ -48,27 +50,24 @@ enum class FileLockError
...
     static constexpr int32_t INVALID_FD = -1;
     static constexpr const char LOCK_FILE_SUFFIX[] = ".lock";
     static constexpr uint64_t FILENAME_LENGTH = platform::IOX_MAX_FILENAME_LENGTH
                                                 - sizeof(platform::IOX_LOCK_FILE_PATH_PREFIX) / sizeof(char)
                                                 - sizeof(LOCK_FILE_SUFFIX) / sizeof(char);
 
     using FileName_t = cxx::string<FILENAME_LENGTH>;
-    using PathName_t = cxx::string<platform::IOX_MAX_PATH_LENGTH>;
+    using FilePath_t = cxx::string<platform::IOX_MAX_PATH_LENGTH>;
+    using PathName_t = cxx::string<platform::IOX_MAX_PATH_LENGTH - FILENAME_LENGTH - 1>;
```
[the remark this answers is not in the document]
elfenpiff: You described it perfectly!

But the path always ends with a slash and if there is none one will be added.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
```diff
@@ -78,21 +77,40 @@ class FileLock : public DesignPattern::Creation<FileLock, FileLockError>
...
+    friend class FileLockBuilder;
+    FileLock(const int32_t fileDescriptor, const FilePath_t& path) noexcept;
 
     void invalidate() noexcept;
 
-    cxx::expected<FileLockError> initializeFileLock() noexcept;
-    FileLockError convertErrnoToFileLockError(const int32_t errnum) const noexcept;
-    cxx::expected<FileLockError> closeFileDescriptor() noexcept;
+    static FileLockError convertErrnoToFileLockError(const int32_t errnum, const FilePath_t& fileLockPath) noexcept;
+    cxx::expected<FileLockError> closeFileDescriptor(const int32_t fileDescriptor) noexcept;
```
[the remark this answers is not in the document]
elfenpiff: During the refactoring it was a static method ... I removed the argument again.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
```diff
@@ -48,27 +50,30 @@ enum class FileLockError
...
+    /// @brief A file name without any containing slash (path separator)
+    ///        For instance "myLock.lock"
     using FileName_t = cxx::string<FILENAME_LENGTH>;
-    using PathName_t = cxx::string<platform::IOX_MAX_PATH_LENGTH>;
+    /// @brief The full path to the file, including the file
+    ///        For instance "/path/to/myLock.lock"
+    using FilePath_t = cxx::string<platform::IOX_MAX_PATH_LENGTH>;
+    /// @brief The directory to the file
+    ///        For instance "/path/to/"
+    using PathName_t = cxx::string<platform::IOX_MAX_PATH_LENGTH - FILENAME_LENGTH - 1>;
```
elBoberido, line 76: I already wrote a lengthy comment that there is a off by one error if `m_path` uses the full size and has not path separator and `m_name` also uses the full size. Then I noticed that the `-1` is not for null termination but for the path separator.

Please use a `constexpr` ... and you owe me 2 minutes of my life :laughing:

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
```diff
@@ -31,50 +32,68 @@ namespace posix
...
+
+    FileLock::FilePath_t fileLockPath = m_path;
+
+    if (!cxx::doesEndWithPathSeparator(fileLockPath))
+    {
+        fileLockPath.unsafe_append(iox::platform::IOX_PATH_SEPARATORS[0]);
+    }
+
+    fileLockPath.unsafe_append(m_name);
+    fileLockPath.unsafe_append(FileLock::LOCK_FILE_SUFFIX);
```
elBoberido, line 57: ... and here I'm able to reuse my reasoning so you already payed your dept. The size of `FileLock::LOCK_FILE_SUFFIX` is not part of the size of `FilePath_t` and might not be appended

elBoberido (commented):
I did not yet review the new tests for the helplets

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
```diff
@@ -64,23 +66,20 @@ inline bool isValidFileName(const string<StringCapacity>& name) noexcept
 template <uint64_t StringCapacity>
 inline bool isValidFilePath(const string<StringCapacity>& name) noexcept
 {
-    if (name.empty())
+    if (doesEndWithPathSeparator(name))
```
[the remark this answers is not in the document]
elfenpiff: I looked at it and decided against implementing something like path. In my opinion it feels like an extended string class with some methods for strings which represents paths and this is something we can fully cover with free functions.

`is_directory` and `is_regular_file` actually determine if the entity exists and if it is a file or directory. Something I would like to add to `cxx::filesystem` at some point. But in this case I look at the string itself and do not care at this stage if it exists or if it is a directory or file. I just would like to analyse the syntax of the string.

But it would make sense to move those free functions in the filesystem header. If you wish I can do this in this PR or in a follow up one.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
elfenpiff: I will ask Hypnotoad if it grants you two additional minutes in your life.

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
elfenpiff: May your 2 minutes be taken again with an additional 50%processing fee. ;)

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Now I think my definition for `FileName_t` is wrong. It shouldn't include the `.lock` suffix, right?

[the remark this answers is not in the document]
elfenpiff: You are right, I adjusted the explanatory comment.

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -150,6 +152,7 @@
 
    delete soSmart; // <- not possible anymore
    ```
+
```
elBoberido, line 155: I guess this superfluous line came in with the merge. In case you need to add further commits to this PR, please remove it but a commit just for this is nor necessary.

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
```diff
@@ -78,21 +86,40 @@ class FileLock : public DesignPattern::Creation<FileLock, FileLockError>
...
-    FileLockError convertErrnoToFileLockError(const int32_t errnum) const noexcept;
+    static FileLockError convertErrnoToFileLockError(const int32_t errnum, const FilePath_t& fileLockPath) noexcept;
     cxx::expected<FileLockError> closeFileDescriptor() noexcept;
+};
 
-    friend class DesignPattern::Creation<FileLock, FileLockError>;
+class FileLockBuilder
+{
+    /// @brief Defines the file name of the lock
+    IOX_BUILDER_PARAMETER(FileLock::FileName_t, name, "")
```
elBoberido, line 111: Can you extend the documentation and state that `LOCK_FILE_SUFFIX` will be appended to the name

iceoryx_hoofs/include/iceoryx_hoofs/posix_wrapper/file_lock.hpp
```diff
@@ -78,21 +86,40 @@ class FileLock : public DesignPattern::Creation<FileLock, FileLockError>
...
+
+    /// @brief Defines the path where the lock is stored
+    IOX_BUILDER_PARAMETER(FileLock::PathName_t, path, platform::IOX_LOCK_FILE_PATH_PREFIX)
+
+    /// @brief Defines the access permissions of the file lock
+    IOX_BUILDER_PARAMETER(cxx::perms, permission, cxx::perms::none)
```
elBoberido, line 117: Could you maybe tell the default values in the documentation in case these functions are not used. Maybe just like `used the platform standard location by default` and `has no permissions by default`

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
elBoberido: Then you can keep your truncation :laughing: 

Could you add a test for this?

elBoberido (commented):
helplets tests still not reviewed

doc/website/release-notes/iceoryx-unreleased.md
elfenpiff: No it is not superfluous I added it explicitly. I like to have a empty line between each entry of an enumeration.  When you take at the previous release notes you see it is consistent, also in this files with the entries above.

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
elfenpiff: > Could you add a test for this?

Please don't. The problem is that I have to create a max path length so I have to create a new directory which has the maximum length with `mkdir` and remove it later with `rmdir`.
Those functions are of course available in Windows with slightly different parameters and are called there `_mkdir` and `_rmdir`.

I could of course some dirty `ifdef`s or implement the c++17 https://en.cppreference.com/w/cpp/filesystem/create_directory and use it here.

Would you be fine when I create an issue for create and remove directory and add a task to implement this unit test when it is available?

doc/website/release-notes/iceoryx-unreleased.md
elBoberido: Aaargh, it looked like there was already an empty line but it's the line with the backticks

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
```diff
@@ -507,13 +513,163 @@ TEST(Helplets_test_isValidFilePath, WhenOneInvalidCharacterIsContainedPathIsInva
...
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("/asda.")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("/bla/../fuu/asda..")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("/bla/./.././xa..")));
+}
+
+TEST(Helplets_test_isValidPath, PathWhichContainsAllValidCharactersIsValid)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "8052b601-c9ad-4cb8-9a87-c301f213d8c4");
+    EXPECT_TRUE(isValidPath(
+        string<FILE_PATH_LENGTH>("/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/0123456789/-.:_")));
```
elBoberido, line 622: Does it mean that removing `:_` from the end makes this path invalid?

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
elBoberido: Okay, that's a good point. I think we can leave it as is for now

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
```diff
@@ -507,13 +513,163 @@ TEST(Helplets_test_isValidFilePath, WhenOneInvalidCharacterIsContainedPathIsInva
...
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("/.")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("./..")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("../.")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("some/path/to/..")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("/another/path/to/.")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("../bla/fuu/../blubb/.")));
+    EXPECT_FALSE(isValidPath(string<FILE_PATH_LENGTH>("./blubb/fuu/../bla/..")));
+}
+
+TEST(Helplets_test_isValidPath, FilePathsWithEndingDotsAreInvalid)
```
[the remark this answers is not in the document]
elBoberido: :grimacing:

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
elBoberido: Because the `FilePathsWithEndingDotsAreInvalid` test suggest that this is the case since the path would end with a dot.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/helplets.hpp
```diff
@@ -249,6 +249,9 @@ constexpr bool isPowerOfTwo(const T n) noexcept
     return n && ((n & (n - 1U)) == 0U);
 }
 
+template <uint64_t StringCapacity>
+bool isValidPathEntry(const string<StringCapacity>& name) noexcept;
```
elBoberido, line 257: Please add documentation

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
```diff
@@ -22,6 +22,40 @@ namespace iox
...
+
+    const auto nameSize = name.size();
+
+    for (uint64_t i = 0; i < nameSize; ++i)
+    {
+        const char c = name.c_str()[i];
+        if (!((internal::ASCII_A <= c && c <= internal::ASCII_Z)
+              || (internal::ASCII_CAPITAL_A <= c && c <= internal::ASCII_CAPITAL_Z)
+              || (internal::ASCII_0 <= c && c <= internal::ASCII_9) || c == internal::ASCII_MINUS
+              || c == internal::ASCII_DOT || c == internal::ASCII_COLON || c == internal::ASCII_UNDERSCORE))
```
elBoberido, line 44: How about `+%<>[](){}~#`? Would we consider some of them valid character?

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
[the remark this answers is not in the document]
elfenpiff: So I splitted the `isValidFilePath` function up so that I can reuse parts of it for `isValidPath`. The new function is called now `isValidPathEntry`. Since the function names started to confuse me I renamed them slightly:
* `isValidFilePath` -> `isValidPathToFile`
* `isValidPath` -> `isValidPathToDirectory`

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
```diff
@@ -119,13 +151,19 @@ inline bool isValidPath(const string<StringCapacity>& name) noexcept
...
 
     return true;
 }
 
+template <uint64_t StringCapacity>
+inline bool isValidPath(const string<StringCapacity>& name) noexcept
+{
+    return isValidPathToDirectory(name);
+}
+
```
elBoberido, line 166: I think this is not used anymore

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
```diff
@@ -48,34 +82,32 @@ inline bool isValidFileName(const string<StringCapacity>& name) noexcept
...
+    return (pathPart.empty() || isValidPathToDirectory(pathPart)) && isValidFileName(filePart);
 }
 
 template <uint64_t StringCapacity>
-inline bool isValidPath(const string<StringCapacity>& name) noexcept
+inline bool isValidFilePath(const string<StringCapacity>& name) noexcept
+{
+    return isValidPathToFile(name);
+}
```
elBoberido, line 107: This neither

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
```diff
@@ -48,34 +82,32 @@ inline bool isValidFileName(const string<StringCapacity>& name) noexcept
     }
```
elBoberido, line 82: A few lines above there is this check
```cpp
const string<StringCapacity> currentDirectory(".");
const string<StringCapacity> parentDirectory("..");

if (name == currentDirectory || name == parentDirectory)
{
    return false;
}
```
Even the code which follows is repeated in `isValidPathEntry`. Is this intentional?

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
elfenpiff: Ah of course then you are right.

The validation of the file names, paths and entries should work platform independently and therefore I restricted the names a bit to support some more file systems or let's say weird applications.

For instance, NTFS supports file names which end with a dot (`blaa.`) but the windows file explorer does not. So you can see the file but not open, modify or delete it since it is not supported by the Windows API. Yeah.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
elfenpiff: For the time being no. If they are valid characters they would have to fulfill the following requirements:

* Valid characters for the file systems:
    - linux: ext3, ext4, btrfs
    - windows: ntfs, exfat, fat
    - freebsd: ufs, ffs
    - apple: apfs
    - qnx: etfs?
    - android: ext3, fat (since we would like to support it in the future)

* The characters must be also supported by the corresponding API. The windows API does not support for instance dots at the end. Who knows why?!

So to conclude this. I would assume that we can add more characters to the list but in the first approach I just wanted to make sure that we have a meaningful minimum viable set of characters available which we can extend later.

And for that we would have to write unit tests by actually creating those files on the different file systems and access those files with the native APIs (most likely only windows). This can be ugly ...

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
elfenpiff: You are right but this was part of the public API of iceoryx 2.0. Therefore I thought to keep this function available and deprecate it for now. Can be deleted in iceoryx 4.0.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
elfenpiff: A there are subtle differences. 
The check in `isValidPathEntry` returns true for the `{current|parent}Directory` and false in `isValidFileName` since `a/../b` is allowed and here `..` would be a path entry but `..` would be an invalid file name.

iceoryx_hoofs/source/posix_wrapper/file_lock.cpp
elfenpiff: I created an issue and linked this comment: https://github.com/eclipse-iceoryx/iceoryx/issues/1430

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
elBoberido: Yes, we should at least deprecate it. I'm not sure what guarantees we currently give for non-posh related API. In the end, the change from the creation pattern to the builder pattern of the FileLock is also a breaking change and there is no migration path via a deprecated API but just the new API. Therefore I would also slightly lean towards just removing this functions since they are not used in our public API.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
elBoberido: How about 
```cpp
enum class CurrentAndParentDirectory {
    Rejected,
    Accepted,
};
```
`isValidFileName` would become to just
```cpp
return isValidPathEntry(name, CurrentAndParrentDirectory::Rejected);
```
and `isValidPathEntry` would have a
```cpp
if (name == currentDirectory || name == parentDirectory)
{
    return currentAndParentDirectoryHandling == CurrentAndParrentDirectory::Accepted;
}
```

Something like this would remove the code duplication and result in 100% MC/DC coverage for this function ;)

iceoryx_posh/include/iceoryx_posh/roudi/roudi_app.hpp
```diff
@@ -47,8 +47,8 @@ class RouDiApp
 
   protected:
     /// @brief waits for the next signal to RouDi daemon
-    [[deprecated(
-        "use iox::posix::waitForTerminationRequest() from 'iceoryx_hoofs/posix_wrapper/signal_watcher.hpp'")]] bool
+    [[deprecated("removed in 4.0, use iox::posix::waitForTerminationRequest() from "
```
elBoberido, line 50: I like this.

Disclaimer: I don't suggest to do this in this PR

It would be great to have this formalized and checked by the CI. A a script ninja could create a script to check for a specific syntax for our deprecation attributes. In combination with the `VERSION` file, this could tell us that there are deprecated functions for the new release and we need to remove them.

Do you think it might make sense to also tell the user which release introduce the deprecation? Just in case we have something that will be remove in next+2, e.g.
`[[deprecated("in 3.0, removed in 6.0, use foo::bar()")]]`

elBoberido (commented):
Almost there

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
```diff
@@ -362,99 +362,105 @@ TEST(Helplets_test_isValidFileName, WhenOneInvalidCharacterIsContainedFileNameIs
...
 }
 
-TEST(Helplets_test_isValidFilePath, WhenOneInvalidCharacterIsContainedPathIsInvalid)
+TEST(Helplets_test_isValidPathToFile, EmptyFilePathIsInvalid)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "a045581c-3a66-4d0e-b2e2-6ed5a97d4f89");
+    EXPECT_FALSE(isValidPathToFile(string<FILE_PATH_LENGTH>("")));
+}
+
+TEST(Helplets_test_isValidPathToFile_and_isValidPath, WhenOneInvalidCharacterIsContainedPathIsInvalid)
```
FerdinandSpitzschnueffler, line 463: The test case name seems to be outdated.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
elfenpiff: I removed it

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
```diff
@@ -504,18 +511,282 @@ TEST(Helplets_test_isValidFilePath, WhenOneInvalidCharacterIsContainedPathIsInva
...
+    {
+        string<FILE_PATH_LENGTH> sut = "HypnotoadAteTheSpagettiMonster";
+        ASSERT_TRUE(sut.unsafe_append(iox::platform::IOX_PATH_SEPARATORS[i]));
+        EXPECT_TRUE(doesEndWithPathSeparator(sut));
+    }
+}
+
+TEST(Helplets_test_isValidPathEntry, EmptyPathEntryIsValid)
+{
+    EXPECT_TRUE(
```
elBoberido, line 692: Please add test IDs to this and the following tests

elBoberido (changes requested):

elBoberido (approved):

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -182,3 +185,26 @@
...
+    // after
+    auto fileLock = iox::posix::FileLockBuilder().name("lockFileName")
+                                                 .path("/Now/I/Can/Add/A/Path")
+                                                 .permission(iox::cxx::perms::owner_all)
+                                                 .create()
+                                                 .expect("Oh no I couldn't create the lock file");
+    ```
+
+10. `isValidFilePath` is deprecated use `isValidPathToFile` instead.
```
elBoberido, line 203: Oh, I think this is not deprecated anymore but removed

elBoberido (changes requested):

elBoberido (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/helplets.inl
```diff
@@ -58,29 +48,48 @@ inline bool isValidFileName(const string<StringCapacity>& name) noexcept
...
+
+template <uint64_t StringCapacity>
+inline bool isValidPathToDirectory(const string<StringCapacity>& name) noexcept
+{
+    if (name.empty())
+    {
+        return false;
     }
 
     auto temp = name;
```
FerdinandSpitzschnueffler, line 95: The next three lines can be removed.

iceoryx_hoofs/test/moduletests/test_cxx_helplets.cpp
```diff
@@ -504,18 +511,290 @@ TEST(Helplets_test_isValidFilePath, WhenOneInvalidCharacterIsContainedPathIsInva
...
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("/.")));
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("./..")));
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("../.")));
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("some/path/to/..")));
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("/another/path/to/.")));
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("../bla/fuu/../blubb/.")));
+    EXPECT_TRUE(isValidPathToDirectory(string<FILE_PATH_LENGTH>("./blubb/fuu/../bla/..")));
+}
+
+TEST(Helplets_test_isValidPathToDirectory, FilePathsWithEndingDotsAreInvalid)
```
FerdinandSpitzschnueffler, line 624:
```suggestion
TEST(Helplets_test_isValidPathToDirectory, PathsToDirectoryWithEndingDotsAreInvalid)
```

PR #1433, opened 2022-06-29
iox-#1419 runtime name must be valid file name

### LINKED ISSUE #1419: Slash in runtime names is interpreted as sub path
## Required information

It is impossible to create a runtime with an app name which has a slash since this is interpreted as path by some underlying constructs.

```cxx
iox::runtime::PoshRuntime::initRuntime("aaargh/aaaaaargh/whoopsie");
```

The problem is caused by multiple posix constructs:
* The `FileLock` which tries to create a file lock under `/tmp/aaargh/aaaaaargh/whoopsie.lock` but the sub folders are not existing
* The `UnixDomainSocket` which tries to create a unix domain socket in `/tmp/aaargh/aaaaaargh/whoopsie`
* The `NamedPipe` which tries to create a shared memory file handle also under such a subpath
* The `NamedSemaphore` in `NamedPipe` with the same issue.

**Solution:**

iceoryx restricts the runtime names to valid file names which the user can verify via `iox::cxx::isValidFileName` from `cxx/helplets.hpp`.

This wouldn't be exactly a public API change since application which non leading slashs in their name would only generate fatal errors in the corresponding posix wrappers when the path does not exist.
But if some smart user would create `mkdir -p /tmp/aaargh/aaaaaargh/` the application would run but since this is a very weird corner case I would dismiss it.

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #1419


iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -81,18 +82,13 @@ const RuntimeName_t& PoshRuntime::verifyInstanceName(cxx::optional<const Runtime
...
+    else if (!cxx::isValidFileName(**name))
     {
-        LogError() << "Cannot initialize runtime. Application name must not be empty!";
-        errorHandler(PoshError::POSH__RUNTIME_NAME_EMPTY, ErrorLevel::FATAL);
-    }
-    else if (name.value()->c_str()[0] == '/')
-    {
-        LogError() << "Cannot initialize runtime. Please remove leading slash from Application name " << *name.value();
-        errorHandler(PoshError::POSH__RUNTIME_LEADING_SLASH_PROVIDED, ErrorLevel::FATAL);
+        LogFatal() << "Cannot initialize runtime. The application name \"" << **name << "\" is not a valid file name.";
```
mossmaurice, line 90: I think we should give the user more info on what is allowed and what not. Where did you derive `isValidFileName` from? Is this defined in some POSIX / UNIX standard?

```suggestion
        LogFatal() << "Cannot initialize runtime. The application name \"" << **name << "\" is not a valid file name according to POSIX blublablubb";
```

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
```diff
@@ -148,12 +148,23 @@ TEST_F(PoshRuntime_test, NoAppName)
...
 
-    EXPECT_DEATH({ PoshRuntime::initRuntime(invalidAppName); }, "POSH__RUNTIME_LEADING_SLASH_PROVIDED");
+    for (auto i : {"/miau",
+                   "/fuu/bar",
+                   "plum/bus",
+                   ".",
+                   "..",
+                   "strawberriesWithMayonnaiseIs/..",
+                   "ohLookADot.",
+                   "amIADirectory/"})
```
mossmaurice, line 162: I think you missed `IAmASpace ` and the empty case. In the end we're just duplicating the tests from `test_cxx_helplets.cpp`. Maybe this is not needed and we just test with one case because we know that the underlying `isValidFileName` works as expected?

iceoryx_posh/source/runtime/posh_runtime.cpp
mossmaurice: Please also update doxygen of `initRuntime` and add there how an valid name looks like e.g. `defined according to POSIX standard blublablubb`

mossmaurice (changes requested):
Thanks for the PR! I left two remarks.

dkroenke (approved):

iceoryx_posh/test/moduletests/test_posh_runtime.cpp
mossmaurice: @elfenpiff Ok. However, if we now support a new file system or platform we have to adapt two tests at the same time and one might forget either this one here or the on in helplets. Maybe we can share the test cases? It's not a biggie and can be solved on another PR.

mossmaurice (approved):

Issue #1196, opened 2022-03-02
Enable clang-tidy checks and fix warnings in code


elfenpiff:
## Brief feature description

At the moment a lot of clang-tidy checks in our root `.clang-tidy` file are disabled to ensure that our codebase is at least warning free with a minimal subset of clang-tidy checks.

We should enable the disabled checks one by one and solve the generated warnings in the same step. Some of the warnings require a larger refactoring. If this happens please create another issue and add a todo link in this issue.

## ToDo

1. Remove the compiler warning suppression from all tests, see `CMakeLists.txt` and the entry `set(TEST_CXX_FLAGS PRIVATE ${ICEORYX_WARNINGS} ${ICEORYX_SANITIZER_FLAGS} -Wno-pedantic -Wno-conversion)` and fix all warnings.
2. Create a `.clang-tidy` file in every test disables warning which originate from the googletest fixtures itself. For instance `TEST_F` and `TYPED_TEST` have a rule of 5 warning.

## Hints

The `.clang-tidy` file may have warnings which do not fit for us. If this is the case then please disable the warning and write a justification why this is rule is not appropriate. Example:
```
* rule: readability-identifier-length
  justification: In some functions single letter variables make sense, this is covered by the review process
  example:
     bad:
       int sum(int summand1, int summand2); // here we would like to have single letter variable arguments
     good:
       int sum(int a, int b); // failure since a and b have less than three letters
```

## CI checks
- `iceoryx_hoofs` diff checks
    - [x] `cxx` @elfenpiff 
    - [x] `concurrent` @elfenpiff 
    - [x] `design_pattern` @elfenpiff 
    - [x] `posix_wrapper` @elfenpiff 
    - [x] `units` @elfenpiff 
- [x] `iceoryx_hoofs` full nightly scan based on `.clang-tidy-diff-scans.txt` @elfenpiff 

## Files

When a file is stated, it means the header, inline file, cpp file and also the tests are refined.

### iceoryx_hoofs/cxx
- [x] `adaptive_wait` @elfenpiff 
- [x] `algorithm` @elfenpiff 
- [x] `attributes.hpp` @elfenpiff  (muahahha there were no warnings but it still counts!)
- [x] `convert` @elfenpiff 
- [x] `deadline_timer` @elfenpiff 
- [x] `expected` @elfenpiff 
- [x] `file_reader` @FerdinandSpitzschnueffler 
- [x] `filesystem` @elfenpiff 
[...]

PR #1473, opened 2022-07-07
iox-#1196 Fix clang-tidy and some AUTOSAR warnings in `cxx::vector`

mossmaurice:
Signed-off-by: Simon Hoinkis <simon.hoinkis@apex.ai>

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Fix clang-tidy warnings and add the respective AUTOSAR suppression

## References

- Closes #1196

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
@@ -192,6 +192,7 @@ inline bool vector<T, Capacity>::emplace(const uint64_t position, Targs&&... arg
         at_unchecked(i) = std::move(at_unchecked(i - 1U));
     }
 
+    at(position).~T();
     new (&at(position)) T(std::forward<Targs>(args)...);
     return true;
 }
@@ -248,7 +249,7 @@ inline bool vector<T, Capacity>::resize(const uint64_t count, const Targs&... ar
 template <typename T, uint64_t Capacity>
 inline T* vector<T, Capacity>::data() noexcept
 {
-    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
+    // AXIVION Next Line AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
     // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
     return const_cast<T*>(const_cast<const vector<T, Capacity>*>(this)->data());
 }
@@ -271,9 +272,7 @@ template <typename T, uint64_t Capacity>
 inline const T& vector<T, Capacity>::at(const uint64_t index) const noexcept
 {
     cxx::Expects((index < m_size) && "Out of bounds access");
-    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
-    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
-    return reinterpret_cast<const T*>(m_data)[index];
+    return at_unchecked(index);
 }
 
 template <typename T, uint64_t Capacity>
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
@@ -35,7 +35,7 @@ inline vector<T, Capacity>::vector(const uint64_t count, const T& value) noexcep
                   << std::endl;
     }
 
-    for (uint64_t i = 0u; i < count && i < Capacity; ++i)
+    for (uint64_t i = 0U; i < count && i < Capacity; ++i)
     {
         emplace_back(value);
     }
@@ -81,7 +81,7 @@ inline vector<T, Capacity>& vector<T, Capacity>::operator=(const vector& rhs) no
 {
     if (this != &rhs)
     {
-        uint64_t i = 0u;
+        uint64_t i = 0U;
         // copy using copy assignment
         for (; i < std::min(rhs.size(), size()); ++i)
         {
@@ -110,7 +110,7 @@ inline vector<T, Capacity>& vector<T, Capacity>::operator=(vector&& rhs) noexcep
 {
     if (this != &rhs)
     {
-        uint64_t i = 0u;
+        uint64_t i = 0U;
         // move using move assignment
         for (; i < std::min(rhs.size(), size()); ++i)
         {
@@ -138,7 +138,7 @@ inline vector<T, Capacity>& vector<T, Capacity>::operator=(vector&& rhs) noexcep
 template <typename T, uint64_t Capacity>
 inline bool vector<T, Capacity>::empty() const noexcept
 {
-    return m_size == 0u;
+    return m_size == 0U;
 }
 
 template <typename T, uint64_t Capacity>
@@ -315,7 +315,7 @@ template <typename T, uint64_t Capacity>
 inline T& vector<T, Capacity>::back() noexcept
 {
     cxx::Expects(!empty() && "Attempting to access the back of an empty vector");
-    return at(size() - 1u);
+    return at(size() - 1U);
 }
 
 template <typename T, uint64_t Capacity>
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
@@ -186,10 +186,13 @@ inline bool vector<T, Capacity>::emplace(const uint64_t position, Targs&&... arg
     {
         return emplace_back(std::forward<Targs>(args)...);
     }
-
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
     emplace_back(std::move(*reinterpret_cast<T*>(m_data[m_size - 1U])));
     for (uint64_t i = m_size - 1U; i > position; --i)
     {
+        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+        // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
         reinterpret_cast<T*>(m_data)[i] = std::move(reinterpret_cast<T*>(m_data)[i - 1U]);
     }
 
@@ -214,6 +217,8 @@ inline bool vector<T, Capacity>::pop_back() noexcept
 {
     if (m_size > 0U)
     {
+        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+        // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
         reinterpret_cast<const T*>(m_data)[--m_size].~T();
         return true;
     }
@@ -249,19 +254,24 @@ inline bool vector<T, Capacity>::resize(const uint64_t count, const Targs&... ar
 template <typename T, uint64_t Capacity>
 inline T* vector<T, Capacity>::data() noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
     return reinterpret_cast<T*>(m_data);
 }
 
 template <typename T, uint64_t Capacity>
 inline const T* vector<T, Capacity>::data() const noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
     return reinterpret_cast<const T*>(m_data);
 }
 
 template <typename T, uint64_t Capacity>
 inline T& vector<T, Capacity>::at(const uint64_t index) noexcept
 {
-    // PRQA S 3066 1 # const cast to avoid code duplication
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
     return const_cast<T&>(const_cast<const vector<T, Capacity>*>(this)->at(index));
 }
 
@@ -269,6 +279,8 @@ template <typename T, uint64_t Capacity>
 inline const T& vector<T, Capacity>::at(const uint64_t index) const noexcept
 {
     cxx::Expects((index < m_size) && "Out of bounds access");
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
     return reinterpret_cast<const T*>(m_data)[index];
 }
 
@@ -294,6 +306,8 @@ inline T& vector<T, Capacity>::front() noexcept
 template <typename T, uint64_t Capacity>
 inline const T& vector<T, Capacity>::front() const noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
     return const_cast<vector<T, Capacity>*>(this)->front();
 }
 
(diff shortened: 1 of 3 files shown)


elfenpiff (commented):
Could you please also add the vector tests here as well.

Otherwise this looks good and ready to be merged.

elBoberido (dismissed):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -186,10 +186,13 @@ inline bool vector<T, Capacity>::emplace(const uint64_t position, Targs&&... arg
     {
         return emplace_back(std::forward<Targs>(args)...);
     }
-
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
     emplace_back(std::move(*reinterpret_cast<T*>(m_data[m_size - 1U])));
```
mossmaurice, line 191: @elfenpiff A big downside to using an aligned, underlying byte array together with `reinterpret_cast`, is that the `cxx::vector` can never become `constexpr` when we move to C++17. [This talk](https://www.youtube.com/watch?v=I8QJLGI0GOE) provides some alternative ideas.

However, I suppose we'll stick with the current approach for now. Instead of implementing `uninitialized_array`, we could think about using `std::aligned_storage`.

elBoberido: @mossmaurice  AFAIK the problem is the destructor not the reinterpret cast. I'm not quite sure if we can easily circumvent this. One option is of course a specialization for trivially destructible types but that might be quite an amount of work.

elBoberido (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: @elBoberido David Stone says something different in the talk and see [this thread](https://stackoverflow.com/a/59913380/18276160). `reinterpret_cast`s are not possible in `constexpr` methods during compile-time.

elBoberido: oh boy, I just love this language :grimacing:

FerdinandSpitzschnueffler (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -186,10 +186,13 @@ inline bool vector<T, Capacity>::emplace(const uint64_t position, Targs&&... arg
...
     for (uint64_t i = m_size - 1U; i > position; --i)
     {
+        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+        // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
         reinterpret_cast<T*>(m_data)[i] = std::move(reinterpret_cast<T*>(m_data)[i - 1U]);
```
elfenpiff, line 196: @mossmaurice @FerdinandSpitzschnueffler @elBoberido 

Isn't this line broken? Here we just move the pointer `T*` and not the underlying object. I think the dereferencing is missing here and the correct line would be:
```cxx
*reinterpret_cast<T*>(m_data)[i] = std::move(*reinterpret_cast<T*>(m_data)[i - 1U]);
```

Do we have tests for it? Or am I blind or something ...

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -214,6 +217,8 @@ inline bool vector<T, Capacity>::pop_back() noexcept
 {
     if (m_size > 0U)
     {
+        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+        // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
         reinterpret_cast<const T*>(m_data)[--m_size].~T();
```
elfenpiff, line 222: Again this seems somehow wrong ... Why does a const cast work here? You cannot call the destructor on a const object since it changes the object itself.
Furthermore, since it is casted to a pointer, shouldn't the line be:
```cxx
reinterpret_cast<T*>(m_data)[--m_size]->~T();
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -294,43 +306,55 @@ inline T& vector<T, Capacity>::front() noexcept
...
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
     return const_cast<vector<T, Capacity>*>(this)->back();
 }
 
 template <typename T, uint64_t Capacity>
 inline typename vector<T, Capacity>::iterator vector<T, Capacity>::begin() noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
```
elfenpiff, line 333: We should use here:
```cxx
return &at(0);
```
Then we do not need to add an axivion and clang-tidy exception.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -294,43 +306,55 @@ inline T& vector<T, Capacity>::front() noexcept
...
     return reinterpret_cast<iterator>(&(m_data[0]));
 }
 
 template <typename T, uint64_t Capacity>
 inline typename vector<T, Capacity>::const_iterator vector<T, Capacity>::begin() const noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
```
elfenpiff, line 341: We should use here:
```cxx
return &at(0);
```
Then we do not need to add an axivion and clang-tidy exception.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -249,26 +254,33 @@ inline bool vector<T, Capacity>::resize(const uint64_t count, const Targs&... ar
...
     return reinterpret_cast<T*>(m_data);
 }
 
 template <typename T, uint64_t Capacity>
 inline const T* vector<T, Capacity>::data() const noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
```
elfenpiff, line 266: We should use here:
```cxx
return &at(0);
```
Then we do not need to add an axivion and clang-tidy exception.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -249,26 +254,33 @@ inline bool vector<T, Capacity>::resize(const uint64_t count, const Targs&... ar
 template <typename T, uint64_t Capacity>
 inline T* vector<T, Capacity>::data() noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
```
elfenpiff, line 258: We should use here:
```cxx
return &at(0);
```
Then we do not need to add an axivion and clang-tidy exception.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elfenpiff, line 191: I think in this method we could replace all `reinterpret_cast` with `at`. Then we also do not need to write axivion and clang-tidy suppressions.

elBoberido: In this case I would suggest to introduce a private `at_unchecked`. There is no need to do the check at every access since we already ensured that the following operations work within the bounds of the vector.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elBoberido: Here I would keep the current code. `&at(0)` will change the semantics since it will terminate when the vector is empty. If this is desired, then it is a different question but with the current semantics it breaks the code. Since there is also `front`, I would keep the current semantic of `data`

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elBoberido: This will break the iterator since an empty vector will crash instead of just returning an iterator past end. `at_unchecked` would work though.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elBoberido: No, this breaks the iterator

iceoryx_hoofs/test/moduletests/test_cxx_vector.cpp
```diff
@@ -311,300 +314,304 @@ TEST_F(vector_test, SizeZeroAfterClear)
...
+    for (uint64_t i = 0U; i < 10U; ++i)
     {
         sut1.emplace_back(i);
     }
 
-    vector<CTorTest, 10> sut2(std::move(sut1));
-    for (uint64_t i = 0; i < 10; ++i)
-    {
-        sut2.at(i) = i;
-    }
```
elBoberido, line 399: same here

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elBoberido: @elfenpiff this must be possible else RAII would not work with `const`
```cpp
{
    const Foo foo;
   // ... do something with foo
} // dtor of foo must be called
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elfenpiff: @elBoberido exactly this is the problem, this line moves a pointer not a reference!

What in essence is written here is `ptr = std::move(anotherPointer);` and this is wrong, we would like to move the content of the type and use the types move assignment operator to create space for the element which is created in the new position. So we would have to write `*ptr = std::move(*anotherPointer);` since we would like to move the type! Take a look at line 191, there it is done correctly.

elfenpiff: Alright @mossmaurice to sum it up @elBoberido is right again but the problem was just that the brackets were set weirdly. Here is the thing:
```cxx
reinterpret_cast<T*>(m_data)[i]; // this is a T&
reinterpret_cast<T*>(m_data[i]); // this is a T* and requires an additional * to dereference it ... do you see the difference
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -186,10 +186,13 @@ inline bool vector<T, Capacity>::emplace(const uint64_t position, Targs&&... arg
...
-
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+    // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
     emplace_back(std::move(*reinterpret_cast<T*>(m_data[m_size - 1U])));
     for (uint64_t i = m_size - 1U; i > position; --i)
     {
+        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Type-safety ensured by template parameter
+        // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
         reinterpret_cast<T*>(m_data)[i] = std::move(reinterpret_cast<T*>(m_data)[i - 1U]);
     }
```
elfenpiff, line 198: But I think @elBoberido agrees that we have in line 199 an actual bug.

The element at `position` was moved in the for loop in line 196. Therefore `position` does contain a valid but moved element. We have to call the destructor first before creating a new element with placement new in line 199.

So we have to add `at(position).~T()` here otherwise we override a valid element without calling its destructor first.

Here an example why this is important: the `smart_lock`. All constructors of the smart lock are creating a mutex and remove this mutex on destruction. If a `smart_lock` is moved the mutex is never moved, it is only locked and unlocked to secure the data but never moved.
So a moved `smart_lock` contains still a `mutex` resource which has to be cleaned up by the destructor otherwise we have a `mutex` leak. It is important that the `smart_lock` contains a valid `mutex` resource since this is always allowed in c++
```cxx
MyType a;
MyType b = std::move(a);
a = std::move(anotherVariable); // I can always move/copy assign another variable to a moved variable
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elBoberido: @elfenpiff hehe
grafik

elBoberido: @elfenpiff one is also not supposed to call the dtor explicitly. This is undefined behavior

```cpp
const Foo foo;
foo.~Foo();
foo.bar();
```

Similar to

```cpp
Foo foo1;
Foo foo2{std::move(foo1)};
foo1.bar();
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: Fixed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: @elfenpiff Thanks for the summary. I replaced it with `at_unchecked`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: @elfenpiff Thanks for pointing that out. I have troubles to understand your example. In my view `std::move` is just passing the ownership and not ending the lifetime of an object. Hence, if someone would store a `smart_lock` in `cxx::vector` (not sure yet if that makes sense at all :D) it would still exist and the d'tor should not be called! 


```
Before
+----+----+----+----+
| 1  | 2  | 3  |    |   m_size = 3
+----+----+----+----+

vec.emplace(2, 10)

After
+----+----+----+----+
| 1  | 10 | 2  | 3  |   m_size = 4
+----+----+----+----+
```

Element `2` is still alive, so why should the d'tor be called?

elfenpiff: @mossmaurice In line 199 you perform a placement new, this overrides an object without calling the destructor.

The lifetime of an object always has to start with the constructor and always has to end with the destructor. When one performs another placement new on a memory position were already an object is living one breaks this essential contract. The destructor must be called before the placement new.

elfenpiff (commented):
Please fix the missing destructor call.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: Used `at_unchecked` here.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: @elBoberido @elfenpiff  would vote for changing the semantics. I can add a note to the release notes. Anyway for now I used `at_unchecked` and let the non-`const` `data()` call the `const` one.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: I used that in the const `data()` and called that one from here.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: I used `at_unchecked`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
mossmaurice: @elfenpiff In the above example this rule is not violated. Once `2` get removed from the vector the d'tor will be called. Overriding a move-from object is ok. Think about the empty-house-without-furniture analogy. We can construct and new house on a empty house.

mossmaurice: Ok, got it now. Although, `std::move` means passing the ownership of ressources according to RAII, there can be other things in the object that are not moved. This needs to be destroyed of course. :see_no_evil: Will write a test case and fix it.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
elBoberido: @mossmaurice this is a good question and I might take counsel with my pillow first.

In general we should try to stay consistent with other constructs, e.g. `string::c_str` which is the equivalent to this `data` function.

Something to keep in mind is also what is dangerous in this situation. Getting the pointer is not dangerous since the memory is there. An out of bounds access is dangerous but that can also happen when the vector contains e.g. 1 item. Therefore the tendency is to keep the current semantic but lets see what the pillow counsel reveals

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/vector.inl
```diff
@@ -248,8 +249,8 @@ inline bool vector<T, Capacity>::resize(const uint64_t count, const Targs&... ar
 template <typename T, uint64_t Capacity>
 inline T* vector<T, Capacity>::data() noexcept
 {
-    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
     // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
+    // AXIVION Next Line AutosarC++19_03-A5.2.3 : const cast to avoid code duplication
```
elBoberido, line 253: This has to be switched

iceoryx_hoofs/test/moduletests/test_cxx_vector.cpp
```diff
@@ -452,6 +452,28 @@ TEST_F(vector_test, DestructorWithFullVector)
...
+        EXPECT_THAT(customCTor, Eq(9U));
+        EXPECT_TRUE(sut.emplace(5U, 42U));
+        EXPECT_THAT(customCTor, Eq(10U));
+        EXPECT_THAT(moveAssignment, Eq(4U));
+        EXPECT_THAT(dTor, Eq(1U));
+    }
+    // Last element in the vector is moved and not constructed, hence #moveCTor+#customCTor = #dTor
+    EXPECT_THAT(moveCTor, Eq(1U));
+    EXPECT_THAT(customCTor, Eq(10U));
+    EXPECT_THAT(dTor, Eq(11U));
```
elBoberido, line 474: Please don't use magic numbers.

elBoberido (approved):

elfenpiff (approved):

PR #1475, opened 2022-07-07
Iox #1196 fix lockfree queue warnings

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Partial #1196

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,6 +18,10 @@ concurrency-*,
 performance-*,
 hicpp-*,
 
+-cppcoreguidelines-non-private-member-variables-in-classes,
+
+-hicpp-named-parameter,
+
 -readability-avoid-const-params-in-decls,
 -readability-identifier-length,
 -readability-redundant-access-specifiers,
@@ -26,12 +30,6 @@ hicpp-*,
 -readability-identifier-naming,
 -readability-use-anyofallof,
 -readability-named-parameter,
-
--cppcoreguidelines-non-private-member-variables-in-classes,
--cppcoreguidelines-avoid-c-arrays,
-
--hicpp-avoid-c-arrays,
--hicpp-named-parameter,
 
 '
 
@@ -101,14 +99,28 @@ hicpp-*,
 #                  Even public members are useful in lightweight classes (effectively structs).
 #                  This unnecessarily limits design options.
 #
-# * rule: cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays
-#   justicfication: C arrays are useful in low level code for buffers etc.
-#                   It should not be mandatory to use std::array as sometimes the STL dependency is not desired.
-#
-# * rule: readability-named-parameter
+# * rule: readability-named-parameter, -hicpp-named-parameter
 #   justicfication: For tag types which just guide overloading no name is needed.
 #                   Declaration of private functions does not require a name (no doxygen comment).
 #                   For callbacks omitting the name is less verbose than to cast an unsued parameter to void.
+#   callbacks
+#   bad:
+#       auto callable = [] (auto /*foo*/) {...};
+#   good:
+#       auto callable = [] (auto) {};
+#   tag types
+#   optional:
+#       void foo(Tag1 tag); // (1) called with Tag1 type
+#       void foo(Tag2 tag); // (2) called with Tag2 type
+#       // dispatch site
+#       Tag1 tag;
+#       foo(tag); // calls (1)
+#   also ok:
+#       void foo(Tag1); // (1)
+#       void foo(Tag2); // (2)
+#       // dispatch site
+#       Tag1 tag;
+#       foo(tag); // calls (1)
 #
 ## Those warnings should be enabled
 ## They are disabled since they require a heavy API refactoring and when we enable it we clutter the code with // NOLINT comments
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -25,6 +25,13 @@ hicpp-*,
 
 -readability-identifier-naming,
 -readability-use-anyofallof,
+-readability-named-parameter,
+
+-cppcoreguidelines-non-private-member-variables-in-classes,
+-cppcoreguidelines-avoid-c-arrays,
+
+-hicpp-avoid-c-arrays,
+-hicpp-named-parameter,
 
 '
 
@@ -88,7 +95,21 @@ hicpp-*,
 #
 # * rule: readability-use-anyofallof
 #   justification: requires C++20 and std::ranges but we only use C++14
-
+#
+# * rule: cppcoreguidelines-non-private-member-variables-in-classes
+#   justification: Sometimes it makes sense to have protected members to extend the base class.
+#                  Even public members are useful in lightweight classes (effectively structs).
+#                  This unnecessarily limits design options.
+#
+# * rule: cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays
+#   justicfication: C arrays are useful in low level code for buffers etc.
+#                   It should not be mandatory to use std::array as sometimes the STL dependency is not desired.
+#
+# * rule: readability-named-parameter
+#   justicfication: For tag types which just guide overloading no name is needed.
+#                   Declaration of private functions does not require a name (no doxygen comment).
+#                   For callbacks omitting the name is less verbose than to cast an unsued parameter to void.
+#
 ## Those warnings should be enabled
 ## They are disabled since they require a heavy API refactoring and when we enable it we clutter the code with // NOLINT comments
 # -bugprone-easily-swappable-parameters
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp
@@ -51,6 +51,8 @@ class Buffer
   private:
     using byte_t = uint8_t;
 
+    // NOLINTJUSTIFICATION required by low level buffer building block and encapsulated in abstraction
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
     alignas(ElementType) byte_t m_buffer[Capacity * sizeof(ElementType)];
 
     ElementType* toPtr(index_t index) const noexcept;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.inl
@@ -14,6 +14,9 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#ifndef IOX_HOOFS_CONCURRENT_LOCKFREE_QUEUE_BUFFER_INL
+#define IOX_HOOFS_CONCURRENT_LOCKFREE_QUEUE_BUFFER_INL
+
 #include "iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp"
 
 namespace iox
@@ -60,3 +63,5 @@ ElementType* Buffer<ElementType, Capacity, index_t>::toPtr(index_t index) const 
 
 } // namespace concurrent
 } // namespace iox
+
+#endif // IOX_HOOFS_CONCURRENT_LOCKFREE_QUEUE_BUFFER_INL
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.inl
@@ -14,6 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include "iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp"
+
 namespace iox
 {
 namespace concurrent
@@ -52,6 +54,7 @@ template <typename ElementType, uint64_t Capacity, typename index_t>
 ElementType* Buffer<ElementType, Capacity, index_t>::toPtr(index_t index) const noexcept
 {
     auto ptr = &(m_buffer[index * sizeof(ElementType)]);
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast, cppcoreguidelines-pro-type-const-cast) type erasure
     return reinterpret_cast<ElementType*>(const_cast<byte_t*>(ptr));
 }
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
@@ -112,10 +112,12 @@ class IndexQueue
     using Index = CyclicIndex<Capacity>;
     using Cell = std::atomic<Index>;
 
-    ///    this member has to be initialized explicitly in the constructor since
-    ///    the default atomic constructor does not call the default constructor of the
-    ///    underlying class.
-    ///    See, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0883r0.pdf
+    /// the array entries have to be initialized explicitly in the constructor since
+    /// the default atomic constructor does not call the default constructor of the
+    /// underlying class.
+    /// See, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0883r0.pdf
+    // NOLINTJUSTIFICATION we need an initialized array here and will not use std::array for now
+    // NOLINTNEXTLINE(*avoid-c-arrays)
     Cell m_cells[Capacity];
 
     std::atomic<Index> m_readPosition;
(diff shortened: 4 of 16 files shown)


.clang-tidy
```diff
@@ -84,6 +90,16 @@ hicpp-*,
...
 #        void bar(Foo& foo) {...}
+#
+# * rule: cppcoreguidelines-non-private-member-variables-in-classes
+#   justification: Sometimes it makes sense to have protected members to extend the base class.
+#                  Even public members are useful in lightweight classes (effectively structs).
+#                  This unnecessarily limits design options.
+#
+# * rule: cppcoreguidelines-avoid-c-arrays
+# * rule: hicpp-avoid-c-arrays
+#   justicfication: C arrays are useful in low level code for buffers etc.
```
MatthiasKillat, line 101: @elBoberido If we do not want to suppress this globally I will add a suppression comment where it applies.

Note that I think using C arrays should be allowed and would be in favor of suppression.

iceoryx_hoofs/test/moduletests/test_lockfree_queue.cpp
```diff
@@ -41,8 +42,11 @@ struct Integer
     int value{0};
 
     // so that it behaves like an int for comparison purposes
+    // NOLINTNEXTLINE(hicpp-explicit-conversions) required for typed tests
     operator int() const
     {
+        /// @todo suppression does not work but it is a false positive
+        // NOLINTNEXTLINE(clang-analyzer-core.unitialized.UndefReturn) false positive - defaults to 0
```
MatthiasKillat, line 49: @elBoberido 

This suppression does not quite work and the warning is false positive or at least not clear to me. The return value should be well defined if the object is constructed (and we do not call it before).

iceoryx_hoofs/test/moduletests/test_lockfree_queue_buffer.cpp
```diff
@@ -32,22 +32,11 @@ class LockFreeQueueBufferTest : public ::testing::Test
...
-    void SetUp()
-    {
-    }
-
-    void TearDown()
-    {
-    }
+    ~LockFreeQueueBufferTest() override = default;
 
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init) false positive
```
MatthiasKillat, line 39: @elBoberido This suppression also does not work for me and again I think it is a false positive. The buffer is default initialized in the default ctor, there should be no problem (at least the way I understand the warning).

The problem is that the rule may find valid problems with uninitialized members as well, but if it catches these kind of false positives it is useless and should be disabled I think.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.inl
```diff
@@ -14,6 +14,11 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#ifndef IOX_HOOFS_CONCURRENT_LOCKFREE_QUEUE_BUFFER_INL
+#define IOX_HOOFS_CONCURRENT_LOCKFREE_QUEUE_BUFFER_INL
+
+#include "iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp"
```
MatthiasKillat, line 20: @elBoberido 
This is necessary to provide definitions locally in this file (but not for the way it is included). If not included, this causes clang-tidy warnings as it does not know about the context of inclusion.

This pseudo-cyclical inclusion (prevented by preprocessor) is weird design though.

.clang-tidy
```diff
@@ -25,6 +25,11 @@ hicpp-*,
 
 -readability-identifier-naming,
 -readability-use-anyofallof,
+-readability-named-parameter,
+
+-cppcoreguidelines-non-private-member-variables-in-classes,
+
+-hicpp-named-parameter,
```
elBoberido, line 32: Could you move this up so that they are alphabetically sorted?

.clang-tidy
```diff
@@ -88,7 +93,17 @@ hicpp-*,
...
+# * rule: cppcoreguidelines-non-private-member-variables-in-classes
+#   justification: Sometimes it makes sense to have protected members to extend the base class.
+#                  Even public members are useful in lightweight classes (effectively structs).
+#                  This unnecessarily limits design options.
+#
+# * rule: readability-named-parameter
+#   justicfication: For tag types which just guide overloading no name is needed.
+#                   Declaration of private functions does not require a name (no doxygen comment).
+#                   For callbacks omitting the name is less verbose than to cast an unsued parameter to void.
+#
```
elBoberido, line 106: Could you please add a small example which shows the violation and in the `readability-named-parameter` how we want it, e.g.
```
bad
  auto callable = [] (auto /*foo*/) {...};
good
  auto callable = [] (auto) {};
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp
```diff
@@ -51,6 +51,7 @@ class Buffer
   private:
     using byte_t = uint8_t;
 
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
     alignas(ElementType) byte_t m_buffer[Capacity * sizeof(ElementType)];
 
     ElementType* toPtr(index_t index) const noexcept;
```
elBoberido, line 57: Please add a justification like `Required for low-level building block and wrapped in the abstraction`

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.inl
elBoberido: yes, it's required for clang-tidy but also other tools require it to cleanly parse the file

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
```diff
@@ -116,6 +116,7 @@ class IndexQueue
     ///    the default atomic constructor does not call the default constructor of the
     ///    underlying class.
     ///    See, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0883r0.pdf
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
     Cell m_cells[Capacity];
```
elBoberido, line 120: Couldn't the `buffer` be used for this? Not in this PR but just for a future refactoring?

iceoryx_hoofs/test/moduletests/test_index_queue.cpp
```diff
@@ -32,21 +32,9 @@ class IndexQueueTest : public ::testing::Test
...
+    IndexQueueTest() = default;
 
-    void SetUp()
-    {
-    }
-
-    void TearDown()
-    {
-    }
+    ~IndexQueueTest() override = default;
```
elBoberido, line 37: Didn't know that `override` also applies to dtors

iceoryx_hoofs/test/moduletests/test_resizeable_lockfree_queue.cpp
```diff
@@ -29,14 +29,16 @@ using namespace ::testing;
...
 struct Integer
 {
+    // NOLINTNEXTLINE(hicpp-explicit-conversions) required for typed tests
     Integer(uint64_t value = 0)
         : value(value)
     {
     }
 
-    uint64_t value{0};
+    uint64_t value = 0;
```
elBoberido, line 38: Out of curiosity, did clang-tidy complain about this or is it just personal preference?

iceoryx_hoofs/test/moduletests/test_resizeable_lockfree_queue.cpp
```diff
@@ -481,7 +471,8 @@ TYPED_TEST(ResizeableLockFreeQueueTest, DecreaseCapacityOfAPartiallyFilledQueue)
 
     uint64_t element = 0U;
     while (q.tryPush(element++))
-        ;
+    {
+    }
```
elBoberido, line 475: We should have a clang-format which creates `while (cond) {}` in a single line

.clang-tidy
MatthiasKillat: I was thinking about sorting them by category/prefix as first key.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
MatthiasKillat: Better use some initialized array here as we want the ctor to be called for each element. Maybe even std::array.

iceoryx_hoofs/test/moduletests/test_index_queue.cpp
MatthiasKillat: Apparently. It is in principle also a function and here the base implementation is virtual so clang-tidy complains to be explicit about overriding (even for default).

iceoryx_hoofs/test/moduletests/test_resizeable_lockfree_queue.cpp
MatthiasKillat: I do not remember. Will check later. Personally I can live with either syntax for primitive types but prefer `=` as it is closer to other languages.

iceoryx_hoofs/test/moduletests/test_resizeable_lockfree_queue.cpp
MatthiasKillat: Or, better, allow the previous code which is not bad to read and it is perfectly fine to have a loop and some small amount of logic in the loop condition. Textbook example would be quicksort (splitting by pivot).

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
```diff
@@ -116,6 +116,7 @@ class IndexQueue
     ///    the default atomic constructor does not call the default constructor of the
     ///    underlying class.
     ///    See, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0883r0.pdf
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
```
FerdinandSpitzschnueffler, line 119: Please add a justification (and maybe remove the white spaces at the beginning of the comments above).

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -14,12 +15,18 @@
...
+
+#include "iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp"
+
 namespace iox
 {
 namespace concurrent
 {
 template <uint64_t Capacity, typename ValueType>
-IndexQueue<Capacity, ValueType>::IndexQueue(ConstructEmpty_t) noexcept
+
```
FerdinandSpitzschnueffler, line 28:
```suggestion
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -30,6 +37,7 @@ IndexQueue<Capacity, ValueType>::IndexQueue(ConstructEmpty_t) noexcept
 }
 
 template <uint64_t Capacity, typename ValueType>
+
```
FerdinandSpitzschnueffler, line 40:
```suggestion
```

.clang-tidy
```diff
@@ -88,7 +93,17 @@ hicpp-*,
...
 # * rule: readability-use-anyofallof
 #   justification: requires C++20 and std::ranges but we only use C++14
-
+#
+# * rule: cppcoreguidelines-non-private-member-variables-in-classes
+#   justification: Sometimes it makes sense to have protected members to extend the base class.
+#                  Even public members are useful in lightweight classes (effectively structs).
+#                  This unnecessarily limits design options.
+#
+# * rule: readability-named-parameter
```
FerdinandSpitzschnueffler, line 102:
```suggestion
# * rule: readability-named-parameter, hicpp-named-parameter
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
MatthiasKillat: Changed to `std::array` (which does not allocate nor throw if used correctly (only in bounds checked access).

iceoryx_hoofs/test/moduletests/test_resizeable_lockfree_queue.cpp
MatthiasKillat: @elBoberido It is not required and hence undone. I believe clang-tidy complained when my language server was not working correctly for some reason but cannot reproduce it.

In my opinion both are fine for primitive types.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
elBoberido: @MatthiasKillat oh, then this was a misunderstanding on my side. I assumed `Cell` to be an alias to something like `uint8_t[N]`. The C array would have been okay here.

iceoryx_hoofs/test/moduletests/test_resizeable_lockfree_queue.cpp
elBoberido: No, not in this but maybe in a later PR.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.inl
```diff
@@ -226,9 +232,9 @@ bool IndexQueue<Capacity, ValueType>::popIfFull(ValueType& index) noexcept
 }
 
 template <uint64_t Capacity, typename ValueType>
-bool IndexQueue<Capacity, ValueType>::popIfSizeIsAtLeast(const uint64_t requiredSize, ValueType& index) noexcept
+bool IndexQueue<Capacity, ValueType>::popIfSizeIsAtLeast(const uint64_t minSize, ValueType& index) noexcept
```
elBoberido, line 235: Could you please also adjust the name in the definition in the corresponding hpp

tools/scripts/used-headers.txt
```diff
@@ -24,6 +24,7 @@ sys/wait.h
 unistd.h
 # iceoryx_posh / iceoryx_hoofs headers
 algorithm
+array
```
elBoberido, line 27: Would it be much work to revert this change?

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
MatthiasKillat: `using Cell = std::atomic<Index>;` is an atomic of lightweight structure, but not a primitive type.

Do we not want to use `std::array` or not? I cannot help but think that our coding or design guidelines combined with the clang-tidy rules are not ideal in this case. We unnecessarily restrict everything and additionally make the code more verbose/ugly.

It is also at odds with the clang tidy rules in this case (which says use something like a std::array). Now we do not have this something of our own yet (and I would argue we do not need it as std::array is fine for us as well).

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
MatthiasKillat: Not needed with the current change to std::array, which is under debate.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.inl
MatthiasKillat: Already the case, they were out of sync and that is why it was changed here.

tools/scripts/used-headers.txt
MatthiasKillat: Of course not, but is it necessary? It is currently needed to pass the checks when using `std::array` as substitute for a C array.

It all depends on whether we want/can use `std::array` which I am in favor of.

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
MatthiasKillat: @FerdinandSpitzschnueffler As I undid the usage of `std::array` after discussion with @elBoberido I added the justification.

Note that I only use an extra line for the justification if it is too long to fit on the suppression command line. Now we may decide to do it differently in the future but I think this should be a separate PR once the details are decided. (I prefer to have this line optional).

iceoryx_hoofs/include/iceoryx_hoofs/internal/concurrent/lockfree_queue/index_queue.hpp
elBoberido: Except that we removed the `std::array` in favor of the C-array to keep the STL dependency low, there is no technical reason not to use it. I would keep the C-array for now and maybe we create a small `cxx::array` wrapper for the `buffer`/`uninitialized_buffer`

elBoberido (approved):

FerdinandSpitzschnueffler (approved):

Issue #1481, opened 2022-07-11
empty PREFIX leads to wrong path in INTERFACE_INCLUDE_DIRECTORIES on CMake


dkroenke:
## Required information

**Operating system:**
E.g. Ubuntu 22.04

**CMake version:**
CMake 3.22.1

**Observed result or behaviour:**
The `INTERFACE_INCLUDE_DIRECTORIES` defined in the generated `iceoryx_poshTargets.cmake`
file contains invalid directory set to `${_IMPORT_PREFIX}/include//iceoryx/v2.90.0`.
This affects all iceoryx packages.

When an external application tries to use the `INTERFACE_INCLUDE_DIRECTORIES` property,
the includes cannot be found.

test code in iceoryx example:
```
set(HOOFS_LIB)
get_target_property(HOOFS_LIB iceoryx_hoofs::iceoryx_hoofs INTERFACE_INCLUDE_DIRECTORIES)
message("HOOFS_LIB: " ${HOOFS_LIB})
```

output in out-of-tree build:
```
HOOFS_LIB: /home/my/folder/iceoryx/build/install/prefix/include//iceoryx/v2.90.0
```

**Expected result or behaviour:**

INTERFACE_INCLUDE_DIRECTORIES evaluates always to a valid directory
with set/unset PREFIX variable.

**Conditions where it occurred / Performed steps:**

The reason lies in this [code snippet]():

```
if(PREFIX STREQUAL "")
  set(PREFIX "${_unique_dir}")
else()
  set(PREFIX "${PREFIX}/${_unique_dir}")
endif()
```

The `PREFIX` expression evaluates here always to false even when the variable
is not set which is the case in iceoryx.

This leads to an attached `/` character
in the install include directories.
The generated include directory seems not to be affected by this.

Solution could be to evaluate the actual content of the Variable:
```
if("${PREFIX}" STREQUAL "")
```

FYI @sloretz

PR #1485, opened 2022-07-11
iox-#1481 Fix interface include directories in CMake

dkroenke:
Signed-off-by: Dietrich Krönke <dietrich.kroenke@apex.ai>

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Relates #1481

### DIFF
--- a/iceoryx_hoofs/cmake/IceoryxPackageHelper.cmake
+++ b/iceoryx_hoofs/cmake/IceoryxPackageHelper.cmake
@@ -111,22 +111,6 @@ Macro(install_package_files_and_export)
     )
 endMacro()
 
-Macro(iox_make_unique_includedir)
-    option(MAKE_UNIQUE_INCLUDEDIR
-      "When ON headers are installed to a path ending with folders called \
-      iceoryx/vX.Y.Z/ . This avoids include directory search order issues when \
-      overriding this package from a merged catkin, ament, or colcon workspace."
-      ON)
-
-    if(MAKE_UNIQUE_INCLUDEDIR)
-      set(_unique_dir "iceoryx/v${IOX_VERSION_STRING}")
-      if(PREFIX STREQUAL "")
-        set(PREFIX "${_unique_dir}")
-      else()
-        set(PREFIX "${PREFIX}/${_unique_dir}")
-      endif()
-    endif()
-endMacro()
 
 Macro(iox_set_rpath)
     set(switches IS_EXECUTABLE)
@@ -205,7 +189,7 @@ endMacro()
 Macro(iox_add_executable)
     set(switches USE_C_LANGUAGE PLACE_IN_BUILD_ROOT)
     set(arguments TARGET STACK_SIZE)
-    set(multiArguments FILES LIBS INCLUDE_DIRECTORIES LIBS_QNX LIBS_LINUX LIBS_UNIX LIBS_WIN32 LIBS_APPLE 
+    set(multiArguments FILES LIBS INCLUDE_DIRECTORIES LIBS_QNX LIBS_LINUX LIBS_UNIX LIBS_WIN32 LIBS_APPLE
         BUILD_INTERFACE INSTALL_INTERFACE)
     cmake_parse_arguments(IOX "${switches}" "${arguments}" "${multiArguments}" ${ARGN} )
 
--- a/doc/aspice_swe3_4/example/iceoryx_component/CMakeLists.txt
+++ b/doc/aspice_swe3_4/example/iceoryx_component/CMakeLists.txt
@@ -21,6 +21,8 @@ find_package(iceoryx_hoofs REQUIRED)
 
 include(IceoryxPlatform)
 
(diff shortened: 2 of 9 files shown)


iceoryx_hoofs/cmake/IceoryxPackageHelper.cmake
```diff
@@ -120,7 +120,7 @@ Macro(iox_make_unique_includedir)
 
     if(MAKE_UNIQUE_INCLUDEDIR)
       set(_unique_dir "iceoryx/v${IOX_VERSION_STRING}")
-      if(PREFIX STREQUAL "")
+      if("${PREFIX}" STREQUAL "")
```
elBoberido, line 123: Not sure what would be more idiomatic cmake. Your fix or `if(NOT ${PREFIX} OR ${PREFIX} STREQUAL ""`. The fix seems to work but if we for whatever reason rename `PREFIX` and forget to rename it in the string, cmake would not complain and happily. Although it would be the same with `if(NOT ${PREFIX} OR ${PREFIX} STREQUAL ""`.

The issue in this case was that `PREFIX` was used instead of `${PREFIX}`. The latter one will always result in a cmake error if it is used but does not exist. The question is if we really use this variable anywhere and if we could just get rid of it?

elBoberido: Okay, I take everything back I wrote above. It seems `${VAR}` and `"${VAR}"` are error prone
https://stackoverflow.com/questions/19982340/cmake-compare-to-empty-string-with-strequal-failed

The most safe solution which also states the intent and prevents our futures selves from removing the quotation marks is probably this
```cmake
if(NOT DEFINED PREFIX OR PREFIX STREQUAL "")
```

dkroenke: This is a possible solution bould could also not prevent that someone use `set(PREFIX " ")` to circumvent this.
One idea would be to hardcode the unique include dir directly without setting the Prefix in every library.

What we currently do is that the `iox_add_library` macro take the `PREFIX` as argument for the `PROJECT_PREFIX`
and `INSTALL_INTERFACE`.
Before that we must try to evaluate somehow that PREFIX is valid and set it to the unique include dir (e.g. `iceoryx/v2.90.0`).
We basically make the include dir unique by adding iceoryx and the version number. So why not use the project version of CMake directly?

We could have something like:
```cmake
iox_add_library(
    TARGET                      iceoryx_hoofs
    NAMESPACE                   iceoryx_hoofs
    PROJECT_PREFIX              iceoryx/v${CMAKE_PROJECT_VERSION}
    PUBLIC_LIBS                 iceoryx_hoofs::iceoryx_platform
    ....
    INSTALL_INTERFACE           include/iceoryx/v${CMAKE_PROJECT_VERSION}
```

With that we can ensure that the version is set properly and can get rid of the macro. If the projects version is empty, the CMake will drop an error here.
Would this make sense

elfenpiff: @dkroenke I like your solution and would go for it!

But when all libraries and executables are always using the same `PROJECT_PREFIX` (`iceoryx/v${CMAKE_PROJECT_VERSION}`) we could also set it directly in `iox_add_library` and `iox_add_executable` and remove the `PROJECT_PREFIX` option from them. Then we have it under our control and the user does not forget it.

Would this work?

elBoberido: @elfenpiff as long as `ìox_add_...` isn't used for something outside of iceoryx, this should work

dkroenke: This will not work with packages like iceoryx_dds which are planned to be moved outside of iceoryx.
We are already defining a lot centrally in iceoryx_hoofs but we should keep in mind to leave configuration options open for later refactorings.

elfenpiff: @dkroenke you are right but in this case we do this centrally for a good reason. It would be great if we could find a way to make this work even in iceoryx_dds otherwise we reintroduce all the issues and bugs we already solved with the approach.

But this is a task for the future ...

dkroenke: @elBoberido @elfenpiff Reworked the PR by setting PREFIX directly and removing the macro check.

elBoberido (commented):
Can you also add the `set(PREFIX iceoryx/v${CMAKE_PROJECT_VERSION})` to `doc/aspice_swe3_4/example/iceoryx_component/CMakeLists.txt`

elfenpiff (approved):

elBoberido (approved):
Will this be backported to release_2.0?

dkroenke:
> Will this be backported to release_2.0?

Yes, here the PR: https://github.com/eclipse-iceoryx/iceoryx/pull/1503

PR #1520, opened 2022-07-18
iox-#1196 Remove conversion operator in `RelativePointer`

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Instead of making the conversion operator `explicit` remove it and use `get()`

## References

- Closes #1196 (partly)

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -174,6 +174,8 @@ class BaseRelativePointer
     static constexpr offset_t NULL_POINTER_OFFSET = std::numeric_limits<offset_t>::max();
 
   protected:
+    ~BaseRelativePointer() noexcept = default;
+
     id_t m_id{NULL_POINTER_ID};
     offset_t m_offset{NULL_POINTER_OFFSET};
 };
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
@@ -82,9 +82,9 @@ class RelativePointer : public BaseRelativePointer
     /// @return a pointer to the underlying object
     T* get() const noexcept;
 
-    /// @brief converts the RelativePointer to a pointer of the type of the underlying object
-    /// @return a pointer of type T pointing to the underlying object
-    operator T*() const noexcept;
+    /// @brief converts the RelativePointer to bool
+    /// @return bool which contains true if the RelativePointer contains a pointer
+    explicit operator bool() const noexcept;
 
     /// @brief checks if this and ptr point to the same pointee
     /// @param[in] ptr is the pointer whose pointee is compared with this' pointee
--- a/iceoryx_hoofs/source/concurrent/loffli.cpp
+++ b/iceoryx_hoofs/source/concurrent/loffli.cpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -42,7 +42,7 @@ void LoFFLi::init(cxx::not_null<Index_t*> freeIndicesMemory, const uint32_t capa
         for (uint32_t i = 0; i < m_size + 1; i++)
         {
             // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) upper limit of index is set by m_size
-            m_nextFreeIndex[i] = i + 1;
+            m_nextFreeIndex.get()[i] = i + 1;
         }
     }
 }
@@ -61,7 +61,7 @@ bool LoFFLi::pop(Index_t& index) noexcept
         }
 
         // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) upper limit of index set by m_size
-        newHead.indexToNextFreeIndex = m_nextFreeIndex[oldHead.indexToNextFreeIndex];
+        newHead.indexToNextFreeIndex = m_nextFreeIndex.get()[oldHead.indexToNextFreeIndex];
         newHead.abaCounter += 1;
     } while (!m_head.compare_exchange_weak(oldHead, newHead, std::memory_order_acq_rel, std::memory_order_acquire));
 
@@ -74,7 +74,7 @@ bool LoFFLi::pop(Index_t& index) noexcept
     ///         thread which performs the cleanup and during this process a synchronization
     ///         is required
     // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
-    m_nextFreeIndex[index] = m_invalidIndex;
+    m_nextFreeIndex.get()[index] = m_invalidIndex;
 
     /// we need to synchronize m_nextFreeIndex with push so that we can perform a validation
     /// check right before push to avoid double free's
@@ -91,7 +91,7 @@ bool LoFFLi::push(const Index_t index) noexcept
     /// we want to avoid double free's therefore we check if the index was acquired
     /// in pop and the push argument "index" is valid
     // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) index is limited by capacity
-    if (index >= m_size || m_nextFreeIndex[index] != m_invalidIndex)
+    if (index >= m_size || m_nextFreeIndex.get()[index] != m_invalidIndex)
     {
         return false;
     }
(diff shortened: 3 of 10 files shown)


iceoryx_hoofs/source/concurrent/loffli.cpp
```diff
@@ -42,7 +42,7 @@ void LoFFLi::init(cxx::not_null<Index_t*> freeIndicesMemory, const uint32_t capa
         for (uint32_t i = 0; i < m_size + 1; i++)
         {
             // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) upper limit of index is set by m_size
-            m_nextFreeIndex[i] = i + 1;
+            m_nextFreeIndex.get()[i] = i + 1;
```
MatthiasKillat, line 45: I will review later, but just looking at this syntax it already feels wrong to remove the operator. It is not really clear from reading either, just more verbose and confusing.

I clearly remember why I opted for the implicit conversion to pointer. If I read what is originally there I just think: "sure, it is some kind of pointer and we use pointer arithmetic and dereference it." For the understanding at this level it does not matter what kind of pointer it has, only that it has pointer syntax (and the corresponding access semantics).

But now I pause a little and need to think about that I probably get some pointer back and then we perform the pointer arithmetic. It would even look worse with explicit conversion.

For me it is the posterchild case where implicit conversion is the way to go.

MatthiasKillat: Sure, making it explicit is worse and I do not advocate this. I advocate keeping it implicit as it shall behave like a pointer from the point of view of the client,  which is exactly what is accomplished by implicit conversion.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
```diff
@@ -82,9 +82,19 @@ class RelativePointer : public BaseRelativePointer
...
-    /// @return a pointer of type T pointing to the underlying object
-    operator T*() const noexcept;
+    /// @brief converts the RelativePointer to bool
+    /// @return bool which contains true if the RelativePointer contains a pointer
+    explicit operator bool() const noexcept;
+
+    /// @brief True if the underyling pointer is smaller than the one in the rhs
+    /// @param[in] rhs Another RelativePointer
+    /// @return True if smaller, otherwise false
+    bool operator<(const RelativePointer<T>& rhs) const noexcept;
```
MatthiasKillat, line 92: Do we need these operators? Usually pointer comparison like this is a bad idea. Equality comparison is useful, but order not so much. One reason is that a order has to satisfy certain properties (basically be at least a irreflexive transitive relation (this is implicitly assumed by various STL constructs).

E.g. `std::map<K,V>` requires that these operators exist for K, but I am not aware that we use these pointers in a map. If the relation < exists but does not have these properties, weird things happen.

One can define it to essentially be the < relation on the corresponding raw pointer.
However, it makes little sense to compare these kinds of pointers as the relation cannot hold across process boundaries (i.e. in one process the pointers may compare as <, in the other as >). This may lead to all kinds of unpredictable problems.

Implicitly this was the case before as well, but not really responsibility of the class.

TLDR: Expecting an order on raw pointers is a bad idea. Expecting an order on fancy pointers is worse (especially those that have to work across process boundaries). The operators should be removed unless really needed. If the user wants to compare the raw pointers after a `get` she should do just that (and be aware of the problems).

MatthiasKillat (commented):
implicit vs. explicit conversion aside (I prefer implicit conversion here), there is the issue whether `operator<` should exist here. If we do not compare those pointers in such a way, it should be removed. If we do, the question is: why? (it was previously possible due to implicit conversion and compared the raw pointer). pointer comparison with < is usually not a good idea.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -174,6 +174,8 @@ class BaseRelativePointer
     static constexpr offset_t NULL_POINTER_OFFSET = std::numeric_limits<offset_t>::max();
 
   protected:
+    ~BaseRelativePointer() noexcept = default;
```
MatthiasKillat, line 177: Update copyright.

iceoryx_hoofs/source/concurrent/loffli.cpp
```diff
@@ -42,7 +42,7 @@ void LoFFLi::init(cxx::not_null<Index_t*> freeIndicesMemory, const uint32_t capa
         for (uint32_t i = 0; i < m_size + 1; i++)
```
MatthiasKillat, line 42: Update copyright.

iceoryx_posh/source/mepoo/mem_pool.cpp
```diff
@@ -90,15 +90,15 @@ void* MemPool::getChunk() noexcept
     m_usedChunks.fetch_add(1U, std::memory_order_relaxed);
     adjustMinFree();
```
MatthiasKillat, line 92: Update copyright.

iceoryx_posh/source/mepoo/shm_safe_unmanaged_chunk.cpp
```diff
@@ -59,7 +59,7 @@ SharedChunk ShmSafeUnmanagedChunk::releaseToSharedChunk() noexcept
     }
     auto chunkMgmt = rp::RelativePointer<mepoo::ChunkManagement>(m_chunkManagement.offset(), m_chunkManagement.id());
     m_chunkManagement.reset();
```
MatthiasKillat, line 61: Update copyright.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
mossmaurice: We store `RelativePointer`s in `cxx::vector`. The operators are required for `std::sort` in file `iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl:179-180`. The code was added for the blocking publisher feature. I understand your point and I'm not sure if we uncovered a conversion problem here. My understanding:

```
// chunk_distributor_data.hpp
using QueueContainer_t =
    cxx::vector<rp::RelativePointer<ChunkQueueData_t>, ChunkDistributorDataProperties_t::MAX_QUEUES>;

// chunk_distributor.inl
std::sort(getMembers()->m_queues.begin(), getMembers()->m_queues.end());

// vector.inl
using const_iterator = const T*; // with T being RelativePointer
return reinterpret_cast<const_iterator>(&at_unchecked(0));

This used to call the conversion operator and use the `operator<(const ChunkQueueData& lhs, ChunkQueueData rhs)`. This is now done via the new operators explicitly.
```

MatthiasKillat: The question is what should be sorted here? If its is just to have a well-defined order it is fine to sort by pointer. In fact I think sorting by data here is not well-defined anyway.

The elements point to something in shared memory, which in a fixed process address space can act as unique identifiers and are sortable by  pointer. I just think that the solution is to use an explicit sorting criterion with `std::sort`, e.g. with a functor struct or function that defines the order.

mossmaurice: @elfenpiff Can you answer the above question? Looking again at the code, I think we want to sort the pointers, correct?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
```
elfenpiff, line 2: Please add all touched files from hoofs into `.clang-tidy-diff-scans.txt`

elfenpiff (changes requested):

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
mossmaurice: Talked to @elfenpiff the other day, the sorting mechanism is indeed wanted here and correct.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -176,14 +176,19 @@ inline uint64_t ChunkDistributor<ChunkDistributorDataType>::deliverToAllStoredQu
...
-            std::sort(getMembers()->m_queues.begin(), getMembers()->m_queues.end());
-            std::sort(remainingQueues.begin(), remainingQueues.end());
+            auto greaterThan = [](rp::RelativePointer<ChunkQueueData_t>& a,
+                                  rp::RelativePointer<ChunkQueueData_t>& b) -> bool {
+                return reinterpret_cast<uint64_t>(a.get()) > reinterpret_cast<uint64_t>(b.get());
+            };
+            std::sort(getMembers()->m_queues.begin(), getMembers()->m_queues.end(), greaterThan);
+            std::sort(remainingQueues.begin(), remainingQueues.end(), greaterThan);
 
             auto iter = std::set_intersection(getMembers()->m_queues.begin(),
```
MatthiasKillat, line 186: I double checked, and it appears that `set_intersection` requires sorted ranges despite in theory being able to sort them itself (by the predicate provided). It is important that the ranges are sorted by the same predicate used in the intersection, which is the case here.

For us it does not even matter much what the order is, just that it is a well-defined order (irreflexive, transitive etc.), which it is if sorting.

MatthiasKillat (dismissed):
Functionally it looks fine, but I still think that implicit conversion and suppression of static analysis findings would be the way to go here (reasons where stated in comments).

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.inl
```diff
@@ -145,7 +145,7 @@ inline void ChunkQueuePopper<ChunkQueueDataType>::unsetConditionVariable() noexc
 template <typename ChunkQueueDataType>
 inline bool ChunkQueuePopper<ChunkQueueDataType>::isConditionVariableSet() const noexcept
 {
-    return getMembers()->m_conditionVariableDataPtr;
```
elBoberido, line 148: Why not just `getMembers()->m_conditionVariableDataPtr != nullptr`?

MatthiasKillat (approved):

elBoberido (approved):

PR #1521, opened 2022-07-18
iox-#1196 Fix clang-tidy warnings in `RelativePointer`

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Fix `bugprone-easily-swappable-parameters` warning
  * Use `NewType` to wrap `iox::rp::id_t` when used as parameter
* Fix some other `clang-tidy` warnings

## References

- Closes #1196 (partly)

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
@@ -18,6 +18,7 @@
 #ifndef IOX_HOOFS_RELOCATABLE_POINTER_BASE_RELATIVE_POINTER_HPP
 #define IOX_HOOFS_RELOCATABLE_POINTER_BASE_RELATIVE_POINTER_HPP
 
+#include "iceoryx_hoofs/cxx/newtype.hpp"
 #include "iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp"
 
 #include <cstdint>
@@ -51,7 +52,16 @@ namespace rp
 class BaseRelativePointer
 {
   public:
-    using id_t = uint64_t;
+    struct id_t : public cxx::NewType<uint64_t,
+                                      cxx::newtype::DefaultConstructable,
+                                      cxx::newtype::Convertable,
+                                      cxx::newtype::ConstructByValueCopy,
+                                      cxx::newtype::MoveConstructable>
+    {
+        using ThisType::ThisType;
+    };
+
+    using id_underlying_t = id_t::value_type;
     using ptr_t = void*;
     using const_ptr_t = const void* const;
     using offset_t = std::uintptr_t;
@@ -59,20 +69,20 @@ class BaseRelativePointer
     /// @brief constructs a BaseRelativePointer pointing to the same pointee as ptr in a segment identified by id
     /// @param[in] ptr the pointer whose pointee shall be the same for this
     /// @param[in] id is the unique id of the segment
-    BaseRelativePointer(ptr_t ptr, id_t id) noexcept;
+    BaseRelativePointer(const ptr_t ptr, const id_t& id) noexcept;
 
     /// @brief constructs a BaseRelativePointer from a given offset and segment id
     /// @param[in] offset is the offset
     /// @param[in] id is the unique id of the segment
-    BaseRelativePointer(offset_t offset, id_t id) noexcept;
+    BaseRelativePointer(const offset_t offset, const id_t& id) noexcept;
 
     /// @brief constructs a BaseRelativePointer pointing to the same pointer as ptr
     /// @param[in] ptr the pointer whose pointee shall be the same for this
-    BaseRelativePointer(ptr_t ptr = nullptr) noexcept;
+    explicit BaseRelativePointer(const ptr_t ptr = nullptr) noexcept;
 
     /// @brief copy constructor
     /// @param[in] other is the copy origin
-    BaseRelativePointer(const BaseRelativePointer& other) noexcept;
+    BaseRelativePointer(const BaseRelativePointer& other) noexcept = default;
 
     /// @brief move constructor
     /// @param[in] other is the move origin
@@ -99,7 +109,7 @@ class BaseRelativePointer
 
     /// @brief returns the id which identifies the segment
     /// @return the id which identifies the segment
-    id_t getId() const noexcept;
+    id_underlying_t getId() const noexcept;
 
     /// @brief returns the offset
     /// @return the offset
@@ -113,24 +123,24 @@ class BaseRelativePointer
     /// @param[in] ptr starting address of the segment to be registered
     /// @param[in] size is the size of the segment
     /// @return id it was registered to
-    static id_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
+    static id_underlying_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
 
     /// @brief tries to register a memory segment with a given size starting at ptr to a given id
     /// @param[in] id is the id of the segment
     /// @param[in] ptr starting address of the segment to be registered
     /// @param[in] size is the size of the segment
     /// @return true if successful (id not occupied), false otherwise
-    static bool registerPtr(const id_t id, const ptr_t ptr, uint64_t size = 0U) noexcept;
+    static bool registerPtr(const id_t& id, const ptr_t ptr, uint64_t size = 0U) noexcept;
 
     /// @brief unregisters ptr with given id
     /// @param[in] id is the id of the segment
     /// @return true if successful (ptr was registered with this id before), false otherwise
-    static bool unregisterPtr(const id_t id) noexcept;
+    static bool unregisterPtr(const id_t& id) noexcept;
 
     /// @brief get the base ptr associated with the given id
     /// @param[in] id is the id of the segment
     /// @return ptr registered at the given id, nullptr if none was registered
-    static ptr_t getBasePtr(const id_t id) noexcept;
+    static ptr_t getBasePtr(const id_t& id) noexcept;
 
     /// @brief unregisters all ptr id pairs (leads to initial state)
     static void unregisterAll() noexcept;
@@ -139,27 +149,27 @@ class BaseRelativePointer
     /// @param[in] id is the id of the segment and is used to get the base pointer
     /// @param[in] ptr is the pointer whose offset should be calculated
     /// @return offset
-    static offset_t getOffset(const id_t id, const_ptr_t ptr) noexcept;
+    static offset_t getOffset(const id_t& id, const_ptr_t ptr) noexcept;
 
     /// @brief get the pointer from id and offset ("inverse" to getOffset)
     /// @param[in] id is the id of the segment and is used to get the base pointer
     /// @param[in] offset is the offset for which the pointer should be calculated
     /// @return the pointer from id and offset
-    static ptr_t getPtr(const id_t id, const offset_t offset) noexcept;
+    static ptr_t getPtr(const id_t& id, const offset_t offset) noexcept;
 
     /// @brief get the id for a given ptr
     /// @param[in] ptr the pointer whose corresponding id is searched for
     /// @return id the pointer was registered to
-    static id_t searchId(ptr_t ptr) noexcept;
+    static id_underlying_t searchId(ptr_t ptr) noexcept;
 
     /// @brief checks if given id is valid
     /// @param[in] id is the id to be checked
     /// @return true if the given id is valid, otherwise false
-    static bool isValid(id_t id) noexcept;
+    static bool isValid(id_t& id) noexcept;
 
     /// @brief returns the pointer repository
     /// @return the pointer repository
-    static PointerRepository<id_t, ptr_t>& getRepository() noexcept;
+    static PointerRepository<id_underlying_t, ptr_t>& getRepository() noexcept;
 
     /// @brief get the offset from the start address of the segment and ptr
     /// @param[in] ptr is the pointer whose offset should be calculated
@@ -170,14 +180,17 @@ class BaseRelativePointer
     /// @return the pointer for stored id and offset
     ptr_t computeRawPtr() const noexcept;
 
-    static constexpr id_t NULL_POINTER_ID = std::numeric_limits<id_t>::max();
+    static constexpr id_underlying_t NULL_POINTER_ID{std::numeric_limits<id_underlying_t>::max()};
     static constexpr offset_t NULL_POINTER_OFFSET = std::numeric_limits<offset_t>::max();
 
   protected:
     ~BaseRelativePointer() noexcept = default;
 
-    id_t m_id{NULL_POINTER_ID};
+    // BaseRelativePointer only used with RelativePointer
+    // NOLINTBEGIN(cppcoreguidelines-non-private-member-variables-in-classes)
+    id_underlying_t m_id{NULL_POINTER_ID};
     offset_t m_offset{NULL_POINTER_OFFSET};
+    // NOLINTEND(cppcoreguidelines-non-private-member-variables-in-classes)
 };
 } // namespace rp
 } // namespace iox
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
@@ -54,6 +54,7 @@ class BaseRelativePointer
   public:
     struct id_t : public cxx::NewType<uint64_t,
                                       cxx::newtype::DefaultConstructable,
+                                      cxx::newtype::CopyConstructable,
                                       cxx::newtype::Convertable,
                                       cxx::newtype::ConstructByValueCopy,
                                       cxx::newtype::MoveConstructable>
@@ -69,12 +70,12 @@ class BaseRelativePointer
     /// @brief constructs a BaseRelativePointer pointing to the same pointee as ptr in a segment identified by id
     /// @param[in] ptr the pointer whose pointee shall be the same for this
     /// @param[in] id is the unique id of the segment
-    BaseRelativePointer(const ptr_t ptr, const id_t& id) noexcept;
+    BaseRelativePointer(const ptr_t ptr, const id_t id) noexcept;
 
     /// @brief constructs a BaseRelativePointer from a given offset and segment id
     /// @param[in] offset is the offset
     /// @param[in] id is the unique id of the segment
-    BaseRelativePointer(const offset_t offset, const id_t& id) noexcept;
+    BaseRelativePointer(const offset_t offset, const id_t id) noexcept;
 
     /// @brief constructs a BaseRelativePointer pointing to the same pointer as ptr
     /// @param[in] ptr the pointer whose pointee shall be the same for this
@@ -130,17 +131,17 @@ class BaseRelativePointer
     /// @param[in] ptr starting address of the segment to be registered
     /// @param[in] size is the size of the segment
     /// @return true if successful (id not occupied), false otherwise
-    static bool registerPtr(const id_t& id, const ptr_t ptr, uint64_t size = 0U) noexcept;
+    static bool registerPtr(const id_t id, const ptr_t ptr, uint64_t size = 0U) noexcept;
 
     /// @brief unregisters ptr with given id
     /// @param[in] id is the id of the segment
     /// @return true if successful (ptr was registered with this id before), false otherwise
-    static bool unregisterPtr(const id_t& id) noexcept;
+    static bool unregisterPtr(const id_t id) noexcept;
 
     /// @brief get the base ptr associated with the given id
     /// @param[in] id is the id of the segment
     /// @return ptr registered at the given id, nullptr if none was registered
-    static ptr_t getBasePtr(const id_t& id) noexcept;
+    static ptr_t getBasePtr(const id_t id) noexcept;
 
     /// @brief unregisters all ptr id pairs (leads to initial state)
     static void unregisterAll() noexcept;
@@ -149,13 +150,13 @@ class BaseRelativePointer
     /// @param[in] id is the id of the segment and is used to get the base pointer
     /// @param[in] ptr is the pointer whose offset should be calculated
     /// @return offset
-    static offset_t getOffset(const id_t& id, const_ptr_t ptr) noexcept;
+    static offset_t getOffset(const id_t id, const_ptr_t ptr) noexcept;
 
     /// @brief get the pointer from id and offset ("inverse" to getOffset)
     /// @param[in] id is the id of the segment and is used to get the base pointer
     /// @param[in] offset is the offset for which the pointer should be calculated
     /// @return the pointer from id and offset
-    static ptr_t getPtr(const id_t& id, const offset_t offset) noexcept;
+    static ptr_t getPtr(const id_t id, const offset_t offset) noexcept;
 
     /// @brief get the id for a given ptr
     /// @param[in] ptr the pointer whose corresponding id is searched for
@@ -165,7 +166,7 @@ class BaseRelativePointer
     /// @brief checks if given id is valid
     /// @param[in] id is the id to be checked
     /// @return true if the given id is valid, otherwise false
-    static bool isValid(id_t& id) noexcept;
+    static bool isValid(id_t id) noexcept;
 
     /// @brief returns the pointer repository
     /// @return the pointer repository
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
@@ -51,6 +51,9 @@ class PointerRepository
     static constexpr id_t MIN_ID{1U};
     static constexpr id_t MAX_ID{CAPACITY - 1U};
 
+    static_assert(MAX_ID >= MIN_ID, "MAX_ID must be greater or equal than MIN_ID!");
+    static_assert(CAPACITY >= 2, "CAPACITY must be at least 2!");
+
   public:
     static constexpr id_t INVALID_ID = std::numeric_limits<id_t>::max();
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
@@ -28,12 +28,14 @@ namespace iox
 {
 namespace rp
 {
+constexpr uint64_t MAX_POINTER_REPO_CAPACITY{10000U};
+
 /// @brief Allows registration of memory segments with their start pointers and size.
 /// This class is used to resolve relative pointers in the corresponding address space of the application.
 /// Up to CAPACITY segments can be registered with MIN_ID = 1 to MAX_ID = CAPACITY - 1
 /// id 0 is reserved and allows relative pointers to behave like normal pointers
 /// (which is equivalent to measure the offset relative to 0).
-template <typename id_t, typename ptr_t, uint64_t CAPACITY = 10000U>
+template <typename id_t, typename ptr_t, uint64_t CAPACITY = MAX_POINTER_REPO_CAPACITY>
 class PointerRepository
 {
   private:
@@ -46,8 +48,8 @@ class PointerRepository
     /// @note 0 is a special purpose id and reserved
     /// id 0 is reserved to interpret the offset just as a raw pointer,
     /// i.e. its corresponding base ptr is 0
-    static constexpr size_t MIN_ID = 1U;
-    static constexpr size_t MAX_ID = CAPACITY - 1U;
+    static constexpr id_t MIN_ID{1U};
+    static constexpr id_t MAX_ID{CAPACITY - 1U};
 
   public:
     static constexpr id_t INVALID_ID = std::numeric_limits<id_t>::max();
@@ -106,7 +108,6 @@ class PointerRepository
     iox::cxx::vector<Info, CAPACITY> m_info;
     uint64_t m_maxRegistered{0U};
 };
-
 } // namespace rp
 } // namespace iox
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
@@ -40,10 +40,11 @@ inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t
     if (m_info[id].basePtr == nullptr)
     {
         m_info[id].basePtr = ptr;
-        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to the
-        // original type
-        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-        m_info[id].endPtr = reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+        m_info[id].endPtr =
+            // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to
+            // the original type
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+            reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
         if (id > m_maxRegistered)
         {
             m_maxRegistered = id;
@@ -61,10 +62,11 @@ inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(const ptr_t pt
         if (m_info[id].basePtr == nullptr)
         {
             m_info[id].basePtr = ptr;
-            // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to
-            // the original type
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-            m_info[id].endPtr = reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+            m_info[id].endPtr =
+                // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back
+                // to the original type
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+                reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
             if (id > m_maxRegistered)
             {
                 m_maxRegistered = id;
@@ -73,7 +75,7 @@ inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(const ptr_t pt
         }
     }
 
-    return INVALID_ID;
+    return id_t{INVALID_ID};
 }
 
 template <typename id_t, typename ptr_t, uint64_t CAPACITY>
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
@@ -40,11 +40,11 @@ inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t
     if (m_info[id].basePtr == nullptr)
     {
         m_info[id].basePtr = ptr;
-        m_info[id].endPtr =
-            // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to
-            // the original type
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-            reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to
+        // the original type
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+        m_info[id].endPtr = reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+
         if (id > m_maxRegistered)
         {
             m_maxRegistered = id;
@@ -62,11 +62,11 @@ inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(const ptr_t pt
         if (m_info[id].basePtr == nullptr)
         {
             m_info[id].basePtr = ptr;
-            m_info[id].endPtr =
-                // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back
-                // to the original type
-                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-                reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+            // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back
+            // to the original type
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+            m_info[id].endPtr = reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+
             if (id > m_maxRegistered)
             {
                 m_maxRegistered = id;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
@@ -51,7 +51,7 @@ class RelativePointer : public BaseRelativePointer
 
     /// @brief constructs a RelativePointer pointing to the same pointee as ptr
     /// @param[in] ptr the pointer whose pointee shall be the same for this
-    RelativePointer(ptr_t ptr) noexcept;
+    explicit RelativePointer(ptr_t ptr) noexcept;
 
     /// @brief assigns the RelativePointer to point to the same pointee as ptr
     /// @param[in] ptr the pointer whose pointee shall be the same for this
@@ -62,20 +62,10 @@ class RelativePointer : public BaseRelativePointer
     /// @tparam U a template parameter to enable the dereferencing operator only for non-void T
     /// @return a reference to the underlying object
     template <typename U = T>
-    typename std::enable_if<!std::is_void<U>::value, U&>::type operator*() noexcept;
-
-    /// @brief access to the underlying object
-    /// @return a pointer to the underlying object
-    T* operator->() noexcept;
-
-    /// @brief dereferencing operator which returns a const reference to the underlying object
-    /// @tparam U a template parameter to enable the dereferencing operator only for non-void T
-    /// @return a const reference to the underlying object
-    template <typename U = T>
     typename std::enable_if<!std::is_void<U>::value, const U&>::type operator*() const noexcept;
 
     /// @brief read-only access to the underlying object
-    /// @return a const pointer to the underlying object
+    /// @return a pointer to the underlying object
     T* operator->() const noexcept;
 
     /// @brief access the underlying object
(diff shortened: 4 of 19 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -170,12 +191,15 @@ class BaseRelativePointer
     /// @return the pointer for stored id and offset
     ptr_t computeRawPtr() const noexcept;
 
-    static constexpr id_t NULL_POINTER_ID = std::numeric_limits<id_t>::max();
```
mossmaurice, line 173: @elBoberido @MatthiasKillat 
Unfortunately, due to the default d'tors in `NewType`, it can't be `constexpr`. We could however remove the default d'tors in `NewType` and then `id_t` could stay `constexpr`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.inl
```diff
@@ -58,35 +58,42 @@ inline typename std::enable_if<!std::is_void<U>::value, U&>::type RelativePointe
...
 
 template <typename T>
-inline T* RelativePointer<T>::get() const noexcept
+inline const T* RelativePointer<T>::get() const noexcept
 {
     return static_cast<T*>(computeRawPtr());
 }
 
 template <typename T>
 inline RelativePointer<T>::operator T*() const noexcept
```
mossmaurice, line 94: @elBoberido @MatthiasKillat 

This operator is removed on #1520

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
```diff
@@ -21,13 +21,13 @@ namespace iox
...
 {
-BaseRelativePointer::BaseRelativePointer(ptr_t ptr, id_t id) noexcept
+BaseRelativePointer::BaseRelativePointer(const ptr_t ptr, const id_t& id) noexcept
     : m_id(id)
     , m_offset(computeOffset(ptr))
 {
 }
 
-BaseRelativePointer::BaseRelativePointer(offset_t offset, id_t id) noexcept
+BaseRelativePointer::BaseRelativePointer(const offset_t offset, const id_t& id) noexcept
```
mossmaurice, line 30: @MatthiasKillat @elBoberido 

Here, and in some other place with `Foo(uint64_t, uint64_t)` the clang tidy warning `bugprone-easily-swappable-parameters`. I think this is important to solve as `Foo(1,1)` is legit but not readable and error-prone.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
elBoberido: @mossmaurice the problem is not the dtor. It must be something else. This compiles and runs
```cpp
#include <cstdint>
#include <iostream>

struct Foo {
    constexpr Foo(uint64_t size) : size(size) {}

    uint64_t size {0};
protected:
    ~Foo() = default;
};

struct Bar : public Foo {
    constexpr Bar(uint64_t size) : Foo(size) {}
};

template <uint64_t N>
uint64_t answer() {
    return N;
}

int main() {
    static constexpr Bar bar{42};

    std::cout << answer<bar.size>() << std::endl;

    return 0;
}
```

Does the `NewType` have `constexpr` ctors?

mossmaurice: @elBoberido I tried to make `NewType` `constexpr` by adding it for all the methods. However, I couldn't make it compile :( We should try that on a separate branch maybe.

elBoberido: btw, the `id_t` is in the `rp` namespace, right?

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
elBoberido: I think it is sufficient for now to have only `id_t`. `offset_t` can be changed in the future if necessary.

MatthiasKillat (commented):
I believe `NewType` as it is is not mature and powerful enough to solve the swappable parameter problem in an efficient way that leads to optimal and expressive code for reasons stated in review. Main reason are that it leads to excessive casts, does not allow constexpr and likely will lead the compiler to miss some optimizations (due to lack of constexpr alone).

I think the cost/benefit calculation is not in its favor (cost = runtime cost, code maintainability).

However, we can decide to go that way but I think some test would be required for the extension of `NewType`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -48,31 +49,49 @@ namespace rp
...
+    using id_t = rp::id_t;
     using ptr_t = void*;
     using const_ptr_t = const void* const;
     using offset_t = std::uintptr_t;
 
     /// @brief constructs a BaseRelativePointer pointing to the same pointee as ptr in a segment identified by id
     /// @param[in] ptr the pointer whose pointee shall be the same for this
     /// @param[in] id is the unique id of the segment
-    BaseRelativePointer(ptr_t ptr, id_t id) noexcept;
+    BaseRelativePointer(const ptr_t ptr, const id_t& id) noexcept;
```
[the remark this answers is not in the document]
mossmaurice: With this change, it's not a POD type anymore and clang-tidy complained that non-POD should be passed by reference. I don't think the compiler will optimize the `NewType` away.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -48,31 +49,49 @@ namespace rp
...
+                                  cxx::newtype::MoveAssignable,
+                                  cxx::newtype::Convertable,
+                                  cxx::newtype::Incrementable,
+                                  cxx::newtype::ConstructByValueCopy,
+                                  cxx::newtype::MoveConstructable>
+{
+    using ThisType::ThisType;
+};
+
+static const id_t NULL_POINTER_ID{std::numeric_limits<uint64_t>::max()};
```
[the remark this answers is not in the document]
mossmaurice: I agree with you that `NewType` should be usable as `constexpr`. That's a basic requirement. I think it's definitely do-able, but I would suggest to implement that in a follow-up PR. For me it would be ok to merge this PR here and continue with a follow-up one.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -43,15 +45,15 @@ class PointerRepository
...
     };
 
+    static const id_t INVALID_ID;
+
     /// @note 0 is a special purpose id and reserved
     /// id 0 is reserved to interpret the offset just as a raw pointer,
     /// i.e. its corresponding base ptr is 0
-    static constexpr size_t MIN_ID = 1U;
-    static constexpr size_t MAX_ID = CAPACITY - 1U;
+    static const id_t MIN_ID;
```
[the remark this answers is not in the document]
mossmaurice: Closing as duplicate

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -53,31 +54,32 @@ inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t
 template <typename id_t, typename ptr_t, uint64_t CAPACITY>
 inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(const ptr_t ptr, uint64_t size) noexcept
 {
-    for (id_t id = 1U; id <= MAX_ID; ++id)
+    for (id_t id{1U}; id <= MAX_ID; ++id)
```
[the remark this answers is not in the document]
mossmaurice: The reason is it's not covertable so the c'tor needs to be called explicitly.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -127,7 +129,7 @@ inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::searchId(ptr_t ptr) const
     /// by setting id to 0
     /// rationale: test cases work without registered shared memory and require
     /// this at the moment to avoid fundamental changes
-    return 0U;
+    return id_t{0U};
```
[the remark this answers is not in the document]
mossmaurice: Here, for me it's clearer what is happening if we write it explicitly. One can not forget that a "special type" `id_t` is returned.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
```diff
@@ -64,23 +64,27 @@ class RelativePointer : public BaseRelativePointer
...
 
     /// @brief read-only access to the underlying object
     /// @return a const pointer to the underlying object
-    T* operator->() const noexcept;
+    const T* operator->() const noexcept;
 
     /// @brief access the underlying object
     /// @return a pointer to the underlying object
-    T* get() const noexcept;
+    T* get() noexcept;
```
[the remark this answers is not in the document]
mossmaurice: @MatthiasKillat So you mean: Only 

```cpp
    T* get() const noexcept;
```

instead of

```cpp
    T* get() noexcept;

```

?

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender_data.hpp
```diff
@@ -39,7 +39,7 @@ struct ChunkSenderData : public ChunkDistributorDataType
 
     using ChunkDistributorData_t = ChunkDistributorDataType;
 
-    const rp::RelativePointer<mepoo::MemoryManager> m_memoryMgr;
+    rp::RelativePointer<mepoo::MemoryManager> m_memoryMgr;
```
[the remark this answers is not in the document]
mossmaurice: IIRC this is because of the `T* get() const noexcept` change.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -38,9 +40,9 @@ struct Data
     uint32_t Data2 = 72;
 };
 
-static constexpr uint64_t SHARED_MEMORY_SIZE = 4096 * 32;
-static constexpr uint64_t NUMBER_OF_MEMORY_PARTITIONS = 2U;
-static uint8_t memoryPatternValue = 1U;
+constexpr uint64_t SHARED_MEMORY_SIZE = 4096UL * 32UL;
```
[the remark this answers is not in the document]
mossmaurice: We should remove all the `U`s and `ULL`s in a separate PR. I can also remove them here for the scope of the PR if you like.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -29,6 +29,8 @@ using namespace iox::rp;
 
 struct Data
 {
+    // NOLINTJUSTIFICATION Ok-ish for tests
```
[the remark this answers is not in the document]
mossmaurice: I don't think something more formal is needed for tests.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -65,10 +68,14 @@ class RelativePointer_test : public Test
 
     uint8_t* partitionPtr(uint32_t partition)
     {
+        // NOLINTJUSTIFICATION Ok-ish for tests
+        // NOLINTNEXTLINE cppcoreguidelines-pro-bounds-constant-array-index)
         return memoryPartition[partition];
```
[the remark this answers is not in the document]
mossmaurice: @MatthiasKillat As you can derive from the rule name, the direct access of C arrays is not recommend and access through `std::span` is preferred in such cases.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/newtype/decrementable.hpp
```diff
@@ -0,0 +1,54 @@
...
+{
+template <typename T>
+// not required since a default'ed destructor does not define a destructor, hence the copy/move operations are
+// not deleted.
+// the only adaptation is that the dtor is protected to prohibit the user deleting the child type by
+// explicitly calling the destructor of the base type.
+// NOLINTNEXTLINE(cppcoreguidelines-special-member-functions, hicpp-special-member-functions)
+struct Decrementable
+{
+    friend T& operator--(T& rhs) noexcept
```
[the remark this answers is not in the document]
mossmaurice: @MatthiasKillat True, I'll add some tests for `NewType`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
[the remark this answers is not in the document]
MatthiasKillat: It maybe complains, but it is more an issue of small type vs. large type. It is still a small type. Either way is fine here.

Note that some objects are inherently value objects, such as many smart pointers. These are also small in general.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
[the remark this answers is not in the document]
MatthiasKillat: A minor downside.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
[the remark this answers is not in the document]
MatthiasKillat: But this is mostly important for the caller who an infer it from the function definition.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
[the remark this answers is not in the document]
MatthiasKillat: That is how `shared_ptr` does it, unless I miss something. And I can see why (as said before). Not super important but we have to keep in mind that it can be instantiated with `const T` and `T` for any `T` which is logically responsible for constness  of the return type.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_sender_data.hpp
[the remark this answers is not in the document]
MatthiasKillat: `memoryMgr` may change, but the pointer to it will not I think. This is about the latter.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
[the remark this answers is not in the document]
MatthiasKillat: You could still just write `4096U * 32U` and it would be fine.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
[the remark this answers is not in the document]
MatthiasKillat: Yes, but what is `ok-ish` supposed to mean? There is no useful info, so it should be shortened.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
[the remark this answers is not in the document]
MatthiasKillat: It has problems with regular array access? Oh my this is kind of a bad rule, as is the general C-array usage one.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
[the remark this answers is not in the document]
mossmaurice: Reverted the changes.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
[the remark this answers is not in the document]
mossmaurice: `constexpr`-ability will happen as part of #1554

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -37,13 +37,14 @@ inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t
     {
         return false;
     }
-    if (m_info[id].basePtr == nullptr)
+    if (m_info[static_cast<uint64_t>(id)].basePtr == nullptr)
```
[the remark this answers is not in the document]
mossmaurice: Removed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -37,13 +37,14 @@ inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t
...
     {
-        m_info[id].basePtr = ptr;
-        m_info[id].endPtr = reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
-        if (id > m_maxRegistered)
+        m_info[static_cast<uint64_t>(id)].basePtr = ptr;
```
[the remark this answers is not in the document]
mossmaurice: Removed

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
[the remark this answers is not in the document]
mossmaurice: Removed

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/newtype/decrementable.hpp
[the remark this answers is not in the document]
mossmaurice: Added the tests, but will move them to a separate PR.

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
```diff
@@ -159,25 +169,28 @@ BaseRelativePointer::id_t BaseRelativePointer::searchId(ptr_t ptr) noexcept
...
-    return getRepository().isValid(id);
+    return getRepository().isValid(static_cast<id_underlying_t>(id));
 }
 
-PointerRepository<BaseRelativePointer::id_t, BaseRelativePointer::ptr_t>& BaseRelativePointer::getRepository() noexcept
+PointerRepository<BaseRelativePointer::id_underlying_t, BaseRelativePointer::ptr_t>&
+BaseRelativePointer::getRepository() noexcept
 {
-    static PointerRepository<id_t, ptr_t> repository;
+    static PointerRepository<id_underlying_t, ptr_t> repository;
```
elBoberido, line 182: Just wondering if the `PointerRepository` should also use `id_t`?

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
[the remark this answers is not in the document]
elBoberido: It seems most of the time `// NOLINTNEXTLINE(bugprone-easily-swappable-parameters) used only for test purposes` is used

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -86,72 +96,88 @@ TYPED_TEST_SUITE(RelativePointer_test, Types);
...
         rp = typedPtr;
         EXPECT_EQ(rp.getOffset(), offset);
-        EXPECT_EQ(rp.getId(), 1);
+        EXPECT_EQ(rp.getId(), 1U);
         EXPECT_NE(rp, nullptr);
     }
 
     {
-        auto typedPtr = reinterpret_cast<TypeParam*>(ptr0);
+        auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(ptr0));
```
elBoberido, line 125: It took me some time to distinguish the difference when `reinterpret_cast` is used and when this construct with two `static_cast` is used. Since this is test code, how about just using `auto* typedPtr = reinterpret_cast<TypeParam*>(ptr0); // NOLINT`. It would help with readability and should be sufficient for this specific file. Maybe with a note like
```
// using only NOLINT for reinterpret_cast instead of NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-reinterpret-cast)
```

elBoberido (commented):
Oh boy, it must have been tedious to adjust the tests.

In general just a few minor things.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -51,28 +52,38 @@ namespace rp
...
+                                      cxx::newtype::DefaultConstructable,
+                                      cxx::newtype::CopyConstructable,
+                                      cxx::newtype::Convertable,
+                                      cxx::newtype::ConstructByValueCopy,
+                                      cxx::newtype::MoveConstructable>
+    {
+        using ThisType::ThisType;
+    };
+
+    using id_underlying_t = id_t::value_type;
```
MatthiasKillat, line 65: fairly verbose.

A potential solution could be to call the struct `Id` (it is in the namespace so it is not ambiguous) and the underlying type `id_t`. This is consistent with the rule too only use  `_t` to aliases (which currently is violated). I personally only use `_t` for aliases to primitive types.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -46,8 +48,8 @@ class PointerRepository
     /// @note 0 is a special purpose id and reserved
     /// id 0 is reserved to interpret the offset just as a raw pointer,
     /// i.e. its corresponding base ptr is 0
-    static constexpr size_t MIN_ID = 1U;
-    static constexpr size_t MAX_ID = CAPACITY - 1U;
+    static constexpr id_t MIN_ID{1U};
+    static constexpr id_t MAX_ID{CAPACITY - 1U};
```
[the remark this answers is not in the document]
MatthiasKillat: Can be done, though we control it and it is private. I know it is about future changes, but someone could also change the assert. Anyway, it does not hurt and only affects compile time performance minimally.

I am neutral on this.

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
```diff
@@ -21,13 +21,17 @@ namespace iox
 {
 namespace rp
 {
-BaseRelativePointer::BaseRelativePointer(ptr_t ptr, id_t id) noexcept
+// NOLINTJUSTIFICATION NewType is trivially copyable, hence pass by value is preferred
```
MatthiasKillat, line 24: It is also small if the wrapped type is only an integer of some kind. Unless the compiler creates more hidden data than I would expect.

Trivially copyable should not be the reason, size of the object and object semantics are (some types are naturally value types). There can be large trivially copyable objects which should not be passed by value.

Please change the justification.

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
```diff
@@ -21,13 +21,17 @@ namespace iox
...
+// NOLINTNEXTLINE(performance-unnecessary-value-param)
+BaseRelativePointer::BaseRelativePointer(const ptr_t ptr, const id_t id) noexcept
     : m_id(id)
     , m_offset(computeOffset(ptr))
 {
 }
 
-BaseRelativePointer::BaseRelativePointer(offset_t offset, id_t id) noexcept
+// NOLINTJUSTIFICATION NewType is trivially copyable, hence pass by value is preferred
```
MatthiasKillat, line 32: It is again about size. Please change the justification. This occurs more often later in the file, but I will not mark it.

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
MatthiasKillat: I personally would prefer to consider this in a larger refactoring. It is implementation detail.

The overall design is not ideal, but it works and I see no critical defects nor static code analysis warnings.

iceoryx_posh/source/mepoo/shm_safe_unmanaged_chunk.cpp
```diff
@@ -41,13 +41,14 @@ ShmSafeUnmanagedChunk::ShmSafeUnmanagedChunk(mepoo::SharedChunk chunk) noexcept
...
     {
-        rp::RelativePointer<mepoo::ChunkManagement> ptr = chunk.release();
+        rp::RelativePointer<mepoo::ChunkManagement> ptr{chunk.release()};
         auto id = ptr.getId();
         auto offset = ptr.getOffset();
         cxx::Ensures(id <= rp::RelativePointerData::ID_RANGE && "RelativePointer id must fit into id type!");
         cxx::Ensures(offset <= rp::RelativePointerData::OFFSET_RANGE
                      && "RelativePointer offset must fit into offset type!");
-        m_chunkManagement = rp::RelativePointerData(static_cast<rp::RelativePointerData::id_t>(id), offset);
+        /// @todo #1196 Unify types to uint64_t
```
MatthiasKillat, line 50: What is supposed to happen here in the todo?

iceoryx_posh/source/roudi/memory/memory_provider.cpp
```diff
@@ -117,9 +117,9 @@ cxx::expected<MemoryProviderError> MemoryProvider::destroy() noexcept
 
     if (!destructionResult.has_error())
     {
-        rp::BaseRelativePointer::unregisterPtr(m_segmentId);
+        rp::BaseRelativePointer::unregisterPtr(rp::BaseRelativePointer::id_t{m_segmentId});
         m_memory = nullptr;
-        m_size = 0u;
+        m_size = 0U;
```
MatthiasKillat, line 122: This is also an opportunity to start to get rid of the U suffix.

iceoryx_posh/test/moduletests/test_roudi_memory_provider.cpp
```diff
@@ -349,7 +349,7 @@ TEST_F(MemoryProvider_Test, SegmentIdValueAfterCreationIsValid)
     ASSERT_THAT(segmentId.has_value(), Eq(true));
     // the segment id being monotonic increasing is an implementation detail, in the case that the implementation
     // changes, just remove this check, since we already check to get a valid result
-    EXPECT_THAT(segmentId.value(), Eq(segmentIdOffset + 1));
+    EXPECT_THAT(segmentId.value(), Eq(segmentIdOffset + 1U));
```
MatthiasKillat, line 352: Is clang tidy complaining about the missing U? Otherwise we should not introduce them but slowly eliminate them.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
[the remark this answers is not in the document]
MatthiasKillat: It is not necessary to have the same justification comments, they just need to be sufficiently formal and convey the intention.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -40,10 +40,11 @@ inline bool PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(id_t id, ptr_t
...
         m_info[id].basePtr = ptr;
-        // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to the
-        // original type
-        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-        m_info[id].endPtr = reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
+        m_info[id].endPtr =
+            // AXIVION Next Construct AutosarC++19_03-A5.2.4 : Cast is needed for pointer arithmetic and casted back to
+            // the original type
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+            reinterpret_cast<ptr_t>(reinterpret_cast<uintptr_t>(ptr) + size - 1U);
```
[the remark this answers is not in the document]
mossmaurice: 🤷

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
mossmaurice: Could be done in future. It wasn't done now for the sake of easier refactoring.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
MatthiasKillat: We are really reinterpreting pointers here so a reinterpret cast is justified. Unlike static_cast it also cannot change the pointer (static_cast does this in some complex inheritance cases). static_cast should not be used here.

static_cast should be used for conversion between primitive types and casting within an inheritance hierarchy only. raw byte buffer casts (source `uint8_t*`, `void*` `char*`) to some other type T* are reinterpret_casts by nature (and requires us to know the actual type which we do here).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -86,72 +96,88 @@ TYPED_TEST_SUITE(RelativePointer_test, Types);
...
         RelativePointer<TypeParam> rp(typedPtr);
         EXPECT_EQ(rp.getOffset(), offset);
-        EXPECT_EQ(rp.getId(), 1);
+        EXPECT_EQ(rp.getId(), 1U);
         EXPECT_NE(rp, nullptr);
     }
 
     {
-        auto typedPtr = reinterpret_cast<TypeParam*>(ptr1);
+        auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(ptr1));
```
MatthiasKillat, line 155: please use reinterpret_cast

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -171,67 +199,75 @@ TYPED_TEST(RelativePointer_test, ConstrTests)
...
 
-    EXPECT_EQ(BaseRelativePointer::registerPtr(1, ptr0, SHARED_MEMORY_SIZE), true);
-    EXPECT_EQ(BaseRelativePointer::registerPtr(2, ptr1, SHARED_MEMORY_SIZE), true);
+    EXPECT_EQ(BaseRelativePointer::registerPtr(iox::rp::BaseRelativePointer::id_t{1U}, ptr0, SHARED_MEMORY_SIZE), true);
+    EXPECT_EQ(BaseRelativePointer::registerPtr(iox::rp::BaseRelativePointer::id_t{2U}, ptr1, SHARED_MEMORY_SIZE), true);
 
     {
         RelativePointer<TypeParam> rp;
-        auto typedPtr = reinterpret_cast<TypeParam*>(ptr0);
+        auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(ptr0));
```
MatthiasKillat, line 210: please use reinterpret_cast

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -171,67 +199,75 @@ TYPED_TEST(RelativePointer_test, ConstrTests)
...
         EXPECT_EQ(rp.getOffset(), offset);
-        EXPECT_EQ(rp.getId(), 1);
+        EXPECT_EQ(rp.getId(), 1U);
         EXPECT_NE(rp, nullptr);
     }
 
     {
         RelativePointer<TypeParam> rp;
-        auto typedPtr = reinterpret_cast<TypeParam*>(ptr1);
+        auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(ptr1));
```
MatthiasKillat, line 243: please use reinterpret_cast

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -243,158 +279,162 @@ TYPED_TEST(RelativePointer_test, AssignmentOperatorTests)
...
-    RelativePointer<TypeParam> rp1(typedPtr, 1);
-    EXPECT_EQ(rp1.registerPtr(1, typedPtr), true);
-    EXPECT_EQ(rp1.getOffset(), reinterpret_cast<std::ptrdiff_t>(ptr));
-    EXPECT_EQ(rp1.getId(), 1);
-
-    int offset = SHARED_MEMORY_SIZE / 2;
-    auto addressAtOffset = reinterpret_cast<TypeParam*>(ptr + offset);
-    RelativePointer<TypeParam> rp2(addressAtOffset, 1);
+    auto* ptr = this->partitionPtr(0U);
+    auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(ptr));
```
MatthiasKillat, line 295: please use reinterpret_cast

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -243,158 +279,162 @@ TYPED_TEST(RelativePointer_test, AssignmentOperatorTests)
...
-    RelativePointer<TypeParam> rp1(typedPtr, 1);
-    EXPECT_EQ(rp1.registerPtr(1, typedPtr), true);
-    EXPECT_EQ(BaseRelativePointer::getOffset(1, ptr), 0);
-
-    int offset = SHARED_MEMORY_SIZE / 2;
-    auto addressAtOffset = reinterpret_cast<TypeParam*>(ptr + offset);
-    RelativePointer<TypeParam> rp2(addressAtOffset, 1);
-    EXPECT_EQ(BaseRelativePointer::getOffset(1, addressAtOffset), offset);
+    auto* ptr = this->partitionPtr(0U);
+    auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(ptr));
```
MatthiasKillat, line 318: please use reinterpret_cast

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -243,158 +279,162 @@ TYPED_TEST(RelativePointer_test, AssignmentOperatorTests)
...
+TYPED_TEST(RelativePointer_test, GetPtrReturnsAddressWithCorrectOffset)
 {
     ::testing::Test::RecordProperty("TEST_ID", "4fadf89f-69c0-4058-8995-a98e2e3334b2");
-    auto ptr = this->partitionPtr(0);
-    auto typedPtr = reinterpret_cast<TypeParam*>(ptr);
-    RelativePointer<TypeParam> rp1(typedPtr, 1);
-    EXPECT_EQ(rp1.registerPtr(1, typedPtr), true);
-    EXPECT_EQ(BaseRelativePointer::getPtr(1, 0), ptr);
+    auto* ptr = this->partitionPtr(0U);
+    auto* typedPtr = static_cast<TypeParam*>(static_cast<void*>(this->partitionPtr(0U)));
```
MatthiasKillat, line 336: please use reinterpret_cast

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -243,158 +279,162 @@ TYPED_TEST(RelativePointer_test, AssignmentOperatorTests)
...
 
-    RelativePointer<TypeParam> rp1(typedPtr, 1);
+    RelativePointer<TypeParam> rp1(typedPtr, iox::rp::BaseRelativePointer::id_t{1U});
 
-    EXPECT_EQ(rp1.registerPtr(1, typedPtr), true);
-    EXPECT_EQ(rp1.registerPtr(1, typedPtr), false);
-    EXPECT_EQ(rp1.unregisterPtr(1), true);
-    EXPECT_EQ(rp1.registerPtr(1, typedPtr), true);
+    EXPECT_EQ(rp1.registerPtr(iox::rp::BaseRelativePointer::id_t{1U}, typedPtr), true);
```
MatthiasKillat, line 356: If this is deemed to verbose we could use an alias in the test file (`id_t` or `Id` etc.). I would prefer this.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: For me tests are often exemplary code where newbies can learn from. Hence, I would prefer to only use `reinterpret_cast` if really needed. I added a comment above the all the lines that use the cast via `void*` to make the intention clear.

MatthiasKillat (changes requested):

iceoryx_posh/source/roudi/memory/memory_provider.cpp
mossmaurice: That will be done on a separate PR. Feel free to go for it :grin:

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -65,10 +68,17 @@ class RelativePointer_test : public Test
 
     uint8_t* partitionPtr(uint32_t partition)
     {
+        // NOLINTJUSTIFICATION Ok-ish for tests
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
mossmaurice: `id_t` clashes with some system header definition and it not recommended to be used (see #1549). The current naming clearly states the intention, I'll leave it as it is.

iceoryx_posh/test/moduletests/test_roudi_memory_provider.cpp
mossmaurice: That will be done on a separate PR. Wanna give it a go? ;)

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: `static_cast` has several advantages e.g. it can be used during compile-time, see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1521#discussion_r936463947).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: `static_cast` has several advantages e.g. it can be used during compile-time, see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1521#discussion_r936463947).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: `static_cast` has several advantages e.g. it can be used during compile-time, see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1521#discussion_r936463947).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: `static_cast` has several advantages e.g. it can be used during compile-time, see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1521#discussion_r936463947).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: `static_cast` has several advantages e.g. it can be used during compile-time, see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1521#discussion_r936463947).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: `static_cast` has several advantages e.g. it can be used during compile-time, see [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1521#discussion_r936463947).

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: Once, the AUTOSAR rule violations get fixed `id_t` needs to be renamed anyway. I would suggest `identifier_t`.

iceoryx_hoofs/source/relocatable_pointer/base_relative_pointer.cpp
mossmaurice: True indeed, adapted.

iceoryx_posh/source/mepoo/shm_safe_unmanaged_chunk.cpp
mossmaurice: `RelativePointerData` defines it as `using id_t = uint16_t` and `BaseRelativePointer` as `uint64_t`. This should be addressed in a follow-up PR.

elBoberido: I'd rather make a `uint16_t` instead of a `uint64_t`. We have to store this together with the offset as 64 bit in the shared memory. When the id is an `uin16_t` there is less to check for in the `RelativePointerData` :)

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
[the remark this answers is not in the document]
elBoberido: Could you move the comment again above the line with reinterpret_cast and format the file? You have to keep in mind that some older devs might already have forgotten what happened before the comment when there are so many commented lines :sweat_smile:

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
elBoberido: Well, `reinterpret_cast<T*>(uPtr)` and `static_cast<T*>(static_cast<void*>(uPtr))` are doing the exact same thing and there is no harm in doing the former one, except a tool telling us not to use it. I'm fine either way.

Personally I'm more annoyed by all this justifications and NOLINTS. Maybe we can just use one big `NOLINTBEGIN` block to disable the warning in cases where it is unavoidable to use a specific construct?

I'm fine to keep it as is for this test.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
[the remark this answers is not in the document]
mossmaurice: @elBoberido I suppose that's not possible, because when using "next line" both clang-tidy and Axivion really only suppress the next line and that wouldn't work in that case.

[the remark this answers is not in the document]
mossmaurice: @elBoberido Ok got it now. Fixed!

elBoberido (dismissed):

elBoberido (approved):

MatthiasKillat (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
MatthiasKillat: would not clash here due to namespace / in class definition though

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: @elBoberido I've not looked at the implementation of either `reinterpret_cast` or `static_cast` but I'm sure they are not doing exactly the same thing, because `static_cast` can be used during compile-time and `reinterpret_cast` can not. It's also likely highly compiler dependent. We can think about writing a wrapper that beforms the trampoline cast.

Issue #1394, opened 2022-06-13
Address Autosar Violations using Axivion


saif-at-github:
## Brief feature description

Fix the Autosar violations reported by Axivion

## Detailed information

The static code analysis using Axivion shall be performed and all the potential bug and violations shall be addressed.

#### `cxx`

* [x] `algorithm.hpp` @FerdinandSpitzschnueffler 
* [x] `deadline_timer.hpp` @FerdinandSpitzschnueffler 
* [x] `duration.hpp` @ithier 
* [x] `expected.hpp` @saif-at-github
* [x] `filesystem.hpp` @elfenpiff 
* [x] `function.hpp` @MatthiasKillat 
* [x] `function_ref.hpp` @mossmaurice 
* [x] `functional_interface.hpp` @FerdinandSpitzschnueffler  
* [x] `helplets.hpp` @saif-at-github
* [x] `new_type.hpp` @FerdinandSpitzschnueffler 
* [x] `optional.hpp` @saif-at-github
* [ ] `requires.hpp`
* [x] `scoped_static.hpp` @saif-at-github
* [x]  `scope_guard` @FerdinandSpitzschnueffler 
* [x] `stack` @FerdinandSpitzschnueffler 
* [x] `string.hpp` @saif-at-github
* [x] `string_internal.hpp` @saif-at-github
* [x] `type_traits.hpp` @saif-at-github
* [x] `types.hpp` @saif-at-github
* [x] `unique_ptr.hpp` @saif-at-github
* [x] `variant.hpp` @FerdinandSpitzschnueffler  
* [x] `vector.hpp` @saif-at-github 
* [x] `attributes.hpp` @saif-at-github
* [x] `base_relative_pointer.hpp` @saif-at-github
* [x] `pointer_repository.hpp` @saif-at-github
* [x] `relative_pointer.hpp` @saif-at-github
* [x] `relative_pointer_data.hpp` @saif-at-github
* [x] `uninitialized_array.hpp` @saif-at-github 

### `rp`

* [x] `RelativePointerData` @mossmaurice

PR #1582, opened 2022-08-17
iox #1394 fix axivion violation for string

saif-at-github:

1. [x] ~~Tests follow the [best practice for testing][testing]~~
1. [x] Update the PR title
   - Link to the relevant issue
1. [x] All touched (C/C++) source code files are added to `./clang-tidy-diff-scans.txt`
1. [ ] Assign PR to reviewer

- [ ] All touched (C/C++) source code files have been added to `./clang-tidy-diff-scans.txt`

## References

- Relates #1394

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
@@ -30,20 +30,22 @@ namespace iox
 {
 namespace cxx
 {
+// AXIVION DISABLE STYLE AutosarC++19_03-A18.1.1 : C-array type usage is intentional
+
 template <typename T, typename ReturnType>
 using IsStringOrCharArrayOrChar =
-    typename std::enable_if<(is_cxx_string<T>::value || is_char_array<T>::value || std::is_same<T, std::string>::value
-                             || std::is_same<T, char>::value),
+    typename std::enable_if<((is_cxx_string<T>::value || is_char_array<T>::value)
+                             || (std::is_same<T, std::string>::value || std::is_same<T, char>::value)),
                             ReturnType>::type;
 
 template <typename T, typename ReturnType>
-using IsStringOrCharArray =
-    typename std::enable_if<(is_cxx_string<T>::value || is_char_array<T>::value || std::is_same<T, std::string>::value),
-                            ReturnType>::type;
+using IsStringOrCharArray = typename std::enable_if<((is_cxx_string<T>::value || is_char_array<T>::value)
+                                                     || std::is_same<T, std::string>::value),
+                                                    ReturnType>::type;
 
 template <typename T, typename ReturnType>
 using IsStdStringOrCharArrayOrChar =
-    typename std::enable_if<(is_char_array<T>::value || std::is_same<T, std::string>::value
+    typename std::enable_if<((is_char_array<T>::value || std::is_same<T, std::string>::value)
                              || std::is_same<T, char>::value),
                             ReturnType>::type;
 
@@ -53,8 +55,8 @@ using IsCxxStringOrCharArray =
 
 template <typename T1, typename T2, typename ReturnType>
 using IsCxxStringOrCharArrayOrChar =
-    typename std::enable_if<(is_char_array<T1>::value || is_cxx_string<T1>::value || std::is_same<T1, char>::value)
-                                && (is_char_array<T2>::value || is_cxx_string<T2>::value
+    typename std::enable_if<((is_char_array<T1>::value || is_cxx_string<T1>::value) || std::is_same<T1, char>::value)
+                                && ((is_char_array<T2>::value || is_cxx_string<T2>::value)
                                     || std::is_same<T2, char>::value),
                             ReturnType>::type;
 
@@ -62,8 +64,8 @@ template <typename T1, typename T2, typename ReturnType>
 using IsCxxStringAndCxxStringOrCharArrayOrChar =
     typename std::enable_if<((is_char_array<T1>::value || std::is_same<T1, char>::value) && is_cxx_string<T2>::value)
                                 || (is_cxx_string<T1>::value
-                                    && (is_char_array<T2>::value || std::is_same<T2, char>::value))
-                                || (is_cxx_string<T1>::value && is_cxx_string<T2>::value),
+                                    && ((is_char_array<T2>::value || std::is_same<T2, char>::value)
+                                        || (is_cxx_string<T1>::value && is_cxx_string<T2>::value))),
                             ReturnType>::type;
 
 /// @brief concatenates two iox::cxx::strings/string literals/chars
@@ -94,6 +96,8 @@ template <typename T1, typename T2, typename... Targs>
 IsCxxStringOrCharArrayOrChar<T1, T2, string<internal::SumCapa<T1, T2, Targs...>::value>>
 concatenate(const T1& str1, const T2& str2, const Targs&... targs) noexcept;
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : operator+ is defined within iox::cxx namespace which prevents easy
+// misuse
 /// @brief concatenates two iox::cxx::strings or one iox::cxx::string and one string literal/char; concatenation of two
 /// string literals/chars is not possible
 ///
@@ -115,7 +119,7 @@ constexpr TruncateToCapacity_t TruncateToCapacity{};
 /// @brief string implementation with some adjustments in the API, because we are not allowed to throw exceptions or use
 /// heap.
 template <uint64_t Capacity>
-class string
+class string final
 {
     static_assert(Capacity > 0U, "The capacity of the fixed string must be greater than 0!");
 
@@ -227,7 +231,7 @@ class string
     /// @endcode
     // TruncateToCapacity_t is a compile time variable to distinguish between constructors
     // NOLINTNEXTLINE(hicpp-named-parameter, readability-named-parameter)
-    string(TruncateToCapacity_t, const char* const other) noexcept;
+    string(TruncateToCapacity_t, const char* const other) noexcept(false);
 
     /// @brief conversion constructor for std::string to string which truncates characters if the std::string size is
     /// greater than the string capacity
@@ -271,7 +275,7 @@ class string
     /// @endcode
     // TruncateToCapacity_t is a compile time variable to distinguish between constructors
     // NOLINTNEXTLINE(hicpp-named-parameter, readability-named-parameter)
-    string(TruncateToCapacity_t, const char* const other, const uint64_t count) noexcept;
+    string(TruncateToCapacity_t, const char* const other, const uint64_t count) noexcept(false);
 
     /// @brief assigns a char array to string with compile time check if the array size is less than or equal
     /// to the string capacity
@@ -359,54 +363,6 @@ class string
     template <typename T>
     IsStringOrCharArray<T, int64_t> compare(const T& other) const noexcept;
 
-    /// @brief checks if self is equal to rhs
-    ///
-    /// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
-    ///
-    /// @return true if both strings are equal, otherwise false
-    template <typename T>
-    IsStringOrCharArrayOrChar<T, bool> operator==(const T& rhs) const noexcept;
-
-    /// @brief checks if self is not equal to rhs
-    ///
-    /// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
-    ///
-    /// @return true if both strings are not equal, otherwise false
-    template <typename T>
-    IsStringOrCharArrayOrChar<T, bool> operator!=(const T& rhs) const noexcept;
-
-    /// @brief checks if self is less than rhs, in lexicographical order
-    ///
-    /// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
-    ///
-    /// @return true if self is less than rhs, otherwise false
-    template <typename T>
-    IsStringOrCharArrayOrChar<T, bool> operator<(const T& rhs) const noexcept;
-
-    /// @brief checks if self is less than or equal to rhs, in lexicographical order
-    ///
-    /// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
-    ///
-    /// @return true if self is less than or equal to rhs, otherwise false
-    template <typename T>
-    IsStringOrCharArrayOrChar<T, bool> operator<=(const T& rhs) const noexcept;
-
-    /// @brief checks if self is greater than rhs, in lexicographical order
-    ///
-    /// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
-    ///
-    /// @return true if self is greater than rhs, otherwise false
-    template <typename T>
-    IsStringOrCharArrayOrChar<T, bool> operator>(const T& rhs) const noexcept;
-
-    /// @brief checks if self is greater than or equal to rhs, in lexicographical order
-    ///
-    /// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
-    ///
-    /// @return true if self is greater than or equal to rhs, otherwise false
-    template <typename T>
-    IsStringOrCharArrayOrChar<T, bool> operator>=(const T& rhs) const noexcept;
-
     /// @brief compares self and a char
     ///
     /// @param [in] other is the char to compare with self
@@ -441,10 +397,12 @@ class string
     /// @brief clears the content of the string
     constexpr void clear() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A13.5.2 , AutosarC++19_03-A13.5.3: used for interoperability with
+    // std::string
     /// @brief converts the string to a std::string
     ///
     /// @return a std::string with data equivalent to those stored in the string
-    // NOLINTNEXTLINE(hicpp-explicit-conversions) @todo iox-#260 remove this conversion and implement toStdString method
+    // NOLINTNEXTLINE(hicpp-explicit-conversions) todo #1196 remove this conversion and implement toStdString method
     operator std::string() const noexcept;
 
     /// @brief since there are two valid options for what should happen when appending a string larger than this'
@@ -585,7 +543,7 @@ class string
 
     template <typename T1, typename T2>
     friend IsCxxStringOrCharArrayOrChar<T1, T2, string<internal::SumCapa<T1, T2>::value>>
-    concatenate(const T1& t1, const T2& t2) noexcept;
+    concatenate(const T1& str1, const T2& str2) noexcept;
 
   private:
     /// @brief copies rhs fixed string to this with compile time check whether rhs capacity is less than or equal to
@@ -612,6 +570,8 @@ class string
     uint64_t m_rawstringSize{0U};
 };
 
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with std::string, char array or
+// char is also intended
 /// @brief checks if a lhs std::string, char array or char is equal to a rhs iox::cxx::string
 ///
 /// @param [in] rhs is the iox::cxx::string
@@ -672,7 +632,63 @@ IsStdStringOrCharArrayOrChar<T, bool> operator>=(const T& lhs, const string<Capa
 ///
 /// @return the stream output of the fixed string
 template <uint64_t Capacity>
-inline std::ostream& operator<<(std::ostream& stream, const string<Capacity>& str) noexcept;
+inline std::ostream& operator<<(std::ostream& stream, const string<Capacity>& str) noexcept(false);
+
+/// @brief checks if self is equal to rhs
+///
+/// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
+///
+/// @return true if both strings are equal, otherwise false
+template <typename T, uint64_t Capacity>
+IsStringOrCharArrayOrChar<T, bool> operator==(const string<Capacity>& lhs, const T& rhs) noexcept;
+
+/// @brief checks if self is not equal to rhs
+///
+/// @param [in] lhs is the iox::cxx::string
+/// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
+///
+/// @return true if both strings are not equal, otherwise false
+template <typename T, uint64_t Capacity>
+IsStringOrCharArrayOrChar<T, bool> operator!=(const string<Capacity>& lhs, const T& rhs) noexcept;
+
+/// @brief checks if self is less than rhs, in lexicographical order
+///
+/// @param [in] lhs is the iox::cxx::string
+/// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
+///
+/// @return true if self is less than rhs, otherwise false
+template <typename T, uint64_t Capacity>
+IsStringOrCharArrayOrChar<T, bool> operator<(const string<Capacity>& lhs, const T& rhs) noexcept;
+
+/// @brief checks if self is less than or equal to rhs, in lexicographical order
+///
+/// @param [in] lhs is the iox::cxx::string
+/// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
+///
+/// @return true if self is less than or equal to rhs, otherwise false
+template <typename T, uint64_t Capacity>
+IsStringOrCharArrayOrChar<T, bool> operator<=(const string<Capacity>& lhs, const T& rhs) noexcept;
+
+/// @brief checks if self is greater than rhs, in lexicographical order
+///
+/// @param [in] lhs is the iox::cxx::string
+/// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
+///
+/// @return true if self is greater than rhs, otherwise false
+template <typename T, uint64_t Capacity>
+IsStringOrCharArrayOrChar<T, bool> operator>(const string<Capacity>& lhs, const T& rhs) noexcept;
+
+/// @brief checks if self is greater than or equal to rhs, in lexicographical order
+///
+/// @param [in] lhs is the iox::cxx::string
+/// @param [in] rhs is the iox::cxx::string, std::string, char array or char to compare with self
+///
+/// @return true if self is greater than or equal to rhs, otherwise false
+template <typename T, uint64_t Capacity>
+IsStringOrCharArrayOrChar<T, bool> operator>=(const string<Capacity>& lhs, const T& rhs) noexcept;
+// AXIVION ENABLE STYLE AutosarC++19_03-A13.5.5
+// AXIVION ENABLE STYLE AutosarC++19_03-A18.1.1
+
 } // namespace cxx
 } // namespace iox
 #include "iceoryx_hoofs/internal/cxx/string.inl"
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
@@ -231,7 +231,7 @@ class string final
     /// @endcode
     // TruncateToCapacity_t is a compile time variable to distinguish between constructors
     // NOLINTNEXTLINE(hicpp-named-parameter, readability-named-parameter)
-    string(TruncateToCapacity_t, const char* const other) noexcept(false);
+    string(TruncateToCapacity_t, const char* const other) noexcept;
 
     /// @brief conversion constructor for std::string to string which truncates characters if the std::string size is
     /// greater than the string capacity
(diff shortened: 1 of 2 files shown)


mossmaurice:
@saif-at-github There are some conflicts in `string_internal.hpp`. Could you please rebase onto `master`?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -228,110 +230,110 @@ inline bool string<Capacity>::unsafe_assign(const std::string& str) noexcept
 
 template <uint64_t Capacity>
 template <typename T>
-inline IsStringOrCharArray<T, int64_t> string<Capacity>::compare(const T& other) const noexcept
+inline auto string<Capacity>::compare(const T& other) const noexcept -> IsStringOrCharArray<T, int64_t>
 {
-    uint64_t otherSize = internal::GetSize<T>::call(other);
-    auto result = memcmp(c_str(), internal::GetData<T>::call(other), std::min(m_rawstringSize, otherSize));
+    uint64_t otherSize{internal::GetSize<T>::call(other)};
+    auto result(memcmp(c_str(), internal::GetData<T>::call(other), std::min(m_rawstringSize, otherSize)));
```
FerdinandSpitzschnueffler, line 236: I think here Autosar actually complains about `result` not being `const`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -535,57 +538,60 @@ string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count)
     return true;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A14.7.1:
```
FerdinandSpitzschnueffler, line 541: I couldn't find a violation of this rule. Can you please check if this suppression can be removed?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -535,57 +538,60 @@ string<Capacity>::insert(const uint64_t pos, const T& str, const uint64_t count)
...
-    uint64_t length = std::min(count, m_rawstringSize - pos);
+    uint64_t length{std::min(count, m_rawstringSize - pos)};
     string subString;
     std::memcpy(&(subString.m_rawstring[0]), &m_rawstring[pos], length);
     subString.m_rawstring[length] = '\0';
     subString.m_rawstringSize = length;
     return subString;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A14.7.1:
```
FerdinandSpitzschnueffler, line 559: same here

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -227,7 +230,7 @@ class string
     /// @endcode
     // TruncateToCapacity_t is a compile time variable to distinguish between constructors
     // NOLINTNEXTLINE(hicpp-named-parameter, readability-named-parameter)
-    string(TruncateToCapacity_t, const char* const other) noexcept;
+    string(TruncateToCapacity_t, const char* const other) noexcept(false);
```
FerdinandSpitzschnueffler, line 233: @saif-at-github Can you please try to replace the `std::cerr` in the implementation with our `LogError()`? This may resolves the issue and you can remove the `false` again. You can find an example for the usage of `LogError` in the `SharedMemoryObject`.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -580,12 +588,14 @@ class string
     /// @note out of bounds access leads to program termination
     constexpr const char& operator[](const uint64_t pos) const noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A11.3.1: It's Ok
```
FerdinandSpitzschnueffler, line 591: For the justification you could state that:
- it's a template class
- the string needs to be constructible from a string with another capacity, comparable with another string, etc.
- this doesn't break the encapsulation

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -580,12 +588,14 @@ class string
...
     template <uint64_t N>
     friend class string;
 
+    // AXIVION Next Construct AutosarC++19_03-A11.3.1: It's Ok
```
FerdinandSpitzschnueffler, line 595: Here you could state that this is a factory method for the `cxx::string` and this doesn't break the encapsulation.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -53,19 +55,18 @@ using IsCxxStringOrCharArray =
 
 template <typename T1, typename T2, typename ReturnType>
 using IsCxxStringOrCharArrayOrChar =
-    typename std::enable_if<(is_char_array<T1>::value || is_cxx_string<T1>::value || std::is_same<T1, char>::value)
-                                && (is_char_array<T2>::value || is_cxx_string<T2>::value
-                                    || std::is_same<T2, char>::value),
+    typename std::enable_if<((is_char_array<T1>::value || is_cxx_string<T1>::value) || std::is_same<T1, char>::value)
+                                && ((is_char_array<T2>::value || is_cxx_string<T2>::value)
+                                    || (std::is_same<T2, char>::value)),
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler sure.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
saif-at-github: @FerdinandSpitzschnueffler I had tried it, it doesn't resolve the issue.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -500,8 +504,9 @@ class string
     ///
     /// @return true if the insertion was successful, otherwise false
     template <typename T>
-    IsCxxStringOrCharArray<T, bool> insert(const uint64_t pos, const T& str, const uint64_t count) noexcept;
+    auto insert(const uint64_t pos, const T& str, const uint64_t count) noexcept -> IsCxxStringOrCharArray<T, bool>;
 
+    // AXIVION Next Construct AutosarC++19_03-A14.7.1:
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler right, I see it is suppressed globally.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -513,6 +518,7 @@ class string
     /// string
     optional<string<Capacity>> substr(const uint64_t pos, const uint64_t count) const noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A14.7.1:
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler right, I see it is suppressed globally.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
saif-at-github: @FerdinandSpitzschnueffler Sure.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
saif-at-github: @FerdinandSpitzschnueffler Sure.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
saif-at-github: @FerdinandSpitzschnueffler right, it is suppressed globally

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
saif-at-github: @FerdinandSpitzschnueffler right, it is suppressed globally

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -634,14 +640,15 @@ inline constexpr char& string<Capacity>::at(const uint64_t pos) noexcept
 {
     // const_cast to avoid code duplication, safe since it's first casted to a const type and then the const is removed
     // again
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3: cast char to avoid implicit conversion
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler Sure.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
FerdinandSpitzschnueffler: @saif-at-github I think it's because `ostream` can throw. I suggest to replace all occurrences of `std::cout` and `std::cerr` in the `cxx:string` with `LogWarn`/`LogError` and ignore the warning for now since @elBoberido's logger refactoring should not throw.

FerdinandSpitzschnueffler: @MatthiasKillat Since @elBoberido is on vacation, can you maybe confirm?

FerdinandSpitzschnueffler (commented):
LGTM. Once we clarified how to deal with the `noexcept` related warning, it's ready to merge IMO.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
MatthiasKillat: @FerdinandSpitzschnueffler @saif-at-github I think keep the `noexcept` specifier as before for now. And yes, the streams can throw since they may allocate memory, so at least `bad_alloc` is possible.

I am not sure whether the tools deduce correctly that the exception specifier is incorrect. Probably because some function not marked as `noexcept` is called (which is too weak, as it only indicates that it cannot conclude that it does not throw).

TLDR: Keep the exception specifier for now, we need to take a closer look at the rule and maybe disable it.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -357,55 +359,55 @@ class string
     /// the contents of both strings are equal, an integer > 0 if the first character that does not match has a greater
     /// value in self than in other
     template <typename T>
-    IsStringOrCharArray<T, int64_t> compare(const T& other) const noexcept;
+    auto compare(const T& other) const noexcept -> IsStringOrCharArray<T, int64_t>;
```
MatthiasKillat, line 362: @saif-at-github @FerdinandSpitzschnueffler Why is this notation preferred? It is equivalent but usually `auto` return type is used if the return type depends on compile time magic. In this case it cannot be stated after `->` explicitly but here it is additionally stated as well so the usefulness of `auto`is lost to me (it is just a more verbose way to express the same thing).

I suggest to not use this notation together with `auto` except in the case I stated above (where it is unavoidable).

MatthiasKillat: @saif-at-github I know, but this is a purely syntactic rule and I would argue we should not follow it.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -608,7 +618,7 @@ class string
 
     // safe access is guaranteed since the char array is wrapped inside the string class
     // NOLINTNEXTLINE(hicpp-avoid-c-arrays, cppcoreguidelines-avoid-c-arrays)
-    char m_rawstring[Capacity + 1U]{'\0'};
+    char m_rawstring[Capacity + 1U]{};
```
MatthiasKillat, line 621: Ok, empty initializer list initializes it to `\0` as well.

@FerdinandSpitzschnueffler I am fairly certain that we can safe initialization effort by only initializing in the ctor (i.e. the actual data followed by a `\0`.). Ideally this would be the case, as 0 initialization is not free and should be avoided if not needed.

Nothing for now, but just to maybe consider in some future refactoring.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -632,16 +637,17 @@ inline IsStringOrCharArray<T, optional<uint64_t>> string<Capacity>::find_last_of
 template <uint64_t Capacity>
 inline constexpr char& string<Capacity>::at(const uint64_t pos) noexcept
 {
-    // const_cast to avoid code duplication, safe since it's first casted to a const type and then the const is removed
-    // again
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3: cast char to avoid implicit conversion.
```
FerdinandSpitzschnueffler, line 640:
```suggestion
    // AXIVION Next Construct AutosarC++19_03-A5.2.3: 
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -30,20 +30,21 @@ namespace iox
 {
 namespace cxx
 {
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: TODO
```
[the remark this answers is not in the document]
MatthiasKillat: I also tend to keeping the rule and moving the implementation out of class right in this pull request.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -77,8 +77,8 @@ using IsCxxStringAndCxxStringOrCharArrayOrChar =
 ///     auto bar = iox::cxx::concatenate(fuu, "ahc");
 /// @endcode
 template <typename T1, typename T2>
-IsCxxStringOrCharArrayOrChar<T1, T2, string<internal::SumCapa<T1, T2>::value>> concatenate(const T1& str1,
-                                                                                           const T2& str2) noexcept;
+auto concatenate(const T1& str1, const T2& str2) noexcept
```
MatthiasKillat, line 80: I would like to disable the rule about trailing return syntax for template types globally. If we decide to do so, then this change and others are not necessary anymore.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -138,14 +140,14 @@ class string
     /// @param [in] rhs is the copy origin
     ///
     /// @return reference to self
-    string& operator=(const string& rhs) noexcept;
+    string& operator=(const string& rhs) & noexcept;
```
MatthiasKillat, line 143: I think we have decided to disable this rule as well, so the `&` overload is not required.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -115,7 +117,7 @@ constexpr TruncateToCapacity_t TruncateToCapacity{};
 /// @brief string implementation with some adjustments in the API, because we are not allowed to throw exceptions or use
 /// heap.
 template <uint64_t Capacity>
-class string
+class string final
```
MatthiasKillat, line 120: I am torn about this and do not like to define classes as `final` (usually there is no need, and the STL also does not do this).

The idea appears to be to prevent a special case of inheritance that involves polymorphism when the dtor is not virtual. The downside is that it also prevents perfectly fine inheritance where this is not an issue. While we do not derive from this class, we also have no real issue with this kind of problems for multiple reasons. 

The problems require explicitly calling `delete` or the `dtor`, in the automatic case the correct dtor is called for potential derived classes.

TLDR: we can keep this `final` here but I still think the rule is unnecessary and counterproductive for useful designs.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
MatthiasKillat: @elBoberido @saif-at-github If I am not mistaken the plan is to not change the exception specifier anywhere for now, even if it means there are still some Axivion findings.

We will decide how to deal with this separately as it affects the API on a semantic level.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
MatthiasKillat: @FerdinandSpitzschnueffler @saif-at-github We should disable this rule after discussion if necessary.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
mossmaurice: @MatthiasKillat Is it not more about saying "Hey, use this class as it is. Please use composition rather than inheritance."? From a pure user-perspective I don't think there are many use-cases to inherit from classes in  `iceoryx_hoofs`.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
MatthiasKillat: @saif-at-github The rule is now disabled.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
elBoberido: @FerdinandSpitzschnueffler @saif-at-github @MatthiasKillat good news everyone. With the merge of the logger this warning is gone on master :partying_face:

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
MatthiasKillat: Needs to be undone.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -441,6 +443,7 @@ class string
     /// @brief clears the content of the string
     constexpr void clear() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A13.5.2 , AutosarC++19_03-A13.5.3: TODO
```
MatthiasKillat, line 446: We need to deal with the TODO. If we cannot define it as explicit (as it breaks other ocde) we need a justification.

Justification: used for interoperability with `std::string`

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -673,6 +631,60 @@ IsStdStringOrCharArrayOrChar<T, bool> operator>=(const T& lhs, const string<Capa
 /// @return the stream output of the fixed string
 template <uint64_t Capacity>
 inline std::ostream& operator<<(std::ostream& stream, const string<Capacity>& str) noexcept;
+
```
FerdinandSpitzschnueffler, line 634: Can you please move the above `operator<<`  above the `AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5:` line? The reason is that operator does not violate the AUTOSAR rule and all the comparison operators are grouped together.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -639,16 +567,17 @@ inline IsStringOrCharArray<T, optional<uint64_t>> string<Capacity>::find_last_of
...
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
     return const_cast<char&>(const_cast<const string<Capacity>*>(this)->at(pos));
 }
 
 template <uint64_t Capacity>
 inline constexpr const char& string<Capacity>::at(const uint64_t pos) const noexcept
 {
-    Expects(pos < size() && "Out of bounds access!");
+    // AXIVION Next Construct AutosarC++19_03-M5.3.1 : operand of logical operator is non-bool, enhances readability
+    Expects((pos < size()) && ("Out of bounds access!"));
```
FerdinandSpitzschnueffler, line 580: You could replace this line with
```suggestion
    ExpectsWithMsg((pos < size()), "Out of bounds access!");
```
Then you need no suppression and the test coverage will also increase.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -227,7 +229,7 @@ class string
     /// @endcode
     // TruncateToCapacity_t is a compile time variable to distinguish between constructors
     // NOLINTNEXTLINE(hicpp-named-parameter, readability-named-parameter)
-    string(TruncateToCapacity_t, const char* const other) noexcept;
+    string(TruncateToCapacity_t, const char* const other) noexcept(false);
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Please use `noexcept` everywhere without `(false)`.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
FerdinandSpitzschnueffler: Can you please explain why you reject this before resolving the conversation?

saif-at-github: @FerdinandSpitzschnueffler may I have missed the comments, not intended. I'll address this on next update.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler   I have removed all the  `(false)` used. There will be the another violations `A15-4-2` and `A15-5-3` are reported. I'm little unclear should suppress them. Could you suggest me ?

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @saif-at-github The other violations are all related to each other. I would propose to remove the `(false)` everywhere and ignore the related violations for now. We need to handle the problem that some functions can throw but are marked as `noexcept` probably in several places in our code base. So this could be part of a dedicated issue.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -441,10 +397,13 @@ class string
     /// @brief clears the content of the string
     constexpr void clear() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A13.5.2 , AutosarC++19_03-A13.5.3: used for interoperability with
+    // std::string
     /// @brief converts the string to a std::string
     ///
     /// @return a std::string with data equivalent to those stored in the string
-    // NOLINTNEXTLINE(hicpp-explicit-conversions) @todo iox-#260 remove this conversion and implement toStdString method
+    // NOLINTNEXTLINE(hicpp-explicit-conversions) @ iox-#260 remove this conversion and
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Please add the `todo` again.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -131,14 +132,17 @@ inline string<Capacity>::string(TruncateToCapacity_t, const char* const other, c
     }
     else if (Capacity < count)
     {
-#if defined(__GNUC__) && __GNUC__ == 8 && __GNUC_MINOR__ == 3
+// AXIVION DISABLE STYLE AutosarC++19_03-A16.0.1: conditional compilation is required for setting gcc diagnostics
```
FerdinandSpitzschnueffler, line 135: I think we should state why gcc 8 incorrectly warns here and therefore we chose the conditional compilation. On the other hand I'm not sure if maybe remove this constructor soon, see #1767 ...

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -400,14 +326,16 @@ inline string<Capacity>& string<Capacity>::move(string<N>&& rhs) noexcept
...
-    uint64_t strSize = rhs.size();
+    const uint64_t strSize{rhs.size()};
     std::memcpy(&(m_rawstring[0]), rhs.c_str(), strSize);
     m_rawstring[strSize] = '\0';
     m_rawstringSize = strSize;
     rhs.clear();
     return *this;
 }
 
+// AXIVION DISABLE Style AutosarC++19_03-M5.17.1: This is not used as shift operator but as stream operator and does not require to implement '<<='
```
FerdinandSpitzschnueffler, line 337: Doesn't this warning only belong to the next construct? Why did you choose `AXIVION DISABLE Style` instead of `AXIVION Next Construct`?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -400,14 +326,16 @@ inline string<Capacity>& string<Capacity>::move(string<N>&& rhs) noexcept
...
+
```
FerdinandSpitzschnueffler, line 338:
```suggestion
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -448,7 +376,10 @@ concatenate(const T1& str1, const T2& str2, const Targs&... targs) noexcept
     return concatenate(concatenate(str1, str2), targs...);
 }
 
+
```
FerdinandSpitzschnueffler, line 379:
```suggestion
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
```diff
@@ -664,6 +595,81 @@ inline constexpr const char& string<Capacity>::operator[](const uint64_t pos) co
...
+{
+    return (lhs.compare(rhs) > 0);
+}
+
+template <typename T, uint64_t Capacity>
+inline IsStringOrCharArrayOrChar<T, bool> operator>=(const string<Capacity>& lhs, const T& rhs) noexcept
+{
+    return (lhs.compare(rhs) >= 0);
+}
+// AXIVION ENABLE Style AutosarC++19_03-M5.17.1
```
FerdinandSpitzschnueffler, line 672: Shouldn't this be
```suggestion
// AXIVION ENABLE Style AutosarC++19_03-A13.5.5
```
?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
saif-at-github: @FerdinandSpitzschnueffler should I keep this suppression for now? once the c'tor removed this may revisit, based on this #1767

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
saif-at-github: @FerdinandSpitzschnueffler `M5.17.1` violation has been reported multiple place in the file so decided to use the `DISABLE Style` instead of `AXIVION Next Construct` let me check and  may use the  `AXIVION Next Construct`

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl
FerdinandSpitzschnueffler: Since the discussion is still ongoing and I'm quite sure that we will forget to revisit if it will be decided to keep the constructor, I would propose to extend the justification to
``` conditional compilation is required for setting gcc diagnostics since ggc 8 incorrectly warns here about out of bounds array access```

File: iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl [hunk 28]
```diff
@@ -23,6 +23,7 @@ namespace iox
 {
 namespace cxx
 {
+// AXIVION DISABLE STYLE AutosarC++19_03-A12.6.1 : false positive, field intentionally not initialized
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler this was reported in my local machine for `m_rawstring` but not in dashboad, I shall remove it.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/string.hpp
```diff
@@ -442,10 +398,13 @@ class string
...
     constexpr void clear() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A13.5.2 , AutosarC++19_03-A13.5.3: used for interoperability with
+    // std::string
     /// @brief converts the string to a std::string
     ///
     /// @return a std::string with data equivalent to those stored in the string
-    // NOLINTNEXTLINE(hicpp-explicit-conversions) @todo iox-#260 remove this conversion and implement toStdString method
+    // NOLINTNEXTLINE(hicpp-explicit-conversions) @todo iox-#260 remove this conversion and
+    // implement toStdString method
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler I see that, corrected now.

FerdinandSpitzschnueffler (approved):

MatthiasKillat (approved):

PR #1597, opened 2022-08-30
iox-#1596 Add check for `nullptr` in `RelativePointer`

### LINKED ISSUE #1596: `rp::RelativePointer::get()` does not check for `nullptr` before access
## Required information

**Operating system:**
 Ubuntu 20.04.3 LTS

**Compiler version:**
 GCC 8.4.0

 Accessing an empty `RelativePointer` leads to a crash

 Accessing an empty `RelativePointer` shall lead to a fatal error (defined behaviour)

Call `RelativePointer::get()` on empty object

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Add check for `nullptr` in `RelativePointer`
  * Activation/ deactivation of this check will be possible by the user after #1032 was implemented 

## References

- Closes #1596

### DIFF
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -46,6 +46,7 @@
 - Fix RouDi crash due to uninitialized `ServiceRegistry` chunk [\#1575](https://github.com/eclipse-iceoryx/iceoryx/issues/1575)
 - Add check in `cxx::unique_ptr::get` to avoid `nullptr` dereferencing [\#1571](https://github.com/eclipse-iceoryx/iceoryx/issues/1571)
 - Pass `CleanupCapacity` to underlying `cxx::function` in `ScopeGuard` (formerly known as `cxx::GenericRAII`) [\#1594](https://github.com/eclipse-iceoryx/iceoryx/issues/1594)
+- Add check in `RelativePointer::get` to avoid `nullptr` dereferencing [\#1596](https://github.com/eclipse-iceoryx/iceoryx/issues/1596)
 
 **Refactoring:**
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.inl
@@ -67,13 +67,15 @@ inline T* RelativePointer<T>::operator->() const noexcept
 template <typename T>
 inline T* RelativePointer<T>::get() const noexcept
 {
-    return static_cast<T*>(computeRawPtr());
+    auto ptr = static_cast<T*>(computeRawPtr());
+    cxx::Ensures(ptr != nullptr);
+    return ptr;
 }
 
 template <typename T>
 inline RelativePointer<T>::operator bool() const noexcept
 {
-    return get() != nullptr;
+    return computeRawPtr() != nullptr;
 }
 
 template <typename T>
--- a/iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
@@ -27,19 +27,6 @@ namespace
 using namespace ::testing;
 using namespace iox::rp;
 
-struct Data
-{
-    // NOLINTJUSTIFICATION Used only for test purposes
-    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
-    Data(uint32_t i, uint32_t j)
-        : Data1(i)
-        , Data2(j)
-    {
-    }
-    uint32_t Data1 = 27;
-    uint32_t Data2 = 72;
-};
-
 constexpr uint64_t SHARED_MEMORY_SIZE = 4096UL * 32UL;
 constexpr uint64_t NUMBER_OF_MEMORY_PARTITIONS = 2U;
 uint8_t memoryPatternValue = 1U;
@@ -185,7 +172,7 @@ TYPED_TEST(RelativePointer_test, ConstrTests)
 
     {
         RelativePointer<TypeParam> rp(nullptr);
-        EXPECT_EQ(rp, nullptr);
+        EXPECT_FALSE(rp);
     }
 
     {
@@ -194,7 +181,7 @@ TYPED_TEST(RelativePointer_test, ConstrTests)
         // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-reinterpret-cast)
         auto* typedPtr = reinterpret_cast<TypeParam*>(ptr1 + offset);
         RelativePointer<TypeParam> rp(typedPtr);
-        EXPECT_NE(rp, nullptr);
+        EXPECT_TRUE(rp);
     }
 }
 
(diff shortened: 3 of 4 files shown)


doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -45,6 +45,7 @@
 - Add "inline" keyword to smart_lock method implementation [\#1551](https://github.com/eclipse-iceoryx/iceoryx/issues/1551)
 - Fix RouDi crash due to uninitialized `ServiceRegistry` chunk [\#1575](https://github.com/eclipse-iceoryx/iceoryx/issues/1575)
 - Add check in `cxx::unique_ptr::get` to avoid `nullptr` dereferencing [\#1571](https://github.com/eclipse-iceoryx/iceoryx/issues/1571)
+- Add check in `RelativePointer::get` to avoid `nullptr` dereferencing [\#1596](https://github.com/eclipse-iceoryx/iceoryx/issues/1596)
```
MatthiasKillat, line 48: I am not a fan of this, it affects the hot paths (data transfer) and is not needed if used correctly. Note that the argument I often hear that it costs almost nothing due to branch prediction is not quite true.

In usual implementations (to my knowledge), the branch prediction buffer is limited, i.e. more branching in code will let the buffer overflow and evict some other branch prediction. If the potential branch is not in the buffer, it is like having no branch prediction at all.

Even with branch prediction, there are extra instructions. I can give more details, but this is certainly not the way to go to increase performance nor does it really provide safety.

Dereferencing a regular nullptr or smart pointer that is null is just as bad and can happen. It is possible for good reason, the responsibility is on the user side.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.inl
```diff
@@ -67,13 +67,15 @@ inline T* RelativePointer<T>::operator->() const noexcept
 template <typename T>
 inline T* RelativePointer<T>::get() const noexcept
 {
-    return static_cast<T*>(computeRawPtr());
+    auto ptr = static_cast<T*>(computeRawPtr());
+    cxx::Ensures(ptr != nullptr);
```
MatthiasKillat, line 71: This runtime check is bad IMO. It is also not like smart pointers work, e.g. STL `shared_ptr` and friends. I am against this check, not having this check is a conscious design decision.

MatthiasKillat (changes requested):
I do not think we should introduce this check in get as it is inconsistent, inefficient and just not required (see comments).

While I marked changes requested, the change would be to not introduce the check at all.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -194,7 +181,7 @@ TYPED_TEST(RelativePointer_test, ConstrTests)
         // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-reinterpret-cast)
         auto* typedPtr = reinterpret_cast<TypeParam*>(ptr1 + offset);
         RelativePointer<TypeParam> rp(typedPtr);
-        EXPECT_NE(rp, nullptr);
+        EXPECT_TRUE(rp);
```
MatthiasKillat, line 184: This is probably required now that implicit conversion is disabled. I wonder why it worked in the first place (without implicit conversion).

doc/website/release-notes/iceoryx-unreleased.md
mossmaurice: @MatthiasKillat Performance is important and this change is very likely in a hot path. However, I would appreciate if we would measure before stating "What about performance?". Same goes for me, if I implement such a change, in future I should attach some benchmark showing that the performance does not severely decrease.

As discussed today, I think such a check reflects our philosophy of `Design APIs in such a way they are easy to use and hard to misuse` and I would also vote for adding such a check if we can do it easily and it can avoid failures. However, I assumed our idea of #1032 included the features for users to disable `assert` / `Expects` / `Ensures` like constructs. In systems I worked with before those typically are active in debug mode and are not compile in with release mode (using `NDEBUG`). We could also think about offering a specific compiler switch. That would make sense for me for all the sanity check e.g. out-of-bounds in `cxx::vector`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.inl
mossmaurice: See my [answer](https://github.com/eclipse-iceoryx/iceoryx/pull/1597#discussion_r959976184) above. I think there was a misunderstanding. Closing here, so we can continue the discussion in the other thread.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
```diff
@@ -185,7 +172,7 @@ TYPED_TEST(RelativePointer_test, ConstrTests)
 
     {
         RelativePointer<TypeParam> rp(nullptr);
-        EXPECT_EQ(rp, nullptr);
+        EXPECT_FALSE(rp);
```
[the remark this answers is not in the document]
mossmaurice: You are right, `RelativePointer`'s should be able to be null as they behave like regular pointers. However, a "logical `nullptr`" communicates to the user that he should never call `operator->` or `get` on it.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
mossmaurice: Good question.. :thinking:

mossmaurice:
> While I marked changes requested, the change would be to not introduce the check at all.

@MatthiasKillat As written in [this thread](https://github.com/eclipse-iceoryx/iceoryx/pull/1597#discussion_r959357497) my suggestion is to go ahead and merge this PR. As part of #1032 we can then add the option to disable the assert-like checks.

doc/website/release-notes/iceoryx-unreleased.md
MatthiasKillat: @mossmaurice We do not need to measure to know that extra checks incur a performance hit. We only measure to understand how much on a **particular system** under specific circumstances.

It is also known how branch prediction works and one can derive an understanding about the incurred costs from this. And as I said, it does not only affect the particular branch introduced by the check, but other branches in the code as well (not enough time/space to elaborate here).

Regarding assert: yes, with a debug assert it is a different matter as the check can then be disabled. We will provide a `DEBUG_ASSERT` that can be used for this later which will not incur any cost if disabled.

iceoryx_hoofs/test/moduletests/test_relative_pointer.cpp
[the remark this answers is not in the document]
MatthiasKillat: `operator->` if used in the standard sense is questionable (actually it is also only a `get` with syntactic sugar and this sugar is the cause of problems). But `get` itself is not immediately harmful, only if the returned pointer is dereferenced.

MatthiasKillat (approved):
We can change it like this but we should replace the checks with a `DEBUG_ASSERT` (name not final) in the future.

We will distinguish between debug and release asserts, the latter being always active.

FerdinandSpitzschnueffler (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.inl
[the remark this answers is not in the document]
mossmaurice: @FerdinandSpitzschnueffler I'll add that on #1599

PR #1598, opened 2022-08-30
iox-1394 fix AUTOSAR violations in algorithm files

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

Despite other changes this PR
- renames `algorithm::max` and `algorithm::min` to make a clear distinction to `std::max` and `std::min`
- suppresses the warning about an equality check for floating-point-types in `doesContainValue`

## References

- Partially closes #1394

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
@@ -26,6 +26,8 @@ namespace iox
 {
 namespace algorithm
 {
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : standard library function intentionally overridden to retrieve the
+// maximum of an arbitrary amount of arguments without using an initializer list which uses the heap
 /// @brief Returns the maximum gained with operator<() of an arbitrary amount
 ///          of variables of the same type. Helper function which is required as generic
 ///          recursive template endpoint.
@@ -35,6 +37,8 @@ namespace algorithm
 template <typename T>
 constexpr T max(const T& left) noexcept;
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : standard library function intentionally overridden to retrieve the
+// maximum of an arbitrary amount of arguments without using an initializer list which uses the heap
 /// @brief Returns the maximum gained with operator<() of an arbitrary amount
 ///          of variables of the same type. Helper function which takes two arguments and returns the
 ///          greater one.
@@ -45,6 +49,9 @@ constexpr T max(const T& left) noexcept;
 template <typename T>
 constexpr T max(const T& left, const T& right) noexcept;
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : standard library function intentionally overridden to retrieve the
+// maximum of an arbitrary amount of arguments without using an initializer list which uses the heap
+// max value of more than two values, initializer list from STL uses heap
 /// @brief Returns the maximum gained with operator<() of an arbitrary amount
 ///          of variables of the same type.
 /// @param T type which implements operator<()
@@ -55,6 +62,8 @@ constexpr T max(const T& left, const T& right) noexcept;
 template <typename T, typename... Targs>
 constexpr T max(const T& left, const T& right, const Targs&... args) noexcept;
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : standard library function intentionally overridden to retrieve the
+// minimum of an arbitrary amount of arguments without using an initializer list which uses the heap
 /// @brief Returns the minimum gained with operator<() of an arbitrary amount
 ///          of variables of the same type. Helper function which is required as generic
 ///          recursive template endpoint.
@@ -64,6 +73,8 @@ constexpr T max(const T& left, const T& right, const Targs&... args) noexcept;
 template <typename T>
 constexpr T min(const T& left) noexcept;
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : standard library function intentionally overridden to retrieve the
+// minimum of an arbitrary amount of arguments without using an initializer list which uses the heap
 /// @brief Returns the minimum gained with operator<() of an arbitrary amount
 ///          of variables of the same type. Helper function which takes two arguments and returns the
 ///          smaller one.
@@ -74,6 +85,8 @@ constexpr T min(const T& left) noexcept;
 template <typename T>
 constexpr T min(const T& left, const T& right) noexcept;
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : standard library function intentionally overridden to retrieve the
+// minimum of an arbitrary amount of arguments without using an initializer list which uses the heap
 /// @brief Returns the minimum gained with operator<() of an arbitrary amount
 ///          of variables of the same type.
 /// @param T type which implements operator<()
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
@@ -99,10 +99,9 @@ template <typename T, typename CompareType, typename Next, typename... Remainder
 constexpr bool doesContainType() noexcept;
 
 /// @brief Finalizes the recursion of doesContainValue
-/// @param[in] v type of the value to check
 /// @return always false
 template <typename T>
-inline constexpr bool doesContainValue(const T v) noexcept;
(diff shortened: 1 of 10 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
```diff
@@ -81,46 +82,47 @@ template <typename T, typename... ValueList>
 inline constexpr bool
 doesContainValue(const T value, const T firstValueListEntry, const ValueList... remainingValueListEntries) noexcept
 {
+    // AXIVION Next Line AutosarC++19_03-M6.2.2 : intentional check for exact equality
```
mossmaurice, line 85: You could add an doxygen comment, that this method probably does not make sense for float or even restrict it only to be used for non-float types using a type trait/static_assert. See [here](https://rules.sonarsource.com/c/RSPEC-1244) for an example.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
```diff
@@ -71,6 +71,7 @@ inline constexpr bool doesContainType() noexcept
     return doesContainType<T, CompareType>() || doesContainType<T, Next, Remainder...>();
 }
 
+// AXIVION Next Construct AutosarC++19_03-A0.1.4 : function finalizes compile time recursion of doesContainValue
 template <typename T>
 inline constexpr bool doesContainValue(const T v IOX_MAYBE_UNUSED) noexcept
```
mossmaurice, line 76: Why not?

```suggestion
inline constexpr bool doesContainValue(const T) noexcept
```

mossmaurice (commented):
LGTM, left one question.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
FerdinandSpitzschnueffler: Unfortunately I can't restrict it because we use it with floating-point types and there it makes sense. If you're interested, `doesContainValue` is for instance called in `PosixCallVerificator::failureReturnValue` which in turn is called with `(-)HUGE_VAL(F)` in the `convert::fromString` specializations. Since the exact return value from the posix call is stored and we search for the exact failure return value, this is equality check is safe.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
FerdinandSpitzschnueffler: Seems like the variable name was introduced to fix a warning, see [here](https://github.com/eclipse-iceoryx/iceoryx/commit/928d02660cee561c4096d6bf461e4cc6be98c5c2) :thinking:  But let's try it again.

FerdinandSpitzschnueffler: I thought so too, but the clang tidy check doesn't complain.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
```diff
@@ -110,6 +110,9 @@ inline constexpr bool doesContainValue(const T v) noexcept;
 /// @param[in] value to look for in the ValueList
 /// @param[in] firstValueListEntry is the first variadic argument of ValueList
 /// @param[in] remainingValueListEntries are the remaining variadic arguments of ValueList
+/// @return true if value is contained in the ValueList, otherwise false
+/// @note be aware that value is tested for exact equality with the entries of ValueList which might lead to unexpected
+/// results for floating-point types
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I wasn't entirely happy with the wording either. Will rephrase the sentence.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
```diff
@@ -81,46 +82,47 @@ template <typename T, typename... ValueList>
...
-    uint64_t indexV2 = 0U;
-    uint64_t v1Size = v1.size();
-    uint64_t v2Size = v2.size();
+    uint64_t indexV1{0U};
+    uint64_t indexV2{0U};
+    const uint64_t v1Size{v1.size()};
+    const uint64_t v2Size{v2.size()};
 
-    while (indexV1 < v1Size && indexV2 < v2Size)
+    while ((indexV1 < v1Size) && (indexV2 < v2Size))
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Agreed, but ... :shrug:

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
mossmaurice: @FerdinandSpitzschnueffler Ok, got it. We are only compare those special floats and hence can count on an exact match.

@MatthiasKillat Well, I partly disagree here. This MISRA rule `Floating-point expressions shall not be directly or indirectly tested for equality or inequality` assumes some developers might not be aware that it doesn't make much sense to check equality/ inequality with `floats`. I think it could make sense to either

1) Document that this method should be only used with care for `float`
2) Rename it e.g. `s/doesContainValue/doesContainExactValue`

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
mossmaurice: @FerdinandSpitzschnueffler That means we can remove the variable?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
```diff
@@ -99,17 +99,19 @@ template <typename T, typename CompareType, typename Next, typename... Remainder
...
+inline constexpr bool doesContainValue(const T) noexcept;
 
 /// @brief Returns true if value of T is found in the ValueList, otherwise false
 /// @tparam T type of the value to check
 /// @tparam ValueList is a list of values to check for a specific value
 /// @param[in] value to look for in the ValueList
 /// @param[in] firstValueListEntry is the first variadic argument of ValueList
 /// @param[in] remainingValueListEntries are the remaining variadic arguments of ValueList
+/// @return true if value is contained in the ValueList, otherwise false
+/// @note be aware that value is tested for exact equality with the entries of ValueList and regular floating-point
```
mossmaurice, line 113: :heart: Maybe mention that comparing float exactly typically doesn't make much sense?

mossmaurice (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
FerdinandSpitzschnueffler: @mossmaurice Just to clarify. The equality check does not only work with those special floats like `HUGE_VAL`. If one stores 4.4, they will find 4.4 in the comparison. But if one stores a floating-point value resulting from a calculation, like 4*1.1, the comparison with 4.4 may return false. But this is the nature of floating-point operations. Regarding your suggestions: 1) I think it's now documented sufficiently in the header file. 2) Renaming the function is IMO unnecessary. Reading the current function name, I would expect an exact match and not a match "within a range". In the latter case, I would expect a different function name instead. Additionally, if developers are not aware of floating-point comparison behavior then a renaming would give them no hint either.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/algorithm.inl
FerdinandSpitzschnueffler: @mossmaurice So it seems.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/algorithm.hpp
FerdinandSpitzschnueffler: In my opinion the current documentation is sufficient. @MatthiasKillat What do you think?

MatthiasKillat: Whether it makes sense or not depends on context. In generic code w do not have this context so this is fine.

MatthiasKillat (approved):

PR #1599, opened 2022-08-31
iox-#605 Refactor `PointerRepository`

### LINKED ISSUE #605: Refactor Relocatable Pointer and Relative Pointer

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
* Remove `INVALID_ID` and introduce `RAW_POINTER_BEHAVIOUR`
* Rename `registerPtr` method and return `cxx::optional`
* Remove `isValid` and `print` methods
* Add sanity check if user didn't call `init` when using `LoFFLi`

## References

- Closes #605 (partly)


iceoryx_posh/include/iceoryx_posh/internal/mepoo/mepoo_segment.inl
```diff
@@ -76,8 +76,12 @@ inline SharedMemoryObjectType MePooSegment<SharedMemoryObjectType, MemoryManager
...
             .create()
             .and_then([this](auto& sharedMemoryObject) {
-                this->setSegmentId(static_cast<uint64_t>(iox::rp::BaseRelativePointer::registerPtr(
-                    sharedMemoryObject.getBaseAddress(), sharedMemoryObject.getSizeInBytes())));
+                auto maybeSegmentId = iox::rp::BaseRelativePointer::registerPtr(sharedMemoryObject.getBaseAddress(),
+                                                                                sharedMemoryObject.getSizeInBytes());
+                if (maybeSegmentId.has_value())
+                {
+                    this->setSegmentId(static_cast<uint64_t>(maybeSegmentId.value()));
+                }
```
mossmaurice, line 84: @MatthiasKillat As the sad path is a fatal error, I would create a new error enum entry here. What do you think?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
```diff
@@ -64,11 +64,13 @@ class RelativePointer : public BaseRelativePointer
     template <typename U = T>
     typename std::enable_if<!std::is_void<U>::value, const U&>::type operator*() const noexcept;
 
-    /// @brief read-only access to the underlying object
+    /// @brief read-only access to the underlying object. If the RelativePointer does not point to anything the
+    /// application might terminate.
     /// @return a pointer to the underlying object
     T* operator->() const noexcept;
```
elfenpiff, line 70: This breaks the const correctness and it should be 
`const T* operator->() const noexcept;`
Furthermore, we should introduce a `T* operator->() noexcept` to provide write access.

I know that the STL `unique_ptr` and `shared_ptr` does it differently but here I totally disagree with their behavior. I know that the relative pointer does not own the memory but it provide access to it. When a function would like to have a pointer as argument one can make sure with:
```cxx
void myFunc(const int * ptr);
```
That it is not changed but in here it would mean to generate a completely new type `RelativePointer<const T>` where the memory is always const. But most likely users will misuse it quickly and assume:
```cxx
void myFunc(const RelativePointer<int> ptr);
```
would assume that `myFunc` cannot change the content when it actually can.

See also the `cxx::unique_ptr` where we already ensure the const correctness with the proposal from above.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
```diff
@@ -64,11 +64,13 @@ class RelativePointer : public BaseRelativePointer
...
 
-    /// @brief access the underlying object
+    /// @brief access the underlying object. If the RelativePointer does not point to anything the
+    /// application might terminate.
     /// @return a pointer to the underlying object
     T* get() const noexcept;
```
elfenpiff, line 75: Same here, breaks the const correctness and it should be:
```cxx
T* get() noexcept;
```

elfenpiff (changes requested):
I would propose to add the suggested methods from my comment in this pull request. This should only add ~20 lines of additional code and 2 more unit tests for the const case.

As an alternative I would also be happy with an issue where this is tracked but I think writing the issue, tracking and following up on it takes more time than actually fixing it right away.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
mossmaurice: @elfenpiff I know what you mean, but I still disagree here. I believe a lot of effort has gone into designing the STL APIs and it's also quite proven in use. We might not even be aware for certain corner cases.

I would suggest to leave it as it is and also adapt it accordingly in `cxx::unique_ptr` as the user can control the constness like:

```
RelativePointer<T> p             ===> T* p
const RelativePointer<T> p       ===> T* const p
RelativePointer<const T> p       ===> const T* p
const RelativePointer<const T> p ===> const T* const p
```

To prevent the potential misuse, I would suggest to add the above to doxygen. What do you think?

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
mossmaurice: Closing this thread as it's a duplicate of [this one](https://github.com/eclipse-iceoryx/iceoryx/pull/1599#discussion_r963478931).

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
MatthiasKillat: @mossmaurice This is the intention and consistent with other fancy pointer APIs to my knowledge, so we should keep it like this. Adding the comment would help.

@elfenpiff The user assumption would be wrong then and she should use 

```
void myFunc(RelativePointer<const int> ptr);
```
The system is consistent and documentation is available.

Another change I would make is revert the name to its original `relative_ptr` for STL style consistency if possible given the current tooling (was like this before some changes before this PR).

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -362,6 +363,25 @@
...
     constexpr uint32_t MAX_VAL = algorithm::maxVal(3, 1890, 57);
     constexpr uint32_t MIN_VAL = algorithm::minVal(3, 1890, 57);
+
+20. `ReleativePointer::registerPtr` returns `cxx::optional`
+
+    ```cpp
+    // before
+    uint64_t id = RelativePointer::register(myPtr, 42);
+
+    if(id == INVALID_ID)
```
MatthiasKillat, line 373: Note that normally that should never happen if there is sufficient space in the repository. If we see that the current setting is too limiting we need to increase it, but it is optimized for thousands of ids but not millions (brute force search during registration, but we could hash later if needed).

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -45,33 +46,40 @@ class PointerRepository
...
 
     static_assert(MAX_ID >= MIN_ID, "MAX_ID must be greater or equal than MIN_ID!");
     static_assert(CAPACITY >= 2, "CAPACITY must be at least 2!");
 
   public:
-    static constexpr id_t INVALID_ID = std::numeric_limits<id_t>::max();
+    /// @note 0 is a special purpose id and reserved
+    /// id 0 is reserved to interpret the offset just as a raw pointer,
+    /// i.e. its corresponding base ptr is 0
+    static constexpr id_t RAW_POINTER_BEHAVIOUR{0};
```
MatthiasKillat, line 59:
```suggestion
    static constexpr id_t RAW_POINTER_BEHAVIOUR_ID{0};
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -45,33 +46,40 @@ class PointerRepository
...
+    ~PointerRepository() noexcept = default;
+
+    PointerRepository(const PointerRepository&) = delete;
+    PointerRepository(PointerRepository&&) = delete;
+    PointerRepository& operator=(const PointerRepository&) = delete;
+    PointerRepository& operator=(PointerRepository&&) = delete;
 
     /// @brief registers the start pointer of the segment in another application with a specific id
-    /// @param[in] id identifies the segment
+    /// @param[in] id identifies the segment which the pointer should be added to
```
MatthiasKillat, line 71:
```suggestion
    /// @param[in] id identifies the segment that the pointer should be added to
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -45,33 +46,40 @@ class PointerRepository
...
     /// @param[in] ptr is the start pointer of the segment
     /// @param[in] size is the size of the segment
     /// @return true if the registration was successful, otherwise false
-    bool registerPtr(id_t id, ptr_t ptr, uint64_t size) noexcept;
+    bool registerPtrWithId(id_t id, ptr_t ptr, uint64_t size) noexcept;
```
MatthiasKillat, line 75: I prefer overloading to long names. This is how it was done in C in the past (and had to, because overloading did not exist).

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -45,33 +46,40 @@ class PointerRepository
...
 
     /// @brief registers the start pointer of a segment with a specific size
     /// @param[in] ptr is the start pointer of the segment
     /// @param[in] size is the size of the segment
-    /// @return the id that identifies the segment
-    id_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
+    /// @return the segment id wrapped in an cxx::optional to which the pointer was added, cxx::nullopt if pointer was
```
MatthiasKillat, line 80:
```suggestion
    /// @return the segment id to which the pointer was added wrapped in an cxx::optional, cxx::nullopt if pointer was
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -71,11 +72,11 @@ inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::registerPtr(const ptr_t pt
             {
                 m_maxRegistered = id;
             }
-            return id;
+            return cxx::make_optional<id_t>(id);
```
MatthiasKillat, line 75: I think it is not needed, we can just use 
```suggestion
            return id;
```
and the ctor will be invoked (as the compiler knows the return type). Unless it is explicit ... which is another reason I do not like explicit as this verbosity does not help here. While reading the impl code the only relevant information is that some id is returned (and not a null-value). We only care about the wrapper type at call site.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.inl
```diff
@@ -132,30 +133,8 @@ inline id_t PointerRepository<id_t, ptr_t, CAPACITY>::searchId(ptr_t ptr) const
         }
     }
     /// @note implicitly interpret the pointer as a regular pointer if not found
```
MatthiasKillat, line 135:
```suggestion
    /// @note implicitly interpret the pointer as a regular pointer if not found
```
```suggestion
    /// @note treat the pointer as a regular pointer if not found
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
```diff
@@ -64,11 +64,13 @@ class RelativePointer : public BaseRelativePointer
...
 
-    /// @brief read-only access to the underlying object
+    /// @brief read-only access to the underlying object. If the RelativePointer does not point to anything the
+    /// application might terminate.
     /// @return a pointer to the underlying object
     T* operator->() const noexcept;
 
-    /// @brief access the underlying object
+    /// @brief access the underlying object. If the RelativePointer does not point to anything the
+    /// application might terminate.
```
MatthiasKillat, line 73: I think id **does terminate** here currently for a logical nullptr, so we should state this. Note that I think it is not a good idea to do so, because just getting the pointer (as `nullptr`)) is not the problem. The problem is dereferencing it later, as with regular pointers so we should remove the check.

iceoryx_posh/source/roudi/memory/memory_provider.cpp
```diff
@@ -86,14 +86,19 @@ cxx::expected<MemoryProviderError> MemoryProvider::create() noexcept
 
     m_memory = memoryResult.value();
     m_size = totalSize;
-    m_segmentId = rp::BaseRelativePointer::registerPtr(m_memory, m_size);
+    auto maybeSegmentId = rp::BaseRelativePointer::registerPtr(m_memory, m_size);
+
+    if (maybeSegmentId.has_value())
```
MatthiasKillat, line 91: If it does not work it can be considered fatal and we would actually not be able to continue. This cannot really happen for a reasonable number of segments though.

We should generate a fatal error in this case, or at least have a placeholder terminate and a todo for it (will add some `IOX_PANIC` or similar once available) .

iceoryx_posh/include/iceoryx_posh/internal/mepoo/mepoo_segment.inl
MatthiasKillat: @mossmaurice I would at least add a terminate, a todo and later call `IOX_PANIC` (for a fatal error). But we can also have an error code for it (INSUFFICIENT_SEGMENT_IDS or something like this).

iceoryx_hoofs/source/concurrent/loffli.cpp
```diff
@@ -55,7 +55,7 @@ bool LoFFLi::pop(Index_t& index) noexcept
     do
     {
         // we are empty if next points to an element with index of Size
-        if (oldHead.indexToNextFreeIndex >= m_size)
+        if (oldHead.indexToNextFreeIndex >= m_size || !m_nextFreeIndex)
```
MatthiasKillat, line 58: Why is this needed?

Note that the whole index terminology in loffli is kind of inconsistent or broken which makes it hard to follow (not well-defined, as there are at least two levels of indices from what I remember).

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -362,6 +363,25 @@
     #include "iceoryx_hoofs/cxx/algorithm.hpp"
     constexpr uint32_t MAX_VAL = algorithm::maxVal(3, 1890, 57);
     constexpr uint32_t MIN_VAL = algorithm::minVal(3, 1890, 57);
+
+20. `ReleativePointer::registerPtr` returns `cxx::optional`
+
+    ```cpp
+    // before
+    uint64_t id = RelativePointer::register(myPtr, 42);
```
MatthiasKillat, line 371:
```suggestion
    uint64_t id = RelativePointer::register(startAddress, numBytes);
```

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -362,6 +363,25 @@
...
+
+    if(id == INVALID_ID)
+    {
+        // Early exit
+    }
+
+    // after
+    auto maybeId = RelativePointer::register(myPtr, 42);
```
MatthiasKillat, line 379:
```suggestion
    auto maybeId = RelativePointer::register(startAddress, numBytes);
```

We register a segment beginning at `startAddress` with size `numBytes` and get some id now associated with this segment (if successful).

MatthiasKillat (changes requested):

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
mossmaurice: @MatthiasKillat In general I agree with you, but in this case the parameters are rather similar so it's nicer if you type `.` in your IDE and then get two different methods offered.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
mossmaurice: @MatthiasKillat True, I moved the assert-style check to `operator->()` and adapted the wording.

iceoryx_hoofs/source/concurrent/loffli.cpp
mossmaurice: @MatthiasKillat By change, I ran into an issue in `LoFFLi` when `init()` was not called. This check prevents `pop()` to work on the data if the `RelativePointer` was not initialized.

FerdinandSpitzschnueffler (commented):
I wonder why you didn't need to adapt the relative ptr test files.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -45,33 +46,40 @@ class PointerRepository
...
+    PointerRepository& operator=(PointerRepository&&) = delete;
 
     /// @brief registers the start pointer of the segment in another application with a specific id
-    /// @param[in] id identifies the segment
+    /// @param[in] id identifies the segment that the pointer should be added to
     /// @param[in] ptr is the start pointer of the segment
     /// @param[in] size is the size of the segment
     /// @return true if the registration was successful, otherwise false
-    bool registerPtr(id_t id, ptr_t ptr, uint64_t size) noexcept;
+    bool registerPtrWithId(id_t id, ptr_t ptr, uint64_t size) noexcept;
```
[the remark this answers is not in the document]
MatthiasKillat: They can. But value parameters should not be and I am fairly sure static code analysis agrees, unless this is disabled.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
```diff
@@ -64,11 +64,12 @@ class RelativePointer : public BaseRelativePointer
...
     typename std::enable_if<!std::is_void<U>::value, const U&>::type operator*() const noexcept;
 
-    /// @brief read-only access to the underlying object
+    /// @brief read-only access to the underlying object. If the RelativePointer does not point to anything the
+    /// application terminates.
     /// @return a pointer to the underlying object
     T* operator->() const noexcept;
 
-    /// @brief access the underlying object
+    /// @brief access the underlying object.
```
MatthiasKillat, line 72: Actually 

```suggestion
    /// @brief Access the underlying object.
```

but this is inconsistent with the other documentation so we should not change it.

iceoryx_hoofs/source/concurrent/loffli.cpp
mossmaurice: As there is already a todo, I didn't add a further one.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/base_relative_pointer.hpp
```diff
@@ -124,7 +124,7 @@ class BaseRelativePointer
     /// @param[in] ptr starting address of the segment to be registered
     /// @param[in] size is the size of the segment
     /// @return id it was registered to
-    static id_underlying_t registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
+    static cxx::optional<id_underlying_t> registerPtr(const ptr_t ptr, uint64_t size = 0U) noexcept;
```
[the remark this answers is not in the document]
mossmaurice: @FerdinandSpitzschnueffler As `BaseRelativePointer` will be merged with `RelativePointer` I would prefer to change doxygen on #1603

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/pointer_repository.hpp
```diff
@@ -18,10 +18,11 @@
 #ifndef IOX_HOOFS_RELOCATABLE_POINTER_POINTER_REPOSITORY_HPP
 #define IOX_HOOFS_RELOCATABLE_POINTER_POINTER_REPOSITORY_HPP
 
+#include "iceoryx_hoofs/cxx/optional.hpp"
 #include "iceoryx_hoofs/cxx/vector.hpp"
-#include <iostream>
 
 #include <cassert>
+#include <iostream>
```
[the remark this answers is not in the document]
mossmaurice: @FerdinandSpitzschnueffler Good catch, thanks! Removed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/relocatable_pointer/relative_pointer.hpp
mossmaurice: @MatthiasKillat I capitalized all sentences in the beginning.

MatthiasKillat (dismissed):

iceoryx_hoofs/source/concurrent/loffli.cpp
FerdinandSpitzschnueffler: My inner chicken fears that we will forget to remove the unnecessary checks. Could you please add a hint to the todo in the header file? Something like "move init code to c'tor, remove unnecessary `!m_nextfreeIndex` check".

FerdinandSpitzschnueffler (approved):

MatthiasKillat (approved):

PR #1607, opened 2022-09-06
Iox #1394 fix axivion violation for optional hpp inl

saif-at-github:

    - [ x Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #1394 (partially)

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
@@ -167,10 +167,11 @@ class optional final : public FunctionalInterface<optional<T>, T, void>
     /// @return reference of type T to the underlying type
     T& operator*() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A13.5.3: Implemented to be as close to the STL interface as possible.
+    // In combination with the keyword explicit accidental casts can be excluded and the usage is well known in the
+    // C++ community.
     /// @brief Will return true if the optional contains a value, otherwise false.
     /// @return true if optional contains a value, otherwise false
-    // AXIVION Next Construct AutosarC++19_03-A13.5.3: designed implementation shall be as close
-    // as std::optional and it requires operator bool
     constexpr explicit operator bool() const noexcept;
 
     /// @brief Will return true if the optional contains a value, otherwise false.
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
@@ -80,12 +80,11 @@ class optional final : public FunctionalInterface<optional<T>, T, void>
     ///         terminates.
     optional() noexcept;
 
-    // AXIVION Next Construct AutosarC++19_03-A8.4.7 : reference parameter is efficient and idiomatic for copy ctors
     /// @brief Creates an optional which has no value. If you access such an
     ///         optional via .value() or the arrow operator the application
     ///         terminates.
     // NOLINTNEXTLINE(hicpp-explicit-conversions) for justification see doxygen
-    optional(const nullopt_t&) noexcept;
+    optional(const nullopt_t) noexcept;
 
     /// @brief Creates an optional by forwarding value to the constructor of
     ///         T. This optional has a value.
@@ -170,7 +169,8 @@ class optional final : public FunctionalInterface<optional<T>, T, void>
 
     /// @brief Will return true if the optional contains a value, otherwise false.
     /// @return true if optional contains a value, otherwise false
-    // AXIVION Next Construct AutosarC++19_03-A13.5.3: requires a user-defined conversion
+    // AXIVION Next Construct AutosarC++19_03-A13.5.3: designed implementation shall be as close
+    // as std::optional and it requires operator bool
     constexpr explicit operator bool() const noexcept;
 
     /// @brief Will return true if the optional contains a value, otherwise false.
@@ -253,14 +253,14 @@ optional<OptionalBaseType> make_optional(Targs&&... args) noexcept;
 
 /// @brief Compare two optionals for equality.
 /// @param[in] lhs cxx::optional
-/// @param[in] rhs value to which lhs should be compared to
+/// @param[in] rhs optional to which lhs should be compared to
 /// @return true if the contained values are equal or both have no value, otherwise false
 template <typename T>
 bool operator==(const optional<T>& lhs, const optional<T>& rhs) noexcept;
 
 /// @brief Compare two optionals for inequality.
 /// @param[in] lhs cxx::optional
-/// @param[in] rhs value to which lhs should be compared to
+/// @param[in] rhs optional to which lhs should be compared to
 /// @return true if the contained values are not equal, otherwise false
 template <typename T>
 bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
@@ -268,27 +268,27 @@ bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
 // AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
 /// @brief Comparison for equality with nullopt_t for easier unset optional comparison
 /// @param[in] lhs empty optional, cxx::nullopt_t
-/// @param[in] rhs value to which lhs should be compared to
-/// @return true if the optional is set, otherwise false
+/// @param[in] rhs optional to which lhs should be compared to
+/// @return true if the rhs is not set, otherwise false
 template <typename T>
 bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
 
 /// @brief Comparison for equality with nullopt_t for easier unset optional comparison
-/// @param[in] lhs value to which lhs should be compared to
+/// @param[in] lhs optional which should be compared to cxx::nullopt_t
 /// @param[in] rhs empty optional
-/// @return true if the optional is set, otherwise false
+/// @return true if the lhs is not set, otherwise false
 template <typename T>
 bool operator==(const optional<T>& lhs, const nullopt_t) noexcept;
 
 /// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
 /// @param[in] lhs empty optional, cxx::nullopt_t
-/// @param[in] rhs value to which lhs should be compared to
+/// @param[in] rhs optional to which lhs should be compared to
 /// @return true if the optional is set, otherwise false
 template <typename T>
 bool operator!=(const nullopt_t, const optional<T>& rhs) noexcept;
 
 /// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
-/// @param[in] lhs value to which lhs should be compared to
+/// @param[in] lhs optional which should be compared to cxx::nullopt_t
 /// @param[in] rhs empty optional
 /// @return true if the optional is set, otherwise false
 template <typename T>
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
@@ -33,12 +33,17 @@ namespace cxx
 struct nullopt_t
 {
 };
-constexpr nullopt_t nullopt = nullopt_t();
+// AXIVION Next Construct AutosarC++19_03-M17.0.2 : nullopt is defined within iox::cxx namespace which prevents easy
+// misuse
+constexpr nullopt_t nullopt{nullopt_t()};
 
 /// @brief helper struct which is used to call the in-place-construction constructor
 struct in_place_t
 {
 };
+
+// AXIVION Next Construct AutosarC++19_03-M17.0.2 : in_place is defined within iox::cxx namespace which prevents easy
+// misuse
 constexpr in_place_t in_place{};
 
 /// @brief Optional implementation from the C++17 standard with C++11. The
@@ -65,7 +70,7 @@ constexpr in_place_t in_place{};
 ///     }
 /// @endcode
 template <typename T>
-class optional : public FunctionalInterface<optional<T>, T, void>
+class optional final : public FunctionalInterface<optional<T>, T, void>
 {
   public:
     using type = T;
@@ -75,11 +80,12 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     ///         terminates.
     optional() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A8.4.7 : reference parameter is efficient and idiomatic for copy ctors
     /// @brief Creates an optional which has no value. If you access such an
     ///         optional via .value() or the arrow operator the application
     ///         terminates.
     // NOLINTNEXTLINE(hicpp-explicit-conversions) for justification see doxygen
-    optional(const nullopt_t& noValue) noexcept;
+    optional(const nullopt_t&) noexcept;
 
     /// @brief Creates an optional by forwarding value to the constructor of
     ///         T. This optional has a value.
@@ -128,26 +134,6 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     /// @return reference to the current optional
     optional& operator=(optional&& rhs) noexcept;
 
-    /// @brief If the optionals have values it compares these values by using
-    ///         their comparison operator.
-    /// @param[in] rhs value to which this optional should be compared to
-    /// @return true if the contained values are equal, otherwise false
-    constexpr bool operator==(const optional<T>& rhs) const noexcept;
-
-    /// @brief Comparison with nullopt_t for easier unset optional comparison
-    /// @return true if the optional is unset, otherwise false
-    constexpr bool operator==(const nullopt_t& rhs) const noexcept;
-
-    /// @brief If the optionals have values it compares these values by using
-    ///         their comparison operator.
-    /// @param[in] rhs value to which this optional should be compared to
-    /// @return true if the contained values are not equal, otherwise false
-    constexpr bool operator!=(const optional<T>& rhs) const noexcept;
-
-    /// @brief comparison with nullopt_t for easier unset optional comparison
-    /// @return true if the optional is set, otherwise false
-    constexpr bool operator!=(const nullopt_t& rhs) const noexcept;
-
     /// @brief Direct assignment of the underlying value. If the optional has no
     ///         value then a new T is constructed by forwarding the assignment to
     ///         T's constructor.
@@ -156,7 +142,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     /// @return reference to the current optional
     template <typename U = T>
     // NOLINTNEXTLINE(cppcoreguidelines-c-copy-assignment-signature) return type is optional&
-    typename std::enable_if<!std::is_same<U, optional<T>&>::value, optional>::type& operator=(U&& value) noexcept;
+    typename std::enable_if<!std::is_same<U, optional<T>&>::value, optional>::type& operator=(U&& newValue) noexcept;
 
     /// @brief Returns a pointer to the underlying value. If the optional has no
     ///         value the application terminates. You need to verify that the
@@ -184,6 +170,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
 
     /// @brief Will return true if the optional contains a value, otherwise false.
     /// @return true if optional contains a value, otherwise false
+    // AXIVION Next Construct AutosarC++19_03-A13.5.3: requires a user-defined conversion
     constexpr explicit operator bool() const noexcept;
 
     /// @brief Will return true if the optional contains a value, otherwise false.
@@ -242,7 +229,10 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     //     initHandle(&handle);
     //   }
     bool m_hasValue{false};
-    // safe access is guaranteed since the array is wrapped inside the optional
+
+  private:
+    // AXIVION Next Construct AutosarC++19_03-A18.1.1 : safe access is guaranteed since the array
+    // is wrapped inside the optional
     // NOLINTNEXTLINE(hicpp-avoid-c-arrays, cppcoreguidelines-avoid-c-arrays)
     alignas(T) byte_t m_data[sizeof(T)];
 
@@ -252,6 +242,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     void destruct_value() noexcept;
 };
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : make_optional is defined within iox::cxx which prevents easy misuse
 /// @brief Creates an optional which contains a value by forwarding the arguments
 ///         to the constructor of T.
 /// @param[in] args arguments which will be perfectly forwarded to the constructor
@@ -259,6 +250,50 @@ class optional : public FunctionalInterface<optional<T>, T, void>
 /// @return optional which contains T constructed with args
 template <typename OptionalBaseType, typename... Targs>
 optional<OptionalBaseType> make_optional(Targs&&... args) noexcept;
+
+/// @brief Compare two optionals for equality.
+/// @param[in] lhs cxx::optional
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the contained values are equal or both have no value, otherwise false
+template <typename T>
+bool operator==(const optional<T>& lhs, const optional<T>& rhs) noexcept;
+
+/// @brief Compare two optionals for inequality.
+/// @param[in] lhs cxx::optional
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the contained values are not equal, otherwise false
+template <typename T>
+bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
+
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
+
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs value to which this optional should be compared to
+/// @param[in] rhs empty optional, cxx::nullopt_t
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator==(const optional<T>& lhs, const nullopt_t) noexcept;
+
+/// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator!=(const nullopt_t, const optional<T>& rhs) noexcept;
+
+/// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs value to which this optional should be compared to
+/// @param[in] rhs empty optional, cxx::nullopt_t
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator!=(const optional<T>& lhs, const nullopt_t) noexcept;
+// AXIVION ENABLE STYLE AutosarC++19_03-A13.5.5
 } // namespace cxx
 } // namespace iox
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
@@ -253,14 +253,14 @@ optional<OptionalBaseType> make_optional(Targs&&... args) noexcept;
 
 /// @brief Compare two optionals for equality.
 /// @param[in] lhs cxx::optional
-/// @param[in] rhs value to which this optional should be compared to
+/// @param[in] rhs value to which lhs should be compared to
 /// @return true if the contained values are equal or both have no value, otherwise false
 template <typename T>
 bool operator==(const optional<T>& lhs, const optional<T>& rhs) noexcept;
 
 /// @brief Compare two optionals for inequality.
 /// @param[in] lhs cxx::optional
-/// @param[in] rhs value to which this optional should be compared to
+/// @param[in] rhs value to which lhs should be compared to
 /// @return true if the contained values are not equal, otherwise false
 template <typename T>
 bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
@@ -268,28 +268,28 @@ bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
 // AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
 /// @brief Comparison for equality with nullopt_t for easier unset optional comparison
 /// @param[in] lhs empty optional, cxx::nullopt_t
-/// @param[in] rhs value to which this optional should be compared to
+/// @param[in] rhs value to which lhs should be compared to
 /// @return true if the optional is set, otherwise false
 template <typename T>
 bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
 
 /// @brief Comparison for equality with nullopt_t for easier unset optional comparison
-/// @param[in] lhs value to which this optional should be compared to
-/// @param[in] rhs empty optional, cxx::nullopt_t
+/// @param[in] lhs value to which lhs should be compared to
+/// @param[in] rhs empty optional
 /// @return true if the optional is set, otherwise false
 template <typename T>
 bool operator==(const optional<T>& lhs, const nullopt_t) noexcept;
 
 /// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
 /// @param[in] lhs empty optional, cxx::nullopt_t
-/// @param[in] rhs value to which this optional should be compared to
+/// @param[in] rhs value to which lhs should be compared to
 /// @return true if the optional is set, otherwise false
 template <typename T>
 bool operator!=(const nullopt_t, const optional<T>& rhs) noexcept;
 
 /// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
-/// @param[in] lhs value to which this optional should be compared to
-/// @param[in] rhs empty optional, cxx::nullopt_t
+/// @param[in] lhs value to which lhs should be compared to
+/// @param[in] rhs empty optional
 /// @return true if the optional is set, otherwise false
 template <typename T>
 bool operator!=(const optional<T>& lhs, const nullopt_t) noexcept;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
@@ -27,10 +27,9 @@ namespace cxx
 // AXIVION DISABLE STYLE AutosarC++19_03-A12.6.1 : m_data is not initialized here, since this is a
 // constructor for an optional with no value; an access of the value would lead to the
 // application's termination
-// AXIVION Next Construct AutosarC++19_03-A8.4.7 : reference type input parameter is fine here
 // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 template <typename T>
-inline optional<T>::optional(const nullopt_t&) noexcept
+inline optional<T>::optional(const nullopt_t) noexcept
 {
 }
 
@@ -93,7 +92,10 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
 {
     construct_value(std::forward<Targs>(args)...);
 }
+// AXIVION ENABLE STYLE AutosarC++19_03-A12.6.1
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : False positive. Overloading excluded via std::enable_if in
+// typename std::enable_if<!std::is_same<U, optional&>::value, optional>::type& operator=(U&& value) noexcept.
 template <typename T>
 inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
 {
@@ -113,12 +115,14 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
         }
         else
         {
-            // do nothing
+            // do nothing since this and rhs contain no values
         }
     }
     return *this;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : False positive. Overloading excluded via std::enable_if in typename
+// std::enable_if<!std::is_same<U, optional&>::value, optional>::type& operator=(U&& value) noexcept.
 template <typename T>
 inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
 {
@@ -138,7 +142,7 @@ inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
         }
         else
         {
-            // do nothing
+            // do nothing since this and rhs contain no values
         }
         if (rhs.m_hasValue)
         {
@@ -266,7 +270,7 @@ template <typename T>
 inline const T&& optional<T>::value() const&& noexcept
 {
     // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast) const_cast to avoid code duplication
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
     return std::move(*const_cast<optional<T>*>(this)->value());
 }
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
@@ -94,7 +94,6 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
     construct_value(std::forward<Targs>(args)...);
 }
 
-// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
 template <typename T>
 inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
 {
@@ -120,7 +119,6 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
     return *this;
 }
 
-// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
 template <typename T>
 inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
 {
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
@@ -24,11 +24,13 @@ namespace iox
 {
 namespace cxx
 {
-// m_data is not initialized since this is a constructor for an optional with no value; an access of the value would
-// lead to the application's termination
+// AXIVION DISABLE STYLE AutosarC++19_03-A12.6.1 : m_data is not initialized here, since this is a
+// constructor for an optional with no value; an access of the value would lead to the
+// application's termination
+// AXIVION Next Construct AutosarC++19_03-A8.4.7 : reference type input parameter is fine here
 // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 template <typename T>
-inline optional<T>::optional(const nullopt_t& noVal IOX_MAYBE_UNUSED) noexcept
+inline optional<T>::optional(const nullopt_t&) noexcept
 {
 }
 
@@ -46,6 +48,8 @@ inline optional<T>::optional() noexcept
 template <typename T>
 inline optional<T>::optional(T&& value) noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2 : Perfect forwarding is intended here and
+    // std::forward is the idiomatic way for this
     construct_value(std::forward<T>(value));
 }
 
@@ -90,6 +94,7 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
     construct_value(std::forward<Targs>(args)...);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
 template <typename T>
 inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
 {
@@ -107,10 +112,15 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
         {
             construct_value(rhs.value());
         }
+        else
+        {
+            // do nothing
+        }
     }
     return *this;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
 template <typename T>
 inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
 {
@@ -127,6 +137,10 @@ inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
         else if (rhs.m_hasValue && !m_hasValue)
         {
             construct_value(std::move(rhs.value()));
+        }
+        else
+        {
+            // do nothing
         }
         if (rhs.m_hasValue)
         {
@@ -153,49 +167,13 @@ optional<T>::operator=(U&& newValue) noexcept
 {
     if (m_hasValue)
     {
-/// @todo iox-#1694 broken msvc compiler, see:
-///  https://developercommunity.visualstudio.com/content/problem/858688/stdforward-none-of-these-2-overloads-could-convert.html
-/// remove this as soon as it is fixed;
-#ifdef _WIN32
-        value() = newValue;
-#else
         value() = std::forward<T>(newValue);
-#endif
     }
     else
     {
-/// @todo iox-#1694 again broken msvc compiler
-#ifdef _WIN32
-        construct_value(newValue);
-#else
         construct_value(std::forward<T>(newValue));
-#endif
     }
     return *this;
-}
-
-template <typename T>
-constexpr inline bool optional<T>::operator==(const optional<T>& rhs) const noexcept
-{
-    return (!m_hasValue && !rhs.m_hasValue) || ((m_hasValue && rhs.m_hasValue) && (value() == rhs.value()));
-}
-
-template <typename T>
-constexpr inline bool optional<T>::operator==(const nullopt_t& rhs IOX_MAYBE_UNUSED) const noexcept
-{
-    return !m_hasValue;
-}
-
-template <typename T>
-constexpr inline bool optional<T>::operator!=(const optional<T>& rhs) const noexcept
-{
-    return !(*this == rhs);
-}
-
-template <typename T>
-constexpr inline bool optional<T>::operator!=(const nullopt_t& rhs) const noexcept
-{
-    return !(*this == rhs);
 }
 
 template <typename T>
@@ -260,13 +238,18 @@ template <typename T>
 inline T& optional<T>::value() & noexcept
 {
     Expects(has_value());
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : The optional has the type T defined
+    // during compile time and the type is unchangeable during the lifetime of the object.
+    // All accesses to the underlying data is done via the same static type and therefore the
+    // casts are always valid
     return *static_cast<T*>(static_cast<void*>(&m_data));
 }
 
 template <typename T>
 inline const T& optional<T>::value() const& noexcept
 {
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast) const_cast to avoid code duplication
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
     return const_cast<optional<T>*>(this)->value();
 }
 
@@ -274,12 +257,17 @@ template <typename T>
 inline T&& optional<T>::value() && noexcept
 {
     Expects(has_value());
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : The optional has the type T defined
+    // during compile time and the type is unchangeable during the lifetime of the object.
+    // All accesses to the underlying data is done via the same static type and therefore the
+    // casts are always valid
     return std::move(*static_cast<T*>(static_cast<void*>(&m_data)));
 }
 
 template <typename T>
 inline const T&& optional<T>::value() const&& noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast) const_cast to avoid code duplication
     return std::move(*const_cast<optional<T>*>(this)->value());
 }
@@ -288,7 +276,7 @@ template <typename T>
 template <typename... Targs>
 inline void optional<T>::construct_value(Targs&&... args) noexcept
 {
-    new (static_cast<T*>(static_cast<void*>(&m_data))) T(std::forward<Targs>(args)...);
+    new (&m_data) T(std::forward<Targs>(args)...);
     m_hasValue = true;
 }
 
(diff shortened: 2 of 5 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -107,12 +112,16 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
         {
             construct_value(rhs.value());
         }
+        else
+        {
+            // do nothing
+        }
```
elBoberido, line 118: @saif-at-github this does not look nice. Which rule does cause the addition of this branch?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -80,7 +83,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     ///         optional via .value() or the arrow operator the application
     ///         terminates.
     // NOLINTNEXTLINE(hicpp-explicit-conversions) for justification see doxygen
-    optional(const nullopt_t& noValue) noexcept;
+    optional(const nullopt_t noVal) noexcept;
```
elfenpiff, line 86: I think the clean C++ way is to remove the variable name. With this the compiler knows that this is not used.
Removing the reference is the wrong approach.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -103,7 +106,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     explicit optional(in_place_t, Targs&&... args) noexcept;
 
     /// @brief The destructor will call the destructor of T if a value is set.
-    ~optional() noexcept;
+    virtual ~optional() noexcept;
```
elfenpiff, line 109: Please remove this! The optional is used in the shared memory and this would introduce a bug and weird segfaults.

The problem is that the `virtual` keyword introduces a vtable which is in the end a list of function pointers. But the function pointers are only valid in one process space. Another process may has the same function stored in a different place and then we either have undefined behavior or a crash.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -120,14 +123,14 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     ///         value a new value is constructed with the copy constructor.
     /// @param[in] rhs source of the copy
     /// @return reference to the current optional
-    optional& operator=(const optional& rhs) noexcept;
+    optional& operator=(const optional& rhs) & noexcept;
```
elfenpiff, line 126: Please remove this! I know this is an autosar rule but this would actually make our C++ code non compliant to the c++ standard.

Take a look at: https://isocpp.org/files/papers/N4860.pdf Section 6.2 part 3.

The move and copy constructors are defined like:
```cxx
class C {
  C(const C& rhs); // copy ctor
  C(C&& rhs); // move ctor
  C& operator=(const C& rhs); // copy assignment
  C& operator=(C&& rhs); // move assignment
}
```

There is also no possible bug which can occur when the lvalue marker is missing.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -120,14 +123,14 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
 
     /// @brief Moves an optional. If the optional has a value then the move
     ///         assignment of that value is called. If the optional has no
     ///         value a new value is constructed with the move constructor.
     /// @param[in] rhs source of the move
     /// @return reference to the current optional
-    optional& operator=(optional&& rhs) noexcept;
+    optional& operator=(optional&& rhs) & noexcept;
```
elfenpiff, line 133: Remove this as well. This is not compliant to the C++ standard.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -137,7 +140,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
 
     /// @brief Comparison with nullopt_t for easier unset optional comparison
     /// @return true if the optional is unset, otherwise false
-    constexpr bool operator==(const nullopt_t& rhs) const noexcept;
+    constexpr bool operator==(const nullopt_t rhs) const noexcept;
```
elfenpiff, line 143: Please remove the argument completely since it is not used at all.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -157,7 +160,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     /// @return reference to the current optional
     template <typename U = T>
     // NOLINTNEXTLINE(cppcoreguidelines-c-copy-assignment-signature) return type is optional&
-    typename std::enable_if<!std::is_same<U, optional<T>&>::value, optional>::type& operator=(U&& value) noexcept;
+    typename std::enable_if<!std::is_same<U, optional<T>&>::value, optional>::type& operator=(U&& newValue) & noexcept;
```
elfenpiff, line 163: Please remove it since it is not compliant to the C++ standard.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -259,7 +265,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
 ///             of T
 /// @return optional which contains T constructed with args
 template <typename OptionalBaseType, typename... Targs>
-optional<OptionalBaseType> make_optional(Targs&&... args) noexcept;
+auto make_optional(Targs&&... args) noexcept -> optional<OptionalBaseType>;
```
elfenpiff, line 268: Please revert this. Here we have to write more and destroy the normal reading flow to follow a rule.

The rational behind the rule is that this prevents `typename` usage. But here is no `typename` and it also does not prevent the `typename` usage. The only thing which happens here is that the code becomes more complex and the normal reading flow of `returntype`, `functionname()`  and `qualifiers` is destroyed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -24,11 +24,15 @@ namespace iox
...
+// AXIVION DISABLE STYLE AutosarC++19_03-A12.6.1 : field intentionally not initialized
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.3.1 : that is constrained (via SFINAE) to not match
+// any other overloads
+// AXIVION Next Construct AutosarC++19_03-A0.1.4 : TODO
 // m_data is not initialized since this is a constructor for an optional with no value; an access of the value would
 // lead to the application's termination
 // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 template <typename T>
-inline optional<T>::optional(const nullopt_t& noVal IOX_MAYBE_UNUSED) noexcept
+inline optional<T>::optional(const nullopt_t noVal IOX_MAYBE_UNUSED) noexcept
```
elfenpiff, line 35: When we remove the variable name we can also get rid of `IOX_MAYBE_UNUSED` and the code would become more readable.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -46,6 +50,7 @@ inline optional<T>::optional() noexcept
 template <typename T>
 inline optional<T>::optional(T&& value) noexcept
 {
+    // AXIVION Next Line AutosarC++19_03-A18.9.2 : No need to explicitly compare err with bool
     construct_value(std::forward<T>(value));
```
elfenpiff, line 54: According to this document: https://www.autosar.org/fileadmin/user_upload/standards/adaptive/18-10/AUTOSAR_RS_CPP14Guidelines.pdf

This piece of code is compliant with the autosar rules.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
elfenpiff: I agree, this does not improve readability or decreases complexity.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -149,13 +162,15 @@ inline optional<T>::~optional() noexcept
...
-optional<T>::operator=(U&& newValue) noexcept
+optional<T>::operator=(U&& newValue) & noexcept
 {
     if (m_hasValue)
     {
 /// @todo broken msvc compiler, see:
 ///  https://developercommunity.visualstudio.com/content/problem/858688/stdforward-none-of-these-2-overloads-could-convert.html
 /// remove this as soon as it is fixed;
+// AXIVION DISABLE STYLE AutosarC++19_03-A16.0.1 : this macro is used to prevent a large amount
+// of boilerplate code
```
elfenpiff, line 173: Instead of suppressing the warning I suggest to remove the following three lines and check if the msvc compiler has fixed the issue in the meantime.
When the CI is green than off it goes.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -179,9 +194,9 @@ constexpr inline bool optional<T>::operator==(const optional<T>& rhs) const noex
 {
     return (!m_hasValue && !rhs.m_hasValue) || ((m_hasValue && rhs.m_hasValue) && (value() == rhs.value()));
 }
-
+// AXIVION Next Construct AutosarC++19_03-A0.1.4 : TODO
 template <typename T>
-constexpr inline bool optional<T>::operator==(const nullopt_t& rhs IOX_MAYBE_UNUSED) const noexcept
+constexpr inline bool optional<T>::operator==(const nullopt_t rhs IOX_MAYBE_UNUSED) const noexcept
```
elfenpiff, line 199: When we remove the variable name here we can get rid of `IOX_MAYBE_UNUSED` as well.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -193,7 +208,7 @@ constexpr inline bool optional<T>::operator!=(const optional<T>& rhs) const noex
 }
 
 template <typename T>
-constexpr inline bool optional<T>::operator!=(const nullopt_t& rhs) const noexcept
+constexpr inline bool optional<T>::operator!=(const nullopt_t rhs) const noexcept
```
elfenpiff, line 211: Please get rid of `rhs`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -260,12 +275,14 @@ template <typename T>
 inline T& optional<T>::value() & noexcept
 {
     Expects(has_value());
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : The caller ensures this is a valid cast
```
elfenpiff, line 278: The rationale is not so fitting.

The `optional` has the type `T` defined during compile time and the type is unchangeable during the lifetime of the object. All accesses to the underlying data is done via the same static type and therefore the casts are always valid.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -274,12 +291,14 @@ template <typename T>
 inline T&& optional<T>::value() && noexcept
 {
     Expects(has_value());
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : The caller ensures this is a valid cast
```
elfenpiff, line 294: See rationale in previous comment.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -288,6 +307,7 @@ template <typename T>
 template <typename... Targs>
 inline void optional<T>::construct_value(Targs&&... args) noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : Pointer conversion is safe here.
```
elfenpiff, line 310: Here I would use the same rational I provided in the comments above.

saif-at-github: Sure.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: Sure

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: Sure

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: the rule is `AutosarC++19_03-M6-4-2` - "All if ... else if constructs shall be terminated with an else clause."

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: Since the input argument is unused hence the reference was removed, but the suggestion looks more apt here. thanks

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: @elfenpiff My understanding is that a `vtable` is a class level entity, and the entries shall be same across the usages.
we can remove `virtual` only if the class is made a final. otherwise, we need to suppress `AutosarC++19_03-A12-4-2'

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: @elfenpiff let me know about this. thanks.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: Agree, it is false positive

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
elfenpiff: @saif-at-github Making the class `final` should not pose a problem for most `cxx` constructs. Can you please try it out.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: @elfenpiff Sure. Thanks

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
elBoberido: @elfenpiff then clang-tidy will complain. The agreed solution was to use `IOX_MAYBE_UNUSED`

elfenpiff: intro-import

There is always that one tool.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: @elfenpiff that works for sure. thanks.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -260,12 +271,17 @@ template <typename T>
...
+    // All accesses to the underlying data is done via the same static type and therefore the
+    // casts are always valid
     return *static_cast<T*>(static_cast<void*>(m_data));
 }
 
 template <typename T>
 inline const T& optional<T>::value() const& noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast) const_cast to avoid code duplication
```
mossmaurice, line 285:
```suggestion
    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,6 +261,9 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     void destruct_value() noexcept;
 };
 
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : make_optional is defined within iox::cxx and
+// is not overridden
+// overridden.
```
mossmaurice, line 266:
```suggestion
// AXIVION Next Construct AutosarC++19_03-M17.0.3 : make_optional is defined within iox::cxx and
// is not overridden
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -260,26 +271,36 @@ template <typename T>
 inline T& optional<T>::value() & noexcept
 {
     Expects(has_value());
-    return *static_cast<T*>(static_cast<void*>(&m_data));
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : The optional has the type T defined
+    // during compile time and the type is unchangeable during the lifetime of the object.
+    // All accesses to the underlying data is done via the same static type and therefore the
+    // casts are always valid
+    return *static_cast<T*>(static_cast<void*>(m_data));
```
MatthiasKillat, line 278: We need to use `&m_data` here (for portability reasons).

Maybe it is necessary to use a `reinterpret_cast` (then one should be used).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -288,7 +309,11 @@ template <typename T>
 template <typename... Targs>
 inline void optional<T>::construct_value(Targs&&... args) noexcept
 {
-    new (static_cast<T*>(static_cast<void*>(&m_data))) T(std::forward<Targs>(args)...);
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : The optional has the type T defined
+    // during compile time and the type is unchangeable during the lifetime of the object.
+    // All accesses to the underlying data is done via the same static type and therefore the
+    // casts are always valid
+    new (static_cast<T*>(static_cast<void*>(m_data))) T(std::forward<Targs>(args)...);
```
MatthiasKillat, line 316: `&` not needed. Also the cast s are not needed here, `new (&m_data) ... `. Note that the pointer is already aligned for T by design. Maybe this makes the comment unnecessary.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -128,13 +137,18 @@ inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
...
+            // do nothing
+        }
         if (rhs.m_hasValue)
         {
             rhs.destruct_value();
         }
     }
     return *this;
 }
+// AXIVION ENABLE STYLE AutosarC++19_03-A13.3.1
```
MatthiasKillat, line 151: Remove. We should not disable rules for entire regions.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -288,7 +276,7 @@ template <typename T>
 template <typename... Targs>
 inline void optional<T>::construct_value(Targs&&... args) noexcept
 {
-    new (static_cast<T*>(static_cast<void*>(&m_data))) T(std::forward<Targs>(args)...);
+    new (&m_data) T(std::forward<Targs>(args)...);
```
MatthiasKillat, line 279: Correct. New does not require a typed pointer and `m_data` is properly aligned by design.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -298,14 +286,56 @@ inline void optional<T>::destruct_value() noexcept
...
+    optional<OptionalBaseType> returnValue{nullopt_t()};
     returnValue.emplace(std::forward<Targs>(args)...);
     return returnValue;
 }
+
+template <typename T>
+bool operator==(const optional<T>& lhs, const optional<T>& rhs) noexcept
+{
+    auto bothNull = !lhs.has_value() && !rhs.has_value();
+    auto bothEqual = (lhs.has_value() && rhs.has_value()) && (*lhs == *rhs);
```
MatthiasKillat, line 303:
```suggestion
    const auto bothValuesEqual = (lhs.has_value() && rhs.has_value()) && (*lhs == *rhs);
```

Alternatively a one liner but the compiler should optimize it anyway and this is "intentional programming".
Now the question for optimization nerds (as myself) is whether it still is able to benefit from shortcircuiting. I would think so, with optimization, but one would need to fire up godbolt to confirm.

Note I am fine either way. I just think some analysis will complain if the helper expressions are not `const`. I consider this unnecessary for such a simple functions but the tools do not.

TLDR: either directly return the expanded expression or add const as suggested.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -89,7 +92,8 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
 {
     construct_value(std::forward<Targs>(args)...);
 }
-
+// AXIVION ENABLE STYLE AutosarC++19_03-A12.6.1
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: This should be fine. I think Axivion refers to 
```
template <typename U = T>
typename std::enable_if<!std::is_same<U, optional<T>&>::value, optional>::type& operator=(U&& value) noexcept;
```
which we need to directly assign the underlying value.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -298,14 +286,56 @@ inline void optional<T>::destruct_value() noexcept
...
-    optional<OptionalBaseType> returnValue = nullopt_t();
+    optional<OptionalBaseType> returnValue{nullopt_t()};
     returnValue.emplace(std::forward<Targs>(args)...);
     return returnValue;
 }
+
+template <typename T>
+bool operator==(const optional<T>& lhs, const optional<T>& rhs) noexcept
+{
+    auto bothNull = !lhs.has_value() && !rhs.has_value();
```
MatthiasKillat, line 302:
```suggestion
    const auto bothNull = !lhs.has_value() && !rhs.has_value();
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -298,14 +286,56 @@ inline void optional<T>::destruct_value() noexcept
     value().~T();
     m_hasValue = false;
 }
-
+// AXIVION Next Construct AutosarC++19_03-M17.0.3 : make_optional is defined within iox::cxx and
+// is not overridden
```
FerdinandSpitzschnueffler, line 290:
```suggestion
// AXIVION Next Construct AutosarC++19_03-M17.0.3 : make_optional is defined within iox::cxx which prevents easy misuse
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -298,14 +286,56 @@ inline void optional<T>::destruct_value() noexcept
...
+    return bothNull || bothEqual;
+}
+
+template <typename T>
+bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept
+{
+    return !(lhs == rhs);
+}
+
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t for easier unset
```
MatthiasKillat, line 313:
```suggestion
// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
```

This is also a questionable rule. If one needs comparison with additional types as here, it should be allowed without having to write a comment... 

@FerdinandSpitzschnueffler @saif-at-github 

Probably OK here to disable it for the whole (short) section. I would favor to disable it globally, but we should not delay the PR with discussion about this.

iceoryx_hoofs/test/moduletests/test_cxx_stack.cpp
```diff
@@ -250,8 +250,8 @@ TEST_F(stack_test, CopyConstructorWorksAndCallsTestClassCopyConstructor)
 
     cxx::stack<TestClass, STACK_SIZE> testStack(m_sut);
     EXPECT_THAT(TestClass::copyCTor, Eq(1));
-    EXPECT_THAT(testStack.size(), Eq(1));
-    EXPECT_THAT(testStack.pop(), Eq(TestClass(ELEMENT, ELEMENT, ELEMENT)));
+    ASSERT_THAT(testStack.size(), Eq(1));
```
MatthiasKillat, line 253: Note that here I am willing to risk a test crash for the sake of brevity (avoid a result variable). The size might be wrong or there could be a bug, but the same could be said for an additional `hasValue` check. 

If there is a serious bug a crash can never be avoided (nor can a silent failure ...).

Nothing to do here.

MatthiasKillat (changes requested):
See comments. The main point is that we should now test `operator==` and `operator!=` for the case where the first argument is of type `nullopt_t`.

Mainly tests to compare nullopt against nullopt and against some non-null, so not much is needed. 

Something like `CompareWithEqualNullopt` in `test_cxx_optional.cpp`.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,44 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+/// @brief If the optionals have values it compares these values by using
+///         their comparison operator.
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the contained values are not equal, otherwise false
+template <typename T>
+bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
+
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t for easier unset
+// optional comparison
+template <typename T>
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I would keep the operator as well.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -24,11 +24,13 @@ namespace iox
...
-// m_data is not initialized since this is a constructor for an optional with no value; an access of the value would
-// lead to the application's termination
+// AXIVION DISABLE STYLE AutosarC++19_03-A12.6.1 : m_data is not initialized here, since this is a
+// constructor for an optional with no value; an access of the value would lead to the
+// application's termination
+// AXIVION Next Construct AutosarC++19_03-A8.4.7 : reference type input parameter is fine here
 // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 template <typename T>
-inline optional<T>::optional(const nullopt_t& noVal IOX_MAYBE_UNUSED) noexcept
+inline optional<T>::optional(const nullopt_t&) noexcept
```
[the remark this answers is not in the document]
MatthiasKillat: I guess it is optimized away regardless if not used in the implementation. Essentially it acts as a tag type. From this point of view, there should be no reference. nullopt_t should have size 1 BTW and fit in a register. But there is no need to pass it anyway, it is just needed for overloading.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -76,11 +79,12 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     ///         terminates.
     optional() noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-A8.4.7 : reference type input parameter is fine here
     /// @brief Creates an optional which has no value. If you access such an
     ///         optional via .value() or the arrow operator the application
     ///         terminates.
     // NOLINTNEXTLINE(hicpp-explicit-conversions) for justification see doxygen
-    optional(const nullopt_t& noValue) noexcept;
+    optional(const nullopt_t&) noexcept;
```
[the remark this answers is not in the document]
saif-at-github: Yes it is. CI is passing.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -242,8 +227,13 @@ class optional : public FunctionalInterface<optional<T>, T, void>
     //     // mistakes can be introduced quickly in low level c abstraction classes
     //     initHandle(&handle);
     //   }
+    // AXIVION Next Construct AutosarC++19_03-M0.1.3 : false positive,it's not a unused variable
```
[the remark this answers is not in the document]
saif-at-github: Am not sure, it shows in local machine, but not in dashboard :(, anyways am removing it for now.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -242,8 +227,13 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
     bool m_hasValue{false};
-    // safe access is guaranteed since the array is wrapped inside the optional
+
+  private:
+    // AXIVION Next Construct AutosarC++19_03-M0.1.3 : false positive,it's not a unused variable
```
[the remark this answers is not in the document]
saif-at-github: sure

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,44 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
 ///         to the constructor of T.
 /// @param[in] args arguments which will be perfectly forwarded to the constructor
 ///             of T
 /// @return optional which contains T constructed with args
 template <typename OptionalBaseType, typename... Targs>
 optional<OptionalBaseType> make_optional(Targs&&... args) noexcept;
+
+/// @brief If the optionals have values it compares these values by using
+///         their comparison operator.
+/// @param[in] rhs value to which this optional should be compared to
```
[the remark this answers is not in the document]
saif-at-github: sure

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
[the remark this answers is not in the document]
saif-at-github: sure

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,44 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+
+template <typename T>
+bool operator==(const optional<T>& lhs, const nullopt_t) noexcept;
+
+template <typename T>
+bool operator!=(const nullopt_t, const optional<T>& rhs) noexcept;
+
+template <typename T>
+bool operator!=(const optional<T>& lhs, const nullopt_t) noexcept;
+// AXIVION ENABLE STYLE AutosarC++19_03-A13.5.5
```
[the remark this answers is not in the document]
saif-at-github: It is used, not for the whole file. it is only for all the comparison operators with nullopt_t argument

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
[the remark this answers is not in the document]
saif-at-github: So, no change here.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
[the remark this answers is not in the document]
saif-at-github: Sure we can use `Axivion Next Construct` here, it is only for two instances.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: Sure.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: Thanks.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: sure

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: sure

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: this is a good suggestion. thanks.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: This should be fine for now. Thanks.

iceoryx_hoofs/test/moduletests/test_cxx_stack.cpp
saif-at-github: Noted.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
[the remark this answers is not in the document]
saif-at-github: updated

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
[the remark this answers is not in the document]
MatthiasKillat: Fine for me.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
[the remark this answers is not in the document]
MatthiasKillat: Fine with me to leave as is, but the unfortunate part is that we require the comment above.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -90,6 +94,7 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
     construct_value(std::forward<Targs>(args)...);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
```
MatthiasKillat, line 97: I think this is a false positive. Not sure SFINAE is in play here, if there is no overload resolution set where ill-formed overloads are eliminated. I cannot see an overload with a forwarding reference (this would require `optional<T>` to be a template parameter).

@FerdinandSpitzschnueffler Maybe I am wrong here. Could you please double check?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -107,10 +112,15 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
...
         }
+        else
+        {
+            // do nothing
+        }
     }
     return *this;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's safe and constrained by SFINAE
```
MatthiasKillat, line 123: Same as above, I also think it is a false positive. If so, the justification should be adapted.

MatthiasKillat (changes requested):
We should still test the new operator== and operator!= overloads where the first argument has type nullopt_t.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator==(const optional<T>& lhs, const nullopt_t) noexcept;
+
+/// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator!=(const nullopt_t, const optional<T>& rhs) noexcept;
```
MatthiasKillat, line 289: Needs to be tested in `test_cxx_optional.cpp`.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+/// @brief Compare two optionals for equality.
+/// @param[in] lhs cxx::optional
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the contained values are equal or both have no value, otherwise false
+template <typename T>
+bool operator==(const optional<T>& lhs, const optional<T>& rhs) noexcept;
+
+/// @brief Compare two optionals for inequality.
+/// @param[in] lhs cxx::optional
+/// @param[in] rhs value to which this optional should be compared to
```
MatthiasKillat, line 264:
```suggestion
/// @param[in] rhs value to which lhs should be compared to
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs value to which this optional should be compared to
+/// @param[in] rhs empty optional, cxx::nullopt_t
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator==(const optional<T>& lhs, const nullopt_t) noexcept;
+
+/// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
```
MatthiasKillat, line 285:
```suggestion
/// @param[in] lhs empty optional
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
+
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs value to which this optional should be compared to
+/// @param[in] rhs empty optional, cxx::nullopt_t
```
MatthiasKillat, line 279:
```suggestion
/// @param[in] rhs empty optional
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+template <typename T>
+bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
+
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs value to which this optional should be compared to
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
```
MatthiasKillat, line 275: Needs to be tested in `test_cxx_optional.cpp` (in existing test case or new one).

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: sure

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: Addressed

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: @MatthiasKillat New tests are added.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: @MatthiasKillat New tests are added.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: @MatthiasKillat It is false positive, and is now been disabled for these scenarios. i.e "overloaded_fref":

/cc @FerdinandSpitzschnueffler

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: as addressed above

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I would prefer to remove the reference so that the suppression lines here and in the header file can be removed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -24,11 +24,13 @@ namespace iox
 {
 namespace cxx
 {
-// m_data is not initialized since this is a constructor for an optional with no value; an access of the value would
-// lead to the application's termination
+// AXIVION DISABLE STYLE AutosarC++19_03-A12.6.1 : m_data is not initialized here, since this is a
```
FerdinandSpitzschnueffler, line 27: Where is this rule enabled again?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+/// @param[in] rhs value to which lhs should be compared to
+/// @return true if the contained values are not equal, otherwise false
+template <typename T>
+bool operator!=(const optional<T>& lhs, const optional<T>& rhs) noexcept;
+
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs value to which lhs should be compared to
+/// @return true if the optional is set, otherwise false
```
FerdinandSpitzschnueffler, line 273:
```suggestion
/// @return true if the rhs is not set, otherwise false
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +243,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+template <typename T>
+bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
+
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs value to which lhs should be compared to
+/// @param[in] rhs empty optional
+/// @return true if the optional is set, otherwise false
```
FerdinandSpitzschnueffler, line 280:
```suggestion
/// @return true if lhs is not set, otherwise false
```

FerdinandSpitzschnueffler (commented):
Review not finished, I just took a quick look.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -107,6 +111,10 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
         {
             construct_value(rhs.value());
         }
+        else
+        {
+            // do nothing
```
FerdinandSpitzschnueffler, line 116:
```suggestion
            // do nothing since this and rhs contain no values
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -128,6 +136,10 @@ inline optional<T>& optional<T>::operator=(optional&& rhs) noexcept
         {
             construct_value(std::move(rhs.value()));
         }
+        else
+        {
+            // do nothing
```
FerdinandSpitzschnueffler, line 141:
```suggestion
            // do nothing since this and rhs contain no values
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -260,26 +236,36 @@ template <typename T>
...
+    // All accesses to the underlying data is done via the same static type and therefore the
+    // casts are always valid
     return std::move(*static_cast<T*>(static_cast<void*>(&m_data)));
 }
 
 template <typename T>
 inline const T&& optional<T>::value() const&& noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast) const_cast to avoid code duplication
```
FerdinandSpitzschnueffler, line 269:
```suggestion
    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
```
Since it is stated in the line above.

iceoryx_hoofs/test/moduletests/test_cxx_optional.cpp
```diff
@@ -215,6 +217,8 @@ TEST_F(Optional_test, NotCompareWithNullopt)
 {
     ::testing::Test::RecordProperty("TEST_ID", "0668eb1b-bc73-4549-9829-ece80f3700cd");
     EXPECT_THAT(m_sutWithValue != iox::cxx::nullopt_t(), Eq(true));
+    EXPECT_THAT(iox::cxx::nullopt_t() != m_sutWithValue, Eq(true));
+    EXPECT_THAT(iox::cxx::nullopt_t() != m_sutNoValue, Eq(false));
```
FerdinandSpitzschnueffler, line 221: You could also add
`EXPECT_THAT(m_sutNoValue != iox::cxx::nullopt_t(), Eq(false));`

saif-at-github: Sure

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +242,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+// AXIVION DISABLE STYLE AutosarC++19_03-A13.5.5: Comparison with nullopt_t is required
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs optional to which lhs should be compared to
+/// @return true if the rhs is not set, otherwise false
+template <typename T>
+bool operator==(const nullopt_t, const optional<T>& rhs) noexcept;
+
+/// @brief Comparison for equality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs optional to which lhs should be compared to
```
FerdinandSpitzschnueffler, line 277:
```suggestion
/// @param[in] lhs optional which should be compared to cxx::nullopt_t
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -253,13 +242,58 @@ class optional : public FunctionalInterface<optional<T>, T, void>
...
+
+/// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs empty optional, cxx::nullopt_t
+/// @param[in] rhs optional to which lhs should be compared to
+/// @return true if the optional is set, otherwise false
+template <typename T>
+bool operator!=(const nullopt_t, const optional<T>& rhs) noexcept;
+
+/// @brief Comparison for inequality with nullopt_t for easier unset optional comparison
+/// @param[in] lhs optional to which lhs should be compared to
```
FerdinandSpitzschnueffler, line 291:
```suggestion
/// @param[in] lhs optional which should be compared to cxx::nullopt_t
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -89,7 +92,9 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
 {
     construct_value(std::forward<Targs>(args)...);
 }
+// AXIVION ENABLE STYLE AutosarC++19_03-A12.6.1
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : It's an abstract call and cause no confusion
```
FerdinandSpitzschnueffler, line 97: I think it's a false positive, please see [this comment](https://github.com/eclipse-iceoryx/iceoryx/pull/1607#discussion_r984384205). @elfenpiff What do you think? 
If this is true, the justification could be something like "False positive. Overloading excluded via std::enable_if in typename std::enable_if<!std::is_same<U, optional<T>&>::value, optional>::type& operator=(U&& value) noexcept"

saif-at-github: @FerdinandSpitzschnueffler updated the justification as per @elfenpiff . thanks

saif-at-github:
@mossmaurice @FerdinandSpitzschnueffler I think all the comments are fixed. ready to merge.

saif-at-github:
> We should still test the new operator== and operator!= overloads where the first argument has type nullopt_t.

This is addressed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
elfenpiff: @saif-at-github @FerdinandSpitzschnueffler maybe I put this justification in the wrong thread or answered to the wrong question but the justification was meant for instance for the functional interface class.

FerdinandSpitzschnueffler: @saif-at-github Can you please update the justification? Maybe you can reuse my first comment on this.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -89,7 +92,10 @@ inline optional<T>::optional(in_place_t, Targs&&... args) noexcept
 {
     construct_value(std::forward<Targs>(args)...);
 }
+// AXIVION ENABLE STYLE AutosarC++19_03-A12.6.1
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : False positive. Overloading excluded via std::enable_if in typename
+// std::enable_if<!std::is_same<U, optional&>::value, optional>::type& operator=(U&& value) noexcept.
```
FerdinandSpitzschnueffler, line 98: For better readability:
```suggestion
// AXIVION Next Construct AutosarC++19_03-A13.3.1 : False positive. Overloading excluded via std::enable_if in
// typename std::enable_if<!std::is_same<U, optional&>::value, optional>::type& operator=(U&& value) noexcept.
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
saif-at-github: @FerdinandSpitzschnueffler  this is addressed. Thanks

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/optional.inl
```diff
@@ -107,10 +113,17 @@ inline optional<T>& optional<T>::operator=(const optional& rhs) noexcept
...
         }
+        else
+        {
+            // do nothing since this and rhs contain no values
+        }
     }
     return *this;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A13.3.1 : Used in an internal abstract interface class with default
```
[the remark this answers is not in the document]
saif-at-github: @FerdinandSpitzschnueffler Sure.

FerdinandSpitzschnueffler (approved):

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -185,6 +170,7 @@ class optional : public FunctionalInterface<optional<T>, T, void>
 
     /// @brief Will return true if the optional contains a value, otherwise false.
     /// @return true if optional contains a value, otherwise false
+    // AXIVION Next Construct AutosarC++19_03-A13.5.3: requires a user-defined conversion
```
elfenpiff, line 173: This is in my opinion not a valid justification. Please remember to write (when it is fitting) why we are doing it and why it is safe.
One example would be:
Implemented to be as close to the STL interface as possible. In combination with the keyword explicit accidental casts can be excluded and the usage is well known in the C++ community.

elfenpiff (changes requested):
The public API of the optional comparision operators has changed. Please add this to `doc/website/release-notes/iceoryx-unreleased.md` as refactoring item and an additional migration path for users at the end of the file. Here I would say that on example with a comparison operator suffices.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -185,6 +170,8 @@ class optional : public FunctionalInterface<optional<T>, T, void>
 
     /// @brief Will return true if the optional contains a value, otherwise false.
     /// @return true if optional contains a value, otherwise false
+    // AXIVION Next Construct AutosarC++19_03-A13.5.3: designed implementation shall be as close
+    // as std::optional and it requires operator bool
```
FerdinandSpitzschnueffler, line 174: I think the justification is still not precise enough. Can you please copy @elfenpiff's suggestion?

FerdinandSpitzschnueffler:
> The public API of the optional comparision operators has changed. Please add this to `doc/website/release-notes/iceoryx-unreleased.md` as refactoring item and an additional migration path for users at the end of the file. Here I would say that on example with a comparison operator suffices.

Since the usage has not changed I think we don't necessarily need to mention this in the release note. @elfenpiff What do you think?

elfenpiff:
> > The public API of the optional comparision operators has changed. Please add this to `doc/website/release-notes/iceoryx-unreleased.md` as refactoring item and an additional migration path for users at the end of the file. Here I would say that on example with a comparison operator suffices.
> 
> Since the usage has not changed I think we don't necessarily need to mention this in the release note. @elfenpiff What do you think?

I am alright with this!

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
saif-at-github: @FerdinandSpitzschnueffler Have updated the justification

saif-at-github:
> The public API of the optional comparision operators has changed. Please add this to `doc/website/release-notes/iceoryx-unreleased.md` as refactoring item and an additional migration path for users at the end of the file. Here I would say that on example with a comparison operator suffices.

Since the usage has not changed we don't need to mention this in the release note. Thanks.

FerdinandSpitzschnueffler (approved):

MatthiasKillat (approved):

elfenpiff (approved):

PR #1620, opened 2022-09-08
[DONOTMERGE] Iox 1619 iceoryx_platform PART 1: add cmake switch to set custom iceoryx platform

### LINKED ISSUE #1619: Introduce cmake switch to set custom iceoryx platform
## Brief feature description

At the moment the platform like windows/linux/mac etc. is autodetected by cmake. This prevents the user to add new custom platforms for instance setups like windows with gcc. A switch in cmake which overrides the platform autodetection and provides a path to a user platform would solve the issue.

Extracted from: #1432

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
I introduced a `settings.cmake` in every platform where the `ICEORYX_PLATFORM_STRING`, `ICEORYX_CXX_STANDARD` and `ICEORYX_{C|CXX}_WARNINGS` are defined.
The problem I encountered was that custom platforms were unable to set or disable certain warning flags. For instance when they are not available on a platform and that every platform was forced on C++17 (or default value for unknown platforms).
On certified platforms we would like to restrict the compiler to C++14 on any other platform the developer is free to use any other standard when it is at least c++14.

## References

- Closes #1619


iceoryx_meta/build_options.cmake
```diff
@@ -29,6 +29,7 @@ option(DOWNLOAD_TOML_LIB "Download cpptoml via the CMake ExternalProject module"
 option(EXAMPLES "Build all iceoryx examples" OFF)
 option(INTROSPECTION "Builds the introspection client which requires the ncurses library with an activated terminfo feature" OFF)
 option(ONE_TO_MANY_ONLY "Restricts communication to 1:n pattern" OFF)
+option(PLATFORM_PATH "Overrides integrated platform detection and uses provided custom path" OFF)
```
dkroenke, line 32: I think it is not a good idea to mix a CMake option (which is treated as boolean) with a CMake argument (given as string value).

For me the `PLATFORM_PATH` look similar to the `CMAKE_TOOLCHAIN_FILE` argument which we also treat not as option.

What you need to do is to check if the variable is set and have a value (two different things in CMake)

Maybe something like this could work (not tested):
```
if(DEFINED PREFIX OR NOT PREFIX STREQUAL "")
```

dkroenke (changes requested):

iceoryx_hoofs/platform/qnx/settings.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+if(BUILD_STRICT)
+    set_global(VAR ICEORYX_C_WARNINGS       VALUE ${ICEORYX_C_WARNINGS} -Werror)
+    set_global(VAR ICEORYX_CXX_WARNINGS     VALUE "${ICEORYX_CXX_WARNINGS} -Werror)
+endif()
+
+
+# check platform requirements
+
+if(NOT CMAKE_CXX_COMPILER_ID MATCHES "GNU")
```
dkroenke, line 31: We need to consider here `QCC` as `CMAKE_CXX_COMPILER_ID`

doc/website/advanced/custom-iceoryx-platform.md
```diff
@@ -0,0 +1,37 @@
+# Custom iceoryx platforms
+
+The `iceoryx_hoofs/platform` represents the lowest layer in iceoryx. Its task is to establish
+a uniform behavior of low level system calls across all supported platforms. When one would like
+to add support for another platform one can use the cmake switch `-DPLATFORM_PATH` to provide
```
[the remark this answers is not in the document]
elfenpiff: I used `set(PLATFORM_PATH "" CACHE PATH "...")` instead of `option(PLATFORM_PATH...`. `set` transforms every user provided path into an absolute path, see: https://cmake.org/cmake/help/latest/command/set.html

iceoryx_meta/build_options.cmake
elfenpiff: I used `set(PLATFORM_PATH "" CACHE PATH "description")` which solves also the relative path issue.

iceoryx_integrationtest/package.xml
```diff
@@ -15,6 +15,7 @@
   <build_depend>ros_testing</build_depend>
 
   <test_depend>ros_testing</test_depend>
+  <test_depend>iceoryx_platform</test_depend>
```
dkroenke, line 18: This will not work since CMake library does not have an own package.xml file. Therefore colcon will not know where the package is located and cannot build it before the integration tests.
Just adding a package.xml to `iceoryx_hoofs/platform` will also not work since iceoryx_hoofs already contains a package.xml file and having a ROS 2 package inside of another package is not allowed.
The only chance here is to flesh out platform into an own package beside iceoryx_hoofs and posh to make it work.

We should think carefully before doing that to consider aspects of integration etc.

elfenpiff: The follow up PR #1627 addresses exactly this issue by introducing a `package.xml` for the iceoryx platform.

PR #1634, opened 2022-09-13
iox-1469 implement copy and move operations for stack

### LINKED ISSUE #1469: stack issues with non trivially copyable types
## Required information

The stack is movable and copyable by default. The raw data array is copied implicitly in the compiler generated function which does not take the copy or move operations into account as well as the destructor when the stack is destroyed or pop'ed elements are overridden with push.

This can be solved by explicitly implementing the copy and move operations as well as the destructor to handle non trivially copyable types inside the stack. 
Another approach could be to restrict the stack to trivially copyable and destructable types.

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR implements the destructor, copy and move operations for the `cxx::stack`.

## References

- Closes #1469

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
@@ -32,7 +32,7 @@ namespace cxx
 /// @tparam T type which the stack contains
 /// @tparam Capacity the capacity of the stack
 template <typename T, uint64_t Capacity>
-class stack final // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member-init)
+class stack final // NOLINT(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 {
   public:
     stack() noexcept = default;
@@ -53,6 +53,9 @@ class stack final // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member
     template <typename... Targs>
     bool push(Targs&&... args) noexcept;
 
+    /// @brief calls the destructor of all contained elements in reverse creation order
+    void clear() noexcept;
+
     /// @brief returns the stack size
     uint64_t size() const noexcept;
 
@@ -62,6 +65,8 @@ class stack final // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member
   private:
     T& getUnchecked(const uint64_t index) noexcept;
     const T& getUnchecked(const uint64_t index) const noexcept;
+
+    void clearUntil(const uint64_t index) noexcept;
 
     // AXIVION Next Construct AutosarC++19_03-A18.1.1 : safe access is guaranteed since the char array is wrapped inside
     // the stack class
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
@@ -30,7 +30,7 @@ namespace cxx
 /// @tparam T type which the stack contains
 /// @tparam Capacity the capacity of the stack
 template <typename T, uint64_t Capacity>
-class stack // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member-init)
+class stack final // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member-init)
 {
   public:
     stack() noexcept = default;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
@@ -33,6 +33,8 @@ template <typename T, uint64_t Capacity>
 class stack // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member-init)
 {
   public:
+    ~stack() noexcept;
+
     /// @brief returns the last pushed element when the stack contains elements
     ///         otherwise a cxx::nullopt
     cxx::optional<T> pop() noexcept;
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
@@ -34,8 +34,10 @@ class stack // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member-init)
 {
   public:
     stack() noexcept = default;
-    stack& operator=(const stack& rhs) noexcept;
     stack(const stack& rhs) noexcept;
+    stack(stack&& rhs) noexcept;
+    stack& operator=(const stack& rhs) noexcept;
+    stack& operator=(stack&& rhs) noexcept;
     ~stack() noexcept;
 
     /// @brief returns the last pushed element when the stack contains elements
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
@@ -16,7 +16,9 @@
 #ifndef IOX_HOOFS_CXX_STACK_HPP
 #define IOX_HOOFS_CXX_STACK_HPP
 
+#include "iceoryx_hoofs/cxx/algorithm.hpp"
 #include "iceoryx_hoofs/cxx/optional.hpp"
+
 #include <cstdint>
 
 namespace iox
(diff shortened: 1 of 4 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/stack.inl
```diff
@@ -22,6 +22,119 @@ namespace iox
...
+        // delete remaining elements
+        for (; i < m_size; i++)
+        {
+            getUnchecked(i).~T();
+        }
+
+        m_size = rhsSize;
+
+        // clear rhs
+        for (uint64_t j{0}; j < rhsSize; j++)
```
elfenpiff, line 103: Could you please introduce a public method called clear which calls the destructor of all elements and sets the size to zero and call this method in the copy and move assignment as well as the destructor.

Furthermore, we have to remove the elements in revert direction from `m_size - 1` ... `0`.
When you for instance declare 3 variables or have 3 members a, b and c like:
```
int a;
int b;
int c;
```

The destruction order is always in the opposite creation order, so first `c` is destructed then `b` and then `a`. It is strange but possible for instance that element `c` inside the stack has a reference to `b` and requires that reference during destruction. This would be completely legal since the stack resides on the heap and never relocates.

With the current stack destruction behavior we would have a problem.

Furthermore, see the C++ standard: https://isocpp.org/files/papers/N4860.pdf, section 7.6.2.8 Delete, subsection 6 (page 129)
```
In the case of an array, the elements will
be destroyed in order of decreasing address (that is, in reverse order of the completion of their constructor;
see 11.10.2).
```

Fun fact, when I compiled this piece of code:
```cxx
#include <iostream>
#include <stack>
#include <vector>

struct A {
    A(int a) : a{a} { std::cout << "create: " << a << std::endl; }
    ~A() { std::cout << "destroy: " << a << std::endl; }

    int a;
};

// namespace iox
int main() {
    std::vector<A> a;
    a.push_back(A(1));
    a.push_back(A(2));
    a.push_back(A(3));
}
```

It showed me that my STL implementation does not destroy the elements in reverse order. (Same goes when you replace `std::vector` with `std::stack`). This is maybe a bug.

elfenpiff: When we agree that this has to be adjusted we also would require a unit tests which verifies the reverse destruction order.
@elBoberido what do you think?

elfenpiff: When I am right we have another funky problem since we are doing it wrong in every container.

elBoberido: @elfenpiff I think you are right and it should be done in reverse order, at least for container where the items are stored consecutively. Not sure what to do in a `list`. A double linked list could also be destroyed in reverse order but with a single linked list this is hard to achieve

@MatthiasKillat what do you think?

FerdinandSpitzschnueffler: @elfenpiff I think you're right. I will adapt the destruction order here, check our other containers and create an issue.

FerdinandSpitzschnueffler: I created an issue for `cxx::vector`: #1643. The other containers seem to be alright, although the destruction order for the lists may be debatable.

elfenpiff (approved):

iceoryx_hoofs/test/moduletests/test_cxx_stack.cpp
```diff
@@ -118,6 +209,269 @@ TEST_F(stack_test, popCreatesSpaceForAnotherElement)
...
+    EXPECT_THAT(TestClass::dTor, Eq(1));
 }
 
+TEST_F(stack_test, TestClassDTorIsCalledWhenStackGoesOutOfScope)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "3c496cb7-898b-4a65-a405-c42cbd7f0d7b");
+    {
+        cxx::stack<TestClass, STACK_SIZE> sut;
+        sut.push();
+        sut.push(1U, 2U, 3U);
```
MatthiasKillat, line 221: remove the U? It currently is inconsistent with later tests.

iceoryx_hoofs/test/moduletests/test_cxx_stack.cpp
```diff
@@ -118,6 +209,269 @@ TEST_F(stack_test, popCreatesSpaceForAnotherElement)
...
+
+TEST_F(stack_test, CopyConstructorWithEmptyStackWorks)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "08bfe7d9-233e-47cc-a7ca-5520eb6b99df");
+    cxx::stack<TestClass, STACK_SIZE> testStack(m_sut);
+    EXPECT_THAT(TestClass::copyCTor, Eq(0));
+    EXPECT_THAT(testStack.size(), Eq(0));
+}
+
+TEST_F(stack_test, CopyConstructorWithFullVectorWorks)
```
MatthiasKillat, line 265:
```suggestion
TEST_F(stack_test, CopyConstructorWithFullStackWorks)
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
```diff
@@ -30,9 +32,16 @@ namespace cxx
 /// @tparam T type which the stack contains
 /// @tparam Capacity the capacity of the stack
 template <typename T, uint64_t Capacity>
-class stack // NOLINT (cppcoreguidelines-pro-type-member-init,hicpp-member-init)
+class stack final // NOLINT(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: In my opinion we should use the `final` keyword. The stack class includes a public d'tor that is not virtual and therefore should not be used as a base class. Once we need the class as a base class, we can remove the `final` keyword and make the d'tor virtual. Since we discussed this topic in other PRs as well, I suggest to find a consensus in a separate issue and, if we decide to disable the rule, create a PR in which we remove all `final` occurrences.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/stack.inl
```diff
@@ -22,6 +22,122 @@ namespace iox
...
+    return *this;
+}
+
+// AXIVION Next Construct AutosarC++19_03-A12.6.1 : the remaining m_data fields are explicitly initialized when a new
+// element is pushed
+// NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
+template <typename T, uint64_t Capacity>
+inline stack<T, Capacity>::stack(const stack& rhs) noexcept
+{
+    copy(rhs);
```
elfenpiff, line 84: Why did you introduce `copy` and `move` methods and do not use `*this = rhs` or `*this = std::move(rhs)`

FerdinandSpitzschnueffler: To avoid the self assignment check in the c'tors.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/stack.hpp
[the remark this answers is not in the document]
MatthiasKillat: We can do this. Its surely is no issue now, but the AUTOSAR rule itself is obsolete in modern C++ with smart pointers.

iceoryx_hoofs/test/moduletests/test_cxx_stack.cpp
MatthiasKillat: I see. Not a big deal but weird.

MatthiasKillat (approved):

elfenpiff (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/stack.inl
```diff
@@ -22,6 +22,104 @@ namespace iox
...
+        const uint64_t rhsSize{rhs.size()};
+        const uint64_t minSize{algorithm::minVal(m_size, rhsSize)};
+
+        // copy assignment
+        for (; i < minSize; i++)
+        {
+            getUnchecked(i) = rhs.getUnchecked(i);
+        }
+        // copy c'tor
+        for (; i < rhsSize; i++)
```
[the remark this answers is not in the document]
MatthiasKillat: It is not about worst case BTW. It is about only doing what is needed in all cases. At leas if some of them can be considered likely.

PR #1644, opened 2022-09-15
Iox 1104 remove nullability from unique ptr

### LINKED ISSUE #1104: `cxx::function_ref`, `cxx::function` and `cxx::unique_ptr` should not be nullable
## Brief feature description

The feature of a nullable `function_ref` and `function` can lead to a very easy misuse since one has to always check the `function`/`function_ref` if it is set before using it.

This causes:
* extra overhead
* increases the probability for undefined behavior/bugs

If one would like to have a nullable `function`/`function_ref` one can always wrap this into `cxx::optional` to make it clear that this is maybe null.

That this can happen very easily is demonstrated in the approved PR: https://github.com/eclipse-iceoryx/iceoryx/pull/1094#discussion_r805948235

ToDo:
 - [x] `function_ref`
 - [ ] `function`
 - [x] `unique_ptr`

elfenpiff:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #1104

### DIFF
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -84,6 +84,7 @@
 - Renamed `cxx::GenericRAII` to `cxx::ScopeGuard` [\#1450](https://github.com/eclipse-iceoryx/iceoryx/issues/1450)
 - Rename `algorithm::max` and `algorithm::min` to `algorithm::maxVal` and `algorithm::minVal` [\#1394](https://github.com/eclipse-iceoryx/iceoryx/issues/1394)
 - Extract `iceoryx_hoofs/platform` into separate package `iceoryx_platform` [\#1615](https://github.com/eclipse-iceoryx/iceoryx/issues/1615)
+- `cxx::unique_ptr` is no longer nullable [\#1104](https://github.com/eclipse-iceoryx/iceoryx/issues/1104)
 
 **Workflow:**
 
@@ -399,4 +400,24 @@
 
     // after
     #include "iceoryx_platform/some_header.hpp"
+    ```
+
+22. `cxx::unique_ptr` is no longer nullable.
+
+    ```cxx
+    // before
+    cxx::unique_ptr<int> myPtr(ptrToInt, someDeleter);
+    cxx::unique_ptr<int> emptyPtr(nullptr, someDeleter);
+
+    myPtr.reset(ptrToOtherInt);
+    myPtr.release();
+
+
+    // after
+    cxx::unique_ptr<int> myPtr(*ptrToInt, someDeleter); // requires reference to object to ensure non nullptr
+    cxx::optional<cxx::unique_ptr<int>> emptyPtr(cxx::nullopt); // if unique_ptr shall be nullable use cxx::optional
+
+    myPtr.reset(*ptrToOtherInt); // requires reference to object to ensure non nullptr
+    cxx::unique_ptr<int>::release(std::move(myPtr)); // release consumes myPtr
+
     ```
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -409,15 +409,25 @@
     cxx::unique_ptr<int> myPtr(ptrToInt, someDeleter);
     cxx::unique_ptr<int> emptyPtr(nullptr, someDeleter);
 
+    if (myPtr) { // required since the object could always be null
+        std::cout << *myPtr << std::endl;
+    }
+
     myPtr.reset(ptrToOtherInt);
     myPtr.release();
 
 
     // after
-    cxx::unique_ptr<int> myPtr(*ptrToInt, someDeleter); // requires reference to object to ensure non nullptr
+    cxx::unique_ptr<int> myPtr(ptrToInt, someDeleter);
     cxx::optional<cxx::unique_ptr<int>> emptyPtr(cxx::nullopt); // if unique_ptr shall be nullable use cxx::optional
 
-    myPtr.reset(*ptrToOtherInt); // requires reference to object to ensure non nullptr
-    cxx::unique_ptr<int>::release(std::move(myPtr)); // release consumes myPtr
+    // no more null check required since it is no longer nullable
+    std::cout << *myPtr << std::endl;
 
+    myPtr.reset(ptrToOtherInt);
+    cxx::unique_ptr<int>::release(std::move(myPtr)); // release consumes myPtr
     ```
+
+    Compilers like ``gcc-12>`` and `clang>14` as well as static code analysis tools like `clang-tidy`
+    will warn the user with a used after move warning when one accesses a moved object. Accessing
+    a moved `unique_ptr` is well defined and behaves like dereferencing a `nullptr`.
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
@@ -84,16 +84,17 @@ class unique_ptr
 
     ///
     /// @brief release Release ownership of the underlying pointer.
-    /// @return Pointer to the managed object or nullptr if none owned.
+    /// @param[in] ptrToBeReleased unique_ptr which is destroyed without deleting its underlying data
+    /// @return Pointer to the managed object.
     ///
     static T* release(unique_ptr&& ptrToBeReleased) noexcept;
 
     ///
     /// @brief reset Reset the unique pointer to take ownership of the given pointer.
-    /// @details Any previously owned objects will be deleted. If no pointer given then points to nullptr.
-    /// @param ptr Pointer to object to take ownership on.
+    /// @details Any previously owned objects will be deleted.
+    /// @param ptr Pointer to object to take ownership on. It is forbidden to provide the value nullptr!
     ///
-    void reset(T* const object) noexcept;
+    void reset(T* const ptr) noexcept;
 
     ///
     /// @brief swap Swaps object ownership with another unique_ptr (incl. deleters)
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
@@ -19,6 +19,7 @@
 #define IOX_HOOFS_CXX_UNIQUE_PTR_HPP
 
 #include "iceoryx_hoofs/cxx/function.hpp"
+#include "iceoryx_hoofs/cxx/requires.hpp"
 
 namespace iox
 {
@@ -40,10 +41,10 @@ class unique_ptr
     /// @brief unique_ptr Creates a unique pointer that takes ownership of an object.
     /// @details A deleter must always be provided as no default can be provided given that no heap is used.
     /// The unique_ptr must know how to delete the managed object when the pointer goes out of scope.
-    /// @param object The reference to the object to be managed.
+    /// @param object The pointer to the object to be managed.
     /// @param deleter The deleter function for cleaning up the managed object.
     ///
-    unique_ptr(T& object, const function<void(T*)>& deleter) noexcept;
+    unique_ptr(T* const object, const function<void(T*)>& deleter) noexcept;
 
     unique_ptr(const unique_ptr& other) = delete;
     unique_ptr& operator=(const unique_ptr&) = delete;
@@ -92,7 +93,7 @@ class unique_ptr
     /// @details Any previously owned objects will be deleted. If no pointer given then points to nullptr.
     /// @param ptr Pointer to object to take ownership on.
     ///
-    void reset(T& object) noexcept;
+    void reset(T* const object) noexcept;
 
     ///
     /// @brief swap Swaps object ownership with another unique_ptr (incl. deleters)
--- a/iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
@@ -40,10 +40,10 @@ class unique_ptr
     /// @brief unique_ptr Creates a unique pointer that takes ownership of an object.
     /// @details A deleter must always be provided as no default can be provided given that no heap is used.
     /// The unique_ptr must know how to delete the managed object when the pointer goes out of scope.
-    /// @param ptr The raw pointer to the object to be managed.
+    /// @param object The reference to the object to be managed.
     /// @param deleter The deleter function for cleaning up the managed object.
     ///
-    unique_ptr(T* const ptr, const function<void(T*)>& deleter) noexcept;
+    unique_ptr(T& object, const function<void(T*)>& deleter) noexcept;
 
     unique_ptr(const unique_ptr& other) = delete;
     unique_ptr& operator=(const unique_ptr&) = delete;
@@ -54,9 +54,6 @@ class unique_ptr
     /// @brief Automatically deletes the managed object on destruction.
     ///
     ~unique_ptr() noexcept;
-
-
-    unique_ptr<T>& operator=(std::nullptr_t) noexcept;
 
     ///
     /// @brief operator -> Transparent access to the managed object.
@@ -69,11 +66,6 @@ class unique_ptr
     /// @return Const pointer to the stored object
     ///
     const T* operator->() const noexcept;
-
-    ///
-    /// @brief operator bool Returns true if it points to something.
-    ///
-    explicit operator bool() const noexcept;
 
     ///
     /// @brief get Retrieve the underlying raw pointer.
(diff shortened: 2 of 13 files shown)


doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -400,3 +401,23 @@
...
+
+
+    // after
+    cxx::unique_ptr<int> myPtr(*ptrToInt, someDeleter); // requires reference to object to ensure non nullptr
+    cxx::optional<cxx::unique_ptr<int>> emptyPtr(cxx::nullopt); // if unique_ptr shall be nullable use cxx::optional
+
+    myPtr.reset(*ptrToOtherInt); // requires reference to object to ensure non nullptr
+    cxx::unique_ptr<int>::release(std::move(myPtr)); // release consumes myPtr
+
+    ```
```
elBoberido, line 423: Could you maybe document somewhere that gcc/clang will issue a warning if a moved-from object is used and since there is no other way the unique_ptr can be invalid, it also does not need to be checked for null?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
```diff
@@ -40,10 +40,10 @@ class unique_ptr
     /// @brief unique_ptr Creates a unique pointer that takes ownership of an object.
     /// @details A deleter must always be provided as no default can be provided given that no heap is used.
     /// The unique_ptr must know how to delete the managed object when the pointer goes out of scope.
-    /// @param ptr The raw pointer to the object to be managed.
+    /// @param object The reference to the object to be managed.
     /// @param deleter The deleter function for cleaning up the managed object.
     ///
-    unique_ptr(T* const ptr, const function<void(T*)>& deleter) noexcept;
+    unique_ptr(T& object, const function<void(T*)>& deleter) noexcept;
```
elBoberido, line 46: Please not. This is not idiomatic C++. A reference does not transfer ownership but the class takes the ownership of the pointer/reference. I know what you try to do but this is the wrong way to do it. Can you please use `cxx::non_null`? Maybe we can extend it to have a ctor which takes a reference.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
```diff
@@ -93,26 +85,47 @@ class unique_ptr
...
-    T* release() noexcept;
+    static T* release(unique_ptr&& ptrToBeReleased) noexcept;
 
     ///
     /// @brief reset Reset the unique pointer to take ownership of the given pointer.
     /// @details Any previously owned objects will be deleted. If no pointer given then points to nullptr.
     /// @param ptr Pointer to object to take ownership on.
     ///
-    void reset(T* const ptr = nullptr) noexcept;
+    void reset(T& object) noexcept;
```
elBoberido, line 95: same here

iceoryx_posh/include/iceoryx_posh/internal/popo/smart_chunk.hpp
```diff
@@ -42,7 +42,7 @@ struct SmartChunkPrivateData
     SmartChunkPrivateData& operator=(const SmartChunkPrivateData&) = delete;
     ~SmartChunkPrivateData() = default;
 
-    cxx::unique_ptr<T> smartChunkUniquePtr;
+    cxx::optional<cxx::unique_ptr<T>> smartChunkUniquePtr;
```
elBoberido, line 45: Do we already have an issue to make the smart chunk also non-nullable?

iceoryx_posh/include/iceoryx_posh/internal/popo/smart_chunk.inl
```diff
@@ -129,7 +129,9 @@ inline const R& SmartChunk<TransmissionInterface, T, H>::getUserHeader() const n
 template <typename TransmissionInterface, typename T, typename H>
 inline T* SmartChunk<TransmissionInterface, T, H>::release() noexcept
 {
-    return m_members.smartChunkUniquePtr.release();
+    auto ptr = cxx::unique_ptr<T>::release(std::move(*m_members.smartChunkUniquePtr));
+    m_members.smartChunkUniquePtr.reset();
```
elBoberido, line 133: How about overloading `unique_ptr::release` for `optional<unique_ptr<T>>&`? Then this would just become
```cpp
auto ptr = cxx::unique_ptr<T>::release(m_members.smartChunkUniquePtr);
```
The the move and reset would happen within the release function.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
elfenpiff: Here I would disagree, especially since it is idiomatic C++ in my opinion. Stating the C++ FAQ from isocpp.org: 

```
Use references when you can, and pointers when you have to.

References are usually preferred over pointers whenever you don’t need [“reseating”]. This usually means that references are most useful in a class’s public interface. References typically appear on the skin of an object, and pointers on the inside.
```

see: https://isocpp.org/wiki/faq/references#refs-vs-ptrs

Furthermore it states at the end:
```
Note: Old line C programmers sometimes don’t like references since they provide reference semantics that isn’t explicit in the caller’s code. After some C++ experience, however, one quickly realizes this is a form of information hiding, which is an asset rather than a liability. E.g., programmers should write code in the language of the problem rather than the language of the machine.
```

So I suspect that you are an old line c programmer :smile:

elBoberido: I disagree. The intention of this rule is to pass references instead of raw pointer when you have an object.
```cpp
Foo foo;
bar(&foo); // bad
bar(foo) // good
```
It is different if you already have a pointer, which is the case when you use a `unique_ptr`.
```cpp
Foo* foo {nullptr};
bar(foo); // good when the nullptr check is done in bar
bar(*foo); // bad since we have a null reference
```

But it seems we are both wrong [I.11: Never transfer ownership by a raw pointer (T*) or reference (T&)](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i11-never-transfer-ownership-by-a-raw-pointer-t-or-reference-t) :laughing: 
That link refers to returning a pointer not to create a smart pointer. I would argue that it is more idiomatic to pass a raw pointer to as constructor of a smart pointer than a reference. Which leads us to [I.12: Declare a pointer that must not be null as not_null](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i12-declare-a-pointer-that-must-not-be-null-as-not_null)

```cpp
Foo* flup {new Foo()}; 
bar(non_null<T>(flup)); // better
bar(make_non_null(flup).value()) // make_non_null would return a `optional<non_null<T>>`
bar(make_non_null_unsafe_unchecked(flup)); // for if performance matters
```

iceoryx_posh/include/iceoryx_posh/internal/popo/smart_chunk.hpp
elfenpiff: I created one: https://github.com/eclipse-iceoryx/iceoryx/issues/1654

iceoryx_posh/include/iceoryx_posh/internal/popo/smart_chunk.inl
elfenpiff: I wouldn't do this. This is a corner case we currently only have for the smart chunk. I would keep it as simplistic as possible. Otherwise we have to do a `nullopt` check in release and what kind of pointer do you return when one puts in a `nullopt` - or call we the error handler which would introduce nice layers of error handling.

Otherwise the `nullptr` could be returned but then a user may easily misuse this api. They think oh the ptr are always valid therefore release always returns a valid non-null pointer and boom they are proved wrong.

On the other hand a call to the error handler will be also surprising.

elBoberido: Well, release could return a `optional<T*>`

elBoberido (approved):

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
```diff
@@ -93,26 +86,47 @@ class unique_ptr
     /// @brief release Release ownership of the underlying pointer.
     /// @return Pointer to the managed object or nullptr if none owned.
     ///
-    T* release() noexcept;
+    static T* release(unique_ptr&& ptrToBeReleased) noexcept;
 
     ///
     /// @brief reset Reset the unique pointer to take ownership of the given pointer.
     /// @details Any previously owned objects will be deleted. If no pointer given then points to nullptr.
```
FerdinandSpitzschnueffler, line 93: With the current changes one have to provide a pointer and this must not be a nullptr.

iceoryx_hoofs/test/moduletests/test_cxx_unique_ptr.cpp
```diff
@@ -168,67 +141,41 @@ TEST_F(UniquePtrTest, MoveAssignmentOverwriteAUniquePtrWithAnotherOneAndCallsAno
...
-    auto sut = iox::cxx::unique_ptr<Position>(object, deleter);
+    auto sut = iox::cxx::unique_ptr<Position>(&object, deleter);
 
     auto* objectPtr = sut.get();
 
     EXPECT_EQ(objectPtr->x, X_POS);
 }
 
-TEST_F(UniquePtrTest, ReleaseAnObjectResultsInUniquePtrBeingInvalidAndReturnOfObjectPtr)
+TEST_F(UniquePtrTest, ReleaseAnObjectResultsInDeleterBeingCalled)
```
FerdinandSpitzschnueffler, line 159: The deleter is not called.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
```diff
@@ -91,28 +84,50 @@ class unique_ptr
...
     ///
-    T* release() noexcept;
+    static T* release(unique_ptr&& ptrToBeReleased) noexcept;
 
     ///
     /// @brief reset Reset the unique pointer to take ownership of the given pointer.
-    /// @details Any previously owned objects will be deleted. If no pointer given then points to nullptr.
-    /// @param ptr Pointer to object to take ownership on.
+    /// @details Any previously owned objects will be deleted.
+    /// @param ptr Pointer to object to take ownership on. It is forbidden to provide the value nullptr!
```
mossmaurice, line 95: You could think about adding a hint, that the application might terminate in debug mode when a `nullptr` is provided.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
```diff
@@ -70,11 +68,6 @@ class unique_ptr
     ///
     const T* operator->() const noexcept;
 
-    ///
-    /// @brief operator bool Returns true if it points to something.
-    ///
-    explicit operator bool() const noexcept;
```
mossmaurice, line 76: It will not be possible anymore to find out if a `cxx::unique_ptr` is in a moved from object by using `operator bool()`. Not sure if we should force the user to call `get()` to check :thinking:

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
```diff
@@ -25,148 +25,111 @@ namespace iox
 namespace cxx
 {
 template <typename T>
-unique_ptr<T>::unique_ptr(T* const ptr, const function<void(T*)>& deleter) noexcept
-    : m_ptr(ptr)
+inline unique_ptr<T>::unique_ptr(T* const object, const function<void(T*)>& deleter) noexcept
+    : m_ptr(object)
     , m_deleter(std::move(deleter))
 {
+    Ensures(object != nullptr);
```
mossmaurice, line 32: Why not use `not_null`?

iceoryx_posh/include/iceoryx_posh/internal/popo/client_impl.inl
```diff
@@ -46,7 +46,7 @@ cxx::expected<Request<Req>, AllocationError> ClientImpl<Req, Res, BaseClientT>::
     }
     auto requestHeader = result.value();
     auto payload = mepoo::ChunkHeader::fromUserHeader(requestHeader)->userPayload();
-    auto request = cxx::unique_ptr<Req>(reinterpret_cast<Req*>(payload), [this](auto* payload) {
+    auto request = cxx::unique_ptr<Req>(static_cast<Req*>(payload), [this](auto* payload) {
```
mossmaurice, line 49: :heart:

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
elfenpiff: I think `not_null` is not the perfect type here. @elBoberido and me thought about a type which signals ownership like `Owns` or so but for now, keep it as minimal as possible and merge this thing.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
elfenpiff: No. The compiler will warn you if you do this. This is the use after move warning we are explicitly suppressing in our tests.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
elfenpiff: I would say also in release mode it should terminate. Otherwise it segfaults as soon as one calls `operator->` and then we should terminate as soon as possible

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
mossmaurice: @elfenpiff Compiler? You mean clang-tidy?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
mossmaurice: @elfenpiff Okidoki, no biggie. Can you elaborate your thoughts about `Owns` a bit more and/or provide and example?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
mossmaurice: @elfenpiff After a discussion with @MatthiasKillat we came to the conclusion that we would prefer the "traditional":

* Release mode: all assert / Expects / Ensures are no-ops for performance reason e.g. out of bounds check in `cxx::vector`
* Debug mode: all assert / Expects / Ensures are active

It's recommend for iceoryx user to test their software in debug mode extensively and only then switch to the release mode.

What do you think?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
elfenpiff: No I mean gcc>=12, clang>=14 and of course also clang-tidy

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
elfenpiff: It originated in a comment in this PR: https://github.com/eclipse-iceoryx/iceoryx/pull/1644#discussion_r974315972

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
```diff
@@ -25,148 +25,111 @@ namespace iox
 namespace cxx
 {
 template <typename T>
-unique_ptr<T>::unique_ptr(T* const ptr, const function<void(T*)>& deleter) noexcept
-    : m_ptr(ptr)
+inline unique_ptr<T>::unique_ptr(T* const object, const function<void(T*)>& deleter) noexcept
+    : m_ptr(object)
```
FerdinandSpitzschnueffler, line 29: For consistency reasons I would rename the parameter to `ptr`, but no changes required now.

FerdinandSpitzschnueffler (approved):

elBoberido (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
elBoberido: @mossmaurice the core guidelines talk about and `owned<T*>` in cases one does not use a `unique_ptr`. It might be slightly better than `not_null`. It might be able to implement an `owned<T*>` and even a `optional<owned<T*>>` as efficient as a raw pointer but with the disadvantage that we must rely on the compiler/clang-tidy to detect use after move.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/unique_ptr.hpp
elfenpiff: I think this discussion is not part of this pull request. But to clarify, I call `Ensures(ptr != nullptr)` in the implementation and this can be replaced with anything which we come up with in the non recoverable error handling design.

I think `IOX_ASSERT` would be its successor and at this point I do not care about its behavior.

mossmaurice: @elfenpiff Ok, makes sense. 

@MatthiasKillat Friendly reminder that we might to update few doxygen places e.g. here or in `cxx::optional` where it states `Will terminate if empty object is accessed`.

PR #1648, opened 2022-09-16
Iox 1640 create singleton abstraction

### LINKED ISSUE #1659: Create singleton abstraction to avoid static destruction order problems
## Brief feature description

The standard singleton implementation (static variable in function scope) does not allow multiple constructors and causes problems with destruction order if static variables depend on each other.

To avoid static destruction order problems a more general singleton abstraction can be created.

## Detailed information

- singleton abstraction shall allow destruction on demand
- singleton abstraction shall allow arbitrary multi argument ctors

TBD

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

A more powerful singleton.

This is not the polymorphic singleton, which I will add in another pull request (implemented but tests are missing).
Nevertheless this is a useful construction to solve problems originating from the life-time of static variables that may depend on each other, specifically in different translation units.

- Closes #1659


MatthiasKillat:
This is a more powerful singleton I used in some private applications (but added tests here etc.). We may not want it right now but it solves some problems I recently saw again internally.

In contrast to standard Meyers singletons it allows:
1. multi argument, non-default, multi ctor construction
2. reinitialization
3. better control over the point destruction

Unless there are bugs, it should have similar thread-safety guarantees and fast access in the fast path where the instance is already initialized (using the double checked locking pattern).

I will add the polymorphic singleton separately, it solves a different problem with different constraints. Although they can probably be combined, it could be messy (I did not try yet) and lead to less efficiency, so I would avoid that.

Whether to add it to dust or hoofs is a major question. For this PR dust is appropriate because we will not use it right away (but can consider it some time for e.g. Runtime and others).

For the upcoming polymorphic singleton I tend to hoofs, as we want to use it in error handling and maybe logging (although in the user/platform defined part).

FYI @elBoberido

iceoryx_dust/include/iceoryx_dust/singleton.hpp
```diff
@@ -0,0 +1,99 @@
...
+    static T& instance();
+
+  private:
+    using storage_t = typename std::aligned_storage_t<sizeof(T), alignof(T)>;
+
+    Singleton() = default;
+
+    // avoid the external definitions of regular statics (especially with templates)
+    // and use lazy initialization
+    static auto& storage();
```
elfenpiff, line 85: In my opinion the singleton is a little too hard to understand. One major problem is that I have to scroll through the implementation to see what kind of hidden members the singleton has.

I would suggest to introduce the members `m_lock`, `m_ptr`, and `m_storage` as well as a private method called `getInstance()` which contains the actual `static Singleton instance;`

Furthermore I would implement `initialize`, `init`, `destroy`, `instance`, `isInitialized` as non static private member functions so that the logic and structure of the object itself is clear. Since it is possible to have multiple singletons when using multiple types.

The public API static methods I would keep for convenience but under the roof I would perform a simple forwarding to the actual implementation in combination with `getInstance()`. Something like: 
```cxx
emplate <typename T>
template <typename... Args>
T& Singleton<T>::init(Args&&... args) {
  return getInstance().init_impl(std::forward<Args>(args)...);
}
```

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
```diff
@@ -0,0 +1,123 @@
...
+{
+    return ptr().load(std::memory_order_relaxed) != nullptr;
+}
+
+template <typename T>
+template <typename... Args>
+T& Singleton<T>::init(Args&&... args)
+{
+    std::lock_guard<std::mutex> g(lock());
+    if (!isInitialized())
```
elfenpiff, line 36: Shouldn't we report this with an error? I would return here an `cxx::expected<>` with an error enum stating that it is already initialized. What when I initialize the singleton again with some other arguments then I may wrongly assume that this is initialized differently. Especially since there is a race condition:
```cxx
if ( !Singleton::isInitialized() ) {
   // if this is called by multiple threads with different parameters some arbitrary thread wins
  Singleton::init(1,2,3);  // in init only one gets the lock the others are then dismissed feels wrong.
}
```

I know that you defined the behavior in the doxygen comment. But I would bet that during a review nearly no developer reads the full documentation of all function calls which they are reviewing and then such a bug sneaks in faster than one thinks.

iceoryx_dust/include/iceoryx_dust/singleton.hpp
```diff
@@ -0,0 +1,99 @@
...
+    /// @brief explicitly destroy the singleton instance if it is initialized
+    /// @note thread-safe
+    /// @note must ONLY be called if the instance is no longer accessed
+    ///       (this is not a severe restriction, as it is true for standard static singletons as well)
+    /// @note benefit: better control of the life-time of a singleton
+    /// @note if destroy is not explictly called, it will be called implictly after main,
+    ///       the usual rules (and problems) of static destruction order apply
+    static void destroy();
+
+    /// @brief explicitly destroy the singleton instance if it is initialized
```
elfenpiff, line 72: Copy and paste fail.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
```diff
@@ -0,0 +1,123 @@
...
+
+template <typename T>
+template <typename... Args>
+T* Singleton<T>::initialize(Args&&... args)
+{
+    static Singleton singleton; // dtor will be called later and call destroy
+    // NOLINTJUSTIFICATION implicit conversion from raw pointer is intentional in design of relocatable structures
+    // NOLINTNEXTLINE(cppcoreguidelines-owning-memory) T dtor will be called by singleton dtor
+    auto p = new (&storage()) T(std::forward<Args>(args)...);
+    ptr().store(p, std::memory_order_relaxed); // memory synced by lock
```
elfenpiff, line 117: In my opinion the synchronization order is wrong.

Let's assume the following steps:
1. Thread 1: `Singleton::init()` , mutex lock/unlock syncs with every other mutex lock/unlock
2. Thread 2: `Singleton::instance()`. `auto p = ptr().load(std::memory_order_acquire);` syncs with the corresponding `store(std::memory_order_release)`. which is in this case not existing. Therefore the memory is not synced.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
```diff
@@ -0,0 +1,123 @@
...
+
+template <typename T>
+bool Singleton<T>::isInitialized()
+{
+    return ptr().load(std::memory_order_relaxed) != nullptr;
+}
+
+template <typename T>
+template <typename... Args>
+T& Singleton<T>::init(Args&&... args)
```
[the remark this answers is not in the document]
elBoberido: This won't be a problem if we restrict T to not be copyable.

elBoberido (commented):
@MatthiasKillat sorry if I disappoint you but I stopped reviewing this when I noticed that it is not the PolymorphicHandler you sketched in your error handler draft. While this might be useful at some time, we won't make use of this in hoofs and that's my current focus. Since this contains non-trivial code it would cost me too much time to review it now. I will continue after the focus moves again away from hoofs.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
```diff
@@ -0,0 +1,123 @@
...
+        p->~T();
+        ptr().store(nullptr);
+    }
+}
+
+template <typename T>
+T& Singleton<T>::instance()
+{
+    // need to sync the memory at *p as well
+    auto p = ptr().load(std::memory_order_acquire);
```
elBoberido, line 61: @MatthiasKillat just to give some context to my previous statement why I think this construct is quite complex and it will cost me too much time to review it now.
This line breaks the singleton since one thread can call `instance()` and obtain the pointer while another thread can easily call `destroy` followed by `init`. This is totally valid with the pubic API but leaves the first thread with a dangling reference. One cannot know if someone calls `destroy`, therefore `instance` is inherently unsafe to use.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
```diff
@@ -0,0 +1,123 @@
...
+
+template <typename T>
+auto& Singleton<T>::storage()
+{
+    static storage_t s;
+    return s;
+}
+
+template <typename T>
+auto& Singleton<T>::ptr()
```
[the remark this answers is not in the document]
MatthiasKillat: It is a private function and will never return null, so there is no reason to do this.

MatthiasKillat:
@elBoberido No problem. We can defer this one. The apporach will still solve some internal problems and can be extended (with singleton dependency trees and whatnot).

 The polymorphic handler is here https://github.com/eclipse-iceoryx/iceoryx/pull/1656.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
MatthiasKillat: I know. This is intentional and cannot easily be guarded against (without huge performance hits, if even possible). A real Meyers Singleton has the same issue, just more subtle (another static can hold a reference to it and use this reference after the dtor was called).

The main purpose is to call destroy only during some shutdown function ad not defer destruction to some point after main.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
MatthiasKillat: The problem is we currently do not use it. I propose to use it later in posh.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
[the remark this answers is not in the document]
elfenpiff: I would state that assumptions can always change. Maybe someone else adopts the contract and then this has weird side-effects.
Therefore, I would future proof it and return a pointer here.

@elBoberido it is a tie, you decide.

iceoryx_dust/include/iceoryx_dust/internal/singleton.inl
elBoberido: @MatthiasKillat right. The main issue is that the API can easily be misused but we talked already about a possible alternative.

MatthiasKillat:
I am closing this and will recreate it for #1659. I initially thought I could combine this feature with the polymorphism but it turned out to be inefficient and overly complex to address both issues at once.

PR #1667, opened 2022-09-26
iox-#1666 Fix GatewayGeneric running state variable setting

### LINKED ISSUE #1666: GatewayGeneric sometimes terminates discovery and forward threads immediately
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Compiler version:**
GCC 7.4.0

When using the GatewayGeneric type to implement a new gateway, the discovery and forward thread sometimes terminate immediately, because the `m_isRunning` variable is only set to `true` after the threads are initialized. In other words, there is a race whether the worked threads or the thread calling `runMultithreaded` access the `m_isRunning` variable sooner.

Discovery and forward threads should run until terminated properly.

Run DDS gateway enough times.

JakubSosnovec:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

Closes #1666

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
@@ -37,9 +37,9 @@ inline GatewayGeneric<channel_t, gateway_t>::~GatewayGeneric() noexcept
 template <typename channel_t, typename gateway_t>
 inline void GatewayGeneric<channel_t, gateway_t>::runMultithreaded() noexcept
 {
+    m_isRunning.store(true, std::memory_order_relaxed);
     m_discoveryThread = std::thread([this] { this->discoveryLoop(); });
     m_forwardingThread = std::thread([this] { this->forwardingLoop(); });
-    m_isRunning.store(true, std::memory_order_relaxed);
 }
 
 template <typename channel_t, typename gateway_t>
--- a/iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
@@ -1,5 +1,6 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2020 - 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2022 by NXP. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -37,7 +38,7 @@ inline GatewayGeneric<channel_t, gateway_t>::~GatewayGeneric() noexcept
 template <typename channel_t, typename gateway_t>
 inline void GatewayGeneric<channel_t, gateway_t>::runMultithreaded() noexcept
 {
-    m_isRunning.store(true, std::memory_order_relaxed);
+    m_isRunning.store(true);
     m_discoveryThread = std::thread([this] { this->discoveryLoop(); });
     m_forwardingThread = std::thread([this] { this->forwardingLoop(); });
 }
@@ -45,11 +46,13 @@ inline void GatewayGeneric<channel_t, gateway_t>::runMultithreaded() noexcept
 template <typename channel_t, typename gateway_t>
 inline void GatewayGeneric<channel_t, gateway_t>::shutdown() noexcept
 {
-    if (m_isRunning.load(std::memory_order_relaxed))
+    m_isRunning.store(false);
+    if (m_discoveryThread.joinable())
     {
-        m_isRunning.store(false, std::memory_order_relaxed);
-
         m_discoveryThread.join();
+    }
+    if (m_forwardingThread.joinable())
+    {
         m_forwardingThread.join();
     }
 }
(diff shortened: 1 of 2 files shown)


JakubSosnovec:
@elBoberido I agree that this is very difficult to test.

I have modified the PR description and added a line to the release notes as you asked.

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
```diff
@@ -37,9 +37,9 @@ inline GatewayGeneric<channel_t, gateway_t>::~GatewayGeneric() noexcept
 template <typename channel_t, typename gateway_t>
 inline void GatewayGeneric<channel_t, gateway_t>::runMultithreaded() noexcept
 {
+    m_isRunning.store(true, std::memory_order_relaxed);
```
elfenpiff, line 40: I think this is still bugged. As far as I understand the as-if rule https://en.cppreference.com/w/cpp/language/as_if it is still legal for the compiler to reorder this code to:
```cxx
    m_discoveryThread = std::thread([this] { this->discoveryLoop(); });
    m_forwardingThread = std::thread([this] { this->forwardingLoop(); });
    m_isRunning.store(true, std::memory_order_relaxed);
```

since this does not violate the as-if rule.

What we actually have to implement is much more complex. Either we use the memory order sequentially consistent for the load and store for `m_isRunning` or we introduce an atomic barrier which prevents an reordering based on the as-if rule.

We need to add the sequentially consistent for load and store since I think that one requires always a corresponding counterpart with the correct memory order otherwise one has again undefined behavior but I am unsure if this is also required for sequentially consistent - maybe not.

elfenpiff (changes requested):

elfenpiff:
@JakubSosnovec thanks for your contribution. An you picked the right spot ... this innocent little line which brings one right into the dark labyrinth of the most complex C++ rules.

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
elBoberido: AFAIK, reordering across thread starts is not allowed

https://en.cppreference.com/w/cpp/thread/thread/thread
https://stackoverflow.com/questions/73001485/c-compiler-reordering-read-and-write-across-thread-start

elBoberido (commented):
@JakubSosnovec I had a quick look at the `GatewayGeneric` class. Could you also change the logic in the `shutdown` method? A more robust approach would be to unconditionally set `m_running` to `false` and the for each thread a
```Cpp
if(m_thread.joinable())
{
    m_thread.join();
}
```

With that change we could merge the PR.

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
JakubSosnovec: I see. Well, I am actually wondering whether the use of the `std::memory_order_relaxed` memory order for the `m_isRunning` field is really needed at all. Obviously, this is not a very hot code, and using the default `std::memory_order_seq_cst` might be much easier to reason about, while losing completely negligible performance (if at all). So, I would propose to just use the sequentially consistent memory order and avoiding any premature optimization pitfalls.

Or am I missing some detail?

elBoberido: @JakubSosnovec I'm fine either way. Using `std::memory_order_relaxed` should not be an issue and as you mentioned it is not a hot path so `std::memory_order_seq_cst` does not hurt. If it it changed, it should also be changed in `shutdown`, tough. The `while` loops should keep the current memory ordering.

JakubSosnovec: Alright, I updated it now.

elfenpiff (approved):

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2020 - 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2022 by NXP. All rights reserved.
```
elBoberido, line 3: @JakubSosnovec NXP, like the semiconductor company?

elBoberido (approved):
Thanks for your contribution and patience with us :)

iceoryx_posh/include/iceoryx_posh/internal/gateway/gateway_generic.inl
JakubSosnovec: That one, indeed :) we also do software to run on those semiconductors.

elBoberido: Cool. A few years ago I was also designing PCBs and used quite a lot of NXP products :)

JakubSosnovec: Very nice! Hopefully, they served you well :)

To be honest, working for a semiconductor vendor doesnt necessarily mean understanding anything about hardware design, which I am a bright example of 😄 .

Thanks a lot to you guys for keeping this stuff open source!

PR #1685, opened 2022-09-28
iox-#1655 Add `DeleterType` template parameter to `cxx::unique_ptr`

### LINKED ISSUE #1655: Easy misuse in `cxx::unique_ptr`
## Required information

Since the deleter is part of the constructor argument but otherwise unchangeable it is extremely hard to track which deleter is currently in use.

Do you spot the bug:
```cxx
cxx::unique_ptr<int> a(new int(1), [](auto*ptr) {delete ptr;});
cxx::unique_ptr<int> b(custom_alloc.new<int>(1), [](auto * ptr) { custom_alloc.free(ptr); });

a = std::move(b);
auto old_int_ptr = a.release(new int(2));
```

The deleter changes silently when `b` is moved to `a` but the user may assumes that `a` is still a unique_ptr which calls delete on destruction. This can be mixed up extremely easily!

The solution would be to make the deleter part of the `unique_ptr` type like the STL does. The deleter would be a callable type which releases the underlying resource and would look like:

```cxx
template<typename T>
struct CallDelete {
  operator()(T*ptr) { delete ptr; }
};

cxx::unique_ptr<int, CallDelete> a(new int(1));
cxx::unique_ptr<int, CustomDelete> b(custom_alloc.new<int>(1));

a = std::move(b); // compile failure ... bug from above cannot happen.

```

mossmaurice:
Signed-off-by: Simon Hoinkis <simon.hoinkis@apex.ai>

1. [x] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## Notes for Reviewer

* This prevents the misuse described in #1655 by introducing `DeleterType` as a 2nd template paramter to `cxx::unique_ptr`
  * This approach is similar to the one in the STL, a `DefaultDeleter` was not added on purpose as we don't use the heap and the user should instead explicitly write it's own deleter, which could also be `delete ptr`.

## References

- Closes #1655

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
@@ -24,16 +24,16 @@ namespace iox
 {
 namespace cxx
 {
-template <typename T>
-inline unique_ptr<T>::unique_ptr(T* const object, const function<void(T*)>& deleter) noexcept
+template <typename T, typename D>
+inline unique_ptr<T, D>::unique_ptr(T* const object, const function<D>& deleter) noexcept
     : m_ptr(object)
     , m_deleter(deleter)
 {
     Ensures(object != nullptr);
 }
 
-template <typename T>
-inline unique_ptr<T>& unique_ptr<T>::operator=(unique_ptr&& rhs) noexcept
+template <typename T, typename D>
+inline unique_ptr<T, D>& unique_ptr<T, D>::operator=(unique_ptr&& rhs) noexcept
 {
     if (this != &rhs)
     {
@@ -45,67 +45,67 @@ inline unique_ptr<T>& unique_ptr<T>::operator=(unique_ptr&& rhs) noexcept
     return *this;
 }
 
-template <typename T>
-inline unique_ptr<T>::unique_ptr(unique_ptr&& rhs) noexcept
+template <typename T, typename D>
+inline unique_ptr<T, D>::unique_ptr(unique_ptr&& rhs) noexcept
     : m_ptr{rhs.m_ptr}
     , m_deleter{std::move(rhs.m_deleter)}
 {
     rhs.m_ptr = nullptr;
 }
 
-template <typename T>
-inline unique_ptr<T>::~unique_ptr() noexcept
+template <typename T, typename D>
+inline unique_ptr<T, D>::~unique_ptr() noexcept
 {
     destroy();
 }
 
-template <typename T>
-inline T* unique_ptr<T>::operator->() noexcept
+template <typename T, typename D>
+inline T* unique_ptr<T, D>::operator->() noexcept
 {
     cxx::Expects(m_ptr != nullptr);
     return get();
 }
 
-template <typename T>
-inline const T* unique_ptr<T>::operator->() const noexcept
+template <typename T, typename D>
+inline const T* unique_ptr<T, D>::operator->() const noexcept
 {
     // Avoid code duplication
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
-    return const_cast<unique_ptr<T>*>(this)->operator->();
+    return const_cast<unique_ptr<T, D>*>(this)->operator->();
 }
 
-template <typename T>
-inline T* unique_ptr<T>::get() noexcept
+template <typename T, typename D>
+inline T* unique_ptr<T, D>::get() noexcept
 {
     return m_ptr;
 }
 
-template <typename T>
-inline const T* unique_ptr<T>::get() const noexcept
+template <typename T, typename D>
+inline const T* unique_ptr<T, D>::get() const noexcept
 {
     // AXIVION Next Construct AutosarC++19_03-A5.2.3 : Avoid code duplication
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
-    return const_cast<unique_ptr<T>*>(this)->get();
+    return const_cast<unique_ptr<T, D>*>(this)->get();
 }
 
-template <typename T>
-inline T* unique_ptr<T>::release(unique_ptr&& ptrToBeReleased) noexcept
+template <typename T, typename D>
+inline T* unique_ptr<T, D>::release(unique_ptr&& ptrToBeReleased) noexcept
 {
     auto ptr = ptrToBeReleased.m_ptr;
     ptrToBeReleased.m_ptr = nullptr;
     return ptr;
 }
 
-template <typename T>
-inline void unique_ptr<T>::reset(T* const ptr) noexcept
+template <typename T, typename D>
+inline void unique_ptr<T, D>::reset(T* const ptr) noexcept
 {
     Ensures(ptr != nullptr);
     destroy();
     m_ptr = ptr;
 }
 
-template <typename T>
-inline void unique_ptr<T>::destroy() noexcept
+template <typename T, typename D>
+inline void unique_ptr<T, D>::destroy() noexcept
 {
     if (m_ptr)
     {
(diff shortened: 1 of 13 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
```diff
@@ -45,67 +45,67 @@ inline unique_ptr<T>& unique_ptr<T>::operator=(unique_ptr&& rhs) noexcept
...
 {
     auto ptr = ptrToBeReleased.m_ptr;
     ptrToBeReleased.m_ptr = nullptr;
     return ptr;
 }
 
-template <typename T>
-inline void unique_ptr<T>::reset(T* const ptr) noexcept
+template <typename T, typename D>
+inline void unique_ptr<T, D>::reset(T* const ptr) noexcept
```
mossmaurice, line 100: @elBoberido Regarding [your comment](https://github.com/eclipse-iceoryx/iceoryx/issues/1655#issuecomment-1254918642): I don't think avoiding such a misuse makes sense. It would be over-the-top. The problem in your example are the `static_cast`s and not the API of `memory::unique_ptr`. The user is responsible for providing the correct types. Furthermore, users should not use the untyped API directly, that's something to be used inside of frameworks and not directly.

elfenpiff (changes requested):
Oh I think here went something wrong and I forgot it to communicate it clearly. But I was also not aware that you fix it like this.

Your approach does not fix the problem at all, one can easily trigger it again. If you would like to use a template approach then you have to follow the STL approach where the allocator/deallocator is stateless.
This would be unusable for us and restrict the `unique_ptr` since the allocator becomes a part of the type.

The only mitigation I see here is to add the deleter as an additional argument to `reset`. Then the user must always provide a valid deleter for its current pointer and it is impossible to imply by accident that the wrong deleter is set.

I would suggest the following here. Please drop your 3 commits and add the deleter as second argument to `reset` and the problem is successfully mitigated.

Sorry for the miscommunication from my side!

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/unique_ptr.inl
elBoberido: @mossmaurice it's not the `static_cast` since both of the types are the same. The problem is that the pointer must be released with a different object which is captured by the lambda. I used the casts since the `unique_ptr` cannot handle `void*` and I didn't want to create a different example. We have this situation in a lower layer though.

mossmaurice:
@elfenpiff @elBoberido Thanks for the review!

You are right, even if `T` and `D` have exactly the same type, having a `reset(T* ptr)` method may lead to situations where a `ptr` and `deleter` which don't belong together might be mixed as depicted in [this comment](https://github.com/eclipse-iceoryx/iceoryx/issues/1655#issuecomment-1254918642).
I don't know why the STL designed the API like this, but I agree that we should change this to avoid misuse.


What's currently implemented:

```
template <typename T, typename D> // No DefaultDeleter
class unique_ptr
{
  // .. snip ..
}

// Avoids assigning `cxx::unique_ptr`s with different `DeleterTypes`
auto ptr1 = iox::cxx::unique_ptr<Foo, DeleterType1>(&myObject1, myDeleter1);
auto ptr2 = iox::cxx::unique_ptr<Foo, DeleterType2>(&myObject2, myDeleter2);

ptr1 = std::move(ptr2);
```

However, this wouldn't be a problem as `myDeleter2` can be called before the move, if we use the same deleter signature for all `cxx::unique_ptr`.


My proposal going forward:

1. Keep the 2nd template parameter because it's more idomatic and gives the users more flexibility e.g. adding cv-qualifiers
1. Do not implement a `DefaultDeleter` to force users to explicitly reason about it (iceoryx is targeting heap-free development anyway, hence `delete ptr` wouldn't make much sense)

Two options for solving the remaining issue:
1. Remove `cxx::unique_ptr::reset(T* ptr)` and recommend users to use the move assignment operator instead
1. Change the signature to `cxx::unique_ptr::reset(cxx::unique_ptr&& ptr)`, which would then call `operator=(cxx::unique_ptr&& ptr)`

I would tend to got for option 2) for convenience of the users.

@ymarcov What are your thoughts on this topic? Do you have any other nifty ideas?

elfenpiff:
> My proposal going forward:
> 
> 1. Keep the 2nd template parameter because it's more idomatic and gives the users more flexibility e.g. adding cv-qualifiers

The deleter takes the ownership of the memory which the pointer is pointing to and the `unique_ptr` always provides a pointer to a non const type. What kind of cv-qualifier can you then set here? Make the pointer const (not part of the signature so irrelevant). Make the underlying type constant - compile failure.
I think I do not understand what you want to achieve here and why it should be idiomatic.

> 2. Do not implement a `DefaultDeleter` to force users to explicitly reason about it (iceoryx is targeting heap-free development anyway, hence `delete ptr` wouldn't make much sense)

I support this.

> Two options for solving the remaining issue:
> 
> 1. Remove `cxx::unique_ptr::reset(T* ptr)` and recommend users to use the move assignment operator instead

Possible, but why not the discussed solution where the user has to provide the new deleter to the new pointer as second argument. So `cxx::unique_ptr::reset(T* ptr, const DeleterCallback & deleter)`

> 2. Change the signature to `cxx::unique_ptr::reset(cxx::unique_ptr&& ptr)`, which would then call `operator=(cxx::unique_ptr&& ptr)`

Are you aware that this method is already implemented and called `release`?

ymarcov:
> @mossmaurice
> 
> > @ymarcov What are your thoughts on this topic? Do you have any other nifty ideas?
> 
> I think it's important to minimize the mixing of different styles in one codebase, and it may be that I have a different style and therefore it's not so relevant here, but in any case here are some thoughts:
> 
> 1. Some problems can be classified as "mistypes". If these can be avoided it's a good thing, but one should also look at the price of solving any particular problem in a certain way, i.e. the trade-off. For me if you reduce the likelihood of _mistypes_ for _one_ certain API call from 10% to 5% or maybe less, but you make 80% of all of its usages take on more boilerplate, I would probably not go for this trade-off, personally. You guys know better than me though if this is the case here or not.
> 2. Leaving it up to the user to figure out how to store and release a pointer you're providing him with is basically a violation of RAII. APIs should avoid returning "ownerless" handles in the first place. If generic types such as `unique_ptr`, which are provided for generic (i.e. not specialized) use cases, do not handle the subtleties of the use case very well, then some kind of a proper type should be provided that handles ownership for a special handle. This type should be simple, straightforward, and thoroughly unit-tested, which takes care of its correctness. Then, all usages of this type should work seamlessly, without requiring extra alertness, nor boilerplate or formalisms, from the user of the API.
> 
> Just my 2 cents.

@mossmaurice 

As a side note, I'm not too big on ambiguity in APIs. If you have a type called `unique_ptr`, it should be as identical as possible (if not fully identical) to the STL, where nowadays it is one of the most foundational and frequently used building blocks. If anything is asking for trouble and mistypes, it is having different subtle characteristics for an identically named type such as this.

elfenpiff:
@ymarcov @mossmaurice 

> As a side note, I'm not a fan of ambiguity in APIs.

In general you are right and we started with the idea that you could always transfer your code from the STL to iceoryx and it would still compile. After a while we saw all the little weaknesses in the API which could lead to undefined behavior or easy misuse and then we started to adjust some minor things to solve those problems. Also getting rid of exceptions or heap leads inevitable to API changes.

Let's stay at this particular example. You come from the STL and would like to use iceoryx you will encounter some compile time failures.
1. The unique_ptr is no longer nullable. But then you will suddenly get rid of all the bugs throughout the code where you accessed an empty unique_ptr. Furthermore, no more null checks so your code becomes faster and less complex. If you would like to have a nullable unique_ptr you can always use the optional.
2. We somehow implemented the polymorphic allocator idea in a simplistic way with the deleter callable. Yes the user has to think about where to release it's resource, but since we are aiming for the safety critical use case a default heap deleter makes no sense. The memory is coming from some custom deleter like shared memory, publisher or for dynamic types from an internal allocator.

We are clearly aiming for the safety critical case and do not intend to be used as generic STL. So our paradigm is a safe API first and then take care of performance (but do not get me wrong, performance is still an important issue).
With the STL performance is the highest priority which can bring some serious drawbacks. Some examples:
1. `std::optional` does not reset its state after move but `std::unique_ptr` does, so in one class `if ( myOptional )` is undefined but in `if ( myUniqueptr)` it is.
2. `std::vector` iterators are valid after move but invalidated when a `std::string` is moved
3. Our just take a look at the condition variable or mutex where other languages already have much safer concepts which we already adapted in our hoofs library (see smart_lock for instance which behaves like a rust mutex)

We try to be as consistent as possible on all levels and restrict the API maybe a bit but the user gains a safer uniform usage from it - sometimes even with performance increase. And we are defined! Move means that the source is in a defined state and this goes for so many more API adjustments.

But as with every API, it is not perfect and sometimes our changes open the door for another flood of bugs which we didn't think of so I appreciate your input very much! We have to always challenge our ideas and concepts.

mossmaurice:
@ymarcov Thanks for sharing your thoughts! :pray: 

@elfenpiff Re

> What kind of cv-qualifier can you then set here? 

Although, being the owner sometimes `unique_ptr`s are not allowed to change the memory. For example in the `Subscriber`:

https://github.com/eclipse-iceoryx/iceoryx/blob/5870d172746a93460f59f34690fb9bb6f286dd73/iceoryx_posh/include/iceoryx_posh/internal/popo/subscriber_impl.inl#L44-L45

> Are you aware that this method is already implemented and called `release`?

`release` is not the same as `reset`. `release` just releases the ownership and does not call the deleter, whereas `reset` calls the deleter and then replaces the managed object (like `operator=(cxx::unique_ptr&& ptr)`).

Conclusion:

1. Only use one template param `T`, but derive cv-qualifier for it with `add_const_conditionally_t` from `T`
2. Remove `reset` and recommend users to use the move assign operator instead

Closing this PR as it is super-seeded by #1707

PR #1698, opened 2022-09-29
Iox 1394 fix axivion violations in cxx function

MatthiasKillat:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #1394 (partial)

### DIFF
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
@@ -45,7 +45,7 @@ class storable_function;
 /// @tparam ReturnType  The return type of the stored callable.
 /// @tparam Args        The arguments of the stored callable.
 template <uint64_t Capacity, typename ReturnType, typename... Args>
-class storable_function<Capacity, signature<ReturnType, Args...>>
+class storable_function<Capacity, signature<ReturnType, Args...>> final
 {
   public:
     using StorageType = static_storage<Capacity>;
@@ -56,6 +56,7 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
               typename = typename std::enable_if<std::is_class<Functor>::value
                                                      && is_invocable_r<ReturnType, Functor, Args...>::value,
                                                  void>::type>
+    // AXIVION Next Construct AutosarC++19_03-A12.1.4: implicit conversion of functors is intentional
     /// @NOLINTJUSTIFICATION the storable function should implicitly behave like any generic constructor, adding
     ///                      explicit would require a static_cast. Furthermore, the storable_functor stores a copy
     ///                      which avoids implicit misbehaviors or ownership problems caused by implicit conversion.
@@ -93,19 +94,13 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
     /// @param args arguments to invoke the stored function with
     /// @return return value of the stored function
     ///
-    /// @note 1) Invoking the function if there is no stored function (i.e. operator bool returns false)
-    ///          leads to terminate being called.
-    ///
-    ///       2) Deliberately not noexcept but can only throw if the stored callable can throw an exception (hence
-    ///          will never throw if we use only our own noexcept functions).
-    ///
-    ///       3) If arguments are passed by value, the copy constructor may be invoked twice:
+    /// @note 1) If arguments are passed by value, the copy constructor may be invoked twice:
     ///          once when passing the arguments to operator() and once when they are passed to the stored callable
     ///          itself. This appears to be unavoidable and also happens in std::function.
     ///          The user can always provide a wrapped callable which takes a reference,
     ///          which is generally preferable for large objects anyway.
     ///
-    ///       4) Arguments of class type cannot have the move constructor explicitly deleted since the arguments
+    ///       2) Arguments of class type cannot have the move constructor explicitly deleted since the arguments
     ///          must be forwarded internally which is done by move or, if no move is specified, by copy.
     ///          If the move operation is explicitly deleted the compiler will not fall back to copy but emit an error.
     ///          Not specifying move or using a default implementation is fine.
@@ -136,7 +131,7 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
     // This means storable_function cannot be used where pointers become invalid, e.g. across process boundaries
     // Therefore we cannot store a storable_function in shared memory (the same holds for std::function).
     // This is inherent to the type erasure technique we (have to) use.
-    struct operations
+    struct operations final
     {
         // function pointers defining copy, move and destroy semantics
         void (*copyFunction)(const storable_function& src, storable_function& dest){nullptr};
@@ -150,11 +145,11 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
         operations& operator=(operations&& other) noexcept = default;
         ~operations() = default;
 
-        void copy(const storable_function& src, storable_function& dest) noexcept;
+        void copy(const storable_function& src, storable_function& dest) const noexcept;
 
-        void move(storable_function& src, storable_function& dest) noexcept;
+        void move(storable_function& src, storable_function& dest) const noexcept;
 
-        void destroy(storable_function& f) noexcept;
+        void destroy(storable_function& f) const noexcept;
     };
 
   private:
@@ -187,12 +182,15 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
     static void destroy(storable_function& f) noexcept;
 
     template <typename CallableType>
-    static ReturnType invoke(void* callable, Args&&... args);
+    static ReturnType invoke(void* callable, Args&&... args) noexcept;
 
     static void copyFreeFunction(const storable_function& src, storable_function& dest) noexcept;
 
     static void moveFreeFunction(storable_function& src, storable_function& dest) noexcept;
 
+    // AXIVION Next Construct AutosarC++19_03-M7.1.2: callable cannot be const void* since
+    // m_invoker is initialized with this function and has to work with functors as well
+    // (functors may change due to invocation)
     static ReturnType invokeFreeFunction(void* callable, Args&&... args);
 };
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
@@ -56,10 +56,10 @@ class storable_function<Capacity, signature<ReturnType, Args...>> final
               typename = typename std::enable_if<std::is_class<Functor>::value
                                                      && is_invocable_r<ReturnType, Functor, Args...>::value,
                                                  void>::type>
-    // AXIVION Next Construct AutosarC++19_03-A12.1.4: implicit conversion of functors is intentional
-    /// @NOLINTJUSTIFICATION the storable function should implicitly behave like any generic constructor, adding
-    ///                      explicit would require a static_cast. Furthermore, the storable_functor stores a copy
-    ///                      which avoids implicit misbehaviors or ownership problems caused by implicit conversion.
+    // AXIVION Next Construct AutosarC++19_03-A12.1.4: implicit conversion of functors is intentional,
+    // the storable function should implicitly behave like any generic constructor, adding
+    // explicit would require a static_cast. Furthermore, the storable_functor stores a copy
+    // which avoids implicit misbehaviors or ownership problems caused by implicit conversion.
     /// @NOLINTNEXTLINE(hicpp-explicit-conversions)
     storable_function(const Functor& functor) noexcept;
 
@@ -191,7 +191,7 @@ class storable_function<Capacity, signature<ReturnType, Args...>> final
     // AXIVION Next Construct AutosarC++19_03-M7.1.2: callable cannot be const void* since
     // m_invoker is initialized with this function and has to work with functors as well
     // (functors may change due to invocation)
-    static ReturnType invokeFreeFunction(void* callable, Args&&... args);
+    static ReturnType invokeFreeFunction(void* callable, Args&&... args) noexcept;
 };
 
 /// @brief swap two storable functions
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
@@ -35,15 +35,15 @@ inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_fun
 
 // AXIVION Next Construct AutosarC++19_03-A12.1.5: constructor delegation is not feasible here due
 // to lack of sufficient common initialization
-// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are default initialized
+// AXIVION Next Construct AutosarC++19_03-M5.2.6: the converted pointer is only used
+// as its original function pointer type after reconversion (type erasure)
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(
     ReturnType (*function)(Args...)) noexcept
-    : // AXIVION Next Construct AutosarC++19_03-M5.2.6: the converted pointer is only used
-      // as its original function pointer type after reconversion (type erasure)
-      // AXIVION Next Construct AutosarC++19_03-A5.2.4: reinterpret_cast is required for type erasure
-      /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
-      ///                      the correct type whenever the callable is used
+    : // AXIVION Next Construct AutosarC++19_03-A5.2.4: reinterpret_cast is required for type erasure,
+      // we use type erasure in combination with compile time template arguments to restore
+      // the correct type whenever the callable is used
       /// @NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
     m_callable(reinterpret_cast<void*>(function))
     , m_invoker(invokeFreeFunction)
@@ -55,7 +55,7 @@ inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_fun
     // destroy is not needed for free functions
 }
 
-// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are default initialized
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename T, typename>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(
@@ -68,7 +68,7 @@ inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_fun
     storeFunctor(functor);
 }
 
-// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are default initialized
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename T, typename>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(const T& object,
@@ -92,6 +92,7 @@ inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_fun
     m_operations.copy(other, *this);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: m_storage is default initialized
 // AXIVION Next Construct AutosarC++19_03-A12.8.4: we copy only the operation pointer table
 // (required) and will perform a move with its type erased move function
 template <uint64_t Capacity, typename ReturnType, typename... Args>
@@ -265,7 +266,6 @@ inline void
 storable_function<Capacity, signature<ReturnType, Args...>>::moveFreeFunction(storable_function& src,
                                                                               storable_function& dest) noexcept
 {
-    // AXIVION Next Construct AutosarC++19_03-A18.9.2: false positive, only a pointer is copied
     dest.m_invoker = src.m_invoker;
     dest.m_callable = src.m_callable;
     src.m_invoker = nullptr;
@@ -279,7 +279,7 @@ template <typename CallableType>
 inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invoke(void* callable,
                                                                                       Args&&... args) noexcept
 {
-    // AXIVION Next Construct AutosarC++19_03-A18.9.2: false positive, perfect forwarding
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2: we use idiomatic perfect forwarding
     // AXIVION Next Construct AutosarC++19_03-A5.3.2: callable is guaranteed not to be nullptr
     // when invoke is called (it is private and only used for type erasure)
     // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
@@ -291,15 +291,16 @@ inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::i
 // m_invoker is initialized with this function and has to work with functors as well
 // (functors may change due to invocation)
 template <uint64_t Capacity, typename ReturnType, typename... Args>
-inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invokeFreeFunction(void* callable,
-                                                                                                  Args&&... args)
+inline ReturnType
+storable_function<Capacity, signature<ReturnType, Args...>>::invokeFreeFunction(void* callable, Args&&... args) noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2: we use idiomatic perfect forwarding
     // AXIVION Next Construct AutosarC++19_03-A5.3.2: callable is guaranteed not to be nullptr
     // when invokeFreeFunction is called (it is private and only used for type erasure)
     // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
     // AXIVION Next Construct AutosarC++19_03-A5.2.4: reinterpret_cast is required for type erasure
-    /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
-    ///                      the correct type whenever the callable is used
+    // type erasure in combination with compile time template arguments to restore the correct type
+    // when the callable is called
     /// @NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
     return (reinterpret_cast<ReturnType (*)(Args...)>(callable))(std::forward<Args>(args)...);
 }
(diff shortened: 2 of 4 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -93,19 +94,13 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
     /// @param args arguments to invoke the stored function with
     /// @return return value of the stored function
     ///
-    /// @note 1) Invoking the function if there is no stored function (i.e. operator bool returns false)
```
MatthiasKillat, line 96: This was incorrect. 
1. It will now not terminate anymore since it cannot be null by design (except after move, but this is illegal).

2. It now has noexcept (already had it) but actually should not since the callable we store can happily throw. This is an issue in all our template code IMO.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -25,50 +25,64 @@ namespace iox
...
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename Functor, typename>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(const Functor& functor) noexcept
 {
     storeFunctor(functor);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A12.1.5: constructor delegation is not feasible here due
+// to lack of sufficient common initialization
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
```
MatthiasKillat, line 38: This rule is questionable. It cannot deal properly with default initialization and (as is the case here as well) with the fact that something will be properly initialized before use.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -25,50 +25,64 @@ namespace iox
...
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(
     ReturnType (*function)(Args...)) noexcept
-    : /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
+    : // AXIVION Next Construct AutosarC++19_03-M5.2.6: the converted pointer is only used
```
MatthiasKillat, line 42: Many many findings are due to the fact that generic type erasure requires low level pointer casts and Axivion complains. However, they are fine and portable if the object is only accessed through a pointer of its actual type later.

This happens here, it is just not possible to prove for Axivion but to some degree, type erasure cannot avoid this so we need those comments.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -25,50 +25,64 @@ namespace iox
...
       ///                      the correct type whenever the callable is used
       /// @NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
     m_callable(reinterpret_cast<void*>(function))
     , m_invoker(invokeFreeFunction)
 {
     cxx::Expects(function);
 
-    m_operations.copyFunction = copyFreeFunction;
-    m_operations.moveFunction = moveFreeFunction;
+    m_operations.copyFunction = &copyFreeFunction;
```
MatthiasKillat, line 53: A questionable rule. A function name is no function pointer but it implicitly converts to one (mandated by standard). There is no other interpretation when assigning a function to a function pointer (signature is checked). The `&` is superfluous as is the rule IMO.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -25,50 +25,64 @@ namespace iox
...
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename T, typename>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(
     T& object, ReturnType (T::*method)(Args...)) noexcept
 {
-    auto p = &object;
+    const auto p = &object;
+    // AXIVION Next Construct AutosarC++19_03-A7.1.1: type erased functor lambda cannot be const
+    // as it may change by calling its operator() later (hence stored as non-const)
```
MatthiasKillat, line 66: Will be stored in a non-const object, so this would not work. I am fairly sure it is not easy to fix for all cases at once, again due to type erasure. We cannot store the object in a `const void*` later, and even if we could, this would be lying as invoking anything on it later may change the stored object.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -78,6 +92,8 @@ inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_fun
     m_operations.copy(other, *this);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A12.8.4: we copy only the operation pointer table
```
MatthiasKillat, line 95: The move is tricky due to type erasure. We need to copy a few function pointers (moving them is wrong and I think leads to complaints as well). The actual move is then performed according to one of the function pointers.

Essentially a `vtable` dependent on the stored object and not the class.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -153,18 +172,21 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::storeFu
     auto ptr = m_storage.template allocate<StoredType>();
     cxx::Expects(ptr != nullptr);
 
+    // AXIVION Next Construct AutosarC++19_03-A18.5.10: false positive, m_storage.allocate<T>
```
MatthiasKillat, line 175: Pointer trickery with invariants not known to Axivion. Requires semantic understanding that allocate provides an aligned pointer for `StoredType` which is non-trivial to infer automatically.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -153,18 +172,21 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::storeFu
...
-    m_operations.copyFunction = copy<StoredType>;
-    m_operations.moveFunction = move<StoredType>;
-    m_operations.destroyFunction = destroy<StoredType>;
+    m_invoker = &invoke<StoredType>;
+    m_operations.copyFunction = &copy<StoredType>;
+    m_operations.moveFunction = &move<StoredType>;
+    m_operations.destroyFunction = &destroy<StoredType>;
 }
 
+// AXIVION Next Construct AutosarC++19_03-A8.4.8: output parameter required by design and for efficiency
```
MatthiasKillat, line 189: Not a good rule. Disabling candidate.

Output parameters have their place, and here it is more that the intention is to modify one or both input objects which previously exist and hence should not be constructed and returned by the function. This is due to design and efficiency reasons.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -173,14 +195,21 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::copy(co
     auto ptr = dest.m_storage.template allocate<CallableType>();
     cxx::Expects(ptr != nullptr);
 
-    auto obj = static_cast<CallableType*>(src.m_callable);
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
+    const auto obj = static_cast<CallableType*>(src.m_callable);
+
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: obj is guaranteed not to be
```
MatthiasKillat, line 201: Torn on this rule. Of course we do not want to derefernce a `nullptr`. 
However, this is often guaranteed by invariants, e.g. we know `m_callable` is not null by design (except for moved from objects, which are illegal to access). Axivion does often fails to infer this, which is understandable as it is a non-trivial problem.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -204,9 +239,11 @@ template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename CallableType>
 inline void storable_function<Capacity, signature<ReturnType, Args...>>::destroy(storable_function& f) noexcept
 {
-    if (f.m_callable)
+    if (f.m_callable != nullptr)
```
MatthiasKillat, line 242: Being explicit here may be useful, but IMO should be optional. The meaning is well-defined to my knowledge.

MatthiasKillat:
Dear future reviewers. I took the liberty to comment on some rules and add a little more reasoning why I think the suppression comments are necessary.

Long story short, it is mainly due to Axivion having difficulties with low level pointer casts required for type erasure which should be perfectly safe here (potential bugs excluded). Also, I do not see an easy way around deferring some initialization to the ctor body here, which should be fine.

I commented on some rules I consider problematic or superfluous for various reasons.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -56,6 +56,7 @@ class storable_function<Capacity, signature<ReturnType, Args...>>
               typename = typename std::enable_if<std::is_class<Functor>::value
                                                      && is_invocable_r<ReturnType, Functor, Args...>::value,
                                                  void>::type>
+    // AXIVION Next Construct AutosarC++19_03-A12.1.4: implicit conversion of functors is intentional
```
FerdinandSpitzschnueffler, line 59: I would prefer if you replace `@NOLINTJUSTIFICATION` in the next line with `AXIVION Next Construct AutosarC++19_03-A12.1.4 :` and reuse the justification.

FerdinandSpitzschnueffler (commented):
Review not finished

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -25,50 +25,64 @@ namespace iox
...
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 inline storable_function<Capacity, signature<ReturnType, Args...>>::storable_function(
     ReturnType (*function)(Args...)) noexcept
-    : /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
+    : // AXIVION Next Construct AutosarC++19_03-M5.2.6: the converted pointer is only used
+      // as its original function pointer type after reconversion (type erasure)
+      // AXIVION Next Construct AutosarC++19_03-A5.2.4: reinterpret_cast is required for type erasure
+      /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
       ///                      the correct type whenever the callable is used
```
FerdinandSpitzschnueffler, line 46: The same justification can be found for the Axivion warning one line above. So please remove this one here. For readability I would suggest to move the Axivion suppressions above the c'tor.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -221,28 +258,46 @@ storable_function<Capacity, signature<ReturnType, Args...>>::copyFreeFunction(co
...
 }
 
+// AXIVION Next Construct AutosarC++19_03-A8.4.4, AutosarC++19_03-A8.4.8: output parameter required by design and for
+// efficiency
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 inline void
 storable_function<Capacity, signature<ReturnType, Args...>>::moveFreeFunction(storable_function& src,
                                                                               storable_function& dest) noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2: false positive, only a pointer is copied
```
FerdinandSpitzschnueffler, line 268: I couldn't find this violation in the Axivion scan. Can you please check again? If we have to keep it, can `Construct` be replaced with `Line`?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -221,28 +258,46 @@ storable_function<Capacity, signature<ReturnType, Args...>>::copyFreeFunction(co
...
 
+// AXIVION Next Construct AutosarC++19_03-M7.1.2: callable cannot be const void* since
+// m_invoker is initialized with this function and has to work with functors as well
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename CallableType>
-inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invoke(void* callable, Args&&... args)
+inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invoke(void* callable,
+                                                                                      Args&&... args) noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2: false positive, perfect forwarding
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: `std::forward` is even recommended by the rule...

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -221,28 +258,46 @@ storable_function<Capacity, signature<ReturnType, Args...>>::copyFreeFunction(co
...
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invokeFreeFunction(void* callable,
                                                                                                   Args&&... args)
 {
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: callable is guaranteed not to be nullptr
+    // when invokeFreeFunction is called (it is private and only used for type erasure)
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4: reinterpret_cast is required for type erasure
     /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
     ///                      the correct type whenever the callable is used
```
FerdinandSpitzschnueffler, line 302: Can be removed since this already stated in the lines above.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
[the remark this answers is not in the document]
MatthiasKillat: But it still complains. I do not get it, but to my knowledge this is idiomatic perfect forwarding. `std::forward` exists precisely for this. Without going into details about reference collapsing and so on, its purpose is to chose the correct "cast" automatically, which could have the same effect as `std::move` in specific circumstances but we do not need to care about it.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: Personally I would avoid using Next Line always, because of future comments. One can argue the same for code and Next Construct, but I think this is less lf an issue.

Sometimes I would like a macro like `JUSTIFY(rule number, justification)` which expands to  what we need and affects the next construct (only for Axivion).

MatthiasKillat:
@FerdinandSpitzschnueffler Fixed what was missing, may have missed something. Local scan is inconclusive on its own.

FerdinandSpitzschnueffler (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -125,16 +142,19 @@ inline storable_function<Capacity, signature<ReturnType, Args...>>::~storable_fu
     m_operations.destroy(*this);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A2.10.1: false positive, args does not hide anything
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::operator()(Args... args) const noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-M0.3.1: m_invoker is initialized in ctor or assignment,
+    // can only be nullptr if this was moved from (calling operator() is illegal in this case)
```
mossmaurice, line 150: Previously a `cxx::Expects(!empty())` was here. I think it's still needed here as a user could invoke a moved-from `cxx::function` and then `m_invoker` would be a nullptr. Can you please re-add it?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
mossmaurice: @MatthiasKillat I disagree here. I think `A5.3.2` is there for a reason. A user could run into the issue of using a moved-from object and then copying it, hence we should check for `nullptr` here with Expects/assert.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -189,7 +219,13 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::move(st
     auto ptr = dest.m_storage.template allocate<CallableType>();
     cxx::Expects(ptr != nullptr);
 
-    auto obj = static_cast<CallableType*>(src.m_callable);
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
+    const auto obj = static_cast<CallableType*>(src.m_callable);
+
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: obj is guaranteed not to be nullptr
```
mossmaurice, line 225: See above. A `nullptr` check should be added.

mossmaurice (changes requested):
Thanks for the PR, LGTM in general! I think it's important to adhere to the rule `AutosarC++19_03-A5.3.2`, checking pointers for `nullptr` before using them is a best practise and in the current state can be violated by a user when using a moved-from object. Please remove all suppressions for this rule and replace them with an `Expects` check.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -25,50 +25,64 @@ namespace iox
 {
 namespace cxx
 {
+// AXIVION Next Construct AutosarC++19_03-A12.6.1: members are initialized in body before read access
```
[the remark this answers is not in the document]
MatthiasKillat: No file wide suppression, this is just masking a tool issue (problem with detection of initialized values). What a ctor is added and the tool is right but the rule is suppressed file-wide?

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: We can add an an assert here but no always enabled runtime check for those kind of misuse.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: Yes but then much of the advantage of non-nullability goes right out of the window (being able to assume it is not null).

Note that I do not agree to have this construct not be null in the first place, as it is useful in some designs where we want to store a null function (this better should have been an extension/separate implementation). This is another issue, however.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
mossmaurice: @MatthiasKillat Bad design and mistakes to use a moved-from object can always occur, but this is not acceptable in a safety-critical software! I know this might have a performance impact, but as you can read in the following post even the C++ community (here LLVM) is moving in this direction: https://discourse.llvm.org/t/rfc-c-buffer-hardening/65734

But you are completely right, this is the perfect place for a `Expects` (later `IOX_ASSERT`). The user can then decide if to compile out the checks in release mode or keep them active (similar to what the LLVM folks are planning)-

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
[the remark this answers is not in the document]
mossmaurice: @MatthiasKillat Ok, no hard feelings here. Up to you.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -173,14 +197,22 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::copy(co
     auto ptr = dest.m_storage.template allocate<CallableType>();
     cxx::Expects(ptr != nullptr);
 
-    auto obj = static_cast<CallableType*>(src.m_callable);
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
+    const auto obj = static_cast<CallableType*>(src.m_callable);
+    cxx::Expects(obj != nullptr); // should not happen unless src is incorrectly used after move
+
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: obj is guaranteed not to be
+    // nullptr unless src was invalidated by move (illegal use of src)
```
FerdinandSpitzschnueffler, line 205: Please check if this suppression can be removed now that you add the check in line 202.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -173,14 +197,22 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::copy(co
...
+    // AXIVION Next Construct AutosarC++19_03-A18.5.10: false positive, m_storage.allocate<T>
+    // provides a properly aligned ptr and sufficient memory (static_storage satisfies this)
     /// @NOLINTJUSTIFICATION ownership is encapsulated in storable_function and will be released in destroy
     /// @NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
     ptr = new (ptr) CallableType(*obj);
     dest.m_callable = ptr;
     dest.m_invoker = src.m_invoker;
 }
-
+// AXIVION Next Construct AutosarC++19_03-A8.4.4, AutosarC++19_03-A8.4.8: output parameter required by design and for
```
FerdinandSpitzschnueffler, line 214:
```suggestion

// AXIVION Next Construct AutosarC++19_03-A8.4.4, AutosarC++19_03-A8.4.8: output parameter required by design and for
```

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -189,7 +221,14 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::move(st
     auto ptr = dest.m_storage.template allocate<CallableType>();
     cxx::Expects(ptr != nullptr);
 
-    auto obj = static_cast<CallableType*>(src.m_callable);
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
+    const auto obj = static_cast<CallableType*>(src.m_callable);
+    cxx::Expects(obj != nullptr); // should not happen unless src is incorrectly used after move
+
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: obj is guaranteed not to be nullptr
+    // unless src was invalidated by move (illegal use of src)
```
FerdinandSpitzschnueffler, line 229: Please check if the suppression is still needed.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: Safety critical software needs more careful design to prevent the mistakes in the first place but is no excuse for inefficient code. It is not safer just because of additional checks for conditions that cannot happen in regular use (such conditions always exist one way or another for moderately complex systems).

However, I am all for asserts to state intention/assumptions.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: According to local scan it is not. Removed.

Once we replace the always existing runtime check with an a compile time dependent assert it will probably be needed again (depending on the compile flag ...).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
MatthiasKillat: According to local scan it is not. Removed.

FerdinandSpitzschnueffler (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -232,19 +272,35 @@ storable_function<Capacity, signature<ReturnType, Args...>>::moveFreeFunction(st
...
-inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invokeFreeFunction(void* callable,
-                                                                                                  Args&&... args)
+inline ReturnType
+storable_function<Capacity, signature<ReturnType, Args...>>::invokeFreeFunction(void* callable, Args&&... args) noexcept
 {
-    /// @NOLINTJUSTIFICATION we use type erasure in combination with compile time template arguments to restore
-    ///                      the correct type whenever the callable is used
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2: we use idiomatic perfect forwarding
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: callable is guaranteed not to be nullptr
+    // when invokeFreeFunction is called (it is private and only used for type erasure)
```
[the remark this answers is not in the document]
MatthiasKillat: The private function is only called if this is the case.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -232,19 +272,35 @@ storable_function<Capacity, signature<ReturnType, Args...>>::moveFreeFunction(st
...
+// AXIVION Next Construct AutosarC++19_03-M7.1.2: callable cannot be const void* since
+// m_invoker is initialized with this function and has to work with functors as well
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename CallableType>
-inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invoke(void* callable, Args&&... args)
+inline ReturnType storable_function<Capacity, signature<ReturnType, Args...>>::invoke(void* callable,
+                                                                                      Args&&... args) noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A18.9.2: we use idiomatic perfect forwarding
+    // AXIVION Next Construct AutosarC++19_03-A5.3.2: callable is guaranteed not to be nullptr
```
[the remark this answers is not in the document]
MatthiasKillat: Yes, it is checked before invocation with an assert.

PR #1731, opened 2022-10-12
iox-#1725 Add refcounting to the PoshRuntime

### LINKED ISSUE #1725: Nifty counter for PoshRuntimeImpl
## Brief feature description

The `PoshRuntimeImpl` is a static variable (Meyer's singleton). When used with Cyclone DDS, sometimes (unfortunately) there are other static objects containing a DDS entity (e.g. domain participant), which on destruction uses the `PoshRuntime`. The destruction order of these static variables is impossible to predict, so it would be good to have a way to extend the lifetime of the `PoshRuntimeImpl` with reference-counting until its last use, as a short-term fix.

In the long term, a better solution might be to eliminate this singleton altogether, but that's not the focus of this issue.

## Detailed information

It was proposed by @elfenpiff to use the [nifty counter idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Nifty_Counter) to implement refcounting for `PoshRuntimeImpl`.

nnmm:
This uses manual creation and destruction of the runtime, like in the nifty counter idiom. However, unlike the nifty counter idiom, the scope guard object is not a global static variable, but a local static, so that the factory can be set before the runtime is created.

1. [X] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [X] Branch follows the naming format (`iox-123-this-is-a-branch`)
1. [X] Commits messages are according to this [guideline][commit-guidelines]
    - [X] Commit messages have the issue ID (`iox-#123 commit text`)
    - [X] Commit messages are signed (`git commit -s`)
1. [X] Update the PR title
   - Link to the relevant issue
1. [X] Relevant issues are linked
1. [X] All touched (C/C++) source code files from `iceoryx_hoofs` are added to `./clang-tidy-diff-scans.txt`
1. [ ] Assign PR to reviewer

## References

- Closes #1725

### DIFF
--- a/iceoryx_posh/source/runtime/posh_runtime.cpp
+++ b/iceoryx_posh/source/runtime/posh_runtime.cpp
@@ -21,11 +21,20 @@
 #include "iceoryx_posh/internal/runtime/posh_runtime_impl.hpp"
 
 #include <cstdint>
+#include <new>
+#include <type_traits>
 
 namespace iox
 {
 namespace runtime
 {
+
+// A refcount for use in getLifetimeParticipant().
+static int refcount;
+// Track whether the refcount lifetime mechanism is used by this particular
+// implementation of the abstract PoshRuntime class.
+static bool manual_lifetime_management = false;
+
 PoshRuntime::factory_t& PoshRuntime::getRuntimeFactory() noexcept
 {
     static factory_t runtimeFactory = PoshRuntimeImpl::defaultRuntimeFactory;
@@ -47,8 +56,13 @@ void PoshRuntime::setRuntimeFactory(const factory_t& factory) noexcept
 
 PoshRuntime& PoshRuntime::defaultRuntimeFactory(cxx::optional<const RuntimeName_t*> name) noexcept
 {
-    static PoshRuntimeImpl instance(name);
-    return instance;
+    static typename std::aligned_storage<sizeof(PoshRuntimeImpl), alignof(PoshRuntimeImpl)>::type buf;
+    static cxx::ScopeGuard ltp = [](auto name){
+        new(&buf) PoshRuntimeImpl(name);
+        manual_lifetime_management = true;
+        return getLifetimeParticipant();
+    }(name);
+    return reinterpret_cast<PoshRuntimeImpl&>(buf);
 }
 
 // singleton access
@@ -65,6 +79,20 @@ PoshRuntime& PoshRuntime::initRuntime(const RuntimeName_t& name) noexcept
 PoshRuntime& PoshRuntime::getInstance(cxx::optional<const RuntimeName_t*> name) noexcept
 {
     return getRuntimeFactory()(name);
+}
+
+cxx::ScopeGuard PoshRuntime::getLifetimeParticipant() noexcept
+{
+    return cxx::ScopeGuard(
+        [](){
+            ++refcount;
+        },
+        [](){
+            if (0 == --refcount && manual_lifetime_management) {
+                getInstance().~PoshRuntime();
+            }
+        }
+    );
 }
 
 PoshRuntime::PoshRuntime(cxx::optional<const RuntimeName_t*> name) noexcept
--- a/iceoryx_posh/source/runtime/posh_runtime.cpp
+++ b/iceoryx_posh/source/runtime/posh_runtime.cpp
@@ -34,7 +34,8 @@ namespace runtime
 static std::atomic<uint64_t> s_refcount{0U};
 // Track whether the refcount lifetime mechanism is used by this particular
 // implementation of the abstract PoshRuntime class.
-static std::atomic<bool> manual_lifetime_management{false};
+// This is necessary to avoid calling the destructor twice when
+static std::atomic<bool> s_manualLifetimeManagement{false};
 
 PoshRuntime::factory_t& PoshRuntime::getRuntimeFactory() noexcept
 {
@@ -60,7 +61,7 @@ PoshRuntime& PoshRuntime::defaultRuntimeFactory(cxx::optional<const RuntimeName_
     static typename std::aligned_storage<sizeof(PoshRuntimeImpl), alignof(PoshRuntimeImpl)>::type buf;
     static cxx::ScopeGuard ltp = [&buf](auto name) {
         new (&buf) PoshRuntimeImpl(name);
-        manual_lifetime_management = true;
+        s_manualLifetimeManagement = true;
         return getLifetimeParticipant();
     }(name);
     return reinterpret_cast<PoshRuntimeImpl&>(buf);
@@ -86,7 +87,7 @@ cxx::ScopeGuard PoshRuntime::getLifetimeParticipant() noexcept
 {
     return cxx::ScopeGuard([]() { ++s_refcount; },
                            []() {
-                               if (0 == --s_refcount && manual_lifetime_management)
+                               if (0 == --s_refcount && s_manualLifetimeManagement)
                                {
                                    getInstance().~PoshRuntime();
                                }
--- a/iceoryx_posh/source/runtime/posh_runtime.cpp
+++ b/iceoryx_posh/source/runtime/posh_runtime.cpp
@@ -20,6 +20,7 @@
 #include "iceoryx_posh/internal/log/posh_logging.hpp"
 #include "iceoryx_posh/internal/runtime/posh_runtime_impl.hpp"
 
+#include <atomic>
 #include <cstdint>
 #include <new>
 #include <type_traits>
@@ -30,10 +31,10 @@ namespace runtime
 {
 
 // A refcount for use in getLifetimeParticipant().
-static int refcount;
+static std::atomic<uint64_t> s_refcount{0U};
 // Track whether the refcount lifetime mechanism is used by this particular
 // implementation of the abstract PoshRuntime class.
-static bool manual_lifetime_management = false;
+static std::atomic<bool> manual_lifetime_management{false};
 
 PoshRuntime::factory_t& PoshRuntime::getRuntimeFactory() noexcept
 {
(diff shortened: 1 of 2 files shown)


nnmm:
@elBoberido @elfenpiff What do you think of this approach?

If this approach is fine it, does this need tests? I think it might be quite difficult to do, since it deals with life past main, but GTest probably finishes before the end of main.

Anything else missing?

elfenpiff:
@nnmm Thanks for the contribution. Nicely done!

But you need to sign the ECA: https://www.eclipse.org/legal/ECA.php with the e-mail address you used in your commits.

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -21,11 +21,20 @@
...
+#include <new>
+#include <type_traits>
 
 namespace iox
 {
 namespace runtime
 {
+
+// A refcount for use in getLifetimeParticipant().
+static int refcount;
```
elfenpiff, line 33: We always use types with explicit sizes. Could you please use `uint64_t` here and initialize the `refcount` and rename it into `s_refcount`.
```
static uint64_t s_refcount{0U};
```

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -21,11 +21,20 @@
...
+// Track whether the refcount lifetime mechanism is used by this particular
+// implementation of the abstract PoshRuntime class.
+static bool manual_lifetime_management = false;
```
elfenpiff, line 36: We use the universal initialization approach and lower camel case for variables. Please write
```
static bool manualLifetimeManagement{false};
```

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -47,8 +56,13 @@ void PoshRuntime::setRuntimeFactory(const factory_t& factory) noexcept
...
 PoshRuntime& PoshRuntime::defaultRuntimeFactory(cxx::optional<const RuntimeName_t*> name) noexcept
 {
-    static PoshRuntimeImpl instance(name);
-    return instance;
+    static typename std::aligned_storage<sizeof(PoshRuntimeImpl), alignof(PoshRuntimeImpl)>::type buf;
+    static cxx::ScopeGuard ltp = [](auto name){
+        new(&buf) PoshRuntimeImpl(name);
+        manual_lifetime_management = true;
+        return getLifetimeParticipant();
+    }(name);
```
elBoberido, line 64: To make gcc5 happy you need to explicitly capture `buf`. Maybe also `this` or use `PoshRuntime::manual_lifetime_management` and `PoshRuntime::getLifetimeParticipant`. Not sure about the latter one since I didn't use static member/methods in lambdas but I'm pretty sure gcc5 will complain about the former one. At least it complains when accessing non-static member.

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -67,6 +81,20 @@ PoshRuntime& PoshRuntime::getInstance(cxx::optional<const RuntimeName_t*> name)
...
 
+cxx::ScopeGuard PoshRuntime::getLifetimeParticipant() noexcept
+{
+    return cxx::ScopeGuard(
+        [](){
+            ++refcount;
+        },
+        [](){
+            if (0 == --refcount && manual_lifetime_management) {
+                getInstance().~PoshRuntime();
```
elBoberido, line 92: I'm wondering if we could get rid of `manual_lifetime_management` but I guess we need it.

It also took me some time to understand why exactly it is needed. If I understood it correctly, this is for cases where `getLifetimeParticipant` is called before `initRuntime`. In this case we cannot rely on the refcount being 0 and return an empty scope guard but must cover the case the `initRuntime` is called later on.
Could you maybe add a comment to make it more clear?

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: This must be a `std::atomic` since `getLifetimeParticipant` could be called from any thread. In addition, we might have the [static initialization fiasco](https://www.cs.technion.ac.il/users/yechiel/c++-faq/static-init-order.html), which is the reason this mechanism is so cumbersome. It just occurred to me that I might have introduced this fiasco by myself lately :grimacing:

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: This must also be an atomic.

elBoberido:
@nnmm btw. you can install the git hooks from `tools/git-hooks` then clang-format will automatically run with the commit and also take care of other things like adding the correct issue number to the commit message and you can just do a `git commit -m"Foo"` :)

nnmm:
I'm surprised that CI doesn't like the formatting, since I ran `clang-format`. Is `clang-format -i <file>` not the correct command?

iceoryx_posh/source/runtime/posh_runtime.cpp
nnmm: Yesterday after work I also realized that these things should be atomic :)

iceoryx_posh/source/runtime/posh_runtime.cpp
nnmm: We do rely on the refcount being always 0 before everything else since it is a load-time constant.

Yes, this protects against the lifetime participant going out of scope before the factory was ever called, but primarily the purpose is for child classes like `PoshRuntimeSingleProcess`, which live in a regular static variable instead of a manually-managed buffer, and where we need to prevent the destructor from being called twice (by the scope guard and the static destruction).

iceoryx_posh/source/runtime/posh_runtime.cpp
nnmm: Should it also have a `s_` prefix? I assume it stands for "static", right?

elBoberido:
@nnmm something is really weird. I guess it is a CI problem. @MatthiasKillat also had a problem with clang-format in one of his open PRs. I create a separate PR with the fix @dkroenke provided and try to get it merged soon. I will ping you once it is in

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: Happens also way to often with me in C++. It's just too easy to make it wrong. We should RiiR :grin:

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: Yes. Just to have an indicator that it is not a local variable.

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: ... oh, you could also wrap it into a anonymous namespace then it should not be accessible from other translation units.

nnmm: I think it's not necessary: https://stackoverflow.com/a/5622848/1726797 So I'd leave it like it is, but add a comment, ok?

elBoberido:
I have no idea how I closed this

iceoryx_posh/source/runtime/posh_runtime.cpp
elBoberido: @nnmm well, that is for an integer and the solution was to not initialize it but atomics have some weird/unintuitive initialization behavior and there is always Axivion which complains about not initialized variables. We also already had initialization issues with the runtime so the chicken in me leans towards the approach with the function :sweat_smile:

iceoryx_posh/source/runtime/posh_runtime.cpp
nnmm: Seems like the opposite is true, CI complains about capturing buf.

iceoryx_posh/source/runtime/posh_runtime.cpp
nnmm: Ah okay, I didn't know that atomics have weird initialization behavior (do you have a link for that?). I applied your proposal.

elfenpiff: Oh I heared c++ atomics and weird initialization behavior, just look here: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0883r2.pdf

iceoryx_posh/source/runtime/posh_runtime.cpp
```diff
@@ -47,8 +76,15 @@ void PoshRuntime::setRuntimeFactory(const factory_t& factory) noexcept
 
 PoshRuntime& PoshRuntime::defaultRuntimeFactory(cxx::optional<const RuntimeName_t*> name) noexcept
 {
-    static PoshRuntimeImpl instance(name);
-    return instance;
+    // Manual construction and destruction of the PoshRuntimeImpl, inspired by
+    // the nifty counter idiom.
+    static typename std::aligned_storage<sizeof(PoshRuntimeImpl), alignof(PoshRuntimeImpl)>::type buf;
+    static cxx::ScopeGuard ltp = [](auto name) {
```
elBoberido, line 82: Can you think of a more descriptive name than `ltp`? Maybe `mainLifetimeParticipant` ... in better :)

Oh, it also seems I gave you a bad advice with capturing. Sorry for that. Didn't grasp that it's not a member.

elBoberido (approved):
Just a minor naming thing. If the second reviewer does not care about it, this can be merged.

elBoberido:
Since it is hard to write unit tests for this and we want to get rid of the singleton anyway, I checked the unit test box.

iceoryx_posh/source/runtime/posh_runtime.cpp
nnmm: Ok, I named it `staticLifetimeParticipant` and added a comment.

elBoberido (approved):

elfenpiff (approved):

nnmm:
@elBoberido @elfenpiff Can we merge this? I verified on my branch that this solves the past-main segfault with the rclcpp default context.

elBoberido:
@nnmm sure, totally forgot that you cannot merge yourself, yet ... a few more PR and maybe some reviews and you could become a eclipse-comitter too ;)

nnmm:
Thanks :)

PR #1765, opened 2022-10-28
iox-#1394 Address AUTOSAR warnings in `RelativePointer`

mossmaurice:

1. [x] ~~Tests follow the [best practice for testing][testing]~~
1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Address AUTOSAR warnings in `RelativePointer`

## References

- Closes #1394 (partly)


iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.hpp
```diff
@@ -207,6 +198,69 @@ using UntypedRelativePointer = RelativePointer<void>;
...
+/// @param[in] rhs Relative Pointer
+/// @return True if the pointees are not equal, otherwise false
+// AXIVION Next Construct AutosarC++19_03-A13.5.5 : The RelativePointer shall explicitly be comparable to raw pointers
+template <typename T>
+bool operator!=(T* const lhs, const RelativePointer<T> rhs) noexcept;
+
+/// @brief Checks if rhs is a nullptr
+/// @param[in] lhs nullptr
+/// @param[in] rhs Relative Pointer
+/// @return True if the pointees are equal, otherwise false
```
MatthiasKillat, line 252:
```suggestion
/// @return True if rhs is equivalent to a null pointer, otherwise false
```
Equivalent to a null pointer means that it points to nothing.

Similar for other null pointer comparisons.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -27,25 +27,24 @@ namespace memory
...
+    : RelativePointer(getOffset(id, ptr), id)
 {
 }
 
 template <typename T>
+// AXIVION Next Construct AutosarC++19_03-A12.1.5 : This is the main c'tor which the other c'tors use
 // NOLINTJUSTIFICATION NewType size is comparable to an integer, hence pass by value is preferred
 // NOLINTNEXTLINE(performance-unnecessary-value-param)
-inline RelativePointer<T>::RelativePointer(offset_t offset, segment_id_t id) noexcept
+inline RelativePointer<T>::RelativePointer(const offset_t offset, const segment_id_t id) noexcept
```
MatthiasKillat, line 39: Yes, here the `const` can be to the left (due to the type of `offset_t`) but this will certainly be confusing compared to other functions where this is not the case.

As it really serves no purpose, that would be another argument to just not use `const` here.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -99,17 +98,17 @@ inline typename std::enable_if<!std::is_void<U>::value, const U&>::type Relative
 }
 
 template <typename T>
-inline T* RelativePointer<T>::operator->() const noexcept
+inline typename RelativePointer<T>::ptr_t RelativePointer<T>::operator->() const noexcept
 {
-    auto* ptr = get();
+    auto* const ptr{get()};
```
MatthiasKillat, line 103: The `const` is also questionable, only more text to express the same thing (bad, to me). The pointer is obviously not changed and only returned. There is just no need to do this.

Maybe Axivion requires it but ideally we can do without IMO.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -99,17 +98,17 @@ inline typename std::enable_if<!std::is_void<U>::value, const U&>::type Relative
...
     cxx::Ensures(ptr != nullptr);
     return ptr;
 }
 
 template <typename T>
-inline T* RelativePointer<T>::get() const noexcept
+inline typename RelativePointer<T>::ptr_t RelativePointer<T>::get() const noexcept
```
MatthiasKillat, line 109: I think it would be better to write `T*` as return type. For example https://en.cppreference.com/w/cpp/memory/shared_ptr/get does specify it like this.

Forcing any reader to look up `ptr_t` is annoying and also leads to more code here. I know that it is already necessary in the arguments and it could be argued it should be `T*` there everywhere as well. At some point it only existed to distinguish from `const_ptr_t`, but this was already redundant before and is now.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -239,11 +230,71 @@ inline typename RelativePointer<T>::ptr_t RelativePointer<T>::computeRawPtr() co
     return getPtr(segment_id_t{m_id}, m_offset);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A15.5.3, AutosarC++19_03-A15.4.2, FaultDetection-NoexceptViolations : False
+// positive, std::terminate is not called in the c'tor of PointerRepository and noexcept-specification is not violated
 inline PointerRepository<segment_id_underlying_t, UntypedRelativePointer::ptr_t>& getRepository() noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A3.3.2 : PointerRepository can't be constexpr, usage of the static
```
MatthiasKillat, line 237: This rule appears not to work correctly. It will request making `static` variables `const` (or even `constexpr`) when the intention is that the variable is just that, variable i.e. mutable.

This leads to lots of suppression comments for static variables which is not useful. Having static variables has its use (e.g. lazy singleton as here).

I suggest disabling the rule.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -239,11 +230,71 @@ inline typename RelativePointer<T>::ptr_t RelativePointer<T>::computeRawPtr() co
     return getPtr(segment_id_t{m_id}, m_offset);
 }
 
+// AXIVION Next Construct AutosarC++19_03-A15.5.3, AutosarC++19_03-A15.4.2, FaultDetection-NoexceptViolations : False
+// positive, std::terminate is not called in the c'tor of PointerRepository and noexcept-specification is not violated
```
MatthiasKillat, line 234: It may not be able to prove that no exception is thrown. Note that the code for static initialization generated by a compiler is not trivial and may involve locks of some kind (depends on type). I am not sure whether this code can throw but I do not think so. 

So I do not know how the analysis concludes that exceptions may be thrown. Static analysis of throw specifications is hard ...

Maybe disable the rule, depending on how many cases there are where it emits false positives.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -239,11 +230,71 @@ inline typename RelativePointer<T>::ptr_t RelativePointer<T>::computeRawPtr() co
...
 inline PointerRepository<segment_id_underlying_t, UntypedRelativePointer::ptr_t>& getRepository() noexcept
 {
+    // AXIVION Next Construct AutosarC++19_03-A3.3.2 : PointerRepository can't be constexpr, usage of the static
+    // object is encapsulated in the RelativePointer class
     static PointerRepository<segment_id_underlying_t, UntypedRelativePointer::ptr_t> repository;
     return repository;
 }
+
+template <typename T>
+// AXIVION Next Line AutosarC++19_03-A13.5.5 : The RelativePointer shall explicitly be comparable to raw pointers
```
MatthiasKillat, line 244: Why does static analysis have a problem with that? It is a legitimate use case and no problem at all?

Especially since it just reads values and compares them without changing the underlying objects. To make this clear we need `const T* rhs` though. Maybe this would remove the Axivion warning as well.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -239,11 +230,71 @@ inline typename RelativePointer<T>::ptr_t RelativePointer<T>::computeRawPtr() co
...
+inline bool operator==(T* const lhs, const RelativePointer<T> rhs) noexcept
+{
+    return rhs == lhs;
+}
+
+template <typename T>
+// AXIVION Next Line AutosarC++19_03-A13.5.5 : The RelativePointer shall explicitly be comparable to nullptrs
+inline bool operator==(std::nullptr_t, const RelativePointer<T> rhs) noexcept
```
MatthiasKillat, line 259: I am not sure we introduced comparisons that were not possible here, but if we did we need to add tests for them. 

If we could not compare `nullptr == RelativePointer<T>` before, we need to add this comparison to a test case. There should already be a symmetric one.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -239,11 +230,71 @@ inline typename RelativePointer<T>::ptr_t RelativePointer<T>::computeRawPtr() co
...
+inline bool operator!=(T* const lhs, const RelativePointer<T> rhs) noexcept
+{
+    return rhs != lhs;
+}
+
+template <typename T>
+// AXIVION Next Line AutosarC++19_03-A13.5.5 : The RelativePointer shall explicitly be comparable to nullptrs
+inline bool operator!=(std::nullptr_t, const RelativePointer<T> rhs) noexcept
+{
+    return rhs.get() != nullptr;
```
MatthiasKillat, line 289: Suggestion:
Could in theory be done in terms of the equality operator (and would be inlined), but I also prefer it as is.

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
```diff
@@ -103,7 +103,9 @@ inline cxx::expected<ChunkDistributorError> ChunkDistributor<ChunkDistributorDat
 {
     typename MemberType_t::LockGuard_t lock(*getMembers());
 
-    const auto iter = std::find(getMembers()->m_queues.begin(), getMembers()->m_queues.end(), queueToRemove);
+    const auto iter = std::find(getMembers()->m_queues.begin(),
+                                getMembers()->m_queues.end(),
+                                static_cast<ChunkQueueData_t* const>(queueToRemove));
```
MatthiasKillat, line 108: So `not_null` does not implicitly convert to its template type? Looks more like a problem that should be addressed there then (and probably would lead to  other changes).

I do not see why the changes in `RelativePointer` imply this change though...

MatthiasKillat (commented):
I am not at all sure the value const definitions are a good idea and consistent, see comments. Even disregarding this, there are a few other issues, especially in the comparison operators.

iceoryx_hoofs/include/iceoryx_hoofs/internal/memory/pointer_repository.hpp
```diff
@@ -81,7 +81,7 @@ class PointerRepository
     /// @param[in] id is the id to be unregistered
     /// @return true if successful, otherwise false
     /// @attention the relative pointers corresponding to this id become unsafe to use
-    bool unregisterPtr(id_t id) noexcept;
+    bool unregisterPtr(const id_t id) noexcept;
```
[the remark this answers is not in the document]
mossmaurice: @MatthiasKillat Sorry, but I have to disagree here strongly. This approach is already used quite consistently in iceoryx. IMHO it's a good thing to have everything `const` by default and selectively make things non-`const` if the variable needs to be changed (like it is done in Rust). For me the contract is also encoded in the interface more clearly, here passing a `const` identifier which shall not be modified as this wouldn't make much sense for plain value parameters.

> Does Axivion require it?

Nope, it's not.

I know that you have a different opinion about this topic. Would be acceptable for you we bring this up in the next developer meetup, discuss it there globally for the code base and until then add the `const`?

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.hpp
```diff
@@ -207,6 +198,69 @@ using UntypedRelativePointer = RelativePointer<void>;
 /// @return the static pointer repository
 PointerRepository<segment_id_underlying_t, UntypedRelativePointer::ptr_t>& getRepository() noexcept;
 
+/// @brief Checks if lhs and rhs point to the same pointee
+/// @param[in] lhs Relative Pointer
+/// @param[in] rhs Is the pointer whose pointee is compared with this' pointee
+/// @return True if the pointees are equal, otherwise false
+// AXIVION Next Construct AutosarC++19_03-A13.5.5 : The RelativePointer shall explicitly be comparable to raw pointers
+template <typename T>
+bool operator==(const RelativePointer<T> lhs, T* const rhs) noexcept;
```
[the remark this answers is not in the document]
mossmaurice: True, I changed it to ` const T* const rhs`

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.hpp
```diff
@@ -207,6 +198,69 @@ using UntypedRelativePointer = RelativePointer<void>;
 /// @return the static pointer repository
 PointerRepository<segment_id_underlying_t, UntypedRelativePointer::ptr_t>& getRepository() noexcept;
 
+/// @brief Checks if lhs and rhs point to the same pointee
+/// @param[in] lhs Relative Pointer
+/// @param[in] rhs Is the pointer whose pointee is compared with this' pointee
+/// @return True if the pointees are equal, otherwise false
```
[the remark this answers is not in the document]
mossmaurice: Fixed

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: See my comment [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1765#discussion_r1011455154), closing this thread.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: This is a finding of Axivion using the AUTOSAR ruleset.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: Agreed, I changed it. However, I would keep the alias `ptr_t` for users to find out the type of a `RelativePointer`.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: @MatthiasKillat I agree, let's monitor this a bit more and then come a decision.

cc @elBoberido

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: @MatthiasKillat I agree, it's a bit of a weird false positive. Let's monitor how Axivion behaves and then do a decision on the rule.

cc @elBoberido

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: Quote from the AUTOSAR document:

```
Comparison operators shall be non-member functions with identical parameter types and noexcept.
```

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: @MatthiasKillat This is used in `loffli.cpp:34`. I decided the implement the operator as e.g. `std::shared_ptr` does it too https://en.cppreference.com/w/cpp/memory/shared_ptr/operator_cmp

iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_distributor.inl
mossmaurice: We want to avoid any implicit conversion and make the conversion explicit as defined in both the AUTOSAR and the MISRA ruleset. That makes sense to me. The change is needed because `std::find` needs the comparison operator, which can use implicit conversion if implemented inside the class but not if the implementation is outside of the class. This is also the reason AUTOSAR and MISRA have this rule.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -189,36 +178,36 @@ template <typename T>
...
     if (static_cast<segment_id_underlying_t>(id) == NULL_POINTER_ID)
     {
         return NULL_POINTER_OFFSET;
     }
-    auto* basePtr = getBasePtr(id);
-    // NOLINTJUSTIFICATION Cast needed for pointer arithmetic
+    const auto* const basePtr = getBasePtr(id);
+    // AXIVION Next Construct AutosarC++19_03-A5.2.4, AutosarC++19_03-M5.2.9 : Cast needed for pointer arithmetic
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
     return reinterpret_cast<offset_t>(ptr) - reinterpret_cast<offset_t>(basePtr);
```
MatthiasKillat, line 190: This is weird. Why does this cast even work, as it tries to cast `const` away (should not work with `reinterpret_cast`). Probably because it is a cast to a non-ptr type and we do not try (cannot) access the data from the destination type.

It is correct, but the const required by Axivion is really too much, semi-defeats the purpose of `auto` (the brevity).

Nothing to do if we want the rule, but enforcing `const` for local variables (that are not changed) is not a good idea IMO. They should be optional for readability of longer functions.

MatthiasKillat (approved):

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: @MatthiasKillat I think in general for let's say higher-level application development it's an ok-ish rule. However, you are right that for an base library it's not very adequate.

elBoberido: The main question for me is how often we need to suppress this rule. If it is only in the lower single digit order then I wouldn't care but if it is in the double digit order then it would be more honest to just disable it.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -174,9 +161,11 @@ inline bool RelativePointer<T>::unregisterPtr(const segment_id_t id) noexcept
 template <typename T>
 // NOLINTJUSTIFICATION NewType size is comparable to an integer, hence pass by value is preferred
 // NOLINTNEXTLINE(performance-unnecessary-value-param)
-inline typename RelativePointer<T>::ptr_t RelativePointer<T>::getBasePtr(const segment_id_t id) noexcept
+inline T* RelativePointer<T>::getBasePtr(const segment_id_t id) noexcept
 {
-    return static_cast<RelativePointer<T>::ptr_t>(getRepository().getBasePtr(static_cast<segment_id_underlying_t>(id)));
+    // AXIVION Next Construct AutosarC++19_03-M5.2.8 : Cast to the underyling pointer type is safe as this is
+    // encapsulated in the RelativePointer class and type safety is ensured by using templates
+    return static_cast<T*>(getRepository().getBasePtr(static_cast<segment_id_underlying_t>(id)));
```
elBoberido, line 168: This could also be `ptr_t` instead of `T*` or `RelativePointer<T>::ptr_t`

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
elBoberido: Okay, it's not even in the lower single digit in this file :sweat_smile:

elBoberido (commented):
Just a few minor questions

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.hpp
```diff
@@ -139,20 +130,20 @@ class RelativePointer
 
     /// @brief Get the base pointer associated with this' id
     /// @return The registered base pointer of the RelativePointer object
-    ptr_t getBasePtr() const noexcept;
+    T* getBasePtr() const noexcept;
```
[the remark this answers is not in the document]
mossmaurice: After discussing this topic with @MatthiasKillat we concluded that it's better to use `T*` here because otherwise due to `M3-9-1` we would have to write in the `.inl` `typename RelativePointer<T>::ptr_t` which is not really nice.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
mossmaurice: See [my comment above](https://github.com/eclipse-iceoryx/iceoryx/pull/1765#discussion_r1013754839).

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
```diff
@@ -27,25 +27,24 @@ namespace memory
...
     , m_offset(offset)
 {
 }
 
 template <typename T>
-inline RelativePointer<T>::RelativePointer(ptr_t ptr) noexcept
-    : m_id(searchId(ptr))
-    , m_offset(computeOffset(ptr))
+inline RelativePointer<T>::RelativePointer(ptr_t const ptr) noexcept
+    : RelativePointer(getOffset(segment_id_t{searchId(ptr)}, ptr), segment_id_t{searchId(ptr)})
```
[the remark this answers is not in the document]
mossmaurice: AFAIU this would not be compliant with `A12-1-5` because it is explicitly required to use a delegating c'tor. I think it's not super nice, but an good middle ground.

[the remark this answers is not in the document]
elBoberido: In this case, please create a private constructor like `RelativPointer(prt_t const ptr, const segment_id_t id)` and call that from this ctor. The new ctor would then call the one expecting a `offset_t` and and `segment_id_t`.

Regarding the delegating constructors, I thought we refrained from this but I might not recall correctly.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.hpp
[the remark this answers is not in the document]
elBoberido: Well, then we have an inconsistency since we use `typename RelativPointer<T>::offset_t` instead of just `std::uintptr_t`. With your argumentation we should also not use that.

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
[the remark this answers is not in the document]
elBoberido: ... oh and I think the immediately invoked lambda would still work
````cpp
template <typename T>
inline RelativePointer<T>::RelativePointer(ptr_t ptr) noexcept
    : RelativePointer<T>([ptr] {
        auto id = segment_id_t{searchId(ptr)};
        auto offset = getOffset(id, ptr);
        return RelativePointer{offset, id};
    }())
{
}
```

The copy ctor would be used which is defaulted and should be cheap or even be optimized away :)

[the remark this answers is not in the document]
mossmaurice: @elBoberido Nice idea, that worked with a few adaptions for be AUTOSAR compliant ;) Thanks!

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.inl
MatthiasKillat: In any case I do not think it is a good rule, regardless how often it applies.

MatthiasKillat (approved):

iceoryx_hoofs/include/iceoryx_hoofs/memory/relative_pointer.hpp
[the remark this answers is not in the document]
MatthiasKillat: @elBoberido `T*` return was mainly for `consistency` of how the STL [smart pointers}(https://en.cppreference.com/w/cpp/memory/shared_ptr). This in particular defines no pointer type but only an element_type (which is T). The important point is to easily see that `T*` is the return type, as this is part of the contract.

Writing less is a side benefit.

elBoberido (approved):

Issue #1593, opened 2022-08-26
Create a flat include structure for `iceoryx_hoofs`


mossmaurice:
## Brief feature description

Often headers are moved inside `iceoryx_hoofs` both in folder and namespaces. This burdens the user with unnecessary code changes, having to adapt the include path.

## Detailed information

The idea would be such a structure

```
include
  |-iox
    |-vector
    |-string
    |-...
```

with e.g. `vector` containing

```
#include "iceoryx_hoofs/cxx/vector.hpp"

namesapce iox
{
    using cxx::vector;
}
```

On the user side it would look like

```
#include "iox/vector"

iox::vector<int, 42> foo;
```

### From the developer meetup https://github.com/eclipse-iceoryx/iceoryx/wiki/2023-11-16-Eclipse-iceoryx-developer-meetup#minutes

3. Relax flat include structure
    - posix is flat include and flat `iox`namespace
    - concurrent is flat include and `iox::concurrent` namespace
    - units is flat include and `iox::units` namespace
    - non public API has a `iox/detail/foo.hpp` include and is in the `iox::detail` namespace
    - non public API of bigger modules, like `cli` can opt in for `iox/detail/cli` and `iox::detail::cli`
    - public API is in `iox/header.hpp`

### From https://github.com/eclipse-iceoryx/iceoryx/pull/2142#discussion_r1439803183

We can do this also at a later point, it was just my personal taste.

My personal favorite would be:

    iox::lock_free::{mpmc|spsc} - for all lockfree code
    iox::concurrent for generic thread safe code

But I would suggest that we talk about it in a later meetup and go from there.

## Tasks

- [x] check which classes should be public
- [x] determine how to proceed with namespaces

mossmaurice:
## Sync with @elBoberido and @elfenpiff

How to proceeded further with the module structure?

1. Differentiation between user-view and iceoryx developer-view
   1. First step: #1391 iceoryx developer-view aka new folder structure (possibly without CMake changes)
   2. Second step: #1593 user-view to decouple the user from any future changes
 
## Action items

* Independently brainstorm for a iceoryx developer-wie module structure
   * Have a look at [Folly](https://github.com/facebook/folly) and [Abseil](https://github.com/abseil/abseil-cpp)

elBoberido:
@mossmaurice @elfenpiff what do you think about this

```console
iceoryx_hoofs
|- concurrent
|- core
|- posix
```
[...]

PR #1774, opened 2022-11-01
iox-#1614 Move uninitialized array

### LINKED ISSUE #1614: implement uninitialized array
## Brief feature description

In containers like the `vector`, `list`, `optional`, `string` etc. we have a byte array which stores the actual type. This byte array, can be extracted and used in those containers. This would reduce bugs since we may forget to define alignment or size properties correctly.
Furthermore we would get rid of the axivion/clang-tidy exceptions which would lead to more readable and cleaner code.

There is already the `Buffer` container in `iceoryx_hoofs/internal/concurrent/lockfree_queue/buffer.hpp` which partly implements what is required.

The container should be moved to `cxx/uninitialized_array.hpp` and refactored so that it fits the universal requirements to be used in the remaining containers.

Extracted from: #1196

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR moves the UninitializedArray to `iceoryx_hoofs/containers/include/iox/detail/` and removes the namespace `containers`.

## References

- Closes #1614

### DIFF
--- a/.clang-tidy-diff-scans.txt
+++ b/.clang-tidy-diff-scans.txt
@@ -37,7 +37,7 @@
 ./iceoryx_hoofs/test/moduletests/test_posix*
 ./iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/builder.hpp
 
-./iceoryx_hoofs/include/iceoryx_hoofs/internal/containers/*
+./iceoryx_hoofs/containers/include/iox/detail/*
 ./iceoryx_hoofs/test/moduletests/test_containers_*
 
 # IMPORTANT:
--- a/.clang-tidy-diff-scans.txt
+++ b/.clang-tidy-diff-scans.txt
@@ -37,8 +37,7 @@
 ./iceoryx_hoofs/test/moduletests/test_posix*
 ./iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/builder.hpp
 
-./iceoryx_hoofs/container/include/iox/*
-./iceoryx_hoofs/container/include/iox/detail/*
+./iceoryx_hoofs/container/include/iox/**/*
 ./iceoryx_hoofs/test/moduletests/test_container_*
 
 # IMPORTANT:
--- a/.clang-tidy-diff-scans.txt
+++ b/.clang-tidy-diff-scans.txt
@@ -37,8 +37,9 @@
 ./iceoryx_hoofs/test/moduletests/test_posix*
 ./iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/builder.hpp
 
-./iceoryx_hoofs/containers/include/iox/detail/*
-./iceoryx_hoofs/test/moduletests/test_containers_*
+./iceoryx_hoofs/container/include/iox/*
+./iceoryx_hoofs/container/include/iox/detail/*
+./iceoryx_hoofs/test/moduletests/test_container_*
 
 # IMPORTANT:
 #    after the first # everything is considered a comment, add new files and
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -785,7 +785,7 @@
     * `iox::bar::foo` to `iox::foo`
         * `iceoryx_hoofs/bar/foo.hpp` to `iox/foo.hpp`
 
-41. Use proper aligned `containers::UninitializedArray` instead of C-style array
+41. Use proper aligned `iox::UninitializedArray` instead of C-style array
 
     ```cpp
     // before
@@ -795,10 +795,10 @@
     alignas(T) element_t myAlignedArray[Capacity];
 
     // after
-    #include "iceoryx_hoofs/containers/uninitialized_array.hpp"
+    #include "iox/detail/uninitialized_array.hpp"
 
-    containers::UninitializedArray<char, Capacity, containers::ZeroedBuffer> myCharArray;
+    iox::UninitializedArray<char, Capacity, containers::ZeroedBuffer> myCharArray;
 
-    containers::UninitializedArray<T, Capacity> myAlignedArray;
+    iox::UninitializedArray<T, Capacity> myAlignedArray;
 
     ```
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -795,9 +795,9 @@
     alignas(T) element_t myAlignedArray[Capacity];
 
     // after
-    #include "iox/detail/uninitialized_array.hpp"
+    #include "iox/uninitialized_array.hpp"
 
-    iox::UninitializedArray<char, Capacity, containers::ZeroedBuffer> myCharArray;
+    iox::UninitializedArray<char, Capacity, iox::ZeroedBuffer> myCharArray;
 
     iox::UninitializedArray<T, Capacity> myAlignedArray;
 
--- a/iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
+++ b/iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
@@ -30,12 +30,15 @@ namespace iox
 template <typename ElementType, uint64_t Capacity>
 struct ZeroedBuffer
 {
+    struct alignas(ElementType) element_t
+    {
+        // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+        cxx::byte_t data[sizeof(ElementType)];
+    };
     // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
-    // NOLINTBEGIN(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
-    using element_t = cxx::byte_t[sizeof(ElementType)];
-    // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
-    alignas(ElementType) element_t value[Capacity]{};
-    // NOLINTEND(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    element_t value[Capacity]{};
 };
 
 /// @brief struct used as policy parameter in UninitializedArray to wrap an uninitialized array
@@ -44,12 +47,15 @@ struct ZeroedBuffer
 template <typename ElementType, uint64_t Capacity>
 struct NonZeroedBuffer
 {
+    struct alignas(ElementType) element_t
+    {
+        // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+        cxx::byte_t data[sizeof(ElementType)];
+    };
     // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
-    // NOLINTBEGIN(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
-    using element_t = cxx::byte_t[sizeof(ElementType)];
-    // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
-    alignas(ElementType) element_t value[Capacity];
-    // NOLINTEND(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    element_t value[Capacity];
 };
 
 /// @brief Wrapper class for a C-style array of type ElementType and size Capacity. Per default it is uninitialized but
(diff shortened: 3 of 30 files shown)


elfenpiff (approved):

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -795,10 +795,10 @@
     alignas(T) element_t myAlignedArray[Capacity];
 
     // after
-    #include "iceoryx_hoofs/containers/uninitialized_array.hpp"
+    #include "iox/detail/uninitialized_array.hpp"
```
elBoberido, line 798: Since this wasn't in `internal` before, would it make sense to keep it in `iox/uninitialized_array.hpp`?

We did not specify how to handle this in https://github.com/eclipse-iceoryx/iceoryx/issues/1593#issuecomment-1258266337 so I guess it would be fine. I think I also need to move the comment from that issue to a more prominent location

elBoberido (commented):
Looks fine. Just one question related to the location and it seems bazel is not happy

elBoberido:
Oh, the clang-tidy warning will disappear once you rebase to master :)

doc/website/release-notes/iceoryx-unreleased.md
FerdinandSpitzschnueffler: Well, it was in `internal` before but the include path in the release notes have been wrong :grimacing:

.clang-tidy-diff-scans.txt
```diff
@@ -37,8 +37,9 @@
 ./iceoryx_hoofs/test/moduletests/test_posix*
 ./iceoryx_hoofs/include/iceoryx_hoofs/design_pattern/builder.hpp
 
-./iceoryx_hoofs/include/iceoryx_hoofs/internal/containers/*
-./iceoryx_hoofs/test/moduletests/test_containers_*
+./iceoryx_hoofs/container/include/iox/*
+./iceoryx_hoofs/container/include/iox/detail/*
```
elBoberido, line 41:
```suggestion
./iceoryx_hoofs/container/include/iox/**/*
```

elBoberido (approved):
I think the clang-tidy-diff file could be simplified

elfenpiff (dismissed):

elBoberido (dismissed):

iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
```diff
@@ -15,29 +15,30 @@
...
 {
-namespace containers
-{
 /// @brief struct used as policy parameter in UninitializedArray to wrap an array with all elements zeroed
 /// @tparam ElementType is the array type
 /// @tparam Capacity is the array size
 template <typename ElementType, uint64_t Capacity>
 struct ZeroedBuffer
 {
+    struct alignas(ElementType) element_t
```
MatthiasKillat, line 33: @FerdinandSpitzschnueffler Good, this allows using it with `char` as expected. 

This should prevent the optimization that we stumbled about. It was a `strlen` optimization in gcc 8.x (specifically 8.4) that could maybe happen also in other circumstances, as the compiler may rightfully assume that `strlen` of a `char[1]` must be 0 (due to zero-termination).

The effects of various optimization can be observed in https://godbolt.org/z/ooqd9M5T3

iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
```diff
@@ -15,29 +15,30 @@
...
+    {
+        // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+        cxx::byte_t data[sizeof(ElementType)];
```
MatthiasKillat, line 37: Adding any other fields to the struct will break things. Maybe worth a comment.

I am also wondering whether the zero-initialization braces `{ }` are required here (instead of array of `element_t`).

iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
```diff
@@ -15,29 +15,30 @@
...
+    };
     // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
-    // NOLINTBEGIN(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
-    using element_t = cxx::byte_t[sizeof(ElementType)];
-    // AXIVION Next Construct AutosarC++19_03-A18.1.1 : required by low level UninitializedArray building block and encapsulated in abstraction
-    alignas(ElementType) element_t value[Capacity]{};
-    // NOLINTEND(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    element_t value[Capacity]{};
```
MatthiasKillat, line 41: If we move the braces to `element_t`, they should not be required here. I am generally not sure how it works with padding without trying it out.

There can be trailing padding in `element_t` (to make size a multiple of alignment). Depends on our requirements of the padding initialization. Padding initialization which would not happen if we move it I think, but would it as it is now?

@elBoberido FYI

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
```diff
@@ -229,7 +230,14 @@ class optional final : public FunctionalInterface<optional<T>, T, void>
     //     initHandle(&handle);
     //   }
     bool m_hasValue{false};
-    typename std::aligned_storage<sizeof(T), alignof(T)>::type m_data;
+    struct alignas(T) element_t
```
MatthiasKillat, line 233: Should not be necessary to use our own struct. Do we want to remove the dependency on `std::aligned_storage` or is there another reason?

iceoryx_hoofs/include/iceoryx_hoofs/cxx/variant.hpp
```diff
@@ -265,10 +267,13 @@ class variant
     constexpr uint64_t index() const noexcept;
 
   private:
-    // NOLINTJUSTIFICATION safe access is guaranteed since the c-array is wrapped inside the variant class
-    // NOLINTBEGIN(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays)
-    alignas(algorithm::maxVal(alignof(Types)...)) internal::byte_t m_storage[TYPE_SIZE]{0U};
-    // NOLINTEND(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays)
+    struct alignas(TYPE_ALIGNMENT) storage_t
```
MatthiasKillat, line 270: This is almost a aligned_storage alternative. Can be factored  out later (in another issue) if we do not want to use `std::aligned_storage`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
```diff
@@ -31,7 +31,7 @@ inline constexpr variant<Types...>::variant(const variant& rhs) noexcept
 {
     if (m_type_index != INVALID_VARIANT_INDEX)
     {
-        internal::call_at_index<0, Types...>::copyConstructor(m_type_index, rhs.m_storage, m_storage);
+        internal::call_at_index<0, Types...>::copyConstructor(m_type_index, &rhs.m_storage.data[0], &m_storage.data[0]);
```
MatthiasKillat, line 34: We could probably get rid of the `.data` indirection by defining a conversion to a pointer to the start of the storage (as part of the storage). This is also only a suggestion if we decide to implement our own aligned_storage in another issue.

Generally the address of the struct will be the same as the address of data, so if a type erased pointer (e.g. `void*` or `char*`) is used this should not be needed.

Observation, not to be done in this issue.

MatthiasKillat (dismissed):
The brace zero initialization question is the only relevant comment. Everything else is only a style question or potential future refactoring in a different issue.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/variant.hpp
elBoberido, line 270: If the `TYPE_ALIGNMENT` isn't used anywhere else, we could get rid of `algorithm::maxVal(alignof(Types)...);` and just use `alignas(Types...)`

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
elBoberido, line 34: I think this leads to warnings with `-Wcast-align=strict`. Unfortunately this is baked into the `call_at_index` struct.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
```diff
@@ -175,7 +177,7 @@ inline bool variant<Types...>::emplace_at_index(CTorArguments&&... args) noexcep
     using T = typename internal::get_type_at_index<0, TypeIndex, Types...>::type;
 
     call_element_destructor();
-    new (m_storage) T(std::forward<CTorArguments>(args)...);
+    new (&m_storage.data) T(std::forward<CTorArguments>(args)...);
```
elBoberido, line 180: Why not just `new (&m_storage) ...`?

elBoberido (commented):
Looks good. Just a few minor remarks

iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
elBoberido: Since the `UninitialiizedArray` is typed and `element_t` uses the size and alignment of that type, there are no issues with additional padding bytes

iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
FerdinandSpitzschnueffler: Thanks for the example!

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
elBoberido: @FerdinandSpitzschnueffler after some thinking, the `call_at_index` should use `void*` instead of `byte_t*`. Could you please change that?

With that change there is only two offender when using `-Wcast_aling=strict`

One is in `types.h` in the C binding
```cpp
/// @brief handle of the chunk header
typedef struct
{
    // could be empty but then we get `struct has no members` warning
    uint8_t do_not_touch_me[1];
} iox_chunk_header_t;
```
to
```cpp
/// @brief handle of the chunk header
typedef struct iox_chunk_header_opaque_t iox_chunk_header_t;
```

And the others in the UDS and MQ iceperf code ... which is by the way broken.

I think `types.h` and the iceperf example could be fixed in another PR.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/optional.hpp
FerdinandSpitzschnueffler: I replaced `std::aligned_storage` since it is deprecated in C++23 (I know, not a problem now but maybe later) and already added similar constructs in `variant` and `UninitializedArray`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
MatthiasKillat: I also think this should work if data is the only member, at least here where `void*` is expected.

FerdinandSpitzschnueffler: @elBoberido I don't know, maybe some copy/paste mistake. I removed `.data`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
FerdinandSpitzschnueffler: I removed `.data` now that I changed the `byte_t*` parameter in `call_at_index` to `void*` (before it was necessary).

FerdinandSpitzschnueffler: I changed `call_at_index` to take a `void*` and created #1778 for the warnings in the C binding and the iceperf example.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant_internal.hpp
```diff
@@ -109,7 +109,7 @@ struct call_at_index
         }
     }
 
-    static void move(const uint64_t index, byte_t* source, byte_t* destination) noexcept
+    static void move(const uint64_t index, void* source, void* destination) noexcept
```
elBoberido, line 112: Oh boy, I hate to do this but
```suggestion
    static void move(const uint64_t index, const void* const source, void* const destination) noexcept
```
:sweat_smile:

elBoberido (approved):
My last remark can also be done in a separate PR since it is not really related to the uninitialized array

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant.inl
elBoberido: @MatthiasKillat I like the idea of a `iox::byte` similar to `std::byte`. Would be something for a separate issue.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant_internal.hpp
FerdinandSpitzschnueffler: @elBoberido I didn't follow your approach completely for move so that it doesn't degrade to a copy operation. Hope I didn't mess up the const correctness - it is late...

elBoberido (approved):

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/variant_internal.hpp
elBoberido: @FerdinandSpitzschnueffler yeah, I mixed it up :sweat_smile:

elfenpiff (approved):

Issue #180, opened 2020-07-08
Add implementation of std::span to iceoryx_hoofs


ithier:
## Brief feature description
The std::span STL feature from C++20 will be useful for working with byte arrays in iceoryx.

## Detailed information
See: 
* https://en.cppreference.com/w/cpp/container/span
* https://en.cppreference.com/w/cpp/container/span/begin

One use-case is for passing around pointers to byte arrays to a function.
Usually, some checks need to be done in the function to ensure the array is valid (i.e. nullptr check, size validation).
With std::span, all of this logic is done automatically and the function doesn't need to reimplement it.

patdhlk:
I'm working on the it, I hope a fist version is available beginning next week

PR #1785, opened 2022-11-10
Iox-#180-add-span-implementation

patdhlk:

1. [ ] Branch follows the naming format (`iox-123-this-is-a-branch`)
1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Closes **TBD**


iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,583 @@
...
+/// @brief  C++11 compatible span implementation. The class template span describes an object that can refer to a
+///         contiguous sequence of objects with the first element of the sequence at position zero. A span can either
+///         have a static extent, in which case the number of elements in the sequence is known at compile-time and
+///         encoded in the type, or a dynamic extent. If a span has dynamic extent, a typical implementation holds two
+///         members: a pointer to T and a size. A span with static extent may have only one member: a pointer to T.
+///         http://eel.is/c++draft/views contains the latest C++20 draft
+/// @tparam T - element type; must be a complete object type that is not an abstract class type
+/// @tparam Extent - the number of elements in the sequence, or std::dynamic_extent if dynamic
+template <typename T, uint64_t Extent>
+class span : public internal::SpanStorage<Extent>
```
patdhlk, line 216: @patdhlk TODO: move implementatio so seperate *.inl file

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+#include "iceoryx_hoofs/cxx/requires.hpp"
+
+namespace iox
+{
+namespace cxx
+{
+
+
+inline constexpr bool ConstexprCheckTrue(bool condition)
```
patdhlk, line 29: @patdhlk 

- [ ] create ExpectsWithMsg variant

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,414 @@
...
+{
+namespace cxx
+{
+// constants
+constexpr uint64_t dynamic_extent = std::numeric_limits<uint64_t>::max();
+
+template <class From, class To>
+constexpr bool is_convertible_v = std::is_convertible<From, To>::value;
+
+// Implementation of C++17 std::size() and std::data()
```
patdhlk, line 44: Q: where to put the implementation of C++17 std::size() and std::data()? Any recommendations? `iox::cxx:span` is not the correct place

patdhlk: @mossmaurice if I understand [n4280](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4280.pdf) correctly they are part of `<iterator>`

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,414 @@
...
+#include "iceoryx_hoofs/cxx/span_iterator.hpp"
+#include "iceoryx_hoofs/cxx/type_traits.hpp"
+
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+#include <limits>
+#include <type_traits>
```
mossmaurice, line 29: The trampoline already includes this here, so an additional include is not needed.

```suggestion
```

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
mossmaurice: @patdhlk Ok. For use I would say they fit best in `type_traits.hpp`.

MatthiasKillat: @patdhlk In the STL, they are part of the individual container headers (vector.hpp etc.) and this is logical. We cannot have `type_traits` depend on those.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,414 @@
...
+#include <utility>
+
+using namespace std;
+
+namespace iox
+{
+namespace cxx
+{
+// constants
+constexpr uint64_t dynamic_extent = std::numeric_limits<uint64_t>::max();
```
MatthiasKillat, line 37: Not the correct capitalization for constants as of the current rules (should be `DYNAMIC_EXTENT`).

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,414 @@
...
+    return array;
+}
+
+template <class E>
+constexpr const E* data(std::initializer_list<E> il) noexcept
+{
+    return il.begin();
+}
+
+template <typename T, uint64_t Extent = dynamic_extent>
```
MatthiasKillat, line 79: What is Extend used for? (I will probably find out later in the review).

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,414 @@
...
+constexpr uint64_t dynamic_extent = std::numeric_limits<uint64_t>::max();
+
+template <class From, class To>
+constexpr bool is_convertible_v = std::is_convertible<From, To>::value;
+
+// Implementation of C++17 std::size() and std::data()
+template <class C>
+constexpr auto size(const C& c) -> decltype(c.size())
+{
+    return c.size();
```
MatthiasKillat, line 46: Function definition in header files is currently discouraged, should be moved to an inl file (only declared here).

IMO we have to move away from this as it leads to much boilerplate. The advantage is that the interface in the header is concise (but that is actually the task of doxygen).

Doxygen documentation is missing.

MatthiasKillat (commented):
@patdhlk I just head a quick look and left some comments. Did not review since it is a draft.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
patdhlk: @MatthiasKillat thank you for the input. I will leave them in the `span.hpp` and move the implementation to the `span.inl` file

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
patdhlk: It is a requirement for `std::span`:

> A span can either have a static extent, in which case the number of elements in the sequence is known at compile-time and encoded in the type, or a dynamic extent

see [en.cppreference.com](https://en.cppreference.com/w/cpp/container/span)

mossmaurice:
@patdhlk Can you rebase the branch to resolve the conflicts? Thanks :pray:

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+        EXPECT_EQ(vector[i], const_span[i]);
+
+    span<int> dyn_span(vector);
+    EXPECT_EQ(vector.data(), dyn_span.data());
+    EXPECT_EQ(vector.size(), dyn_span.size());
+
+    for (size_t i = 0; i < dyn_span.size(); ++i)
+        EXPECT_EQ(vector[i], dyn_span[i]);
+
+    span<int, 6> static_span(vector.data(), vector.size());
```
mossmaurice, line 96: Same here.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+    static_assert(nullptr == static_span.data(), "");
+    static_assert(0u == static_span.size(), "");
+}
+
+TEST_F(span_test, NewSpanFromIterator)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "92de7ff5-d03e-41ae-bfdd-893892381b24");
+    constexpr int* kNull = nullptr;
+    constexpr span<int> empty_span(kNull, 0);
+    EXPECT_TRUE(empty_span.empty());
```
mossmaurice, line 57: This will end the test if not fulfilled and not continue to work on corrupted data.

```suggestion
    ASSERT_TRUE(empty_span.empty());
```

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+    EXPECT_EQ(nullptr, empty_span.data());
+
+    std::vector<int> vector = {1, 1, 2, 3, 5, 8};
+
+    span<int> dyn_span(vector.begin(), vector.size());
+    EXPECT_EQ(vector.data(), dyn_span.data());
+    EXPECT_EQ(vector.size(), dyn_span.size());
```
mossmaurice, line 64:
```suggestion
    ASSERT_EQ(vector.size(), dyn_span.size());
    ASSERT_EQ(vector.data(), dyn_span.data());
```

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+        EXPECT_EQ(array[i], static_span[i]);
+}
+
+TEST_F(span_test, NewSpanCreatedFromConstexprArray)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "9ec9c31c-b97f-43a3-9669-3bdff3a82b9e");
+    static constexpr int arr[] = {5, 4, 3, 2, 1};
+
+    constexpr span<const int> dyn_span(arr);
+    static_assert(arr == dyn_span.data(), "");
```
mossmaurice, line 134: Why not?

```suggestion
    EXPECT_EQ(arr, dyn_span.data());
```

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+    void SetUp() override
+    {
+    }
+
+    void TearDown() override
+    {
+    }
+};
+
+TEST_F(span_test, NewSpanCreatedWithTheDefaultConstructor)
```
mossmaurice, line 40:
```suggestion
TEST_F(span_test, NewSpanCreatedWithTheDefaultConstructorHasEqualSizeAndDataPtr)
```

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    span<const int, 6> static_span(vector.data(), vector.size());
+    EXPECT_EQ(vector.data(), static_span.data());
+    EXPECT_EQ(vector.size(), static_span.size());
+
+    for (size_t i = 0; i < static_span.size(); ++i)
+        EXPECT_EQ(vector[i], static_span[i]);
+}
+
+TEST_F(span_test, CheckFrontOfSpanIfItReturnsTheElementAtIndex0)
```
mossmaurice, line 200: :heart:

mossmaurice (commented):
Thanks for the PR and your work! Much appreciated. Looking forward to having a `span` finally available in iceoryx :tada:

I left a few nitpicks and was rather inclined to comment a bit more for your to understand our typical guidelines.

MatthiasKillat:
@mossmaurice @patdhlk I have not forgotten this and will get to this after some priority tasks.

mossmaurice:
@patdhlk I'm taking over this PR and will get it merged in the next days.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span_iterator.hpp
mossmaurice: Added a comment to refactor this check with an `if constexpr` once moving to C++17.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
```diff
@@ -0,0 +1,414 @@
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/include/iceoryx_dust/internal/cxx/span.inl
```diff
@@ -0,0 +1,296 @@
...
+    : span(other.data(), other.size())
+{
+}
+
+template <typename T, uint64_t Extent>
+template <uint64_t Count>
+inline constexpr span<T, Count> span<T, Extent>::first() const noexcept
+{
+    static_assert(Count <= Extent, "Count must not exceed Extent");
+    static_assert(Extent != DYNAMIC_EXTENT || Count <= size());
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/internal/cxx/span.inl
```diff
@@ -0,0 +1,296 @@
...
+    return {data(), Count};
+}
+
+template <typename T, uint64_t Extent>
+inline constexpr span<T, DYNAMIC_EXTENT> span<T, Extent>::first(uint64_t count) const noexcept
+{
+    static_assert(count <= size());
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+
+template <typename Container, typename T, uint64_t Extent>
+using enable_if_compatible_dynamic_container_t =
+    std::enable_if_t<is_compatible_container_t<Container, T>::value && Extent == DYNAMIC_EXTENT>;
+
+template <uint64_t Extent>
+class span_storage
+{
+  public:
+    constexpr explicit span_storage(uint64_t size) noexcept
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
```diff
@@ -159,6 +160,85 @@ struct is_cxx_string<::iox::cxx::string<N>> : std::true_type
...
+{
+};
+
+/// @brief Implementation of C++17 std::conjunction
+template <class...>
+struct conjunction : std::true_type
+{
+};
+
+template <class _Arg>
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+        : size_(size)
+    {
+    }
+    constexpr uint64_t size() const noexcept
+    {
+        return size_;
+    }
+
+  private:
+    uint64_t size_;
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+    /// @param count
+    template <typename It>
+    constexpr span(It first, uint64_t count) noexcept;
+
+    /// @brief Constructs a span that is a view over the range [first, last);
+    /// @tparam It
+    /// @tparam End
+    /// @tparam
+    /// @param begin
+    /// @param end
```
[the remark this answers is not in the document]
mossmaurice: Done :heavy_check_mark:

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+    constexpr T& front() const noexcept;
+
+    /// @brief access the last element
+    /// @return
+    constexpr T& back() const noexcept;
+
+    /// @brief accesses an element of the sequence
+    /// @param idx
+    /// @return
+    constexpr T& operator[](uint64_t idx) const noexcept;
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/include/iceoryx_dust/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+    template <class Type2, std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
+    constexpr bool operator>=(const span_iterator<Type2>& rhs) const noexcept
+    {
+        return !(*this < rhs);
+    }
+
+    const pointer m_begin = nullptr;
+    const pointer m_end = nullptr;
+    pointer m_current = nullptr;
```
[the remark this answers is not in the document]
mossmaurice: Done

iceoryx_dust/include/iceoryx_dust/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    constexpr span_iterator& operator+=(const difference_type n) noexcept
+    {
+        if (n != 0)
+            iox::cxx::ConstexprCheckTrue(m_begin && m_current && m_end);
+        if (n > 0)
+            iox::cxx::ConstexprCheckTrue(m_end - m_current >= n);
+        if (n < 0)
+            iox::cxx::ConstexprCheckTrue(m_current - m_begin >= -n);
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/include/iceoryx_dust/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+    friend constexpr span_iterator operator+(const difference_type n, const span_iterator& rhs) noexcept
+    {
+        return rhs + n;
+    }
+
+    constexpr span_iterator& operator-=(const difference_type n) noexcept
+    {
+        if (n != 0)
+            iox::cxx::ConstexprCheckTrue(m_begin && m_current && m_end);
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/include/iceoryx_dust/internal/cxx/span.inl
```diff
@@ -0,0 +1,296 @@
...
+template <typename T, uint64_t Extent>
+inline constexpr T& span<T, Extent>::back() const noexcept
+{
+    static_assert(Extent == DYNAMIC_EXTENT || Extent > 0, "Extent must not be 0");
+    static_assert(Extent != DYNAMIC_EXTENT || !empty());
+    return *(data() + size() - 1);
+}
+
+template <typename T, uint64_t Extent>
+inline constexpr T& span<T, Extent>::operator[](uint64_t idx) const noexcept
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+  public:
+    void SetUp() override
+    {
+    }
+
+    void TearDown() override
+    {
+    }
+};
```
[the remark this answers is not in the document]
mossmaurice: Changed.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    for (size_t i = 0; i < dyn_span.size(); ++i)
+        EXPECT_EQ(vector[i], dyn_span[i]);
+
+    span<int, 6> static_span(vector.begin(), vector.size());
+    EXPECT_EQ(vector.data(), static_span.data());
+    EXPECT_EQ(vector.size(), static_span.size());
+
+    for (size_t i = 0; i < static_span.size(); ++i)
+        EXPECT_EQ(vector[i], static_span[i]);
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+TEST_F(span_test, NewSpanCreatedWithTheDefaultConstructor)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "6a0589e2-b2fb-4f81-8082-6d3ea411e659");
+    span<int> dyn_span;
+    EXPECT_EQ(nullptr, dyn_span.data());
+    EXPECT_EQ(0u, dyn_span.size());
+
+    constexpr span<int, 0> static_span;
+    static_assert(nullptr == static_span.data(), "");
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    void TearDown() override
+    {
+    }
+};
+
+TEST_F(span_test, NewSpanCreatedWithTheDefaultConstructor)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "6a0589e2-b2fb-4f81-8082-6d3ea411e659");
+    span<int> dyn_span;
```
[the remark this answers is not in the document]
mossmaurice: Changed for all occurrences.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+    std::vector<int> vector = {1, 1, 2, 3, 5, 8};
+
+    span<int> dyn_span(vector.begin(), vector.size());
+    EXPECT_EQ(vector.data(), dyn_span.data());
+    EXPECT_EQ(vector.size(), dyn_span.size());
+
+    for (size_t i = 0; i < dyn_span.size(); ++i)
+        EXPECT_EQ(vector[i], dyn_span[i]);
+
+    span<int, 6> static_span(vector.begin(), vector.size());
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
mossmaurice: Done.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
mossmaurice: Done.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
mossmaurice: Done.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
mossmaurice: That can't be done as this must be a compile-time check.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
mossmaurice: Done.

iceoryx_dust/test/moduletests/test_cxx_span.cpp
```diff
@@ -0,0 +1,237 @@
...
+    span<int> dyn_span;
+    EXPECT_EQ(nullptr, dyn_span.data());
+    EXPECT_EQ(0u, dyn_span.size());
+
+    constexpr span<int, 0> static_span;
+    static_assert(nullptr == static_span.data(), "");
+    static_assert(0u == static_span.size(), "");
+}
+
+TEST_F(span_test, NewSpanFromIterator)
```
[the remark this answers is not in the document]
mossmaurice: Done.

iceoryx_dust/include/iceoryx_dust/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
...
+    constexpr span_iterator(const span_iterator& other) = default;
+
+    constexpr operator span_iterator<const T>() const noexcept
+    {
+        return {m_begin, m_end, m_current};
+    }
+
+    constexpr reference operator*() const noexcept
+    {
+        iox::cxx::ConstexprCheckTrue(m_begin && m_end);
```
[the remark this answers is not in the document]
mossmaurice: This check will be replace with a `if constexpr(..)` one once switched to C++17

iceoryx_dust/include/iceoryx_dust/internal/cxx/span.inl
```diff
@@ -0,0 +1,296 @@
...
+    static_assert(Extent == DYNAMIC_EXTENT || Extent == 0, "Invalid Extent");
+}
+
+template <typename T, uint64_t Extent>
+template <typename It>
+inline constexpr span<T, Extent>::span(It first, uint64_t count) noexcept
+    : span_storage_t(count)
+    , m_data(internal::to_address(first))
+{
+    iox::cxx::ConstexprCheckTrue(Extent == DYNAMIC_EXTENT || Extent == count);
```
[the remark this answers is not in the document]
mossmaurice: Same here.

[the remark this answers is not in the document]
mossmaurice: See above, `ConstexprCheckTrue` replace with `assert`, which can be executed during compile-time with C++14.

iceoryx_dust/include/iceoryx_dust/cxx/span_iterator.hpp
```diff
@@ -0,0 +1,214 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+#ifndef IOX_DUST_CXX_SPAN_ITERATOR_HPP
+#define IOX_DUST_CXX_SPAN_ITERATOR_HPP
+
+#include <iterator> // for reverse_iterator, distance, random_access_...
```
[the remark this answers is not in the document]
mossmaurice: Fixed.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+    using pointer = T*;
+    using reference = T&;
+    using iterator = iox::cxx::span_iterator<T>;
+    using reverse_iterator = std::reverse_iterator<iterator>;
+    static constexpr uint64_t extent = Extent;
+
+    // constructors, copy, assignment, and destructor
+
+    /// @brief Constructs an empty span whose data() == nullptr and size() == 0.
+    constexpr span() noexcept;
```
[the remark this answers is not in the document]
mossmaurice: Removed as described above.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+
+    /// @brief Constructs a span that is a view over the array arr; the resulting span has size() == N and data() ==
+    /// std::data(arr)
+    /// @tparam U
+    /// @tparam
+    /// @tparam N
+    /// @param array
+    template <typename U,
+              uint64_t N,
+              typename = internal::enable_if_compatible_array_t<iox::UninitializedArray<U, N>&, T, Extent>>
```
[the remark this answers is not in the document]
mossmaurice: Turns this is needed in order to make sure that a `data` and `size` specialisation is there for `UninitializedArray`.

iceoryx_dust/include/iceoryx_dust/cxx/span.hpp
```diff
@@ -0,0 +1,438 @@
...
+
+    /// @brief Constructs a span that is a view over the array arr; the resulting span has size() == N and data() ==
+    /// std::data(arr)
+    /// @tparam U
+    /// @tparam
+    /// @tparam N
+    /// @param array
+    template <typename U,
+              uint64_t N,
+              typename = internal::enable_if_compatible_array_t<const iox::UninitializedArray<U, N>&, T, Extent>>
```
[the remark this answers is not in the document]
mossmaurice: See above.

iceoryx_hoofs/include/iceoryx_hoofs/cxx/span.hpp
mossmaurice: Done :heavy_check_mark:

iceoryx_hoofs/include/iceoryx_hoofs/cxx/type_traits.hpp
```diff
@@ -23,6 +23,7 @@
 
 #include "iceoryx_hoofs/cxx/attributes.hpp"
 #include "iceoryx_platform/platform_settings.hpp"
+#include "iox/uninitialized_array.hpp"
```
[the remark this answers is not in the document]
mossmaurice: Done :heavy_check_mark:

patdhlk:
> @patdhlk I'm taking over this PR and will get it merged in the next days.

@mossmaurice thank you :pray:

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+template <typename Container>
+constexpr auto size(const Container& container) -> decltype(container.size());
+
+/// @brief Returns N
+/// @tparam T Type of the array
+/// @tparam N Size of the array
+/// @param An array
+/// @return Size of the array
+template <typename T, std::uint64_t N>
+constexpr std::uint64_t size(const T (&)[N]) noexcept;
```
elBoberido, line 53: This is already available in `iox/size.hpp`

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+constexpr uint64_t DYNAMIC_EXTENT = std::numeric_limits<uint64_t>::max();
+
+// Implementation of C++17 std::size() and std::data()
+
+/// @brief Returns c.size(), converted to the return type if necessary.
+/// @tparam C Type of the container
+/// @param c A container or view with a size member function
+/// @return The size of c
+template <typename Container>
+constexpr auto size(const Container& container) -> decltype(container.size());
```
elBoberido, line 45: I'm a bit torn here since the overload might have a different return type than the function below or in `size.hpp`

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+namespace iox
+{
+// constants
+constexpr uint64_t DYNAMIC_EXTENT = std::numeric_limits<uint64_t>::max();
+
+// Implementation of C++17 std::size() and std::data()
+
+/// @brief Returns c.size(), converted to the return type if necessary.
+/// @tparam C Type of the container
+/// @param c A container or view with a size member function
```
elBoberido, line 42:
```suggestion
/// @param container A container or view with a size member function
```

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+
+namespace iox
+{
+// constants
+constexpr uint64_t DYNAMIC_EXTENT = std::numeric_limits<uint64_t>::max();
+
+// Implementation of C++17 std::size() and std::data()
+
+/// @brief Returns c.size(), converted to the return type if necessary.
+/// @tparam C Type of the container
```
elBoberido, line 41:
```suggestion
/// @tparam Container Type of the container
```

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+using namespace std;
+
+namespace iox
+{
+// constants
+constexpr uint64_t DYNAMIC_EXTENT = std::numeric_limits<uint64_t>::max();
+
+// Implementation of C++17 std::size() and std::data()
+
+/// @brief Returns c.size(), converted to the return type if necessary.
```
elBoberido, line 40:
```suggestion
/// @brief Returns container.size(), converted to the return type if necessary.
```

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+/// @tparam T Type of the iox::UninitializedArray
+/// @tparam N Size of the iox::UninitializedArray
+/// @param An iox::UninitializedArray
+/// @return Size of the iox::UninitializedArray
+template <typename T, std::uint64_t N, template <typename, uint64_t> class Buffer>
+constexpr std::uint64_t size(const UninitializedArray<T, N, Buffer>&) noexcept;
+
+/// @brief Returns a pointer to the block of memory containing the elements of the range.
+/// @tparam C Type of the container
+/// @param c A container or view with a data() member function
```
elBoberido, line 65:
```suggestion
/// @tparam Container Type of the container
/// @param container A container or view with a data() member function
```

Please also change the other occurrences

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+template <typename T, std::uint64_t N>
+constexpr std::uint64_t size(const T (&)[N]) noexcept;
+
+/// @brief Returns N
+/// @tparam T Type of the iox::UninitializedArray
+/// @tparam N Size of the iox::UninitializedArray
+/// @param An iox::UninitializedArray
+/// @return Size of the iox::UninitializedArray
+template <typename T, std::uint64_t N, template <typename, uint64_t> class Buffer>
+constexpr std::uint64_t size(const UninitializedArray<T, N, Buffer>&) noexcept;
```
elBoberido, line 61: Just thinking out loud. I'm wondering which of the constructs is more basic, the `UninitializedArray` or the `span`. It might make sense to move this specialization to the `UninitializedArray`

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+template <typename T, std::uint64_t N, template <typename, uint64_t> class Buffer>
+constexpr T* data(UninitializedArray<T, N, Buffer>& uninit_array) noexcept;
+
+template <typename T, uint64_t Extent = DYNAMIC_EXTENT>
+class span;
+
+namespace detail
+{
+template <uint64_t I>
+using size_constant = std::integral_constant<uint64_t, I>;
```
elBoberido, line 100: I'm wondering if it would make sense to move this to a more central location. `size_constant` is a quite generic name and I fear some issues regarding double definition ... not like compile errors but having this in multiple translation units and then getting some weird linker issues

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,451 @@
...
+template <typename T>
+using extent_t = extent_impl<remove_cvref_t<T>>;
+
+template <typename T>
+struct is_span : std::false_type
+{
+};
+
+template <typename T, uint64_t Extent_t>
+struct is_span<span<T, Extent_t>> : std::true_type
```
elBoberido, line 132:
```suggestion
template <typename T, uint64_t Extent>
struct is_span<span<T, Extent>> : std::true_type
```

It is a little bit inconsistent since the other template parameter also do not have a `_t` suffix

elBoberido (commented):
Did not yet fully finish the header and the implementation/tests

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: Good point, I moved the specialization.

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: Renamed.

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: @elBoberido That's how it's defined [in the STL](https://en.cppreference.com/w/cpp/iterator/size). Therefore I would leave it as it is.

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: Good catch, I removed the duplication & moved the generic template to the file.

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,429 @@
...
+    return std::pointer_traits<Ptr>::to_address(p);
+}
+
+template <typename Ptr, typename... None>
+constexpr auto to_address(const Ptr& p, None...) noexcept
+{
+    return to_address(p.operator->());
+}
+
+} // namespace detail
```
elBoberido, line 215: Quite a lot of the functions in `detail` have generic names which could lead to issues later on. I'm myself not sure what would be the best to do. Prefix, another namespace or just keep it as is?

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,429 @@
...
+
+/// @brief  C++11 compatible span implementation. The class template span describes an object that can refer to a
+///         contiguous sequence of objects with the first element of the sequence at position zero. A span can either
+///         have a static extent, in which case the number of elements in the sequence is known at compile-time and
+///         encoded in the type, or a dynamic extent. If a span has dynamic extent, a typical implementation holds two
+///         members: a pointer to T and a size. A span with static extent may have only one member: a pointer to T.
+///         http://eel.is/c++draft/views contains the latest C++20 draft
+/// @tparam T - element type; must be a complete object type that is not an abstract class type
+/// @tparam Extent - the number of elements in the sequence, or std::dynamic_extent if dynamic
```
elBoberido, line 224:
```suggestion
/// @tparam Extent - the number of elements in the sequence, or 'iox::DYNAMIC_EXTENT' if dynamic
```

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,429 @@
...
+    // constructors, copy, assignment, and destructor
+
+    constexpr span() noexcept = delete;
+
+    /// @brief Constructs a span that is a view over the range [first, first + count);
+    /// @tparam It Type of the iterator
+    /// @param first iterator of where to start creating the span
+    /// @param count number of elements counting from first
+    template <typename It>
+    constexpr span(It first, uint64_t count) noexcept;
```
elBoberido, line 251: Nitpicking. In other occasions  the full name is used, i.e. here `Iterator`. No strong feelings about this though

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,429 @@
...
+    /// data() == std::data(arr)
+    /// @tparam U implicit element type stored in the uninitialized array
+    /// @tparam N implicit capacity of the uninitialized array
+    /// @param array uninitialized array used to create the span
+    template <typename U,
+              uint64_t N,
+              template <typename, uint64_t>
+              class Buffer,
+              typename = detail::enable_if_compatible_array_t<iox::UninitializedArray<U, N, Buffer>&, T, Extent>>
+    constexpr explicit span(iox::UninitializedArray<U, N, Buffer>& array) noexcept;
```
elBoberido, line 280: Just as food for thought to get rid of the dependency to `UninitializedArray`. One could have something like
```cpp
template <typename T>
SpanCompatibleContainerAdapter {
    static_assert(always_false_v<T>, "Please specialize for your container");
    
    // some methods which are called by the ctor
}

// in the span
template <typename Container, typename = detail::enable_if_compatible_dynamic_container_t<Container&, T, Extent>>
constexpr explicit span(SpanCompatibleContainerAdapter<Container>& container) noexcept;
```

The nice thing about this approach would be to be able to create new Container and make them useable with the `span` without having to adjust the `span` itself.

Disclainer: I did not yet look at the implementation so the above might not make sense or would need adjustments.

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,429 @@
...
+    constexpr T& front() const noexcept;
+
+    /// @brief access the last element
+    /// @return reference to first element
+    constexpr T& back() const noexcept;
+
+    /// @brief accesses an element of the sequence
+    /// @param index which should be accessed
+    /// @return reference to respective element
+    constexpr T& operator[](uint64_t index) const noexcept;
```
elBoberido, line 391: Not part of the STL but since other container have an `at` which is also bounds checked, should we add such a method. Follow up question, should we have an `unchecked_at` where the bounds check is omitted but the user has to ensure that the access is within bounds? This would be useful for code in the hot path.

elBoberido (commented):
Another intermediate step. Impl, tests and iterators are still missing

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+class span_iterator
+{
+  public:
+    using iterator_category = std::random_access_iterator_tag;
+    using value_type = std::remove_cv_t<T>;
+    using difference_type = std::ptrdiff_t;
+    using pointer = T*;
+    using reference = T&;
+
+    constexpr span_iterator() = default;
```
elBoberido, line 36: Is there a use case for a defaulted span_iterator?

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+        : span_iterator(start, end, start)
+    {
+    }
+
+    constexpr span_iterator(const pointer begin, const pointer end, pointer current)
+        : m_begin(begin)
+        , m_end(end)
+        , m_current(current)
+    {
+    }
```
elBoberido, line 53: How about a range check, e.g. `begin <= end` and `current >= begin && current <=end`?

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+    explicit constexpr operator span_iterator<const T>() const noexcept
+    {
+        return {m_begin, m_end, m_current};
+    }
+
+    constexpr reference operator*() const noexcept
+    {
+        assert(m_begin && m_end);
+        assert(m_begin <= m_current && m_current < m_end);
```
elBoberido, line 63: These checks could be omitted if they are only performed when ´m_current´ is changed

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: I think for now leave it as it is, but I would move the generic things outside in future. Is that ok from your side?

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: Which occasions do you mean?

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: @elBoberido That's a nice future improvement, thanks! Let's for now stick with the direct dependency.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+    return uninit_array.begin();
+}
+
+template <typename T, uint64_t Extent>
+template <typename It>
+inline constexpr span<T, Extent>::span(It first, uint64_t count) noexcept
+    : span_storage_t(count)
+    , m_data(detail::to_address(first))
+{
+    assert(Extent == DYNAMIC_EXTENT || Extent == count);
```
elfenpiff, line 54: Since it is a `Extent` is a compile time value shouldn't this be a `static_assert`?

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+    using reference = T&;
+
+    constexpr span_iterator() = default;
+    constexpr span_iterator(const span_iterator& other) = default;
+    constexpr span_iterator(span_iterator&& other) noexcept = default;
+    constexpr span_iterator& operator=(const span_iterator& other) = default;
+    constexpr span_iterator& operator=(span_iterator&& other) noexcept = default;
+    ~span_iterator() = default;
+
+    constexpr span_iterator(pointer start, const pointer end)
```
elfenpiff, line 43: `noexcept` is missing

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+        : span_iterator(start, end, start)
+    {
+    }
+
+    constexpr span_iterator(const pointer begin, const pointer end, pointer current)
```
elfenpiff, line 48: `noexcept` is missing

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+        span_iterator ret = *this;
+        --*this;
+        return ret;
+    }
+
+    constexpr span_iterator& operator+=(const difference_type n) noexcept
+    {
+        if (n != 0)
+        {
+            assert(m_begin && m_current && m_end);
```
elBoberido, line 112: Why doing this not unconditional? It feels weird to add and assign 0 to a `nullptr`. Alternatively an early return for `n == 0` would also work

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+        if (n > 0)
+        {
+            assert(m_end - m_current >= n);
+        }
+        if (n < 0)
```
elBoberido, line 118: With the early return this would become just an `else` branch

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    constexpr span_iterator operator-(const difference_type n) const noexcept
+    {
+        span_iterator ret = *this;
+        ret -= n;
+        return ret;
+    }
+
+    template <class Type2, std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
```
elBoberido, line 164: Just as food for thought. Initially I did not realize why `Type2`. What do you think about `OtherType`. No strong feelings though

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+    constexpr difference_type operator-(const span_iterator<Type2>& rhs) const noexcept
+    {
+        assert(m_begin == rhs.m_begin && m_end == rhs.m_end);
+        return m_current - rhs.m_current;
```
elBoberido, line 168: Is it possible that `m_current` is still a `nullptr` at this time?

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: I think you are mistaken `operator[]` is [part of the STL](https://en.cppreference.com/w/cpp/container/span/operator_at).

Regarding

1. `at`: `std::span` does not have this method, so I would not add it either
2. `at_unchecked`: I would not add it right away due to the principle `Always put safety over performance`

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+    constexpr reference operator[](const difference_type n) const noexcept
+    {
+        return *(*this + n);
+    }
+
+    template <class Type2, std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
+    constexpr bool operator==(const span_iterator<Type2>& rhs) const noexcept
+    {
+        assert(m_begin == rhs.m_begin && m_end == rhs.m_end);
```
elBoberido, line 179: Thinking out loud. Can this be removed?

iceoryx_dust/vocabulary/include/iox/detail/span.inl
mossmaurice: @elfenpiff Unfortunately that's not possible as count is passed during runtime:

```
iceoryx/iceoryx_dust/vocabulary/include/iox/detail/span.inl:54:44: error: non-constant condition for static assertion
     static_assert(Extent == DYNAMIC_EXTENT || Extent == count, "foo");
                   ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
/home/simon.hoinkis/iceoryx/iceoryx_dust/vocabulary/include/iox/detail/span.inl:54:44: error: ‘count’ is not a constant expression
```

iceoryx_dust/vocabulary/include/iox/span.hpp
elBoberido: Okay, my comment could be mistaken. I meant that `at` is not part of the STL :sweat_smile: 

Well, due to `Always put safety over performance` we have bound checks by default for `operator[]`. There are already a few container who have an `unchecked_at` ... or maybe `at_unchecked`. No strong feelings here though. Can also be added later on.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+{
+    static_assert(Count <= Extent, "Count must not exceed Extent");
+    static_assert(Extent != DYNAMIC_EXTENT || Count <= size(), "Count must not exeed size");
+    return {data(), Count};
+}
+
+template <typename T, uint64_t Extent>
+inline constexpr span<T, DYNAMIC_EXTENT> span<T, Extent>::first(uint64_t count) const noexcept
+{
+    static_assert(count <= size(), "");
```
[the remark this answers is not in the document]
mossmaurice: @elfenpiff Good catch, I made this an `assert` due to the reason described [here](https://github.com/eclipse-iceoryx/iceoryx/pull/1785#discussion_r1054828379).

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: Thanks, fixed.

elBoberido (commented):
Only 600 lines missing with the impl and tests :)

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+    constexpr span_iterator& operator=(span_iterator&& other) noexcept = default;
+    ~span_iterator() = default;
+
+    constexpr span_iterator(pointer start, const pointer end)
+        : span_iterator(start, end, start)
+    {
+    }
+
+    constexpr span_iterator(const pointer begin, const pointer end, pointer current)
+        : m_begin(begin)
```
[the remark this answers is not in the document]
mossmaurice: @elfenpiff I think so far throughout the iceoryx code base `m_begin(begin)` was used. If you don't mind I would like to stick to that.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+    : span(iox::data(array), N)
+{
+}
+
+template <typename T, uint64_t Extent>
+template <typename U, uint64_t N, template <typename, uint64_t> class Buffer, typename>
+inline constexpr span<T, Extent>::span(const iox::UninitializedArray<U, N, Buffer>& array) noexcept
+    : span(iox::data(array), N)
+{
+}
```
elBoberido, line 85: Since `data` and `size` are already specialized for `T (&array)[N]` and `UninitializedArray`, are these c`tors really necessary?

elBoberido (commented):
Now it's only the tests

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    explicit constexpr operator span_iterator<const T>() const noexcept
+    {
+        return {m_begin, m_end, m_current};
+    }
+
+    constexpr reference operator*() const noexcept
+    {
+        assert(m_begin && m_end);
```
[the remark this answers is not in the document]
mossmaurice: @elfenpiff That is unfortunately not possible as `Require` is not `constexpr`:

```
iceoryx/iceoryx_hoofs/include/iceoryx_hoofs/cxx/requires.hpp:46:45: error: call to non-‘constexpr’ function ‘void iox::cxx::internal::Require(bool, const char*, int, const char*, const char*)’
```


`assert`s can be executed both during [compile- and run-time](https://stackoverflow.com/questions/32401256/how-can-i-do-a-runtime-assert-in-a-constexpr-function/32403291#32403291):

https://github.com/ApexAI/iceoryx/blob/25b09bb053cd789aff43d04eb18e4200052844ff/iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp#L19-L20

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: I can't think of any use-cases, removed.

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: Let's add it later.

iceoryx_hoofs/container/include/iox/detail/uninitialized_array_type_traits.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+};
+
+template <typename T, uint64_t N, template <typename, uint64_t> class Buffer>
+struct is_iox_array<iox::UninitializedArray<T, N, Buffer>> : std::true_type
+{
+};
+
+template <typename T>
+using is_not_iox_array_t = iox::negation<is_iox_array<std::decay_t<T>>>;
```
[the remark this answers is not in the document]
mossmaurice: Let's keep it as it ensures better readability as you stated. Added the doxygen.

iceoryx_hoofs/container/include/iox/uninitialized_array.hpp
```diff
@@ -133,6 +133,14 @@ class UninitializedArray final
     Buffer<ElementType, Capacity> m_buffer;
 };
 
+/// @brief Returns N
+/// @tparam T Type of the iox::UninitializedArray
+/// @tparam N Size of the iox::UninitializedArray
+/// @param An iox::UninitializedArray
+/// @return Size of the iox::UninitializedArray
+template <typename T, uint64_t N, template <typename, uint64_t> class Buffer>
+constexpr uint64_t size(const UninitializedArray<T, N, Buffer>&) noexcept;
```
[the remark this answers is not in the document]
mossmaurice: @elfenpiff I know what you mean. However, this is is consistent with a plain array, see here:

https://github.com/eclipse-iceoryx/iceoryx/blob/a6133e4b5b5d2d3ac93002684d64377c6ad3803a/iceoryx_hoofs/primitives/include/iox/size.hpp#L33-L40

> Is it possible to use capacity here in this context?

I fear not, the STL [treats this topic in the same way](https://en.cppreference.com/w/cpp/iterator/size).

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: I think it's safer if we check them, when actually using them.

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: @elBoberido Well, because the checks are only needed under certain conditions :upside_down_face: 

1. When `n` is zero, no arithmetic operation is performed on the iterator, hence the check is not needed
2. When decrementing the iterator (negative `n`) the check is not needed because the iterator can go before the beginning of the `span`
3. When incrementing the iterator, the check is not needed because the iterator can go behind the end of `span`

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: @elBoberido The first `assert` checks `m_begin` and `m_end` in both `span_iterator`s are equal. That means both `span_iterator`s are constructed with the same `span`. Hence, `m_current` can't be `nullptr` because the `span` is not empty, in such a case also `m_begin` and `m_end` would both be `nullptr`.

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: We can do a lot of things :grin: However, in this case, I'd stick with the `assert` to make sure both `span_iterator`s refer to the same `span`, otherwise UB can happen.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+
+template <typename T, uint64_t N, template <typename, uint64_t> class Buffer>
+inline constexpr T* data(UninitializedArray<T, N, Buffer>& uninit_array) noexcept
+{
+    return uninit_array.begin();
+}
+
+template <typename T, uint64_t Extent>
+template <typename It>
+inline constexpr span<T, Extent>::span(It first, uint64_t count) noexcept
```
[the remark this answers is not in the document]
mossmaurice: There is no chance to check if users are ~dumb~ uninformed users :stuck_out_tongue:

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+    assert(Extent == DYNAMIC_EXTENT || Extent == count);
+}
+
+template <typename T, uint64_t Extent>
+template <typename It, typename End, typename>
+inline constexpr span<T, Extent>::span(It begin, End end) noexcept
+    : span(begin, static_cast<uint64_t>(end - begin))
+{
+    static_assert(begin <= end, "begin shall not be smaller than end");
+}
```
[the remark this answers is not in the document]
mossmaurice: Good catch, added.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
mossmaurice: Yes, they are. Otherwise such code will not compile `span<const int> const_sut(array)`.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+    static_assert(Extent == DYNAMIC_EXTENT || Extent > 0, "Extent must not be 0");
+    assert(Extent != DYNAMIC_EXTENT || !empty());
+    return *data();
+}
+
+template <typename T, uint64_t Extent>
+inline constexpr T& span<T, Extent>::back() const noexcept
+{
+    static_assert(Extent == DYNAMIC_EXTENT || Extent > 0, "Extent must not be 0");
+    static_assert(Extent != DYNAMIC_EXTENT || !empty(), "Span must not be empty");
```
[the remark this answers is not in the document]
mossmaurice: Correct, the latter needs to be an `assert` because `empty` can also be runtime dependent. I changed also all other occurrences.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
```diff
@@ -0,0 +1,271 @@
...
+template <typename T, uint64_t Extent>
+constexpr uint64_t span<T, Extent>::extent;
+
+// object representation
+template <typename T, uint64_t X>
+inline span<const uint8_t, (X == DYNAMIC_EXTENT ? DYNAMIC_EXTENT : sizeof(T) * X)> as_bytes(span<T, X> s) noexcept
+{
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    return {reinterpret_cast<const uint8_t*>(s.data()), s.size_bytes()};
+}
```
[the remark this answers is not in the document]
mossmaurice: @elBoberido Makes sense, added.

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,429 @@
...
+#include "iox/type_traits.hpp"
+#include "iox/uninitialized_array.hpp"
+
+#include <algorithm>
+#include <cassert>
+#include <cstdint>
+#include <cstdio>
+#include <limits>
+
+using namespace std;
```
elBoberido, line 32: Please remove this :)

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
elBoberido: Sure, but since one line further down pointer are compared I thought it is superfluous. On the other hand, if both `m_current` are `nullptr` it would lead to a wrong result when the iterators accidentally refer to different `span`

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
elBoberido: But this check is done for all conditions except `1` and then there is a potential assignment to a `nullptr` although it is `0`. In theory it should not cause any issue but I'm quite cautious with such things.

iceoryx_dust/vocabulary/include/iox/detail/span.inl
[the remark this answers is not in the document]
elBoberido: I guess then the same is then required for the c'tor on line 50?

iceoryx_dust/test/moduletests/test_vocabulary_span.cpp
```diff
@@ -0,0 +1,346 @@
...
+    ::testing::Test::RecordProperty("TEST_ID", "05db30c2-e13d-4116-ba05-668b30ba4a23");
+    std::vector<int> vector = {1, 1, 2, 3, 5, 8};
+
+    span<int> dyn_sut(vector.begin(), vector.size());
+
+    EXPECT_EQ(vector.data(), dyn_sut.data());
+    EXPECT_EQ(vector.size(), dyn_sut.size());
+    for (size_t i = 0; i < dyn_sut.size(); ++i)
+    {
+        EXPECT_EQ(vector[i], dyn_sut[i]);
```
elBoberido, line 50: While I like the Fibonacci sequence it would be good to have an initial check whether the creation of the `span` does not corrupt the container. With the current check this is not ensured. It would be nice to have e.g. a vector from 0 to 5 and to loop could also check for 0 to 5. Additionally, the data pointer and size of the vector could be obtained before the `span` is created. Maybe I'm a bit paranoid but one does never know :sweat_smile: ... and it has to be done only once ... alternatively, there could be a
```cpp
const std::vector<int> expectedData = {1, 1, 2, 3, 5, 8};
std::vector<int> vector = expectedData;
```
and the checks could be performed with `expectedData`

iceoryx_dust/vocabulary/include/iox/detail/span_iterator.hpp
mossmaurice: Added.

iceoryx_dust/test/moduletests/test_vocabulary_span.cpp
mossmaurice: Done.

iceoryx_dust/vocabulary/include/iox/span.hpp
```diff
@@ -0,0 +1,427 @@
...
+    }
+    span_storage& operator=(span_storage&& other) noexcept
+    {
+        if (this != &other)
+        {
+            m_size = other.m_size;
+            other.m_size = 0;
+        }
+        return *this;
+    }
```
elBoberido, line 187: Just thinking out loud. We can keep it as is but is it really necessary to manually implement the move c'tor and move assignment operator? Since the `span` does not do resource management, the default move c'tor and assignment operators would also do the job and mirror the non `DYNAMIC_EXTENT` behavior.

Btw, it also seems to be the behavior of `std::span`. I tried out this
```cpp
#include <iostream>
#include <span>
#include <vector>

void print(std::span<int, std::dynamic_extent>&& s) {
    std::cout << "====" << std::endl;
    for(auto i: s) {
        std::cout << i << " ";
    }
    std::cout << std::endl;
}

int main() {
    std::vector<int> v {1, 1, 2, 3, 5, 8};
    std::span<int, std::dynamic_extent> s{v};

    print(std::move(s));
    print(std::move(s));

    return EXIT_SUCCESS;
}
```
and got two times `1 1 2 3 5 8`.

iceoryx_dust/test/moduletests/test_vocabulary_span.cpp
```diff
@@ -0,0 +1,347 @@
...
+    IOX_MAYBE_UNUSED constexpr span<const int> span(arr);
+
+    // Explicitly not use EXPECT_TRUE here to be able to execute the test case during compile-time
+    // 'static_assert' is not possible as not being supported with GCC5
+    assert(1 == span.begin()[0]);      // First element needs to be '1'
+    assert(1 == *(span.begin() += 0)); // First element needs to be '1'
+    assert(6 == *(span.begin() += 1)); // Second element needs to be '6'
+
+    assert(1 == *((span.begin() + 1) -= 1)); // First element needs to be '1'
+    assert(6 == *((span.begin() + 1) -= 0)); // Second element needs to be '6'
```
elBoberido, line 328: In this case please use `static_assert` and disable the test for gcc5 with an ifdef

iceoryx_dust/vocabulary/include/iox/span.hpp
mossmaurice: @elBoberido Good pointed. They are defaulted now.

iceoryx_dust/test/moduletests/test_vocabulary_span.cpp
mossmaurice: @elBoberido IIRC for some reason it's not only GCC5 but also one of the sanitiser jobs wasn't able to build this `constexpr`. I assume that has something to do with the optimization levels. Would you be ok with keeping the current approach?

elBoberido: @mossmaurice `assert` is only a runtime check and `assert(false)` will happily compile. Do you know which job had problems? This indicates that something is wrong with the constexpr.

mossmaurice: @elBoberido As I stated in several threads above, `assert`s can be executed both during compile- and run-time, see [this Godbolt snippet](https://godbolt.org/z/13dThfbd9). If you don't mind I would go ahead and merge it as it is.

mossmaurice (dismissed):

iceoryx_dust/test/moduletests/test_vocabulary_span.cpp
elBoberido: @mossmaurice the linked example clearly shows that it is runtime only. The code compiles and only during runtime the assertion is checked. If it would be compile time, the output would be similar to this
```
<source>: In function 'int main()':
<source>:12:19: error: static assertion failed
   12 |     static_assert(false);
      |                   ^~~~~
ASM generation compiler returned: 1
<source>: In function 'int main()':
<source>:12:19: error: static assertion failed
   12 |     static_assert(false);
      |                   ^~~~~
Execution build compiler returned: 1
```
but with the `assert` it is
```
ASM generation compiler returned: 0
Execution build compiler returned: 0
Program returned: 139
output.s: /app/example.cpp:9: int main(): Assertion `false' failed.
 compile-time
```
I don't know why the string `compile-time` time occurs in the output. It is also there when there are no assertions. The assert itself occurred at runtime which is indicated by `Program returned: 139`

elBoberido: @mossmaurice I take (almost) everything back I said about `assert` and `only runtime`. With C++14 it can indeed be used in a constexpr and it still works when the `assert` can be evaluated to `true` at compile time. But in your example it has lead to a runtime check and the string `compile-time` comes from the `std::cout`. I overlooked that before.

This godbolt snippet shows how it works with constexpr.
https://godbolt.org/z/s83TT54d9

elBoberido (dismissed):

mossmaurice (approved):

mossmaurice (approved):

elBoberido (approved):

PR #1820, opened 2022-12-07
iox-1732 implement bump allocator

### LINKED ISSUE #1732: Implement BumpAllocator
## Brief feature description

Currently we have (at least) two classes to allocate memory: `posix::Allocator` and `cxx::static_storage`, where `cxx::static_storage` is basically `posix::Allocator` but with internal storage. To have a more generic and reusable allocator, we can remove the internal storage from `cxx::static_storage`, merge it with `posix::Allocator` and rename it to `BumpAllocator` (as known in literature). This allocator class can use the `unintialized_array` (#1614) underneath.

For the `storable_function`, this means that the storage needs then to be moved to the `storable_function` and the `BumpAllocator` needs to be initialized with a pointer to this storage.

FerdinandSpitzschnueffler:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR implements the `BumpAllocator`. Therefor the `Allocator` form the shared memory object directory is moved to `memory`, `deallocate` is implemented and the option to finalize the allocation is moved to the shared memory object. Additionally, the `static_storage` is moved to `iceoryx_dust` and a simple byte array is used in the `storable_function` instead.

## References

- Closes #1732

### DIFF
--- a/iceoryx_binding_c/test/moduletests/test_notification_info.cpp
+++ b/iceoryx_binding_c/test/moduletests/test_notification_info.cpp
@@ -102,7 +102,7 @@ class iox_notification_info_test : public Test
     static constexpr uint32_t CHUNK_SIZE = 128U;
     static constexpr size_t MEMORY_SIZE = 1024 * 1024 * 100;
     uint8_t m_memory[MEMORY_SIZE];
-    Allocator m_memoryAllocator{m_memory, MEMORY_SIZE};
+    BumpAllocator m_memoryAllocator{m_memory, MEMORY_SIZE};
     MePooConfig m_mempoolconf;
     MemoryManager m_memoryManager;
     SubscriberOptions m_subscriberOptions{MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY, 0U};
--- a/iceoryx_dust/include/iceoryx_dust/internal/cxx/static_storage.hpp
+++ b/iceoryx_dust/include/iceoryx_dust/internal/cxx/static_storage.hpp
@@ -14,8 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-#ifndef IOX_HOOFS_STATIC_STORAGE_HPP
-#define IOX_HOOFS_STATIC_STORAGE_HPP
+#ifndef IOX_DUST_STATIC_STORAGE_HPP
+#define IOX_DUST_STATIC_STORAGE_HPP
 
 #include <cstdint>
 #include <cstring>
@@ -109,5 +109,5 @@ class static_storage final
 } // namespace cxx
 } // namespace iox
 
-#include "iceoryx_hoofs/internal/cxx/static_storage.inl"
-#endif // IOX_HOOFS_STATIC_STORAGE_HPP
+#include "iceoryx_dust/internal/cxx/static_storage.inl"
+#endif // IOX_DUST_STATIC_STORAGE_HPP
--- a/iceoryx_dust/include/iceoryx_dust/internal/cxx/static_storage.hpp
+++ b/iceoryx_dust/include/iceoryx_dust/internal/cxx/static_storage.hpp
@@ -98,7 +98,7 @@ class static_storage final
     static constexpr uint64_t allocation_size() noexcept;
 
   private:
-    /// @NOLINTJUSTIFICATION @todo iox-#1732 will be refactored with BumpAllocator
+    /// AXIVION Next Construct AutosarC++19_03-A18.1.1 : safe access is guaranteed since the c-array is wrapped inside the static_storage
     /// @NOLINTNEXTLINE(hicpp-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays)
     alignas(Align) uint8_t m_bytes[Capacity];
     void* m_ptr{nullptr};
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
@@ -18,6 +18,7 @@
 #define IOX_HOOFS_STORABLE_FUNCTION_HPP
 
 #include "iceoryx_hoofs/cxx/type_traits.hpp"
+#include "iceoryx_hoofs/iceoryx_hoofs_types.hpp"
 #include "iceoryx_hoofs/internal/cxx/static_storage.hpp"
 
 #include <iostream>
@@ -155,7 +156,9 @@ class storable_function<Capacity, signature<ReturnType, Args...>> final
   private:
     operations m_operations; // operations depending on type-erased callable (copy, move, destroy)
 
-    StorageType m_storage;                              // storage for the callable
+    // AXIVION Next Construct AutosarC++19_03-A18.1.1 : safe access is guaranteed since the c-array is wrapped inside the storable_function
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, hicpp-avoid-c-arrays)
+    byte_t m_storage[Capacity];                         // storage for the callable
     void* m_callable{nullptr};                          // pointer to stored type-erased callable
     ReturnType (*m_invoker)(void*, Args&&...){nullptr}; // indirection to invoke the stored callable,
                                                         // nullptr if no callable is stored
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 - 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2020 - 2023 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -193,6 +193,9 @@ class storable_function<Capacity, signature<ReturnType, Args...>> final
     // m_invoker is initialized with this function and has to work with functors as well
     // (functors may change due to invocation)
     static ReturnType invokeFreeFunction(void* callable, Args&&... args) noexcept;
+
+    template <typename T>
+    static constexpr void* safeAlign(byte_t* startAddress);
 };
 
 /// @brief swap two storable functions
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
@@ -19,7 +19,6 @@
 
 #include "iceoryx_hoofs/cxx/type_traits.hpp"
 #include "iceoryx_hoofs/iceoryx_hoofs_types.hpp"
-#include "iceoryx_hoofs/internal/cxx/static_storage.hpp"
 
 #include <iostream>
 #include <utility>
@@ -49,7 +48,6 @@ template <uint64_t Capacity, typename ReturnType, typename... Args>
 class storable_function<Capacity, signature<ReturnType, Args...>> final
 {
   public:
-    using StorageType = static_storage<Capacity>;
     using signature_t = signature<ReturnType, Args...>;
 
     /// @brief construct from functor (including lambdas)
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
@@ -166,22 +166,23 @@ inline void swap(storable_function<Capacity, T>& f, storable_function<Capacity, 
     f.swap(g);
 }
 
+template <typename T>
+void* allocate(byte_t* startAddress)
+{
+    uint64_t alignedPosition = cxx::align(reinterpret_cast<uint64_t>(startAddress), alignof(T));
+    return reinterpret_cast<void*>(alignedPosition);
+}
+
 template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename Functor, typename>
 inline void storable_function<Capacity, signature<ReturnType, Args...>>::storeFunctor(const Functor& functor) noexcept
 {
     using StoredType = typename std::remove_reference<Functor>::type;
-    auto ptr = m_storage.template allocate<StoredType>();
-    cxx::Expects(ptr != nullptr);
-
-    // AXIVION Next Construct AutosarC++19_03-A18.5.10: false positive, m_storage.allocate<T>
-    // provides a properly aligned ptr and sufficient memory (static_storage satisfies this)
-    /// @NOLINTJUSTIFICATION ownership is encapsulated in storable_function and will be released in destroy
-    /// @NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
-    ptr = new (ptr) StoredType(functor);
+    m_callable = allocate<StoredType>(m_storage);
 
     // erase the functor type and store as reference to the call in storage
-    m_callable = ptr;
+    new (m_callable) StoredType(functor);
+
     m_invoker = &invoke<StoredType>;
     m_operations.copyFunction = &copy<StoredType>;
     m_operations.moveFunction = &move<StoredType>;
@@ -194,19 +195,13 @@ template <typename CallableType>
 inline void storable_function<Capacity, signature<ReturnType, Args...>>::copy(const storable_function& src,
                                                                               storable_function& dest) noexcept
 {
-    auto ptr = dest.m_storage.template allocate<CallableType>();
-    cxx::Expects(ptr != nullptr);
+    dest.m_callable = allocate<CallableType>(dest.m_storage);
 
     // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
     const auto obj = static_cast<CallableType*>(src.m_callable);
     cxx::Expects(obj != nullptr); // should not happen unless src is incorrectly used after move
 
-    // AXIVION Next Construct AutosarC++19_03-A18.5.10: false positive, m_storage.allocate<T>
-    // provides a properly aligned ptr and sufficient memory (static_storage satisfies this)
-    /// @NOLINTJUSTIFICATION ownership is encapsulated in storable_function and will be released in destroy
-    /// @NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
-    ptr = new (ptr) CallableType(*obj);
-    dest.m_callable = ptr;
+    new (dest.m_callable) CallableType(*obj);
     dest.m_invoker = src.m_invoker;
 }
 
@@ -217,19 +212,14 @@ template <typename CallableType>
 inline void storable_function<Capacity, signature<ReturnType, Args...>>::move(storable_function& src,
                                                                               storable_function& dest) noexcept
 {
-    auto ptr = dest.m_storage.template allocate<CallableType>();
-    cxx::Expects(ptr != nullptr);
+    //  m_callable = static_cast<void*>(align(reinterpret_cast<uint64_t>(m_storage), alignof(CallableType)));
+    dest.m_callable = allocate<CallableType>(dest.m_storage);
 
     // AXIVION Next Construct AutosarC++19_03-M5.2.8: type erasure - conversion to compatible type
     const auto obj = static_cast<CallableType*>(src.m_callable);
     cxx::Expects(obj != nullptr); // should not happen unless src is incorrectly used after move
 
-    // AXIVION Next Construct AutosarC++19_03-A18.5.10: false positive, m_storage.allocate<T>
-    // provides a properly aligned ptr and sufficient memory (static_storage satisfies this)
-    /// @NOLINTJUSTIFICATION ownership is encapsulated in storable_function and will be released in destroy
-    /// @NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
-    ptr = new (ptr) CallableType(std::move(*obj));
-    dest.m_callable = ptr;
+    new (dest.m_callable) CallableType(std::move(*obj));
     dest.m_invoker = src.m_invoker;
     src.m_operations.destroy(src);
     src.m_callable = nullptr;
@@ -246,7 +236,6 @@ inline void storable_function<Capacity, signature<ReturnType, Args...>>::destroy
         const auto ptr = static_cast<CallableType*>(f.m_callable);
         // AXIVION Next Construct AutosarC++19_03-A5.3.2: ptr is guaranteed not to be nullptr
         ptr->~CallableType();
-        f.m_storage.deallocate();
     }
 }
 
--- a/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
+++ b/iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
@@ -172,14 +172,13 @@ inline void swap(storable_function<Capacity, T>& f, storable_function<Capacity, 
     f.swap(g);
 }
 
-template <typename T>
-void* allocate(byte_t* startAddress, uint64_t Capacity)
+template <typename T, uint64_t Capacity>
+void* allocate(byte_t* startAddress)
 {
+    static_assert(sizeof(T) + alignof(T) - 1 <= Capacity, "type does not fit into storage");
     // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast, performance-no-int-to-ptr) required for low level pointer alignment
     uint64_t alignment = alignof(T);
     uint64_t alignedPosition = cxx::align(reinterpret_cast<uint64_t>(startAddress), alignment);
-    cxx::Expects(alignedPosition + sizeof(T) - reinterpret_cast<uint64_t>(startAddress) < Capacity
-                 && "type does not fit into storage");
     return reinterpret_cast<void*>(alignedPosition);
     // NOLINTEND(cppcoreguidelines-pro-type-reinterpret-cast, performance-no-int-to-ptr) required for low level pointer alignment
 }
@@ -189,7 +188,7 @@ template <typename Functor, typename>
 inline void storable_function<Capacity, signature<ReturnType, Args...>>::storeFunctor(const Functor& functor) noexcept
 {
     using StoredType = typename std::remove_reference<Functor>::type;
-    m_callable = allocate<StoredType>(m_storage, Capacity);
+    m_callable = allocate<StoredType, Capacity>(m_storage);
 
     // erase the functor type and store as reference to the call in storage
     new (m_callable) StoredType(functor);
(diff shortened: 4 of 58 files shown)


iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -166,22 +172,28 @@ inline void swap(storable_function<Capacity, T>& f, storable_function<Capacity,
     f.swap(g);
 }
 
+template <typename T>
+void* allocate(byte_t* startAddress, uint64_t Capacity)
+{
+    // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast, performance-no-int-to-ptr) required for low level pointer alignment
+    uint64_t alignment = alignof(T);
+    uint64_t alignedPosition = cxx::align(reinterpret_cast<uint64_t>(startAddress), alignment);
+    cxx::Expects(alignedPosition + sizeof(T) - reinterpret_cast<uint64_t>(startAddress) < Capacity
```
MatthiasKillat, line 181: @elBoberido @FerdinandSpitzschnueffler

Now we have a regression. Previously it could be detected at compile time that the allocation will fail due to insufficient size for a callable (if used with some type that was too large). This was one reason for the static storage. Now we have checks at runtime for this, which leads to some issues which are just not necessary.

1.) Performance overhead (if the check is enabled, which it currently is)
2.) We may not know at compile time that we have insufficient space and cannot guarantee that it runs at compile time (with sufficient stack, but this is still required).

This is not good, we want to follow a fail early philosophy and you cannot really fail much earlier than compile time. Note that we could add the compile time check here also directly, as it was done in `static_storage`. However, I do not quite see the improvement. We shifted more of the storage responsibility to `function`, instead of abstracting it in something like `static_storage`.

MatthiasKillat (commented):
Only reviewed the initial changes up to `cxx::function` changes. Will continue tomorrow.

iceoryx_posh/source/roudi/memory/mempool_collection_memory_block.cpp
```diff
@@ -51,8 +51,10 @@ uint64_t MemPoolCollectionMemoryBlock::alignment() const noexcept
 
 void MemPoolCollectionMemoryBlock::onMemoryAvailable(cxx::not_null<void*> memory) noexcept
 {
-    posix::Allocator allocator(memory, size());
-    auto memoryManager = allocator.allocate(sizeof(mepoo::MemoryManager), alignof(mepoo::MemoryManager));
+    BumpAllocator allocator(memory, size());
+    auto allocationResult = allocator.allocate(sizeof(mepoo::MemoryManager), alignof(mepoo::MemoryManager));
+    cxx::Expects(!allocationResult.has_error());
```
MatthiasKillat, line 56: Same pattern as before. If allocation fails we terminate. This may be tolerable here. It was not checked before and checking is better than continuing without resources.

It would be even better to continue, but this is probably not possible. The next best option is graceful termination, but we lack facilities to do so. This will be corrected soon. :-)

This pattern may occur more often and I will not further comment on it as there is no immediate solution here.

Just an observation, nothing to do for now (and no todo needed as searching for `Expects` is sufficient).

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+    ::testing::Test::RecordProperty("TEST_ID", "4252ddcc-05d4-499f-ad7c-30bffb420e08");
+    iox::BumpAllocator sut(memory, memorySize);
+    auto allocationResult = sut.allocate(5, MEMORY_ALIGNMENT);
+    ASSERT_FALSE(allocationResult.has_error());
+    auto* bla = static_cast<uint8_t*>(allocationResult.value());
+
+    allocationResult = sut.allocate(5, MEMORY_ALIGNMENT);
+    ASSERT_FALSE(allocationResult.has_error());
+    auto* bla2 = static_cast<uint8_t*>(allocationResult.value());
+    EXPECT_THAT(bla2 - bla, Eq(8U));
```
MatthiasKillat, line 116: That only shows that consecutive allocations are 8 bytes offset from each other but not that they are correctly aligned (more precisely it also assumes `bla2 > bla` which is OK for a bump allocator but not necessarily for others). Theoretically this condition does not need to hold, the gap could be larger and both addresses still correctly aligned. In this case some of the bytes are not used, but allocators do this all the time, depending on type.

The test should be changed to test the alignment of the addresses, meaning that they are multiples of whatever alignment is requested. The test name would be `allocationIsCorrectlyAligned`.

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+TEST_F(BumpAllocator_Test, allocateElementOfSizeZero)
```
MatthiasKillat, line 119:
```suggestion
TEST_F(BumpAllocator_Test, allocationOfZeroBytesTerminates)
```

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+    ::testing::Test::RecordProperty("TEST_ID", "17caa50c-94bf-4a1d-a1ec-dfda563caa0b");
+    iox::BumpAllocator sut(memory, memorySize);
+
+    // @todo iox-#1613 remove EXPECT_DEATH
+    // NOLINTBEGIN(hicpp-avoid-goto, cppcoreguidelines-avoid-goto, cert-err33-c, cppcoreguidelines-pro-type-vararg,
+    // hiccpp-vararg)
+    EXPECT_DEATH(IOX_DISCARD_RESULT(sut.allocate(0, MEMORY_ALIGNMENT)), ".*");
```
MatthiasKillat, line 127: We could also leave this as unspecified and hence not test it if we want to get rid of (death) tests. Allocating 0 bytes is not part of the contract.

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // NOLINTEND(hicpp-avoid-goto, cppcoreguidelines-avoid-goto, cert-err33-c, cppcoreguidelines-pro-type-vararg,
+    // hiccpp-vararg)
+}
+
+TEST_F(BumpAllocator_Test, allocateAfterDeallocateWorks)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "323fc1af-481f-4732-b7d3-fa32da389cef");
+    iox::BumpAllocator sut(memory, memorySize);
+    ASSERT_FALSE(sut.allocate(memorySize, 1).has_error());
```
MatthiasKillat, line 136: Maybe extend to a more general case with multiple allocations. However, `deallocate` reclaims (resets) them all anyway by design.

We should test if the following allocation would be the start address (or more precisely the next aligned address after this) of the memory the bump allocator manages (i.e. `memory`).

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    sut.deallocate();
+
+    auto allocationResult = sut.allocate(memorySize, 1);
+    ASSERT_FALSE(allocationResult.has_error());
+    int* bla = static_cast<int*>(allocationResult.value());
```
MatthiasKillat, line 142: We can fill the memory with something and check (but this was already done before, albeit without previous deallocation). It would be more important to check that we start with the "first" address before, i.e. the bump allocator is correctly reset.

Test name could be `deallocationResetsAllocator`.

MatthiasKillat (commented):
The tests need some rework.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
elfenpiff: I think we can improve this function so that it is compile time again.
1. `Capacity` becomes the second template argument
2. We add this `static_assert(sizeof(T) + alignof(T) - 1 <= Capacity, "type does not fit into storage");`

With this it should be possible to remove the `cxx::Expects`.

iceoryx_hoofs/memory/include/iox/bump_allocator.hpp
```diff
@@ -0,0 +1,64 @@
...
+    BumpAllocator& operator=(const BumpAllocator&) noexcept = delete;
+    BumpAllocator& operator=(BumpAllocator&&) noexcept = default;
+    ~BumpAllocator() noexcept = default;
+
+    /// @brief allocates on the memory supplied with the ctor
+    /// @param[in] size of the memory to allocate, must be greater than 0
+    /// @param[in] alignment of the memory to allocate
+    /// @return returns an expected containing a pointer to the allocated memory if memory is available, otherwise
+    /// BumpAllocatorError::OUT_OF_MEMORY
+    cxx::expected<void*, BumpAllocatorError> allocate(const uint64_t size, const uint64_t alignment) noexcept;
```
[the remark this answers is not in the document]
MatthiasKillat: I am just putting it up for debate. And Rust has better builtin support for its result types, e.g. `?` and match which makes it less cumbersome to use

iceoryx_posh/source/mepoo/mem_pool.cpp
```diff
@@ -41,18 +41,23 @@ constexpr uint64_t MemPool::CHUNK_MEMORY_ALIGNMENT;
...
     , m_minFree(numberOfChunks)
 {
     if (isMultipleOfAlignment(chunkSize))
     {
-        m_rawMemory = static_cast<uint8_t*>(chunkMemoryAllocator.allocate(
-            static_cast<uint64_t>(m_numberOfChunks) * m_chunkSize, CHUNK_MEMORY_ALIGNMENT));
-        auto memoryLoFFLi =
+        auto allocationResult = chunkMemoryAllocator.allocate(static_cast<uint64_t>(m_numberOfChunks) * m_chunkSize,
+                                                              CHUNK_MEMORY_ALIGNMENT);
+        cxx::Expects(!allocationResult.has_error());
```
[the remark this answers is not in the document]
elBoberido: @MatthiasKillat yes, this is handled during initialization and it should be ensured that enough memory is available.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
elBoberido: @MatthiasKillat yes, we replace the dependency on one class with another class but in total there is less code since one dependency had a large overlap with the one that replaced it.

iceoryx_hoofs/memory/source/bump_allocator.cpp
```diff
@@ -15,32 +15,25 @@
...
 // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
-void* Allocator::allocate(const uint64_t size, const uint64_t alignment) noexcept
+cxx::expected<void*, BumpAllocatorError> BumpAllocator::allocate(const uint64_t size, const uint64_t alignment) noexcept
 {
-    cxx::Expects(size > 0);
-
-    cxx::Expects(
-        !m_allocationFinalized
-        && "allocate() call after finalizeAllocation()! You are not allowed to acquire shared memory chunks anymore");
+    cxx::ExpectsWithMsg(size > 0, "Cannot allocate memory of size 0");
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: It is correct that `ExpectsWithMsg` causes termination (in the future only in debug mode). I removed it and return an error now.

iceoryx_binding_c/test/moduletests/test_notification_info.cpp
```diff
@@ -102,7 +102,7 @@ class iox_notification_info_test : public Test
     static constexpr uint32_t CHUNK_SIZE = 128U;
     static constexpr size_t MEMORY_SIZE = 1024 * 1024 * 100;
     uint8_t m_memory[MEMORY_SIZE];
-    Allocator m_memoryAllocator{m_memory, MEMORY_SIZE};
+    BumpAllocator m_memoryAllocator{m_memory, MEMORY_SIZE};
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I created #1849

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
FerdinandSpitzschnueffler: Since `BumpAllocator::allocate` returns now an `expected`, allocating 0 bytes is tested without an death test in `AllocateFailsWithZeroSize`.

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+    *bla = 123;
+    EXPECT_THAT(*bla, Eq(123));
+}
+
+TEST_F(BumpAllocator_Test, allocateEverythingWithMultipleElements)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "21d0fa61-54f9-41a0-8e53-e3448784497b");
+    iox::BumpAllocator sut(memory, memorySize);
+    for (size_t i = 0; i < memorySize; i += 32)
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Please see `AllocateCompleteMemoryWithEquallySizedChunksWorks` and `AllocateCompleteMemoryWithDifferentSizedChunksWorks`.

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+    int* bla = static_cast<int*>(allocationResult.value());
+    *bla = 123;
+    EXPECT_THAT(*bla, Eq(123));
+}
+
+TEST_F(BumpAllocator_Test, allocateEverythingWithSingleElement)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "f2e1085b-08fe-4b08-b022-0385b5a53fca");
+    iox::BumpAllocator sut(memory, memorySize);
+    auto allocationResult = sut.allocate(memorySize, 1);
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I reworked all the `BumpAllocator` tests. For this test, please see `AllocateCompleteMemoryAndStoreDataWorks`.

iceoryx_hoofs/test/moduletests/test_memory_bump_allocator.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    void* memory{nullptr};
+    size_t memorySize = 10016;
+};
+
+TEST_F(BumpAllocator_Test, allocateOneSmallElement)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "f689e95c-5743-4370-93f0-8a23b909c75a");
+    iox::BumpAllocator sut(memory, memorySize);
+    auto allocationResult = sut.allocate(sizeof(int), MEMORY_ALIGNMENT);
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Please see `AllocationIsCorrectlyAligned` and `AllocateSmallMemoryChunkAndStoreDataWorks`.

iceoryx_posh/source/roudi/memory/mempool_collection_memory_block.cpp
FerdinandSpitzschnueffler: Just for info: the `Expects` was hidden inside `Allocator::allocate` before.

iceoryx_hoofs/test/moduletests/test_posix_shared_memory_object.cpp
```diff
@@ -54,6 +54,8 @@ class SharedMemoryObject_Test : public Test
         // hiccpp-vararg)
         internal::CaptureStderr();
     }
+
+    static constexpr uint64_t MEMORY_ALIGNMENT{8};
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Confirmed.

iceoryx_hoofs/memory/include/iox/bump_allocator.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    /// @brief mark the memory as unused
+    void deallocate() noexcept;
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I tend to keep the current name.

iceoryx_hoofs/memory/include/iox/bump_allocator.hpp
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: I tend to keep the return type, also added another allocation error.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -166,22 +172,28 @@ inline void swap(storable_function<Capacity, T>& f, storable_function<Capacity,
     f.swap(g);
 }
 
+template <typename T>
+void* allocate(byte_t* startAddress, uint64_t Capacity)
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Changed the name to `safeAlign`.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
FerdinandSpitzschnueffler: @MatthiasKillat @elfenpiff I added the compile time check.

elfenpiff (dismissed):

iceoryx_dust/include/iceoryx_dust/internal/cxx/static_storage.hpp
```diff
@@ -14,8 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-#ifndef IOX_HOOFS_STATIC_STORAGE_HPP
-#define IOX_HOOFS_STATIC_STORAGE_HPP
+#ifndef IOX_DUST_STATIC_STORAGE_HPP
+#define IOX_DUST_STATIC_STORAGE_HPP
```
MatthiasKillat, line 19: @FerdinandSpitzschnueffler We may consider fixing the last 2 remaining Axivion warnings right here, I can add a commit myself. It would be the C array usage in this file (questionable rule but OK).

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.hpp
```diff
@@ -155,7 +154,9 @@ class storable_function<Capacity, signature<ReturnType, Args...>> final
   private:
     operations m_operations; // operations depending on type-erased callable (copy, move, destroy)
 
-    StorageType m_storage;                              // storage for the callable
+    // AXIVION Next Construct AutosarC++19_03-A18.1.1 : safe access is guaranteed since the c-array is wrapped inside the storable_function
```
MatthiasKillat, line 157: @elBoberido @FerdinandSpitzschnueffler I still think we need the storage abstraction in the future to avoid comments like this etc. I would like to distinguish between allocator and storage as different concepts.

Storage either magically exists or is provided by some other mechanism. In the end it abstracts memory, often but not necessarily contiguous. In a way it wraps memory but abstracts the access and other functionality.

An allocator is a means to get memory, or maybe storage. It can serve multiple allocations and internally generally does not need to manage contiguous memory.

The reason for storage abstraction is also to move to a more general concept where we decouple the storage and the object logic. Anyway, we can keep it for now but I do not see the benefit. Now the function implementation is more tightly coupled again to the storage concept and implements the core functionality (now called `safeAlign`).

Considerations for future refactoring. Resolve.

iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/storable_function.inl
```diff
@@ -309,14 +311,14 @@ template <uint64_t Capacity, typename ReturnType, typename... Args>
 template <typename T>
 inline constexpr uint64_t storable_function<Capacity, signature<ReturnType, Args...>>::required_storage_size() noexcept
 {
-    return StorageType::template allocation_size<T>();
+    return sizeof(T) + alignof(T) - 1;
```
MatthiasKillat, line 314: This could be made more exact if the alignment of storage is taken in account (may safe a little memory). But this worst case estimate is correct, just not tight in general. It assumes that we need to maximally realign by `alignof(T) - 1`, but generally it will be less.

It would look like the `align` method we refactored in helplets to get an aligned address from the start of storage and the look whether there is space for `sizeof(T)` from this address onward.

This is why it was previously part of the storage and hence knew about the alignment directly (and another reason for the abstraction). I consider reworking the storage concepts in the future to better reflect this, as it existed before it was functional but not fully developed.

TLDR: can be optimized but does not have to be right now. A change would need to take the starting address of `m_storage` into account.

iceoryx_hoofs/include/iceoryx_hoofs/internal/posix_wrapper/shared_memory_object.hpp
```diff
@@ -40,6 +40,14 @@ enum class SharedMemoryObjectError
     INTERNAL_LOGIC_FAILURE,
 };
 
+enum class SharedMemoryAllocationError
+{
+    REQUESTED_MEMORY_AFTER_FINALIZED_ALLOCATION,
+    NOT_ENOUGH_MEMORY,
+    REQUESTED_ZERO_SIZED_MEMORY
```
MatthiasKillat, line 47: Requesting zero memory should be a misuse case. We do not have to change it now as it is a more fundamental issue, but I think we should not mix errors that can happen during correct use and incorrect use. Note that some allocator implementations just treat requesting zero bytes as requesting one byte by convention, but we should not do this IMO.

I have some notion of this in the new error handling and I think the distinction is needed to develop better/safe code.

Nothing to do now.

iceoryx_hoofs/memory/include/iox/bump_allocator.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <cstdint>
+
+namespace iox
+{
+
+enum class BumpAllocatorError
+{
+    OUT_OF_MEMORY,
+    REQUESTED_ZERO_SIZED_MEMORY
```
MatthiasKillat, line 31: See my comment about misuse cases vs. legitimate errors. No change required now but fundamentally there should be a distinction.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object.cpp
```diff
@@ -166,17 +166,36 @@ SharedMemoryObject::SharedMemoryObject(SharedMemory&& sharedMemory,
...
 
-void* SharedMemoryObject::allocate(const uint64_t size, const uint64_t alignment) noexcept
+cxx::expected<void*, SharedMemoryAllocationError> SharedMemoryObject::allocate(const uint64_t size,
+                                                                               const uint64_t alignment) noexcept
 {
-    return m_allocator.allocate(size, alignment);
+    if (size == 0)
+    {
+        IOX_LOG(WARN) << "Cannot allocate memory of size 0.";
+        return cxx::error<SharedMemoryAllocationError>(SharedMemoryAllocationError::REQUESTED_ZERO_SIZED_MEMORY);
```
MatthiasKillat, line 175: IMO this should be an assert like check, it is a violation of the precondition size>0.

Precondition violations indicate bugs and should never lead to error propagation but (as graceful as possible) termination and a log entry indicating the bug.

We can solve this later while reworking how we deal with errors though.

iceoryx_hoofs/source/posix_wrapper/shared_memory_object.cpp
```diff
@@ -166,17 +166,36 @@ SharedMemoryObject::SharedMemoryObject(SharedMemory&& sharedMemory,
...
+    }
+    if (m_allocationFinalized)
```
MatthiasKillat, line 177: This error condition in turn is a bit on the fence. Arguably one can request the system designer to ensure to never call it after finalization (if he controls finalization as well). On the other hand it is much harder to do when finalization is controlled elsewhere. So this is a legitimate error case that needs to be communicated somehow.

Same with being out of memory at runtime. Controllable to some degree, but not in all usage scenarios.

Resolve me.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
```diff
@@ -252,9 +253,16 @@ TEST_F(function_test, ConstructionFromAnotherFunctionIsCallable)
 TEST_F(function_test, FunctionStateIsIndependentOfSource)
 {
     ::testing::Test::RecordProperty("TEST_ID", "8302046f-cd6a-4527-aca6-3e6408f87a6b");
-    constexpr uint32_t INITIAL_STATE = 73U;
-    static_storage<1024U> storage;
-    auto* p = storage.allocate<Functor>();
+    constexpr uint32_t INITIAL_STATE = 73;
+    constexpr uint32_t MEMORY_SIZE = 1024;
+
+    // NOLINTNEXTLINE(hicpp-no-malloc, cppcoreguidelines-no-malloc) low-level memory management for testing purpose
```
MatthiasKillat, line 259: @elBoberido 

I was not aware that we enforce static analysis rules in test code as well. I think we should not (we already established that we use certain constructs in tests that are not allowed in library code, such as dynamic memory). If we do enforce the rules we need the comment of course.

MatthiasKillat (approved):
The Bumballocator is fine IMO (apart from defining zero allocation as an error).

I still do not see the benefit of moving the storage and reimplementing the core functionality in the `function` implementation itself now. Also, since some changes before to modularity/policies of `storable_function`, there is no need anymore to have different names `storable_function` and `function` (only function suffices now). The idea was that a `storable_function` with `static_storage` policy is the general purpose function wrapper we use - `function`. But there can be other ones, with different storage policy. This is obsolete now due to the many changes but the idea is still sound I wuld argue.

Since this is not a major part of `BumpAllocator` refactoring we should not change this here and now, but I recommend we reconsider the storage concept in the future.

MatthiasKillat:
@FerdinandSpitzschnueffler There are conflicts unfortunately. Will approve again once they are resolved.

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
elBoberido: @MatthiasKillat we enforce a subset of what we do for production code. In the past we also had some issues in the tests which at some point resulted in weird test failures. If the current list of checks is too strict for the tests, we can of course talk about relaxing them.

MatthiasKillat (approved):

iceoryx_hoofs/test/moduletests/test_cxx_function.cpp
MatthiasKillat: I see. Dynamic memory allocation might be one of them, since we allow `new` and `delete` for objects created in tests `malloc` and `free` should be allowed for similar reasons without a comment.

elfenpiff (approved):

PR #1856, opened 2023-01-24
iox-1855 Return in `WaitSet::wait` if data was send before `WaitSet::attachState`

### LINKED ISSUE #1855: `WaitSet::wait` does not return if data was send before `WaitSet::attachState`
## Required information

**Operating system:**
Ubuntu 20.04.3 LTS

**Compiler version:**
GCC 8.4.0

**Affected release(s):**
* `master` branch
* iceoryx Blueberry `v2.0.2`

`WaitSet::wait` does not return when there is data in the queue aka `hasData()` / `hasRespones()` / `hasRequests()` returning `true` at the time of attaching an object to the `WaitSet` and without further notification.

If data is available in the queue when attached, the `WaitSet::wait` should immediately return.

The issue was uncovered while working on https://github.com/ros2/rmw_iceoryx/pull/84 with request/response communication. The same problem should also occur with pub/sub communication, but I have not explicitly reproduced that.

1. Start app Alice with `iox::popo::Server`
2. Start app Bob with `iox::popo::Client`
3. Send a request with `iox::popo::Client` from app Bob
4. From app Alice `iox::popo::Server` sends response
5. Only now, app Bob creates a `ìox::popo::WaitSet`
6. App Bob attaches the client with `iox::popo::ClientState::HAS_RESPONSE`
7. App Bob then calls `WaitSet::wait` which never returns because the `ConditionNotifier::notify()` call and its trigger is missing for the data which is already stored in the queue

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] ~~All touched (C/C++) source code files from `iceoryx_hoofs` are added to `./clang-tidy-diff-scans.txt`~~
1. [x] Assign PR to reviewer

## Notes for Reviewer

* After attaching a state to the `WaitSet`, `isStateConditionSatisfied()` is called to see whether the state condition is still fulfilled, if so the `ConditionNotifier` is triggered
   * This allows the usage in ROS 2's rmw where `rmw_wait` attaches all wait-able entities just before waiting on every run (see https://github.com/ros2/rmw_iceoryx/pull/84)

## References

- Closes #1855

### DIFF
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -63,6 +63,7 @@
 - Avoid UB when accessing `iox::expected` [\#1750](https://github.com/eclipse-iceoryx/iceoryx/issues/1750)
 - Fix double move in `vector::emplace` [\#1823](https://github.com/eclipse-iceoryx/iceoryx/issues/1823)
 - Default roudi_config.toml path is not used [\#1826](https://github.com/eclipse-iceoryx/iceoryx/issues/1826)
+- `WaitSet::wait` returns if data was send before `WaitSet::attachState(.., State::HAS_{DATA, REQUEST, RESPONSE})` [\#1855](https://github.com/eclipse-iceoryx/iceoryx/issues/1855)
 
 **Refactoring:**
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/wait_set.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/wait_set.inl
@@ -233,6 +233,12 @@ WaitSet<Capacity>::attachState(T& stateOrigin,
                 stateOrigin,
                 TriggerHandle(*m_conditionVariableDataPtr, {*this, &WaitSet::removeTrigger}, uniqueId),
                 stateType);
+
+            auto& trigger = m_triggerArray[uniqueId];
+            if (trigger->isStateConditionSatisfied())
+            {
+                ConditionNotifier(*m_conditionVariableDataPtr, uniqueId).notify();
+            }
         });
 }
 
@@ -259,6 +265,12 @@ inline expected<WaitSetError> WaitSet<Capacity>::attachState(
         .and_then([&](auto& uniqueId) {
             NotificationAttorney::enableState(
                 stateOrigin, TriggerHandle(*m_conditionVariableDataPtr, {*this, &WaitSet::removeTrigger}, uniqueId));
+
+            auto& trigger = m_triggerArray[uniqueId];
+            if (trigger->isStateConditionSatisfied())
+            {
+                ConditionNotifier(*m_conditionVariableDataPtr, uniqueId).notify();
+            }
         });
 }
 
--- a/iceoryx_posh/test/moduletests/test_popo_waitset.cpp
+++ b/iceoryx_posh/test/moduletests/test_popo_waitset.cpp
@@ -1761,4 +1761,30 @@ TEST_F(WaitSet_test, TimedWaitUnblocksAfterMarkForDestructionCall)
     t.join();
 }
 
+TEST_F(WaitSet_test, WaitSetReturnsIfStateTriggeredBeforeAttachingWithEventType)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "407dc160-a50c-45b3-84bc-92a5f186fc14");
+    m_simpleEvents[0U].m_autoResetTrigger = false;
+    m_simpleEvents[0U].trigger();
+
+    ASSERT_FALSE(m_sut->attachState(m_simpleEvents[0U], SimpleState1::STATE1).has_error());
+
+    auto triggerVector = m_sut->timedWait(iox::units::Duration::fromSeconds(1337));
+    ASSERT_THAT(triggerVector.size(), Eq(1U));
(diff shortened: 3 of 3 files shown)


doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -63,6 +63,7 @@
 - Avoid UB when accessing `iox::expected` [\#1750](https://github.com/eclipse-iceoryx/iceoryx/issues/1750)
 - Fix double move in `vector::emplace` [\#1823](https://github.com/eclipse-iceoryx/iceoryx/issues/1823)
 - Default roudi_config.toml path is not used [\#1826](https://github.com/eclipse-iceoryx/iceoryx/issues/1826)
+- `WaitSet::wait` returns if data was send before `WaitSet::attachState` [\#1855](https://github.com/eclipse-iceoryx/iceoryx/issues/1855)
```
budrus, line 66: Only for HAS_DATA?

mossmaurice: True, I extended the call argument.

budrus:
@elfenpiff @mossmaurice If for `HAS_DATA` the waitset shall return if there are any samples in the queue, I fear this change is not the whole story. Because if it would be like that, why would we need this fix? And I always had the assumption that `HAS_DATA` checks whether there are samples in the queue without the need for an active trigger

mossmaurice:
@budrus 

> If for `HAS_DATA` the waitset shall return if there are any samples in the queue, I fear this change is not the whole story. Because if it would be like that, why would we need this fix? And I always had the assumption that `HAS_DATA` checks whether there are samples in the queue without the need for an active trigger

Good point, with the changes on this PR, `WaitSet::wait` would not return on a 2nd call and only on `WaitSet::attachState`. Currently, the `triggers` are only returned if someone rang the bell at the `m_conditionListener`:

https://github.com/eclipse-iceoryx/iceoryx/blob/5f4dd127311024cff4cd00ef2c5e119b3c717d41/iceoryx_posh/include/iceoryx_posh/internal/popo/wait_set.inl#L374-L389


I'll further investigate.

mossmaurice:
@budrus Turns out everything is fine.

In the following scenario:

1. `Publisher` is created
2. `Subscriber` is created
3. `Publisher` sends two samples to queue
4. `WaitSet` is created
5. `Publisher` is attached to `WaitSet`
6. `WaitSet::wait` immediately returns due to the fix on this PR
7. Calling `WaitSet::wait` again will, despite the missing `ConditionNotifier::notify` call return again immediately due to the fact that the `m_activeNotifications` still contains the notification from the last time, see below.

https://github.com/eclipse-iceoryx/iceoryx/blob/5f4dd127311024cff4cd00ef2c5e119b3c717d41/iceoryx_posh/include/iceoryx_posh/internal/popo/wait_set.inl#L326-L355

`WaitSet::wait`

budrus:
@mossmaurice Thanks, I got it now. The `HAS_DATA` mechanism relies on a set bit in the `activationNotifications` which needs a triggered Semaphore to get set. So it is only a kind of startup problem and the fix is valid. The benefit with this fix is that we do not change the complex algorithm that deals with `activationNotifications` and queue size polling during runtime

budrus (dismissed):

budrus (approved):

elBoberido (approved):

elfenpiff:
@elBoberido @mossmaurice @budrus I just saw the PR now and it is fine like the problem is solved BUT it demonstrates that states are an abomination directly from hell. It is the wrong layer to handle this, the waitset should only handle events and the layer above (the user using the waitset) handles states if they like. This would make things much more faster and the design more readable.

The responsibilities were:
* a waitset listens on a trigger
* events (also for states) on attachments will fire such a trigger

With this implementation we break those clear responsibilities and the waitset suddenly triggers. But I see no better alternative. Good job @mossmaurice but we have to get rid of those states they do not make sense.

budrus:
@elfenpiff In the specific `rmw_iceoryx` case this would mean the `rwm_wait()` implementation has to take care of this. Which would be possible. A ROS user doesn't know anything about this and an event-only behavior of the waitset led to the deadlock. In DDS you have all these possibilities to configure your conditions, and the entity is called waitset ;-)  

So the `state` functionality is needed here. Now the question is if the iceoryx waitset should provide this or if users or framework layers like rmw have to care of this. For me it is not really a sin that the waitset provides this functionality. Otherwise many people would have to build it on top

elBoberido:
@budrus @elfenpiff I see this as layered approach. One layer provides the basics and does not even start a thread like the `Listener` and the next layer provides what we call now `WaitSet` and `Listener`. Depending on where one is in the layer, she can choose the right component.

cc @mossmaurice

elfenpiff:
@elBoberido You are right and with the upcoming iceoryx rust next gen version I implemented it exactly like this. 
* First layer provides triggers and notifiers
* Second layer provides a reactor (a sinkhole where multiple events are coming together at one waiting point, I know the DDS/ROS guys call this waitset but the literature and the rest of the world calls it reactor)
* Third layer can add threads and more advanced functionality. Here one would add the states but in my opinion this is still a design flaw. There is a better approach of handling this which I can present in the near future but would be to much for this issue here.

But to be clear. ROS requires it (sadly) so therefore I would implemented it somewhere between here and the end user layer so that high level users do not have to take care of this.

Issue #1391, opened 2022-06-10
Split `iceoryx_hoofs` into logical modules


mossmaurice:
## Brief feature description

Currently, `iceoryx_hoofs` uses a monolithic `CMakeLists.txt`. To model and control the dependencies of the `iceoryx_hoofs` modules / namespace like `cxx`, `concurrent`, etc. the `CMakeLists.txt` shall be split up into several ones. The overall goal is a cleaner architecture.
An interface library shall be used to wrap the different libraries to the target `iceoryx_hoofs`, so that the user does not have to adapt any CMake code.

## Detailed information

Split up the `iceoryx_hoofs` [modules / namspaces](https://iceoryx.io/v2.0.2/images/iceoryx_components_diagram_v2_0_0.svg) into several CMake targets.

There are two steps
- [x] move the source files to separate folders with an own include/source structure
- [x] ~create object libs for this folders the prevent cyclic dependencies between the sub-packages~ split out to #2204 

## References

* Comes from #590

mossmaurice:
Modules were created in:

- #2143 
- #2142
- #2061
- #2133 

This should be discussed in the next developer meetup if this is still wanted for `v4.0`.

PR #1600, opened 2022-09-01
iox-#1295 Update component and layer diagram for upcoming release

mossmaurice:
## Pre-Review Checklist for the PR Author

1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] ~~Changelog updated [in the unreleased section][changelog] including API breaking changes~~ No user API changed
1. [x] Branch follows the naming format (`iox-123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit messages are signed (`git commit -s`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] All touched (C/C++) source code files from `iceoryx_hoofs` are added to `./clang-tidy-diff-scans.txt`
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/concepts/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/release-notes/iceoryx-unreleased.md

## Notes for Reviewer

* Add component and layer diagram for upcoming `v3.x` release

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] All touched (C/C++) source code files from `iceoryx_hoofs` have been added to `./clang-tidy-diff-scans.txt`
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

- Closes #1295 (partly)

PR #1863, opened 2023-01-31
iox-1391 Move time-related classes to `Time` module

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Move `deadline_timer`, `adaptive_wait` and `Duration` to `Time` module
* Remove 2nd level namespace `cxx`
* Create legacy header for backwards compatibility

## References

- Closes #1391 (Partly)


doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -907,3 +907,15 @@
     // iox::string to std::string
     std::string myConvertedIoxString = iox::cxx::into<std::string>(myIoxString);
     ```
+
+44. Move and rename `DeadlineTimer`
```
mossmaurice, line 911: Both `adaptive_wait` and `units::Duration` are labeled as `internal` and hence not mentioned in the release note.

elfenpiff (changes requested):
No please do not name this `Time` it is a `Units` module. Otherwise we will create in the near future a lot of additional tiny modules with one or two classes of content.

It may makes sense that in this `Units` module is later also a unit for `velocity`, `acceleration` etc. for our users or we have an abstraction for of the unit `memory`

elfenpiff:
> No please do not name this `Time` it is a `Units` module. Otherwise we will create in the near future a lot of additional tiny modules with one or two classes of content.
> 
> It may makes sense that in this `Units` module is later also a unit for `velocity`, `acceleration` etc. for our users or we have an abstraction for of the unit `memory`

@mossmaurice The high level classes like `adaptive_wait` etc. would not be stored in a units module but here I would try to find something more generic than `Time`.

mossmaurice:
@elfenpiff  Thanks for your review and remarks.

> No please do not name this `Time` it is a `Units` module. Otherwise we will create in the near future a lot of additional tiny modules with one or two classes of content.
> 
> It may makes sense that in this `Units` module is later also a unit for `velocity`, `acceleration` etc. for our users or we have an abstraction for of the unit `memory`

I remember we had this discussion in #1600 during October 2022. I can definitely see your point that `Units` is something different and might be extended later. Back in Oct we concluded that, as you mentioned, in order to keep the number of overall modules as low as possible, we will not further differentiate and just use one module for all time- and units-related classes. IMHO that's still a good idea and we should not change this decision. I would be ok with renaming the current module to something like `Units and time utilities`. The STL has a similar module [`Date and time utilites`](https://en.cppreference.com/w/cpp/chrono), which contains the clocks, time points, duration, etc.

@elBoberido What's your take? Shall we change the plan?

elfenpiff:
> @elBoberido What's your take? Shall we change the [plan](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_hoofs/README.md?plain=1#L156-L172)?

@mossmaurice Let it be `Time` then. In the readme you linked there we also concluded that every posix construct is stored inside the `posix` folder and that we there replicate the component structure.
`Duration` is a posix abstraction (see `timeval`, `timespec` and `clock_gettime`) and belongs therefore in `posix/time` but I would also agree to simply `posix`

The deadline timer uses the duration and has no posix dependency and can therefore be stored simply in `time`.

The adaptive wait is a little bit more complicated. It uses `std::this_thread::yield()` and `std::this_thread::sleep_for` calls which are not allowed for us in the final implementation and they have to be replaced with the posix calls `sched_yield` and `nanosleep`.
So I would recommend to move `adaptive_wait` into `posix/time` since this is the correct final resting place for it. If you have time you can use the posix call and replace `std::this_thread::yield()` and `std::this_thread::sleep_for` for the previously mentioned calls - then it is cleaned up correctly. For me it makes sense since those are only 3 additional line changes.

elBoberido:
@mossmaurice while `chrono` can be moved to `dust`, `timespec` and `timeval` must remain in `hoofs` since it is our native way to convert a `Duration` to a time value for POSIX calls like `sem_timedwait`. These could potentially be moved to 'posix/time/whatever.hpp'

doc/website/release-notes/iceoryx-unreleased.md
elBoberido: @mossmaurice since `unints::Duration` might be used quite heavily despite being labeled `internal` I think it would nevertheless be a good idea to also mention that class.

mossmaurice: @elBoberido Good point, added.

mossmaurice:
@elBoberido 

> while `chrono` can be moved to `dust`, `timespec` and `timeval` must remain in hoofs since it is our native way to convert a Duration to a time value for POSIX calls like `sem_timedwait`. These could potentially be moved to 'posix/time/whatever.hpp'

That's true, I'll address that on the follow-up PR.

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -998,3 +998,25 @@
...
+    // before
+    #include "iceoryx_hoofs/cxx/deadline_timer.hpp"
+    iox::cxx::DeadlineTimer myTimer;
+
+    // after
+    #include "iox/deadline_timer.hpp"
+    iox::deadline_timer myTimer;
+    ```
+
+45. Changed include path of `iox::units::Duration`
```
elBoberido, line 1014: Out of curiosity, is it now `iox::Duration` or still `iox::units::Duration`?

elBoberido (approved):

doc/website/release-notes/iceoryx-unreleased.md
mossmaurice: The latter, as we hopefully will extend `iox::units` with more SI units.

mossmaurice: It's similar in the STL with `std::chrono`. For me it's ok to have nested namespace, but I agreed currently it does not make a lot of sense.

elfenpiff (approved):

PR #1894, opened 2023-02-14
iox-1893 fix memory leak in stack

### LINKED ISSUE #1893: `iox::stack` copies `T` on `pop()` instead of moving
## Required information

**Operating system:**
Ubuntu 20.04.3 LTS

**Compiler version:**
GCC 8.4.0

**Eclipse iceoryx version:**
`master` branch

* Memory leak in the `iox::stack`
  * `stack::pop` returns a copy of the value instead of the moving value
  * `stack::pop` does not call `T::~T` on the moved-from value

https://github.com/eclipse-iceoryx/iceoryx/blob/f4c81abf003d2ca47ffdd6323f3fbc3824c1aa98/iceoryx_hoofs/buffer/include/iox/detail/stack.inl#L150

* No memory leak
  * Moving the value out on `stack::pop` call
  * Calling `T::~T` on the moved-from value after `stack::pop`

```cpp
    {
        stack<TestClass, STACK_SIZE> other;
        other.push(TestClass(37, 37, 37));

        stack<TestClass, STACK_SIZE> sut(other);

        EXPECT_EQ(other.size(), 1);
        EXPECT_EQ(sut.size(), 1);

        EXPECT_THAT(TestClass::dTor, Eq(1));
        EXPECT_THAT(TestClass::copyCTor, Eq(1));
        EXPECT_THAT(TestClass::moveCTor, Eq(1));

        auto element = sut.pop(); // A copy is performed here and the TestClass::~TestClass is not called
        ASSERT_TRUE(element.has_value());
        EXPECT_THAT(other.size(), Eq(1));
        EXPECT_THAT(sut.size(), Eq(0));
    }
    // The below expectations are violated
    EXPECT_THAT(TestClass::dTor, Eq(5));
    EXPECT_THAT(TestClass::cTor, Eq(5));
    EXPECT_THAT(TestClass::copyCTor, Eq(1));
    EXPECT_THAT(TestClass::moveCTor, Eq(3));
```

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Fix potential memory leak in `iox::stack` by
  * Moving the value out on `pop()`
  * Calling `T::~T` after `pop()` on the moved-from value inside the `iox::stack` 

## References

- Closes #1893

### DIFF
--- a/iceoryx_hoofs/buffer/include/iox/detail/stack.inl
+++ b/iceoryx_hoofs/buffer/include/iox/detail/stack.inl
@@ -144,10 +144,9 @@ inline optional<T> stack<T, Capacity>::pop() noexcept
         return nullopt;
     }
 
-    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : low level memory management with access to the topmost element on
-    // the untyped buffer; reinterpret_cast is safe since the size and the alignment of each array element is guaranteed
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-    return *reinterpret_cast<T*>(&m_data[--m_size]);
+    optional<T> element{std::move(m_data[--m_size])};
+    m_data[m_size].~T();
+    return element;
 }
 
 template <typename T, uint64_t Capacity>
--- a/iceoryx_hoofs/buffer/include/iox/detail/stack.inl
+++ b/iceoryx_hoofs/buffer/include/iox/detail/stack.inl
@@ -130,10 +130,7 @@ inline T& stack<T, Capacity>::getUnchecked(const uint64_t index) noexcept
 template <typename T, uint64_t Capacity>
 inline const T& stack<T, Capacity>::getUnchecked(const uint64_t index) const noexcept
 {
-    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : reinterpret_cast is safe since the size and the alignment of each
-    // array element is guaranteed
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-    return *reinterpret_cast<const T*>(&m_data[index]);
+    return m_data[index];
 }
 
 template <typename T, uint64_t Capacity>
--- a/iceoryx_hoofs/buffer/include/iox/stack.hpp
+++ b/iceoryx_hoofs/buffer/include/iox/stack.hpp
@@ -18,12 +18,18 @@
 
 #include "iox/algorithm.hpp"
 #include "iox/optional.hpp"
+#include "iox/type_traits.hpp"
 #include "iox/uninitialized_array.hpp"
 
 #include <cstdint>
 
 namespace iox
 {
+
+// Due to 'pop' returning an 'iox::optional'
+template <typename T>
+using is_storable_in_iox_stack_t = typename std::is_move_constructible<T>::type;
+
 // AXIVION Next Construct AutosarC++19_03-A12.1.1 : it is guaranteed that the array elements are initialized before read
 // access
 // AXIVION Next Construct AutosarC++19_03-A9.6.1 : false positive since no bit-fields are involved
@@ -34,6 +40,8 @@ template <typename T, uint64_t Capacity>
 class stack final // NOLINT(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 {
   public:
+    static_assert(is_storable_in_iox_stack_t<T>::value, "T has to be move constructible");
+
     stack() noexcept = default;
     stack(const stack& rhs) noexcept;
     stack(stack&& rhs) noexcept;
--- a/iceoryx_hoofs/test/moduletests/test_buffer_stack.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_buffer_stack.cpp
@@ -211,9 +211,17 @@ TEST_F(stack_test, popCreatesSpaceForAnotherElement)
     ::testing::Test::RecordProperty("TEST_ID", "3ebf7f6d-81ef-45d6-83a6-80f8588cbba6");
     pushElements(STACK_SIZE);
 
-    EXPECT_THAT(m_sut.pop(), Ne(nullopt));
+    EXPECT_TRUE(m_sut.pop().has_value());
     EXPECT_TRUE(m_sut.push());
-    EXPECT_THAT(TestClass::dTor, Eq(1));
+
+// clang >= 15 performs mandatory elision of copy/move operations (C++17 requirement)
+#if defined(__clang__) && __clang_major__ >= 15
+    constexpr uint64_t ELISION_CORRECTION{0};
+#else
+    constexpr uint64_t ELISION_CORRECTION{1};
+#endif
+
+    EXPECT_THAT(TestClass::dTor, Eq(2 + ELISION_CORRECTION));
 }
 
 TEST_F(stack_test, TestClassDTorIsCalledWhenStackGoesOutOfScope)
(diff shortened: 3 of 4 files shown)


iceoryx_hoofs/buffer/include/iox/detail/stack.inl
```diff
@@ -144,10 +144,9 @@ inline optional<T> stack<T, Capacity>::pop() noexcept
         return nullopt;
     }
 
-    // AXIVION Next Construct AutosarC++19_03-A5.2.4 : low level memory management with access to the topmost element on
```
mossmaurice, line 147: The Axivion suppression is not needed here anymore, as the `reinterpret_cast` is being performed inside the `UninitializedArray`.

iceoryx_hoofs/test/moduletests/test_buffer_stack.cpp
```diff
@@ -254,6 +258,34 @@ TEST_F(stack_test, CopyConstructorWorksAndCallsTestClassCopyConstructor)
...
+        EXPECT_THAT(TestClass::copyCTor, Eq(1));
+        EXPECT_THAT(TestClass::moveCTor, Eq(1));
+
+        auto element = sut.pop();
+        ASSERT_TRUE(element.has_value());
+        EXPECT_THAT(other.size(), Eq(1));
+        EXPECT_THAT(sut.size(), Eq(0));
+    }
+    EXPECT_THAT(TestClass::dTor, Eq(5));
+    EXPECT_THAT(TestClass::cTor, Eq(5));
```
mossmaurice, line 284: @elBoberido I have troubles to understand why the d'tor and c'tor call are one less each in case of the [ASAN job](https://github.com/eclipse-iceoryx/iceoryx/actions/runs/4175619432/jobs/7230810479). Wasn't able to reproduce this locally.

It also happens in `popCreatesSpaceForAnotherElement`.

mossmaurice: It's related to `clang-15` and [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision).

iceoryx_hoofs/buffer/include/iox/detail/stack.inl
```diff
@@ -144,10 +141,9 @@ inline optional<T> stack<T, Capacity>::pop() noexcept
...
-    // the untyped buffer; reinterpret_cast is safe since the size and the alignment of each array element is guaranteed
-    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-    return *reinterpret_cast<T*>(&m_data[--m_size]);
+    optional<T> element{std::move(m_data[--m_size])};
```
elfenpiff, line 144: Mmmhh does it mean we loosen the type requirements here a bit so that types only have to be move constructible?

elfenpiff (changes requested):
Could you please add a 
```cxx
template<typename T>
constexpr bool FULFILLS_STACK_TYPE_REQUIREMENTS = std::is_move_constructible<T>;
```

And verify this inside the stack class with
```cxx
static_assert(FULFILLS_STACK_TYPE_REQUIREMENTS<T>, "T has to be move constructible");
```

This would help the user a lot otherwise they get a complex and weird compile time error when they try to move elements out of the stack and post bug reports. With this we make clear that those types are not supported since you cannot pop them out of the stack

iceoryx_hoofs/buffer/include/iox/detail/stack.inl
mossmaurice: @elfenpiff Yep, that's correct. I added the `static_assert`.

elBoberido (dismissed):

iceoryx_hoofs/buffer/include/iox/stack.hpp
```diff
@@ -34,6 +35,9 @@ template <typename T, uint64_t Capacity>
 class stack final // NOLINT(cppcoreguidelines-pro-type-member-init, hicpp-member-init)
 {
   public:
+    // Due to 'pop' returning an 'iox::optional'
+    static_assert(std::is_move_constructible<T>::value, "T has to be move constructible");
```
elfenpiff, line 39: Could you please move the condition to an external constexpr variable so that we can check in different instances if the type is compatible with the stack beforehand. If it is directly in here we have no other way than fail instead of providing a nice message.

So:
```cxx
template<typename T>
constexpr bool TYPE_FULFILLS_STACK_REQUIREMENTS = std::is_move_constructible<T>::value;

// this line in the stack

static_assert(TYPE_FULFILLS_STACK_REQUIREMENTS<T>, "..."); 
```

Now we can ask in other templates with `TYPE_FULFILLLS_STACK_REQUIREMENTS` if the type works with a stack.

elfenpiff (commented):
Looks good. When the constexpr variable is introduced we can merge this thing.

Could you maybe squash some commits? From the first look it seems a little bit too fine grained.

iceoryx_hoofs/buffer/include/iox/stack.hpp
mossmaurice: @elfenpiff Got your point. What I don't like that too much about that we'll end up with inconsistent naming of helper templates, I'm really missing the formality of C++20 concepts here :pensive: Right now, most of the time the `static_assert`s are in the class e.g. `CyclicIndex`, `SoFi`, `PolymorphicHandler`, etc.

I would stick with the idiomatic C++ way of naming the helper template if you don't mind:

```cpp
template <typename T>
using is_storable_in_iox_stack_t = typename std::is_move_constructible<T>::type;

// the following line in the stack

static_assert(is_storable_in_iox_stack_t<T>::value, "T has to be move constructible");
```

Does that make sense?

elfenpiff: @mossmaurice Good idea. Let's do it in your proposed way.

iceoryx_hoofs/buffer/include/iox/stack.hpp
```diff
@@ -18,12 +18,18 @@
...
 #include "iox/uninitialized_array.hpp"
 
 #include <cstdint>
 
 namespace iox
 {
+
+// Due to 'pop' returning an 'iox::optional'
+template <typename T>
+using is_storable_in_iox_stack_t = typename std::is_move_constructible<T>::type;
```
elfenpiff, line 31: Why not write:
```cxx
template<typename T>
constexpr bool is_storable_in_iox_stack_v = std::is_move_constructible<T>::value;
```

elfenpiff (commented):
Why did you create a type alias instead of a templated constexpr

mossmaurice:
@elfenpiff 

> Why did you create a type alias instead of a templated constexpr

Because that's what I wrote here in [my post this morning](https://github.com/eclipse-iceoryx/iceoryx/pull/1894#discussion_r1111785058) :grin: 

I might have confused you, though. My aim is to prevent the instantiation of an variable for the check, but instead provide C++-y way of checking if the restrictions on `T` are fulfilled without the need to instantiate the `iox::stack` and then get stuck in the `static_assert`. E.g. a refactoring could lead to more restrictions or less restrictions and then someone using the `iox::stack` could nicely use SFINAE with `is_storable_in_iox_stack_t` to draw a conclusion during compile-time.

elfenpiff:
> @elfenpiff
> 
> > Why did you create a type alias instead of a templated constexpr
> 
> Because that's what I wrote here in [my post this morning](https://github.com/eclipse-iceoryx/iceoryx/pull/1894#discussion_r1111785058) grin
> 
> I might have confused you, though. My aim is to prevent the instantiation of an variable for the check, but instead provide C++-y way of checking if the restrictions on `T` are fulfilled without the need to instantiate the `iox::stack` and then get stuck in the `static_assert`. E.g. a refactoring could lead to more restrictions or less restrictions and then someone using the `iox::stack` could nicely use SFINAE with `is_storable_in_iox_stack_t` to draw a conclusion during compile-time.

But providing a constexpr variable is the C++-y way of checking such things! Isn't this why C++17 has introduced variables like `std::is_copy_constructible_v`? It is just a little issue but why force everyone to type more without gaining any benifit.

>  but instead provide C++-y way of checking if the restrictions on `T` are fulfilled without the need to instantiate the `iox::stack` 

```cxx
template<typename T>
constexpr bool fulfills_stack_type_requirements = std::is_copy_constructible<T>::value
```

fulfills exactly this requirement. Please try it out and you will see.

elfenpiff (approved):

elBoberido (approved):

PR #1898, opened 2023-02-15
Introspection-app Wrap c calls in posixCall

### LINKED ISSUE #1692: Wrap all C calls in `posixCall` in IntrospectionApp
## Brief feature description

The introspection app uses C calls without using the `posixCall`. This should be adjusted to ensure a correct error handling.

zmostafa:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #1692

### DIFF
--- a/tools/introspection/include/iceoryx_introspection/introspection_app.hpp
+++ b/tools/introspection/include/iceoryx_introspection/introspection_app.hpp
@@ -62,7 +62,6 @@ static const std::map<PrettyOptions, uint32_t> prettyMap = {
 
 
 /// @brief base class for introspection
-/// @todo iox-#1692 Wrap all C calls with posixCall
 class IntrospectionApp
 {
   public:
--- a/tools/introspection/source/introspection_app.cpp
+++ b/tools/introspection/source/introspection_app.cpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2019 - 2020 by Robert Bosch GmbH. All rights reserved.
-// Copyright (c) 2020 - 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2020 - 2023 by Apex.AI Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -228,15 +228,16 @@ void IntrospectionApp::waitForUserInput(int32_t timeoutMs)
     fileDesc.fd = STDIN_FILENO;
     fileDesc.events = POLLIN;
     constexpr size_t nFileDesc = 1u;
-    /// @todo iox-#1692 Wrap kernel calls with posixCall
-    int32_t eventCount = poll(&fileDesc, nFileDesc, timeoutMs);
-
-    // Event detected
-    if ((eventCount == nFileDesc) && (fileDesc.revents == POLLIN))
-    {
-        updateDisplayYX();
-        refreshTerminal();
-    }
+    iox::posix::posixCall(poll)(&fileDesc, nFileDesc, timeoutMs)
+        .failureReturnValue(-1)
+        .evaluate()
+        .and_then([&](auto eventCount) {
+            if (static_cast<size_t>(eventCount.value) == nFileDesc && fileDesc.revents == POLLIN)
+            {
+                this->updateDisplayYX();
+                this->refreshTerminal();
+            }
+        });
 }
 
 void IntrospectionApp::prettyPrint(const std::string& str, const PrettyOptions pr)
@@ -791,7 +792,8 @@ void IntrospectionApp::runIntrospection(const iox::units::Duration updatePeriod,
         }
     }
 
-    getchar();
+    iox::posix::posixCall(getchar)().failureReturnValue(EOF).evaluate().expect(
+        "unable to exit the introspection client since getchar failed");
     closeTerminal();
 }
 
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -119,6 +119,7 @@
 - Remove `algorithm::uniqueMergeSortedContainers` from `algorithm.hpp`
 - Move `std::string` conversion function to `iceoryx_dust` [\#1612](https://github.com/eclipse-iceoryx/iceoryx/issues/1612)
 - The posix call `unlink` is directly used in `UnixDomainSocket` [\#1622](https://github.com/eclipse-iceoryx/iceoryx/issues/1622)
+- Wrap all C calls in posixCall in IntrospectionApp [\#1692](https://github.com/eclipse-iceoryx/iceoryx/issues/1692)
 
 **Workflow:**
 


zmostafa:
@elfenpiff I could be missing something, but the only api I found in the introspection app was `poll` and that is what I did in the MR, is there any other c call or kernel calls that I might have missed ?

tools/introspection/include/iceoryx_introspection/introspection_app.hpp
```diff
@@ -62,7 +62,6 @@ static const std::map<PrettyOptions, uint32_t> prettyMap = {
 
 
 /// @brief base class for introspection
-/// @todo iox-#1692 Wrap all C calls with posixCall
```
mossmaurice, line 65: I think this comment here meant to wrap all C operating system calls e.g. those in `initTerminal`. There might be others.

```
    initscr();
    curs_set(0);
    start_color();
    init_pair(static_cast<uint8_t>(ColorPairs::redOnBlack), COLOR_RED, COLOR_BLACK);
    init_pair(static_cast<uint8_t>(ColorPairs::whiteOnRed), COLOR_WHITE, COLOR_RED);
```

@elfenpiff Can you confirm?

zmostafa: @elfenpiff Can you confirm if my understanding to the use of `posixCall` is right ? or I need to wrap all calls from c written libraries like `ncurses` ?

cc. @mossmaurice

elfenpiff: @zmostafa You are completely right with your understanding. The posix call function is for posix c function calls only. It requires that the either the return value is the errno or the return value indicates an errno and the errno is set.

C functions which can never fail (even when they are posix function) do not need to be wrapped.

The intention of the posixCall was, that the error handling and all the side conditions are hard to handle and it is easy to miss something. The posixCall does some rudimentary error handling already for you, resets the errno = 0 before calling the posix function and repeats the call if it has failed and the errno is set to EINTR. Something which you definitely do not want from an ncurses function.

tools/introspection/source/introspection_app.cpp
```diff
@@ -228,14 +228,17 @@ void IntrospectionApp::waitForUserInput(int32_t timeoutMs)
     fileDesc.fd = STDIN_FILENO;
     fileDesc.events = POLLIN;
     constexpr size_t nFileDesc = 1u;
-    /// @todo iox-#1692 Wrap kernel calls with posixCall
-    int32_t eventCount = poll(&fileDesc, nFileDesc, timeoutMs);
+
+    auto eventCount = iox::posix::posixCall(poll)(&fileDesc, nFileDesc, timeoutMs).failureReturnValue(-1).evaluate();
```
elfenpiff, line 232: You could also use here our awesome `and_then` API :smile: and rewrite the code like this:

```cxx
iox::posix::posixCall(poll)(&fileDesc, nFileDesc, timeoutMs).failureReturnValue(-1).evaluate()
  .and_then([&](auto eventCount){
    if ( eventCount == nFileDesc && fileDesc.revents == POLLIN ) {
      updateDisplayYX();
      refreshTerminal();
    }
  });
```

tools/introspection/include/iceoryx_introspection/introspection_app.hpp
zmostafa: @elfenpiff Thanks for the detailed explanation.

So, my understanding is like:
 1. `posixCall` can be used with non-posix calls, to use its error handling features.
 2. I should look on `ncurses` methods and their returned error number to wrap them in a `posixCall`

Does this make sense now ?

elfenpiff: > The posixCall does some rudimentary error handling already for you, resets the errno = 0 before calling the posix function and **repeats the call if it has failed and the errno is set to EINTR. Something which you definitely do not want from an ncurses function.**

@zmostafa From my previous comment. This is why I would not use the `posixCall` to wrap non-posix calls like for instance ncurses functions.

zmostafa: @elfenpiff my bad, I thought you do not want a failure from ncureses, and re-trying make sense here :sweat_smile: Thanks again for your explanation.

I wrapped `getchar()` already and I think nothing needs to be done more here,unless I missed a concrete posix call here :eyes:

elfenpiff: @zmostafa No problem at all! I think you finished them all and now lets get it merged.

tools/introspection/source/introspection_app.cpp
```diff
@@ -791,7 +792,10 @@ void IntrospectionApp::runIntrospection(const iox::units::Duration updatePeriod,
         }
     }
 
-    getchar();
+    iox::posix::posixCall(getchar)().failureReturnValue(EOF).evaluate().or_else([](auto& r) {
```
zmostafa, line 795: @elfenpiff I had an error for dropping the return value, so I used it to report the error using `cout` I know you mentioned before that this might not be needed as `posixCall` take care of formatting a nice error message to the use, but I was using `iox::log::Logger`.

Is there a better way to do it ? handle the return value I mean?

elfenpiff: @zmostafa Yes there is! The method `expect` is exactly what you want here. If the call fails it prints the error message and terminates. So in this case:
```cxx
iox::posix::posixCall(getchar)().failureReturnValue(EOF).evaluate().expect("This error message should provide some context where the error is occurring");
```

Please try not to write error messages like this:
```cxx
std::cout << "getchar() call failed with EOF error" << r.getHumanReadableErrnum() << std::endl;
```

`posixCall` will print exactly this but in a more detailed fashion. It will provide file, line, the posix function which one has called, the return value, the errno value as integer and the errno as human readable string. So in the end when this would fail you would get the same error message more or less twice.

In every level you should provide some context where the error is occurring and state what is affected. I think in this case `getchar` is used to wait for user input to refresh something?! So it would be nice to state this. This would result in a nice and readable log output which reads somehow like a stack trace for humans.

In this case for instance:
```
introspection_app.cpp:795 (getchar) failed with errno XX, human readable errno message
unable to refresh whatever in introspection since getchar failed
```

The first line would come from the posix call the second would originate here. Now a fellow developer knows that getchar failed and that this lead to a failure in the introspection refresh thingy.

zmostafa: @elfenpiff Thanks, I am using this [API page](https://iceoryx.io/v2.0.3/API-reference/hoofs/Files/posix__call_8hpp/#define-posixcall) to find about `posixCall` and search in the code to find usage examples,and could not find anything about using `expect`.

I think we need to update the page with what you explained me already, this is more than what is written there. :sweat_smile:

elfenpiff: @zmostafa You are right. In the end the `posixCall` returns an `expected` and this provides the method `expect`

elfenpiff (dismissed):

tools/introspection/source/introspection_app.cpp
```diff
@@ -228,15 +228,16 @@ void IntrospectionApp::waitForUserInput(int32_t timeoutMs)
...
-    /// @todo iox-#1692 Wrap kernel calls with posixCall
-    int32_t eventCount = poll(&fileDesc, nFileDesc, timeoutMs);
-
-    // Event detected
-    if ((eventCount == nFileDesc) && (fileDesc.revents == POLLIN))
-    {
-        updateDisplayYX();
-        refreshTerminal();
-    }
+    iox::posix::posixCall(poll)(&fileDesc, nFileDesc, timeoutMs)
```
mossmaurice, line 231: @zmostafa Could you please update the year in the copyright header at the very top of this file?

zmostafa: @mossmaurice Done. 
Is that the rule now ? once I touch a file I have to update the year in the header ?

mossmaurice: @zmostafa Yes, correct.

mossmaurice (dismissed):
LGTM! @zmostafa There is one checkbox you still have to tick ;)

elfenpiff (approved):

mossmaurice (dismissed):

zmostafa:
@mossmaurice @elfenpiff just rebased to latest master to get the pipeline fixed. Can you please check and approve again ?!

mossmaurice (approved):

PR #1929, opened 2023-02-28
iox-536 Move `std::chrono` dependency to `iceoryx_dust`

### LINKED ISSUE #536: Improving `units::Duration`
## Brief feature description

The `units::Duration` class could have same additional features and improvements.

## Detailed information

- [x] rename the creation methods like `milliseconds` to `fromMilliseconds`
    - some getter methods should then be renamed, e.g. `milliSeconds` -> `milliseconds` or `toMilliseconds`
    - no final decision, please ask before starting with this
- [ ] conversion to/from all `chrono` types
- [x] `operator+=`, `operator-=` and `operator*=` https://github.com/eclipse-iceoryx/iceoryx/pull/501#discussion_r564463012
- [ ] abstractions for `TimePoint` for monotonic and system clock https://github.com/eclipse-iceoryx/iceoryx/pull/501#discussion_r560061177
    - needs to be discussed and maybe another issue should be used for this

mossmaurice:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

* Move `std::chrono` dependency to `iceoryx_dust`

## References

- Closes #536 (partly)


iceoryx_hoofs/posix/time/source/deadline_timer.cpp
```diff
@@ -48,11 +51,21 @@ iox::units::Duration deadline_timer::remainingTime() const noexcept
...
-    return iox::units::Duration(std::chrono::milliseconds(0));
+    using namespace iox::units::duration_literals;
+    return 0_s;
 }
 
 iox::units::Duration deadline_timer::getCurrentMonotonicTime() noexcept
 {
-    return iox::units::Duration{std::chrono::steady_clock::now().time_since_epoch()};
+    timespec time_since_epoch{0, 0};
+    if (posix::posixCall(clock_gettime)(CLOCK_REALTIME, &time_since_epoch)
```
FerdinandSpitzschnueffler, line 61: Not sure but doesn't `std::chrono::steady_clock` correspond more to `CLOCK_MONOTONIC`?

iceoryx_dust/include/iceoryx_dust/cxx/std_chrono_support.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+/// @brief Construct a Duration object from std::chrono::milliseconds
+/// @param[in] value as milliseconds
+/// @attention Since negative durations are not allowed, the duration will be clamped to 0
+Duration make_duration(const std::chrono::milliseconds value) noexcept;
+
+/// @brief Construct a Duration object from std::chrono::nanoseconds
+/// @param[in] value as nanoseconds
+/// @attention Since negative durations are not allowed, the duration will be clamped to 0
+Duration make_duration(const std::chrono::nanoseconds value) noexcept;
```
elBoberido, line 35: Why not just `into<Duration>`? It is similar to the `std::string -> iox::string` approach and when we use it for other type conversions a user only needs to learn the pattern once. The same can be done for other STL types if necessary.

In addition we could also have a `into<std::chrono::milliseconds>` later on.

iceoryx_hoofs/posix/time/source/deadline_timer.cpp
elBoberido: Yes, it must be `CLOCK_MONOTONIC`

iceoryx_hoofs/posix/time/source/deadline_timer.cpp
```diff
@@ -48,11 +51,21 @@ iox::units::Duration deadline_timer::remainingTime() const noexcept
...
+            .failureReturnValue(-1)
+            .evaluate()
+            .has_error())
+    {
+        using namespace iox::units::duration_literals;
+        return 0_s;
```
elBoberido, line 67: I'm hesitant to silently fail. Can you check under which conditions `clock_gettime` can fail?

My tendency is to call the error handler since we do not now what the user does after the call to this method. It might lead to a failure on the user side. Mid term we should add a `tryGetCurrentMonotonicTime` and encourage that use instead of something that can terminate.

iceoryx_dust/include/iceoryx_dust/cxx/std_chrono_support.hpp
mossmaurice: Good point, I'll refactor that :sweat_smile:

iceoryx_hoofs/posix/time/source/deadline_timer.cpp
mossmaurice: I would argue that this doesn't really matter because [`time_since_epoch`](https://en.cppreference.com/w/cpp/chrono/time_point/time_since_epoch) has the same behaviour. The caller needs to deal with the result of `0`.

elBoberido: I just had a look at `clock_gettime` and it seems there are three possible errors
- EFAULT: tp points outside the accessible address space. 
- EPERM: clock_settime() does not have permission to set the clock indicated. 
- EINVAL: The clk_id specified is not supported on this system. 

The first one cannot happen since `tp` is on the stack. The second one also cannot happen since we are not using `clock_settime`. The last one might happen when someone ports iceoryx to a esoteric platform or someone creates a faulty image for a supported platform. In this case the class is completely broken and will never work. Therefore it makes sense to call the error handler here and fail early. In another PR an `IOX_UNREACHABLE` macro is in the works and that one would be ideal for this case since it will be unreachable in a properly configured system. In the meantime, could you add an `Ensures` or at least a todo to add the `IOX_UNREACHABLE`?

doc/website/release-notes/iceoryx-unreleased.md
```diff
@@ -1045,3 +1046,16 @@
...
+    ```cpp
+    //before
+    std::chrono::milliseconds chronoDuration = 1_ms;
+    iox::units::Duration ioxDuration(chronoDuration);
+
+    //after
+    #include "iceoryx_dust/cxx/std_chrono_support.hpp"
+
+    std::chrono::milliseconds chronoDuration = 1_ms;
+    iox::units::Duration ioxDuration = into<iox::units::Duration>(chronoDuration);
```
elBoberido, line 1060: With this code one would introduce Axivian warnings ;)

If you need another commit please consider changing this either to `auto ioxDuration = ...` or ``iox::units::Duration ioxDuration {...};

iceoryx_dust/include/iceoryx_dust/internal/cxx/std_chrono_support.inl
```diff
@@ -0,0 +1,38 @@
...
+#include "iceoryx_dust/cxx/std_chrono_support.hpp"
+
+namespace iox
+{
+
+inline units::Duration
+FromImpl<std::chrono::milliseconds, units::Duration>::fromImpl(const std::chrono::milliseconds& value) noexcept
+{
+    return units::Duration::fromMilliseconds(value.count());
+}
```
elBoberido, line 29: At some point we really have to reconsider our guidelines regarding the `*.inl` files. There should be exceptions for cases like this whit just some one-liner

iceoryx_dust/include/iceoryx_dust/internal/cxx/std_string_support.inl
```diff
@@ -30,7 +30,7 @@ template <uint64_t N>
 inline string<N> FromImpl<std::string, string<N>>::fromImpl(const std::string&) noexcept
 {
     static_assert(always_false_v<std::string> && always_false_v<string<N>>, "\n \
-        The conversion from 'std::string' to 'iox::sring<N>' is potentially lossy!\n \
+        The conversion from 'std::string' to 'iox::string<N>' is potentially lossy!\n \
```
elBoberido, line 33: Just googled it and sring is an actual word :laughing:

iceoryx_hoofs/posix/time/source/deadline_timer.cpp
mossmaurice: I think that's very esoteric and nothing can go really wrong here, because a properly constructed `Duration` is returned.

If this is changed I would propose to change it consistently at all locations:

https://github.com/eclipse-iceoryx/iceoryx/blob/619514f1e98088f978935e5ec1601649024d2fdc/iceoryx_hoofs/time/source/duration.cpp#L55-L58

https://github.com/eclipse-iceoryx/iceoryx/blob/619514f1e98088f978935e5ec1601649024d2fdc/iceoryx_hoofs/source/log/building_blocks/console_logger.cpp#L58-L62

Is it ok for you if we take this to a follow-up PR?

iceoryx_posh/test/moduletests/test_runtime_ipc_interface.cpp
```diff
@@ -303,7 +304,7 @@ TYPED_TEST(IpcInterface_test, TimedSendWorks)
         if (!result)
         {
             // Do not exceed timeout
-            auto timeDiff = units::Duration(after - before);
+            auto timeDiff = units::make_duration(after - before);
```
elBoberido, line 307: I guess this is a leftover from the previous approach?

elBoberido (dismissed):

iceoryx_hoofs/posix/time/source/deadline_timer.cpp
elBoberido: Being esoteric is exactly the reason I'd like it to have an early failure. I'd rather have a bug report with a `I get this error code` than `I get a zero Duration`. The former tells me exactly what happens. The latter would be either `clock_gettime` returning 0 or an error.

I don't mind if you change it for the logger but there the situation is a bit different. Having a timestamp is not the crucial part of the functionality and the logger would have some value even if `clock_gettime` is broken. `getCurrentMonotonicTime` only has one functionality and when `clock_gettime` is broken all of the functionality is broken.

Therefore I'd like to have the fast failure for `getCurrentMonotonicTime`. If you also want to have it in `createLogMessageHeader` feel free to change it.

elBoberido: @mossmaurice if you have to commit anyway please either add a TODO or an `Ensures` and a `IOX_LOG(FATAL)` ... oh, in this case it might be a good idea not to have an `Ensures` in the `createLogMessageHeader`. I just remember why I needed the resilience :sweat_smile:

elBoberido: ... oh, another reason why an `Ensures` would be a good idea is that one could have code like this
```cpp
deadline_timer t{1_ms};
while(!t.has_expired()) {
    // polling code
}
```

One wouldn't even easily notice that the code is broken but just notice a high CPU load.

mossmaurice: @elBoberido Makes sense, added. I chose to wrap the call in a `EnsuresWithMsg` as this our strategy for non-recoverable errors (see #1902). I left the logger untouched and added a further comment.

iceoryx_posh/test/moduletests/test_runtime_ipc_interface.cpp
```diff
@@ -333,7 +334,7 @@ TYPED_TEST(IpcInterface_test, TimedReceiveWorks)
     ASSERT_FALSE(this->server->timedReceive(timeout, receivedMessage));
     auto after = system_clock::now();
 
-    auto timeDiff = units::Duration(after - before);
-    EXPECT_GE(timeDiff, timeout);
+    auto timeDiff = std::chrono::duration_cast<std::chrono::milliseconds>(after - before);
+    EXPECT_GE(into<units::Duration>(timeDiff), timeout);
```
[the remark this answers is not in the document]
mossmaurice: @FerdinandSpitzschnueffler For Windows and MSVC an explicit `duration_cast` is needed otherwise an implementation for `into` from the type is needed `iox::into<iox::units::Duration,std::chrono::duration<std::chrono::system_clock::rep,std::chrono::system_clock::period>>`, see [here](https://github.com/eclipse-iceoryx/iceoryx/actions/runs/4374027889/jobs/7652951288).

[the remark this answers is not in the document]
elBoberido: @FerdinandSpitzschnueffler on top of what @mossmaurice wrote, I'm not even fully sure why it worked before. `now` returns a `time_point` and subtracting two `time_points` returns some `common_type::type` template magic which tries to deduce the correct type. I guess it just happens to work by accident.

iceoryx_dust/include/iceoryx_dust/cxx/std_chrono_support.hpp
```diff
@@ -0,0 +1,65 @@
...
+namespace iox
+{
+/// @brief Construct a Duration object from std::chrono::nanoseconds
+/// @param[in] value as nanoseconds
+/// @attention Since negative durations are not allowed, the duration will be clamped to 0
+template <>
+struct FromImpl<std::chrono::nanoseconds, units::Duration>
+{
+    static units::Duration fromImpl(const std::chrono::nanoseconds& value) noexcept;
+};
```
elBoberido, line 33: Nothing for now but something for an interested contributor. I think all four specializations could be combined into one with a partial specialization for `std::chrono::duration<Rep, Period>`. Although, this might also open a can of worms.

iceoryx_hoofs/posix/time/source/deadline_timer.cpp
```diff
@@ -48,11 +51,18 @@ iox::units::Duration deadline_timer::remainingTime() const noexcept
...
 
 iox::units::Duration deadline_timer::getCurrentMonotonicTime() noexcept
 {
-    return iox::units::Duration{std::chrono::steady_clock::now().time_since_epoch()};
+    timespec time_since_epoch{0, 0};
+    cxx::EnsuresWithMsg(!posix::posixCall(clock_gettime)(CLOCK_MONOTONIC, &time_since_epoch)
+                             .failureReturnValue(-1)
+                             .evaluate()
+                             .has_error(),
+                        "An error which should never happen occured during 'clock_gettime'!");
```
elBoberido, line 65: Just in case you will be the reviewer when this is refactored with the new error handling. Please keep in mind to point out that `IOX_UNREACHABLE` shall be used :)

elBoberido (approved):

FerdinandSpitzschnueffler (approved):

PR #1964, opened 2023-04-05
iox-#1963 Fix UB in publishCopyOf and publishResultOf

### LINKED ISSUE #1963: Undefined behaviour in `publishCopyOf` and `publishResultOf`
## Required information

**Operating system:**
Ubuntu 22

**Compiler version:**
GCC 9.4

**Eclipse iceoryx version:**
Master branch

There is undefined behaviour in `iox::popo::Publisher::publishCopyOf` and `iox::popo::Publisher::publishResultOf`.

For `publishCopyOf`, copy-assignment operator is used to copy the user-given instance of `T` into the sample memory. However, there might not be any validly constructed instance of `T` in the sample memory before this copy-assignment! So, for example if the copy-assignment operator is implemented with some non-trivial destruction of existing data (before the assignment part), things can easily go wrong (because there might be some left-over random data from previous usages of this chunk). In general, I believe it is UB to do this. See here: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl#L64

For `publishResultOf`, it is similar. The user-given callable must have the first argument `T*`. And again, there might not be any validly constructed instance of `T` when it is passed to the callable. So, if the user callable is using copy-assignment operator to assign to the sample memory, this will be UB again. See here: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl#L55

Note that this is in contrast to the usual `loan()` method, where there is placement new operator called, which ensures that a valid instance is constructed in the memory: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl#L40

No UB.

I met this when using a `iox::vector` as `T`, and the copy assignment operator got stuck trying to destroy existing samples, because it believed that its size was huge.

JakubSosnovec:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer
Not sure whether we need a test for this bug. We could do something like this: Configure roudi to have just one buffer in its memory pool. Then publish some random data of data type U. Then publish data of T , and in the copy-assigment operator of T check whether there is the random data from publication of U. Should I try to write such a test?

## References

- Closes #1963

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl
@@ -1,5 +1,6 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2020 - 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2023 NXP. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl
+++ b/iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl
@@ -52,7 +52,7 @@ inline expected<AllocationError> PublisherImpl<T, H, BasePublisherType>::publish
                   "callable provided to Publisher<T>::publishResultOf must have signature void(T*, ArgsTypes...)");
 
     return loanSample().and_then([&](auto& sample) {
-        c(sample.get(), std::forward<ArgTypes>(args)...);
+        c(new (sample.get()) T, std::forward<ArgTypes>(args)...);
         sample.publish();
     });
 }
@@ -61,7 +61,8 @@ template <typename T, typename H, typename BasePublisherType>
 inline expected<AllocationError> PublisherImpl<T, H, BasePublisherType>::publishCopyOf(const T& val) noexcept
 {
     return loanSample().and_then([&](auto& sample) {
-        *sample.get() = val; // Copy assignment of value into sample's memory allocation.
+        new (sample.get()) T(val); // Placement new copy-construction of sample, avoid copy-assigment because there
+                                   // might not be an existing instance of T in the sample memory
         sample.publish();
     });
 }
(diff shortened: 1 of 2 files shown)


JakubSosnovec:
I just noticed ECA check failed for me, that is because I changed my email in my Eclipse account, due to some legal stuff. I have now changed it back to jakub.sosnovec@nxp.com which should match with the commits. If there is some way to re-trigger the ECA check, it should pass now. Sorry for the mess.

elfenpiff (approved):

elfenpiff (commented):
As soon as the ECA is signed and it is added to the release notes we can get it merged.

JakubSosnovec:
> As soon as the ECA is signed and it is added to the release notes we can get it merged.

ECA seems green now. And I just pushed a commit with release notes updated.

iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl
```diff
@@ -52,7 +53,7 @@ inline expected<AllocationError> PublisherImpl<T, H, BasePublisherType>::publish
                   "callable provided to Publisher<T>::publishResultOf must have signature void(T*, ArgsTypes...)");
 
     return loanSample().and_then([&](auto& sample) {
-        c(sample.get(), std::forward<ArgTypes>(args)...);
+        c(new (sample.get()) T, std::forward<ArgTypes>(args)...);
```
elBoberido, line 56: This is tricky. If `T` does not have a default ctor this breaks. On the other hand, the code is already broken. Ideally we would have something like Rusts `MaybeUninit<T>` and pass that to the callable.

I admire the problem but don't have a solution.

@elfenpiff what's your opinion on this? Live with the restriction of requiring a default ctor for now and get rid of the undefined behavior?

elBoberido (commented):
Good catch. Just one little thing to think about.

iceoryx_posh/include/iceoryx_posh/internal/popo/publisher_impl.inl
JakubSosnovec: Yeah, I noticed that too. If `T` doesnt have default ctor, then the signature of the callable doesnt make much sense. We could pass in the sample address as `void*`, with the expectation that the callable will placement-new-construct `T` there, but that doesnt play nice with the modern C++ style of iceoryx. Alternatively, we could SFINAE-ize `publishResultOf` so that it is only available when `T` does have default ctor. Or, finally, the callable could just return `T`, which we would again placement-new-copy-construct then, and if the copy-elision mechanisms kick in, that should be efficient. But that would be a breaking public API change...

JakubSosnovec: Or rewrite it in Rust! 🚀 😆

elBoberido: I think your current solution is okay. Didn't see that @elfenpiff already approved. Since he seems to also be okay with this LGTM

elBoberido (approved):

elfenpiff (approved):

PR #2020, opened 2023-09-05
iox-#1030 Speed up posh tests

### LINKED ISSUE #1030: Speed up posh tests
## Brief feature description

Some of our tests, e.g. when publisher subscriber communication is tested, call `InterOpWait` provided by the `RouDiEnvironment` to ensure that the discovery loop has been run through. Adding a function to the `RouDiEnvironment` to trigger the discovery loop manually would make these tests faster and more reliable.

In addition quite a lot of time is spent in `memset` to zero the shared memory. This time could be reduced by having a minimal mempool config for the tests. In addition, the introspection mempools should also be configurable via the `RoudiConfit_t`

## Tasks

- [x] add manual trigger for the RouDi discovery loop (e.g. via a `WaitSet` and a `UserTrigger`)
- [x] provide a `RouDiConfig_t` with a minimal memory footprint
- [x] add an option to `RouDiConfit_t` to configure the number of chunks for the introspection

### LINKED ISSUE #1359: ServiceDiscovery uses instrospection MemPools
## Required information

**Operating system:**
all

**Compiler version:**
all

The ServiceDiscovery uses the MemPools which are dedicated to the introspection. Currently this works by luck since there is an introspection topic which requires larger chunks than the service discovery data. When this changes, e.g. due to a refactoring, the ServiceDiscovery won't be able to allocate chunks and will be broken.

Additionally, the introspection will interfere with the ServiceDiscovery since it uses the same chunks, which will reduce the number of possible clients using the ServiceDiscovery.

The ServiceDiscovery should have a dedicated MemPool. This could of course be placed in the same MemoryManager like the introspection MemPools. A renaming of some members and methods would then be necessary to prevent confusion.

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR reduces the test time by more than two minutes. Some CI jobs are now even four minutes faster compared to the latest master build.

This is also getting a bit out of hand since the faster shutdown of RouDi uncovered some issues in the examples which happened previously only sporadically.

## References

- Closes #1030
- Closes #1359


iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -1071,13 +1078,11 @@ void PortManager::addPublisherToServiceRegistry(const capro::ServiceDescription&
         IOX_LOG(WARN) << "Could not add publisher with service description '" << service << "' to service registry!";
         errorHandler(PoshError::POSH__PORT_MANAGER_COULD_NOT_ADD_SERVICE_TO_REGISTRY, ErrorLevel::MODERATE);
     });
-    publishServiceRegistry();
 }
 
 void PortManager::removePublisherFromServiceRegistry(const capro::ServiceDescription& service) noexcept
 {
     m_serviceRegistry.removePublisher(service);
-    publishServiceRegistry();
```
elBoberido, line 1080: Oh boy, I remember the bikeshedding regarding the right (i.e. fastest) data structure for the `ServiceRegistry` well enough but it seems after the implementation nobody bothered to look at the slow tests with flamegraph to discover that there is something an order of magnitude slower than the most naive data structure one can use. This is a prime example why one should measure first and then optimize the bottleneck.

iceoryx_examples/icediscovery/iox_wait_for_service.cpp
```diff
@@ -44,14 +46,15 @@ void printSearchResult(const discovery::ServiceContainer& result)
...
 }
 
-Discovery* discoveryPtr{nullptr};
+volatile Discovery* discoverySigHandlerAccess{nullptr};
 
 void sigHandler(int)
 {
     //! [unblock wait]
-    if (discoveryPtr)
+    keepRunning = false;
```
FerdinandSpitzschnueffler, line 54: Why do we need `keepRunning` here? Shouldn't `discoverySigHandlerAccess` suffice?

elBoberido: They have two different purposes. `discoverySigHandlerAccess` could be set to a `nullptr` to indicate the while-loop to end. But then `discoverySigHandlerAccess` would be misused to indicate the loop condition and it would be less clear than with `keepRunning`. In general I try to avoid surrogates to not frustrate my future self too much.

iceoryx_posh/include/iceoryx_posh/roudi/roudi_config.hpp
```diff
@@ -26,6 +26,13 @@ namespace config
 {
 struct RouDiConfig
 {
+    // have some spare chunks to still deliver introspection data in case there are multiple subscriber to the data
```
FerdinandSpitzschnueffler, line 29:
```suggestion
    // have some spare chunks to still deliver introspection data in case there are multiple subscribers to the data
```

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -704,4 +704,46 @@ TYPED_TEST(ServiceRegistry_test, FindWithMixOfPublishersAndServersWorks)
...
 }
 
+TYPED_TEST(ServiceRegistry_test, HasDataChangedSinceLastCallReturnsTrueOnInitialCall)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "51398abb-53b2-4dce-9267-73f02f9d7574");
+
+    EXPECT_TRUE(this->sut.registry.hasDataChangedSinceLastCall());
+}
+
+TYPED_TEST(ServiceRegistry_test, HasDataChangedSinceLastCallReturnsFalsOnSubsequentCall)
```
FerdinandSpitzschnueffler, line 714:
```suggestion
TYPED_TEST(ServiceRegistry_test, HasDataChangedSinceLastCallReturnsFalseOnSubsequentCall)
```

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -704,4 +704,46 @@ TYPED_TEST(ServiceRegistry_test, FindWithMixOfPublishersAndServersWorks)
...
+
+    EXPECT_TRUE(this->sut.registry.hasDataChangedSinceLastCall());
+    EXPECT_FALSE(this->sut.registry.hasDataChangedSinceLastCall());
+}
+
+TYPED_TEST(ServiceRegistry_test, HasDataChangedSinceLastCallReturnsTrueAfterAddingService)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "17d5b84a-abe0-46e0-aa06-2d049c716b22");
+
+    iox::capro::ServiceDescription service1("a", "a", "a");
```
FerdinandSpitzschnueffler, line 726: Why not just `service`?

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
```diff
@@ -704,4 +704,46 @@ TYPED_TEST(ServiceRegistry_test, FindWithMixOfPublishersAndServersWorks)
...
+    ASSERT_FALSE(this->sut.add(service1).has_error());
+
+    EXPECT_TRUE(this->sut.registry.hasDataChangedSinceLastCall());
+}
+
+TYPED_TEST(ServiceRegistry_test, HasDataChangedSinceLastCallReturnsTrueAfterRemovingService)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "a4f0c9e2-2549-4fa0-88d4-75a2ef8714b8");
+
+    iox::capro::ServiceDescription service1("a", "a", "a");
```
FerdinandSpitzschnueffler, line 739: Same here.

iceoryx_posh/test/moduletests/test_roudi_iceoyx_roudi_memory_manager.cpp
```diff
@@ -94,6 +103,17 @@ TEST_F(IceoryxRoudiMemoryManager_test, AcquiringIntrospectionMemoryManagerAfterC
     EXPECT_THAT(result, Not(Eq(iox::nullopt_t())));
 }
 
+TEST_F(IceoryxRoudiMemoryManager_test, AcquiringDiscoveryMemoryManagerAfterCreateAndAnnounceMemoryIsSuccessful)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "ce42929a-0721-41b9-bf08-b48e53fe901e");
+    auto tr = m_roudiMemoryManagerTest->createAndAnnounceMemory();
```
FerdinandSpitzschnueffler, line 109: What does `tr` stand for?

iceoryx_posh/test/moduletests/test_roudi_iceoyx_roudi_memory_manager.cpp
```diff
@@ -94,6 +103,17 @@ TEST_F(IceoryxRoudiMemoryManager_test, AcquiringIntrospectionMemoryManagerAfterC
...
+
+    EXPECT_THAT(tr.has_error(), Eq(false));
+
+    auto result = m_roudiMemoryManagerTest->discoveryMemoryManager();
+    EXPECT_THAT(result, Not(Eq(iox::nullopt_t())));
```
FerdinandSpitzschnueffler, line 114:
```suggestion
    EXPECT_TRUE(result.has_value());
```

FerdinandSpitzschnueffler (commented):
I'm not finished with the review yet, will continue later.

elBoberido:
I think I found the reason (or at least one) for sporadic failures of the integration tests. It seems there was a race between launch_testing and RouDi trying to terminate the applications. I fixed the issue at the cost of additional 200 lines to review.

iceoryx_posh/test/moduletests/test_roudi_service_registry.cpp
elBoberido: c&p with shutdown of main brain functionality :)

iceoryx_posh/test/moduletests/test_roudi_iceoyx_roudi_memory_manager.cpp
elBoberido: I guess short for `test result`. I just c&p a test and adjusted it.

FerdinandSpitzschnueffler: Forget my question ;)

iceoryx_examples/icediscovery/iox_wait_for_service.cpp
FerdinandSpitzschnueffler: Okay, I got it, but I have another question: shouldn't we use an atomic volatile bool to ensure that `keepRunning` is synchronized?

elBoberido: The variable does not need to be thread safe but only async signal safe, i.e. just atomic with respect to signals. This means that read/write operations on variables must be executed in one instruction. Furthermore, `volatile` tells the compiler that the variable can be changed from another location not visible to the compiler and therefore it always needs to be read from the memory and not from the register and also cannot be optimized away. Keep in mind, the OS does not start a new thread for the execution of the signal handler but executes it in an already [running thread](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap02.html#tag_22_02_04_02). You can kind of think of it like a `goto` to the signal handler when the signal is raised. This is also the reason why e.g. `pthread_mutex_lock` is not [signal safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html) since there might be nobody left to unlock the mutex if it is already locked.

In [many](https://en.cppreference.com/w/c/program/sig_atomic_t) [examples](https://www.gnu.org/software/libc/manual/html_node/Handler-Returns.html) for the signal handler one can find a definition like `volatile sig_atomic_t keepRunning = 1;`. In this case `sig_atomic_t` is not an atomic data type but just a type alias to an integer type which can be accessed as an [atomic](https://en.cppreference.com/w/c/program/sig_atomic_t) [entity](https://www.gnu.org/software/libc/manual/html_node/Atomic-Types.html). This is also true for pointer and since a boolean is smaller than an `int`, it is also true for `bool`.

[Here](https://www.gnu.org/software/libc/manual/html_node/Non_002datomic-Example.html) you can find a good explanation about non-atomic access. Basically it boils down to torn writes and torn reads.

If you are more interested in the topic and have some time, this is an interesting read -> https://www.gnu.org/software/libc/manual/html_node/Signal-Handling.html

iceoryx_posh/test/integrationtests/test_service_discovery.cpp
```diff
@@ -1019,10 +1037,14 @@ TYPED_TEST(ServiceDiscoveryFindService_test, FindInMaximumMixedServices)
         this->add(randomService("Ferdinand", "Spitz"));
     }
 
+    IOX_LOG(FATAL) << "#### start";
```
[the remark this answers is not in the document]
elBoberido: No, no. I just wanted to check it the reviewer pay attention ;)

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -170,7 +250,7 @@ void RouDi::processRuntimeMessages() noexcept
     {
         // read RouDi's IPC channel
         runtime::IpcMessage message;
-        if (roudiIpcInterface.timedReceive(m_runtimeMessagesThreadTimeout, message))
+        if (roudiIpcInterface.timedReceive(10_ms /*m_runtimeMessagesThreadTimeout*/, message))
```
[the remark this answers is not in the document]
elBoberido: Ahem :sweat_smile:

iceoryx_examples/waitset_in_c/ice_c_waitset_timer_driven_execution.c
```diff
@@ -62,10 +64,16 @@ void* cyclicTriggerCallback(void* dontCare)
...
     {
-        iox_user_trigger_trigger(cyclicTrigger);
-        sleep_for(1000);
+        if (countdownToTrigger == 0)
+        {
+            iox_user_trigger_trigger(cyclicTrigger);
+            countdownToTrigger = 100;
+        }
+        sleep_for(10);
+        --countdownToTrigger;
```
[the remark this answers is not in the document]
elBoberido: I changed that before I discovered the real issue with the failing integration tests. This change resulted in a faster shutdown of the application which "solved" the problem for this example but actually only made it less likely to occur. I'll change it back.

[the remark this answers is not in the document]
elBoberido: ... on a second thought, it will speed up the tests so if you don't mind I would keep it anyway.

iceoryx_examples/icediscovery/iox_wait_for_service.cpp
elBoberido: Btw, here is a short example https://godbolt.org/z/deqsPjcsY
```cpp
volatile int v = 0;             // (1)
volatile sig_atomic_t va = 0;   // (2)
std::atomic<int> a = 0;         // (3)

struct Point {
    uint64_t x {0};
    uint64_t y {0};
} p;                            // (4)

void setVolatile(int num) {
    v = num;
}

void setVolatileSigAtomic(int num) {
    va = num;
}

void setAtomic(int num) {
    a.store(num, std::memory_order_relaxed);
}

void setPoint(Point point) {
    p = point;
}
```
When compiled with `-O1` this is the result
```asm
setVolatile(int):
        mov     DWORD PTR v[rip], edi
        ret
setVolatileSigAtomic(int):
        mov     DWORD PTR va[rip], edi
        ret
setAtomic(int):
        mov     DWORD PTR a[rip], edi
        ret
setPoint(Point):
        mov     QWORD PTR p[rip], rdi
        mov     QWORD PTR p[rip+8], rsi
        ret
```
As you can see on , the first three variants compile to the same instruction and are equivalent. The last one does not fulfill the requirement that the operation is atomic. Contrary to atomic data types, the usage of the first two variants is limited since they do not prevent reordering and don't have CAS operations but for our purpose it is equivalent to an atomic which supports only relaxed read and write operations.

FerdinandSpitzschnueffler: Ah okay, thanks for the explanation and the links you provided!

iceoryx_examples/waitset/ice_waitset_gateway.cpp
```diff
@@ -24,20 +24,24 @@
 #include <chrono>
 #include <iostream>
 
+constexpr uint64_t NUMBER_OF_SUBSCRIBERS = 2U;
+
 std::atomic_bool keepRunning{true};
```
FerdinandSpitzschnueffler, line 29: Is there a specific reason why you keep the `atomic bool` here and the following waitset examples, but changed it everywhere else to a `volatile bool`?

iceoryx_posh/source/roudi/roudi.cpp
```diff
@@ -147,15 +175,67 @@ void RouDi::cyclicUpdateHook() noexcept
...
+      public:
+        DiscoveryWaitSet(popo::ConditionVariableData& condVarData) noexcept
+            : WaitSet(condVarData)
+        {
+        }
+    };
+
+    popo::ConditionVariableData conditionVariableData;
+    DiscoveryWaitSet discoveryLoopWaitset{conditionVariableData};
+    discoveryLoopWaitset.attachEvent(m_discoveryLoopTrigger).expect("Successfully attaching a single event");
```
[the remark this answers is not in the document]
elBoberido: Oh boy, git blame tells me I was the only one using the non-non-expected variant for the string :sweat_smile:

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Well... Reading the line of code, `expect("what to expect")` reads better, but when it fails, I'd be confused to read "Successfully whatever" on the console :thinking:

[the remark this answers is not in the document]
elBoberido: What do you think about something like `Expected 'what to expect' but failed` or `Expected 'Successfully whatever' but failed` on the console?

Btw, in Rust the output on the console is like `panic at 'what to expect'`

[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Sounds good. Btw. in my opinion, you don't have to change this if there's nothing else you want to commit (or have to, depending on the second reviewer :wink:)

FerdinandSpitzschnueffler (approved):

iceoryx_posh/include/iceoryx_posh/roudi/roudi_cmd_line_parser.hpp
```diff
@@ -63,12 +63,7 @@ class CmdLineParser
           const CmdLineArgumentParsingMode cmdLineParsingMode = CmdLineArgumentParsingMode::ALL) noexcept;
 
   protected:
-    bool m_run{true};
-    iox::log::LogLevel m_logLevel{iox::log::LogLevel::INFO};
-    roudi::MonitoringMode m_monitoringMode{roudi::MonitoringMode::OFF};
-    version::CompatibilityCheckLevel m_compatibilityCheckLevel{version::CompatibilityCheckLevel::PATCH};
-    optional<uint16_t> m_uniqueRouDiId;
-    units::Duration m_processKillDelay{roudi::PROCESS_DEFAULT_KILL_DELAY};
+    CmdLineArgs_t m_cmdLineArgs;
```
mossmaurice, line 66: :heart:

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -53,13 +53,13 @@ PortManager::PortManager(RouDiMemoryInterface* roudiMemoryInterface) noexcept
     }
```
mossmaurice, line 53: @elBoberido Was the `chmod` on this file on purpose?

iceoryx_posh/source/roudi/port_manager.cpp
```diff
@@ -53,13 +53,13 @@ PortManager::PortManager(RouDiMemoryInterface* roudiMemoryInterface) noexcept
...
     m_portPool = maybePortPool.value();
 
-    auto maybeIntrospectionMemoryManager = m_roudiMemoryInterface->introspectionMemoryManager();
-    if (!maybeIntrospectionMemoryManager.has_value())
+    auto maybediscoveryMemoryManager = m_roudiMemoryInterface->discoveryMemoryManager();
```
mossmaurice, line 56: Nitpick:

```suggestion
    auto maybeDiscoveryMemoryManager = m_roudiMemoryInterface->discoveryMemoryManager();
```

mossmaurice (approved):
@elBoberido LGTM, thanks for the PR! 

One wish: Could we please move the C binding move to another repository before the `v3.0` C release like we did with the other optional bindings? This PR would have been much smaller and I prefer reviewing things step-by-step. Also, like with the other bindings we only have to update them if there is a need and only once after the release.

Shall I create issue for that?

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: Actually no. I have no idea how I did this but it actually makes sense. All the other files also have `644` and the executable does not make sense for a source file

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: Will change this in the follow up PR

elBoberido:
@mossmaurice I would keep the bindings for the v3 release and only move them afterwards. In general it does not cost us much and especially if we agree on the reduced scope of the v3 release it would not be much extra work. It might even be less work than to remove them. At least from some comments in some issues I have the feeling some people are using Cyclone DDS with the latest iceoryx master and for them it would be nice to have the bindings in the v3 release. Doing the move right after the v3 release would also result in a better transition path for these users.

cc @dkroenke

mossmaurice:
> I would keep the bindings for the v3 release and only move them afterwards. In general it does not cost us much and especially if we agree on the reduced scope of the v3 release it would not be much extra work. It might even be less work than to remove them. At least from some comments in some issues I have the feeling some people are using Cyclone DDS with the latest iceoryx master and for them it would be nice to have the bindings in the v3 release. Doing the move right after the v3 release would also result in a better transition path for these users.

@elBoberido That makes sense to me.

Could please still create the issue for the move to a separate repository and assign it to the `v4.0` project? Thanks :pray:

iceoryx_posh/source/roudi/port_manager.cpp
elBoberido: Fixed with https://github.com/eclipse-iceoryx/iceoryx/pull/2021/commits/2251d635530ef2b7c44bb1b97aca20fb5662cf9b

elBoberido:
@mossmaurice done with #2030

PR #2064, opened 2023-10-26
iox-#1391 Move dust header from legacy path to module path part 3

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR moves the posix wrapper in iceoryx dust to the new module location.

Additionally `IOX_EXPECTS_WITH_MSG` and `IOX_DEPRECATED_SINCE` is used where appropriate.

Finally it also fixes a race condition in the `PortUser_IntegrationTest` which was uncovered on the aarch64 build.

## References

- Relates to #1391

### DIFF
--- a/.cirrus.yaml
+++ b/.cirrus.yaml
@@ -13,6 +13,18 @@
 ---
 
 #
+# Filter to run the CI only on the master and release branches or for pull request to the master, release* and iox-* branches
+#
+
+only_if: $CIRRUS_BRANCH == 'master'
+         || $CIRRUS_BRANCH == 'release*'
+         || ($CIRRUS_PR != '' && ($CIRRUS_BASE_BRANCH == 'master'
+                                  || $CIRRUS_BASE_BRANCH == 'release*'
+                                  || $CIRRUS_BASE_BRANCH == 'iox-*'
+                                  )
+            )
+
+#
 # Templates
 #
 
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,4 +1,5 @@
 https://github.com/eclipse-iceoryx/iceoryx/compare/vx.x.x...vx.x.x
 https://github.com/eclipse-iceoryx/iceoryx/tree/vx.x.x
+https://www.misra.org.uk/
 
 iceoryx-dev@eclipse.org
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -1006,7 +1006,7 @@
     #include "iceoryx_hoofs/posix_wrapper/named_pipe.hpp"
 
     // after
-    #include "iceoryx_dust/posix_wrapper/named_pipe.hpp"
+    #include "iox/named_pipe.hpp"
     ```
 
     ```cpp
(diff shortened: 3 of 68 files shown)


elBoberido:
@mossmaurice @elfenpiff after moving the first posix wrapper to the new module structure and making the include structure flat I'm a bit hesitant with the flat namespace and includes for posix and concurrent.

I think it might make sense to have a `iox/posix/foo.hpp` include and a respective `iox::posix::FOO` access. There are quite some helper types which might class when we don't do this.

There is also half of a precedence for this, we use `iox::units::Duration` instead of just `iox::Duration`. Although the include is `iox/duration.hpp`. This is also confusing and we should move the include to `iox/units/duration.hpp`.

In general I thin we should have the flat structure for general constructs like `vector` or when the class is not part of a bigger logical group of its own. `posix` and `concurrent` are bigger logical groups which have some implications when used and would warrant an additional layer in the namespace and include path.

Do you agree to keep those namespaces and have the include path mirror the namespace (with the exception of the detail folder)?

elBoberido:
@dkroenke @MatthiasKillat @FerdinandSpitzschnueffler  ^^^ although you were not part of the original discussion, your input is also welcome

FerdinandSpitzschnueffler:
> @mossmaurice @elfenpiff after moving the first posix wrapper to the new module structure and making the include structure flat I'm a bit hesitant with the flat namespace and includes for posix and concurrent.
> 
> I think it might make sense to have a `iox/posix/foo.hpp` include and a respective `iox::posix::FOO` access. There are quite some helper types which might class when we don't do this.
> 
> There is also half of a precedence for this, we use `iox::units::Duration` instead of just `iox::Duration`. Although the include is `iox/duration.hpp`. This is also confusing and we should move the include to `iox/units/duration.hpp`.
> 
> In general I thin we should have the flat structure for general constructs like `vector` or when the class is not part of a bigger logical group of its own. `posix` and `concurrent` are bigger logical groups which have some implications when used and would warrant an additional layer in the namespace and include path.
> 
> Do you agree to keep those namespaces and have the include path mirror the namespace (with the exception of the detail folder)?

I agree that it makes sense to keep the `posix` and the `concurrent` namespace. Since you mentioned `Duration`: I don't see why we should keep the `units` namespace. Can't we just remove it? If something speaks against it, can we at least make the include path (`time`) and the namespace (`units`) consistent?

elBoberido:
@FerdinandSpitzschnueffler @mossmaurice @dkroenke @MatthiasKillat @elfenpiff @budrus feel free to add yourself as reviewer

I kept `posix` in the include path and as namespace for now. We also need to think about clear guidelines for this or remove it as part of #1593 because it is not as clear as I hoped. For example, there is a `deadline_timer` directly in `iox`. Initially it used `chrono` to get the current time but eventually was ported to `clock_gettime` which would make it a candidate to be moved to `posix` but then the include paths and namespaces would change. I'll put this on the agenda for the next meeting. We have to weight in all the pros and cons.

I'll continue to keep `posix` as namespace and in the include paths while moving the files to the new module structure for now. It can be changed later on with #1593 if we decide so.

FerdinandSpitzschnueffler (approved):

elfenpiff (approved):

Issue #2052, opened 2023-10-21
Implement copy/move constructors and assignment operators for 'FixedPositionContainer'


elBoberido:
## Brief feature description

The `FixedPositionContainer` is neither move nor copy constructible or assignable. Currently it is not required but would be a nice enhancement.

Dennis40816:
Hi, @elBoberido  I'm interesting about this one. Could I give it a try?

elBoberido:
@Dennis40816 sure. The PR which introduce the class is not merged yet. I would ping you once #2051 is merged. You can of course start with that branch but there might be some force pushes until it is merged.

Dennis40816:
@elBoberido Sure, I'll be around. Feel free to notify me once it's completed.

elBoberido:
@Dennis40816 it is merged. Looking forward to your PR :)

Dennis40816:
Hi, [elBoberido](https://github.com/elBoberido). As this is my first time making modifications to iceoryx, there might be areas I haven't fully considered, so I'd appreciate your assistance in reviewing. I have initially completed the target implementation and added unit tests.

Dennis40816:
In the move feature, there's an issue concerning the original container instance after move constructor or move assignment. Currently, I've only modified `m_size`, `m_begin_free`, and `m_begin_used`. However, I realized that not resetting `m_next` and `m_status` might lead to errors when the instance is reused. Should we reset the entire container, including `m_data`, `m_next`, and `m_status`, to make it reusable? Or is it better to let the user manually reset it or even prohibit its reuse?

elBoberido:
@Dennis40816 looking at other containers, we also call `clear` at the end of the move operations. Therefore it would be good to do the same here.

When you think you are done, just open an PR :)

PR #2069, opened 2023-10-29
iox-#2052 Implement move/copy constructor and assignment for `FixedPositionContainer` 

Dennis40816:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

- Move/copy constructors and assignment for `FixedPositionContainer` are implemented.
- At present, when invoking the move/copy constructor or assignment for `FixedPositionContainer`, only the move/copy assignment of elements within `FixedPositionContainer::m_data` gets triggered (without calling the constructor). This is 
because the `FixedPositionContainer` constructor is realized using assignment, mirroring the behavior of `vector`.
- Unit tests are added. To test the newly added code under the following scenarios, you can execute it via `build/dust/test/dust_moduletests`:
  - From an empty container
  - From a single-element container
  - From a multi-element container
  - From a full-capacity container
- For move cases, the criteria for determining whether `FixedPositionContainer::clear()` has been executed is based on detecting if the element (`ComplexType`) has invoked the `DTor`.

## References

- Closes **#2052**

### DIFF
--- a/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
+++ b/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
@@ -52,6 +52,84 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
 }
 
 template <typename T, uint64_t CAPACITY>
+FixedPositionContainer<T, CAPACITY>::FixedPositionContainer(const FixedPositionContainer& other) noexcept
+{
+    m_size = other.m_size;
+    m_begin_free = other.m_begin_free;
+    m_begin_used = other.m_begin_used;
+
+    for (IndexType i = 0; i < CAPACITY; ++i)
+    {
+        m_data[i] = T{other.m_data[i]};
+        m_status[i] = other.m_status[i];
+        m_next[i] = other.m_next[i];
+    }
+}
+
+template <typename T, uint64_t CAPACITY>
+FixedPositionContainer<T, CAPACITY>::FixedPositionContainer(FixedPositionContainer&& other) noexcept
+{
+    m_size = other.m_size;
+    m_begin_free = other.m_begin_free;
+    m_begin_used = other.m_begin_used;
+
+    for (IndexType i = 0; i < CAPACITY; ++i)
+    {
+        m_data[i] = T{std::move(other.m_data[i])};
+        m_status[i] = other.m_status[i];
+        m_next[i] = other.m_next[i];
+    }
+
+    other.m_size = 0;
+    other.m_begin_free = Index::FIRST;
+    other.m_begin_used = Index::INVALID;
+}
+
+template <typename T, uint64_t CAPACITY>
+FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(const FixedPositionContainer& other) noexcept
+{
+    if (this != &other)
+    {
+        m_size = other.m_size;
+        m_begin_free = other.m_begin_free;
+        m_begin_used = other.m_begin_used;
+
+        for (IndexType i = 0; i < CAPACITY; ++i)
+        {
+            m_data[i] = other.m_data[i];
+            m_status[i] = other.m_status[i];
+            m_next[i] = other.m_next[i];
+        }
+    }
+    return *this;
+}
+
+template <typename T, uint64_t CAPACITY>
+FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(FixedPositionContainer&& other) noexcept
+{
+    if (this != &other)
+    {
+        m_size = other.m_size;
+        m_begin_free = other.m_begin_free;
+        m_begin_used = other.m_begin_used;
+
+        for (IndexType i = 0; i < CAPACITY; ++i)
+        {
+            m_data[i] = std::move(other.m_data[i]);
+            m_status[i] = other.m_status[i];
+            m_next[i] = other.m_next[i];
+        }
+
+        other.m_size = 0;
+        other.m_begin_free = Index::FIRST;
+        other.m_begin_used = Index::INVALID;
+    }
+    return *this;
+}
+
+template <typename T, uint64_t CAPACITY>
 inline void FixedPositionContainer<T, CAPACITY>::clear() noexcept
 {
     for (IndexType i = 0; i < CAPACITY;)
--- a/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
+++ b/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
@@ -71,45 +71,51 @@ FixedPositionContainer<T, CAPACITY>::operator=(const FixedPositionContainer& rhs
     if (this != &rhs)
     {
         IndexType i = Index::FIRST;
-        auto lhs_size = size();
-        auto min_size = algorithm::minVal(lhs_size, rhs.size());
+        ConstIterator rhs_it = rhs.begin();
 
-        // copy using copy assignment
-        for (; i < min_size; ++i)
+        // copy using copy assignment util rhs iter becomes invalid
+        for (; rhs_it.to_index() != Index::INVALID; i++, rhs_it++)
         {
-            m_data[i] = rhs.m_data[i];
-            m_status[i] = rhs.m_status[i];
-            m_next[i] = rhs.m_next[i];
-        }
-
-        // with rhs.size bigger than this.size: copy further elements from rhs to this
-        for (; i < rhs.size(); ++i)
-        {
-            m_data[i] = rhs.m_data[i];
-            m_status[i] = rhs.m_status[i];
-            m_next[i] = rhs.m_next[i];
-        }
-
-        // with rhs.size smaller than this.size: delete remaining elements of this (erase)
-        for (; i < lhs_size; ++i)
-        {
-            erase(i);
-        }
-
-        // init rest status to avoid garbage value
-        for (; i < CAPACITY; ++i)
-        {
-            m_status[i] = SlotStatus::FREE;
+            if (m_status[i] == SlotStatus::USED)
+            {
+                // clean original data by it's dtor
+                erase(i);
+            }
+            m_data[i] = *rhs_it;
+            m_status[i] = SlotStatus::USED;
+            // can this line be removed?
             m_next[i] = static_cast<IndexType>(i + 1U);
         }
-        m_next[Index::LAST] = Index::INVALID;
+
+        // correct next
+        m_next[i] = Index::INVALID;
+
+        // erase rest USED element in rhs, m_next and m_status will be also modified by erase()
+        while (i < Index::INVALID)
+        {
+            if (m_status[i] == SlotStatus::USED)
+            {
+                break;
+            }
+
+            i++;
+        }
+
+        // if rest USED slot found
+        if (i != Index::INVALID)
+        {
+            for (Iterator lhs_it = this->iter_from_index(i); lhs_it.to_index() != Index::INVALID; lhs_it++)
+            {
+                erase(lhs_it);
+            }
+        }
 
         // member update
-        m_begin_free = rhs.m_begin_free;
-        m_begin_used = rhs.m_begin_used;
+        // may this cause problem when size == capacity?
+        m_begin_free = static_cast<IndexType>(rhs.size());
+        m_begin_used = Index::FIRST;
         m_size = rhs.m_size;
     }
-
     return *this;
 }
 
@@ -120,45 +126,51 @@ FixedPositionContainer<T, CAPACITY>::operator=(FixedPositionContainer&& rhs) noe
     if (this != &rhs)
     {
         IndexType i = Index::FIRST;
-        auto lhs_size = size();
-        auto min_size = algorithm::minVal(lhs_size, rhs.size());
+        Iterator rhs_it = rhs.begin();
 
-        // move using move assignment
-        for (; i < min_size; ++i)
+        // copy using copy assignment util rhs iter becomes invalid
+        for (; rhs_it.to_index() != Index::INVALID; i++, rhs_it++)
         {
-            m_data[i] = std::move(rhs.m_data[i]);
-            m_status[i] = rhs.m_status[i];
-            m_next[i] = rhs.m_next[i];
-        }
-
-        // with rhs.size bigger than this.size: move further elements from rhs to this
-        for (; i < rhs.size(); ++i)
-        {
-            m_data[i] = std::move(rhs.m_data[i]);
-            m_status[i] = rhs.m_status[i];
-            m_next[i] = rhs.m_next[i];
-        }
-
-        // with rhs.size smaller than this.size: delete remaining elements of this (erase)
-        for (; i < lhs_size; ++i)
-        {
-            erase(i);
-        }
-
-        // init rest status to avoid garbage value
-        for (; i < CAPACITY; ++i)
-        {
-            m_status[i] = SlotStatus::FREE;
+            if (m_status[i] == SlotStatus::USED)
+            {
+                // clean original data by it's dtor
+                erase(i);
+            }
+            m_data[i] = std::move(*rhs_it);
+            m_status[i] = SlotStatus::USED;
+            // can this line be removed?
             m_next[i] = static_cast<IndexType>(i + 1U);
         }
-        m_next[Index::LAST] = Index::INVALID;
+
+        // correct next
+        m_next[i] = Index::INVALID;
+
+        // erase rest USED element in rhs, m_next and m_status will be also modified by erase()
+        while (i < Index::INVALID)
+        {
+            if (m_status[i] == SlotStatus::USED)
+            {
+                break;
+            }
+            i++;
+        }
+
+        // if rest USED slot found
+        if (i != Index::INVALID)
+        {
+            for (Iterator lhs_it = this->iter_from_index(i); lhs_it.to_index() != Index::INVALID; lhs_it++)
+            {
+                erase(lhs_it);
+            }
+        }
 
         // member update
-        m_begin_free = rhs.m_begin_free;
-        m_begin_used = rhs.m_begin_used;
+        // may this cause problem when size == capacity?
+        m_begin_free = static_cast<IndexType>(rhs.size());
+        m_begin_used = Index::FIRST;
         m_size = rhs.m_size;
 
-        // clear rhs to avoid garbage value
+        // clear rhs
         rhs.clear();
     }
     return *this;
--- a/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
+++ b/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
@@ -52,79 +52,113 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
 }
 
 template <typename T, uint64_t CAPACITY>
-FixedPositionContainer<T, CAPACITY>::FixedPositionContainer(const FixedPositionContainer& other) noexcept
+inline FixedPositionContainer<T, CAPACITY>::FixedPositionContainer(const FixedPositionContainer& rhs) noexcept
 {
-    m_size = other.m_size;
-    m_begin_free = other.m_begin_free;
-    m_begin_used = other.m_begin_used;
-
-    for (IndexType i = 0; i < CAPACITY; ++i)
-    {
-        m_data[i] = T{other.m_data[i]};
-        m_status[i] = other.m_status[i];
-        m_next[i] = other.m_next[i];
-    }
+    *this = rhs;
 }
 
 template <typename T, uint64_t CAPACITY>
-FixedPositionContainer<T, CAPACITY>::FixedPositionContainer(FixedPositionContainer&& other) noexcept
+inline FixedPositionContainer<T, CAPACITY>::FixedPositionContainer(FixedPositionContainer&& rhs) noexcept
 {
-    m_size = other.m_size;
-    m_begin_free = other.m_begin_free;
-    m_begin_used = other.m_begin_used;
-
-    for (IndexType i = 0; i < CAPACITY; ++i)
-    {
-        m_data[i] = T{std::move(other.m_data[i])};
-        m_status[i] = other.m_status[i];
-        m_next[i] = other.m_next[i];
-    }
-
-    other.m_size = 0;
-    other.m_begin_free = Index::FIRST;
-    other.m_begin_used = Index::INVALID;
+    *this = std::move(rhs);
 }
 
 template <typename T, uint64_t CAPACITY>
-FixedPositionContainer<T, CAPACITY>&
-FixedPositionContainer<T, CAPACITY>::operator=(const FixedPositionContainer& other) noexcept
+inline FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(const FixedPositionContainer& rhs) noexcept
 {
-    if (this != &other)
+    if (this != &rhs)
     {
-        m_size = other.m_size;
-        m_begin_free = other.m_begin_free;
-        m_begin_used = other.m_begin_used;
+        IndexType i = 0U;
+        auto lhsSize = size();
+        auto minSize = algorithm::minVal(lhsSize, rhs.size());
 
-        for (IndexType i = 0; i < CAPACITY; ++i)
+        // copy using copy assignment
+        for (; i < minSize; ++i)
         {
-            m_data[i] = other.m_data[i];
-            m_status[i] = other.m_status[i];
-            m_next[i] = other.m_next[i];
+            m_data[i] = rhs.m_data[i];
+            m_status[i] = rhs.m_status[i];
+            m_next[i] = rhs.m_next[i];
         }
+
+        // with rhs.size bigger than this.size: copy further elements from rhs to this
+        for (; i < rhs.size(); ++i)
+        {
+            m_data[i] = rhs.m_data[i];
+            m_status[i] = rhs.m_status[i];
+            m_next[i] = rhs.m_next[i];
+        }
+
+        // with rhs.size smaller than this.size: delete remaining elements of this (erase)
+        for (; i < lhsSize; ++i)
+        {
+            erase(i);
+        }
+
+        // init rest status to avoid garbage value
+        for (; i < CAPACITY; ++i)
+        {
+            m_status[i] = SlotStatus::FREE;
+            m_next[i] = static_cast<IndexType>(i + 1U);
+        }
+        m_next[Index::LAST] = Index::INVALID;
+
+        // member update
+        m_begin_free = rhs.m_begin_free;
+        m_begin_used = rhs.m_begin_used;
+        m_size = rhs.m_size;
     }
+
     return *this;
 }
 
 template <typename T, uint64_t CAPACITY>
-FixedPositionContainer<T, CAPACITY>&
-FixedPositionContainer<T, CAPACITY>::operator=(FixedPositionContainer&& other) noexcept
+inline FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(FixedPositionContainer&& rhs) noexcept
 {
-    if (this != &other)
+    if (this != &rhs)
     {
-        m_size = other.m_size;
-        m_begin_free = other.m_begin_free;
-        m_begin_used = other.m_begin_used;
+        IndexType i = 0U;
+        auto lhsSize = size();
+        auto minSize = algorithm::minVal(lhsSize, rhs.size());
 
-        for (IndexType i = 0; i < CAPACITY; ++i)
+        // move using move assignment
+        for (; i < minSize; ++i)
         {
-            m_data[i] = std::move(other.m_data[i]);
-            m_status[i] = other.m_status[i];
-            m_next[i] = other.m_next[i];
+            m_data[i] = std::move(rhs.m_data[i]);
+            m_status[i] = rhs.m_status[i];
+            m_next[i] = rhs.m_next[i];
         }
 
-        other.m_size = 0;
-        other.m_begin_free = Index::FIRST;
-        other.m_begin_used = Index::INVALID;
+        // with rhs.size bigger than this.size: move further elements from rhs to this
+        for (; i < rhs.size(); ++i)
+        {
+            m_data[i] = std::move(rhs.m_data[i]);
+            m_status[i] = rhs.m_status[i];
+            m_next[i] = rhs.m_next[i];
+        }
+
+        // with rhs.size smaller than this.size: delete remaining elements of this (erase)
+        for (; i < lhsSize; ++i)
+        {
+            erase(i);
+        }
+
+        // init rest status to avoid garbage value
+        for (; i < CAPACITY; ++i)
+        {
+            m_status[i] = SlotStatus::FREE;
+            m_next[i] = static_cast<IndexType>(i + 1U);
+        }
+        m_next[Index::LAST] = Index::INVALID;
+
+        // member update
+        m_begin_free = rhs.m_begin_free;
+        m_begin_used = rhs.m_begin_used;
+        m_size = rhs.m_size;
+
+        // clear rhs to avoid garbage value
+        rhs.clear();
     }
     return *this;
 }
--- a/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
+++ b/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
@@ -93,6 +93,8 @@ template <typename T, uint64_t CAPACITY>
 template <typename RhsType>
 inline void FixedPositionContainer<T, CAPACITY>::copy_and_move_impl(RhsType&& rhs) noexcept
 {
+    // we already make sure rhs is always passed by std::move() for move case, therefore
+    // the result of "decltype(rhs)" is same as "decltype(std::forward<RhsType>(rhs))"
     static_assert(std::is_rvalue_reference<decltype(rhs)>::value
                       || (std::is_lvalue_reference<decltype(rhs)>::value
                           && std::is_const<std::remove_reference_t<decltype(rhs)>>::value),
@@ -100,14 +102,15 @@ inline void FixedPositionContainer<T, CAPACITY>::copy_and_move_impl(RhsType&& rh
 
     IndexType i = Index::FIRST;
     auto rhs_it = (std::forward<RhsType>(rhs)).begin();
-    bool is_move = std::is_rvalue_reference<RhsType&&>::value;
+    constexpr bool is_move = std::is_rvalue_reference<decltype(rhs)>::value;
 
     // transfer src data to destination
     for (; rhs_it.to_index() != Index::INVALID; ++i, ++rhs_it)
     {
         if (m_status[i] == SlotStatus::USED)
         {
-            if (is_move)
+#if __cplusplus >= 201703L
+            if constexpr (is_move)
             {
                 m_data[i] = std::move(*rhs_it);
             }
@@ -115,11 +118,20 @@ inline void FixedPositionContainer<T, CAPACITY>::copy_and_move_impl(RhsType&& rh
             {
                 m_data[i] = *rhs_it;
             }
+#else
+            // We introduce a helper struct primarily due to the test case
+            // e1cc7c9f-c1b5-4047-811b-004302af5c00. It demands compile-time if-else branching
+            // for classes that are either non-copyable or non-moveable.
+            // Note: With C++17's 'if constexpr', the need for these helper structs (labeled "C++ 14 feature"
+            // in fixed_position_container.hpp) can be eliminated.
+            AssignmentHelper<is_move>::assign(m_data[i], MoveHelper<is_move>::move_or_copy(rhs_it));
+#endif
         }
         else
         {
-            // use ctor to avoid UB for non-initialized free slots
-            if (is_move)
+// use ctor to avoid UB for non-initialized free slots
+#if __cplusplus >= 201703L
+            if constexpr (is_move)
             {
                 new (&m_data[i]) T(std::move(*rhs_it));
             }
@@ -127,6 +139,10 @@ inline void FixedPositionContainer<T, CAPACITY>::copy_and_move_impl(RhsType&& rh
             {
                 new (&m_data[i]) T(*rhs_it);
             }
+#else
+            // Same as above
+            CtorHelper<is_move>::construct(m_data[i], MoveHelper<is_move>::move_or_copy(rhs_it));
+#endif
         }
 
         m_status[i] = SlotStatus::USED;
--- a/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
+++ b/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
@@ -19,6 +19,9 @@
 #define IOX_DUST_CONTAINER_DETAIL_FIXED_POSITION_CONTAINER_INL
 
 #include "iox/fixed_position_container.hpp"
+#if __cplusplus < 201703L
+#include "iox/detail/fixed_position_container_helper.hpp"
+#endif
 
 namespace iox
 {
--- a/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
+++ b/iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
@@ -29,18 +29,7 @@ namespace iox
 template <typename T, uint64_t CAPACITY>
 inline FixedPositionContainer<T, CAPACITY>::FixedPositionContainer() noexcept
 {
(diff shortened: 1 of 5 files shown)


iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +51,118 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+}
+
+template <typename T, uint64_t CAPACITY>
+inline FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(const FixedPositionContainer& rhs) noexcept
+{
+    if (this != &rhs)
+    {
+        IndexType i = 0U;
+        auto lhsSize = size();
```
elBoberido, line 73: nitpick: we want to move to snake_case for methods and variables. It would save us some time later on if you could change this now :)

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +51,118 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+        auto minSize = algorithm::minVal(lhsSize, rhs.size());
+
+        // copy using copy assignment
+        for (; i < minSize; ++i)
+        {
+            m_data[i] = rhs.m_data[i];
+            m_status[i] = rhs.m_status[i];
+            m_next[i] = rhs.m_next[i];
```
elBoberido, line 81: This is not quite correct. You could have the following situation on either of the two containers
``` 
+---+---+---+---+---+---+ - +
| 0 | 1 | 2 | 3 | 4 | 5 | I |   <- index
+---+---+---+---+---+---+ - +
| F | F | U | U | U | F |   |   <- status (F = free; U = used)
+---+---+---+---+---+---+ - +
| 1 | 5 | 3 | 4 | I | I |   |   <- next index with same status
+---+---+---+---+---+---+ - +
```

I think you need to iterate over both container and use the iterator for the copy operation. With this you even don't need to change the status and next index.

Another option would be to iterate over `rhs` and fill `this` via the indices but then you have to distinguish if the slot is free or used.

The latter one might result in a slightly better performance since all used slots will be consecutive.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +51,118 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+        }
+
+        // with rhs.size bigger than this.size: copy further elements from rhs to this
+        for (; i < rhs.size(); ++i)
+        {
+            m_data[i] = rhs.m_data[i];
+            m_status[i] = rhs.m_status[i];
+            m_next[i] = rhs.m_next[i];
```
elBoberido, line 89: This has a similar problem than above.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +51,118 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+        }
+
+        // with rhs.size smaller than this.size: delete remaining elements of this (erase)
+        for (; i < lhsSize; ++i)
+        {
+            erase(i);
```
elBoberido, line 95: This too and it will even terminate when the slot behind `i` has the free status

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +51,118 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+        m_begin_used = rhs.m_begin_used;
+        m_size = rhs.m_size;
+    }
+
+    return *this;
+}
+
+template <typename T, uint64_t CAPACITY>
+inline FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(FixedPositionContainer&& rhs) noexcept
```
elBoberido, line 117: This needs to be changed as well, similar to the copy assignment operator.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
 
+// BEGIN test copy constructor
+
+TEST_F(FixedPositionContainer_test, UsingCopyCtorEmptyContainerResultsInEmptyContainer)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "6c528ef3-9c2d-4eb2-93a9-2d998d0db380");
+
+    SutComplex copy_sut_complex{sut_complex};
+
+    EXPECT_THAT(copy_sut_complex.empty(), Eq(true));
```
elBoberido, line 105: Please also check `full` and `size` here. No need to create another test for this.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+{
+    ::testing::Test::RecordProperty("TEST_ID", "f3aaf452-77fa-4535-bf0b-37bedefc2bf6");
+
+    constexpr DataType EXPECTED_VALUE{42U};
+    constexpr SutComplex::IndexType EXPECTED_SIZE{1U};
+
+    sut_complex.emplace(EXPECTED_VALUE);
+    SutComplex copy_sut_complex{sut_complex};
+
+    EXPECT_THAT(copy_sut_complex.size(), Eq(EXPECTED_SIZE));
```
elBoberido, line 118: Please also check for `empty` and `full` here.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+
+    // actually call copyAssignment
+    EXPECT_THAT(copy_sut_complex.begin()->stats.copyAssignment, Eq(EXPECTED_SIZE));
```
elBoberido, line 121: You can use just `stats`. It's a static member and all elements share the same value.
```suggestion
    EXPECT_THAT(stats.copyAssignment, Eq(EXPECTED_SIZE));
```

Please also change further down.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+TEST_F(FixedPositionContainer_test, UsingCopyCtorFullCapacityContainerPreservesAllElements)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "028704df-b2f3-4133-9c16-b9d2c6a79916");
+
+    fillSutComplex();
+    constexpr SutComplex::IndexType EXPECTED_SIZE{CAPACITY};
+
+    SutComplex copy_sut_complex{sut_complex};
+
+    EXPECT_THAT(copy_sut_complex.full(), Eq(true));
```
elBoberido, line 157: Please also check `size` here.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+    EXPECT_THAT(copy_sut_complex.begin()->stats.copyAssignment, Eq(EXPECTED_SIZE));
+    EXPECT_THAT(copy_sut_complex.begin()->value, Eq(EXPECTED_VALUE));
+}
+
+TEST_F(FixedPositionContainer_test, UsingCopyCtorMultipleElementsContainerPreservesAllElements)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "6261f53e-8089-4b9b-9b2d-9da0016a2f1e");
+
+    constexpr SutComplex::IndexType EXPECTED_SIZE{4U};
+    std::vector<DataType> EXPECTED_VALUE = {0U, 1U, 2U, 3U};
```
elBoberido, line 130: Please use random values when you manually add them. Just in case the container was constructed and the same memory location than the one from the test before and the memory accidentally have the same values.

I just realized I also did a mistake by not adding an `offset` parameter to the fill methods :/

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+    // actually call copyAssignment
+    EXPECT_THAT(copy_sut_complex.begin()->stats.copyAssignment, Eq(EXPECTED_SIZE));
+    for (Sut::IndexType i = 0; i < EXPECTED_SIZE; ++i)
+    {
+        EXPECT_THAT(copy_sut_complex.iter_from_index(i)->value, Eq(sut_complex.iter_from_index(i)->value));
+        EXPECT_THAT(copy_sut_complex.iter_from_index(i), Ne(sut_complex.iter_from_index(i)));
+    }
+}
+
+// END test copy constructor
```
elBoberido, line 168: There are some missing tests:
- copy empty  container to non empty one
- copy non-empty container with `size` being less than, equal and greater than than the destination container
- copy non-empty container with `Index::FIRST` erased to an empty one
- copy non-empty container with `Index::FIRST` erased to non empty one
- copy non-empty container with `Index::FIRST` erased to non empty one with `Index::FIRST` erased
- copy non-empty container with `Index::FIRST` erased to non empty one with index `Index::FIRST` and `Index::FIRST + 1` erased
- copy non-empty container with `Index::FIRST`, `Index::LAST / 2` and `Index::LAST` erased to an empty one
- copy non-empty container with `Index::FIRST`, `Index::LAST / 2` and `Index::LAST` erased to a non empty one
- copy non-empty container with `Index::FIRST`, `Index::LAST / 2` and `Index::LAST` erased to a non empty one with `Index::FIRST` erased
- copy non-empty container with `Index::FIRST`, `Index::LAST / 2` and `Index::LAST` erased to a non empty one with `Index::FIRST + 1`,  `Index::LAST / 2 - 1` and `Index::LAST - 1` erased
- copy non-empty container with `Index::LAST` erased to full container but with `Index::FIRST` erased
- add a test to check that the source container did not change (just in case someone does a const_cast)

This also need to be added to the other move ctor and assignment operator tests.

You can have a look at https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/concepts/best-practice-for-testing.md on how to come up with all these test cases :)

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+
+// BEGIN test move constructor
+
+TEST_F(FixedPositionContainer_test, UsingMoveCtorFromEmptyContainerResultsInEmptyContainer)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "af8958fb-9a09-4987-b290-ce41abdc2354");
+
+    SutComplex move_sut_complex{std::move(sut_complex)};
+
+    EXPECT_THAT(move_sut_complex.empty(), Eq(true));
```
elBoberido, line 178: Please check that the source container also remains empty

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +94,326 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+    SutComplex move_sut_complex{std::move(sut_complex)};
+
+    ASSERT_THAT(move_sut_complex.size(), Eq(EXPECTED_SIZE));
+
+    // actually call moveAssignment
+    EXPECT_THAT(move_sut_complex.begin()->stats.moveAssignment, Eq(EXPECTED_SIZE));
+
+    // make sure clear() been called
+    EXPECT_THAT(move_sut_complex.begin()->stats.dTor, Eq(EXPECTED_SIZE));
+    EXPECT_THAT(sut_complex.empty(), Eq(true));
```
elBoberido, line 198: Please also check `size`, `empty` and `full`. It is sufficient to do it only once for `sut_complex`

elBoberido (commented):
Feel free to add yourself as copyright holder to the files with significant changes :)

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: Apologies, but I am still somewhat unclear.

In the first scenario, I don't quite comprehend why there is no need to modify the `status` and `next` when using iterators to execute the copy operation. The dereferenced iterator `(*iter)` returns a reference to `m_data[i]`. If the value of `m_data` is updated via an assignment operation, it appears that the values for `status` and `next index` won't get updated. Or are you suggesting that these values get updated automatically through the use of functions like `emplace()`?

For the second scenario, I am not quite sure about the meaning of "all used slots will be consecutive".

```
// this
+---+---+---+---+---+---+ - +
| 0 | 1 | 2 | 3 | 4 | 5 | I |   <- index
+---+---+---+---+---+---+ - +
| F | F | U | U | U | F |   |   <- status (F = free; U = used)
+---+---+---+---+---+---+ - +
| 1 | 5 | 3 | 4 | I | I |   |   <- next index with same status
+---+---+---+---+---+---+ - +

// rhs
+---+---+---+---+---+---+ - +
| 0 | 1 | 2 | 3 | 4 | 5 | I |   <- index
+---+---+---+---+---+---+ - +
| F | U | F | U | F | F |   |   <- status (F = free; U = used)
+---+---+---+---+---+---+ - +
| 2 | 3 | 4 | I | 4 | I |   |   <- next index with same status
+---+---+---+---+---+---+ - +
```
Do you mean `this` will become following after copy?

```
// new this
+---+---+---+---+---+---+ - +
| 0 | 1 | 2 | 3 | 4 | 5 | I |   <- index
+---+---+---+---+---+---+ - +
| U | U | F | F | F | F |   |   <- status (F = free; U = used)
+---+---+---+---+---+---+ - +
| 1 | I | 3 | 4 | 5 | I |   |   <- next index with same status
+---+---+---+---+---+---+ - +
```

Also, if you think any document that I missed, please tell me. Thank you.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
Dennis40816: By "random", are you referring to a fixed random value like `constexpr DataType tmp = 42U`, or do you mean using a pseudorandom number generator like `rand()`? If needed, I can also add an offset parameter to both `fillSut` and `fillComplex`.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
Dennis40816: I think I get it now :O

Dennis40816: Some of the tests mentioned above may be applicable only to assignment operations. For instance, I believe the "empty container to non-empty one" scenario can't be accomplished using a constructor; it can only be done via assignment.

Here's how the code comment might look, revised for clarity:

```c
SutComplex tmp;
tmp.emplace(rand_num);

// I believe this can only be accomplished through assignment
tmp = sut_complex;

EXPECT_THAT(tmp.empty(), Eq(true));
// ...
```

Please correct me if I'm wrong, otherwise I'm inclined to implement tests only where necessary :)

elBoberido: Yes, that's correct. I think it would be even fine to have just one copy/move ctor test since they are implemented by delegating to the corresponding assignment operators anyway.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
elBoberido: No need to apologize :)

If you use the iterator on the source container you would iterate through all slots which contain valid data and therefore you could use the copy assignment operator on the dereferenced iterator, then increment the iterator and continue as long as it is not the end iterator. This will only change the `data` and and `status` and `next index` are already pointing to a valid used location. When the end iterator is reached, you would need to continue with `emplace` and if the destination container had more elements than the source container `erase` must be used on the remaining used slots.

Regarding the second option. Yes, that's exactly what I meant. I guess this option would result in less complex code, especially if `clear` would be called on the destination container before any of the other operations.

I think you now know about all the relevant documents for this PR :)

Dennis40816: Appreciate for your clear reply~

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +52,130 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+inline FixedPositionContainer<T, CAPACITY>&
+FixedPositionContainer<T, CAPACITY>::operator=(const FixedPositionContainer& rhs) noexcept
+{
+    if (this != &rhs)
+    {
+        IndexType i = Index::FIRST;
+        ConstIterator rhs_it = rhs.begin();
+
+        // copy using copy assignment util rhs iter becomes invalid
+        for (; rhs_it.to_index() != Index::INVALID; i++, rhs_it++)
```
elBoberido, line 77: nitpick: if the return value is not used, please use pre-increment :)

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +52,130 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+        {
+            if (m_status[i] == SlotStatus::USED)
+            {
+                // clean original data by it's dtor
+                erase(i);
+            }
+            m_data[i] = *rhs_it;
```
elBoberido, line 84: Using the copy assignment after deletion is undefined behavior. You can use copy assignment when the slot is used and placement new when it is free

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +52,130 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+        }
+
+        // if rest USED slot found
+        if (i != Index::INVALID)
+        {
+            for (Iterator lhs_it = this->iter_from_index(i); lhs_it.to_index() != Index::INVALID; lhs_it++)
+            {
+                erase(lhs_it);
+            }
+        }
```
elBoberido, line 111: `m_next` must also be updated for the free slots. I guess a simple for loop similar to the one above were the values are copied will work.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +52,130 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+
+        // member update
+        // may this cause problem when size == capacity?
+        m_begin_free = static_cast<IndexType>(rhs.size());
```
elBoberido, line 115: Theoretically not, since the index at capacity is the `INVALID` index. As long as the implementation does not change this we are safe and if the implementation changes the tests will fail so we are also safe :)

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +52,130 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+
+        // copy using copy assignment util rhs iter becomes invalid
+        for (; rhs_it.to_index() != Index::INVALID; i++, rhs_it++)
+        {
+            if (m_status[i] == SlotStatus::USED)
+            {
+                // clean original data by it's dtor
+                erase(i);
+            }
+            m_data[i] = std::move(*rhs_it);
```
elBoberido, line 139: This must be changed similarly to the copy assignment.

When looking at the code, it is almost identical the copy assignment, except the line with `m_data[i] = std::move(*rhs_it);` and the `rhs.clear()` at the bottom. Can you think of a way how to reduce the code duplication?

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: If I understand correctly, this would fix the problem:

```c
// assign directly, no need to call erase (all memory located on stack)
m_data[i] = (m_status[i] == SlotStatus::USED) ? (*rhs_it) : new(&m_data[i]) T(*rhs_it);
```

elBoberido: Hahaha, yes. I had to think a bit to realize that it works. Interesting solution but for readability reasons an if-else statement would be better.

Dennis40816: Sure :)

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: See e28c2d7

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: See e28c2d7.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: See https://github.com/eclipse-iceoryx/iceoryx/commit/e28c2d798c936d69fc90e82d37abaf50d6b97017.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: See https://github.com/eclipse-iceoryx/iceoryx/commit/e28c2d798c936d69fc90e82d37abaf50d6b97017.

iceoryx_dust/container/include/iox/fixed_position_container.hpp
```diff
@@ -324,6 +324,9 @@ class FixedPositionContainer final
     IndexType m_size{0};
     IndexType m_begin_free{Index::FIRST};
     IndexType m_begin_used{Index::INVALID};
+
+    template <typename RhsType>
+    void init(RhsType&& rhs) noexcept;
```
elBoberido, line 329: The implementation looks sound. Now it's up to the tests to find out if there are still some corner cases :)

btw, you cannot know this but we usually have the members at the very bottom. Could you please move it up. Something between line 179 and 320. Oh, and can you find a name that better describe what the function does? If you need suggestions feel free to ask :)

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
elBoberido: I think I have on additional test. Could you please add a test where a partially filled container is moved or copied and then add new elements to the moved-to container till it is full. The moved-to container should also not be empty but partially filled and the slot and `Index::LAST` should also be occupied. This should ensure that the free list is not broken after the move/copy.

iceoryx_dust/container/include/iox/fixed_position_container.hpp
Dennis40816: Well, nothing comes to mind. Do you have any suggestions for a name :>

elBoberido: Well, you know that naming is one of the two big problems in computer science. The other ones are cache coherency and off-by-1 errors :) 

How about `copy_and_move_impl`?

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
Dennis40816: You mean something like:
```c
// arrange: add some elements
constexpr DataType offset{32U};
sut_complex.emplace(offset);
sut_complex.emplace(offset+1);

// action:
Sutcomplex copy_or_move_sut_complex;
copy_or_move_sut_complex = sut_complex;

// assert: 
EXPECT_THAT(copy_or_move_sut_complex.full(), Eq(false));
EXPECT_THAT(copy_or_move_sut_complex.empty(), Eq(false));
EXPECT_THAT(copy_or_move_sut_complex.size(), Eq(2U));

// action: fill rest slots
for(IndexType i = 2U; i < CAPACITY; ++i)
{
    copy_or_move_sut_complex.emplace(offset+i);
}

// assert:
EXPECT_THAT(copy_or_move_sut_complex.full(), Eq(true));
EXPECT_THAT(copy_or_move_sut_complex.iter_from_index(SutComplex::Index::LAST)->value,
                        offset + CAPACITY - 1);

```
If something goes wrong, please tell me:)
If everything is fine, I'll append it to the todo list of test cases.

iceoryx_dust/container/include/iox/fixed_position_container.hpp
Dennis40816: Hahaha, that's true. I'll take it:)

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
elBoberido: More like
```cpp
SutComplex  sut2;
fillComplex(sut2); // aaargh, should have called this fill_complex()

sut2.erase(Sut::Index::FIRST);
sut2.erase(Sut::Index::LAST / 2U);

// do something similar with sut_complex; important is only that it is neither empty nor full

sut2 = sut_complex;

// now insert two elements and the third insertion should fail
// sut2 should also be full now and the size should equal the capacity
```
Something similar to above.

It is not necessary to do some asserts after the copy since that scenario is already tested. The new scenario is the filling of the destination container after the copy and only for this the checks need to be done.

Btw, there is the iceoryx developer meetup tomorrow around this time. Your welcome to join if you like :)

Dennis40816: Oh, I got it:)

Thank you for the invitation to the iceoryx developer meetup! I'm very interested in joining, but I'm not sure if I'll be near a computer at that time. Is this a regularly scheduled event?

elBoberido: Yes, it is once a month. Currently the first Thursday at 5pm CET. Theoretically it should show up in the [ROS calendar](https://calendar.google.com/calendar/u/0/embed?src=agf3kajirket8khktupm9go748@group.calendar.google.com) but it seems it is currently removed.

Dennis40816:
I found a weird behavior when using git. I noticed that using `git add -p` followed by ```git commit -s``` doesn't behave as I expect. It changes the content that I've added. For example, I create a file called ```test.txt```.

```bash
# Track test.txt
git add test.txt
git commit -s

# Modify separate parts of test.txt, denoted as A and B
git add -p test.txt      # Use y and n (add A and ignore B)
git commit -s

# Both A and B are included in the last commit (though only A was supposed to be included)
git status
```

So far, I've been using `stash` to avoid the problem. Am I doing anything wrong?

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
Dennis40816: See #[af61c72](https://github.com/eclipse-iceoryx/iceoryx/pull/2069/commits/af61c72e066778b0ab153971299baa25778b6e29).

Dennis40816:
The test ran on the Linux platform (Ubuntu 22.04 LTS via WSL) for both C++14 and C++17.

elBoberido:
@Dennis40816 I don't use `git add -p` myself so I cannot tell if you did something wrong. Usually I always commit the whole file.

Could you please rebase to master and fix the conflict in the files. I try to start reviewing tomorrow.

elBoberido:
With the latest merge it seems there is another merge conflict. Please update again. It also seems that the gcc builds are broken. I guess you need to move the specialization to the .inl file

Dennis40816:
Ok, I'll switch to g++ 8.4.0 and take a took what's going on.

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +55,166 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+
+    constexpr bool is_move = std::is_rvalue_reference<decltype(rhs)>::value;
+
+    IndexType i{Index::FIRST};
+
+    // this can update member faster
+    IndexType next_used_cache{Index::INVALID};
+    IndexType next_free_cache{Index::INVALID};
+    bool overwrite_used_slot{false};
+    bool overwrite_free_slot{true};
```
elBoberido, line 114: Shouldn't this be `false`?

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
```diff
@@ -51,6 +55,166 @@ inline FixedPositionContainer<T, CAPACITY>::~FixedPositionContainer() noexcept
...
+            // update next free cache
+            next_free_cache = static_cast<IndexType>(m_next[i]);
+            overwrite_free_slot = true;
+        }
+
+        m_status[i] = SlotStatus::USED;
+        m_next[i] = static_cast<IndexType>(i + 1U);
+    }
+
+    // update the member only if rhs is not empty to ensure that erase() functions correctly.
```
elBoberido, line 171: I think the following code can be massively simplified if you take inspiration from the `clear` method. The `erase` method is not required since the internal structure is rebuild anyway.

iceoryx_dust/container/include/iox/detail/fixed_position_container_helper.hpp
```diff
@@ -0,0 +1,94 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef IOX_DUST_CONTAINER_DETAIL_FIXED_POSITION_CONTAINER_HELPER_HPP
+#define IOX_DUST_CONTAINER_DETAIL_FIXED_POSITION_CONTAINER_HELPER_HPP
+
+#include <utility>
+
+namespace iox
+{
```
elBoberido, line 23: Please add a `detail` namespace for this helper classes.

elBoberido (commented):
It's too late for the tests today. Something to look forward to for tomorrow

iceoryx_dust/container/include/iox/detail/fixed_position_container.inl
Dennis40816: See 9eff9ce.

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +131,1542 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+    for (const auto& value : EXPECTED_VALUE)
+    {
+        copy_sut_complex.emplace(value);
+    }
+
+    copy_sut_complex = sut_complex;
+
+    EXPECT_THAT(copy_sut_complex.full(), Eq(0));
+    EXPECT_THAT(copy_sut_complex.empty(), Eq(true));
+    EXPECT_THAT(copy_sut_complex.size(), Eq(0));
```
elBoberido, line 574: Please also check that `begin` points to `end` in this test

iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp
```diff
@@ -92,6 +131,1542 @@ TEST_F(FixedPositionContainer_test, Capacity)
...
+    {
+        copy_sut_complex.emplace(value);
+    }
+
+    copy_sut_complex = sut_complex;
+
+    EXPECT_THAT(copy_sut_complex.full(), Eq(false));
+    EXPECT_THAT(copy_sut_complex.empty(), Eq(false));
+    EXPECT_THAT(copy_sut_complex.size(), Eq(EXPECTED_SIZE));
+    auto it = copy_sut_complex.iter_from_index(SutComplex::Index::FIRST);
```
elBoberido, line 600:
```suggestion
    auto it = copy_sut_complex.begin();
```

elBoberido (commented):
Only a few minor things left and from my point of view it could be merged. Thanks for being so patient :)

Oh, could you add one more test, e.g. IteratorsAfterMoveWorkAsExpected
- source container `begin()` == `end()`
- destination container iterates until `end()` <- if I did not overlook something then then it was never checked that `++it` results in `end()`

Dennis40816:
> Only a few minor things left and from my point of view it could be merged. Thanks for being so patient :)
> 
> Oh, could you add one more test, e.g. IteratorsAfterMoveWorkAsExpected
> 
> * source container `begin()` == `end()`
> * destination container iterates until `end()` <- if I did not overlook something then then it was never checked that `++it` results in `end()`

Sure. Thank you so much for your patience and kindness. Your willingness to accommodate and review my mistakes has been incredibly beneficial.

elfenpiff:
@Dennis40816 I reviewed your PR as well, and it looks like a great job! 
As soon as the requested additional test cases are in, I would approve it.

Dennis40816:
@elfenpiff  Thank you so much! I am happy to contribute to the project.
Btw, occasionally the tests fail (as with the recent builds and tests), and the reasons are still unclear.

elfenpiff:
@Dennis40816 can you create an issue with the sporadically failing test names. When we have time we can look into it, except when you are eager to dig deeper. 😀

elBoberido (approved):

elBoberido:
> Btw, occasionally the tests fail (as with the recent builds and tests), and the reasons are still unclear.

I saw only two failures during one CI run. It was the bazel and thread sanitizer and looked like the test file was too large but after triggering the CI again it passed. Did you see more of these? If yes, maybe the file with the tests needs to be split into two.

Dennis40816:
> I saw only two failures during one CI run. It was the bazel and thread sanitizer and looked like the test file was too large but after triggering the CI again it passed. 

You're right. But something weird happened.



In [build-test-ubuntu-with-thread-sanitizer-clang-latest](https://github.com/eclipse-iceoryx/iceoryx/actions/runs/6793755775/job/18478181732#step:4:9820), all constructors were supposed to be invoked, as the destination container was empty, indicating that constructors should be called initially.

```
[ RUN      ] FixedPositionContainer_test.UsingCopyCtorFullCapacityContainerPreservesAllElements
/home/runner/work/iceoryx/iceoryx/iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp:214: Failure
Value of: stats.copyCTor
Expected: is equal to '\n' (10, 0xA)
  Actual: 9 (of type unsigned long)
/home/runner/work/iceoryx/iceoryx/iceoryx_dust/test/moduletests/test_container_fixed_position_container.cpp:216: Failure
Value of: stats.copyAssignment
Expected: is equal to 0
  Actual: 1 (of type unsigned long)
[  FAILED  ] FixedPositionContainer_test.UsingCopyCtorFullCapacityContainerPreservesAllElements (1 ms)
```
And something similar in [build-test-ubuntu-bazel](https://github.com/eclipse-iceoryx/iceoryx/actions/runs/6793755775/job/18478181226#step:4:1730)

elBoberido (dismissed):

elBoberido:
Since this PR is already quite large and `copy_and_move_impl` is already a bit complex I think it would be best to add the initialization of `m_status` in this PR and if you like, you could create a follow up PR to refactor `copy_and_move_impl`.

I like what you did with the `Helper` structs and the `copy_and_move_impl`. Maybe you can come up with a solution which does not need the presence of the assignment operators when the copy/move ctor is used and still share the code.

I haven't fully thought it through but having something like this
```cpp
enum class SpecialMemberOperation {
    CopyConstructor,
    CopyAssignment,
    MoveConstructor,
    MoveAssignment,
};
```
Could be used as template parameter for the `Helper` instead of the boolean and this could then also be used for `copy_and_move_impl`. If this works out, the `Helper` structs could be moved to the design module and a small design document could document how they can be used with a `copy_and_move_impl`.

What do you think?

elBoberido:
Oh, and I'm a bit disappointed that the undefined behavior sanitizer did not find the issue with the uninitialized access. I was hoping that it could find issues like this.

Dennis40816:
> Since this PR is already quite large and `copy_and_move_impl` is already a bit complex I think it would be best to add the initialization of `m_status` in this PR and if you like, you could create a follow up PR to refactor `copy_and_move_impl`.
> 
> I like what you did with the `Helper` structs and the `copy_and_move_impl`. Maybe you can come up with a solution which does not need the presence of the assignment operators when the copy/move ctor is used and still share the code.
> 
> I haven't fully thought it through but having something like this
> 
> ```c++
> enum class SpecialMemberOperation {
>     CopyConstructor,
>     CopyAssignment,
>     MoveConstructor,
>     MoveAssignment,
> };
> ```
> 
> Could be used as template parameter for the `Helper` instead of the boolean and this could then also be used for `copy_and_move_impl`. If this works out, the `Helper` structs could be moved to the design module and a small design document could document how they can be used with a `copy_and_move_impl`.
> 
> What do you think?

Seems Interesting! I'll try to create a PR(should the issue be created first or not?) about refactoring `copy_and_move_impl`. We can discuss about the details in that issue or PR.

Dennis40816:
Do you mean something like below?

```cpp
// transfer src data to destination
    for (; rhs_it.to_index() != Index::INVALID; ++i, ++rhs_it)
    {
        if (is_move && m_status[i] == SlotStatus::USED)
        // ...

       for (; i < CAPACITY; ++i)
    {
        if (is_move && m_status[i] == SlotStatus::USED)
        {
            m_data[i].~T();
        }


```

elBoberido:
> Sounds Interesting! I'll try to create a PR(should the issue be created first or not?) about refactoring `copy_and_move_impl`. We can discuss about the details in that issue or PR.

It's up to you. You can either as additional PR of #2052 and if you are successful create another issue to move the helper classes to the `design` module and write a small design document. Alternatively, you could create the issue for the helper classes and try to make it work (including a small design document) and afterward adjust the `copy_and_move_impl` for this class. Whatever fits your workflow best

elBoberido:
> Do you mean something like below?
> 
> ```c++
> // transfer src data to destination
>     for (; rhs_it.to_index() != Index::INVALID; ++i, ++rhs_it)
>     {
>         if (is_move && m_status[i] == SlotStatus::USED)
>         // ...
> 
>        for (; i < CAPACITY; ++i)
>     {
>         if (is_move && m_status[i] == SlotStatus::USED)
>         {
>             m_data[i].~T();
>         }
> ```

I'm a bit lost :sweat_smile: 
This is code for `copy_and_move_impl`, right? It won't really work since `is_move` wil also be true for the move assignment and then the move assignment operator should be used. The issue exists only for the copy and move ctor.

The more I think about it, the more I come to the conclusion that the code of the default ctor needs to be moved to something like `init` and this should be called at the top of the copy and move ctors. With this change we can merge the PR. Then we have to think if it is possible have some elegant solution for `copy_and_move_impl` without calling `init` beforehand

Dennis40816:
Oh my god. You're right. I don't even understand what I was thinking just now. I think I might need to make these modifications tomorrow. I'm sorry for not thinking it through. I think I will follow your suggestion and use an `init` function (which might do the same thing as the constructor function). Thank you for the reminder.

elBoberido:
> Oh my god. You're right. I don't even understand what I was thinking just now. I think I might need to make these modifications tomorrow. I'm sorry for not thinking it through. I think I will follow your suggestion and use an `init` function (which might do the same thing as the constructor function). Thank you for the reminder.

No problem :)
I guess it is already late at your location. Take your time.

iceoryx_dust/container/include/iox/fixed_position_container.hpp
```diff
@@ -317,6 +317,10 @@ class FixedPositionContainer final
         IndexType m_index;
     };
 
+  private:
+    /// @brief Initialize member variables to prevent undefined or erroneous values.
+    void reset_member() noexcept;
```
elBoberido: Just one last nitpick. Since you already used `initialize` in the comment could you rename `reset_member` to `init_member` or similar?

Dennis40816: Well, I decided to remove the function in the last commit. Currently,  a `for` loop is used to initialize `m_status` in ctor.
I think this can be improved in the following PR (refactoring one).

elBoberido (approved):

elfenpiff (approved):

elBoberido:
@Dennis40816 congratulations to your first merged PR in iceoryx. Thanks for the patience with me and I'm looking forward to more contributions from you :)

Issue #2040, opened 2023-10-05
zero runtime memory allocation application


lucabart97:
## Required information

**Operating system:**
Ubuntu 20.04.6 LTS

**Compiler version:**
g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0

**Eclipse iceoryx version:**
v2.0.3 (Blueberry)

**Heaptrack version.**
heaptrack 1.4.80

## Description
I have an application in C++ that uses Iceoryx that needs zero runtime memory allocation.
Checking the memory usage using heaptrack, there seems to be a periodic task inside Iceoryx that allocates a string to send the node's status. 
There is a way to mitigate it?

### Heaptrack flame graph

![Selection_002](https://github.com/eclipse-iceoryx/iceoryx/assets/55996294/f4c5e1bc-6afa-4118-abc6-c005e1887305)

elfenpiff:
@lucabart97 Yes, you are right. This is some technical debt from the past and is originating here: iceoryx_dust/include/iceoryx_dust/cxx/serialization.hpp

In the end, all occurrences of `std::string` has to be replaced with `iox::string<1024>` (1024 is a lucky guess, but it seems to fit our use case). The easiest approach would be to introduce a type alias in the `Serialization` class called `using SerializedString = iox::string<1024>;` and do a search and replace.

@lucabart97 since you unraveled the issue, would you like to have the chance to fix it? We would guide you through the pull request process, and we would be happy to have a new iceoryx contributer :smile:

lucabart97:
@elfenpiff Thank you for your quick response. I'll try to fix the ""issue"" myself, and I'll make a pull request once I'm done.

elfenpiff:
> @elfenpiff Thank you for your quick response. I'll try to fix the ""issue"" myself, and I'll make a pull request once I'm done.

@lucabart97 Awesome! Please add `iox-#2020 ` as prefix for every commit (required for traceability) and sign the eclipse ECA (https://www.eclipse.org/legal/ECA.php) then this could be "easy" - if I forgot something, and it turns out to be hard, do not hesitate to ask!

elBoberido:
@lucabart97 I think it won't be that easy to remove the last `std::string` from the code base. We need some bigger refactoring for the serialization and Unix Domain Socket abstraction to get rid of `std::string`.

[...]

Issue #1431, opened 2022-06-29
Convenience functions for `cxx::string`


FerdinandSpitzschnueffler:
## Brief feature description

This is an issue to collect convenience functions for the `cxx::string` which would be nice to have but are not urgently needed.

## Detailed information

- add c'tor that constructs the `iox::string` with count copies of a character
    ``` 
    iox::string<10> str(TruncateToCapacity, count, 'M');    // count as runtime parameter
    iox::string<10> str(count, 'M');    // count as compile time constant
    ```
- remove characters from the `iox::string`
   ```
   iox::string<10> str("someString");
   str.erase(5, 2);    // str == someSing
   str.erase(7);    // str == someSin
   ```

PR #2093, opened 2023-11-14
iox-#2040: add unsafe_raw_access to iox::string

lucabart97:
Introduce a function to directly access the string pointer. This function enables checking the string's size and capacity after each call, minimizing the risk of encountering invalid strings.

1. [ ] Add a second reviewer for complex new features or larger refactorings
1. [x] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Refer https://github.com/eclipse-iceoryx/iceoryx/issues/2040


iceoryx_hoofs/vocabulary/include/iox/detail/string.inl
```diff
@@ -238,6 +238,22 @@ inline bool string<Capacity>::unsafe_assign(const char* const str) noexcept
     return true;
 }
 
+template <uint64_t Capacity>
+inline void
+string<Capacity>::unsafe_raw_access(const iox::function<void(char*, const uint64_t, const uint64_t)>& func) noexcept
+{
+    func(m_rawstring, m_rawstringSize, Capacity);
```
elBoberido, line 245: What do you think about setting `m_rawstrin[Capacity]` to `'\0'` before calling `func`?

iceoryx_hoofs/vocabulary/include/iox/detail/string.inl
```diff
@@ -238,6 +238,22 @@ inline bool string<Capacity>::unsafe_assign(const char* const str) noexcept
...
+
+    const uint64_t strSize{strnlen(m_rawstring, Capacity + 1U)};
+    if (Capacity < strSize)
+    {
+        IOX_LOG(FATAL, "unsafe_raw_access failed. Data wrote outside the maximun string capacity of " << Capacity);
+        return;
+    }
```
elBoberido, line 252: Please use `IOX_EXPECTS_WITH_MSG` in order to terminate if there was a buffer overflow.

I'm also a bit torn on the automatic size calculation. On one hand it is more safe for simple cases but if someone does a memcpy into `m_rawstring` and the raw data contains zeros, then then strnlen will not determine the correct size and also not a potential buffer overflow. One could argue that it is a misuse if binary data is copied into a string and something different needs to be used, e.g. a `vector`. Here I would totally agree.

Another reason I'm torn is consistency with a potentially similar API for the `vector`. Here we could not automatically detect the new size and the user would need to return it from the lambda. The question is now whether the `string` API should work similarly and return the new size. What's your thoughts on this question?

iceoryx_hoofs/vocabulary/include/iox/string.hpp
```diff
@@ -423,6 +424,13 @@ class string final
     template <typename T>
     IsStringOrCharArrayOrChar<T, bool> unsafe_append(const T& str) noexcept;
 
+    /// @brief direct access to the string's raw pointer. The access resizes the data correctly after the function call.
+    /// If the data written in the function exceeds the capacity, and the code does not segfault beforehand due to an
+    /// invalid access, a FATAL error occurs.
+    ///
+    /// @param [in] function composed by the raw data, current string size and the capacity
```
elBoberido, line 431: nitpick
```suggestion
    /// @param [in] func is a function composed by the raw data pointer, current string size and the capacity
```

Could you also add a small example how this is used, nothing fancy just something like this
```cpp
iox::string<100> s;
s.unsafe_raw_access([] (auto* str, const uint64_t size, const uint64_t capacity) {
    strncpy(str, "Hello World", capacity);
});
```

Maybe also mention that is is safe to write the null-termination at `str[capacity]`.

iceoryx_hoofs/vocabulary/include/iox/string.hpp
```diff
@@ -423,6 +424,13 @@ class string final
...
+    void unsafe_raw_access(const iox::function<void(char*, const uint64_t, const uint64_t)>& func) noexcept;
```
elBoberido, line 432: Please use `iox::function_ref` instead of `ioc::function` when the function is not stored.

Oh, and could you define an alias fur the function and use that in `string.hpp` and here.

elBoberido (commented):
For traceability reasons it might be better to have this PR to be part of #1431 which is more related to strings than #2040. Maybe we can make #2040 kind of a meta issue and have more specific issues for the tasks which need to be done. There will be a few preparatory PRs like this until everything is in place to finally get rid of the runtime allocations and not relying on small object optimizations to not have them.

lucabart97:
> For traceability reasons it might be better to have this PR to be part of https://github.com/eclipse-iceoryx/iceoryx/issues/1431 which is more related to strings than https://github.com/eclipse-iceoryx/iceoryx/issues/2040. Maybe we can make https://github.com/eclipse-iceoryx/iceoryx/issues/2040 kind of a meta issue and have more specific issues for the tasks which need to be done. There will be a few preparatory PRs like this until everything is in place to finally get rid of the runtime allocations and not relying on small object optimizations to not have them.

@elBoberido if you prefer I can force-push the commit, but the branch style name will be the same, otherwise, I can close this PR and reopen another with the correct branch identification (using 1431 instead of 2040).

Yes, maybe is better to create a sort of tasklist for the 2040

iceoryx_hoofs/vocabulary/include/iox/detail/string.inl
lucabart97: I do not see any case where having the termination at the capacity size will help to not have troubles.

iceoryx_hoofs/vocabulary/include/iox/detail/string.inl
lucabart97: You are using a ```char``` string, so if you are using binary data you should use an ```uint8_t``` vector. I like the proposal to return the auto-calculated size, in this way, you could also check if the calculated value is correct.
```

Comment (elBoberido) [reply iceoryx_hoofs/vocabulary/include/iox/detail/string.inl:252]: It would be even more performant since one only needs to check whether there is a `\0` at the returned position. The user might already know the size and does not need to call `strnlen`.

**Comment by elBoberido** [comment]
> @elBoberido if you prefer I can force-push the commit, but the branch style name will be the same, otherwise, I can close this PR and reopen another with the correct branch identification (using 1431 instead of 2040).
> 
> Yes, maybe is better to create a sort of tasklist for the 2040

I would be fine with a force push and keeping the branch name with 2040 for this PR. The most important part for traceability is what git blame shows on the respective line and to easily find the corresponding issue where the line was changed. The branch name is not that important and we must not be more catholic than the pope :)

**Comment by elBoberido** [comment]
Btw, the clang-tidy check fails. You can run it locally with `tools/scripts/clang_tidy_check.sh scan_list .clang-tidy-diff-scans.txt "path/to/file1.cpp path/to/file2.cpp ..."`

**Comment by lucabart97** [comment]
> Btw, the clang-tidy check fails. You can run it locally with tools/scripts/clang_tidy_check.sh scan_list .clang-tidy-diff-scans.txt "path/to/file1.cpp path/to/file2.cpp ..."

I see, the ```strcpy``` should be fine in tests, so I'll add ```//NOLINT..``` if you agree. Instead, about the other "error", what do you suggest?

```
adjacent parameters of 'operator()' of similar type ('const uint64_t') are easily swapped by mistake
[bugprone-easily-swappable-parameters,-warnings-as-errors]
    this->testSubject.unsafe_raw_access([this](char* str, const uint64_t size, const uint64_t capacity) {
                                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

**Comment by elBoberido** [comment]
> I see, the `strcpy` should be fine in tests, so I'll add `//NOLINT..` if you agree. Instead, about the other "error", what do you suggest?
> 
> ```
> adjacent parameters of 'operator()' of similar type ('const uint64_t') are easily swapped by mistake
> [bugprone-easily-swappable-parameters,-warnings-as-errors]
>     this->testSubject.unsafe_raw_access([this](char* str, const uint64_t size, const uint64_t capacity) {
>                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

`NOLINT` wouldn't be a problem in this case. I think using `strlcpy` is not available on windows so that would require additional work in the platform layer. If you are interested you could try add it but `NOLINT` is fine for the tests.

I might end up making `strlcpy` available on all platforms later on anyway since we are using `strncpy` in some places where `strlcpy` would be more safe.

For the other issue, in theory we decided deactivated the rule for the time being but actually forgot to actually do it. I think the reason we forgot about it was that it did not show up in the production files we scan. If `auto* str, const auto size, const auto capacity` is used it also does not show up. This would be a workaround but I tend to think of a solution with a struct with two member. We could then re-use that struct also for the complementary function in `vector` later on. chatGPT suggest `BufferInfo` as name :)

It could look like this and could live in `iceoryx_hoofs/buffer/`
```cpp
struct BufferInfo {
    uint64_t used_size{0};
    uint64_t total_size{0};
};
```
It would be a bit different than what we have now since it supplies the whole buffer size, including the space for the null-termination, but this would make it easier to work with functions like `strlcpy` or `strlcat` which automatically adds the null-termination.

What do you think?

lucabart97:
@elBoberido 

If possible, I suggest changing the clang-tidy rules, almost to set at 3 the ```bugprone-easily-swappable-parameters```. 

I think this workaround increases the code complexity because when I see the method prototype, I can't understand all parameters, I need to find and read the ```BufferInfo``` struct.

elBoberido:
@lucabart97 if it is only for convenience I would take the more safe approach. Our long term plan is to use specific types as much as possible to elevate small oversights to compile time failures.

With autocompletion from the clang LSP it also just shows `const iox::function<void(char*, const uint64_t, const uint64_t)>& func`. A `BufferInfo` instead of the two `uint64_t` would give the developer some hints on what to expect.

The reason you understand the current signature is because you just wrote it and it is fresh in your mind but these things get blurry and in a few months it is not anymore as clear as now. User which are new to the API have to look at the documentation anyway or their IDE will tell them the meaning of the members of the struct.

Btw, have you already seen comments like this https://www.reddit.com/media?url=https%3A%2F%2Fi.redd.it%2Fhwqj7yx9vm211.jpg? Only slightly related but when writing about fading memories I remembered that I wrote a similar comment to something which was fully clear when I wrote the code :)

lucabart97:
@elBoberido got it

I add `BufferInfo` and two string access functions:
 - `unsafe_raw_access`
 - `unsafe_auto_raw_access`

Tell me your feedback :)

iceoryx_hoofs/vocabulary/include/iox/detail/string.inl
```diff
@@ -238,6 +238,35 @@ inline bool string<Capacity>::unsafe_assign(const char* const str) noexcept
...
+                         "unsafe_auto_raw_access failed. Data wrote outside the maximun string capacity of "
+                             << Capacity);
+    m_rawstringSize = strSize;
+}
+
+template <uint64_t Capacity>
+inline void
+string<Capacity>::unsafe_raw_access(const iox::function_ref<uint64_t(char*, const iox::BufferInfo info)>& func) noexcept
+{
+    iox::BufferInfo info{m_rawstringSize, Capacity};
```
elBoberido, line 259: Please use `Capacity + 1`. This is a low level API and the user might need to write a zero termination at the end of the buffer which is larger than the `Capacity` by one element.

iceoryx_hoofs/vocabulary/include/iox/string.hpp
```diff
@@ -423,6 +425,36 @@ class string final
...
+    ///         strncpy(str, "Hello World", info.total_size);
+    ///     });
+    /// @endcode
+    void unsafe_auto_raw_access(const iox::function_ref<void(char*, const iox::BufferInfo info)>& func) noexcept;
+
+    /// @brief direct access to the string's raw pointer. The access resizes the data with the value returned by the
+    /// passed function. If the data written has not the terminator at the returned size, a FATAL error occurs.
+    ///
+    /// @param [in] function func is a function composed by the raw data pointer and the BufferInfo with current size
+    /// and capacity. The return value is the string lenght.
```
elBoberido, line 447:
```suggestion
    /// and total size, including the space for the zero termination. The return value is the string length.
```

elBoberido (commented):
Almost there. This can soon be merged :)

elBoberido:
Oh, and there is another formalism. If there wouldn't be other changes then it could be left as is but since you have to commit anyway could you capitalize the commit message to `iox-#1431 Add ...`?

lucabart97:
@elBoberido 
Sorry for the copyright, I just copy-pasted from another file without reading it 😅

I got your comments and I have updated the code/commit message!

iceoryx_hoofs/test/moduletests/test_vocabulary_string.cpp
```diff
@@ -643,6 +644,53 @@ TYPED_TEST(stringTyped_test, UnsafeAssignOfNullptrFails)
...
+    testCharstring.emplace_back('\0');
+    this->testSubject.unsafe_raw_access([&testCharstring](char* str, auto) -> uint64_t {
+        //NOLINTNEXTLINE(clang-analyzer-security.insecureAPI.strcpy,-warnings-as-errors)
+        strcpy(str, testCharstring.data());
+        return STRINGCAP;
+    });
+    EXPECT_THAT(this->testSubject.unsafe_assign(testCharstring.data()), Eq(true));
+    EXPECT_THAT(this->testSubject.size(), Eq(STRINGCAP));
+}
+
```
elBoberido, line 693: Can you add tests for wrong usage
- return info.total_size + 1
- no `\0` at returned position

You can use `IOX_EXPECT_FATAL_FAILURE` for these tests. There are some in this file if you need an example.

elBoberido (commented):
Almost there. Just two additional tests.

iceoryx_hoofs/test/moduletests/test_vocabulary_string.cpp
lucabart97: done!

elBoberido (approved):
Great. Thanks for your patience :)

Once the CI has run this can be merged

elBoberido:
Oh, just notice that there is an issue on the CI. It seems the new tests have an unused `this` capture in the lambda

lucabart97:
@elBoberido  
fixed!

elBoberido:
Oh no, now Windows complains. You can easily find the error by searching for `error C`

lucabart97:
@elBoberido 
I understood the error with ```STRINGCAP``` that was not captured by lambda, but you understand the others errors?

elBoberido:
@lucabart97 I'm not quite sure. If it isn't an aftereffect on the first error it might be a missing `const` on the second lambda parameter  on line 685

elBoberido:
Or maybe there is a weird type conversion for `STRINGCAP` happening?

lucabart97:
Ok, on mac-os the capture of ```STRINGCAP```gives us an error. Let's use directly the ```&```.

Let's see if Windows passes or not, maybe can be a weird type conversion for ```STRINGCAP```

elBoberido:
Welcome to the hell known as Windows support :sweat_smile: 

It seems clang based toolchains consider the capturing of `STRINGCAP` as superfluous. You can just capture all by reference, i.e. a singe `&`. Alternatively `strlen` or the data supplied by buffer info should also work.

elBoberido (approved):

lucabart97:
I hope it will be fine!

(Windows 👎)

elBoberido:
I've a good feeling now :)

Issue #2128, opened 2023-12-07
Map segments by name instead of access permissions, and allow publishers write access to multiple


gpalmer-latai:
## Brief feature description

Shared memory segments are currently mapped strictly by access controls. It would be beneficial to map instead by name or some other key so that processes aren't limited to writing to a single segment and we could better split up communication channels across multiple segments.

## Detailed information

Currently, it is possible to creating multiple shared memory segments with different access permissions, as described in the [configuration guide](https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/advanced/configuration-guide.md).

However, not only do the segments allow one to specify access controls, but they create really strong couplings based on those access controls. The implications include:

- The name of shared memory segments is automatically deduced as the name of the cgroup that has write access to it
- The above means that there can only be one segment that any given cgroup has write access to
- Publishers are allowed write access to only one segment, which is deduced as the segment that matches the cgroup of the process the publisher is contained in.
- It is therefore not possible to have a single publisher access multiple segments (This makes sense - it would be strange to do so).
- It is also not possible to have multiple publishers in the same process which access different segments

The last point is of particular interest to me because I can consider a couple notable reasons why one would want to be able to create many shared memory segments and be able to access them from within the same process:

- Splitting communication channels across separate shared memory segments ensures better [Freedom From Interference](https://heicon-ulm.de/en/iso26262-freedom-from-interference-what-is-that/) because should a bad publisher or subscriber access data past the end of a memory pool, it will usually result in a segfault rather than simply a corrupted read/write of another memory pool. 
- We may wish to make use of the [MemoryInfo member of a segment config](https://github.com/eclipse-iceoryx/iceoryx/blob/6303f2ea7ec18381943e1f3edb150b7dae908fe1/iceoryx_posh/include/iceoryx_posh/mepoo/segment_config.hpp#L37) (currently just defaulted) to differentiate segments based on memory type. This would allow a simpler, cleaner solution to the issue described in #2092 because we could specify that a segment is intended to support `CUDA_PINNED_HOST` memory, and then each publisher / subscriber that needs access to such memory would identify that segment and pin the memory of only that segment. 

**Possible Obstacles** 

I've found a [curious comment in the code](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/mepoo/segment_manager.inl#L66): 

> a user is allowed to be only in one writer group, as we currently only support one memory manager per process

[...]

PR #2140, opened 2023-12-20
iox-#2128 Draft design for named segments

gpalmer-latai:

1. [ ] Add a second reviewer for complex new features or larger refactorings
1. [ ] Branch follows the naming format (`iox-123-this-is-a-branch`)
1. [ ] Update the PR title
   - Link to the relevant issue
1. [ ] Assign PR to reviewer

## References

- Closes **TBD**


doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+* When no segment is specified, the fallback behavior must be the same as it is currently
+  * Publishers will select the *unique* segment they have write access to.
+  * Subscribers will receive messages from all segments they have read access to.
+  * Clients and Servers follow the same rules as Publishers for data they send and the same rules as Subscribers for data they receive.
+* Creation will fail if a communication endpoint requests to use a segment to which it does not
+have the proper access rights.
+
+### Optional additional requirements
+
+* When initializing the runtime, a list of segment names may be provided. In this case, rather than mapping all shared memory segments available in the system, only those that are named shall be mapped.
```
elBoberido, line 60: When we create a new API, the user should explicitely tell what the behavior should be by either passing a list of segments or using a flag which indicates to get all segments with read or write access. The current API will keep its behavior but would be deprecated at the long term.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Design
+
+### Updated RouDi Config
+
+Supporting named segments will require add an additional `name` field to the RouDi config under the `[[segment]]` heading, as well as updating the config version Example:
+
+```
+[general]
+version = 2
```
elBoberido, line 71: Thinking out loud. I'm wondering if the version increment is necessary but since an old RouDi potentially errors out due to the possibility of two segments with the same writer group this ends up being a semantic change and the increment is required.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+
+[[segment]]
+name = "foo"
+
+[[segment.mempool]]
+size = 32
+count = 10000
+```
+
+Updating the version will be necessary for older versions of RouDi to fail informatively when presented with a newer config. Otherwise RouDi can support version 1 configs as if they are version 2 configs with no `name` fields specified.
```
elBoberido, line 81: Hahaha. Could have saved me some time by just reading this sentence before writing the previous comment 😅

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+
+#### When creating a Subscriber
+
+A similar field will be added to the [SubscriberOptions struct](../../../iceoryx_posh/include/iceoryx_posh/popo/subscriber_options.hpp), except that it will support multiple elements
+
+```
+struct SubscriberOptions
+{
+  ...
+  vector<ShmName_t, MAX_SUBSCRIBER_SEGMENTS> segmentNames{};
```
elBoberido, line 104: Here we have again the issue with the fixed size container. We need to find a sane limit without causing the data structures exploding in their memory requirements.

elBoberido (commented):
Half way through it

doc/design/draft/named-segments.md
gpalmer-latai: Yes I agree. The main benefit is that older versions error out on the version number rather than syntax, which makes it more obvious what the problem is.

doc/design/draft/named-segments.md
gpalmer-latai: Yes... 🙁 

In my own experimentation I actually just made it a single string. I think in practice that is usually what would be used since the main benefit of this feature is communicating over specific isolated "channels". But I was trying to leave a bit flexibility for some hypothetical use cases like having multiple publishers publishing on different segments, but the same topic. 

Another idea to consider which I do not recall whether I've mentioned before, but we could instead of introducing a new string field, add some coupling to the service description. For example - we could say that the "group" field is used to identify the shared memory segment to use, and all topics belonging to the same group share the same segment. 

I'm very hesitant/skeptical of this approach because I prefer not to couple features to one another. This would also more likely be a breaking API change unless we were careful to add some flags like "useServiceGroupAsSegmentMapping" or whatever.

gpalmer-latai: I guess you could have a set of flags / bitmap that with mutual exclusivity allows to choose from the following:
```
noSegmentMapping
groupAsSegmentMapping
instanceAsSegmentMapping
topicAsSegmentMapping
```
(topic is fully qualified on the group and instance as well)

gpalmer-latai: I think, in terms of the FFI implications here. In a system where multiple producers for a single topic is allowed, you would still want there to be a way for each producer to be isolated to writing to different shared memory segments and as such the "match by service description" approach would be insufficient.

gpalmer-latai:
@elBoberido I'm back from vacation whenever you want to continue discussing this. If you'd like, we can add an agenda item to one of the next two developer meetups. Though I understand the focus may be on introducing Iceoryx2.

doc/design/draft/named-segments.md
elBoberido: I'm wondering whether it is necessary to have the segments in the `SubscriberOptions` at all. An alternative would be to leave it to the runtime, e.g.
```cpp
for(const auto* name: {"foo", "bar"}) {
    runtime.mapReadOnlySegment(name);
}
```
The same could be done for the publisher. Since the whole process has read/write access to these segments and there might be multiple subscriber/publisher which access the same segments this could be done upfront at the start of the application.

Regarding the `ServiceDescription`. There is the idea to combine the three string into one. See also https://github.com/eclipse-iceoryx/iceoryx/issues/1052. The three separate strings date back to a time where iceoryx was meant to be the base for an AUTOSAR Adaptiv middleware and some design decisions where made with that idea in mind. With iceoryx2 there is also just one string and with a new API we could also move into this direction. The old `ServiceDescription` would the just be concatenated to e.g. `/service/event/instance`.

For iceoryx2 we currently have a separate shm segment for each publisher. Since this can become an issue with many publisher, there is the idea to group some together and create a `SharedMemoryDirectory` which can be used by multiple publisher. I think this goes into the same direction of what you describe as "group". I need to check with @elfenpiff how exactly this is implemented with iceoryx2.

I think we can hit two birds with this refactoring and create a API to ease the transition from iceoryx1 to iceoryx2.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+Currently initialization of the runtime happens in the constructor, and any errors that occur result in a contract violation and program termination. 
+
+While it is not directly in scope of this design, it would be beneficial to refactor the Runtime implementations to use the builder pattern such that any errors may be handled through returning errors types.
+
+Tying this together with requesting segments, this could look something like:
+```
+class DefaultRuntimeBuilder
+{
+  ...
+  IOX_BUILDER_PARAMETER(vector<ShmName_t>, shmSegmentNames, {});
```
elBoberido, line 144: I think this can be done by subsequent calls. With this we don't need to find a suitable size for the vector.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+  IOX_BUILDER_PARAMETER(bool, allowUnmappedSegments, false);
```
elBoberido, line 145: Although the boolean would also work in this case, in general I prefer enums since they are more future proof and have more than two options. It could be something like
```cpp
enum class UnmappedSegmentBehavior {
    Ignore,
    LoadOnDemand,
    TerminateProcess,
    ReleaseSampleAndReturnError
};
```

`UnmappedSegmentBehavior::Ignore` could be the default and allows to create a subscriber and afterwards calling `runtime.mapReadOnlySegments`. Even if nothing is called, it might be interesting when the application is not interested in the data itself but might only forward the sample.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,228 @@
...
+  ...
+  public:
+    expected<DefaultRuntime, RuntimeBuilderError> create() noexcept;
+```
+
+The `create` method of this builder would then return a custom error if a 
+segment was requested that either does not exist or the current process does not have access to.
```
elBoberido, line 152: I would decouple the creation of the runtime from the mapping of the segments. This makes it easier to continue execution and react on unavailable segments than just aborting the creation of the runtime.

doc/design/draft/named-segments.md
gpalmer-latai: This seems reasonable but it would be a somewhat breaking API change as users would need to include the flag, or explicitly pass an empty list that they are not currently passing. 

This is in contrast to simply not providing a list, which can be interpreted as the deprecated behavior. I'm a bit ambivalent here. I'm open to a small initial breaking change followed by dropping support for the old behavior. Or we could simply start by printing a warning when no list/flag is provided.

doc/design/draft/named-segments.md
gpalmer-latai: Yeah so let me think out loud a little here.

- If we do not specify which segments we expect the subscriber to receive messages from, it could receive them from any of the mapped segments.
- If we then connect to a channel whose segment has not been mapped, we need to either map it dynamically, or ensure some sort of error is triggered. This follows from the other comment you make about having an enum. 
- We will lose the ability somewhat to control who we may allow to talk to a subscriber. Though I suppose this can be solved through other means of controlling which channels use which segments (e.g. segment names are coupled to channels through some configuration scheme so you know ahead of time anyway that channel data can only come across a specific segment). 

I think dropping the explicit segment list from the subscriber works for my purposes and is simpler. The other somewhat ambiguous concerns I'm surmising here can be readdressed if necessary.

gpalmer-latai: The other tricky thing I've encountered is the fact that a subscriber may succceed in "connecting" with a publisher, but since the publisher publishes messages to a segment the subscriber has no access to, it doesn't actually receive any messages, which can be a really surprising silent error.

gpalmer-latai: For example, if we added a list of segment ids to the subscriber port data, and a single segment id to the publisher port data, we could use this additional information in [isCompatiblePubSub](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/source/roudi/port_manager.cpp#L521) to determine if a publishers messages can actually reach a subscriber.

And we can then treat this error the same way we treat policy mismatches.

doc/design/draft/named-segments.md
gpalmer-latai: Good point. An enum would work fine. Note that this value only applies to create of Publishers/Clients/Servers when they request a segment. I haven't considered the possibility of detecting we have a sample from an unregistered segment and suspect it would be non-trivial to add the ability in this scenario to map the segment on demand (the smart chunk data type almost certainly doesn't have access to the segment manager).

doc/design/draft/named-segments.md
gpalmer-latai: How about adding a flag indicating whether none, all read-only, all write, or all read and write segments should be mapped upon creation? And then additionally the ability to map specific ones? 

We can of course also expose those as separate functions as well.

gpalmer-latai:
I've made some changes to the design to remove the subscriber segment requesting feature and change how segments are requested to be mapped in the runtime. I'll go through again on Monday and see if I can clean up the design again some more.

elBoberido:
> I've made some changes to the design to remove the subscriber segment requesting feature and change how segments are requested to be mapped in the runtime. I'll go through again on Monday and see if I can clean up the design again some more.

I don't have time to look at it today and tomorrow might also be difficult. On Wednesday it shouldn't be a problem.

doc/design/draft/named-segments.md
elBoberido: I think I  need some pseudo code for this :sweat_smile:

It also feels a bit complicated with the implicit behavior when something is omitted.

doc/design/draft/named-segments.md
elBoberido: Also just thinking out loud. The question is whether this check should be done in RouDi or the subscriber. A subscriber might receive a sample from a segment which is not mapped but the process might have read access. The subscriber process would then print an error message if it could not map the segment. This would be more visible to the user than a warning from RouDi. We already had some bug reports because of RouDi printing an error message and the user just looking at the terminal with the application.

doc/design/draft/named-segments.md
elBoberido: I think we need to add some hooks to the building blocks the be able to gracefully handle this situation. We should be able to get the information from the pointer repository and then inform someone that we cannot read the payload. This should be possible without reading the chunk header. Just dropping the sample and printing a warning would also be an option. After all, not being able to read the payload is most probably just a misconfiguration.

doc/design/draft/named-segments.md
elBoberido: I think that might work.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+  2. If write access segments are requested, map every segment the user has write access to
+  3. If read and write access segmeents are requested, map every segment the user has either read or write access to
+  4. If no segments are requested, do nothing
+2. If `segmentFilter` is not empty
+    1. Iterate over each name in the filter
+    2. If there is a segment that matches the name in the filter
+        1. If the segment has already been mapped, return an error indicating that a segment has been requested twice. The error should contain access permissions and the value of the `premappedSegments` option the user understands why it has already been mapped.
+        2. Otherwise, if the process has access rights to that segment, map it
+        2. If the process does not have access rights, return an error indicating that a segment was requested the runtime does not have access to.
+    3. If there is no segment that matches the name in the filter, return an error indicating that there is no segment matching the one requested to be mapped.
```
elBoberido, line 205: Like already mentioned above. I have the feeling this becomes confusing due to the parameter influencing each other too much.

It might also break applications by just changing the segment config in a compatible way. Let's assume one want's to map all segments with write access with the `premappedSegments` options and one segment with read access from `segmentFilter`, let's call the segment `alice`. Now the config is changed so that the process has also write access to `alice` and suddenly the application does run anymore although it is not a breaking change in the configuration.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Development Roadmap
+
+- [ ] Extend the RouDi config to support specifying segment names to be added to the `SegmentConfig`
+- [ ] Add the name to the `MePooSegment` data structure and allow multiple write-access segments to be mapped.
```
elBoberido, line 210: On this we already agree :)

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+- [ ] Update producer options structs to include segment names
+- [ ] Update producer segment selection logic to take the segment name into account
+
+### Optional
+
+- [ ] Refactor runtime initialization to use builder pattern
+- [ ] Add segment filter and apply it to segment mapping during runtime initialization. 
+- [ ] Add flag to specify whether endpoints requesting non-mapped segments should fail or whether segments should be created dynamically
```
elBoberido, line 218: I think here some further discussion are needed :)

doc/design/draft/named-segments.md
gpalmer-latai: So I'll need to take another look but IIRC one issue I ran into was that the subscriber didn't have direct access to that sort of information. The Runtime is what has access to a lot of the information and the only interaction Subscribers have with it is in the beginning via `getMiddlewareSubscriber()` to get the subscriber port. 

The subscriber could certainly tell that the segment is not mapped, but the only way to get to the information identifying the segment to determine if the subscriber in fact has read access is via the segment manager, which the runtime currently only temporarily gains access to during construction (the identifying information is collected via the IPC connection, but the runtime doesn't hold onto it).

I must confess that in following your suggestion to subclass things, I managed to subclass the runtime and, after moving the [ipcChannelInterface](https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/runtime/posh_runtime_impl.hpp#L114) from private to protected, have now stored the relative pointer to the segment manager as a member of my derived class. 

With this, you CAN easily ask questions like "Do I have read access to this segment?" And then decide "well, lets go map it now then!". But to do this - the subscriber needs to talk to the runtime. I suppose this is alright - it can always invoke the runtime via `getInstance()`. If we ever get to refactoring the runtime away from being a singleton we would instead need to reference it as a member of the subscriber, and maybe do some attorney pattern stuff for the API's touching the segment manager that only the subscriber should be using in only a specific way. 

As for the runtime - I'd recommend taking a look at this `shared_memory_user` and considering what we need to change to expose the info we need, without exposing too much. It seems to try to only temporarily access the segment manager to just get the info it needs and map everything, and then relinquish it back. But if we want to be able to map segments later - we are going to need to access the segment manager again anyway so... either the runtime should just keep a hold of the segment manager or the `shared_memory_user` API also needs to be extended. I guess there is no reason that class can't privately reference the segment manager past the life of the constructor 🤔

gpalmer-latai: Oh another thing I wanted to mention. Having access to the segment manager seems to make accessing introspection information so much easier than setting up a whole other introspection client. Haha.

doc/design/draft/named-segments.md
gpalmer-latai: Yeah I agree. `take` should probably return an informative error, at the very least. 

Looking into the code, it is unclear to me so far where exactly things will fail. At a high level I know that the subscriber queue contains offset pointers. I don't know exactly if/when that pointer might be dereferenced and crash because the segment is not mapped. But presumably this happens somewhere during the `take()` and can be guarded against, propagating an error out. 

Then the question is whether we want to just return the error to the user, or whether we want to allow some sort of hook in the subscriber to ask the runtime to map that segment.

gpalmer-latai: So... I THINK that all the relative pointer stuff and pointer repository stuff will just return nullptrs when referencing an unmapped segment id. Reasoning about taking and releasing chunks in the subscriber in this sense is okay. It can increment and decrement the reference in the management segment no problem, without actually having access to the chunk data itself. 

The first potential problem I see is at this line of code: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/include/iceoryx_posh/internal/popo/building_blocks/chunk_queue_popper.inl#L57

It looks like the subscriber gets a chunk, doesn't actually check if the chunk is a nullptr, and then plows into dereferencing the nullptr with:
```
        auto receivedChunkHeaderVersion = chunk.getChunkHeader()->chunkHeaderVersion();
```

I'd have to do some testing to confirm my suspicion. This is already a bug unto its own right. Outside of this design we should have better error handling for this scenario.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+* When the user attempts to create a producer that requests a segment which is not mapped, this will trigger the error handler and cause program termination.
+  * In the future if we expose a builder-pattern method of creating producers, we could instead have the builder return an error indicating the segment is not mapped.
+  * An alternative to the above is to add an additional enum with configuration options allowing for the runtime to map requested segments on the fly. This increases flexibility, but because it is already possible to tell the runtime to map a specific segment, this would not add much value. A user who wishes to ensure a producer's segment is already mapped may simply call `mapSegmentForWriting(name)`, ignoring the error indicating that the segment has already been mapped.
+* When a consumer is created, an error will only occur if **no** segment has been mapped. Otherwise if a producer on the same channel as the consumer uses a segment the consumer does not have access to (whether because the segment is unmapped or because the consumer does not have read access), this will result in a fatal error.
```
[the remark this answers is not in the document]
gpalmer-latai: Yeah I'll see if I can rephrase but what I mean is:

When you create a subscriber, you don't bother checking whether that subscriber will have appropriate access to what publishers publish. You just hope it does, and, as discussed in other threads, either capture an error or do some dynamic mapping on the fly of that segment (still checking for the error case that the subscriber just doesn't have read access).

The check I'm talking about here is just whether the subscriber has access to any segment at all, in the case where no dynamic mapping is allowed. This would mean that any `take()` will always fail and it would make sense to report this error early. 

For publishers on the other hand, you can more specifically check whether they have access to the segment they specify.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+count = 10000
+```
+
+Updating the version will be necessary for older versions of RouDi to fail informatively when presented with a newer config. Otherwise RouDi can support version 1 configs as if they are version 2 configs with no `name` fields specified.
+
+### Requesting Segments
+
+#### When creating a Publisher
+
+A new field will be added to the [PublisherOptions struct](../../../iceoryx_posh/include/iceoryx_posh/popo/publisher_options.hpp) as follows:
```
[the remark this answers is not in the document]
gpalmer-latai: Well, one advantage to making it a default-able optional is that we might be able to get away with not having to change any code/break any downstream code. I did have to refactor a bunch of tests that initialize the full options struct in a brace initializer list. 

It might have been less annoying if the options structs were instead builders for the publishers because then the way they would be used would be only explicitly changing the fields folks don't want to remain the defaults.

I actually think this might be the way to go in general - have publishers and subscribers created by builders. This way you can capture the miriad of errors that can occur instead of the current behavior which I believe is "error handler and crash".

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+
+While it is not directly in scope of this design, it would be beneficial to refactor the Runtime implementations to use the builder pattern such that any errors may be handled through returning errors types.
+
+Tying this together with requesting segments, this could look something like:
+```
+class DefaultRuntimeBuilder
+{
+  ...
+  IOX_BUILDER_PARAMETER(MappedSegments, premappedSegments, MappedSegments::ReadAndWrite)
+  IOX_BUILDER_PARAMETER(vector<ShmName_t, MAX_SHM_SEGMENTS>, shmSegmentNames, {});
```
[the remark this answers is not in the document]
gpalmer-latai: Yeah I suppose I agree. Map nothing, or some preset, then just explicitly map what you want in subsequent calls. Or you can go a step further and leave out the ability to do ANY mapping in the creation of the runtime. After all... creating the runtime and mapping segments are sort of separate concerns and the error cases could benefit from being disentangled. 

And it's nice avoiding allocating static vectors. Less stack memory bloat.

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+  IOX_BUILDER_PARAMETER(UnmappedSegmentBehavior, unmappedSegmentBehavior, UnmappedSegmentBehavior::LoadOnDemand);
+  ...
+  public:
+    expected<DefaultRuntime, RuntimeBuilderError> create() noexcept;
+```
+
+The `create` method of this builder would then return a custom error if:
+  * A segment is requested which does not exist
+  * A segment is requested which the current process does not have access to
+  * A segment is requested which has already been captured by the `premappedSegments` option
```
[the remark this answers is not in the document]
gpalmer-latai: Yep, agreed. How do you feel about no mapping happening in `create` and just exposing API's for "map specific segment" or "map all read/write segments"?

doc/design/draft/named-segments.md
```diff
@@ -0,0 +1,219 @@
...
+    2. If the containing process has write access to the segment via one of its `userGroups`, return the corresponding segment information
+    3. Otherwise return an error indicating the publisher does not have write access to the requested segment
+2. If it is empty:
+    1. Iterate over all process `userGroups`
+    2. Determine if the user group name matches the name of one of the segments
+    3. If it does:
+        1. If a match has already been previously made, return an error indicating that the publisher must only have write access to one segment
+        2. If not, verify the process has access to the segment and record the match
+    4. At the end of iteration, if a matching segment has been found, return the segment information
+    5. Otherwise return an error indicating that no matching segment has been found
```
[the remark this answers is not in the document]
gpalmer-latai: Yeah... The reasons I mention RouDi doing this is because the logic is already there today: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/source/roudi/process_manager.cpp#L493

The runtime sends an IPC request with publisher info and it is RouDi that checks the segment manager and figures out where that publisher should publish. 

It is an interesting thought to move some of this to the Runtime... I think you'd need to expose that API through the `SharedMemoryUser` class. I'm not sure that the runtime could go so far as to acquire the port data itself (though this moves more towards the decentralized architecture of iox2 and is a nice thought). It could however identify the segment it wishes to publish into and make a more specific request to roudi along the lines of "hey, I want a publisher port and I'm going to publish into THIS segment. Feel free to check on your end that I'm able to, though I already know I'm able to".

doc/design/draft/named-segments.md
gpalmer-latai: Hm yeah. It could be pretty annoying even though you _could_ guard against it.

Another option is to simply ignore, but log a warning, about the case where the segment is already mapped. Most users wouldn't care anyway.

doc/design/draft/named-segments.md
gpalmer-latai: Yeah. It sounds like most of the existential questions revolve around refactoring the runtime. I wonder if it would make sense to move all this optional section to a different design draft and just focus on allowing publishers to specify segments in this design. The feature doesn't actually require any changes to the runtime API - just slight tweaks to the segment matching algorithms for publishers.

gpalmer-latai: I might be biased due to the fact that, the segment naming is the concrete feature I need to deliver to my company to achieve our goals. But more generally I find it good software engineering practice to separate concerns as much as possible and deliver features in bite-sized pieces 😉

doc/design/draft/named-segments.md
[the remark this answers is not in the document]
elBoberido: Yes, this would be ideal and your might already guess it ... alreadysolvediniceoryx2™

doc/design/draft/named-segments.md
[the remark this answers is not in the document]
elBoberido: Indeed. There is currently too much entangled in the ctor/initRuntime call

doc/design/draft/named-segments.md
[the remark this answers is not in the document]
elBoberido: Fuck ack :)

doc/design/draft/named-segments.md
elBoberido: Sure. Go ahead. It basically all boils down to create an API which does not impose a high burden on maintenance. As you know, the community is not that big and getting a lot of bug reports because of [Hyrum's Law](https://www.hyrumslaw.com/) is something we want to avoid.

PR #2146, opened 2023-12-27
iox-#1032 Refactor 'IOX_REQUIRE'

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

Currently `IOX_REQUIRE` can only be used with an error code which makes it semantically similar to `IOX_REPORT_FATAL_IF`. On the other hand, there is no assertion macro which cannot be deactivated via compile time switch. I think it makes more sense to have an `IOX_REQUIRE` which acts like an `IOX_ASSUME` but cannot turned off by a compile time switch.

## References

- Relates to #1032

### DIFF
--- a/doc/design/error_reporting.md
+++ b/doc/design/error_reporting.md
@@ -165,7 +165,7 @@ the case that it does not hold
 ```cpp
 int x;
 // ...
-IOX_REQUIRE(x>=0, Code::OutOfBounds);
+IOX_REQUIRE(x>=0, "required condition violation message");
 ```
 
 The condition is required to hold and this requirement is always checked.
--- a/iceoryx_hoofs/reporting/include/iox/assertions.hpp
+++ b/iceoryx_hoofs/reporting/include/iox/assertions.hpp
@@ -44,18 +44,28 @@
         iox::er::forwardPanic(CURRENT_SOURCE_LOCATION, message);                                                       \
     } while (false)
 
-/// @brief report fatal error if expr evaluates to false
-/// @note for conditions that may actually happen during correct use
-/// @param expr boolean expression that must hold
-/// @param error error object (or code)
-#define IOX_REQUIRE(expr, error) IOX_REPORT_FATAL_IF(!(expr), error)
-
 //************************************************************************************************
 //* For documentation of intent, defensive programming and debugging
 //*
 //* There are no error codes/errors required here on purpose, as it would make the use cumbersome.
 //* Instead a special internal error type is used.
 //************************************************************************************************
+
+/// @brief report fatal required condition violation if expr evaluates to false
+/// @note for conditions that may actually happen during correct use
+/// @param expr boolean expression that must hold
+/// @param message message to be forwarded in case of violation
+#define IOX_REQUIRE(expr, message)                                                                                     \
+    do                                                                                                                 \
+    {                                                                                                                  \
+        if (!(expr))                                                                                                   \
+        {                                                                                                              \
+            iox::er::forwardFatalError(iox::er::Violation::createRequiredConditionViolation(),                         \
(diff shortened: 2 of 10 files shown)


elfenpiff (dismissed):

elfenpiff (approved):

doc/design/error_reporting.md
```diff
@@ -165,7 +165,7 @@ the case that it does not hold
 ```cpp
 int x;
 // ...
-IOX_REQUIRE(x>=0, Code::OutOfBounds);
+IOX_REQUIRE(x>=0, "required condition violation message");
```
MatthiasKillat, line 168: @elBoberido I know it is merged but I think it is useful to have a code or more generic error object if it is to be passed to another framework (this is still the intention).

Some thing like

```IOX_REQUIRE(x>=0, Foo("Text", 73))```.

With sufficient support in the handler, anything can be passed and handled. In particular string literals, runtime constructed strings, objects of any kind, error wrappers around these. As of now, there exist no handling code for anything but codes (or string literals now). We decided this for simplicity.

It is important not to require to pass heavyweight objects, pointers to string literals or error codes are fine, but still allow more heavyweight objects if the user desires. Messages alone may not be descriptive enough at the handler site.

I would also liked to have the message (or any arguments) as optional, but variadic arguments do not play nice in some circumstances, but mainly if the case of no arguments must be supported (not necessarily the case here).

The semantics are also important. `IOX_REQUIRE` is not necessarily on a rare path in the "should not happen category (this would be `IOX_ASSUME`). Therefore performance is a consideration.

It may be a good idea to rethink whether we want more generic objects to be handled in the future. This should include string literals, codes and anything else. Fine for now and can be extended if we see the need.

File: iceoryx_hoofs/reporting/include/iox/assertions.hpp [hunk 2]
```diff
@@ -44,19 +44,29 @@
...
+    do                                                                                                                 \
+    {                                                                                                                  \
+        if (!(expr))                                                                                                   \
+        {                                                                                                              \
+            iox::er::forwardFatalError(iox::er::Violation::createRequiredConditionViolation(),                         \
+                                       iox::er::REQUIRED_CONDITION_VIOLATION,                                          \
+                                       CURRENT_SOURCE_LOCATION,                                                        \
+                                       message); /* @todo iox-#1032 add strigified 'expr' as '#expr' */                \
+        }                                                                                                              \
+    } while (false)
```
MatthiasKillat, line 68: @elBoberido I recently considered as trailing `static_assert` as an alternative to this `do { ... } while` madness: https://stackoverflow.com/questions/35530850/how-to-require-a-semicolon-after-a-macro.

However, it is not equivalent in general as it e.g. does not play nice in `if(...) MACRO() else { }` without parentheses. One could argue this is a good thing but I disagree as the MACRO should not impose these restrictions if possible. Also there may be more limitations. 

In other words I have no better alternative yet and this is pretty standard but also ancient from the C world.

Stringification as in the TODO is a good addition.

doc/design/error_reporting.md
elBoberido: @MatthiasKillat for that there is still `IOX_REPORT_IF` and semantically it fits also better if it is intended to be forwarded. Currently I'm even thinking about simplifying the macros even more. Instead of having `IOX_PRECONDITION`, `IOX_ASSUME` and `IOX_REQUIRE` it could be `IOX_ASSERT` and `IOX_ENFORCE`. The former would be deactivated for release builds by default and nicely match the C assert. The latter would would emphasis the stronger requirements of always being active. With this the `IOX_REQUIRE` would again become available if we want to have it as a shortcut for `IOX_REPORT_IF`. We can discuss this further in the currently active PR.

PR #2147, opened 2024-01-03
iox-#1032 Refine testing with new error reporting

elBoberido:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

This PR resets the `TestingErrorHandler` at each test start to simplify the usage to the error handler for the user in tests. Additionally, the violation error codes are moved to the file with the `Violation` class.

## References

- Relates to #1032

### DIFF
--- a/doc/design/error_reporting.md
+++ b/doc/design/error_reporting.md
@@ -157,15 +157,15 @@ Fatal errors can be conditionally reported in a similar way.
 IOX_REPORT_FATAL_IF(x<0, Code::OutOfBounds);
 ```
 
-### Require That a Condition Holds
+### Enforce a Condition
 
-Similarly we can conditionally check whether a condition does hold and report a fatal error in
+Similarly we can conditionally enforce whether a condition does hold and report a fatal error in
 the case that it does not hold
 
 ```cpp
 int x;
 // ...
-IOX_REQUIRE(x>=0, "required condition violation message");
+IOX_ENFORCE(x>=0, "enforce violation message");
 ```
 
 The condition is required to hold and this requirement is always checked.
@@ -173,67 +173,46 @@ If the condition does not hold, panic is invoked and the execution stops.
 
 This should be used for conditions that may not hold on the correct path, e.g. for error cases.
 It should not be used for assumptions that have to be true in correct code
-(use `IOX_ASSUME` or `IOX_PRECONDITION` for this).
+(use `IOX_ASSERT` for this).
 
 Note that no condition can generally be enforced in the sense that it must be true and no checking is required.
 
-## Using the API to Check Contracts and Assumptions
+### Assert a Condition
 
-The following checks can be disabled and are intended to increase safety during incorrect
+The following check is disabled for release builds and is intended to increase safety during incorrect
 use, specifically detect incorrect use at runtime.
 
-This means they should only be used for checks that are not needed in correct code (i.e. defensive
+This means it should only be used for checks that are not needed in correct code (i.e. defensive
 programming).
 
-If these checks are disabled, there is no overhead in the code, i.e. no checking or reporting takes place.
-
-### Checking Preconditions
-
-A precondition check
+When check is disabled, there is no overhead in the code, i.e. no checking or reporting takes place.
 
 ```cpp
 int f(int x)
 {
-IOX_PRECONDITION(x>=0, "precondition violation message");
+    IOX_ASSERT(x>=0, "precondition violation message");
 
-// ...
+    // some computation
+    int y = g(x);
+
+    IOX_ASSERT(y>=0, "assumption violation message");
+    // proceed assuming that y>=0 holds
 }
 ```
 
-is used to verify assumptions **BEFORE** any logic in the function body is executed. Technically copy
+Is can be used to verify preconditions before any logic in the function body is executed. Technically copy
 constructors may run before any condition can be checked, and there is also the possibility of
 reordering if the following code does not depend on the condition at all.
 This is not a problem since any reordering is not allowed to affect the observable result.
 Specifically it cannot affect the value of the precondition itself as this would change the
 observable behaviour.
 
+When used in the middle of a function it serves as documentation of assumptions that should hold at this
+point in the code before the next statement and can be used e.g. to check for out-of-bounds accesses. It can
+also be used to check postconditions.
+
 In case of violation, the violation and a (potentially empty) message are forwarded to the backend,
 panic is invoked and execution stops.
-
-The verification can be optionally disabled, and hence this also documents assumptions of the
-function itself.
-
-### Checking Assumptions
-
-Checking assumptions is similar to checking preconditions, but can happen anywhere in the code.
-
-```cpp
-int f(int x)
-{
-    // some computation
-    int y = g(x);
-
-    IOX_ASSUME(y>=0, "assumption violation message");
-    // proceed assuming that y>=0 holds
-}
-```
-
-These serve as documentation of assumptions that should hold at this point in the code before the
-next statement and can be used e.g. to check for out-of-bounds accesses. It can also be used to
-check postconditions.
-
-It should not be used at the start of a function body and instead replaced with a precondition check
-in this case.
 
 ## Marking Unreachable Code
 
@@ -394,7 +373,7 @@ Alternatively the shorthand version can be used
 int algorithm(int x)
 {
     // require that the condition holds or raise a fatal error
-    IOX_REQUIRE(!errorCondition(x), SomeError);
+    IOX_REPORT_FATAL_IF(errorCondition(x), SomeError);
     return 42;
 }
 ```
@@ -425,7 +404,7 @@ These are
 8. `errors.hpp` : supported error types and related free functions
 
 Additionally there is the `assertions.hpp` in `iceoryx_hoofs/reporting` which contains the `IOX_PANIC`,
-`IOX_UNREACHABLE`, `IOX_REQUIRE`, `IOX_PRECONDITION` and `IOX_ASSUME` macros
+`IOX_UNREACHABLE`, `IOX_ASSERT` and `IOX_ENFORCE` macros.
 
 All the files focus on singular aspects to allow fine-grained inclusion.
 All definitions have to reside in `iox::er`, which is considered a private (detail) namespace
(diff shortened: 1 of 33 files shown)


iceoryx_hoofs/reporting/include/iox/error_reporting/errors.hpp
```diff
@@ -36,9 +36,21 @@ static constexpr const char* UNKNOWN_ERROR_NAME = "unknown error";
 // By default, there are only error codes and violations.
 // Custom errors can be added but must satisfy the minimal interface.
 
+enum class ViolationErrorCode : uint32_t
+{
+    REQUIRED_CONDITION_VIOLATION,
+    ASSUMPTION_VIOLATION,
+    PRECONDITION_VIOLATION
+};
+
```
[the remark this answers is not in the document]
MatthiasKillat: @elBoberido I looked at #2146 and it is fine (and already merged anyway). One can argue whether more generic arguments including codes would be a good addition in the future, but we can go with this.

iceoryx_hoofs/reporting/include/iox/error_reporting/errors.hpp
```diff
@@ -36,9 +36,21 @@ static constexpr const char* UNKNOWN_ERROR_NAME = "unknown error";
...
 class Violation
```
MatthiasKillat, line 46: @elBoberido I was also considering having distinct types for everything, but decided against it as they are essentially all wrapper around a code for now. It should be considered if this changes.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/error_reporting/testing_error_handler.hpp
```diff
@@ -38,52 +42,68 @@ namespace testing
...
+    ///     iox::testing::ErrorHandler::init();
+    ///
+    ///     return RUN_ALL_TESTS();
+    /// }
+    /// @endcode
+    static void init() noexcept;
 
     /// @brief Defines the reaction on panic.
-    void onPanic() override;
+    void onPanic() noexcept override;
```
MatthiasKillat, line 73: I think all these handlers (`onXYZ`) should not be `noexcept`. In theory the handler code should be flexible to throw if it desires. I do not see a reason to restrict it.

Generally I am not a fan of `noexcept` in generic code. This includes ode that calls a (potentially polymorphic) function that is decided on runtime and could throw for all we know. In these cases the compiler can never know whether it will throw.

But most importantly the code should not restrict any framework configuring a handler to throw. We can argue that in this particular override in the test handler does not throw, but with C++17 `noexcept` is part of the signature. I have not checked, but specifying the derived version as `noexcept` would require the base interface function to be `noexcept` as well. This precludes any overrides that may throw, which is undesirable.

MatthiasKillat (commented):
@elBoberido Generally looks fine. We should reconsider the `noexcept` specification in the handler though unless the base interface is not restricted in such a way (I have to look whether it has to if derived is `noexcept`). There is not much point in proliferation of `noexcept` from my experience and it causes all kinds of issues .

The important thing is that `noexcept` is seen as part of the specification and hence imposes an implementation restriction (that the compiler unfortunately cannot check). We should be careful with these restrictions and leave the user as much flexibility in these generic constructs as reasonably possible without bad trade-offs.

iceoryx_hoofs/reporting/include/iox/error_reporting/errors.hpp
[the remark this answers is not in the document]
elBoberido: Looking at how `Expects` and `Ensures` is used in iceoryx I came to the conclusion that it is easily used the wrong™ way and as you also noted it cannot be enforced. Therefore I would propose to just have an `IOX_ASSERT` which is deactivated in release builds. This is something all the developer know and would be a drop in replacement for `assert`. Furthermore, @elfenpiff suggested `IOX_ENFORCE` instead of `IOX_REQUIRE` to emphasis the stronger requirements of always being active.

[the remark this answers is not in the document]
MatthiasKillat: @elBoberido `IOC_ENFORCE` instead of require is fine. We considered it originally but decided against it for reasons I do not remember. 

The same with `IOX_ASSUME` vs `IOX_ASSERT`, but here it was that we considered `IOX_DEBUG_ASSERT` as opposed to `IOX_ASSERT` that would always be active.  This was misleading and so we opted for a new name `IOX_ASSUME`, since it documents the programmers assumption at this point. In the end I fine with either of them as long as there is at least one that can be disabled (usually for performance release mode).

The way of thinking of a code should be: Should this happen in real code unless there is a bug? If yes, then use `IOX_ENFORCE`. Otherwise use `IOX_ASSERT` to document the intention.

As discussed, `Expects` corresponds to `expects`in C++ contracts that were dropped for some reason despite being a good idea (but a bad name IMO). It was meant to check preconditions in some mode (strict,  performance, axiom or something similar). Same with ensures, but for postconditions. The proposal also contained an `assert` of some kind to be used in the same way as `IOX_ASSUME`.

Now we cannot bake it into the language and hence it can be used incorrectly. I would argue it is easy to use correctly and nice to distinguish between preconditions and general assertions and maybe even disable one but not the other (say disable assertions but not preconditions). Also the distinction allows a more fine grained understanding at the handler side of what went wrong. In both cases it should likely be considered a severe error due to a bug, so that might not be that useful. Nevertheless I like to have the option.

However, I am fine with removing it for a first iteration but still do not see the difficulty. If it is an assertion in the sense of above but must hold upon entry in the function, use `IOX_PRECONDITION` (it is not the end of the world if `IOX_ASSERT` is used in its place though). It also stands out in the code and hence one can search for preconditions etc.

I do not think correct use is too much to ask for and the misuse we prevent by dropping `IOX_PRECONDITION` is not a severe case as long as there is still an `IOX_ASSERT`. In the end the end we cannot force the programmer to use any of those. While we can decide to undo this later, I think if it is now decided it will likely stay like this so I wanted to get this "right" from the beginning.

Observe how (IMO) `Expects` and `Ensures` was used incorrectly in the sense that it was not in the spirit of the C++ proposal. This is easily corrected, as the usage sites in where not that many, at least in hoofs. `Ensures` was used completely the wrong way in most cases AFAIR. Currently it serves no purpose, hence why I opted to drop anything specific for  postconditions, an `IOX_ASSERT` is fine here.

Conclusion: 
We can condense it to `IOX_ASSERT` only, but I think this is the wrong approach. Nevertheless we can start like this and reconsider later. Personally I think we should first use it on a larger scale before we make any major or adjustments, unless something clearly does not work. The machinery behind all this makes it easy to define new macros that suit the needs I think. So if the majority would prefer this simpler but IMO less expressive solution we can do this and see how it works out. I am not convinced for reasons above and would keep the distinction between `IOX_PRECONDITION` and `IOX_ASSERT`. All the other renaming is fine.

doc/design/error_reporting.md
```diff
@@ -157,84 +157,63 @@ Fatal errors can be conditionally reported in a similar way.
...
+    // some computation
+    int y = g(x);
+
+    IOX_ASSERT(y>=0, "assumption violation message");
+    // proceed assuming that y>=0 holds
 }
 ```
 
-is used to verify assumptions **BEFORE** any logic in the function body is executed. Technically copy
+Is can be used to verify preconditions before any logic in the function body is executed. Technically copy
```
MatthiasKillat, line 203:
```suggestion
It can be used to verify preconditions before any logic in the function body is executed. Technically copy
```

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/error_reporting/testing_error_handler.hpp
```diff
@@ -62,28 +82,28 @@ class TestErrorHandler : public iox::er::ErrorHandlerInterface
 
     /// @brief Indicates whether there was a panic call previously.
     /// @return true if there was a panic call, false otherwise
-    bool hasPanicked() const;
+    bool hasPanicked() const noexcept;
```
MatthiasKillat, line 85: @elBoberido What about this noexcept? Is this really useful? I had some objections here before but maybe they got lost. Have to check.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/error_reporting/testing_error_handler.hpp
MatthiasKillat: @elBoberido It is still noexcept. Is there a strong reason why? Maybe I was wrong above or did not consider something. Or it was simply forgotten. I mean it apparently compiles (C++17?) so it is probably fine. I still am not a huge fan of littering everything with `noexcept` in generic code

The specialization here is not generic and we know that it does not throw, so it is probably fine. I cannot point to it but I think I miss something important here. Oh well, if so, we will find out later.

MatthiasKillat (approved):
@elBoberido Except for my doubts about reducing everything to `IOX_ASSERT` thereby dropping `IOX_PRECONDITION` I agree with the changes. Ultimately it comes down to what we want to accomplish and the granularity of violations we want to distinguish. I am fine to merge it as is to move forward to actually use it and make adjustments based on the usage experience if needed (should be relatively easy).

I leave it to you whether the functions in the test handler should be `noexcept`. I believe they should not but if they really cannot be we find out soon enough. The interface they inherit from should not be `noexcept` if we intend to let the client code throw (which IMO we should to be least restrictive). It is no big deal for now either way.

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/error_reporting/testing_error_handler.hpp
elBoberido: I removed them from the `onPanic` and other hooks. The `hasPanicked` method cannot throw and has no hooks and therefore `noexcept` communicates exactly what the method does.

doc/design/error_reporting.md
elBoberido: I fix this in my other open PR

iceoryx_hoofs/testing/include/iceoryx_hoofs/testing/error_reporting/testing_error_handler.hpp
elBoberido: Are you sure ? `onPanic` has no `noexcept` anymore and it is the only override. All the other methods do not throw exceptions and there are now hooks to inject them.

elBoberido:
The typo is fixed in #2151. I had to adjust the testing error handler for MinGW anyway (for some reasons the longjmp did not work in the template method)

Issue #2055, opened 2023-10-23
'convert' is broken for edge cases


elBoberido:
## Required information

**Operating system:**
all

**Compiler version:**
all

**Eclipse iceoryx version:**
all

**Observed result or behaviour:**
Conversions from strings like `nan`, `inf` or `DBL_MAX` for floating point types or `ULLONG_MAX` for uint64_t fail.

**Expected result or behaviour:**
All valid values should be converted.

**Conditions where it occurred / Performed steps:**
Try to convert `inf` to `double` or `18446744073709551615` to `uint64_t`

## Detailed description of the problem

The conversion functions have some weird way to tell the user something went wrong. Let's take `strtoull` for example
```c
unsigned long long int strtoull(const char *nptr, char **endptr,  int base);
```
There are three scenarios:
1. `nptr == endptr`, the conversion failed because there were no valid characters
2.  `nptr != endptr && *endptr != '\0'`, the conversion was (probably) successful and it depends on the user if this is correct because not all characters were converted, e.g. if the user is passed `123ab` then ` *endptr` would be `a` but if the user expected the whole string to be a number, this should be an error
3. `nptr != endptr && *endptr != '\0'` , the conversion was (probably) successful

2 and 3 might still have failed if the return value was `ULLONG_MAX` **and** the `errno` is set to `ERANGE`.

All of this has to be taken care of or the conversion will fail on edge cases

## Test program

This small test program can be used to observe the behavior for string to integer and string to floating point conversion.
```cpp
#include <cstdlib>
#include <iostream>

int main() {
    struct Test {
        const char* name;
        const char* value;
    };

    const Test integer_tests[] {{"ZERO", "0"},
                                {"ULLONG_MAX", "18446744073709551615"},
                                {"ULLONG_MAX + 1", "18446744073709551616"},
                                {"ULLONG_MAX + ascii", "18446744073709551615q"},
                                {"ULLONG_MAX + 1 + ascii", "18446744073709551616q"},
                                {"String 'COFFEE'", "COFFEE"}
                               };

    std::cout << "#### Integer" << std::endl;
    for(const auto& test: integer_tests) {
        std::cout << "==== " << test.name << std::endl;
        char* end{nullptr};
        errno = 0;
        auto retval = strtoull(test.value, &end, 10);
        std::cout << "tst val: " << test.value << std::endl;
        std::cout << "ret val: " << retval << std::endl;
        std::cout << "errno: " << errno << std::endl;
[...]

PR #2151, opened 2024-01-04
iox-#2150 MinGW support for Windows

elBoberido:
## Pre-Review Checklist for the PR Author

1. [x] Add a second reviewer for complex new features or larger refactorings
1. [x] Code follows the coding style of [CONTRIBUTING.md][contributing]
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Changelog updated [in the unreleased section][changelog] including API breaking changes
1. [x] Branch follows the naming format (`iox-123-this-is-a-branch`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`iox-#123 commit text`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Update the PR title
   - Follow the same conventions as for commit messages
   - Link to the relevant issue
1. [x] Relevant issues are linked
1. [x] Add sensible notes for the reviewer
1. [x] All checks have passed (except `task-list-completed`)
1. [x] All touched (C/C++) source code files from `iceoryx_hoofs` are added to `./clang-tidy-diff-scans.txt`
1. [x] Assign PR to reviewer

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/concepts/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/release-notes/iceoryx-unreleased.md

## Notes for Reviewer

With this PR we do now support MSVC as well as MinGW to build iceoryx on Windows.

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Code according to our coding style and naming conventions
- [x] Unit tests have been written for new behavior
- [x] Public API changes are documented via doxygen
- [x] Copyright owner are updated in the changed files
- [x] All touched (C/C++) source code files from `iceoryx_hoofs` have been added to `./clang-tidy-diff-scans.txt`
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

- Closes #2150

PR #2155, opened 2024-01-08
Fix 'covert::from_string' test on aarch64 platform

Dennis40816:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

It seems that there's some unknown issue on aarch64. I disable the specific test and will try to fix it in the following commits.

## References

- #2055

### DIFF
--- a/iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
@@ -69,16 +69,7 @@ class convert_test : public Test
     {
     }
     template <typename T>
-    std::string fp_to_string(T value)
-    {
-        static_assert(std::is_floating_point<T>::value, "requires floating point type");
-
-        std::ostringstream oss;
-        oss << std::scientific << std::setprecision(std::numeric_limits<T>::digits10) << value;
-        return oss.str();
-    }
-    template <typename T>
-    std::string fp_to_string(T value, uint16_t digits)
+    std::string fp_to_string(T value, uint16_t digits = std::numeric_limits<T>::digits10)
     {
         static_assert(std::is_floating_point<T>::value, "requires floating point type");
 
@@ -703,7 +694,7 @@ TEST_F(convert_test, fromString_Float_EdgeCase_InRange_Success)
     EXPECT_THAT(float_max.value(), FloatEq(std::numeric_limits<float>::max()));
 }
 
-TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFailExceptMsvc)
+TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFail)
 {
     ::testing::Test::RecordProperty("TEST_ID", "68d4f096-a93c-406b-b081-fe50e4b1a2c9");
 
@@ -737,7 +728,7 @@ TEST_F(convert_test, fromString_Double_EdgeCase_InRange_Success)
     EXPECT_THAT(double_max.value(), DoubleEq(std::numeric_limits<double>::max()));
 }
 
-TEST_F(convert_test, fromString_Double_EdgeCase_SubNormalDouble_ShouldFailExceptMsvc)
+TEST_F(convert_test, fromString_Double_EdgeCase_SubNormalDouble_ShouldFail)
 {
     ::testing::Test::RecordProperty("TEST_ID", "af7ca2e6-ba7e-41f7-a321-5f68617d3566");
 
@@ -774,7 +765,7 @@ TEST_F(convert_test, fromString_LongDouble_EdgeCase_InRange_Success)
     EXPECT_THAT(LongDouble::Eq(long_double_max.value(), std::numeric_limits<long double>::max()), Eq(true));
 }
 
-TEST_F(convert_test, fromString_LongDouble_EdgeCase_SubNormalLongDouble_ShouldFailExceptMsvc)
+TEST_F(convert_test, fromString_LongDouble_EdgeCase_SubNormalLongDouble_ShouldFail)
 {
     ::testing::Test::RecordProperty("TEST_ID", "fb96e526-8fb6-4af9-87f0-dfd4193237a5");
 
--- a/iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
@@ -45,8 +45,27 @@ class convert_test : public Test
         static_assert(std::is_floating_point<T>::value, "fp_to_string requires floating point type");
 
         std::ostringstream oss;
-        oss << std::scientific << std::setprecision(std::numeric_limits<T>::max_digits10) << value;
+        oss << std::scientific << std::setprecision(get_digits<T>()) << value;
         return oss.str();
+    }
+
+  private:
+    // see https://github.com/eclipse-iceoryx/iceoryx/pull/2155
+    template <typename T>
+    constexpr static uint32_t get_digits()
+    {
+        if constexpr (std::is_same_v<T, float>)
+        {
+            return 9;
+        }
+        else if (std::is_same_v<T, double> || std::is_same_v<T, long double>)
+        {
+            return 19;
+        }
+        else
+        {
+            static_assert("T should be floating point");
+        }
     }
 };
 
@@ -664,8 +683,6 @@ TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFail)
 {
     ::testing::Test::RecordProperty("TEST_ID", "68d4f096-a93c-406b-b081-fe50e4b1a2c9");
 
-    // strtof will trigger ERANGE if the input is a subnormal float, resulting in a nullopt return value.
-    // note that for MSVC, sub normal float is a valid input!
     auto normal_float_min_eps = std::nextafter(std::numeric_limits<float>::min(), 0.0F);
     std::string source = fp_to_string(std::numeric_limits<float>::min() - normal_float_min_eps);
     auto float_min_dec_eps = iox::convert::from_string<float>(source.c_str());
@@ -673,14 +690,11 @@ TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFail)
     GTEST_SKIP() << "@todo iox-#2055 temporarily skipped";
 #else
     ASSERT_THAT(float_min_dec_eps.has_value(), Eq(false));
-#endif
 }
 
 TEST_F(convert_test, fromString_Double_EdgeCase_InRange_Success)
 {
     ::testing::Test::RecordProperty("TEST_ID", "d5e5e5ad-92ed-4229-8128-4ee82059fbf7");
-
-    GTEST_SKIP() << "@todo iox-#2055 Temporarily skipped due to issues in aarch64";
 
     std::string source = fp_to_string(std::numeric_limits<double>::min());
     auto double_min = iox::convert::from_string<double>(source.c_str());
@@ -709,7 +723,6 @@ TEST_F(convert_test, fromString_Double_EdgeCase_SubNormalDouble_ShouldFailExcept
     GTEST_SKIP() << "@todo iox-#2055 temporarily skipped";
 #else
     ASSERT_THAT(double_min_dec_eps.has_value(), Eq(false));
-#endif
 }
 
 TEST_F(convert_test, fromString_LongDouble_EdgeCase_InRange_Success)
@@ -744,7 +757,6 @@ TEST_F(convert_test, fromString_LongDouble_EdgeCase_SubNormalLongDouble_ShouldFa
     GTEST_SKIP() << "@todo iox-#2055 temporarily skipped";
 #else
     ASSERT_THAT(long_double_min_dec_eps.has_value(), Eq(false));
-#endif
 }
 
 /// NORMAL FLOATING POINT TYPE EDGE CASES END
--- a/iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
@@ -30,6 +30,35 @@ using namespace ::testing;
 
 using NumberType = iox::convert::NumberType;
 
+class LongDouble
+{
+  public:
+    static bool Eq(long double a, long double b)
+    {
+        IOX_LOG(DEBUG, "a: " << a << ", b: " << b);
+
+        long double min_val = std::min(std::fabs(a), std::fabs(b));
+        long double epsilon = std::fabs(min_val - std::nextafter(min_val, static_cast<long double>(0)));
+
+        IOX_LOG(DEBUG, "epsilon from min_val: " << epsilon);
+        IOX_LOG(DEBUG, "abs min_val: " << min_val);
+
+        if (epsilon <= 0 || epsilon < std::numeric_limits<long double>::min())
+        {
+            epsilon = std::numeric_limits<long double>::min();
+        }
+        IOX_LOG(DEBUG, "epsilon: " << epsilon);
+
+        long double abs_diff = std::fabs(a - b);
+        IOX_LOG(DEBUG, "fabs result: " << abs_diff);
+
+        bool is_equal = abs_diff <= epsilon;
+        IOX_LOG(DEBUG, "<< a and b " << ((is_equal) ? "IS" : "IS NOT") << " considered equal! >>");
+
+        return is_equal;
+    }
+};
+
 class convert_test : public Test
 {
   public:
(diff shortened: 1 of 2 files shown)


Dennis40816:
We have several issues:

1. **GTest_Skip() Not Being Triggered?**
   - It seems that `GTest_Skip()` is not being triggered? => For now, it will be temporarily disabled through comments.

2. **Double Min Issue on Raspberry 4b Ubuntu**
   - Upon testing on Raspberry Pi 4b with Ubuntu (aarch64), it's confirmed that converting `double min` does indeed trigger `ERANGE`. => Currently, there's no specific solution, but changing the test string from `2.22507385850720138e-308` to `2.22507385850720139e-308` (least significant digit) seems to avoid triggering `ERANGE`.

3. **Issues with Long Double Min Tests**
   - There are errors in the tests for `long double min`. Firstly, there appears to be a difference in how `std::scientific` and `std::setprecision` are handled by aarch64 and x64:
      - On aarch64:
        ```c++
        fp_to_string(std::numeric_limits<long double>::min())
        // we get 3.362103143112093506262677817321752603e-4932
        ```
      - On x64:
        ```c++
        fp_to_string(std::numeric_limits<long double>::min())
        // we get 3.362103143112093506263e-4932
        ```
   Secondly, for both aarch64 and x64, the return values for the `long double min` case are 0 (`errno == 0`), not `long double min`. Since `LongDoubleEq` is not implemented in `GTest`, only `Eq` is used, which failed to detect this issue. Therefore, a solution is needed for comparing `long double` values.

elBoberido:
I guess you found the pipeline :)

It's really weird that GTEST_SKIP did not work. I've the feeling the Cirrus CI is behind a few commits.

elBoberido:
Regarding the long double. AFAIK it is platform dependant and can be 12 or 16 bytes wide which could lead in the difference of precision.

The issue with double min might be related to rounding errors during conversion on that platform. Have you checked if it works when omiting the last digit? It might also just be a conversion issue when converting to a string in the test setup.

Dennis40816:
Yes I already checked the case with the last digit omiting (`2.2250738585072013e-308`). Since `2.2250738585072013e-308` < `2.22507385850720139e-308`, the errno is set to ERANGE, which is reasonable.

elBoberido:
Okay, I have the feeling it might be a rounding error during string conversion. Have you tried `std::setprecision(std::numeric_limits<T>::max_digits10 - 1)`?

Dennis40816:
You're right. It works! So should we then write platform-specific programs for this issue?

elBoberido:
> You're right. It works! So should we then write platform-specific programs for this issue?

My gut feeling tells me it is not the conversion code you wrote in convert.hpp but the stringification. If you are curious, you could try `snprintf(buf, buf_size, "%.17le\n", value)`. I've the feeling it will create something different than ostringstream.

Dennis40816:
Sorry, do you mean something like? 

```cpp
char buf[30];
snprintf(buf, 30, "%.17le\n", 2.22507385850720138e-308);
```

If so, the output seems as same as the oss.str().

elBoberido:
> Sorry, do you mean something like?
> 
> ```c++
> char buf[30];
> snprintf(buf, 30, "%.17le\n", 2.22507385850720138e-308);
> ```
> 
> If so, the output seems as same as the oss.str().

Not quite. I meant
```cpp
char buf[30];
snprintf(buf, 30, "%.17le\n", std::numeric_limits<double>::min());
```
I also noticed that `max_digits10` (which is 17 for double) is not really the upper limit and one could also use e.g. 18. This results in `2.225073858507201383e-308`, which means `2.22507385850720138e-308` is actually out of range. I think the issue could be solved by not using `max_digits10` but cutting the string at a position where the last digit is rounded up, e.g. 17 for double. I think the issue with long double could be solved in a similar fashion.

Unfortunately there is not much we can do since this is not that easily detectable as the subnormal issue with MSVC. On the positive side, this is an absolute edge case and I doubt we will ever have issues with it.

Dennis40816:
An interesting observation: for architectures other than aarch64, any digits beyond the 17th position do not trigger `ERANGE` for `strtod`. However, for aarch64, considering the rounding you mentioned, the following phenomenon occurs:
```cpp
#include <cmath>
#include <iostream>

int main() 
{
    constexpr int SIZE = 100;
    char buf[SIZE];
    char format_str[SIZE];
    bool should_print = false;

    for (int i = 17; i <= 100; i++)
    {
        snprintf(format_str, SIZE, "%%.%dle", i);
        snprintf(buf, SIZE, format_str, std::numeric_limits<double>::min());

        double v = strtod(buf, nullptr);

        if (should_print)
        {
            std::cout << "next i: " << i << std::endl;
            std::cout << "string: " << buf << std::endl;
            std::cout << "<<<<<" << std::endl;
            should_print = false;
        }

        if (errno == 0)
        {
            std::cout << "current i: " << i << std::endl;
            std::cout << "string: " << buf << std::endl;
            should_print = true;
        }

        errno = 0;
    }
}
```

On aarch64, this results in:
```cpp
current i: 19
string: 2.2250738585072013831e-308
next i: 20
string: 2.22507385850720138309e-308
<<<<<
current i: 24
string: 2.225073858507201383090233e-308
next i: 25
string: 2.2250738585072013830902327e-308
<<<<<
current i: 26
string: 2.22507385850720138309023272e-308
next i: 27
string: 2.225073858507201383090232717e-308
<<<<<
current i: 34
string: 2.2250738585072013830902327173324041e-308
next i: 35
string: 2.22507385850720138309023271733240406e-308
<<<<<
```
Do you think we should use the smallest correct rounding digit number exceeding `max_digits10` on aarch64 as the standard for truncation (for double it's `19`)?

elBoberido:
> I think the behavior on aarch64 is more reasonable though it might be more complicated. Do you think we should use the smallest correct rounding digit number exceeding `max_digits10` on aarch64 as the standard for truncation (for double it's `19`)?

I'm not quite sure what you meant :sweat_smile: 

My pragmatic approach would be to use `max_digits10 - 1` for the test. I think we might even get away with appending a "0" as in ASCII 0x30 to the resulting string to force an ERANGE on all platforms.

elBoberido:
@Dennis40816 I think I fixed the problem with the cache with https://github.com/eclipse-iceoryx/iceoryx/pull/2147/commits/41ec0e7cd242d182f02450afb3b083e8cb57abc3. It looks like that always the binaries cached on the master branch were used. I also deactivated the test in that PR. If the PR is not merged today I will create a separate small one to get this in soonish.

Dennis40816:
@elBoberido Glad to hear that you have solved the problem :) If I understand correctly, you are asking if this PR will be concluded today. My answer is no, it won't be finished today. I might complete it in the coming days, but not today. It might be necessary for you to create a PR to resolve this issue promptly. 

> I think the behavior on aarch64 is more reasonable though it might be more complicated. Do you think we should use the smallest correct rounding digit number exceeding max_digits10 on aarch64 as the standard for truncation (for double it's 19)?

Apologies for my limited English proficiency; it sometimes leads to misunderstandings and imprecise wording.

Dennis40816:
What I truly meant to convey is that we can base our approach on the aarch64 results, extracting a successful value that is greater than `max_digits10` but as close to `max_digits10` as possible. In the case of double, this happens when we set it to `%.19le`, so we can choose 19. Your suggestion of `max_digits10 -1` (16) is also viable; we're just using different approaches.

Additionally, adding a 0 does not always 100% lead to an ERANGE. An example is when we use 16 as the dividing point on aarch64. Even if a 0 is added at the 17th position, it still does not cause an ERANGE. Here is a C++ example:

```cpp
#include <iostream>
#include <cmath>
#include <cstring>

constexpr int SIZE = 100;
constexpr int EXPONENT_LEN = 5; // e-308
char buf[SIZE];
int main ()
{
    auto len = snprintf(buf, SIZE, "%.16le", std::numeric_limits<double>::min());

    std::cout << "input before append 0: " << buf << std::endl;

    if (len < SIZE - 1)
    {
        char* exponent_start = &buf[len - EXPONENT_LEN];
        char exponent_buf[EXPONENT_LEN + 1];
        memcpy(exponent_buf, exponent_start, EXPONENT_LEN + 1);

        // add 0
        *exponent_start = '0';

        // append exponent, including '\0'
        memcpy((exponent_start + 1), exponent_buf, EXPONENT_LEN + 1);

        std::cout << "input after  append 0: " << buf << std::endl;
    }

    auto v = strtod(buf, nullptr);

    std::cout << "output: " << v << std::endl << "errno: " << errno << std::endl;
}
```

```
# on aarch64
input before append 0: 2.2250738585072014e-308
input after  append 0: 2.22507385850720140e-308
output: 2.22507e-308
errno: 0
```

Since `2.22507385850720140e-308` is greater than `2.22507385850720138e-308`, it does not result in an ERANGE.

If I have misunderstood your point, please correct me. Thank you.

elBoberido:
> @elBoberido Glad to hear that you have solved the problem :) If I understand correctly, you are asking if this PR will be concluded today. My answer is no, it won't be finished today. I might complete it in the coming days, but not today. It might be necessary for you to create a PR to resolve this issue promptly.

No, I did not meant this PR but the one where I added the fix. If that one will not be merged today I'll create a separate one with only the fix for aarch64.

> > I think the behavior on aarch64 is more reasonable though it might be more complicated. Do you think we should use the smallest correct rounding digit number exceeding max_digits10 on aarch64 as the standard for truncation (for double it's 19)?
> 
> Apologies for my limited English proficiency; it sometimes leads to misunderstandings and imprecise wording.

No, problem. I'm also a non-native speaker and have the same issues :)

iceoryx_hoofs/utility/include/iox/detail/convert.inl
```diff
@@ -363,16 +363,20 @@ inline bool convert::is_within_range(const SourceType& source_val) noexcept
...
     }
-    // is_arithmetic_v
     if constexpr (std::is_floating_point_v<SourceType>)
     {
         // special cases for floating point
         if (std::isnan(source_val) || std::isinf(source_val))
         {
             return true;
         }
+#ifdef _MSC_VER
```
elBoberido, line 373: Is this compiler switch required? If yes, then it needs to be `_WIN32` since I noticed a similar behavior with MinGW where i had to temporarily deactivate some tests on the branch in order to make the CI happy.

iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
```diff
@@ -660,22 +679,14 @@ TEST_F(convert_test, fromString_Float_EdgeCase_InRange_Success)
...
-#ifdef _MSC_VER
-    ASSERT_THAT(float_min_dec_eps.has_value(), Eq(true));
-    ASSERT_THAT(std::fpclassify(float_min_dec_eps.value()), Eq(FP_SUBNORMAL));
-    EXPECT_THAT(float_min_dec_eps.value(), FloatNear(0.0F, std::numeric_limits<float>::min()));
-#else
     ASSERT_THAT(float_min_dec_eps.has_value(), Eq(false));
-#endif
 }
 
 TEST_F(convert_test, fromString_Double_EdgeCase_InRange_Success)
```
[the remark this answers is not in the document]
Dennis40816: I think we don't need GTEST_SKIP anymore if we apply the changes for aarch64.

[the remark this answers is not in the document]
elBoberido: Exactly :)

iceoryx_hoofs/utility/include/iox/detail/convert.inl
Dennis40816: Oh, you're correct. I'll remove the branch cause it seems ok to apply to every circumstance.

elBoberido (approved):

elBoberido:
@Dennis40816 it seems there is now an issue with long double. I'm going to merge #2151 and will check whether it will introduce breakage in your branch. I had to skip some of tests so you will have merge conflicts after the merge but you can just overwrite all of my changes.

Dennis40816:
@elBoberido I wrote a prototype for solving the rounding problem on aarch64. 
[A demo](https://godbolt.org/z/xf5de6Wj5) shows that how I modified the original string to the closest in range string.

elBoberido:
@Dennis40816 I really appreciate your dedication :)

Saying that, I think it is not necessary to increase the complexity that much for this edge case. For the subnormal float on Windows there were a whole range of numbers which would lead to a different behavior and it was easy to unify the behavior. Here it's basically only the min value. Therefore I would not try to catch this at the conversion function but just live with this small platform specific behavior since there will also be other small platform dependent differences when calculating with floats.

I think we can live with manually specifying the string instead of using `fp_to_string` for the min value in the long double tests. Maybe with a comment in the test why it is not used. Alternatively, since the long double test worked before setting the precision to 19, maybe using `max_digits10` for long double will also help to make the test pass on arch.

Dennis40816:
Indeed, there's no need for such complexity. :) Following your suggestion, I will specialize only for the case of LongDouble.

Dennis40816:
Now, all tests pass on aarch64 but there's a small issue on windows-mingw. The test case `fromString_LongDouble_EdgeCase_SubNormalLongDouble_ShouldFailExceptMsvc` failed. However, when I use the following code to test on windows-mingw, I got `ERANGE` as expected.

```bash
# env
# g++ --version                                                                                           ─╯ 
g++.exe (x86_64-posix-seh-rev2, Built by MinGW-W64 project) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```cpp
#include <cstdio>
#include <iostream>
#include <cmath>
#include <sstream>
#include <iomanip>

template <typename T>
std::string fp_to_string(T value)
{
	static_assert(std::is_floating_point<T>::value, "requires floating point type");

	std::ostringstream oss;
	oss << std::scientific << std::setprecision(std::numeric_limits<T>::digits10) << value;
	return oss.str();
}

int main()
{
  auto normal_long_double_min_eps = std::nextafter(std::numeric_limits<long double>::min(), 0.0L);
  std::string source = fp_to_string(std::numeric_limits<long double>::min() - normal_long_double_min_eps);
  std::cout << "source: " << source << std::endl;
  auto long_double_min_dec_eps = strtold(source.c_str(), nullptr);

  char buf[50];
  snprintf(buf, 50, "%.18Le", long_double_min_dec_eps);
  
  std::cout << "digits10: " << std::numeric_limits<long double>::digits10 << std::endl; 
  std::cout << "dec eps: " << buf << std::endl;
  std::cout << "errno: " << errno << std::endl;
}

```
```
source: 3.645199531882474603e-4951
digits10: 18
dec eps: 3.645199531882474603e-4951
errno: 34
```
Any idea why it happened?

elBoberido:
I think something went wrong during the rebase. I cannot find this code anymore
```cpp
        // should be normal or zero
        if (!std::isnormal(source_val) && (source_val != 0.0))
        {
            return false;
        }
```
It once was right below the `isnan` check.

Also, I think all the if-defs can be removed. At least it worked with a previous state of the branch I tested on Windows with MinGW.

If you squash everything to two commits, it should become obvious where things went wrong. The first 4 commits could be squashed to a ´isnormal´ commit and the remaining ones to a `fix tests` commit. In the last one the removal of th `isnormal` then needs to be reverted.

Dennis40816:
Oops, I misunderstood the following:

> For the subnormal float on Windows there were a whole range of numbers which would lead to a different behavior and it was easy to unify the behavior. Here it's basically only the min value. Therefore I would not try to catch this at the conversion function but just live with this small platform specific behavior since there will also be other small platform dependent differences when calculating with floats.

I thought it means to not consider subnormal problem of specific platform. Therefore, I deleted the code you mentioned and readd if-def branches. Sorry for that.

iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
```diff
@@ -30,6 +30,35 @@ using namespace ::testing;
...
+        {
+            epsilon = std::numeric_limits<long double>::min();
+        }
+        IOX_LOG(DEBUG, "epsilon: " << epsilon);
+
+        long double abs_diff = std::fabs(a - b);
+        IOX_LOG(DEBUG, "fabs result: " << abs_diff);
+
+        bool is_equal = abs_diff <= epsilon;
+        IOX_LOG(DEBUG, "<< a and b " << ((is_equal) ? "IS" : "IS NOT") << " considered equal! >>");
```
elBoberido, line 56: Nice usage of the testing logger to give some context when a test fails :)

iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
```diff
@@ -644,107 +683,105 @@ TEST_F(convert_test, fromString_Float_EdgeCase_InRange_Success)
...
 
-    source = fp_to_string(std::numeric_limits<float>::max());
+    source = fp_to_string(std::numeric_limits<float>::max(), PLATFORM_DIGIT_WORKAROUND_MAX);
     auto float_max = iox::convert::from_string<float>(source.c_str());
     ASSERT_THAT(float_max.has_value(), Eq(true));
     EXPECT_THAT(float_max.value(), FloatEq(std::numeric_limits<float>::max()));
 }
 
-TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFail)
+TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFailExceptMsvc)
```
elBoberido, line 706:
```suggestion
TEST_F(convert_test, fromString_Float_EdgeCase_SubNormalFloat_ShouldFail
```
Can you please also change it for the other cases with `ExceptMsvc`

iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
```diff
@@ -42,10 +71,19 @@ class convert_test : public Test
...
-        static_assert(std::is_floating_point<T>::value, "fp_to_string requires floating point type");
+        static_assert(std::is_floating_point<T>::value, "requires floating point type");
 
         std::ostringstream oss;
-        oss << std::scientific << std::setprecision(std::numeric_limits<T>::max_digits10) << value;
+        oss << std::scientific << std::setprecision(std::numeric_limits<T>::digits10) << value;
+        return oss.str();
+    }
+    template <typename T>
+    std::string fp_to_string(T value, uint16_t digits)
```
elBoberido, line 81: Just as food for thought. The two `fp_to_string` methods can be combined like this
```suggestion
    std::string fp_to_string(T value, uint16_t digits = std::numeric_limits<T>::digits10)
```

elBoberido (commented):
Ready to merge. Just some test names do not fit anymore and there is potential for a small code reuse. If it weren't for the test names, I would have already merged it :)

iceoryx_hoofs/test/moduletests/test_utility_convert.cpp
Dennis40816: I don't notice that :<

elBoberido: No problem :)

elBoberido (approved):
Thanks for your patience. Just waiting for the CI to finish and then it'll be merged

PR #2158, opened 2024-01-10
iox-#2157 Replace uint32_t with uint64_t for payload size

### LINKED ISSUE #2157: Allow For memory chunks bigger than 4GB
As discussed this seems to be a "technical unnecessary restriction."
From my experiments it seems possible to fix it by tweaking some of the types 

### Discussed in https://github.com/eclipse-iceoryx/iceoryx/discussions/2149

<div type='discussions-op-text'>

<sup>Originally posted by **kozakusek** January  3, 2024</sup>
When trying to start RouDi with a 4GB segment.mempool I encountered an unpleasant error:

```
 Condition: (t >= Minimum) && (t <= Maximum) in iox::range<T, Minimum, Maximum>::range(T) [with T = unsigned int; T Minimum = 1; T Maximum = 4294967255] is violated. (/workdir/iceoryx/iceoryx_hoofs/primitives/include/iox/algorithm.hpp:208)
```

Then after reading into the source code I noticed that in many cases (e.g. `iox_pub_loan_chunk`) the variable that refers to size is of type `uint32_t`.

Is there a reason for this limitation? Are you planning to add support for bigger memory chunks by adding option to hold size in 64bit integers?

Thanks</div>

kozakusek:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes #2157

### DIFF
--- a/iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
@@ -35,12 +35,12 @@ struct MePooConfig
     struct Entry
     {
         /// @brief set the size and count of memory chunks
-        Entry(uint32_t f_size, uint32_t f_chunkCount) noexcept
+        Entry(uint64_t f_size, uint32_t f_chunkCount) noexcept
             : m_size(f_size)
             , m_chunkCount(f_chunkCount)
         {
         }
-        uint32_t m_size{0};
+        uint64_t m_size{0};
         uint32_t m_chunkCount{0};
     };
 
--- a/iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
@@ -35,9 +35,9 @@ struct MePooConfig
     struct Entry
     {
         /// @brief set the size and count of memory chunks
-        Entry(uint64_t f_size, uint32_t f_chunkCount) noexcept
-            : m_size(f_size)
-            , m_chunkCount(f_chunkCount)
+        Entry(uint64_t size, uint32_t chunkCount) noexcept
+            : m_size(size)
+            , m_chunkCount(chunkCount)
         {
         }
         uint64_t m_size{0};
@@ -56,7 +56,7 @@ struct MePooConfig
 
     /// @brief Function for adding new entry
     /// @param[in] Entry structure of mempool configuration
-    void addMemPool(Entry f_entry) noexcept;
+    void addMemPool(Entry entry) noexcept;
 
     /// @brief Function for creating default memory pools
     MePooConfig& setDefaults() noexcept;
--- a/iceoryx_posh/source/mepoo/chunk_settings.cpp
+++ b/iceoryx_posh/source/mepoo/chunk_settings.cpp
@@ -63,15 +63,20 @@ expected<ChunkSettings, ChunkSettings::Error> ChunkSettings::create(const uint64
         return err(ChunkSettings::Error::USER_HEADER_SIZE_NOT_MULTIPLE_OF_ITS_ALIGNMENT);
     }
 
-    uint64_t requiredChunkSize =
-        calculateRequiredChunkSize(userPayloadSize, adjustedUserPayloadAlignment, userHeaderSize);
+    auto expectChunkSize = calculateRequiredChunkSize(userPayloadSize, adjustedUserPayloadAlignment, userHeaderSize);
+    if (expectChunkSize.has_error())
+    {
+        return err(expectChunkSize.get_error());
+    }
+    uint64_t requiredChunkSize = expectChunkSize.value();
+
 
     return ok(ChunkSettings{
         userPayloadSize, adjustedUserPayloadAlignment, userHeaderSize, adjustedUserHeaderAlignment, requiredChunkSize});
 }
-uint64_t ChunkSettings::calculateRequiredChunkSize(const uint64_t userPayloadSize,
-                                                   const uint32_t userPayloadAlignment,
-                                                   const uint32_t userHeaderSize) noexcept
+
+expected<uint64_t, ChunkSettings::Error> ChunkSettings::calculateRequiredChunkSize(
+    const uint64_t userPayloadSize, const uint32_t userPayloadAlignment, const uint32_t userHeaderSize) noexcept
 {
     // have a look at »Required Chunk Size Calculation« in chunk_header.md for more details regarding the calculation
     if (userHeaderSize == 0)
@@ -79,17 +84,29 @@ uint64_t ChunkSettings::calculateRequiredChunkSize(const uint64_t userPayloadSiz
         // the most simple case with no user-header and the user-payload adjacent to the ChunkHeader
         if (userPayloadAlignment <= alignof(mepoo::ChunkHeader))
         {
+            if (userPayloadSize > std::numeric_limits<uint64_t>::max() - sizeof(ChunkHeader))
+            {
+                return err(ChunkSettings::Error::REQUIRED_CHUNK_SIZE_EXCEEDS_MAX_CHUNK_SIZE);
+            }
+
             uint64_t requiredChunkSize = sizeof(ChunkHeader) + userPayloadSize;
 
-            return requiredChunkSize;
+            return ok(requiredChunkSize);
         }
 
         // the second most simple case with no user-header but the user-payload alignment
         // exceeds the ChunkHeader alignment and is therefore not necessarily adjacent
         uint64_t preUserPayloadAlignmentOverhang = sizeof(ChunkHeader) - alignof(ChunkHeader);
+
+        if (userPayloadSize
+            > std::numeric_limits<uint64_t>::max() - preUserPayloadAlignmentOverhang - userPayloadAlignment)
+        {
+            return err(ChunkSettings::Error::REQUIRED_CHUNK_SIZE_EXCEEDS_MAX_CHUNK_SIZE);
+        }
+
         uint64_t requiredChunkSize = preUserPayloadAlignmentOverhang + userPayloadAlignment + userPayloadSize;
 
-        return requiredChunkSize;
+        return ok(requiredChunkSize);
     }
 
     // the most complex case with a user-header
@@ -98,9 +115,15 @@ uint64_t ChunkSettings::calculateRequiredChunkSize(const uint64_t userPayloadSiz
     uint64_t headerSize = sizeof(ChunkHeader) + userHeaderSize;
     uint64_t preUserPayloadAlignmentOverhang = align(headerSize, ALIGNMENT_OF_USER_PAYLOAD_OFFSET_T);
     uint64_t maxPadding = algorithm::maxVal(SIZE_OF_USER_PAYLOAD_OFFSET_T, static_cast<uint64_t>(userPayloadAlignment));
+
+    if (userPayloadSize > std::numeric_limits<uint64_t>::max() - preUserPayloadAlignmentOverhang - maxPadding)
+    {
+        return err(ChunkSettings::Error::REQUIRED_CHUNK_SIZE_EXCEEDS_MAX_CHUNK_SIZE);
+    }
+
     uint64_t requiredChunkSize = preUserPayloadAlignmentOverhang + maxPadding + userPayloadSize;
 
-    return requiredChunkSize;
+    return ok(requiredChunkSize);
 }
 
 uint64_t ChunkSettings::requiredChunkSize() const noexcept
--- a/iceoryx_posh/source/mepoo/chunk_settings.cpp
+++ b/iceoryx_posh/source/mepoo/chunk_settings.cpp
@@ -23,11 +23,11 @@ namespace iox
 {
 namespace mepoo
 {
-ChunkSettings::ChunkSettings(const uint32_t userPayloadSize,
+ChunkSettings::ChunkSettings(const uint64_t userPayloadSize,
                              const uint32_t userPayloadAlignment,
                              const uint32_t userHeaderSize,
                              const uint32_t userHeaderAlignment,
-                             const uint32_t requiredChunkSize) noexcept
+                             const uint64_t requiredChunkSize) noexcept
     : m_userPayloadSize(userPayloadSize)
     , m_userPayloadAlignment(userPayloadAlignment)
     , m_userHeaderSize(userHeaderSize)
@@ -36,7 +36,7 @@ ChunkSettings::ChunkSettings(const uint32_t userPayloadSize,
 {
 }
 
-expected<ChunkSettings, ChunkSettings::Error> ChunkSettings::create(const uint32_t userPayloadSize,
+expected<ChunkSettings, ChunkSettings::Error> ChunkSettings::create(const uint64_t userPayloadSize,
                                                                     const uint32_t userPayloadAlignment,
                                                                     const uint32_t userHeaderSize,
                                                                     const uint32_t userHeaderAlignment) noexcept
@@ -66,18 +66,10 @@ expected<ChunkSettings, ChunkSettings::Error> ChunkSettings::create(const uint32
     uint64_t requiredChunkSize =
         calculateRequiredChunkSize(userPayloadSize, adjustedUserPayloadAlignment, userHeaderSize);
 
-    if (requiredChunkSize > std::numeric_limits<uint32_t>::max())
-    {
-        return err(ChunkSettings::Error::REQUIRED_CHUNK_SIZE_EXCEEDS_MAX_CHUNK_SIZE);
-    }
-
-    return ok(ChunkSettings{userPayloadSize,
-                            adjustedUserPayloadAlignment,
-                            userHeaderSize,
-                            adjustedUserHeaderAlignment,
-                            static_cast<uint32_t>(requiredChunkSize)});
+    return ok(ChunkSettings{
+        userPayloadSize, adjustedUserPayloadAlignment, userHeaderSize, adjustedUserHeaderAlignment, requiredChunkSize});
 }
-uint64_t ChunkSettings::calculateRequiredChunkSize(const uint32_t userPayloadSize,
+uint64_t ChunkSettings::calculateRequiredChunkSize(const uint64_t userPayloadSize,
                                                    const uint32_t userPayloadAlignment,
                                                    const uint32_t userHeaderSize) noexcept
 {
@@ -111,12 +103,12 @@ uint64_t ChunkSettings::calculateRequiredChunkSize(const uint32_t userPayloadSiz
     return requiredChunkSize;
 }
 
-uint32_t ChunkSettings::requiredChunkSize() const noexcept
+uint64_t ChunkSettings::requiredChunkSize() const noexcept
 {
     return m_requiredChunkSize;
 }
 
-uint32_t ChunkSettings::userPayloadSize() const noexcept
+uint64_t ChunkSettings::userPayloadSize() const noexcept
 {
     return m_userPayloadSize;
 }
--- a/iceoryx_posh/source/mepoo/chunk_settings.cpp
+++ b/iceoryx_posh/source/mepoo/chunk_settings.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2021 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2024 by Bartlomiej Kozaryna <kozarynabartlomiej@gmail.com>. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/iceoryx_posh/source/mepoo/mem_pool.cpp
+++ b/iceoryx_posh/source/mepoo/mem_pool.cpp
@@ -49,6 +49,7 @@ MemPool::MemPool(const greater_or_equal<uint64_t, CHUNK_MEMORY_ALIGNMENT> chunkS
 {
     if (isMultipleOfAlignment(chunkSize))
     {
+        IOX_EXPECTS(m_chunkSize <= std::numeric_limits<uint64_t>::max() / m_numberOfChunks);
         auto allocationResult = chunkMemoryAllocator.allocate(static_cast<uint64_t>(m_numberOfChunks) * m_chunkSize,
                                                               CHUNK_MEMORY_ALIGNMENT);
         IOX_EXPECTS(allocationResult.has_value());
@@ -63,9 +64,8 @@ MemPool::MemPool(const greater_or_equal<uint64_t, CHUNK_MEMORY_ALIGNMENT> chunkS
     else
     {
         IOX_LOG(FATAL,
-                "Chunk size must be multiple of '" << CHUNK_MEMORY_ALIGNMENT << "'! Requested size is "
-                                                   << static_cast<uint64_t>(chunkSize) << " for "
-                                                   << static_cast<uint32_t>(numberOfChunks) << " chunks!");
+                "Chunk size must be multiple of '" << CHUNK_MEMORY_ALIGNMENT << "'! Requested size is " << chunkSize
+                                                   << " for " << numberOfChunks << " chunks!");
         errorHandler(PoshError::MEPOO__MEMPOOL_CHUNKSIZE_MUST_BE_MULTIPLE_OF_CHUNK_MEMORY_ALIGNMENT);
     }
 }
--- a/iceoryx_posh/source/mepoo/mem_pool.cpp
+++ b/iceoryx_posh/source/mepoo/mem_pool.cpp
@@ -29,7 +29,7 @@ namespace mepoo
 MemPoolInfo::MemPoolInfo(const uint32_t usedChunks,
                          const uint32_t minFreeChunks,
                          const uint32_t numChunks,
-                         const uint32_t chunkSize) noexcept
+                         const uint64_t chunkSize) noexcept
     : m_usedChunks(usedChunks)
     , m_minFreeChunks(minFreeChunks)
     , m_numChunks(numChunks)
@@ -39,7 +39,7 @@ MemPoolInfo::MemPoolInfo(const uint32_t usedChunks,
 
 constexpr uint64_t MemPool::CHUNK_MEMORY_ALIGNMENT;
 
-MemPool::MemPool(const greater_or_equal<uint32_t, CHUNK_MEMORY_ALIGNMENT> chunkSize,
+MemPool::MemPool(const greater_or_equal<uint64_t, CHUNK_MEMORY_ALIGNMENT> chunkSize,
                  const greater_or_equal<uint32_t, 1> numberOfChunks,
                  iox::BumpAllocator& managementAllocator,
                  iox::BumpAllocator& chunkMemoryAllocator) noexcept
@@ -64,13 +64,13 @@ MemPool::MemPool(const greater_or_equal<uint32_t, CHUNK_MEMORY_ALIGNMENT> chunkS
     {
         IOX_LOG(FATAL,
                 "Chunk size must be multiple of '" << CHUNK_MEMORY_ALIGNMENT << "'! Requested size is "
-                                                   << static_cast<uint32_t>(chunkSize) << " for "
+                                                   << static_cast<uint64_t>(chunkSize) << " for "
                                                    << static_cast<uint32_t>(numberOfChunks) << " chunks!");
         errorHandler(PoshError::MEPOO__MEMPOOL_CHUNKSIZE_MUST_BE_MULTIPLE_OF_CHUNK_MEMORY_ALIGNMENT);
     }
 }
 
-bool MemPool::isMultipleOfAlignment(const uint32_t value) const noexcept
+bool MemPool::isMultipleOfAlignment(const uint64_t value) const noexcept
 {
     return (value % CHUNK_MEMORY_ALIGNMENT == 0U);
 }
@@ -101,14 +101,14 @@ void* MemPool::getChunk() noexcept
     return indexToPointer(index, m_chunkSize, m_rawMemory.get());
 }
 
-void* MemPool::indexToPointer(uint32_t index, uint32_t chunkSize, void* const rawMemoryBase) noexcept
+void* MemPool::indexToPointer(uint32_t index, uint64_t chunkSize, void* const rawMemoryBase) noexcept
 {
     const auto offset = static_cast<uint64_t>(index) * chunkSize;
     return static_cast<void*>(static_cast<uint8_t*>(rawMemoryBase) + offset);
 }
 
 uint32_t
-MemPool::pointerToIndex(const void* const chunk, const uint32_t chunkSize, const void* const rawMemoryBase) noexcept
+MemPool::pointerToIndex(const void* const chunk, const uint64_t chunkSize, const void* const rawMemoryBase) noexcept
 {
     const auto offset =
         static_cast<uint64_t>(static_cast<const uint8_t*>(chunk) - static_cast<const uint8_t*>(rawMemoryBase));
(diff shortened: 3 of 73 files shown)


elBoberido:
@kozakusek commit message is fine. Now the more tedious work starts to make it work on all platforms :)

elBoberido:
It seems the code is not formatted with clang-format. You can use `tools/git-hooks` to automate some of the stuff which is checked by the CI, e.g. formatting the code or adding the issue number to the commit messages

kozakusek:
@elBoberido the biggest issue I see now  is that 
- *TooLargeUserPayload_Fails tests are no longer relevant and
- I might not fully get the idea behind CHUNK_HEADER_VERSION and the threats in chunk_header.hpp: "m_chunkSize and m_chunkHeaderVersion should therefore neither changed the type, nor the position" which might be the cause for failure for the other three tests

elBoberido:
@kozakusek I did a stupid mistake when I put `chunkSize` at the beginning of the `ChunkHeader` struct instead of `chunkHeaderVersion`. I never thought the type of `chunkSize` would change but only that maybe new members would be added. The reason for `CHUNK_HEADER_VERSION` and the test is to detect when a subscriber receives a chunk with an incompatible `ChunkHeader`. This can be the case when someone implements something like record&replay and stores all the data. If the the `ChunkHeader` would change after storing the data, a replay would result in weird bugs in the iceoryx layer. The issue we need to solve now is to keep the `chunkHeaderVersion` at an offset of 32 bits in the `ChunkHeader` struct. I think this can be achieved by swapping `userHeaderSize` with `chunkSize`. I need to check whether this would cause other issues but I think we should be safe.

kozakusek:
So one last thing before I open this for review - should I add a unit test for memory chunks > 4GB?
Do the CI machines have enough RAM for this?

elBoberido:
> So one last thing before I open this for review - should I add a unit test for memory chunks > 4GB? Do the CI machines have enough RAM for this? @elBoberido could you confirm?

Unfortunately the CI seems to not reliably have that much memory. AFAIR we already had to deactivate on test which required > 4GB. I need to check what changed maybe it is possible to mock a few things and test at least the algorithm.

kozakusek:
Cleaned up the mess in examples and tests.
Hopefully fixed the issue with building on Mac.
Let me know if I and how should I mock 'stuff' for testing

elBoberido:
CI looks good. Thanks. Will have a look at the changes the next days.

elBoberido (commented):
Will continue tomorrow, it's already quite late. Just had a look at the C binding and the release notes

iceoryx_posh/source/mepoo/chunk_settings.cpp
```diff
@@ -66,18 +66,10 @@ expected<ChunkSettings, ChunkSettings::Error> ChunkSettings::create(const uint32
...
-    return ok(ChunkSettings{userPayloadSize,
-                            adjustedUserPayloadAlignment,
-                            userHeaderSize,
-                            adjustedUserHeaderAlignment,
-                            static_cast<uint32_t>(requiredChunkSize)});
+    return ok(ChunkSettings{
+        userPayloadSize, adjustedUserPayloadAlignment, userHeaderSize, adjustedUserHeaderAlignment, requiredChunkSize});
 }
-uint64_t ChunkSettings::calculateRequiredChunkSize(const uint32_t userPayloadSize,
+uint64_t ChunkSettings::calculateRequiredChunkSize(const uint64_t userPayloadSize,
```
elBoberido, line 72: Here we definitely needs  changes since the calculation could easily result in an overflow. It was easier when the input parameter were `uint32_t`.

Luckily we already moved to the static creation method pattern instead of calling the ctor explicitly and therefore we can also use an `iox::expected` here and forward the error in the `create` method

iceoryx_posh/source/mepoo/mem_pool.cpp
```diff
@@ -39,7 +39,7 @@ MemPoolInfo::MemPoolInfo(const uint32_t usedChunks,
 
 constexpr uint64_t MemPool::CHUNK_MEMORY_ALIGNMENT;
 
-MemPool::MemPool(const greater_or_equal<uint32_t, CHUNK_MEMORY_ALIGNMENT> chunkSize,
+MemPool::MemPool(const greater_or_equal<uint64_t, CHUNK_MEMORY_ALIGNMENT> chunkSize,
```
elBoberido, line 42: Here we also need to check for potential overflows when `chunkSize` is multiplied with `numberOfChunks`

iceoryx_posh/test/moduletests/test_popo_server_port_common.hpp
```diff
@@ -222,8 +222,8 @@ class ServerPort_test : public Test
     iox::optional<SutServerPort> clientPortForStateTransitionTests;
 
   public:
-    static constexpr uint32_t USER_PAYLOAD_SIZE{32U};
-    static constexpr uint32_t USER_PAYLOAD_ALIGNMENT{8U};
+    static constexpr uint64_t USER_PAYLOAD_SIZE{32U};
+    static constexpr uint64_t USER_PAYLOAD_ALIGNMENT{8U};
```
elBoberido, line 226: The alignment could stay as `uint32_t`

iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
```diff
@@ -261,8 +261,8 @@ TEST_F(PublisherPort_test, allocatingAChunkWithoutUserHeaderAndLargeUserPayloadA
 TEST_F(PublisherPort_test, allocatingAChunkWithLargeUserHeaderResultsInLargeChunk)
 {
     ::testing::Test::RecordProperty("TEST_ID", "598e04d8-8a37-43ef-b686-64e7b2723ffe");
-    constexpr uint32_t USER_PAYLOAD_SIZE{SMALL_CHUNK / 2};
-    constexpr uint32_t LARGE_USER_HEADER_SIZE{SMALL_CHUNK};
+    constexpr uint64_t USER_PAYLOAD_SIZE{SMALL_CHUNK / 2};
+    constexpr uint64_t LARGE_USER_HEADER_SIZE{SMALL_CHUNK};
```
elBoberido, line 265: This could also stay `uint32_t`

iceoryx_posh/test/moduletests/test_popo_publisher_port.cpp
```diff
@@ -63,10 +63,10 @@ class PublisherPort_test : public Test
     static constexpr size_t MEMORY_SIZE = 1024 * 1024;
     uint8_t m_memory[MEMORY_SIZE];
     static constexpr uint32_t NUM_CHUNKS_IN_POOL = 20;
-    static constexpr uint32_t SMALL_CHUNK = 128;
-    static constexpr uint32_t BIG_CHUNK = 256;
+    static constexpr uint64_t SMALL_CHUNK = 128;
+    static constexpr uint64_t BIG_CHUNK = 256;
 
-    static constexpr uint32_t USER_PAYLOAD_ALIGNMENT = iox::CHUNK_DEFAULT_USER_PAYLOAD_ALIGNMENT;
+    static constexpr uint64_t USER_PAYLOAD_ALIGNMENT = iox::CHUNK_DEFAULT_USER_PAYLOAD_ALIGNMENT;
```
elBoberido, line 69: This also. If you changed it to prevent casts in the tests, then this can be left as it is now.

iceoryx_posh/test/moduletests/test_popo_client_port.cpp
```diff
@@ -206,8 +206,8 @@ class ClientPort_test : public Test
     iox::optional<SutClientPort> clientPortForStateTransitionTests;
 
   public:
-    static constexpr uint32_t USER_PAYLOAD_SIZE{32U};
-    static constexpr uint32_t USER_PAYLOAD_ALIGNMENT{8U};
+    static constexpr uint64_t USER_PAYLOAD_SIZE{32U};
+    static constexpr uint64_t USER_PAYLOAD_ALIGNMENT{8U};
```
elBoberido, line 210: Same here, `uint32_t`

iceoryx_posh/test/moduletests/test_popo_chunk_sender.cpp
```diff
@@ -72,12 +72,12 @@ class ChunkSender_test : public Test
...
     static constexpr uint32_t NUM_CHUNKS_IN_POOL = 20;
-    static constexpr uint32_t SMALL_CHUNK = 128;
-    static constexpr uint32_t BIG_CHUNK = 256;
+    static constexpr uint64_t SMALL_CHUNK = 128;
+    static constexpr uint64_t BIG_CHUNK = 256;
     static constexpr uint64_t HISTORY_CAPACITY = 4;
     static constexpr uint32_t MAX_NUMBER_QUEUES = 128;
 
-    static constexpr uint32_t USER_PAYLOAD_ALIGNMENT = iox::CHUNK_DEFAULT_USER_PAYLOAD_ALIGNMENT;
+    static constexpr uint64_t USER_PAYLOAD_ALIGNMENT = iox::CHUNK_DEFAULT_USER_PAYLOAD_ALIGNMENT;
```
elBoberido, line 80: Also `uint32_t`

iceoryx_posh/test/moduletests/test_mepoo_mempool.cpp
```diff
@@ -183,7 +183,7 @@ TEST_F(MemPool_test, MempoolCtorWhenChunkSizeIsNotAMultipleOfAlignmentReturnErro
 TEST_F(MemPool_test, MempoolCtorWhenChunkSizeIsSmallerThanChunkMemoryAlignmentGetsTerminated)
 {
     ::testing::Test::RecordProperty("TEST_ID", "52df897a-0847-476c-9d2f-99cb16432199");
-    constexpr uint32_t CHUNK_SIZE_SMALLER_THAN_MEMORY_ALIGNMENT = iox::mepoo::MemPool::CHUNK_MEMORY_ALIGNMENT - 1U;
+    constexpr uint64_t CHUNK_SIZE_SMALLER_THAN_MEMORY_ALIGNMENT = iox::mepoo::MemPool::CHUNK_MEMORY_ALIGNMENT - 1U;
```
elBoberido, line 186: Also `uint32_t`

elBoberido (commented):
Not yet fully done. I have saved some of the critical files for tomorrow :)

Btw, please don't squash commits from now on since this will make the review harder and there are subtle thinks to look at.

iceoryx_posh/source/mepoo/chunk_settings.cpp
elfenpiff: @elBoberido I think having an overflow here is most likely caused by just a programming mistake, and therefore I would vote to use `IOX_ENSURES` here and panic when this occurs. I cannot imagine a use case where an overflow can happen, and it is a real world example.

iceoryx_posh/source/mepoo/mem_pool.cpp
elfenpiff: @elBoberido yes, but I would call here again panic and not work with an expected.

elfenpiff (changes requested):
Since this affects a lot of files and from the comments from @elBoberido I saw that there are maybe even more places where we overlooked to change an `uint32_t` to an `uint64_t`. This could result in weird unsigned integer overflow bugs.

Therefore, I would add a test that verifies that payloads larger than 4GB are actually supported. Those tests shall be added to:
* `iceoryx_posh/test/integrationtests/test_client_server.cpp`
* `iceoryx_posh/test/integrationtests/test_publish_subscriber_communication.cpp`

I would introduce a test struct like:
```cxx
struct LargeData {
  uint8_t payload[4294967296 + 100]; // + 100 to be 100 bytes above the original limit
};
```

And use it for the typed communication. Also we need to adjust the roudi config to support those large structs. I think this should work in the test setup

```cxx
RouDi_GTest(MinimalRouDiConfigBuilder().payloadChunkSize(4294967296 + 100).payloadChunkCount(2).create());
```

I would test both cases, one with the typed API and one with the untyped API. If this is too complex or time-consuming, I would also accept just a test with one of the APIs.

iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
```diff
@@ -35,12 +35,12 @@ struct MePooConfig
     struct Entry
     {
         /// @brief set the size and count of memory chunks
-        Entry(uint32_t f_size, uint32_t f_chunkCount) noexcept
+        Entry(uint64_t f_size, uint32_t f_chunkCount) noexcept
```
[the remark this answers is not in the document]
gpalmer-latai: Why is it `f` anyway? I would think that that meant "floating point".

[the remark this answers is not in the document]
mossmaurice: @gpalmer-latai You are right. That's due to historical reasons using an old coding style.

iceoryx_posh/source/mepoo/mem_pool.cpp
gpalmer-latai: I'd also like to see an explicit IOX_EXPECTS verifying that `chunkSize` * `numChunks` isn't somehow greater than `numeric_limits<uint64_t>::max`. Such a check would have helped previously when I discovered the `numeric_limits<uint64_t>::max` overflow bug.

iceoryx_posh/source/mepoo/mem_pool.cpp
```diff
@@ -101,14 +101,14 @@ void* MemPool::getChunk() noexcept
     return indexToPointer(index, m_chunkSize, m_rawMemory.get());
 }
 
-void* MemPool::indexToPointer(uint32_t index, uint32_t chunkSize, void* const rawMemoryBase) noexcept
+void* MemPool::indexToPointer(uint32_t index, uint64_t chunkSize, void* const rawMemoryBase) noexcept
 {
     const auto offset = static_cast<uint64_t>(index) * chunkSize;
     return static_cast<void*>(static_cast<uint8_t*>(rawMemoryBase) + offset);
```
gpalmer-latai, line 107: This is more of a comment on the previous change to fix overflow for memory pools spanning more than 4GB. 

Do we have to care about 32-bit operating systems? I.e. should we be using `size_t` and validating while parsing the config and via a few `IOX_ENSURES` statements that we don't overflow `numeric_limits<size_t>::max`? 

Otherwise the pointer arithmetic will simply overflow on those platforms and probably segfault.

gpalmer-latai: In any case I wouldn't require any action in this particular PR. It should be a separate issue.

iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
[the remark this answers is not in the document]
elBoberido: @gpalmer-latai it was the coding style of we well known German automotive supplier and `f` means `function parameter`. There was also a `l_` prefix for local variables.

iceoryx_posh/source/mepoo/mem_pool.cpp
elBoberido: Well, `chunkSize * numChunks` will never be larger than `numeric_limits<uint64_t>::max` ;)

We have to check whether `numeric_limits<uint64_t>::max / numChunks` is larger or equal to `chunkSize`.

Side note, it is now called `IOX_ENFORCE` :)

I would avoid to panic here, an incorrect size is not different from an incorrect alignment, which also does not panic. Let's keep doing it like the local error handling on the class instead of mixing it. It makes testing also a little bit easier.

iceoryx_posh/source/mepoo/chunk_settings.cpp
elBoberido: I would avoid to panic here, an incorrect size is not different from an incorrect alignment, which also does not panic. Let's keep doing it like the local error handling on the class instead of mixing it. It makes testing also a little bit easier.

iceoryx_posh/source/mepoo/mem_pool.cpp
elBoberido: These methods could also easily become private. The only reason they are public is because it was easier to test.

iceoryx_posh/source/mepoo/memory_manager.cpp
```diff
@@ -58,7 +58,7 @@ void MemoryManager::addMemPool(BumpAllocator& managementAllocator,
                 return log;
             } << "These mempools must be added in an increasing chunk size ordering. The newly added  MemPool [ "
                  "ChunkSize = "
-              << adjustedChunkSize << ", ChunkPayloadSize = " << static_cast<uint32_t>(chunkPayloadSize)
+              << adjustedChunkSize << ", ChunkPayloadSize = " << static_cast<uint64_t>(chunkPayloadSize)
               << ", ChunkCount = " << static_cast<uint32_t>(numberOfChunks) << "] breaks that requirement!");
```
elBoberido, line 62: Small nitpick. The casts can be removed. There once was a bug in the `LogStream` implementation which was fixed a while back.

No hard requirement for a merge since this was already there, just a nice to have :)

iceoryx_posh/test/moduletests/test_mepoo_chunk_settings.cpp
```diff
@@ -115,61 +115,12 @@ TEST(ChunkSettings_test, CallingRequiredChunkSizeReturnsCorrectValue)
 
 // END GETTER METHOD TESTS
 
-// BEGIN EXCEEDING CHUNK SIZE TESTS
-
-TEST(ChunkSettings_test, NoCustomUserPayloadAlignmentAndTooLargeUserPayload_Fails)
```
elBoberido, line 120: These tests must be added again. Wrong input parameter can still lead to an overflow.

elBoberido (commented):
Can you please also update `chunk_header.md` regarding the uint32_t->uint64_t change

iceoryx_posh/include/iceoryx_posh/mepoo/mepoo_config.hpp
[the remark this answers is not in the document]
kozakusek: Does the `m_` prefix, that I've noticed being used in many places, have the same history?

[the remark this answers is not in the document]
gpalmer-latai: I would assume `m_` means member. I think this one is intended to stay? Unless y'all want to adopt the trailing underscore pattern 😉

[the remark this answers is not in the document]
elBoberido: Yes, `m_` stays. We need some pattern for the members :)

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
```diff
@@ -713,4 +743,36 @@ TEST_F(PublisherSubscriberCommunication_test, PublisherUniqueIdMatchesReceivedSa
...
+                         }
+                         sample.publish();
+                     })
+                     .has_error());
+
+    EXPECT_FALSE(subscriber->take()
+                     .and_then([](auto& sample) {
+                         for (uint64_t i = 4242; i < 5353; ++i)
+                         {
+                             EXPECT_THAT(sample->complexType.bigPayload[i], Eq(static_cast<uint8_t>(i % 256U)));
```
elBoberido, line 770: Please use `ASSERT_THAT` to break out of the for loop after the first failed check.

elBoberido (commented):
Looks good. Maybe a few performance improvements for the new tests but they are not that important since we don't run them in the CI.

More important though would be to add the new option to the cmake options. This shouldn't take much time and you can just copy&paste from an existing options.

iceoryx_posh/test/integrationtests/test_client_server.cpp
```diff
@@ -450,4 +482,127 @@ TEST_F(ClientServer_test, ClientTakesResponseUnblocksServerSendingResponse)
...
+    // send request
+    {
+        auto loanResult = client.loan();
+        ASSERT_FALSE(loanResult.has_error());
+        auto& request = loanResult.value();
+        request.getRequestHeader().setSequenceId(SEQUENCE_ID);
+        for (uint64_t i = START; i < END; ++i)
+        {
+            request->bigPayload[i] = static_cast<uint8_t>(i % 256U);
+        }
```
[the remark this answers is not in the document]
kozakusek: Yeah.
Doing it for the whole payload took a lot of time, that is why I reduced it to a smaller part.
However your idea with writing every 4096th element (or even 2048th) makes a lot more sense

iceoryx_posh/test/integrationtests/test_client_server.cpp
```diff
@@ -450,4 +482,127 @@ TEST_F(ClientServer_test, ClientTakesResponseUnblocksServerSendingResponse)
     EXPECT_THAT(wasResponseSent.load(), Eq(true));
 }
 
+#ifdef RUN_BIG_PAYLOAD_TESTS
+
+TEST_F(BigPayloadClientServer_test, TypedApiWithBigPayloadWithMatchingOptionsWorks)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "9838d2dc-bd87-42aa-b581-a9526e35e46a");
+
```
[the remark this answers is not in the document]
kozakusek: Wouldn't it still cause BigPayloadClientServer_test to be constructed?
This leads to the machine running out of memory either way.
At least IIUC this is what happened in CI when I tried to use `GTEST_SKIP()`

iceoryx_posh/test/integrationtests/test_publisher_subscriber_communication.cpp
kozakusek: Oh yeah.
Missed that macro.
Makes a lot more sense than having an if with break

iceoryx_posh/test/integrationtests/test_client_server.cpp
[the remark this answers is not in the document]
elBoberido: Yes, you are right. I totally overlooked that. Then lets keep it as is but only add the option to `iceoryx_meta/build_options.cmake`

elBoberido (approved):

elBoberido:
@elfenpiff @gpalmer-latai please also have a final look

@kozakusek looks good. Thanks for the contributions. Btw, there is a [developer meetup](https://github.com/eclipse-iceoryx/iceoryx/wiki/Developer-meetup) on each first Tuesday of the month. Would be great to hear more about your use case.

gpalmer-latai:
It seems like it would be a good idea for you to take some credit for your contribution by updating the copyright headers: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/source/mepoo/mem_pool.cpp#L3

Basically put your individual name as @elBoberido has if you signed the eclipse agreement as an individual contributor, or if you are doing this for work purposes and your company cares to claim your IP, you can put their name as well.

iceoryx_posh/source/mepoo/chunk_settings.cpp
```diff
@@ -63,41 +63,50 @@ expected<ChunkSettings, ChunkSettings::Error> ChunkSettings::create(const uint32
...
     // have a look at »Required Chunk Size Calculation« in chunk_header.md for more details regarding the calculation
     if (userHeaderSize == 0)
     {
         // the most simple case with no user-header and the user-payload adjacent to the ChunkHeader
         if (userPayloadAlignment <= alignof(mepoo::ChunkHeader))
         {
+            if (userPayloadSize > std::numeric_limits<uint64_t>::max() - sizeof(ChunkHeader))
+            {
+                return err(ChunkSettings::Error::REQUIRED_CHUNK_SIZE_EXCEEDS_MAX_CHUNK_SIZE);
+            }
```
gpalmer-latai, line 90: Since you've added this (unlikely) error case, it'd probably be good to add a unit test as well that calls the function with exactly the bad threshold size and asserts that an error is returned.

gpalmer-latai: Same for the other error case.

iceoryx_posh/source/mepoo/mem_pool.cpp
gpalmer-latai: Let me rephrase - I'd like a check that the multiplication will not overflow. There is pseudocode for this in this stackoverflow: https://stackoverflow.com/a/1815391

gpalmer-latai: And it looks like the check below does indeed implement the check correctly 👍

iceoryx_posh/source/mepoo/mem_pool.cpp
```diff
@@ -49,6 +49,7 @@ MemPool::MemPool(const greater_or_equal<uint32_t, CHUNK_MEMORY_ALIGNMENT> chunkS
 {
     if (isMultipleOfAlignment(chunkSize))
     {
+        IOX_EXPECTS(m_chunkSize <= std::numeric_limits<uint64_t>::max() / m_numberOfChunks);
```
gpalmer-latai, line 52: Nice, thank you.

Very small nit here and in other places. You can use `IOX_EXPECTS_WITH_MSG` to add an explanation of what the failure of this check implies. Like "Requested size of memory pool is greater than the maximum of uint64_t::max()"

gpalmer-latai (approved):
Left a few suggestions about more extension testing/error messaging, but overall this looks great to me.

iceoryx_posh/source/mepoo/chunk_settings.cpp
elBoberido: If I did not overlook something, there are already test cases for all of the possible errors. It just happend that the checks are now partially at a different location. At least the coverage of these files is 100%

iceoryx_posh/source/mepoo/mem_pool.cpp
elBoberido: Oh yes, please. This will make my life easier wenn porting to the new macros which require always an error string

iceoryx_posh/source/mepoo/chunk_settings.cpp
gpalmer-latai: Ah okay. I guess I didn't look hard enough and assumed because there wasn't a new unit test calling this function directly and exercising the expected value, it might not be checked. 

But you are right - this is the same error that existed before.

iceoryx_posh/test/integrationtests/test_client_server.cpp
```diff
@@ -450,4 +482,121 @@ TEST_F(ClientServer_test, ClientTakesResponseUnblocksServerSendingResponse)
...
+    // send request
+    {
+        auto loanResult = client.loan();
+        ASSERT_FALSE(loanResult.has_error());
+        auto& request = loanResult.value();
+        request.getRequestHeader().setSequenceId(SEQUENCE_ID);
+        for (uint64_t i = FIRST; i <= LAST; i += STEP)
+        {
+            request->bigPayload[i] = static_cast<uint8_t>((i / STEP) % 256U);
```
elfenpiff, line 508: (optional) To give more context understanding I would use here `std::numeric_limits<uint8_t>::max()`. Maybe also add a little type alias to `BigPayloadStruct` and use it here.

iceoryx_posh/test/integrationtests/test_client_server.cpp
```diff
@@ -450,4 +482,121 @@ TEST_F(ClientServer_test, ClientTakesResponseUnblocksServerSendingResponse)
...
+
+    // take response
+    {
+        auto takeResult = client.take();
+        ASSERT_FALSE(takeResult.has_error());
+        auto& response = takeResult.value();
+        EXPECT_THAT(response.getResponseHeader().getSequenceId(), Eq(SEQUENCE_ID));
+        for (uint64_t i = FIRST; i <= LAST; i += STEP)
+        {
+            ASSERT_THAT(response->bigPayload[i], Eq(static_cast<uint8_t>(((i / STEP) % 256U) + SHIFT)));
```
elfenpiff, line 537: (optional) Same here, use numeric limits.

iceoryx_posh/test/integrationtests/test_client_server.cpp
```diff
@@ -35,6 +35,8 @@ using namespace iox::capro;
 using namespace iox::runtime;
 using namespace iox::roudi_env;
 
+constexpr uint64_t BIG_PAYLOAD_SIZE = std::numeric_limits<uint32_t>::max() + 105UL;
```
elfenpiff, line 38: (optional) I would rename the variable into `PAYLOAD_LARGER_4GB` so that it becomes clearer that we are testing here a payload that exceeds 32-bit.

elfenpiff (approved):
I have some remarks, but they are purely optional. If you have to touch the PR again, it would be nice if you would incorporate them, otherwise just merge the PR.

iceoryx_posh/test/integrationtests/test_client_server.cpp
```diff
@@ -82,6 +95,25 @@ class ClientServer_test : public RouDi_GTest
 };
 constexpr iox::units::Duration ClientServer_test::DEADLOCK_TIMEOUT;
 
+class BigPayloadClientServer_test : public ClientServer_test
+{
+  public:
+    BigPayloadClientServer_test()
+        : ClientServer_test(
+            MinimalRouDiConfigBuilder().payloadChunkSize(BIG_PAYLOAD_SIZE + 128).payloadChunkCount(2).create())
```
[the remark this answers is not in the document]
elBoberido: No, this is just an arbitrary number to make sure the payload is larger than 4GB.

iceoryx_posh/test/integrationtests/test_client_server.cpp
elBoberido: `std::numeric_limits<uint8_t>::max().` won't work but yeah, a constexpr for this magic number would be nice. But on the other hand, I think this is not even necessary since the cast to uint8_t does already what wants to be achieved by `%256`.

iceoryx_posh/test/integrationtests/test_client_server.cpp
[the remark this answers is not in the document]
elBoberido: Btw, could you increase the size of the payload by at least the step size in the for loop. This is just to ensure to write beyond the 4GB mark.

iceoryx_posh/source/mepoo/mem_pool.cpp
kozakusek: Sure
But what do you mean by other places?
I couldn't find any other places around the touched code that would use IOX_EXPECTS and would benefit from the message.
Could you point me to the code you had in mind?

elBoberido: Don't worry, if there is a place without a message it will be added when I port the code the the new macros, which I would start once this PR is merged.

iceoryx_posh/test/integrationtests/test_client_server.cpp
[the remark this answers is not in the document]
kozakusek: > No, this is just an arbitrary number to make sure the payload is larger than 4GB.

Actually this number is included in the constant `BIG_PAYLOAD_SIZE`
This might have been an arbitrary offset that should manage to fit the header

kozakusek:
> It seems like it would be a good idea for you to take some credit for your contribution by updating the copyright headers: https://github.com/eclipse-iceoryx/iceoryx/blob/master/iceoryx_posh/source/mepoo/mem_pool.cpp#L3
> 
> Basically put your individual name as @elBoberido has if you signed the eclipse agreement as an individual contributor, or if you are doing this for work purposes and your company cares to claim your IP, you can put their name as well.

Well
Why not
I'll add it in the few files that I've done something more than just change 32 to 64 :D

iceoryx_posh/test/integrationtests/test_client_server.cpp
kozakusek: That's what I wanted to say. 
This modulo is unnecessary here.

I am not sure what would you expect the alias to be.
`uint8_t` is precisely what I needed here and what I am used to use when I refer to 'just some bytes in memory'.

elBoberido:
> Well
> Why not
> I'll add it in the few files that I've done something more than just change 32 to 64 :D

That' s also how I handle it. It must be something more complex than something a trained ferret could have done :laughing:

iceoryx_posh/test/integrationtests/test_client_server.cpp
kozakusek: And I can get rid of the cast.
Good idea

elBoberido (approved):
Looks good. Every request was implemented. I'm going to merge once the CI is finished. Thanks very much for your contribution and patience. Hope to see you in the developer meetup :)

PR #2168, opened 2024-01-26
iox-#2167 iox-#2156 Update SpscFiFo

### LINKED ISSUE #2156: Bug in lock-free `SpscFifo::size()` & `SpscFifo::is_full()` method
## Required information

Races that lead to random behavior in `SpscFifo::size()` and `SpscFifo::is_full()` method.
The `SpscFifo` has the contract that at most one thread can call `push` and at most one thread can call `pop` concurrently. All other methods can be called without restriction, since they do not modify the fifo.
One reasonable assumption is that `size()` returns in the range of `[0; capacity[` and that the size it returns was valid at some timepoint in the past.

The implementation of the size method looks like:

```cxx
return m_write_pos.load(std::memory_order_relaxed) - m_read_pos.load(std::memory_order_relaxed);
```

In the background, C++ will call the free function `uint64_t operator-(uint64_t lhs, uint64_t rhs);`. The evaluation order of the input arguments of the function can be freely chosen by the compiler.

Let's assume the following scenario.
1. In the beginning, the value of `m_write_pos == 5` and `m_read_pos == 3`
2. The user calls `SpscFifo::size()` from thread `A`. After the value of `m_write_pos.load(..)` - (**5**) - is acquired the OS interrupts the method.
3. Now, thread `B` and thread `C` are performing multiple `push` and `pop` operations according to the contract.
The new state after the operations is `m_write_pos == 8` and `m_read_pos == 6`.
4. The OS schedules thread `A` again, and it continues by reading the value of `m_read_pos.load(..)` - (**6**) -. Now the subtraction is performed `5 - 6 == -1` which lead to an underflow. The underflow of an unsigned integer is well-defined, and the result is `2^64`.
5. The `size()` method returns `2^64`.

Result: the `size()` method returns garbage values in a concurrent scenario.

`is_full` method has the same problem, but the severity should not be so severe since it returns a bool, but it can still happen that `is_full` returns a state that never existed in the past. If an algorithm is requiring that at least at some point in time the state was valid, it will fail.

## Fix

Update the algorithm to

```cxx
uint64_t write_pos = 0;
uint64_t read_pos = 0;
uint64_t size = 0;
do {
  // load current position value
  write_pos = m_write_pos.load(std::memory_order_relaxed);
  read_pos = m_read_pos.load(std::memory_order_relaxed);
  size = write_pos - read_pos;
  // ensure that the values did not change during the size calculation
} while (write_pos != m_write_pos.load(std::memory_order_relaxed) || read_pos != m_read_pos.load(std::memory_order_relaxed))
```

### LINKED ISSUE #2167: Update SPSCFiFo
## Brief feature description

The goal of this issue is to improve the documentation of the FiFo as well as to fix potential wrong memory orders

## Detailed information

- better understanding of the memory orders
- fix potential data races
- relax memory orders when possible

albtam:

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## Notes for Reviewer

See discussion

## References

- Closes #2167 
- Closes #2156

### DIFF
--- a/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
+++ b/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
@@ -27,37 +27,47 @@ namespace iox
 {
 namespace concurrent
 {
-/// @brief single pusher single pop'er thread safe fifo
+/// @brief single producer single consumer thread safe fifo
+/// @note there are at most one push and one pop threads that can work concurrently on the fifo
 template <typename ValueType, uint64_t Capacity>
 class SpscFifo
 {
   public:
     /// @brief pushes a value into the fifo
-    /// @return if the values was pushed successfully into the fifo it returns
+    /// @note restricted thread-safe: can be called from different threads but not concurrently (i.e.
+    /// simultaneously)
+    /// @return if the value was pushed successfully into the fifo, returns
     ///         true, otherwise false
     bool push(const ValueType& value) noexcept;
 
     /// @brief returns the oldest value from the fifo and removes it
-    /// @return if the fifo was not empty the optional contains the value,
+    /// @note restricted thread-safe: can be called from different threads but not concurrently (i.e.
+    /// simultaneously)
+    /// @return if the fifo was not empty, the optional contains the value,
     ///         otherwise it contains a nullopt
     optional<ValueType> pop() noexcept;
 
     /// @brief returns true when the fifo is empty, otherwise false
+    /// @note thread safe (the result might already be outdated when used)
     bool empty() const noexcept;
 
     /// @brief returns the size of the fifo
+    /// @note thread safe (the result might already be outdated when used)
     uint64_t size() const noexcept;
 
     /// @brief returns the capacity of the fifo
     static constexpr uint64_t capacity() noexcept;
 
   private:
+    // thread safe (the result might already be outdated when used)
     bool is_full() const noexcept;
+    std::pair<uint64_t, uint64_t> get_read_write_positions() const noexcept;
+
 
   private:
     UninitializedArray<ValueType, Capacity> m_data;
-    std::atomic<uint64_t> m_write_pos{0};
-    std::atomic<uint64_t> m_read_pos{0};
+    std::atomic<uint64_t> m_writePos{0};
+    std::atomic<uint64_t> m_readPos{0};
 };
 
 } // namespace concurrent
--- a/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
+++ b/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
@@ -28,41 +28,40 @@ namespace iox
 namespace concurrent
 {
 /// @brief single producer single consumer thread safe fifo
-/// @note there are at most one push and one pop threads that can work concurrently on the fifo
+/// @note there is only one push and one pop thread that can work on the FIFO without additional synchronization
+/// mechanisms like transferring the authorization to push to a third thread
 template <typename ValueType, uint64_t Capacity>
 class SpscFifo
 {
   public:
     /// @brief pushes a value into the fifo
-    /// @note restricted thread-safe: can be called from different threads but not concurrently (i.e.
-    /// simultaneously)
+    /// @note restricted thread-safe: can only be accessed from one thread. The authorization to push into the FIFO can
+    /// be transferred to another thread if appropriate synchronization mechanisms are used.
     /// @return if the value was pushed successfully into the fifo, returns
     ///         true, otherwise false
     bool push(const ValueType& value) noexcept;
 
     /// @brief returns the oldest value from the fifo and removes it
-    /// @note restricted thread-safe: can be called from different threads but not concurrently (i.e.
-    /// simultaneously)
+    /// @note restricted thread-safe: can only be accessed from one thread. The authorization to pop into the FIFO can
+    /// be transferred to another thread if appropriate synchronization mechanisms are used.
     /// @return if the fifo was not empty, the optional contains the value,
     ///         otherwise it contains a nullopt
     optional<ValueType> pop() noexcept;
 
     /// @brief returns true when the fifo is empty, otherwise false
-    /// @note thread safe (the result might already be outdated when used)
+    /// @note thread safe (the result might already be outdated when used). Expected to be called from either the push or the pop thread but not from a third thread
     bool empty() const noexcept;
 
     /// @brief returns the size of the fifo
-    /// @note thread safe (the result might already be outdated when used)
+    /// @note thread safe (the result might already be outdated when used). Expected to be called from either the push or the pop thread but not from a third thread
     uint64_t size() const noexcept;
 
     /// @brief returns the capacity of the fifo
     static constexpr uint64_t capacity() noexcept;
 
   private:
-    // thread safe (the result might already be outdated when used)
+    // thread safe (the result might already be outdated when used). Expected to be called from either the push or the pop thread but not from a third thread
     bool is_full() const noexcept;
-    std::pair<uint64_t, uint64_t> get_read_write_positions() const noexcept;
-
 
   private:
     UninitializedArray<ValueType, Capacity> m_data;
--- a/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
+++ b/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
@@ -42,25 +42,28 @@ class SpscFifo
     bool push(const ValueType& value) noexcept;
 
     /// @brief returns the oldest value from the fifo and removes it
-    /// @note restricted thread-safe: can only be accessed from one thread. The authorization to pop into the FIFO can
+    /// @note restricted thread-safe: can only be accessed from one thread. The authorization to pop from the FIFO can
     /// be transferred to another thread if appropriate synchronization mechanisms are used.
     /// @return if the fifo was not empty, the optional contains the value,
     ///         otherwise it contains a nullopt
     optional<ValueType> pop() noexcept;
 
     /// @brief returns true when the fifo is empty, otherwise false
-    /// @note thread safe (the result might already be outdated when used). Expected to be called from either the push or the pop thread but not from a third thread
+    /// @note thread safe (the result might already be outdated when used). Expected to be called from either the push
+    /// or the pop thread but not from a third thread
     bool empty() const noexcept;
 
     /// @brief returns the size of the fifo
-    /// @note thread safe (the result might already be outdated when used). Expected to be called from either the push or the pop thread but not from a third thread
+    /// @note thread safe (the result might already be outdated when used). Expected to be called from either the push
+    /// or the pop thread but not from a third thread
     uint64_t size() const noexcept;
 
     /// @brief returns the capacity of the fifo
     static constexpr uint64_t capacity() noexcept;
 
   private:
-    // thread safe (the result might already be outdated when used). Expected to be called from either the push or the pop thread but not from a third thread
+    // thread safe (the result might already be outdated when used). Expected to be called from either the push or the
+    // pop thread but not from a third thread
     bool is_full() const noexcept;
 
   private:
--- a/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
+++ b/iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
@@ -27,14 +27,17 @@ namespace concurrent
 template <class ValueType, uint64_t Capacity>
 inline bool SpscFifo<ValueType, Capacity>::push(const ValueType& value) noexcept
 {
-    if (is_full())
-    {
-        return false;
-    }
     // Memory order relaxed is enough since:
     // - there is no concurrent access to this method
     // - the load statement cannot be reordered with writing m_data otherwise it would not compile
     auto currentWritePos = m_writePos.load(std::memory_order_relaxed);
+    bool isFull = (currentWritePos ==
+                   // Memory order acquire is needed to ensure happens-before relationship
+                   m_readPos.load(std::memory_order_acquire) + Capacity);
+    if (isFull)
+    {
+        return false;
+    }
     m_data[currentWritePos % Capacity] = value;
 
     // SYNC POINT WRITE: m_data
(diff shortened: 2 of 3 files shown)


elfenpiff (commented):
@albtam Could you please add an entry to the release notes and then we can merge it.

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
```diff
@@ -27,37 +27,47 @@ namespace iox
...
     bool is_full() const noexcept;
+    std::pair<uint64_t, uint64_t> get_read_write_positions() const noexcept;
+
 
   private:
     UninitializedArray<ValueType, Capacity> m_data;
-    std::atomic<uint64_t> m_write_pos{0};
-    std::atomic<uint64_t> m_read_pos{0};
+    std::atomic<uint64_t> m_writePos{0};
+    std::atomic<uint64_t> m_readPos{0};
```
elBoberido, line 70: Oh no, in the last developer meetup we decided to use move to snake case for members :sweat_smile: 

Well can be changed later on back. Good to have everything in this class following the same coding style :)

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
```diff
@@ -31,27 +31,69 @@ inline bool SpscFifo<ValueType, Capacity>::push(const ValueType& value) noexcept
...
-    m_write_pos.store(currentWritePos + 1, std::memory_order_release);
+    // SYNC POINT WRITE: m_data
+    // We need to make sure that writing the value happens before incrementing the
+    // m_writePos otherwise the following scenario can happen:
+    // 1. m_writePos is increased (but the value has not been written yet)
+    // 2. Another thread calls pop(): we check if the queue is empty => no
+    // 3. In pop(), when we read a value, a data race can occur when at the same time a value is
+    // written by push. With memory_order_release, this cannot happen as it is guaranteed that
+    // writing the data happens before incrementing m_writePos. Note that the following scenario
+    // can still happen (but this is not a problem):
```
elBoberido, line 48: Indeed. This is an inherent race with all concurrent algorithms. It depends on when you look at something and in a blink of an eye this information might be outdated.

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
```diff
@@ -31,27 +31,69 @@ inline bool SpscFifo<ValueType, Capacity>::push(const ValueType& value) noexcept
...
+        // 2. The OS puts the thread A to sleep
+        // 3. Other threads continue to call push and pop
+        // 4. New state m_writePos == 8, m_readPos == 6
+        // 5. The OS schedules again thread A and we read m_readPos = 6
+        // 6. We calculate 5 - 6 == -1  => overflow in uint64_t and size() returns
+        // 18446744073709551615
+        // 7. Another algorithm may depend on the reasonable argument that the size is always less or
+        // equal capacity it may lead to weird bugs.
+        // Note: it is still possible to return a size that is not up-to-date anymore but at least
+        // this size is logically valid
```
elBoberido, line 92: I guess this is related to #2156 where we already concluded that this would be a misuse of the `SpscFifo`. I've the feeling you need a `MpmcFifo`. Do you have a use case for calling `size`, `empty` or `is_full` from another thread than the producer or consumer thread?

Note, you **cannot** rely on the `size` from any other thread than the producer or consumer thread
- the producer knows that the size will not increase but cannot tell whether is will be smaller right after the call returns
- the consumer knows that the size will not decrease but cannot tell whether it will be larger right after the call returns
- a third thread has absolutely no guarantees; it only knows that at some point in time it had this size but cannot tell whether it will be smaller or larger right after the call returns; the information is essentially useless; if you rely on this information you need to re-design your algorithm

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
```diff
@@ -63,27 +105,50 @@ inline constexpr uint64_t SpscFifo<ValueType, Capacity>::capacity() noexcept
...
+    // (e.g. when the queue is full, i.e. m_readPos == m_writePos)
+    // Note that the following situation can still happen but is not a problem:
+    // 1. We read the value
+    // 2. Another thread calls push(): we check if the queue is full => yes
+    // 3. The queue was not really full as the value was already read
+    // Memory order acq_rel to avoid:
+    // - reading m_data to happen after the fence (rel)
+    // - incrementing m_readPos before the fence (acq)
+    std::atomic_thread_fence(std::memory_order::memory_order_acq_rel);
+    m_readPos.store(current_readPos + 1, std::memory_order_relaxed);
```
elBoberido, line 151: I fail to see why this is required. Where is the race in the `push` method?

Are you confusing this with the SOFI?

elBoberido (commented):
Thanks for the documentation update. For some of the code changes we need some use cases as it looks like you want to make this kind of a `MpmcFifo`

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.hpp
albtam: I can change it but I guess the whole repo can be linted with clang-tidy at once

elBoberido: No problem, we change it later all at once. It's good to have the class follow one style.

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
albtam: Ok, thanks for the explanation. It makes sense. I reverted the changes

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
albtam: Initial situation: the queue is full
```
|--1--|--2--|
^
r=0,  w=2
```

We increment the read counter but haven't read the value yet (operation reordering)
```
|--1--|--2--|
^     ^
w=2  r=1
```

The pop thread is paused and push is called in another thread. The queue is not full so the push thread writes a new value:
```
|--3--|--2--|
      ^
   r=1, w=3
```

Now the pop thread resumes and reads the value at the previously stored index 0 => 3 (1 was never read)

albtam: The idea here was to say:

1. there is no memory synchronization needed here (we are just reading) BUT we need to avoid operation reordering
2. We make the memory order `relaxed` to show that there is no memory synchronization
3. Since we still need to prevent operation reordering, we use an atomic fence

elBoberido: Well, it seems Miri agrees with me https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=33665814b2deb0ef8b42cb7fe4255a71

Like already mentioned in another comment, adding either a release ordering on line 72 in the new link or a fence with acquire semantics on line 74 will fix the problem. Instead of the fences I would rather use the atomics as proposed in https://github.com/eclipse-iceoryx/iceoryx/pull/2168#discussion_r1474886638. I'm also fine with keeping the memory order as is and I create a separate PR to fix the issue.

Fences do not work like atomics and have subtle differences. Each time I work with fences I refresh my memory with these blog posts from Jeff Preshing.
- https://preshing.com/20130922/acquire-and-release-fences/
- https://preshing.com/20131125/acquire-and-release-fences-dont-work-the-way-youd-expect/
- https://preshing.com/20150402/you-can-do-any-kind-of-atomic-read-modify-write-operation/
- https://preshing.com/20170612/can-reordering-of-release-acquire-operations-introduce-deadlock/

I would need to dig in again in order to explain you exactly why this code does not work.

albtam: Thanks for the links. I think there is a misunderstanding. Here is the behavior I observed:

| line 72: read_pos.load(order) | line 113: read_pos.store(order) | miri report |
|-------------------------------|---------------------------------|-------------|
| Relaxed                       | Release                         | race        |
| Relaxed                       | Relaxed + fence(Release)        | race        |
| Acquire                       | Relaxed                         | race        |
| Acquire                       | Release                         | no race     |
| Acquire                       | Relaxed + fence(Release)        | no race     |

So it looks like the memory fence is ok

elBoberido: @albtam yes, that's what I meant. Keeping the current memory order and just adding an `Acquire` also fixes the race without the need of a fence. The fence would in this case just make the reasoning harder.

elBoberido: Just to be sure to prevent a misuderstanding.
Current code
```cpp
push(...)
{
    m_readPos.load(relaxed); // via is_full
    // doing stuff
    m_writePos.store(release);
}
pop(...)
{
    m_readPos.load(acquire);
    m_writePos.load(acquire);
    // doing stuff
    m_readPos.store(release);
}
```

Your initial proposal
```cpp
push(...)
{
    m_readPos.load(relaxed); // via is_full
    // doing stuff
    m_writePos.store(release);
}
pop(...)
{
    m_readPos.load(relaxed);
    m_writePos.load(acquire);
    // doing stuff
    fence(acq_rel);
    m_readPos.store(relaxed);
}
```

Your updated proposal
```cpp
push(...)
{
    m_readPos.load(acquire);
    // doing stuff
    m_writePos.store(release);
}
pop(...)
{
    m_readPos.load(relaxed);
    m_writePos.load(acquire);
    // doing stuff
    fence(acq_rel);
    m_readPos.store(relaxed);
}
```

My proposal
```cpp
push(...)
{
    m_readPos.load(acquire);
    // doing stuff
    m_writePos.store(release);
}
pop(...)
{
    m_readPos.load(relaxed);
    m_writePos.load(acquire);
    // doing stuff
    m_readPos.store(release);
}
```

albtam: > The fence would in this case just make the reasoning harder.

I got your point since fences might not work as you expect. My reasoning is the following: when you use a fence, you only want to avoid operation reordering (but you don't need to sync the memory). When you specify the memory order in the load/store operation, you want (either or both) to prevent operation reordering and sync the memory. That's why I believe using fences makes the intention clearer 

Note: I actually edited my comment but you were too fast to answer, my final proposal would be:

```cpp
push(...)
{
    m_readPos.load(relaxed);
    if is_full {
      return false;
   }
   // only at this point we need to prevent operation reordering
  fence(acquire);


pop(...)
{
    m_readPos.load(relaxed);
    m_writePos.load(acquire);
    // doing stuff
    fence(release)
    m_readPos.store(relaxed);
}
```

That being said, I would go with your proposal, since you're the maintainer ;)

albtam: As for explaining why the original code was wrong, here is my interpretation:

even if you have a release/store (line 113), it doesn't do anything (i.e. prevent operation rendering) if there is no corresponding load/acquire. Do you agree?

elBoberido: Here I would go with the save option and don't use fences as long as we do not have an extensive test suite and also run that suite on a weakly ordered CPU. As already mentioned, fences work differently than atomics and one can introduce subtle bugs. With the atomics it is simpler to reason about what happens and the fences can be added later once we are sure it does not break the code, not even on edge cases.

elBoberido: ... also, keep in mind that fences also **do synchronize** memory.

elBoberido: @albtam I think I realize more and more where the confusion came from. You are right that in theory you do not need to synchronize the memory back to the push thread but you need to establish a happens before relationship. This can either be done by an atomic store with release ordering in one thread and an atomic load with acquire ordering in the second thread or with two fences and an atomic with relaxed memory order or a fence and an atomic which has the complementary memory order to the fence. Using just a fence on one thread does not help you since it has no counterpart to establish the happens before relationship. Keep in mind, the only reason you want the ordering is because there is another thread which can observe the ordering and therefore needs to be synced with.

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
elBoberido: Oh, could you change this to ` but, although it is an inherent race with concurrent algorithms, it is not a data race and therefore not a problem.` to make it explicit why it is not a problem.

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
```diff
@@ -27,33 +27,47 @@ namespace concurrent
 template <class ValueType, uint64_t Capacity>
 inline bool SpscFifo<ValueType, Capacity>::push(const ValueType& value) noexcept
 {
-    if (is_full())
+    // Memory order relaxed is enough since:
+    // - there is no concurrent access to this method
+    // - the load statement cannot be reordered with writing m_data otherwise it would not compile
+    auto currentWritePos = m_writePos.load(std::memory_order_relaxed);
+    bool isFull = (currentWritePos ==
+                   // Memory order acquire is needed to ensure happens-before relationship
```
elBoberido, line 35: I think it would be good to also add the corresponding sync point. In this case it would be the `store  with release ordering on m_readPos in the pop method`.

Similar with other non-relaxed memory orderings. It is essential to know the counterparts of the sync operation.

iceoryx_hoofs/concurrent/buffer/include/iox/detail/spsc_fifo.inl
```diff
@@ -27,33 +27,47 @@ namespace concurrent
...
+    // m_writePos otherwise the following scenario can happen:
+    // 1. m_writePos is increased (but the value has not been written yet)
+    // 2. Another thread calls pop(): we check if the queue is empty => no
+    // 3. In pop(), when we read a value, a data race can occur when at the same time a value is
+    // written by push. With memory_order_release, this cannot happen as it is guaranteed that
+    // writing the data happens before incrementing m_writePos. Note that the following scenario
+    // can still happen (but this is not a problem):
+    // 1. A value is written (m_writePos hasn't been incremented yet)
+    // 2. Another thread calls pop(): we check if the queue is empty => yes
+    // 3. An element was already stored so we could have popped the element
```
elBoberido, line 54: I would also rephrase this since it is a bit misleading. Technically the consumer thread cannot read the element, even after the producer thread wrote the data since the producer did not yet synchronized the memory. Stating that the consumer threat could have popped the element is therefore not true. It could only have popped the element if it would have been informed that there is one, everything else is not allowed and leads to a data race.

1. There is an empty queue
2. A push operation is in progress, the value has ben written but 'm_writePos' was not yet advanced
3. The consumer thread performs a pop operation and the check for an empty queue is true resulting in a failed pop
4. The push operation is finished by advancing `m_writePos` and synchronizing the memory
5. The consumer thread missed the chance to pop the element by a blink of an eye

elBoberido (commented):
Sorry by being so picky but in the end it was our fault to not document this properly and now we have to be really careful to use the right wording :)

elBoberido:
@albtam something weird happened with your last force push. It seems the e-mail is now unverified

elBoberido:
@albtam you do not need to close the PR just of the e-mail address :)

Issue #2224, opened 2024-03-13
Can't directly assign `const` underlying value to `iox::optional`


tdouglas-latai:
## Required information

**Operating system:**
Ubuntu 20.04.6 LTS

**Compiler version:**
Clang 11, GCC 9.4.0

**Eclipse iceoryx version:**
`master`

**Observed result or behaviour:**
In the implementation of `iox::optional::operator=` for direct assignment of the underlying value, the `newValue` parameter is forwarded as `T` instead of `U`, which results in a build error when attempting to assign a `const` value.  This can be fixed by changing the two instances of `std::forward<T>` to `std::forward<U>`.

https://github.com/eclipse-iceoryx/iceoryx/blob/1e17b23d50e7ac2243f43969d90e7f5e8b60afca/iceoryx_hoofs/vocabulary/include/iox/detail/optional.inl#L164-L179

**Expected result or behaviour:**
No build error.

**Conditions where it occurred / Performed steps:**

I was able to replicate the error in the `test_vocabulary_optional.cpp` test by adding the following cases:
```
TEST_F(Optional_test, DirectCopyAssignmentWithNoValue)
{
    ::testing::Test::RecordProperty("TEST_ID", "8dddd1c5-e59b-4f3c-9e6c-6fa9ac1daa86");
    iox::optional<TestClass> sut;
    const TestClass value{4711, 1337};

    sut = value;
    ASSERT_THAT(sut.has_value(), Eq(true));
    EXPECT_THAT(sut->value, Eq(4711));
    EXPECT_THAT(sut->secondValue, Eq(1337));
}

TEST_F(Optional_test, DirectCopyAssignmentWithValue)
{
    ::testing::Test::RecordProperty("TEST_ID", "66fa19ab-0a08-48d3-824c-7b259e6f15b0");
    iox::optional<TestClass> sut{TestClass{7474, 33331}};
    const TestClass value{4711, 1337};

    sut = value;
    ASSERT_THAT(sut.has_value(), Eq(true));
    EXPECT_THAT(sut->value, Eq(4711));
    EXPECT_THAT(sut->secondValue, Eq(1337));
}
```
The error occurs when the `TestCase value` is `const` or `constexpr` qualified.

```
[...]

PR #2225, opened 2024-03-13
iox-#2224 Fix const value assignment in optional

tdouglas-latai:
Fixes the forwarding reference type in `iox::optional::operator=` when performing a direct assignment of the underlying value such that assigning `const` values is supported.

1. [ ] ~Add a second reviewer for complex new features or larger refactorings~
1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

## References

- Closes [#2224](https://github.com/eclipse-iceoryx/iceoryx/issues/2224)

### DIFF
--- a/iceoryx_hoofs/vocabulary/include/iox/detail/optional.inl
+++ b/iceoryx_hoofs/vocabulary/include/iox/detail/optional.inl
@@ -169,11 +169,11 @@ optional<T>::operator=(U&& newValue) noexcept
 {
     if (m_hasValue)
     {
-        value() = std::forward<T>(newValue);
+        value() = std::forward<U>(newValue);
     }
     else
     {
-        construct_value(std::forward<T>(newValue));
+        construct_value(std::forward<U>(newValue));
     }
     return *this;
 }
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -125,6 +125,7 @@
 - Fix wrong memory orders in SpscFiFo [#2167](https://github.com/eclipse-iceoryx/iceoryx/issues/2167)
 - Implement missing copy assignment for expected [#2216](https://github.com/eclipse-iceoryx/iceoryx/issues/2216)
 - Add missing type aliases that conform with STL container types [#2220](https://github.com/eclipse-iceoryx/iceoryx/issues/2220)
+- Fix `const` value assignment in `iox::optional` [\#2224](https://github.com/eclipse-iceoryx/iceoryx/issues/2224)
 
 **Refactoring:**
 
--- a/iceoryx_hoofs/test/moduletests/test_vocabulary_optional.cpp
+++ b/iceoryx_hoofs/test/moduletests/test_vocabulary_optional.cpp
@@ -280,6 +280,30 @@ TEST_F(Optional_test, CopyAssignmentFromNoValueToNoValue)
     ASSERT_THAT(sut2.has_value(), Eq(false));
 }
 
+TEST_F(Optional_test, DirectCopyAssignmentWithNoValue)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "8dddd1c5-e59b-4f3c-9e6c-6fa9ac1daa86");
+    iox::optional<TestClass> sut;
+    const TestClass value{4711, 1337};
+
+    sut = value;
+    ASSERT_THAT(sut.has_value(), Eq(true));
+    EXPECT_THAT(sut->value, Eq(4711));
+    EXPECT_THAT(sut->secondValue, Eq(1337));
+}
+
+TEST_F(Optional_test, DirectCopyAssignmentWithValue)
+{
+    ::testing::Test::RecordProperty("TEST_ID", "66fa19ab-0a08-48d3-824c-7b259e6f15b0");
+    iox::optional<TestClass> sut{TestClass{7474, 33331}};
+    TestClass value{4711, 1337};
+
+    sut = value;
+    ASSERT_THAT(sut.has_value(), Eq(true));
+    EXPECT_THAT(sut->value, Eq(4711));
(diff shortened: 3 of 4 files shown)


iceoryx_hoofs/vocabulary/include/iox/detail/optional.inl
```diff
@@ -169,11 +169,11 @@ optional<T>::operator=(U&& newValue) noexcept
 {
     if (m_hasValue)
     {
-        value() = std::forward<T>(newValue);
```
elfenpiff, line 172: I think we have to revert this change. Maybe the solution is to add explicit "copy/move" assignment operators for the underlying value like:

```cxx
// in inl file
template <typename T>
inline optional<T>& optional<T>::operator=(const T& newValue) noexcept
{
    if (m_hasValue)
    {
        value() = newValue;
    }
    else
    {
        construct_value(newValue);
    }
    return *this;
}

template <typename T>
inline optional<T>& optional<T>::operator=(T&& newValue) noexcept
{
    if (m_hasValue)
    {
        value() = std::move(newValue);
    }
    else
    {
        construct_value(std::move(newValue));
    }
    return *this;
}
```

and add the definition to the hpp file:
```cxx
optional& operator=(const T& newValue) noexcept;
optional& operator=(T&& newValue) noexcept;
```

tdouglas-latai: Taking a look at https://github.com/eclipse-iceoryx/iceoryx/blob/1e17b23d50e7ac2243f43969d90e7f5e8b60afca/tools/introspection/source/introspection_app.cpp#L796 where one of the errors comes from, it seems like a side effect of the current behavior is to implicitly move when assigning a non-const reference.

The `std::forward<T>` turns into `T&&`, while `std::forward<U>` with `U = T&` is equivalent to `T&`.  This matches the copy constructor of `Sample` (which is deleted) rather than the move constructor.  It seems like this is not desired behavior (implicitly converting an l-value to r-value reference).  I can fix it here by explicitly doing `memPoolSample = std::move(sample)`, though I'm not sure what implications this might have elsewhere in Iceoryx or for other users.

tdouglas-latai: This was the only file I found with issues when building locally, so I pushed an update to try again.  Let me know how you'd like to proceed if CI passes.

elfenpiff: We should realize this quickly when this was the wrong approach, so I would go for it and test the introspection app afterwards. If it still shows all the mempools we should be fine.

In my opinion, we are with iceoryx at a state with testing and the CI that we can perform fearless refactoring and rely on the CI.

tdouglas-latai: I ran the introspection client with the changes, along with just `iox-roudi` and I see this output: 
image

elfenpiff: Mempool looks exactly like it should. Nice.

tdouglas-latai:
Hmm, it seems like the ASAN build may have crashed or something?
```
2024-03-13T19:15:11.4614668Z [----------] Global test environment tear-down
2024-03-13T19:15:11.4942698Z [==========] 21 tests from 2 test suites ran. (41 ms total)
2024-03-13T19:15:11.4943362Z [  PASSED  ] 21 tests.
2024-03-13T19:15:11.5740157Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5740803Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5741310Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5741823Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5742317Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5742807Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5743292Z AddressSanitizer:DEADLYSIGNAL
2024-03-13T19:15:11.5743782Z AddressSanitizer:DEADLYSIGNAL
```
Is there something I can do about that? The `DEADLYSIGNAL` spew is so much I have to stop the raw log page from loading before it's responsive :sweat_smile:

elfenpiff:
@tdouglas-latai maybe it was a CI hiccup, I will rerun it again. As far as I can see, the STDLIB_tests did run successfully and then the address sanitizer failed. Without any reason ...

tdouglas-latai:
Hmm, seems to have failed in a similar way the second time.  I ran `./tools/ci/build-test-ubuntu-with-sanitizers.sh gcc asan` on my local machine and it seemed to pass without issue.  Maybe rerun a third time?

tdouglas-latai:
@elfenpiff, I tried running those commands and did get some failures during build with the compiler being killed, even got those errors on `master`, but I didn't get the same error as you or CI.  I changed the build jobs to 4 and it passed on my local machine using this branch and `master`.  Do you have any idea from your output if this could be caused by the proposed changes here?

elfenpiff:
@tdouglas-latai I can reproduce it with your branch but not on master. So it is definitely caused by your changes, but I have no idea what the issue here is. Especially, since it is crashing at the end of a successful test run and the test run varies as well, sometimes after the STDLIB tests, sometimes after the hoofs_moduletests.

I tend to say this is some bug within the address sanitizer, but we cannot ignore this, otherwise we have a failing CI on master. We either need to confirm that it is in fact an address sanitizer bug and disable the CI target, or we have to find the issue on our side

tdouglas-latai:
Yes, definitely can't ignore the CI failure, I'll play with it more to try and replicate it locally.

elfenpiff (approved):
The CI issue seems to be gone - most likely some ubuntu problem.

tdouglas-latai:
Oh, sweet, thanks!

PR #2328, opened 2024-08-19
iox-#2327 Replace macro ENUM with IOX_ENUM and CLASS with IOX_CLASS

### LINKED ISSUE #2327: Compilation Conflict with Apache Arrow from ENUM macro
## Required information

**Operating system:**
Ubuntu 20.04 LTS

**Compiler version:**
GCC 12.2.0

**Eclipse iceoryx version:**
v2.90.0

There is a compile-time conflict between this library and [Apache Arrow](https://github.com/apache/arrow).  This is due to the macro definition for `ENUM` in `iceoryx_binding_c/include/iceoryx_binding_c/internal/c2cpp_binding.h`.  In the Apache Arrow library, there is a header that has `ENUM` as a value in one of their enums (see [arrow/cpp/src/parquet/types.h line 78](https://github.com/apache/arrow/blob/9d4dcc903e84732a6e14d61aece1f9a1d096f7c9/cpp/src/parquet/types.h#L78)).  If `c2cpp_binding.h` from this library is included before the other, then we either need to include an `#undef ENUM` before including the next header or else run into a compile-time error.

I expect if you define macros in a public header that it's a unique enough name that it will not conflict with user code or with other third-party code.

```c++
#include <iceoryx_binding_c/types.h>
#include <parquet/types.h>

int main() {
    return 0;
}
```

It is fixed by either reordering includes

```c++
#include <parquet/types.h>
#include <iceoryx_binding_c/types.h>

int main() {
    return 0;
}
```

or by undefining the `ENUM` macro after the include

```c++
#include <iceoryx_binding_c/types.h>
#undef ENUM
#include <parquet/types.h>

int main() {
    return 0;
}
```

Both of these solutions on the user side are unsatisfactory.  It would be really nice if the macros had more specific names such as `ICEORYX_ENUM` and `ICEORYX_CLASS` to avoid name clashes.

mikebentley15:
## Notes for Reviewer

The macros `ENUM` and `CLASS` are very generic names and can potentially cause naming conflicts with user code or other third-party libraries.  These macros are intended to make the headers compatible with both C and C++.  The `ENUM` macro is unnecessary since the `enum` prefix for variable declarations is required in C and optional in C++.  The `CLASS` macro is required to match the Microsoft C++ ABI -- defined as `class` when in C++ mode and `struct` when in C mode.

This PR seeks to solve this by

- Removing the `ENUM` macro, replacing its use with the `enum` keyword.
- Rename `CLASS` macro to `IOX_C_CLASS` to reduce likelihood of naming conflicts
    - Command: `sed -i -e 's/\<CLASS\>/IOX_C_CLASS/g' $(find -type f)`
    - One file was reverted since `CLASS` was in a comment; not referencing the macro.

See issue #2327 for a minimal example to reproduce the issue and existing workarounds.

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/main/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/main/doc/website/concepts/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx/blob/main/doc/website/release-notes/iceoryx-unreleased.md

- [x] Consider a second reviewer for complex new features or larger refactorings

## References

- Closes #2327

### DIFF
--- a/iceoryx_binding_c/include/iceoryx_binding_c/internal/c2cpp_binding.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/internal/c2cpp_binding.h
@@ -1,4 +1,5 @@
 // Copyright (c) 2020 by Robert Bosch GmbH. All rights reserved.
+// Copyright (c) 2024 by Michael Bentley <mikebentley15@gmail.com>. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,16 +22,14 @@
 
 #include <cstdint>
 
-#define CLASS class
-#define ENUM
+#define IOX_C_CLASS class
 
 #else
 
 #include <stdbool.h>
 #include <stdint.h>
 
-#define CLASS struct
-#define ENUM enum
+#define IOX_C_CLASS struct
 
 #endif
 
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -68,6 +68,7 @@
 
 **Bugfixes:**
 
+- Fix compile-time conflict with Apache Arrow [\#2327](https://github.com/eclipse-iceoryx/iceoryx/issues/2327)
 - FreeBSD CI build is broken [\#1338](https://github.com/eclipse-iceoryx/iceoryx/issues/1338)
 - High CPU load in blocked publisher is reduced by introducing smart busy loop waiting (adaptive_wait) [\#1347](https://github.com/eclipse-iceoryx/iceoryx/issues/1347)
 - Compile Error : iceoryx_dds/Mempool.hpp: No such file or directory [\#1364](https://github.com/eclipse-iceoryx/iceoryx/issues/1364)
--- a/iceoryx_binding_c/include/iceoryx_binding_c/client.h
+++ b/iceoryx_binding_c/include/iceoryx_binding_c/client.h
@@ -1,4 +1,5 @@
 // Copyright (c) 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2024 by Michael Bentley <mikebentley15@gmail.com>. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,7 +24,7 @@
 #include "iceoryx_binding_c/types.h"
 
 /// @brief client handle
-typedef CLASS UntypedClient* iox_client_t;
+typedef IOX_C_CLASS UntypedClient* iox_client_t;
 
 /// @brief options to be set for a client
 typedef struct
@@ -38,10 +39,10 @@ typedef struct
     bool connectOnCreate;
 
     /// @brief Sets whether the server blocks when the client response queue is full
-    ENUM iox_QueueFullPolicy responseQueueFullPolicy;
+    enum iox_QueueFullPolicy responseQueueFullPolicy;
 
     /// @brief Sets whether the client blocks when the server request queue is full
-    ENUM iox_ConsumerTooSlowPolicy serverTooSlowPolicy;
+    enum iox_ConsumerTooSlowPolicy serverTooSlowPolicy;
 
     /// @brief this value will be set exclusively by 'iox_client_options_init' and is not supposed to be modified
     /// otherwise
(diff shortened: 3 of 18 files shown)


mikebentley15:
After this review goes through, I may make another PR to redo this change for version v2.90 to integrate with our pipeline.

iceoryx_binding_c/include/iceoryx_binding_c/internal/c2cpp_binding.h
```diff
@@ -21,16 +21,16 @@
...
 
 #include <stdbool.h>
 #include <stdint.h>
 
-#define CLASS struct
-#define ENUM enum
+#define IOX_CLASS struct
+#define IOX_ENUM enum
 
 #endif
```
elBoberido, line 35: I'm not even sure why this is required?

Can you remove this part and the corresponding `#endif`
```cpp
#ifdef __cplusplus

#include <cstdint>

#define CLASS class
#define ENUM

#else
```

We can then run the CI and check if something breaks but I'm quite sure that all of our CI targets should still run. If this is the case, we can remove the defines completely

mikebentley15: It looks like it's the difference between treating the code as C++ versus C.  Do you have parts of your code that need to be compiled in both forms?

In C, if you have an enum, you need to use the `enum` keyword to create an instance, since it's part of the type.

```c
enum Food {
   Apple,
   Cherry
};

enum Food food = Apple;
```

But in C++, the `enum` keyword is optional when specifying the type.

As for `CLASS`, C doesn't have the `class` keyword, only `struct`.  In C++, the `struct` keyword implies public default visibility instead of private default visibility with `class`.

If you no longer have the requirement to compile your headers for both C and C++, then yeah, we can probably get rid of these.

mikebentley15: Enum example in C++: https://godbolt.org/z/P8shnz18j
Enum example in C: https://godbolt.org/z/bc3jMeavG

mikebentley15: Looking closer, I think you're right.  Whether interpreted in C or C++, the

```
#define CLASS struct
#define ENUM enum
```

should work based on how it's used.  The `CLASS` macro is used only to make typedefs.  The `ENUM` macro is only for types.  I'll give it a try, thanks! 👍

mikebentley15: @elBoberido if everything looks good, I'm happy to squash my commits into a single commit and update the PR title and description.

elBoberido: Okay, now I understand why this was done. I'm not sure if it was a good idea but there are typedefs with some C++ classees, e.g.
```C
typedef struct UntypedClient* iox_client_t;
```
The `UntypedClient` is actually a C++ class and if the header is used in a C++ context it needs to be just `typedef UntypedClient* iox_client_t`.

I'm not quite sure why the compiler does not complain since `UntypedClient` is in a C++ namespace.

In order to get rid of the defines, one would need to create something like this
```C
struct iox_client;
typedef struct iox_client* iox_client_t;
```

And the always do a reinterpret_cast between the `iox_client_t` and the `UntypedClient*`.

I would suggest to keep the current approach but name the define `IOX_C_ENUM` and `IOX_C_CLASS`.

mikebentley15: I tested and found that it's fine on Linux using GCC.  I didn't know about problems with the Microsoft ABI:

> error: struct 'UntypedClient' was previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI

That's good to know.  The `enum` should be totally fine, right?  AFAIK, the `enum` keyword in C++ for types are optional, but are required in C.

```c++
enum Food {
  Apple,
  Cherry
};

const char* foodToString(enum Food food) {
  switch(food) {
    case Apple: return "Apple";
    case Cherry: return "Cherry";
  }
  return "??";
}
```

The above is valid C code, and valid C++ code.  If we removed the `enum` from `const char* foodToString(enum Food food)`, then it's no longer valid C code.

elBoberido: I'm not sure if there are some compiler quirks on some platforms. You can try it though.

mikebentley15: I( pushed a version that removes `ENUM` and renames `CLASS`.  Let's see if that passes.

elBoberido: It worked on all platforms with any compiler.

Once you squash a few commits, this can be merged.

mikebentley15: I merged all commits into one, and added the to the copyright headers of modified files.

Thanks!

elBoberido (approved):

elBoberido:
@mikebentley15 thanks very much for your contribution. I think I already mentioned it above, we are currently working on iceoryx2 which is written in Rust and will have C and C++ bindings with the next release. We will also present it in the iceoryx developer meetup next week on Tuesday. We would also be happy to hear what you are using iceoryx for :)

mikebentley15:
@elBoberido, you're welcome!  It's pretty cool that you have iceoryx2 implemented in Rust.  I'm using iceoryx indirectly via Cyclone DDS.  Cheers!

PR #2443, opened 2025-03-17
iox-#2440 Fix race condition that can cause failed chunk management a…

### LINKED ISSUE #2440: Race condition in Memory Manager leads to a nullptr contract violation
## Required information

**Operating system:**
(Have to double check)

**Compiler version:**
(Have to double check)

**Eclipse iceoryx version:**
Based off of a fork of main at commit hash 8a5e08348a8fe830cdb1d92bf5b299a3e4bc8282

Conditions still exist on the current main branch

In a production build, a segfault occurs. IN a debug build a contract check would fail when constructing an `iox::not_null` from a `nullptr` at this line: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/memory_manager.cpp#L203

The resulting warning log plus stack trace is:
```
[0;90m1971-01-01 00:02:25.971 [0;1;93m[Warn ][m: Mempool [m_chunkSize = 56, numberOfChunks = 66, used_chunks = 59 ] has no more space left
Writing crash info data to file: <redacted>
Crash Info is written successfully to a file

Caught signal 11!
signal number : 11 (SIGSEGV)
signal errno  : 0
signal code   : 1 (SEGV_MAPERR)
signal address: 0x0
error         : 0x0

Backtrace (see https://go/backtraces):
<redacted>(Unknown)[0xaaaaba23febc]
linux-vdso.so.1(__kernel_rt_sigreturn+0x0)[0xffffb6dd87d4]
<redacted>(iox::mepoo::ChunkManagement::ChunkManagement(iox::not_null<iox::mepoo::ChunkHeader*, void>, iox::not_null<iox::mepoo::MemPool*, void>, iox::not_null<iox::mepoo::MemPool*, void>)+0x174)[0xaaaaba1e0cfc]
<redacted>(iox::mepoo::MemoryManager::getChunk(iox::mepoo::ChunkSettings const&)+0x204)[0xaaaaba1de474]
<redacted>(iox::popo::ChunkSender<iox::popo::ChunkSenderData<8u, iox::popo::ChunkDistributorData<iox::DefaultChunkDistributorConfig, iox::popo::ThreadSafePolicy, iox::popo::ChunkQueuePusher<iox::popo::ChunkQueueData<iox::DefaultChunkQueueConfig, iox::popo::ThreadSafePolicy> > > > >::tryAllocate(iox::popo::UniquePortId, unsigned int, unsigned int, unsigned int, unsigned int)+0x1cc)[0xaaaaba1d17ec]
<redacted>(iox::popo::PublisherPortUser::tryAllocateChunk(unsigned int, unsigned int, unsigned int, unsigned int)+0x5c)[0xaaaaba1d1c9c]
<redacted ... call stack reaches application code>
...
```

Note here that the chunk size is 56, which is the size of a `ChunkManagement` chunk, just to confirm that a failed allocation happened in the chunk management pool and a nullptr was returned here: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/mem_pool.cpp#L94

What do you expect to happen?

I expect the failure to instead be, 100% of the time:
```
[0;90m1971-01-01 00:02:21.828 [0;1;93m[Warn ][m: Mempool [m_chunkSize = 320, numberOfChunks = 66, used_chunks = 59 ] has no more space left
[0;90m1971-01-01 00:02:21.828 [0;1;31;103m[Error][m: MemoryManager: unable to acquire a chunk with a chunk-payload size of 280The following mempools are available:  MemPool [ ChunkSize = 320, ChunkPayloadSize = 280, ChunkCount = 66 ]
[0;90m1971-01-01 00:02:21.828 [0;1;93m[Warn ][m: ICEORYX error! MEPOO__MEMPOOL_GETCHUNK_POOL_IS_RUNNING_OUT_OF_CHUNKS
```

We have a RouDi config with a segment containing a single memory pool, whose chunk count is **66**. Because there is only one memory pool, this also means that the chunk count for the chunk management pool is **66**: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/memory_manager.cpp#L74

We have two processes - one publishing on a channel configured to use that memory pool, and one which is subscribing to that channel.

The subscriber is configured to have a queue capacity of **64**. Note that, when the subscriber has taken a message and the publisher has taken a message, and the queue is full, the memory pool has exactly **0** chunks left available.

The subscriber processes messages periodically in a loop like so:
```
auto maybe_message = subscriber_.take();
while (maybe_message.has_value())
{
  do_stuff(maybe_message.value());
  maybe_message = subscriber_.take(); // Note that, very briefly, there are two chunks not in the queue
}
```

The publisher similarly publishes in a loop generalized as:
```
while (!incoming_data_queue_.empty())
{
  publisher_.loan().and_then([this]() { publisher_.publishCopyOf(incoming_data_queue_.pop()); });
}
```

Under these conditions, the following interleaving of events can happen:
1. The subscriber queue is full
2.  In the body of the loop, during the assignment operation the subscriber takes a message
3. The publisher publishes a message into the queue, making it full again. The memory pool now has 0 chunks free. The chunk management pool now has 0 chunks free
4. The subscriber _begins_ releasing the message being assigned to. It reaches this line of code: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/shared_chunk.cpp#L63 where it releases the **payload chunk** but has not yet released the **chunk management chunk**. There is now **1** payload chunk available but **0** chunk management chunks available
5. The publisher then attempts to loan another message. It reaches this point in the code where it successfully allocates a **chunk for the payload**: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/memory_manager.cpp#L159. 
6. The publisher then reaches this line of code where it attempts, and **fails** to allocate a chunk management chunk, here: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/memory_manager.cpp#L203
7. Because the allocation failed, the publisher is busy invoking the iox logger, and writing to the stream: https://github.com/eclipse-iceoryx/iceoryx/blob/7592b2adbf425d04a06d24571711e1a19a977d92/iceoryx_posh/source/mepoo/mem_pool.cpp#L91-L93. Arguments passed to the log macro are evaluated only when the stream operator for that argument is evaluated
8. Meanwhile, the subscriber continues draining the subscriber queue
9. By the time that `m_usedChunks` is loaded, 7 more messages have been drained from the queue, hence the weird error message saying 59/66 chunks are used, yet none are available
10. The publisher then proceeds to fail the contract check (or segfault) after the nullptr is returned

## Additional helpful information

I'm pretty sure this bug is solved by flipping the order of: https://github.com/eclipse-iceoryx/iceoryx/blob/2fc7a5fba165c3b9171ac57ab9ae1f43eb648ded/iceoryx_posh/source/mepoo/shared_chunk.cpp#L63-L64 so that the chunks are released in the opposite order that they are acquired. That way it becomes impossible for there ever to be more payload chunks available than management chunks.

gpalmer-latai:
…llocation

## Notes for Reviewer

As described in #2440, the race condition exists between the `MemoryManager` acquiring chunks for payloads before acquiring chunks for management, and the `SharedChunk` releasing chunks for payloads before releasing chunks for management.

The `MemoryManager` relies upon the assumption that if a payload allocation succeeded, the management allocation will succeed. This assumption proves to be false however for a brief moment when `SharedChunk::releaseChunk` releases the payload chunk but has not yet released the management chunk. 

The changes in this pull request resolve the issue by switching the order of release such that the management chunk is released before the payload chunk. This properly uploads the invariant that the number of management chunks is always greater than or equal to the number of payload chunks - since for any given payload/management pair the lifetime of the management chunk is contained within the lifetime of the payload chunk.

With regards to testing - there isn't any formal way to prove these changes resolve the issue nor prevent a new issue, other than running the stress tests.

1. [x] Update the PR title
   - Link to the relevant issue
1. [x] Assign PR to reviewer

[contributing]: https://github.com/eclipse-iceoryx/iceoryx/blob/main/CONTRIBUTING.md#coding-style
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/main/doc/website/concepts/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx/blob/main/doc/website/release-notes/iceoryx-unreleased.md

- [x] Consider a second reviewer for complex new features or larger refactorings

## References

- Closes #2440

### DIFF
--- a/iceoryx_posh/source/mepoo/memory_manager.cpp
+++ b/iceoryx_posh/source/mepoo/memory_manager.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2019 - 2020 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2020 - 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2025 by Latitude AI. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -204,6 +205,19 @@ expected<SharedChunk, MemoryManager::Error> MemoryManager::getChunk(const ChunkS
             ChunkManagement(chunkHeader, memPoolPointer, &m_chunkManagementPool.front());
         return ok(SharedChunk(chunkManagement));
     }
+}
+
+void MemoryManager::freeChunk(ChunkManagement& chunkManagement) noexcept
+{
+    const auto* chunkHeader = static_cast<void*>(chunkManagement.m_chunkHeader.get());
+    const auto mempool = chunkManagement.m_mempool;
+
+    // Here the chunk management must be freed before the chunk itself to maintain
+    // the invariant that there are always at least as many chunk management chunks available as payload chunks
+    chunkManagement.m_chunkManagementPool->freeChunk(&chunkManagement);
+    // NOTE: chunkManagement is a dangling reference from here on out
+
+    mempool->freeChunk(chunkHeader);
 }
 
 std::ostream& operator<<(std::ostream& stream, const MemoryManager::Error value) noexcept
--- a/iceoryx_posh/source/mepoo/shared_chunk.cpp
+++ b/iceoryx_posh/source/mepoo/shared_chunk.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2025 by Latitude AI. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,6 +17,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "iceoryx_posh/internal/mepoo/shared_chunk.hpp"
+#include "iceoryx_posh/internal/mepoo/memory_manager.hpp"
 
 namespace iox
 {
@@ -54,15 +56,9 @@ void SharedChunk::decrementReferenceCounter() noexcept
     if ((m_chunkManagement != nullptr)
         && (m_chunkManagement->m_referenceCounter.fetch_sub(1U, std::memory_order_relaxed) == 1U))
     {
-        freeChunk();
+        MemoryManager::freeChunk(*m_chunkManagement);
+        m_chunkManagement = nullptr;
     }
-}
-
-void SharedChunk::freeChunk() noexcept
-{
-    m_chunkManagement->m_mempool->freeChunk(static_cast<void*>(m_chunkManagement->m_chunkHeader.get()));
-    m_chunkManagement->m_chunkManagementPool->freeChunk(m_chunkManagement);
-    m_chunkManagement = nullptr;
 }
 
 SharedChunk& SharedChunk::operator=(const SharedChunk& rhs) noexcept
--- a/doc/website/release-notes/iceoryx-unreleased.md
+++ b/doc/website/release-notes/iceoryx-unreleased.md
@@ -156,6 +156,7 @@
 - Fix windows performance issue [#2377](https://github.com/eclipse-iceoryx/iceoryx/issues/2377)
 - Max Client and Server cannot be configured independently of Max Publisher and Subscriber [#2394](https://github.com/eclipse-iceoryx/iceoryx/issues/2394)
 - Fix call to non-existing `getService` in channel.inl [#2426](https://github.com/eclipse-iceoryx/iceoryx/issues/2426)
+- Fix chunk management race condition causing failed management allocation [#2440](https://github.com/eclipse-iceoryx/iceoryx/issues/2440)
 
 **Refactoring:**
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/memory_manager.hpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2019 - 2020 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2025 by Latitude AI. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -71,6 +72,10 @@ class MemoryManager
     /// @param[in] chunkSettings for the requested chunk
     /// @return a SharedChunk if successful, otherwise a MemoryManager::Error
     expected<SharedChunk, Error> getChunk(const ChunkSettings& chunkSettings) noexcept;
+
+    /// @brief Release a chunk back to the mempools
+    /// @param[in] chunkManagement Management for the chunk
+    static void freeChunk(ChunkManagement& chunkManagement) noexcept;
 
     uint32_t getNumberOfMemPools() const noexcept;
 
--- a/iceoryx_posh/include/iceoryx_posh/internal/mepoo/shared_chunk.hpp
+++ b/iceoryx_posh/include/iceoryx_posh/internal/mepoo/shared_chunk.hpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2019 by Robert Bosch GmbH. All rights reserved.
 // Copyright (c) 2021 - 2022 by Apex.AI Inc. All rights reserved.
+// Copyright (c) 2025 by Latitude AI. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -64,7 +65,6 @@ class SharedChunk
   private:
     void decrementReferenceCounter() noexcept;
     void incrementReferenceCounter() noexcept;
-    void freeChunk() noexcept;
 
   private:
     ChunkManagement* m_chunkManagement{nullptr};


iceoryx_posh/source/mepoo/shared_chunk.cpp
```diff
@@ -60,9 +61,15 @@ void SharedChunk::decrementReferenceCounter() noexcept
 
 void SharedChunk::freeChunk() noexcept
 {
-    m_chunkManagement->m_mempool->freeChunk(static_cast<void*>(m_chunkManagement->m_chunkHeader.get()));
+    const auto* chunkHeader = static_cast<void*>(m_chunkManagement->m_chunkHeader.get());
+    const auto mempool = m_chunkManagement->m_mempool;
+
+    // Here the chunk management must be freed before the chunk itself to maintain
+    // the invariant that there are always at least as many chunk management chunks available as payload chunks
```
elBoberido, line 68: Good catch. I'm wondering whether the freeing should be moved to the `MemoryManager`, right below the `getChunk` method. It could be a static function which takes a `ChunkManagement` pointer. The logic would be exactly like here, but the code would be much closer to the location where the chunk is allocated and therefore easier to reason about the de-allocation order.

What do you think?

elBoberido: Oh, and I guess writing a test for this bug is basically impossible due to its concurrent nature :/

gpalmer-latai: Having the management contained within a single class instead of implicitly between two different ones sounds like a good idea. I'll look into that. 

Yeah I mean I could invest a lot of time and effort into a stress test which has a publisher publishing slightly faster than a subscriber drains the queue, and that _might_ trigger the race condition over some period of time, but my bandwidth is somewhat limited at the moment to pursue this side quest.

gpalmer-latai: @elBoberido okay so it _might_ be possible in the `chunkManagement` to replace the pointer to the chunk management pool with a pointer to the memory manager, and then add a `freeChunk` call to the memory manager. Then the `freeChunk` method in `SharedChunk` would just call the `freeChunk` method in the memory manager, passing it a reference to the chunk management (after which it sets the chunk management to a nullptr). 

This will achieve the goal of keeping the freeing logic in the same class as the getting logic. The downside is now that the chunk management needs to hold a reference to the memory manager that created it (which will be annoying for updating unit tests). There is still a bit of weirdness around taking a mutable reference, invaliding the thing, and then relying on the caller to not use it further.

elBoberido: @gpalmer-latai I'm not sure if it would be worth the effort. It would also introduce the risk of new bugs. I think having the free function in the memory manager works equally good, without the need to refactor the chunk

gpalmer-latai: Ah wait, this might work because it is being included from the `shared_chunk.cpp`, not the `.hpp`. Let me try...

gpalmer-latai: @elBoberido take a look at the latest commit and let me know what you think. I opted for a static function rather than a free function so as to more closely mirror the `MemoryManager::getChunk`. 

Regarding unit tests - intuitively I _should_ add a unit test to the Memory Manager for the new static method. But it is a little bit annoying because `getChunk` returns a `SharedChunk` which holds the `ChunkManagement` pointer as a private member. Also the `FreeChunk` test in `SharedChunk` still basically tests this code path. 

I don't know - what do you think? Another option is to make the memory manager a friend of the `SharedChunk`. The static method could then take the `SharedChunk` directly and directly access the chunk management.

elBoberido: Yeah, looks already good.

I like the idea with passing the SharedChunk. It becomes symmetric with getChunk.

gpalmer-latai:
@elBoberido I noticed there is no protection around the `m_chunkManagement` pointer in `SharedChunk::freeChunk`. Would you like me to do anything like add an `IOX_ASSERT` to check that it is not null? I've documented the precondition to `freeChunk` that it has not already been freed. 

I can also change `freeChunk` to be a noop if it is called more than once, but perhaps that scenario itself is indicative of a logic error and it would be better to blow up.

iceoryx_posh/source/mepoo/shared_chunk.cpp
gpalmer-latai: Okay I just realized a problem with the whole "friend" approach. If I make `MemoryManager::freeChunk` accept a `SharedChunk`, then that operation is no longer private. I'd need to like, also make that static method private on the memory manager. Which means both the memory manager and free chunk need to be friends of one another? 

But if I wanted to do THAT most cleanly they should probably also only be friends via an attorney so they can only access the relevant data...

gpalmer-latai: Alright I did end up just making them friends of eachother. They are kind of deeply intertwined enough that it seems reasonable. 

I did not add a unit test because the memory manager already has a unit test which is freeing chunks so that code path is effectively still tested there.

gpalmer-latai:
@elBoberido ready for re-review

iceoryx_posh/source/mepoo/shared_chunk.cpp
elBoberido: Why do you want to have the `freeChunk` private? Maybe I overlook something, but I think it would not hurt to have it as public static function.

elBoberido: Forget about it. After looking at the code, I understand the reason. Ideally, `freeChunk` would get an r-value of a `SharedChunk` but since this needs to be called from the `SharedChunk` d'tor, this would look weird.

elBoberido:
@gpalmer-latai `freeChunk` is a private function and only called from the d'tor which does the sanity check.

iceoryx_posh/source/mepoo/shared_chunk.cpp
gpalmer-latai: Yeah I went back to using the chunk management and it looks a lot cleaner :)

iceoryx_posh/source/mepoo/memory_manager.cpp
```diff
@@ -206,6 +207,18 @@ expected<SharedChunk, MemoryManager::Error> MemoryManager::getChunk(const ChunkS
...
+void MemoryManager::freeChunk(SharedChunk& sharedChunk) noexcept
+{
+    const auto* chunkHeader = static_cast<void*>(sharedChunk.m_chunkManagement->m_chunkHeader.get());
+    const auto mempool = sharedChunk.m_chunkManagement->m_mempool;
+
+    // Here the chunk management must be freed before the chunk itself to maintain
+    // the invariant that there are always at least as many chunk management chunks available as payload chunks
+    sharedChunk.m_chunkManagement->m_chunkManagementPool->freeChunk(sharedChunk.m_chunkManagement);
+    sharedChunk.m_chunkManagement = nullptr;
+    mempool->freeChunk(chunkHeader);
```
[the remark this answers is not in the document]
gpalmer-latai: I used a reference instead which is even better.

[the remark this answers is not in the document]
elBoberido: It's kind of a double edged sword. The ideal solution would be to have a `not_null` with a different semantic. It should be possible to create one from a pointer without an assert if one already knows that the pointer is not null, e.g. `not_null::new_unchecked(T* inner)`. But that's for another PR :)

iceoryx_posh/source/mepoo/memory_manager.cpp
```diff
@@ -206,6 +207,17 @@ expected<SharedChunk, MemoryManager::Error> MemoryManager::getChunk(const ChunkS
...
 }
 
+void MemoryManager::freeChunk(ChunkManagement& chunkManagement) noexcept
+{
+    const auto* chunkHeader = static_cast<void*>(chunkManagement.m_chunkHeader.get());
+    const auto mempool = chunkManagement.m_mempool;
+
+    // Here the chunk management must be freed before the chunk itself to maintain
+    // the invariant that there are always at least as many chunk management chunks available as payload chunks
+    chunkManagement.m_chunkManagementPool->freeChunk(&chunkManagement);
```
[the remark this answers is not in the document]
elBoberido: yeah, but how shall we call it? riceoryx?

cc @elfenpiff a lost opportunity but it's maybe not too late

elBoberido (approved):

PR #2525, opened 2026-05-12
fix(posh): bound memcpy in VersionInfo::getCurrentVersion

MerinoSheep:

1. [x] Add a story to the PR title.
2. [x] Format the source code by calling `tools/scripts/clang_format.sh`. *(One-line code change; matches surrounding style.)*
3. [x] Have a look at the assigned labels of the PR (NA — first-time external contributor)
4. [x] Mention the test cases in the PR description.
5. [x] Update the PR title to start with `iox-#?` and end with task description that will be used as commit message. *(Used Conventional-Commit-style title; happy to rename to `iox-#???` after a maintainer assigns an issue number.)*
6. [x] Add presentation of new features in the **iceoryx tech talk**.
7. [x] All checks of GitHub Actions pass.
8. [x] If the PR introduces a new behavior or a new API, it is covered by the corresponding test cases.

## Notes for Reviewer

This PR fixes an AddressSanitizer global-buffer-overflow in
`iox::version::VersionInfo::getCurrentVersion()` that triggers any time iceoryx
is built from a release tarball (i.e. without a `.git` directory) with
`-fsanitize=address`.

### The bug

`iceoryx_posh/cmake/iceoryxversions.cmake` runs `git describe` to populate
`ICEORYX_SHA1`. When the source tree has no `.git` (release tarball, vendored
copy, distro package, Bazel `http_archive`, etc.), the command fails silently
and `ICEORYX_SHA1` is defined as the empty string literal `""` — that's 1 byte
in the binary, including the trailing NUL.

`getCurrentVersion()` then constructs a `CommitIdString_t` like this:

```cpp
CommitIdString_t shortCommitIdString(TruncateToCapacity,
                                     ICEORYX_SHA1,
                                     COMMIT_ID_STRING_SIZE /*=12*/);
```

The matching `cxx::string` `TruncateToCapacity` ctor in
`iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl` (line ~90) does
**not** check the actual length of the source pointer when an explicit `count`
argument is supplied — it `memcpy()`s `count` bytes unconditionally. So the
ctor reads 12 bytes from a 1-byte literal.

ASan reports the resulting overflow:

```
==X==ERROR: AddressSanitizer: global-buffer-overflow on address ...
    READ of size 12 at 0x... thread T0
      #0 __asan_memcpy
      #1 iox::cxx::string<12u>::string<...>(... ICEORYX_SHA1, 12)
            iceoryx_hoofs/include/iceoryx_hoofs/internal/cxx/string.inl:90
      #2 iox::version::VersionInfo::getCurrentVersion()
            iceoryx_posh/source/version/version_info.cpp:111
      #3 iox::roudi::RouDi setup / iox::PoshRuntime::PoshRuntime
0x... is located 0 bytes to the left of global variable ""...
    ICEORYX_SHA1 ... of size 1
```

### Why this matters in practice

This is a hard blocker for ASan-built **ROS 2** workspaces that use
`rmw_cyclonedds_cpp` + iceoryx. The first thing `iox::PoshRuntime` does at
startup is call `getCurrentVersion()` to compare its version info against
RouDi's. With ASan enabled, every node aborts inside `rmw_create_node()`
before user code ever runs. Without this fix, an entire AV / robotics stack
built with `--config=asan` cannot bring up a single node.

### The fix

Cap the byte count to the actual length of the `ICEORYX_SHA1` literal using
`strnlen`:

```cpp
CommitIdString_t shortCommitIdString(TruncateToCapacity,
                                     ICEORYX_SHA1,
                                     strnlen(ICEORYX_SHA1, COMMIT_ID_STRING_SIZE));
```

Real git builds (where `ICEORYX_SHA1` is the full 40-char SHA) are unaffected:
`strnlen` is bounded by `COMMIT_ID_STRING_SIZE`, so the truncation behavior is
identical for any source string of length ≥ 12.

A `#include <cstring>` is added near the existing `<algorithm>` include for
`strnlen`.

### Test cases

Two new tests are added to `iceoryx_posh/test/moduletests/test_version_info.cpp`:

1. **`GetCurrentVersionDoesNotReadPastEndOfShortCommitId`** — calls
   `VersionInfo::getCurrentVersion()`, exercising the same code path that
   previously aborted inside `iox::PoshRuntime` initialization. When the test
   binary is built with `-fsanitize=address` against a tarball checkout (no
   `.git/`), this test would have aborted before the fix; after the fix it
   passes and asserts the returned `VersionInfo` is valid.
2. **`CommitIdStringTruncateToCapacityHandlesShortSourceLiteral`** — pins the
   underlying invariant deterministically. Constructs `CommitIdString_t` with
   an empty source literal `""` and a `strnlen(src, COMMIT_ID_STRING_SIZE)`
   bounded count, and asserts the resulting size is `0`. This regression
   guards against the same shape of mistake (`count > strlen(src)`) being
   reintroduced in any callsite, regardless of how the build environment
   defines `ICEORYX_SHA1`.

### Tested by

Applied this exact change as a local patch on top of the iceoryx **v2.0.6**
release tarball (consumed via `rules_ros2` in a Bazel monorepo), then rebuilt
the ROS 2 stack with `-fsanitize=address`. Before the patch, every ASan-built
node aborted inside `rmw_create_node` with the global-buffer-overflow above.
After the patch, the same nodes initialize cleanly and the regular
(non-ASan) builds are byte-for-byte unchanged in behavior.

### DIFF
--- a/iceoryx_posh/source/version/version_info.cpp
+++ b/iceoryx_posh/source/version/version_info.cpp
@@ -17,6 +17,7 @@
 #include "iceoryx_posh/version/version_info.hpp"
 
 #include <algorithm>
+#include <cstring>
 
 namespace iox
 {
@@ -108,7 +109,7 @@ bool VersionInfo::isValid() noexcept
 VersionInfo VersionInfo::getCurrentVersion() noexcept
 {
     BuildDateString_t buildDateStringCxx(ICEORYX_BUILDDATE);
-    CommitIdString_t shortCommitIdString(TruncateToCapacity, ICEORYX_SHA1, COMMIT_ID_STRING_SIZE);
+    CommitIdString_t shortCommitIdString(TruncateToCapacity, ICEORYX_SHA1, strnlen(ICEORYX_SHA1, COMMIT_ID_STRING_SIZE));
 
     return VersionInfo(static_cast<uint16_t>(ICEORYX_VERSION_MAJOR),
                        static_cast<uint16_t>(ICEORYX_VERSION_MINOR),
--- a/iceoryx_posh/test/moduletests/test_version_info.cpp
+++ b/iceoryx_posh/test/moduletests/test_version_info.cpp
@@ -19,10 +19,15 @@
 #include "iceoryx_versions.hpp"
 #include "test.hpp"
 
+#include <cstring>
+
 namespace
 {
 using namespace ::testing;
 using namespace iox::version;
+using iox::TruncateToCapacity;
+using iox::COMMIT_ID_STRING_SIZE;
+using iox::CommitIdString_t;
 
 class VersionInfo_test : public Test
 {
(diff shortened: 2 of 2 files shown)


MerinoSheep:
Superseded by #2527. Closing in favor of the iox-#2526 branch which conforms to CONTRIBUTING.md naming conventions.

